PR #9, opened 2022-01-24
Adding a function to encrypt multiple matchkeys

benjaminsavage:
Created two structs - meant to encapsulate event reports, with a new function on the User struct to generate a report, and a function on the Helper struct to partially decrypt a report. For now reports only consist of a set of encrypted match keys. We will add the secret shares later.

### DIFF
--- a/raw-ipa-lib/src/helpers/event.rs
+++ b/raw-ipa-lib/src/helpers/event.rs
@@ -161,7 +161,7 @@ mod tests {
         assert_eq!(
             partially_decrypted_1
                 .matchkeys()
-                .count_matches(&partially_decrypted_2.matchkeys()),
+                .count_matches(partially_decrypted_2.matchkeys()),
             0
         );
 
@@ -174,7 +174,7 @@ mod tests {
         assert_eq!(
             decrypted_1
                 .matchkeys()
-                .count_matches(&decrypted_2.matchkeys()),
+                .count_matches(decrypted_2.matchkeys()),
             1
         );
     }
--- a/raw-ipa-lib/src/helpers/event.rs
+++ b/raw-ipa-lib/src/helpers/event.rs
@@ -1,17 +1,13 @@
-#[cfg(feature = "enable-serde")]
-use std::collections::HashMap;
 use crate::error::{Error, Res};
 use crate::helpers::Helpers;
+use crate::report::{DecryptedEventReport, EventReport};
 use crate::threshold::DecryptionKey as ThresholdDecryptionKey;
-use crate::user::DecryptedEventReport;
-use crate::user::EventReport;
 use rand::thread_rng;
-use rand::rngs::StdRng;
-use rand::SeedableRng;
-use rust_elgamal::{Ciphertext, EncryptionKey, RistrettoPoint, Scalar, GENERATOR_POINT};
-pub use rust_elgamal::{Ciphertext, EncryptionKey, RistrettoPoint, Scalar, GENERATOR_POINT};
+use rust_elgamal::EncryptionKey;
 #[cfg(feature = "enable-serde")]
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "enable-serde")]
+use std::collections::HashMap;
 #[cfg(feature = "enable-serde")]
 use std::fs;
 use std::ops::{Deref, DerefMut};
@@ -85,17 +81,25 @@ impl Helper {
     }
 
     pub fn threshold_decrypt_event(&self, r: EventReport) -> EventReport {
-        let partially_decrypted_matchkeys: HashMap<_, _> = r.encrypted_match_keys.iter().map(
-            |(p, emk)| (p.to_string(), self.matchkey_decrypt.threshold_decrypt(*emk))
-        ).collect();
-        EventReport{encrypted_match_keys: partially_decrypted_matchkeys}
+        let partially_decrypted_matchkeys: HashMap<_, _> = r
+            .encrypted_match_keys
+            .iter()
+            .map(|(p, emk)| (p.to_string(), self.matchkey_decrypt.threshold_decrypt(*emk)))
+            .collect();
+        EventReport {
+            encrypted_match_keys: partially_decrypted_matchkeys,
+        }
     }
 
     pub fn decrypt_event(&self, r: EventReport) -> DecryptedEventReport {
-        let partially_decrypted_matchkeys: HashMap<_, _> = r.encrypted_match_keys.iter().map(
-            |(p, emk)| (p.to_string(), self.matchkey_decrypt.decrypt(*emk))
-        ).collect();
-        DecryptedEventReport{decrypted_match_keys: partially_decrypted_matchkeys}
+        let partially_decrypted_matchkeys: HashMap<_, _> = r
+            .encrypted_match_keys
+            .iter()
+            .map(|(p, emk)| (p.to_string(), self.matchkey_decrypt.decrypt(*emk)))
+            .collect();
+        DecryptedEventReport {
+            decrypted_match_keys: partially_decrypted_matchkeys,
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


raw-ipa-lib/src/user.rs
```diff
@@ -253,4 +276,60 @@ mod tests {
...
+
+        assert_ne!(r1.encrypted_match_keys.get(PROVIDER_3), r2.encrypted_match_keys.get(PROVIDER_3));
+
+        let fully_decrypted_r1: HashMap<_, _> = r1.encrypted_match_keys.iter().map(
+            |(p, emk)| (p.to_string(), d2.decrypt(d1.threshold_decrypt(*emk)))
+        ).collect();
+
+        let fully_decrypted_r2: HashMap<_, _> = r2.encrypted_match_keys.iter().map(
+            |(p, emk)| (p.to_string(), d2.decrypt(d1.threshold_decrypt(*emk)))
+        ).collect();
```
martinthomson, line 323: It seems like you want a real object to hold this result.  That would allow for the logic to be held somewhere (other than tests) as well.  That object can have a PartialEq implementation that returns true if any matchkey matches (no Eq implementation though, as that doesn't make sense for that type).  Then all you would need to do is take the encrypted matchkeys, decrypt into this new object, and compare.  (More extensive checking is probably useful to be sure that everything is working, of course; see other comments.)

raw-ipa-lib/src/helpers/event.rs
```diff
@@ -92,3 +109,88 @@ impl DerefMut for Helper {
...
+        let tek = ThresholdEncryptionKey::new(&[h_se.public.matchkey_encrypt, h_te.public.matchkey_encrypt]);
+
+        const PROVIDER_1: &str = "social.example";
+        const PROVIDER_2: &str = "news.example";
+        const PROVIDER_3: &str = "email.example";
+        const PROVIDER_4: &str = "game.example";
+
+        const MATCHING_MATCHKEY: &str = "12345678";
+
+        let mut u123 = User::new(123, tek);
```
martinthomson, line 138: Given that the user ID doesn't matter, maybe just make it
```suggestion
        let mut u1 = User::new(1, tek);
        let mut u2 = User::new(2, tek);
```

That might silence at least *some* of clippy's complaints here.

raw-ipa-lib/src/helpers/event.rs
```diff
@@ -92,3 +109,88 @@ impl DerefMut for Helper {
...
+        u123.set_matchkey(PROVIDER_4, "45678901");
+
+        let mut u234 = User::new(234, tek);
+        u234.set_matchkey(PROVIDER_1, MATCHING_MATCHKEY);
+        u234.set_matchkey(PROVIDER_2, "does_not_match");
+        u234.set_matchkey(PROVIDER_3, "something_else");
+
+        let r123 = u123.generate_event_report(&[PROVIDER_1, PROVIDER_2, PROVIDER_3, PROVIDER_4]);
+        let r234 = u234.generate_event_report(&[PROVIDER_1, PROVIDER_2, PROVIDER_3]);
+
```
martinthomson, line 151: Some checks for equality here are probably useful.  Maybe you can just create a helper function for doing that.

raw-ipa-lib/src/helpers/event.rs
```diff
@@ -92,3 +109,88 @@ impl DerefMut for Helper {
...
+            partially_decrypted_234.encrypted_match_keys.get(PROVIDER_2),
+        );
+        assert_ne!(
+            partially_decrypted_123.encrypted_match_keys.get(PROVIDER_3),
+            partially_decrypted_234.encrypted_match_keys.get(PROVIDER_3),
+        );
+        assert_ne!(
+            partially_decrypted_123.encrypted_match_keys.get(PROVIDER_4),
+            partially_decrypted_234.encrypted_match_keys.get(PROVIDER_4),
+        );
```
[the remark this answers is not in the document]
benjaminsavage: I like that helper function - that'll definitely come in handy in future tests.

raw-ipa-lib/src/helpers/event.rs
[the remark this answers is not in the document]
benjaminsavage: Makes sense.

raw-ipa-lib/src/user.rs
benjaminsavage: That's pretty neat! I like this idea.

raw-ipa-lib/src/helpers/event.rs
benjaminsavage: ok - will do.

martinthomson (commented):
Looking good.  A few places where we can probably improve, but I'd be happy to take this.

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn n_matches<T>(
+    a: impl Iterator<Item = impl PartialEq<T>>,
+    b: impl Iterator<Item = T> + Clone,
+) -> usize {
+    let b = b.into_iter();
+    a.into_iter()
+        .map(|x| b.clone().filter(|y| x.eq(y)).count())
+        .sum()
+}
```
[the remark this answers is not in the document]
martinthomson: try again without the same mistakes as before: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=af9cdaf9e3fd287a5dacf6529ef6b99b

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
...
+        let decrypted_matchkeys = self
+            .match_keys
+            .iter()
+            .map(|(p, emk)| (p.to_string(), matchkey_decrypt.decrypt(*emk)))
+            .collect();
+        DecryptedMatchkeys::from_matchkeys(decrypted_matchkeys)
+    }
+}
+
+impl fmt::Debug for EncryptedMatchkeys {
```
martinthomson, line 52: You should be able to just add `#[derive(Debug)]` to the struct declaration.  It's a little more noise, but less code to write :)

raw-ipa-lib/src/user.rs
```diff
@@ -253,4 +265,42 @@ mod tests {
             User::point_from_matchkey(OTHER_PROVIDER, MATCHKEY.as_bytes())
         );
     }
+
+    #[test]
+    fn test_generate_event_report() {
```
martinthomson, line 270:
```suggestion
    fn generate_event_report() {
```

I'm not a fan of having all the functions in the "tests" module start with "test_" :)

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl DecryptedMatchkeys {
+    pub fn count_matches(&self, other: &Self) -> usize {
+        n_matches(self.match_keys.values(), &other.match_keys.values())
+    }
+
+    pub fn from_matchkeys(match_keys: HashMap<String, RistrettoPoint>) -> DecryptedMatchkeys {
+        DecryptedMatchkeys { match_keys }
+    }
```
martinthomson, line 77: You should be able to do this:

```rust
impl From<HashMap<String, RistrettoPoint>> for DecryptedMatchKeys {
  fn from(match_keys: HashMap<String, RistrettoPoint>) -> Self {
    Self { match_keys }
  }
}
```
which you can then invoke with `DecryptedMatchkeys::from(x)`.  That fits the idioms better.

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl PartialEq for DecryptedMatchkeys {
+    fn eq(&self, other: &Self) -> bool {
+        any_matches(self.match_keys.values(), &other.match_keys.values())
+    }
+}
+
+impl DecryptedMatchkeys {
+    pub fn count_matches(&self, other: &Self) -> usize {
```
martinthomson, line 70: This looks like a function that you might only want to run for testing purposes.  You can move the implementation to the test module (the test module is able to access private fields), or you could do this:

```suggestion
    #[cfg(test)]
    pub fn count_matches(&self, other: &Self) -> usize {
```

I might prefer this in the test code, as then you don't have to deal with dead_code warnings and the like.

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
+use crate::threshold::DecryptionKey as ThresholdDecryptionKey;
+use crate::threshold::{Ciphertext, RistrettoPoint};
+
+use std::collections::HashMap;
+use std::fmt;
+
+pub struct EncryptedMatchkeys {
```
martinthomson, line 7:
```suggestion
#[derive(Debug)]
pub struct EncryptedMatchkeys {
```

See below.

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
...
+    match_keys: HashMap<String, Ciphertext>,
+}
+
+impl PartialEq for EncryptedMatchkeys {
+    fn eq(&self, other: &Self) -> bool {
+        any_matches(self.match_keys.values(), &other.match_keys.values())
+    }
+}
```
martinthomson, line 15: Is there any case in which we would need to compare encrypted matchkeys?  Maybe it makes sense to inspect for test code, but does this need to be something we expose to users of this code?

raw-ipa-lib/src/report.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl EncryptedMatchkeys {
+    pub fn count_matches(&self, other: &Self) -> usize {
+        n_matches(self.match_keys.values(), &other.match_keys.values())
+    }
+
+    pub fn from_matchkeys(match_keys: HashMap<String, Ciphertext>) -> EncryptedMatchkeys {
```
martinthomson, line 24: See comment about `From`.

martinthomson (approved):
Looking good!

I probably shouldn't apply the usual standards to code review here, but I have a few nits left over.  Nothing serious.  Decide what you want to address and then we should just merge this.  I'm going to want to start using this now.

PR #10, opened 2022-01-27
Aggregation

martinthomson:
This adds a key pair and a `sum()` function to the aggregation helpers.

Builds on #8.

### DIFF
--- a/raw-ipa-lib/src/helpers/aggregation.rs
+++ b/raw-ipa-lib/src/helpers/aggregation.rs
@@ -160,7 +160,7 @@ mod tests {
         let sum2 = helper2.sum(encrypted_shares2.iter().map(unref_share));
 
         let total = sum1 + sum2;
-        assert_eq!(total.value(), u128::from(expected_total));
+        assert_eq!(u128::from(total), u128::from(expected_total));
     }
 
     #[test]
@@ -192,7 +192,7 @@ mod tests {
             // The source or trigger helpers perform this on each pair of shares:
             // Add a random offset to the shares and then re-randomize the shared secret.
             // The two values are then sent the two aggregation helpers.
-            let offset = AdditiveShare::from(rng.next_u64());
+            let offset = AdditiveShare::<64>::try_from(u128::from(rng.next_u64())).unwrap();
             encrypted_shares1.push((
                 share1 + offset,
                 secret1.rerandomize(helper1.share_public_key(), &mut rng),
@@ -207,6 +207,6 @@ mod tests {
         let sum2 = helper2.sum(encrypted_shares2.iter().map(unref_share));
 
         let total = sum1 + sum2;
-        assert_eq!(total.value(), u128::from(expected_total));
+        assert_eq!(u128::from(total), u128::from(expected_total));
     }
 }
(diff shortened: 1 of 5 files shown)


raw-ipa-lib/src/helpers/aggregation.rs
```diff
@@ -73,3 +104,112 @@ impl DerefMut for Helper {
...
+        // The helper then sends the shuffled result to its peer, which would add its own offset.
+        // Then the shares would be separated and sent to the aggregation helpers for decryption.
+        // This runs the whole process at once, with a single offset and no shuffle.
+        let offset = AdditiveShare::from(rng.next_u64());
+        let encrypted_shares1: Vec<_> = shares1
+            .into_iter()
+            .map(|share| {
+                let (mut encryptor, secret) = helper1.share_public_key().encryptor(&mut rng);
+                (
+                    encryptor.encrypt(share) + offset,
```
benjaminsavage, line 193: Ah, I see. I hadn't realized the helper adds the same random offset to all of the encrypted secret shares.

This seems like something Helper 2 could feasibly undo and use to un-shuffle the array.

If a colluding agent tells Helper 2 that "Ben has encrypted secret share 456" and "Martin has encrypted secret share 567", then if Helper 1 adds a random value, say 789 to both of these encrypted secret shares, they'll preserve the same difference between them. Helper 2 can still look for a pair of encrypted secret shares where A - B == (567 - 456) == 111.

benjaminsavage: ...but I think it can use a different offset for each event right? Because it has the pair of encrypted secret shares together at once? So perhaps that's what you're trying to say in the comment above, that this isn't exactly how it'll really look?

If that's the case - I'd suggest just making the test look more like how we think things will eventually look.

raw-ipa-lib/src/helpers/aggregation.rs
```diff
@@ -59,6 +80,16 @@ impl Helper {
...
         Ok(())
     }
+
+    pub fn sum<'item, const N: u32>(
+        self,
+        shares: impl IntoIterator<Item = (AdditiveShare<N>, &'item EncryptedSecret)>,
+    ) -> AdditiveShare<N> {
+        shares
+            .into_iter()
+            .map(|(share, secret)| self.share_decryption.decryptor(secret).decrypt(share))
```
benjaminsavage, line 90: I'm trying, but I just can't get used to this use of the term "secret". I associate this word with "secret key" and think of it as a value that cannot be shared. To call a value which is distributed to others a "secret" just just really confusing to me. Maybe I'm alone in this, but I'd prefer to name it something else. I should probably have left this feedback on the last pull request. I think that in Diffie-Helman the "shared-secret" is something both parties can ultimately derive that nobody else can see - and doesn't refer to the value derived from the public key that's being passed around in the clear. I'm not sure what to call that part though.

benjaminsavage (changes requested):
A few comments

raw-ipa-lib/src/helpers/aggregation.rs
martinthomson: Yes, the idea was to use a different offset for every value.  I was having trouble with share mutability of the RNG and left it outside, but then I changed it.

raw-ipa-lib/src/helpers/aggregation.rs
martinthomson: Yeah, I was not happy with this either.

martinthomson:
Abandoning this also.  OBE

PR #11, opened 2022-01-27
Reorganize

martinthomson:
This arrangement should be a lot cleaner to work with.

I was just copying what we did on another project with the first iteration of this, but it doesn't really make sense for something like this.


martinthomson:
@btsavage this might be a good first change if we want to get secret shares in.

Structure is that we have a single library (I'll fix the name up) with a couple of binaries in it as follows:

```
./src/cli/hexarg.rs
./src/cli/mod.rs
./src/cli/helper.rs
./src/cli/stringn.rs
./src/cli/verbosity.rs
./src/user.rs
./src/error.rs
./src/helpers/mod.rs
./src/helpers/event.rs
./src/helpers/aggregation.rs
./src/report.rs
./src/lib.rs
./src/bin/helper.rs
./src/bin/ua.rs
./src/threshold.rs
```

There is just a single crate now, which cuts down the code size a lot.  Same amount of code.

PR #13, opened 2022-03-08
implement "Efficient Signatures on Randomizable Ciphertexts"

thurstonsand:
per [this white paper](https://eprint.iacr.org/2020/524.pdf), we are trying to determine the performance characteristics of randomizable ciphertexts and their signatures. This PR will be an exploration of the algorithm's performance at scale.


martinthomson:
Just a little feedback:

Use tests to test that your code is working correctly rather than a main function.

With #11, which I just merged, you might want to consider just adding a new .rs file.  That's the conflict here.

More feedback when you have something more substantial.

thurstonsand:
thanks, I merged master, and went ahead and implemented the rest of the sign + verify logic

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+    G1Affine::from(dec)
+}
+
+#[allow(dead_code)]
+fn randomize(cm: &CipherText, enc_key: &EncryptionKey, r_prime: Scalar) -> CipherText {
+    let CipherText(c0, c1) = cm;
+    let rc0 = c0 + (G1Affine::generator() * r_prime);
+    let rc1 = c1 + (enc_key * r_prime);
+    CipherText(rc0, rc1)
+}
```
eriktaubeneck, line 81: FWIW, "prime" used with cryptography typically signals a prime number. I know the paper uses `r'`, which I would read as "r prime", but it might help to just call this `r_new`.
```suggestion
#[allow(dead_code)]
fn randomize(cm: &CipherText, enc_key: &EncryptionKey, r_new: Scalar) -> CipherText {
    let CipherText(c0, c1) = cm;
    let rc0 = c0 + (G1Affine::generator() * r_new);
    let rc1 = c1 + (enc_key * r_new);
    CipherText(rc0, rc1)
}
```

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+    let SigningKey(x0, x1) = sign_key;
+    let z = (G1Affine::generator() + (c0 * x0) + (c1 * x1)) * scalar_s_inverted;
+    let s = G1Affine::generator() * scalar_s;
+    let s_hat = G2Affine::generator() * scalar_s;
+    let t = ((G1Affine::generator() * x0) + (enc_key * x1)) * scalar_s_inverted;
+    Signature(z, s, s_hat, t)
+}
+
+#[allow(dead_code)]
+fn adapt(
```
eriktaubeneck, line 102: "adapt" is a less common feature in crypto, so it would be helpful to add a doc string that says something to the effect of "when rerandomizing a cipher text, the original signature breaks. this function adapts the signature to validate with the new cipher text."

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+    rng: &mut (impl RngCore + CryptoRng),
+    signature: &Signature,
+    r_prime: Scalar,
+) -> Signature {
+    let scalar_s_prime = Scalar::from(rng.next_u64());
+    let scalar_s_prime_inverted = scalar_s_prime.invert().unwrap(); // TODO: how to handle?
```
eriktaubeneck, line 108: My understanding is that it should be incredibly rare for this to be non-invertible, so I think it makes sense to handle this in this function, since it's basically just a need to retry. I think we could cap the retries at a reasonable level (say 10 or even 100) so that it does fail if there happens to be a bug in the provided `rng`.

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+    let s_hat = G2Affine::generator() * scalar_s;
+    let t = ((G1Affine::generator() * x0) + (enc_key * x1)) * scalar_s_inverted;
+    Signature(z, s, s_hat, t)
+}
+
+#[allow(dead_code)]
+fn adapt(
+    rng: &mut (impl RngCore + CryptoRng),
+    signature: &Signature,
+    r_prime: Scalar,
```
eriktaubeneck, line 105: same comment as above, it probably makes sense to use something like "new" instead of "prime". we could even note that in the doc string (and it would probably be a good idea to link to the paper in a docstring for the module itself.)

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+    let CipherText(c0, c1) = cm;
+    let VerifyKey(x_hat0, x_hat1) = verify_key;
+
+    let z_s_pair = pairing(&G1Affine::from(z), &G2Affine::from(s_hat));
+    let g_g_hat_pair = pairing(&G1Affine::generator(), &G2Affine::generator());
+    let c0_x_hat0_pair = pairing(&G1Affine::from(c0), &G2Affine::from(x_hat0));
+    let c1_x_hat1_pair = pairing(&G1Affine::from(c1), &G2Affine::from(x_hat1));
+    if z_s_pair != g_g_hat_pair + c0_x_hat0_pair + c1_x_hat1_pair {
+        return false;
+    }
```
eriktaubeneck, line 135:
```suggestion
    let z_s_hat_pair = pairing(&G1Affine::from(z), &G2Affine::from(s_hat));
    let g_g_hat_pair = pairing(&G1Affine::generator(), &G2Affine::generator());
    let c0_x_hat0_pair = pairing(&G1Affine::from(c0), &G2Affine::from(x_hat0));
    let c1_x_hat1_pair = pairing(&G1Affine::from(c1), &G2Affine::from(x_hat1));
    if z_s_hat_pair != g_g_hat_pair + c0_x_hat0_pair + c1_x_hat1_pair {
        return false;
    }
```

eriktaubeneck: Given that we're using `s_hat`, this seems a little more in line with the naming scheme.

src/esrc.rs
```diff
@@ -0,0 +1,195 @@
...
+) -> bool {
+    let Signature(z, s, s_hat, t) = signature;
+    let CipherText(c0, c1) = cm;
+    let VerifyKey(x_hat0, x_hat1) = verify_key;
+
+    let z_s_pair = pairing(&G1Affine::from(z), &G2Affine::from(s_hat));
+    let g_g_hat_pair = pairing(&G1Affine::generator(), &G2Affine::generator());
+    let c0_x_hat0_pair = pairing(&G1Affine::from(c0), &G2Affine::from(x_hat0));
+    let c1_x_hat1_pair = pairing(&G1Affine::from(c1), &G2Affine::from(x_hat1));
+    if z_s_pair != g_g_hat_pair + c0_x_hat0_pair + c1_x_hat1_pair {
```
eriktaubeneck, line 133: Is addition the right operator for pairings? (Just double checking. The paper just has them next to each other, so I'd assume that's multiplication, but it might just be the default operator, and that could be addition in this case.)

eriktaubeneck (approved):
A few comments and small suggestions, but overall this LGTM. I would still want @martinthomson to review this.

src/esrc.rs
thurstonsand: added a couple comments clarifying the meaning here

src/esrc.rs
thurstonsand: done

src/esrc.rs
thurstonsand: yep, done

thurstonsand (commented):
thanks for the review. I believe @eriktaubeneck mentioned I should get someone else to weigh in on the `r_new` vs `r_prime` comments

thurstonsand:
I also moved all the functions into a new `Esrc` struct, so that they were struct methods. Let me know which style you prefer

thurstonsand:
@martinthomson @benjaminsavage do we think this code will ever be used, or should I just close this?

benjaminsavage:
I think we can safely close this now that our updated IPA proposal doesn't involve this.

PR #14, opened 2022-03-10
This module implements privacy budget using a redis cache

richajaindce:
It expects a vector of encrypted blobs alongwith budget to be consumed.
After increasing the budget consumed by all the blobs, it returns blobs which are below TOTAL_PRIVACY_BUDGET

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,7 +2,7 @@ name: CI
 on: [push, pull_request]
 env:
   CARGO_TERM_COLOR: always
-
+  REDIS_VERSION: '6.2.4'
 jobs:
   check:
     name: All Checks
@@ -20,6 +20,14 @@ jobs:
         with:
           command: build
           args: --tests
+
+      - name: Install redis and start redis server 
+        run: |
+          sudo apt-get update
+          wget https://github.com/redis/redis/archive/${{ env.REDIS_VERSION }}.tar.gz;
+          tar -xzvf ${{ env.REDIS_VERSION }}.tar.gz;
+          pushd redis-${{ env.REDIS_VERSION }} && BUILD_TLS=yes make && sudo mv src/redis-server src/redis-cli /usr/bin/ && popd;
+          redis-server &
 
       - uses: actions-rs/cargo@v1
         name: Run Tests
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ hkdf = "0.11"
 log = "0.4"
 rand_core = "0.6"
 rand = "0.8"
+redis = "0.21.5"
 rust-elgamal = "0.4"
 serde = {version = "1.0", optional = true}
 serde_json = {version = "1.0", optional = true}
@@ -31,6 +32,7 @@ structopt = {version = "0.3", optional = true}
 
 [dev-dependencies]
 hex = "0.4"
+lazy_static = "1.4.0"
 
 [lib]
 name = "raw_ipa"
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,6 +6,7 @@ pub enum Error {
     InvalidRole,
     NotEnoughHelpers,
     NotFound,
+    RedisError(redis::RedisError),
     TooManyHelpers,
 
     #[cfg(feature = "cli")]
@@ -49,6 +50,7 @@ forward_errors! {
     std::io::Error => Io,
     #[cfg(feature = "enable-serde")]
     serde_json::Error => Serde,
+    redis::RedisError => RedisError,
 }
 
 pub type Res<T> = Result<T, Error>;
--- a/src/helpers/privacy_budget.rs
+++ b/src/helpers/privacy_budget.rs
@@ -1,0 +1,230 @@
+#![allow(dead_code)]
+
+use crate::error::Res;
+use redis::{transaction, Commands, RedisWrite, ToRedisArgs};
+use rust_elgamal::RistrettoPoint;
+#[derive(Debug, PartialEq, Eq)]
+struct BlindedMatchKey(RistrettoPoint);
+
+impl BlindedMatchKey {
+    pub fn new(key: RistrettoPoint) -> Self {
+        Self(key)
+    }
+}
+
+impl ToRedisArgs for BlindedMatchKey {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        // let (c0, c1) = self.0.inner();
+        out.write_arg(self.0.compress().as_bytes());
+    }
+}
+
+/// This module implements privacy budget using a redis cache
+struct PrivacyBudget {
+    redis_connection: redis::Connection,
+    total_privacy_budget: u32,
+}
+
+impl PrivacyBudget {
+    fn new(redis_host_name: &str, redis_password: &str, is_tls: bool) -> Res<PrivacyBudget> {
+        //if Redis server needs secure connection
+        let uri_scheme = if is_tls { "rediss" } else { "redis" };
+
+        let redis_conn_url = format!("{}://:{}@{}", uri_scheme, redis_password, redis_host_name);
+
+        let redis_connection = redis::Client::open(redis_conn_url)?.get_connection()?;
+
+        Ok(PrivacyBudget {
+            redis_connection,
+            total_privacy_budget: 100,
+        })
+    }
+
+    /// Input : a vector of blinded match keys alongwith budget to be consumed.
+    /// This atomically increases the budget of all of the blinded match keys which have
+    /// sufficient budget to consume. For those which do not pass, match keys are discarded and not passed to the output.
+    pub fn consume_budget_and_return_for_eligible_bmks<'a>(
+        &'a mut self,
+        blinded_match_keys: impl Iterator<Item = &'a BlindedMatchKey>,
+        budget_to_consume: u32,
+    ) -> impl Iterator<Item = &'a BlindedMatchKey> {
+        blinded_match_keys.filter(move |blinded_match_key| {
+            match self.increase_budget_consumed(blinded_match_key, budget_to_consume) {
+                Ok(true) => true,
+                Ok(false) | Err(_) => false,
+            }
+        })
+    }
+
+    pub fn get_total_privacy_budget(&self) -> u32 {
+        self.total_privacy_budget
+    }
+
+    /// Budget is increased for each key atomically in Redis
+    fn increase_budget_consumed(
+        &mut self,
+        blinded_match_key: &BlindedMatchKey,
+        budget_to_consume: u32,
+    ) -> Res<bool> {
+        let total_privacy_budget = self.get_total_privacy_budget();
+
+        Ok(transaction(
+            &mut self.redis_connection,
+            &[blinded_match_key],
+            |con, pipe| {
+                let val: u32 = con.get(&blinded_match_key).unwrap_or(0);
+                if val + budget_to_consume > total_privacy_budget {
+                    pipe.set(&blinded_match_key, val)
+                        .ignore()
+                        .query::<()>(con)?;
+                    Ok(Some(false))
+                } else {
+                    // increment
+                    pipe.set(&blinded_match_key, val + budget_to_consume)
+                        .ignore()
+                        .query::<()>(con)?;
+                    Ok(Some(true))
+                }
+            },
+        )?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use lazy_static::lazy_static;
+    use rand::thread_rng;
+    use std::sync::Mutex;
+
+    lazy_static! {
+        static ref PRIVACY_BUDGET: Mutex<PrivacyBudget> =
+            Mutex::new(PrivacyBudget::new("127.0.0.1", "", false).unwrap());
+    }
+
+    impl Drop for BlindedMatchKey {
+        fn drop(&mut self) {
+            clear_redis_key(self);
+        }
+    }
+
+    fn clear_redis_key(blinded_match_key: &BlindedMatchKey) {
+        let mut privacy_budget = PRIVACY_BUDGET.lock().unwrap();
+        privacy_budget
+            .redis_connection
+            .del::<&BlindedMatchKey, ()>(blinded_match_key)
+            .expect("Failed to delete keys");
+    }
+
+    fn create_blinded_match_keys() -> [BlindedMatchKey; 4] {
+        let mut rng = thread_rng();
+
+        let c1 = BlindedMatchKey::new(RistrettoPoint::random(&mut rng));
+        let c2 = BlindedMatchKey::new(RistrettoPoint::random(&mut rng));
+        let c3 = BlindedMatchKey::new(RistrettoPoint::random(&mut rng));
+        let c4 = BlindedMatchKey::new(RistrettoPoint::random(&mut rng));
+
+        [c1, c2, c3, c4]
+    }
+
+    #[test]
+    fn privacy_budget_didnot_exceed() {
+        let mut privacy_budget = PRIVACY_BUDGET.lock().unwrap();
+        let blinded_match_keys = create_blinded_match_keys();
+        let mut eligible_bmk = privacy_budget
+            .consume_budget_and_return_for_eligible_bmks(blinded_match_keys.iter(), 20);
+        let ok0 = eligible_bmk.next() == Some(&blinded_match_keys[0]);
+        let ok1 = eligible_bmk.next() == Some(&blinded_match_keys[1]);
+        let ok2 = eligible_bmk.next() == Some(&blinded_match_keys[2]);
+        let ok3 = eligible_bmk.next() == Some(&blinded_match_keys[3]);
+
+        drop(eligible_bmk);
+        drop(privacy_budget);
+
+        assert!(ok0 && ok1 && ok2 && ok3);
+    }
+
+    #[test]
+    fn privacy_budget_exceed() {
+        let mut privacy_budget = PRIVACY_BUDGET.lock().unwrap();
+        let blinded_match_keys = create_blinded_match_keys();
+        let mut eligible_bmk = privacy_budget
+            .consume_budget_and_return_for_eligible_bmks(blinded_match_keys.iter(), 120);
+
+        let ok = eligible_bmk.next() == None;
+
+        drop(eligible_bmk);
+        drop(privacy_budget);
+        assert!(ok);
+    }
+
+    #[test]
+    fn privacy_budget_multiple_runs_exceed() {
+        let mut privacy_budget = PRIVACY_BUDGET.lock().unwrap();
+        let blinded_match_keys = create_blinded_match_keys();
+        let mut eligible_bmk = privacy_budget
+            .consume_budget_and_return_for_eligible_bmks(blinded_match_keys.iter(), 20);
+
+        let ok0 = eligible_bmk.next() == Some(&blinded_match_keys[0]);
+        let ok1 = eligible_bmk.next() == Some(&blinded_match_keys[1]);
+        let ok2 = eligible_bmk.next() == Some(&blinded_match_keys[2]);
+        let ok3 = eligible_bmk.next() == Some(&blinded_match_keys[3]);
+
+        drop(eligible_bmk);
+        drop(privacy_budget);
+
+        assert!(ok0 && ok1 && ok2 && ok3);
+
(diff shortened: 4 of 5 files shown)


src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,80 @@
...
+    //format - host:port
+    let redis_host_name =
+        env::var("REDIS_HOSTNAME").expect("missing environment variable REDIS_HOSTNAME");
+    let redis_password = env::var("REDIS_PASSWORD").unwrap_or_default();
+
+    //if Redis server needs secure connection
+    let uri_scheme = match env::var("IS_TLS") {
+        Ok(_) => "rediss",
+        Err(_) => "redis",
+    };
```
martinthomson, line 50: Please don't use the environment for these things.  Pass them in as arguments.  Let the caller decide if they want to read the environment.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,80 @@
+/// This module implements privacy budget using a redis cache
+/// It expects a vector of encrypted blobs alongwith budget to be consumed.
+/// After increasing the budget consumed by all the blobs, it returns blobs which are below ``TOTAL_PRIVACY_BUDGET``
+
+use redis::Commands;
+use std::env;
+
+const TOTAL_PRIVACY_BUDGET: i32 = 100;
```
martinthomson, line 8: This might be better as a property on a struct (see other comments).  It also needs to be public  (with a getter if it is a property) so that the caller can know how much budget is available.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,80 @@
...
+            .expect("Failed to GET budget for packet");
+
+        if budget_consumed < TOTAL_PRIVACY_BUDGET {
+            eligible_packets.push(packet.to_string());
+        }
+    }
+    eligible_packets
+}
+
+fn connect() -> redis::Connection {
```
martinthomson, line 40: This should produce an object (a new struct) that hosts the above functions.  Then the connection can be held open by the caller until they are done.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,80 @@
...
+/// It expects a vector of encrypted blobs alongwith budget to be consumed.
+/// After increasing the budget consumed by all the blobs, it returns blobs which are below ``TOTAL_PRIVACY_BUDGET``
+
+use redis::Commands;
+use std::env;
+
+const TOTAL_PRIVACY_BUDGET: i32 = 100;
+
+#[must_use]
+pub fn get_eligible_packets_with_budget(packets: &[String], budget: i32) -> Vec<String> {
```
martinthomson, line 11: Taking packets as a slice of strings is likely to be quite inconvenient.  What this probably needs is a new trait that produces a sequence of bytes that is used as a key to the database.

Ideally, this takes an iterator and produces an iterator so that this acts as filtering function on that iterator.  Alternatively, if you wanted to be really smart, this could produce a closure that you would pass to `Iterator::filter`.

martinthomson (commented):
Thanks for getting started on this.  This will be a useful function to have.

In addition to the structural comments, can you also look at adding more thorough testing of the budget.  For instance, if I use 1 unit when the budget is at 99, what happens?  What if I ask for 2 units then?  Or 200 units?

src/helpers/privacy_budget.rs
richajaindce: Since we need to keep a track of both eligible(to return) and ineligible packets (to later decrement their budget), chose to keep it as a normal iterator. Also have changed the input type of Ciphertext since we would be expecting match keys to be sent to Privacy Budget module

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,202 @@
...
+            let _: () = privacy_budget
+                .redis_connection
+                .del(packet.to_string())
+                .expect("Failed to delete keys");
+        }
+    }
+
+    fn create_privacy_budget_ciphertexts() -> (PrivacyBudget, [Ciphertext; 4]) {
+        let redis_host_name =
+            env::var("REDIS_HOSTNAME").expect("missing environment variable REDIS_HOSTNAME");
```
martinthomson, line 155: This is where things get tedious.  It's not ideal if someone who runs the tool has to start their own redis server, get it into the expected state, and set environment variables.

What these tests need to do is start the redis server on their own.  I know that's a massive burden, because it involves a bunch of tricky setup (I suggest using the lazy_static crate for executing the database).  The `redis` crate has [a good example of how you might set that up](https://github.com/mitsuhiko/redis-rs/blob/master/tests/support/mod.rs) (you don't need all that complexity, so that can be pared back fairly aggressively).

Then your only requirement for running the tests is that you have the redis server installed.

martinthomson (changes requested):

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    pub fn get_total_privacy_budget(&self) -> i32 {
+        return self.total_privacy_budget;
+    }
+
+    /// Budget is increased for each key atomically in Redis
+    fn increase_budget_consumed(&mut self, packets: &[Ciphertext], budget: i32) {
+        for packet in packets.iter() {
+            let _: () = self
```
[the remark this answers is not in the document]
richajaindce: Do we want to return the budget consumed by the match keys from this function? Isn't that a potential information leak?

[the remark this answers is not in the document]
martinthomson: I don't think that we need to return anything (we provide the amount of budget consumed as an input).

What I was looking for from the redis API was what we discussed when we met.  That is, an atomic add operation returns either the outcome of the addition, or the value before the addition.  Here, we get no return value, so we don't learn whether the addition overflowed.

If this is the case, then what we probably want to do is read then conditionally add instead.  That's two operations that we would need to run anyway, but if we conditionally add then we don't spend time on adding and subtracting for those match keys that exceed the budget.

src/helpers/privacy_budget.rs
richajaindce: Have implemented a setup and teardown  service for each test which starts and stops the redis server. However, while running tests, the tests keep passing and failing randomly in different runs. 
Also tried to run all the tests in a single thread but that didnt help alleviate the issue.
Let me know your thoughts

martinthomson:
Were you creating a single server for each test?  That's probably OK as long as you also ensure that you can run multiple servers at once (which means a completely new server instance for each, with different ports and files) as Rust will run tests in parallel.

The other way to manage that is to run a single server using something like the lazy_static crate; for that you could have each test hit a different part of the database that won't interact with others.

martinthomson:
The only way I know of addressing that is to modify the .github/workflows/check.yml file to include a line that installs the redis server package.

richajaindce:
Have added the redis server in the workflow. So no need to manage it from our tests now :) Let me know if the diff looks ok

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    pub fn get_total_privacy_budget(&self) -> u32 {
+        self.total_privacy_budget
+    }
+
+    /// Budget is increased for each key atomically in Redis
+    fn increase_budget_consumed(&mut self, packet: &BlindedMatchKey, budget: u32) -> Res<bool> {
+        let key = packet.to_string();
+        let (budget_consumed,): (u32,) =
+            transaction(&mut self.redis_connection, &[key.clone()], |con, pipe| {
```
martinthomson, line 104: Read, write, read seems inefficient.  Can't you just bail if the value you read exceeds the budget?  Also, given that the first read is outside of the atomic transaction, aren't you exposed to concurrent updates?

(I don't know redis well, so feel free to tell me I'm wrong.)

src/lib.rs
```diff
@@ -5,3 +5,7 @@ pub mod helpers;
 pub mod report;
 pub mod threshold;
 pub mod user;
+
+#[allow(unused_imports)]
+#[macro_use]
+extern crate lazy_static;
```
martinthomson, line 11: You should be able to use use lazy_static in the test code and access the macro that way.

Cargo.toml
```diff
@@ -17,9 +17,11 @@ digest = "0.9"
 hex = {version = "0.4", optional = true}
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, so pin this
 hkdf = "0.11"
+lazy_static = "1.4.0"
```
martinthomson, line 20: If this is only for tests, add it to the dev-dependencies block.

src/error.rs
```diff
@@ -6,6 +6,7 @@ pub enum Error {
     InvalidRole,
     NotEnoughHelpers,
     NotFound,
+    RedisError,
```
martinthomson, line 9: Add this to the forward_errors macro below as well.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+
+#[allow(dead_code)]
+impl BlindedMatchKey {
+    pub fn new(key: DecryptionKey, message: RistrettoPoint, scalar: Scalar) -> Self {
+        let c = key.encryption_key().encrypt_with(message, scalar);
+        Self(c)
+    }
+}
+
+impl ToString for BlindedMatchKey {
```
martinthomson, line 19: I think that you want to implement ToRedisArgs instead.  Then you don't need hex encoding.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+        c0_hex.push_str(&c1_hex);
+        c0_hex
+    }
+}
+
+#[allow(dead_code)]
+impl PrivacyBudget {
+    fn new(
+        redis_host_name: &str,
+        redis_password: Option<String>,
```
martinthomson, line 52: &str would seem to be sufficient for password (if you have no password, then the empty string can be provided).

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+            redis_connection,
+            total_privacy_budget: 100,
+        })
+    }
+
+    /// Input : a vector of ciphertexts alongwith budget to be consumed.
+    /// This atomically increases the budget of all of the blinded match keys that are received.
+    /// For those that end up exceeding the budget, the budget is then decremented again,
+    /// but the corresponding match key is discarded and not passed to the output.
+    pub fn get_eligible_packets_with_budget(
```
martinthomson, line 77: I'm still not particularly happy with this interface.

A consumer of this API might want to feed data in via an iterator (rather than having you call .iter() on a slice).  And they don't necessarily want to have clones of their inputs provided to them.  Returning an iterator that includes a subset of the inputs is closer to what the caller will need.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.redis_connection.decr(packet.to_string(), budget)?;
+        Ok(())
+    }
+}
+
+impl From<RedisError> for Error {
+    fn from(_e: RedisError) -> Self {
+        Error::RedisError
+    }
+}
```
martinthomson, line 127: The macro in lib.rs does this for you.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+        let packets = create_ciphertexts();
+        clear_redis_keys(&mut *privacy_budget, &packets);
+        packets
+    }
+
+    #[test]
+    fn privacy_budget_didnot_exceed() {
+        let mut privacy_budget = PRIVACY_BUDGET.lock().unwrap();
+        let packets = create_privacy_budget_ciphertexts(&mut *privacy_budget);
+        let eligible_packets = privacy_budget.get_eligible_packets_with_budget(&packets, 20);
```
martinthomson, line 180: The 20 here is relative to a constant.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand::thread_rng;
+    use std::sync::Mutex;
+
+    lazy_static! {
+        static ref PRIVACY_BUDGET: Mutex<PrivacyBudget> =
+            Mutex::new(PrivacyBudget::new("127.0.0.1", None, false).unwrap());
+    }
```
martinthomson, line 138: This is basically OK, but you need to ensure that you cleanup properly or this database will accumulate a lot of old values.  Not a problem in CI, but local runs that fail will leave stuff behind.  Implementing Drop on the test keys you use might help with that.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,217 @@
...
+        let c3 = BlindedMatchKey::new(k1, mk3, r);
+        let c4 = BlindedMatchKey::new(k1, mk4, r);
+
+        [c1, c2, c3, c4]
+    }
+
+    fn clear_redis_keys(privacy_budget: &mut PrivacyBudget, packets: &[BlindedMatchKey]) {
+        for packet in packets.iter() {
+            #![allow(clippy::let_unit_value)]
+            let _: () = privacy_budget
```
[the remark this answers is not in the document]
richajaindce: Actually this is required by the compiler to interpret type. If removed, this gives an error "cannot infer type for type parameter `RV` declared on the associated function `del`"

src/helpers/privacy_budget.rs
richajaindce: It would expose it but I thought we were ok with doing sequential queries for now (based on our discussions in person)?

src/helpers/privacy_budget.rs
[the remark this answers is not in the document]
martinthomson: I think that you can use a turbofish instead.  In this case, I'm guesssing that del() needs the extra type parameters, so `del::<RedisResult<()>>(packet.to_string())` perhaps.

src/helpers/privacy_budget.rs
martinthomson: yes, but if we don't care about concurrency, then why use a transaction?

src/helpers/privacy_budget.rs
richajaindce: 20 here is an absolute budget which was passed to be consumed for all the match keys in the query. Is the understanding correct?

.github/workflows/check.yml
```diff
@@ -15,10 +15,13 @@ jobs:
...
-        uses: supercharge/redis-github-action@1.4.0
-        with:
-          redis-version: 6
+     - name: Install redis and start redis server
+      run: |
+        sudo apt-get update
+        wget https://github.com/redis/redis/archive/${{ env.REDIS_VERSION }}.tar.gz;
+        tar -xzvf ${{ env.REDIS_VERSION }}.tar.gz;
+        pushd redis-${{ env.REDIS_VERSION }} && BUILD_TLS=yes make && sudo mv src/redis-server src/redis-cli /usr/bin/ && popd;
+        redis-server
```
martinthomson, line 24: This probably won't work as the processing won't move on to the next step until redis-server exits (which presumably never happens).

src/helpers/privacy_budget.rs
```diff
@@ -74,23 +63,20 @@ impl PrivacyBudget {
...
+    ) -> Vec<&'a BlindedMatchKey> {
+        packets
+            .filter(
+                |packet| match self.increase_budget_consumed(packet, budget) {
+                    Ok(true) => true,
+                    Ok(false) => false,
+                    Err(_) => false,
+                },
+            )
+            .collect()
```
martinthomson, line 79: You shouldn't need to collect, though it means changing the return value to `impl Iterator<Item = &'a BlindedMatchKey` as well.

src/helpers/privacy_budget.rs
```diff
@@ -100,43 +86,46 @@ impl PrivacyBudget {
...
+            |con, pipe| {
+                let val: u32 = con.get(&key).unwrap_or(0);
+                if val + budget > total_privacy_budget {
+                    pipe.set(&key, val).ignore().query::<()>(con)?;
+                    Ok(Some(false))
+                } else {
+                    // increment
+                    pipe.set(&key, val + budget)
+                        .ignore()
+                        .query::<()>(con)?;
```
martinthomson, line 103: So this still fails if there are concurrent executions, so it might be better to use incr() rather than set() here.  That won't stop a race from happening that results in allowing a request that should be blocked, but it will at least ensure that we don't lose the budget that has been spent.

Note that the transaction handling in the redis crate is extraordinarily loose if you are used to ACID.

I don't know how to stop this from happening except to have this fail closed instead with incr, get, check, but that requires a rollback if the increment blows the budget.

.github/workflows/check.yml
richajaindce: Will put it in a background so that it keeps running.

src/helpers/privacy_budget.rs
richajaindce: 1. Redis doesnt support rollbacks so that is out. 
2 .A transaction would watch over the key for any change and will abort in case someone else changes the value. In that case, we would receive a "false" and discard that packet from being considered while also not burning the budget. --> this should take care of concurrency issues
3. However, if some error happens while executing the transaction, the budget might be burned while we still discard the packet --> Is the concern on this?

richajaindce: Based on https://redis.io/docs/manual/transactions/#watchcommandswatch-explained looks like WATCH would not modify the budget consumed for a key in case of any concurrency issues. Due to this, going to keep this implementation

src/helpers/privacy_budget.rs
richajaindce: Since privacy budget is a mutex, Drop was giving an issue due to over-locking. Have dropped the lock before each assert so the destructor is able to clear the redis keys

src/helpers/privacy_budget.rs
richajaindce: While making the function return an iterator, there are some considerations. Since iterator holds on to the &mut self which is passed on to it, we would need to restrict the lifetime of the packets to that of privacy budget object. Is that an acceptable outcome?

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,231 @@
...
+            |con, pipe| {
+                let val: u32 = con.get(&packet).unwrap_or(0);
+                if val + budget > total_privacy_budget {
+                    pipe.set(&packet, val).ignore().query::<()>(con)?;
+                    Ok(Some(false))
+                } else {
+                    // increment
+                    pipe.set(&packet, val + budget).ignore().query::<()>(con)?;
+                    Ok(Some(true))
+                }
```
benjaminsavage, line 84: This is not how I remember us discussing this working.

I remember us saying that we would start by just atomically incrementing all of the keys by the desired amount, then we would look at the return values to see which had exceeded the privacy budget, and subsequently decrementing those. That approach is more safe in a concurrent environment.

This approach isn't using atomic increments - it's first calling "con.get" to read the current value, then calling pipe.set(val + budget). That's not an atomic increment. If another query runs between those two lines, the value could be read at the pre-update value, then the new value overwritten, thereby allowing a query to exceed the privacy budget.

src/helpers/privacy_budget.rs
```diff
@@ -0,0 +1,231 @@
...
+        let mk1 = RistrettoPoint::random(&mut rng);
+        let mk2 = RistrettoPoint::random(&mut rng);
+        let mk3 = RistrettoPoint::random(&mut rng);
+        let mk4 = RistrettoPoint::random(&mut rng);
+        let r = Scalar::random(&mut rng);
+
+        let c1 = BlindedMatchKey::new(k1, mk1, r);
+        let c2 = BlindedMatchKey::new(k1, mk2, r);
+        let c3 = BlindedMatchKey::new(k1, mk3, r);
+        let c4 = BlindedMatchKey::new(k1, mk4, r);
```
benjaminsavage, line 129: Yeah, this is not how I think about Blinded Match Keys. I do not think they are CipherText, they are not encrypted.

benjaminsavage (changes requested):
Left a few comments.

src/helpers/privacy_budget.rs
richajaindce: Given that this is done in a transaction, the keys will be watched, increment will happen in memory and will be changed only if they are not changed by anyone else.
If the values are observed to be changed, redis will error out. 
This approach makes the entire operation concurrency safe as compared to increment and decrement

benjaminsavage: That is so cool. Thanks for explaining this to me!

benjaminsavage (approved):
Thanks for writing this code Richa!

PR #15, opened 2022-03-17
Test fallback_matchkey

taikiy:
A quick and simple commit and pull request test, adding a couple of lines in .gitignore.

Also testing `fallback_matchkey()` during `generate_event_report()` if a match key has not been generated for the user for a particular provider.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 *~
 /Cargo.lock
 /.vscode/
-/db
+*.code-workspace
+*/db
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
generally, I would prefer that each test cover just one piece of functionality, but this is OK

taikiy:
Thanks Martin! I'll separate this test case out if I add more tests in the future.

I see a warning message "First-time contributors need a maintainer to approve running workflows.". Seems I need you to push a button before I can push?

PR #16, opened 2022-03-24
Implement secure mul between 3 parties

thurstonsand:


Cargo.toml
```diff
@@ -19,6 +19,7 @@ hkdf = "0.11"
 log = "0.4"
 rand_core = "0.6"
 rand = "0.8"
+rug = "1.15.0"
```
thurstonsand, line 22: this dependency uses gmpz under the hood, but includes the gmp source code along with the rust bindings, so this is all that is needed to use it.

according to docs, "[Rug also depends on the GMP, MPFR and MPC libraries through the low-level FFI bindings in the gmp-mpfr-sys crate, which needs some setup to build](https://docs.rs/rug/latest/rug/#using-rug)".

that setup is straightforward, eg on linux you need to have `diffutils`, `gcc`, `m4` and `make` installed. on macos, you just need to run `xcode-select --install`.

thurstonsand: gmp (including rug) can have a significant perf improvement over the pure-rust [num_bigint](https://docs.rs/num-bigint/latest/num_bigint/).

according to [benchmarks here](https://github.com/vks/discrete-log), which solve for discrete log, gmp can be 6-8x faster than pure rust implementation

src/secure_mul/mod.rs
```diff
@@ -0,0 +1,15 @@
+use rug::Integer;
+
+pub mod clear_text;
+
+pub trait H {
```
thurstonsand, line 5: I'd like to get a better overall picture of the algorithm being designed here, so I can better figure out the best abstractions to be used. This is my best interpretation of what we will need, but could need adjusting, and as of now I don't have all the context I need to decide one way or another

PR #19, opened 2022-04-01
[Stacked Diff] Thread pool object for concurrent connections handling

taikiy:
This is a stacked branch of my previous PR: [Thread module to be used for server connection handler #18](https://github.com/martinthomson/raw-ipa/pull/18)
New commits for this pull request begin at [886ac46](https://github.com/martinthomson/raw-ipa/commit/886ac46cd482407befd3e7e78a7e5162997b6cf6)

This diff implements `Pool` which consists of a vector of `Thread`s. Now, a single thread object is a size(1) `Pool`.

### DIFF
--- a/src/net/server.rs
+++ b/src/net/server.rs
@@ -1,19 +1,26 @@
-use crate::net::Thread;
+use crate::net::Pool;
+use log::warn;
 
 pub struct Server {
-    connection_handler_thread: Thread,
+    connection_handler_thread: Pool,
 }
 
 impl Server {
     #[must_use]
     pub fn new() -> Server {
         Server {
-            connection_handler_thread: Thread::new(),
+            connection_handler_thread: Pool::new(1),
         }
     }
 
     pub fn start(&self) {
         self.start_connection_handler();
+    }
+
+    pub fn stop(&mut self) {
+        if let Err(e) = self.connection_handler_thread.shutdown() {
+            warn!("Graceful shutdown failed: {}", e);
+        }
     }
 
     /// Spawns a new thread to handle incoming connections.
@@ -42,7 +49,10 @@ mod tests {
 
     #[test]
     fn no_panic() {
-        let server = Server::new();
+        let mut server = Server::new();
+
         server.start();
+
+        server.stop();
     }
 }
--- a/src/net/thread.rs
+++ b/src/net/thread.rs
@@ -5,10 +5,6 @@ use std::thread;
 
 use crate::error::Res;
 
-pub struct Thread {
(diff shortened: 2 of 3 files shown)


src/net/thread.rs
```diff
@@ -54,11 +81,19 @@ impl Default for Thread {
...
+                break;
+            }
+        }
+
+        for worker in &mut self.workers {
+            info!("Terminating thread {:?}", worker);
+            if let Some(thread) = worker.thread.take() {
                 thread.join().unwrap();
             }
         }
```
martinthomson, line 99: Thinking more about this, it might be worth breaking this down a little more.

What you are looking for here is a cleanup function that ensures that you don't get zombie threads.  For that, sending the Terminate instruction is probably sufficient.  The reason that you want to call join() is to ensure that you get a clean shutdown.  Just sending N terminate instructions won't ensure that each of the threads has time to finish open tasks, so a join is necessary there.

So my suggestion would be to have the Drop implementation only send the Terminate instruction.  Then, if the Pool controller is dropped (by accident, I hope), the threads are not left dangling.  But if you want a graceful shutdown, then you need the join.

So I think you have two functions: terminate() and shutdown().  The first sends the terminate message to all the workers without waiting for them to come back.  The second calls terminate() and then waits for all of the threads to end.  drop() only really needs to do the first.  Then you don't have a call that potentially blocks for a long time in a drop implementation.

src/net/thread.rs
```diff
@@ -24,13 +28,36 @@ pub enum Message {
...
+    ///  If the thread has been terminated.
+    pub fn execute<F>(&self, f: F) -> Res<()>
+    where
+        F: FnOnce() + Send + 'static,
+    {
+        self.thread.execute(f)?;
+
+        Ok(())
+    }
+}
```
martinthomson, line 47: There's a nasty trick you can use here...  You can implement Deref on Thread and have it return Pool.  Then any functions on the pool are automatically available when you hold a reference to Thread.

That said, I think that you can probably just delete Thread.  `Pool::new(1)` is shorter to type than `Thread::new()`

src/net/server.rs
```diff
@@ -0,0 +1,58 @@
+use crate::net::Pool;
+use log::warn;
+
+pub struct Server {
+    connection_handler_thread: Pool,
+}
+
+impl Server {
```
benjaminsavage, line 8: IPAServer

benjaminsavage (approved):
This is a good start =). I'm sure we will evolve this considerably over time.

PR #21, opened 2022-04-11
Three Party Secret Share Shuffling

linanw:
Based on IPA Technical Writeup April 25th

Implemented Three Party Hide and Seek Secret Share Shuffling [Semi-Honest]
sub modules include:
- Secure Coin Tossing
- Rejection Sampling
- Bit XOR Secret Share

**Test plan: **
cargo test -- --nocapture test_three_party_hide_and_seek_shuffling

**Known Issue: ** 
1. Event batch size limited to 16, as generation of a permutation by Rejecting Sampling currently uses a bit string of an u128 integer as the random factor, which maximal to generate a 16-dimensions permutation. It's possible to extend to use multiple u128 integer to generate longer bit string as the random factor. I will do it in coming commit.
2. Related to the issue above. Because that in the process of Rejecting Sampling, there potentially founds duplicated numbers and it makes the same length of bit strings generates inconsistent number of distinct integers. We will need to have extra bits to ensure required number of integers been generated. I'm working on a solution.

### DIFF
--- a/src/ss.rs
+++ b/src/ss.rs
@@ -43,5 +43,7 @@ mod tests {
         // recover from secret shares
         let recovered_integer_matchkey = recover_from_shares(sh1.share, sh2.share, sh3.share);
         println!("recovered_integer_matchkey = {}.", recovered_integer_matchkey);
+
+        assert_eq!(integer_matchkey, recovered_integer_matchkey);
     }
 }
--- a/src/ss.rs
+++ b/src/ss.rs
@@ -1,0 +1,47 @@
+use rand::Rng;
+
+
+pub struct BitXORShare {
+    share: u32,
+}
+
+impl BitXORShare{
+    #[must_use]
+    pub fn generate(value: u32) -> (Self, Self, Self) {
+        let mut rng = rand::thread_rng();
+    
+        let sh1 = rng.gen::<u32>();
+        let sh2 = rng.gen::<u32>();
+        let sh3 = value ^ sh1 ^ sh2 ;
+    
+        (Self {share: sh1}, Self {share: sh2}, Self {share: sh3},)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{BitXORShare};
+    use rand::Rng;
+
+    fn recover_from_shares(share1: u32, share2: u32, share3: u32) -> u32 {
+        share1 ^ share2 ^ share3
+    }
+
+    #[allow(clippy::similar_names)] // deal with it.
+    #[test]
+    fn generate_shares_and_recover() {
+
+        // generate a random unsigned 32bit integer as matchkey for test.
+        let mut rng = rand::thread_rng();
+        let integer_matchkey = rng.gen::<u32>();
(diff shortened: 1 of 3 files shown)


src/ss.rs
```diff
@@ -0,0 +1,330 @@
...
+pub struct RejectionSampling {}
+
+impl RejectionSampling {
+    pub fn generate_permutation(bit_string_as_int: u128, dimension: usize) -> Vec<usize> {
+        // Check if input dimension can be supported.
+        // bit_string_as_int is 128 bit. It only enough to generate 4 bit dimension.
+        // E.g. if demension is 30, we needs 5 bits to generate 1 integer number
+        // between 1 to 30, and we need at least 30 group of 5 bits to generate
+        // 20 number to meet the demension. 5 bits times 30 is bigger 128 bit.
+        let max_dimension = 2usize.pow(4);
```
martinthomson, line 80: use `const`

src/ss.rs
```diff
@@ -0,0 +1,330 @@
...
+        if dimension > max_dimension {
+            panic!(
+                "dimension ({}) > {} is not supported!",
+                dimension, max_dimension
+            );
+        }
+
+        // calculate how many bit we need to generate an integer number in dimension.
+        // ceiling of log2(n), n is the number of dimensions.
+        let bit = ((dimension) as f64).log2().ceil() as u8;
```
martinthomson, line 90: This calculation is problematic as the use of floating point leads to inaccuracies.  It's also slower than it needs to be.  You probably want to look at u128::leading_zeros (and size_of) for this.

src/ss.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    pub fn calculate_shared_random(r1: u128, r2: u128) -> u128 {
+        r1 ^ r2
+    }
+}
+
+pub struct RejectionSampling {}
+
+impl RejectionSampling {
+    pub fn generate_permutation(bit_string_as_int: u128, dimension: usize) -> Vec<usize> {
```
martinthomson, line 74: Your source of entropy is a little scant here.  You probably want `&[u8]` rather than `u128` for that.  Or `rand::Rng`.  Note that your use of rejection sampling will ensure that you will run out of bits if you provide a fixed-length input of any size; you really want an RNG (or really a PRF with a counter that can be extended arbitrarily).

src/ss.rs
```diff
@@ -0,0 +1,330 @@
...
+impl SecureCoinTossing {
+    pub fn get_random_and_hash() -> (u128, String) {
+        // generate a u128 random integer.
+        let mut rng = rand::thread_rng();
+        let r = rng.gen::<u128>();
+
+        // generate hash for the random integer's bit string.
+        let mut hasher = Sha3_256::new();
+        hasher.update(format!("{:b}", r));
+        let h: String = format!("{:X}", hasher.finalize());
```
martinthomson, line 48: Please don't use strings unless you need a string.

martinthomson (changes requested):
I didn't go through the shuffle part as I think there are a few fundamentals that need to be improved here first.

PR #20, opened 2022-04-01
[Stacked Diff] Implement basic networking for stream IO over TCP

taikiy:
This is a stacked branch of my previous PR: [[Stacked Diff] Thread pool object for concurrent connections handling #19
](https://github.com/martinthomson/raw-ipa/pull/19)
New commits for this pull request begin at [dfad8b8](https://github.com/martinthomson/raw-ipa/commit/dfad8b8218909876c440f9b6515aecbab163f56d)

This commit adds a connection handler and basic networking IO.

1. `Server` spawns a new thread (`ConnectionHandler`) to create a `TcpListener` and a pool thread
2. For each accepted connection, `ConnectionHandler` will send the stream to any available thread
3. Each thread will handle the actual IO operation. Once the operation is completed, the thread returns to idle state

For testing purpose, I created `Echo` server command, which just returns whatever the server received.

PR #22, opened 2022-04-11
[Stacked Diff] Implement a client module to send commands to a network server

taikiy:
This is a stacked branch of my previous PR [[Stacked Diff] Implement basic networking for stream IO over TCP #20
](https://github.com/martinthomson/raw-ipa/pull/20)

New commits for this branch begin at [ee4190b](https://github.com/martinthomson/raw-ipa/commit/ee4190b9d4142632b8f0dd18ecc9b7b6d4660763)

This commit adds a simple TCPStream client which enables sending a `Command` to a specified server and receiving the result from it. Client is pretty much the same as what I have in server.rs unit tests, but refactored into its own file.

### DIFF
--- a/src/net/client.rs
+++ b/src/net/client.rs
@@ -1,0 +1,98 @@
+use crate::net::data::Command;
+#[cfg(feature = "enable-serde")]
+use std::io::{Read, Result as IoResult, Write};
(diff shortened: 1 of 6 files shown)


benjaminsavage (approved):
LGMT.

PR #24, opened 2022-04-18
[PRF Compute] Adding function to compute multiplicative inverse of a …

richajaindce:
…number i.e. given m and n it computes t such that (m * t) % n = 1

This copies Extended Euclidean Algorithm and takes care of errors (https://github.com/facebookresearch/Private-ID/blob/main/crypto/src/gcd.rs)

### DIFF
--- a/src/helpers/prf_compute/multiplicative_inverse.rs
+++ b/src/helpers/prf_compute/multiplicative_inverse.rs
@@ -1,0 +1,122 @@
+use num_bigint::{BigInt, Sign::Plus};
+use num_integer::Integer;
+use num_traits::{identities::Zero, Signed};
+use rust_elgamal::Scalar;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ScalarType(Scalar);
+
+impl From<&ScalarType> for BigInt {
+    fn from(value: &ScalarType) -> Self {
+        BigInt::from_bytes_le(Plus, value.0.as_bytes())
+    }
+}
+
+impl From<&BigInt> for ScalarType {
+    fn from(value: &BigInt) -> Self {
+        let (_, mut bytes) = (value).to_bytes_le();
+        while bytes.len() < 32 {
+            bytes.push(0_u8);
+        }
+        let ans = bytes.try_into().unwrap();
+        ScalarType(Scalar::from_bits(ans))
+    }
+}
+
+/// Implementation of Extended Euclidean Algorithm to compute multiplicative inverse of a number
+/// i.e given num and modulus it computes t such that (num * t) % modulus = 1
+///
+/// How it computes :
+/// Finds Bezout Coefficient a and b for coprimes num and modulus i.e.
+/// ``bezout_coefficient_a`` * ``num`` + ``bezout_coefficient_b`` * ``modulus`` = 1
+/// This deduces to ( ``bezout_coefficient_a`` * ``num`` ) % ``modulus`` = 1 % ``modulus``
+/// Hence ``bezout_coefficient_a`` % ``modulus`` is the multiplicative inverse of ``num``
+/// In the code, ``bezout_coefficient_a`` is represented by ``t_current``
+/// Before returning number is converted into a positive number
+///
+/// # Panics
+/// This will assert in case input numbers are zero or num is not smaller than modulus
+/// if the function does not work properly and returns an unexpected bezout coefficient
+
+#[must_use]
+pub fn multiplicative_inverse<'a>(num: &'a ScalarType, modulus: &'a ScalarType) -> ScalarType {
+    // Assume non zero
+    assert!(num.0 != Scalar::zero());
+    assert!(modulus.0 != Scalar::zero());
+    let (num, modulus) = (BigInt::from(num), BigInt::from(modulus));
+
+    assert!(num < modulus);
+
+    let (mut remainder_current, mut remainder_last) = (num, modulus.clone());
+    let (mut t_last, mut t_current) = (BigInt::from(0_i32), BigInt::from(1_i32));
+
+    loop {
+        let (quotient, remainder_new) = remainder_last.div_rem(&remainder_current);
+        if remainder_new.is_zero() {
+            break;
+        }
+        let t_new = &t_last - quotient * t_current.clone();
+        remainder_last = remainder_current;
+        remainder_current = remainder_new;
+        t_last = t_current;
+        t_current = t_new;
+    }
+
+    if t_current.is_negative() {
+        t_current += modulus.clone();
+        assert!(t_current.is_positive());
+    }
+    ScalarType::from(&(t_current % modulus))
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::helpers::prf_compute::multiplicative_inverse::ScalarType;
+
+    use super::{multiplicative_inverse, BigInt, Scalar};
+    use rand::Rng;
+
+    impl From<u32> for ScalarType {
+        fn from(value: u32) -> Self {
+            ScalarType(Scalar::from(value))
+        }
+    }
+
+    impl From<u128> for ScalarType {
+        fn from(value: u128) -> Self {
+            ScalarType(Scalar::from(value))
+        }
+    }
+
+    #[test]
+    fn check_multiplicative_inverse() {
+        let num = 638_204_786_u32;
+        let modulo = 2_147_483_647_u32;
+
+        let inv = multiplicative_inverse(&ScalarType::from(num), &ScalarType::from(modulo));
+        assert_eq!(inv, ScalarType::from(1_237_888_327_u32));
+
+        let result = (num * BigInt::from(&inv)) % modulo;
+        assert_eq!(result, BigInt::from(1_i32));
+    }
+
+    #[test]
+    fn check_multiplicative_inverse_big_primes() {
+        let mut rng = rand::thread_rng();
+        let primes = [
+            999_999_000_001_u128,
+            67_280_421_310_721_u128,
+            4_125_636_888_562_548_868_221_559_797_461_449_u128,
+            170_141_183_460_469_231_731_687_303_715_884_105_727_u128,
+        ];
+
+        for prime in primes {
+            for _i in [1..10] {
+                let num = rng.gen_range(0..prime);
+                let inv = multiplicative_inverse(&ScalarType::from(num), &ScalarType::from(prime));
+                let result = (num * BigInt::from(&inv)) % prime;
+                assert_eq!(result, BigInt::from(1_i32));
+            }
+        }
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /.vscode/
 *.code-workspace
 */db
+dump.rdb
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,10 @@ digest = "0.9"
 hex = {version = "0.4", optional = true}
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, so pin this
 hkdf = "0.11"
+gcd = "2.1.0"
 log = "0.4"
+num-integer = "0.1"
+num-traits = "0.2"
 rand_core = "0.6"
 rand = "0.8"
 redis = "0.21.5"
@@ -29,6 +32,7 @@ stderrlog = {version = "0.5", optional = true}
 structopt = {version = "0.3", optional = true}
 
 [dev-dependencies]
+bigint = "4"
 hex = "0.4"
 lazy_static = "1.4.0"
 
@@ -43,3 +47,7 @@ required-features = ["cli"]
 [[bin]]
 name = "helper"
 required-features = ["cli"]
+
+[dependencies.num-bigint]
+version = "0.4"
+features = ["rand"]
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -1,5 +1,6 @@
 pub mod aggregation;
 pub mod event;
+pub mod prf_compute;
 pub mod privacy_budget;
 
 pub use aggregation::{
--- a/src/helpers/prf_compute/mod.rs
+++ b/src/helpers/prf_compute/mod.rs
@@ -1,0 +1,1 @@
+pub mod multiplicative_inverse;


src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+    // Assume non zero
+    assert!(!m.is_zero());
+    assert!(!n.is_zero());
+
+    let mut x_ebgcd = m.clone();
+    let mut y_ebgcd = n.clone();
+    let mut g_ebgcd = BigUint::one();
+
+    // find common factors of 2
+    let shift = cmp::min(count_twos_multiple(&x_ebgcd), count_twos_multiple(&y_ebgcd));
```
benjaminsavage, line 22: This seems like a funny optimization to throw in. I wouldn't anticipate this being a really common use-case in IPA given that we are finding the GCD of a number and a *prime*... In fact, I'm almost completely confident that we will never shift due to that...

...let's remove this code.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+    y_ebgcd >>= shift;
+    g_ebgcd <<= shift;
+
+    let mut u_ebgcd = x_ebgcd.clone();
+    let mut v_ebgcd = y_ebgcd.clone();
+
+    let mut a_ebgcd = BigInt::one();
+    let mut b_ebgcd = BigInt::zero();
+    let mut c_ebgcd = BigInt::zero();
+    let mut d_ebgcd = BigInt::one();
```
benjaminsavage, line 34: These variable names are really hard for me to work with. I have to keep looking back up and these re-assignments to understand what they all are.

What is "u_ebgcg"? Oh, it's a clone of x_ebgcg... what's that? Oh, it's a clone of m... what was m again? Is it the prime?

I'd love maybe a comment at least (if not better names). For example, what is the "_ebgcd" suffix adding here? (Aside from extra characters...)

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+    if !g.is_one() {
+        return None;
+    }
+
+    let to_add = n.to_bigint().unwrap();
+
+    if a.is_negative() {
+        while a.is_negative() {
+            a += &to_add;
+        }
```
benjaminsavage, line 97: I do not understand why a "while" loop is necessary here. It seems like there should exist some kind of constant time, non-looping approach to convert this to a positive number.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+    use num_integer::Integer;
+    use num_traits::One;
+    use std::mem;
+
+    use crate::helpers::prf_compute::multiplicative_inverse::{extended_binary_gcd, mod_inverse};
+
+    #[test]
+    fn check_mod_inv() {
+        // These numbers are prime
+        let num = 638_204_786_i32.to_biguint().unwrap();
```
benjaminsavage, line 118: The comment is wrong. this number is not prime.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+        let modulo = 2_147_483_647_i32.to_biguint().unwrap();
```
benjaminsavage, line 119: *this* one is prime =)

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let inv = mod_inverse(&num, &modulo);
```
benjaminsavage, line 121: I like the variable names "num" and "modulus" so much better than m and n... should we change them in the mod_inverse function to match?

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+            a_ebgcd -= &c_ebgcd;
+            b_ebgcd -= &d_ebgcd;
+        } else {
+            v_ebgcd -= &u_ebgcd;
+            c_ebgcd -= &a_ebgcd;
+            d_ebgcd -= &b_ebgcd;
+        }
+
+        if u_ebgcd.is_zero() {
+            return (c_ebgcd.clone(), d_ebgcd.clone(), v_ebgcd << shift);
```
benjaminsavage, line 72: Does this function return the Bézout coefficients as a and b? https://en.wikipedia.org/wiki/B%C3%A9zout%27s_identity

One option is to add a comment explaining this and keep things as they are.

However, if we do not actually need a function that returns these, we could just inline this implementation into the mod_inverse function and return the parts to compute b (since it is never used). Honestly, I'm leaning in that direction.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,170 @@
...
+        let num = 638_204_786_i32.to_biguint().unwrap();
+        let modulo = 2_147_483_647_i32.to_biguint().unwrap();
+
+        let (a, b, g) = extended_binary_gcd(&num, &modulo);
+
+        assert_eq!(g, num.gcd(&modulo));
+        assert_eq!(
+            a * num.to_bigint().unwrap() + b * modulo.to_bigint().unwrap(),
+            g.to_bigint().unwrap()
+        );
```
benjaminsavage, line 139: We should decide if we want to import this function or not.

benjaminsavage (changes requested):

src/helpers/prf_compute/multiplicative_inverse.rs
richajaindce: True.. makes sense

src/helpers/prf_compute/multiplicative_inverse.rs
richajaindce: Changed the variable names to being a bit more meaningful + also added comments to the function to tell what it actually does

src/helpers/prf_compute/multiplicative_inverse.rs
richajaindce: Yes it computes x and y for Bezout coefficients.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// This may return None when converting from f32 or f64,
+/// and will always succeed when converting from any integer or unsigned primitive, or ``BigUint``.
+#[must_use]
+pub fn mod_inverse(num: &BigUint, modulus: &BigUint) -> Option<BigInt> {
+    /// This computes the greatest common divisor (``gcd``) of integers ``num1`` and ``num2``,
+    /// and also the coefficients of Bézout's identity, which are integers ``bezout_coefficient_a`` and ``bezout_coefficient_b`` such that
+    /// ``num1`` * ``bezout_coefficient_a`` + ``num2`` * ``bezout_coefficient_b`` = ``gcd``(num1,num2)
+    #[inline]
+    fn extended_binary_gcd(num1: &BigUint, num2: &BigUint) -> (BigInt, BigUint) {
```
benjaminsavage, line 25: This function is only invoked once, within the scope of this mod_inverse function. As such, I do not see any reason to extract it like this. I recommend that we just put this functionality into this function.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+        // Assume non zero
+        assert!(!num1.is_zero());
+        assert!(!num2.is_zero());
+
+        let mut old_gcd = num1.clone();
+        let mut gcd = num2.clone();
```
benjaminsavage, line 31: I wouldn't recommend calling these variable these names. First of all, they are mutable and therefore take on many values, the majority of which are not the gcd. Secondly, I'm not sure they *ever* take on the value of the GCD. Thirdly, I don't think we even care to compute the gcd within the context of our mod_inverse function =). 

I recommend we try to align the naming with the Wikipedia article on the "Extended Euclidean Algorithm" (https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm). I think that would be the simplest way to inform the reader of the code about what is going on. 

I believe these variables map to "remainder_current" and "remainder_last"? But I'm not sure. Aligning the names will help me understand what these represent.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        let mut bezout_coefficient_old_a = BigInt::one();
+        let mut bezout_coefficient_a = BigInt::zero();
```
benjaminsavage, line 34: As with above, I think these map onto "t_current" and "t_last" from the Wikipedia article. Is that right? https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm

Let's call them that.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+#[must_use]
+pub fn mod_inverse(num: &BigUint, modulus: &BigUint) -> Option<BigInt> {
+    /// This computes the greatest common divisor (``gcd``) of integers ``num1`` and ``num2``,
+    /// and also the coefficients of Bézout's identity, which are integers ``bezout_coefficient_a`` and ``bezout_coefficient_b`` such that
+    /// ``num1`` * ``bezout_coefficient_a`` + ``num2`` * ``bezout_coefficient_b`` = ``gcd``(num1,num2)
+    #[inline]
+    fn extended_binary_gcd(num1: &BigUint, num2: &BigUint) -> (BigInt, BigUint) {
+        // Assume non zero
+        assert!(!num1.is_zero());
+        assert!(!num2.is_zero());
```
benjaminsavage, line 28: I would recommend that we add an assert that num1 is strictly less than modulus. That will help us simplify the logic below.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+        let mut bezout_coefficient_a = BigInt::zero();
+
+        loop {
+            if old_gcd < gcd {
+                gcd -= &old_gcd;
+                bezout_coefficient_a -= &bezout_coefficient_old_a;
+            } else {
+                old_gcd -= &gcd;
+                bezout_coefficient_old_a -= &bezout_coefficient_a;
+            }
```
benjaminsavage, line 43: If we add the assert above that ensures num1 is strictly less than modulus, we shouldn't have to check this condition. It should always be the case that we have a decreasing sequence of values.

benjaminsavage: I'm envisioning code where the variable names and algorithm clearly map onto the Wikipedia article and are simple to read and understand:

  quotient 
  remainder_last
  remainder_current
  remainder_new
  t_last
  t_current
  t_new

And here we compute
  quotient = remainder_last / remainder_current
  remainder_new = remainder_last % remainder_current
  t_new = t_last - quotient * t_current

We can check if we are done:
  if remainder_new.is_zero() {
    return t_current;
  }

And if we are not done, we just overwrite the mutable variables before looping:
  remainder_last = remainder_current
  remainder_current = remainder_new
  t_last = t_current
  t_current = t_new

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+
+            if old_gcd.is_zero() {
+                return (bezout_coefficient_a, gcd);
+            }
+        }
+    }
+    let (mut bezout_coefficient_a, gcd) = extended_binary_gcd(num, modulus);
```
benjaminsavage, line 50: As mentioned before, I think we should get rid of this inline function declaration, and just put the loop directly into our mod_inverse function. Thus, we could remove this function call here.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    let to_add = modulus.to_bigint().unwrap();
+
+    if bezout_coefficient_a.is_negative() {
+        bezout_coefficient_a += &to_add;
+        assert!(bezout_coefficient_a.is_positive());
+        Some(bezout_coefficient_a)
+    } else {
+        Some(bezout_coefficient_a % to_add)
+    }
```
benjaminsavage, line 64: I think all of this code can be removed once the inline function declaration above is removed and the loop is within the body of the mod_inverse function. The only exit condition needed is to check when remainder_new.is_zero()

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+#[cfg(test)]
+mod tests {
```
benjaminsavage, line 68: I would recommend adding a few more tests.

Perhaps a good approach would be to hard-code a list of known large primes: (just grab a bunch of largeish ones from this page: https://en.wikipedia.org/wiki/List_of_prime_numbers) and for each prime, make a loop which generates numbers that are strictly less than that prime. Compute the mod_inverse and check that indeed that value is correct by multiplying the two numbers and checking the result is 1 mod P.

benjaminsavage (changes requested):

src/helpers/prf_compute/multiplicative_inverse.rs
richajaindce: Yes thats right

martinthomson (commented):
I had some old comments sitting here from an unfinished review.  I'm sending those now..

Since then I realized that perhaps we don't really want this code.  We can, and perhaps should, use the Scalar from the curve25519 crate for this.

src/helpers/prf_compute/multiplicative_inverse.rs
```diff
@@ -0,0 +1,92 @@
...
+        if remainder_new.is_zero() {
+            break;
+        }
+        remainder_last = remainder_current;
+        remainder_current = remainder_new;
+        t_last = t_current;
+        t_current = t_new;
+    }
+
+    let to_add = modulus.to_bigint().unwrap();
```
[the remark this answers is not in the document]
richajaindce: It is also used to take a modulus of the value before returning in the end . Have renamed it to shadow original modulus variable

benjaminsavage:
I think we can close this PR now that we've moved on to no longer use the PRF approach.

PR #27, opened 2022-04-26
Events and query structure definitions

taikiy:
Quick implementation of source/trigger events and IPA query data models. Actual parameter types require further reviews but submitting the first try for PoC.

### DIFF
--- a/src/helpers/models.rs
+++ b/src/helpers/models.rs
@@ -1,21 +1,23 @@
-use num_bigint::BigInt;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Formatter};
 use std::ops::Range;
 
 // Type aliases to indicate whether the parameter should be encrypted, secret shared, etc.
 // Underlying types are temporalily assigned for PoC.
-type CipherText = BigInt;
+type CipherText = Vec<u8>;
 type PlainText = String;
 type SecretShare = [CipherText; 3];
-type PBRange = Range<u8>;
 
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 struct Event {
     /// Secret shared and then encrypted match keys.
     matchkeys: Vec<SecretShare>,
 
-    /// Date and time of the event occurence. Secret shared and encrypted.
+    /// The epoch which this event is generated. Using an 8-bit value = 256 epochs > 4 years (assuming 1 epoch = 1 week).
+    /// This field is in the clear.
+    epoch: u8,
+
+    /// An offset in seconds into a given offset. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
     timestamp: SecretShare,
 }
 
@@ -23,7 +25,7 @@ struct Event {
 struct SourceEvent {
     event: Event,
 
-    /// A key to group sets of the events into.
+    /// A key to group sets of the events.
     breakdown_key: PlainText,
 }
 
@@ -89,7 +91,7 @@ struct IPAQuery {
     query_type: QueryType,
 
     /// Percentage of epoch-level privacy budget this query should consume. Likely 1-100.
-    privacy_budget: PBRange,
+    privacy_budget: u8,
 
     /// A collection of source events. At least 100 (TBD) unique source events must be provided.
     source_events: Vec<SourceEvent>,
@@ -103,7 +105,8 @@ struct SourceFanoutQuery {
     query: IPAQuery,
 
     /// The maximum number of attributed conversion events that a single person can contribute
-    /// towards the final output.
+    /// towards the final output. We could also express this using sum of trigger values.
+    /// We'll leave it for the future spec to decide.
     cap: u8,
 }
 
@@ -123,7 +126,7 @@ impl Debug for SourceFanoutQuery {
 struct TriggerFanoutQuery {
     query: IPAQuery,
 
-    /// The range which all trigger event's conversion values must lie within.
+    /// The range within which all the trigger event values must lie.
     value_range: Range<u32>,
 }
 
--- a/src/helpers/models.rs
+++ b/src/helpers/models.rs
@@ -17,7 +17,7 @@ struct Event {
     /// This field is in the clear.
     epoch: u8,
 
-    /// An offset in seconds into a given offset. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
+    /// An offset in seconds into a given epoch. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
     timestamp: SecretShare,
 }
 
--- a/src/helpers/models.rs
+++ b/src/helpers/models.rs
@@ -1,0 +1,143 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::{Debug, Formatter};
+use std::ops::Range;
+
+// Type aliases to indicate whether the parameter should be encrypted, secret shared, etc.
+// Underlying types are temporalily assigned for PoC.
+type CipherText = [u8; 32];
+type PlainText = String;
+type SecretShare = [CipherText; 3];
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct Event {
+    /// Name of the business entity that generated this event. For source-fanout,
+    /// it's the business who promotes ads. For trigger-fanout, it's the business
+    /// where a user has made a conversion.
+    entity_name: PlainText,
+
+    /// Secret shared and then encrypted match keys.
+    matchkey: SecretShare,
+
+    /// Date and time of the event occurence. Secret shared and encrypted.
+    timestamp: SecretShare,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct SourceEvent {
+    event: Event,
+
+    /// Campaign ID of the ad served to the user.
+    campaign_id: PlainText,
+}
+
+#[cfg(feature = "debug")]
+impl Debug for SourceEvent {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "SourceEvent:\n  matchkey={:?}\n  campaign_id={}",
+            &self.event.matchkey, &self.campaign_id
+        )
+    }
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct TriggerEvent {
+    event: Event,
+
+    /// Conversion value.
+    value: SecretShare,
+
+    /// Zero knowledge proom that the trigger value lies within a specific range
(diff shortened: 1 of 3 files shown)


src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::{Debug, Formatter};
+use std::ops::Range;
+
+// Type aliases to indicate whether the parameter should be encrypted, secret shared, etc.
+// Underlying types are temporalily assigned for PoC.
+type CipherText = [u8; 32];
```
benjaminsavage, line 7: Isn't there a better type available for representing huge integers? I think there was one with pretty good built-in support.

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+type PlainText = String;
+type SecretShare = [CipherText; 3];
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct Event {
+    /// Name of the business entity that generated this event. For source-fanout,
+    /// it's the business who promotes ads. For trigger-fanout, it's the business
```
benjaminsavage, line 14: I think we do not need this at all for trigger-fanout queries, as it will have the same value for all events (source or trigger)

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// where a user has made a conversion.
+    entity_name: PlainText,
+
+    /// Secret shared and then encrypted match keys.
+    matchkey: SecretShare,
```
benjaminsavage, line 19: If we want to support multiple matchkeys, this should be a vector of SecretShare.

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    /// Date and time of the event occurence. Secret shared and encrypted.
+    timestamp: SecretShare,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct SourceEvent {
+    event: Event,
+
+    /// Campaign ID of the ad served to the user.
```
benjaminsavage, line 29: Per this thread: https://github.com/patcg/private-measurement/issues/9#issuecomment-1121329028 we might want to consider renaming this to "group_id" or even just "breakdown_key".

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct TriggerEvent {
+    event: Event,
+
+    /// Conversion value.
+    value: SecretShare,
+
+    /// Zero knowledge proom that the trigger value lies within a specific range
```
benjaminsavage, line 51: spelling: proom => proof

benjaminsavage: I'm not sure if this is the best type - but we can always change this later.

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+            f,
+            "TriggerEvent:\n  matchkey={:?}\n  value={:?}",
+            &self.event.matchkey, &self.value
+        )
+    }
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+enum QueryType {
+    Source,
```
benjaminsavage, line 69: Should we call them "SourceFanout" and "TriggerFanout"?

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+    leader_node: Node,
+
+    /// List of match key providers to be used by the source and trigger events during an epoch.
+    mk_providers: Vec<String>,
+
+    /// Source-fanout or Trigger-fanout.
+    query_type: QueryType,
+
+    /// Percentage of epoch-level privacy budget this query should consume.
+    privacy_budget: f32,
```
benjaminsavage, line 95: We probably cannot support arbitrarily fine-grained queries. I suspect an integer value would be better than a float. For example, an integer ranging from 1 to 100 might actually be better (i.e. we might want to have a cap on the total number of queries can can make).

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+    trigger_events: Vec<TriggerEvent>,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct SourceFanoutQuery {
+    query: Query,
+
+    /// The maximum number of attributed conversion events that a single person can contribute
+    /// towards the final output.
+    cap: u32,
```
benjaminsavage, line 110: u8 should be fine I think.

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+enum Node {
+    Helper1,
+    Helper2,
+    Helper3,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct Query {
```
benjaminsavage, line 81: To disambiguate this against QueryType, I'd suggest renaming this IPAQuery

benjaminsavage (changes requested):
Thanks for submitting this! A few suggestions.

taikiy:
@benjaminsavage Thanks for reviewing! Fixed the code accordingly.

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
+use num_bigint::BigInt;
+use serde::{Deserialize, Serialize};
+use std::fmt::{Debug, Formatter};
+use std::ops::Range;
+
+// Type aliases to indicate whether the parameter should be encrypted, secret shared, etc.
+// Underlying types are temporalily assigned for PoC.
+type CipherText = BigInt;
```
benjaminsavage, line 8: @martinthomson - if we use some basic encryption approach like AES, what would be the right type signature to store those CipherTexts? Is it BigInt, or is it something else?

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
...
+type PlainText = String;
+type SecretShare = [CipherText; 3];
+type PBRange = Range<u8>;
```
benjaminsavage, line 11: I don't think we want to use a Range for this. A range has a start and an end. The privacy budget a query will consume will be a fixed number, not something within a range.

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
...
+    leader_node: Node,
+
+    /// List of match key providers to be used by the source and trigger events during an epoch.
+    mk_providers: Vec<String>,
+
+    /// Source-fanout or Trigger-fanout.
+    query_type: QueryType,
+
+    /// Percentage of epoch-level privacy budget this query should consume. Likely 1-100.
+    privacy_budget: PBRange,
```
benjaminsavage, line 92: How can we consume a range? We would consume a specific value.

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
...
+type PlainText = String;
+type SecretShare = [CipherText; 3];
+type PBRange = Range<u8>;
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct Event {
+    /// Secret shared and then encrypted match keys.
+    matchkeys: Vec<SecretShare>,
+
+    /// Date and time of the event occurence. Secret shared and encrypted.
```
benjaminsavage, line 18: It wouldn't be date and time. I would expect that the timestamp would just be an offset into a given Epoch. We probably need another field to indicate which Epoch we are in. That wouldn't need to be encrypted or secret shared - it could be in the clear. I'm not sure what's the best type to express the Epoch. @martinthomson - what do you think?

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// A collection of trigger events. At least 10 (TBD) unique trigger events must be provided.
+    trigger_events: Vec<TriggerEvent>,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct SourceFanoutQuery {
+    query: IPAQuery,
+
+    /// The maximum number of attributed conversion events that a single person can contribute
```
benjaminsavage, line 105: I'm wondering if we should express the number of attributed conversion events, or if we should cap their sum of trigger value? I think they are equivalent since we will have range proofs that all the trigger values lie within a given range. I think this is fine for now until we have reason to change it.

src/helpers/models.rs
```diff
@@ -0,0 +1,140 @@
...
+            &self.query.trigger_events.len()
+        )
+    }
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct TriggerFanoutQuery {
+    query: IPAQuery,
+
+    /// The range which all trigger event's conversion values must lie within.
```
benjaminsavage, line 126: nit: the trailing preposition bothers me =).

I suggest: "The range within which all the trigger event values must lie"

benjaminsavage (changes requested):

src/helpers/models.rs
martinthomson: As a buffer of bytes generally, so `Vec<u8>` or `&[u8]` depending on whether you need a reference.

src/helpers/models.rs
martinthomson: Within an epoch, we have 2^20 seconds, so u32 for that part would do.

For the epoch number, we only need to ensure that there are enough values that we can be reasonably sure that events won't be mistaken.  For instance, if you chose a 2 bit epoch, then every 4 epochs you would use the same number.  You would have to force helpers to rotate their keys more often than the rollover occurs.  Given that, an 8-bit value might be enough for the epoch as it is probably OK to force key rotations every 4 years.

Make it 16 bits and you never have to rotate keys, or at least you aren't forced to plan for key rotation.  That might not be good in that building and testing a critical function can be deferred indefinitely, after which dealing with a compromise becomes even worse.

taikiy:
@benjaminsavage I was offline for a several days last week. Updated the PR again. Thanks in advance!

src/helpers/models.rs
```diff
@@ -0,0 +1,143 @@
...
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct Event {
+    /// Secret shared and then encrypted match keys.
+    matchkeys: Vec<SecretShare>,
+
+    /// The epoch which this event is generated. Using an 8-bit value = 256 epochs > 4 years (assuming 1 epoch = 1 week).
+    /// This field is in the clear.
+    epoch: u8,
+
+    /// An offset in seconds into a given offset. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
```
benjaminsavage, line 20: "...into a given epoch."

benjaminsavage (approved):
Love it. Thank you for this!

PR #28, opened 2022-04-26
On a local machine, Install and start Redis in case not running

richajaindce:
In case redis server is not running, we download a source copy and start redis server.
This is done in the pre-commit script to ensure all tests pass (including privacy budget which expects redis server to be running)
On github, this is already covered using Actions

Tip: If you are running tests for the first time, recommend to run pre-commit script before executing tests


pre-commit
```diff
@@ -46,4 +46,6 @@ error() {
 
 cargo clippy --tests -- -D warnings --D clippy::pedantic || error "Clippy errors"
 
+./install_and_start_redis
```
martinthomson, line 49: This only works for the pre-commit hook.

I think that it is reasonable to assume that people already have a redis-server installed locally if they are doing work on this project.  But the tests should be responsible for starting their own server instance as necessary.

If you don't want to assume anything, you can add a build.rs file and have that install redis-server into the target/ directory.

install_and_start_redis
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e
+
+REDIS_VERSION="6.2.4"
+pwd=$PWD
+TEST_DIR=$PWD/test_dependencies
+REDIS_SERVER_SCRIPT=$TEST_DIR/redis-$REDIS_VERSION/src/redis-server
+
+if [ $(ps aux | grep -v grep | grep -c $REDIS_SERVER_SCRIPT) -gt 0 ]
```
martinthomson, line 9: The usual way to manage servers is to track their PID in a file.  That's far less fragile than this, which might end up thinking that a different server instance is running.

pre-commit
richajaindce: Agree that build.rs would be a better approach. For now, all developers are implementing their modules so dont really need to "ensure" a running redis server. Will move it to build.rs

PR #29, opened 2022-05-11
Add cargo-make dependency and makefiles

akoshelev:
This PR suggest to use `cargo-make` as our primary build tool. It has a slightly nicer interface than `build.rs`, supports auto-installing dependencies and is fully integrated with Github actions (not implemented in this PR).

I've added several tasks that can be used to validate code before submitting a PR - these tasks run the same set of checks as Github CI, so it could be convenient for contributors to make sure their code is correct before getting an official feedback from the CI tool.


Makefile.toml
```diff
@@ -0,0 +1,35 @@
...
+[tasks.clippy]
+command = "cargo"
+args = ["clippy", "--tests", "--", "-D","warnings", "-D", "clippy::pedantic"]
+
+[tasks.coverage]
+alias="coverage_grcov"
+
+[tasks.doctest]
+command = "cargo"
+args = ["test", "--doc"]
```
martinthomson, line 19: cargo test runs doc tests already, so aren't you doubling up on that work?

martinthomson (commented):
Not sure about the grcov stuff as that seems like a bit of a white elephant (for example, we're already at 1.60 but it has a bunch of code for the 1.59->1.60 transition; also, it's a bit short on variable quoting).  I guess we can just pull that in occasionally.

Makefile.toml
akoshelev: you're right, `cargo test` does run doc tests, so this is not required

akoshelev:
> Not sure about the grcov stuff as that seems like a bit of a white elephant (for example, we're already at 1.60 but it has a bunch of code for the 1.59->1.60 transition; also, it's a bit short on variable quoting). I guess we can just pull that in occasionally.

I agree, it does look scary. I looked into it a bit more and it seems there is a [better way](https://github.com/BamPeers/rust-ci-github-actions-workflow/blob/main/.github/workflows/test.yaml#L40) to produce coverage reports. Will be removing grcov shortly

akoshelev:
it does not seem that we're in favor of pulling `cargo make`, I will probably just create a bash script that runs the same checks as GH workflow instead.

PR #31, opened 2022-06-07
Enable clippy rule clone_on_ref_ptr

akoshelev:
It is disabled by default.
https://rust-lang.github.io/rust-clippy/master/#clone_on_ref_ptr
Motivated by https://github.com/rust-lang/rfcs/issues/2588

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::clone_on_ref_ptr)]
+
 #[cfg(feature = "cli")]
 pub mod cli;
 pub mod error;


martinthomson:
Given that CI already runs with clippy::pedantic, isn't this covered already?  We just have memorialized it in the code, but we could do that instead.  (Adding one lint at a time seems like it might take a lot of effort.)

akoshelev:
> Given that CI already runs with clippy::pedantic, isn't this covered already? We just have memorialized it in the code, but we could do that instead. (Adding one lint at a time seems like it might take a lot of effort.)

I think the default clippy setting is [`allow`](https://rust-lang.github.io/rust-clippy/master/#clone_on_ref_ptr), so `clippy::pedantic` does not block these. I thought we could use `.clippy.toml` for it but it is marked unstable so given that we don't have code that uses `<arc>.clone()` yet, this might be a good way to prevent it from appearing in the repository. I think the pipeline PR violates this rule already

martinthomson:
Oh, I didn't check the rule's default enforcement.  I wonder why it never made it into the pedantic set...

If we merge this first, the pipeline code will hit it :)

akoshelev:
I think it is considered controversial https://github.com/rust-lang/rust-clippy/issues/2048, but we can see whether it works for us :)

PR #32, opened 2022-06-14
Implement Shamir secret sharing over a finite field

akoshelev:
## Overview

This change adds Shamir secret sharing implementation over a finite field (implementation of FF is provided by [`ff`](https://docs.rs/ff/latest/ff/) crate. This implementation supports share addition if field that is used for coefficients and secrets is a finite prime field. 

The goal is to be able to execute secure addition (and later secure multiplication) protocol over the wire by running helper instances that perform MPC and client that submits the job. 

## Existing implementations

There are plenty of [SSS implementations](https://crates.io/keywords/shamir) written in Rust, however most of them use binary fields GF(2^n) which makes it impossible to implement share addition in O(1). See [this](https://en.wikipedia.org/wiki/Finite_field_arithmetic#Addition_and_subtraction) for more information.

## Roadmap

- [x] Shamir secret sharing implementation (this change)
- [ ] Network/Communication layer for helpers to be able to exchange messages
- [ ] Helper and client binaries in IPA
- [ ] Secure addition protocol
- [ ] Secure multiplication protocol
- [ ] Test harness that executes addition and multiplication

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -48,6 +48,10 @@ impl Field for Fp31 {
     const ONE: Self = Fp31(1);
 
     fn invert(&self) -> Self {
+        debug_assert!(
+            self != &Self::ZERO,
+            "Multiplicative inverse is not defined for Fp31(0)"
+        );
         self.pow(Self::PRIME - 2)
     }
 
@@ -201,5 +205,13 @@ mod test {
                 "{field_element:?}*1/{field_element:?} != 1"
             );
         }
+    }
+
+    #[test]
+    #[should_panic]
+    fn invert_panics_if_called_on_zero() {
+        // assertion does not matter here, test should panic when `invert` is called.
+        // it is here to silence #must_use warning
+        assert_ne!(Fp31::ZERO, Fp31(0).invert());
     }
 }
--- a/src/field.rs
+++ b/src/field.rs
@@ -20,6 +20,18 @@ pub trait Field:
 {
     type Integer;
     const PRIME: Self::Integer;
+    /// Additive identity element
+    const ZERO: Self;
+    /// Multiplicative identity element
+    const ONE: Self;
+
+    /// computes the multiplicative inverse of `self`. It is UB if `self` is 0.
+    #[must_use]
+    fn invert(&self) -> Self;
+
+    /// Computes modular exponentiation, self^exp (mod PRIME)
+    #[must_use]
+    fn pow(&self, exp: Self::Integer) -> Self;
 }
 
 // TODO(mt) - this code defining fields can be turned into a macro if we ever
@@ -31,6 +43,24 @@ pub struct Fp31(<Self as Field>::Integer);
 impl Field for Fp31 {
     type Integer = u8;
     const PRIME: Self::Integer = 31;
+    const ZERO: Self = Fp31(0);
+    const ONE: Self = Fp31(1);
+
+    fn invert(&self) -> Self {
+        self.pow(Self::PRIME - 2)
+    }
+
+    fn pow(&self, exp: Self::Integer) -> Self {
+        let mut t = Self::ONE;
+        for i in (0..Self::Integer::BITS - exp.leading_zeros()).rev() {
+            t *= t;
+            if (exp >> i) & 1 != 0 {
+                t *= *self;
+            }
+        }
+
+        t
+    }
 }
 
 impl Add for Fp31 {
@@ -114,6 +144,7 @@ impl Debug for Fp31 {
 #[cfg(test)]
 mod test {
     use crate::field::Field;
+    use std::ops::Mul;
 
     use super::Fp31;
 
@@ -142,5 +173,32 @@ mod test {
         assert_eq!(Fp31(0), Fp31(0) + Fp31(0));
         assert_eq!(Fp31(<Fp31 as Field>::PRIME - 1), Fp31(0) - Fp31(1));
         assert_eq!(Fp31(0), Fp31(0) * Fp31(1));
+    }
+
+    #[test]
+    fn pow() {
+        let zero = Fp31::ZERO;
+        let one = Fp31::ONE;
+
+        assert_eq!(Fp31(2), Fp31(2).pow(1));
+        assert_eq!(one, Fp31(2).pow(0));
+        assert_eq!(one, one.pow(0));
+        assert_eq!(one, one.pow(2));
+        assert_eq!(zero, zero.pow(2));
+
+        assert_eq!(Fp31(Fp31::PRIME - 1), Fp31(Fp31::PRIME - 1).pow(1));
+        assert_eq!(one, Fp31(2).pow(Fp31::PRIME - 1));
+    }
+
+    #[test]
+    fn invert() {
+        for i in 1..Fp31::PRIME {
+            let field_element = Fp31(i);
+            assert_eq!(
+                Fp31::ONE,
+                field_element.invert().mul(field_element),
+                "{field_element:?}*1/{field_element:?} != 1"
+            );
+        }
     }
 }
--- a/src/field.rs
+++ b/src/field.rs
@@ -18,7 +18,8 @@ pub trait Field:
     + Debug
     + Sized
 {
-    type Integer;
+    type Integer: Debug + Into<u128>;
+
     const PRIME: Self::Integer;
     /// Additive identity element
     const ZERO: Self;
--- a/src/shamir.rs
+++ b/src/shamir.rs
@@ -14,23 +14,29 @@ use std::num::NonZeroU8;
 use std::ops::Add;
 use thiserror::Error;
 
-
+/// Interpolating polynomial used to reconstruct secrets encoded with Shamir secret scheme.
 pub struct LagrangePolynomial<F> {
-    coefficients: Vec<F>
+    coefficients: Vec<F>,
 }
 
-impl <F: PrimeField> LagrangePolynomial<F> {
-    pub fn new(n: NonZeroU8) -> Result<Self, Error> {
-        let mut coefficients = Vec::with_capacity(n.get() as usize);
-        let n = n.get();
+impl<F: PrimeField> LagrangePolynomial<F> {
+    /// Constructs new polynomial of a given degree.
+    ///
+    /// ## Errors
+    /// Returns an error if it fails to evaluate at least one of the coefficients
+    pub fn new(degree: NonZeroU8) -> Result<Self, Error> {
+        let mut coefficients = Vec::with_capacity(degree.get() as usize);
+        let n = degree.get();
         for i in 1..=n {
             let mut x = F::one();
             let mut denom = F::one();
 
             for j in 1..=n {
                 if i != j {
-                    x *= F::from(j as u64);
-                    denom *= F::from(j as u64) - F::from(i as u64);
+                    let x_i = F::from(u64::from(i));
+                    let x_j = F::from(u64::from(j));
+                    x *= x_j;
+                    denom *= x_j - x_i;
                 }
             }
 
@@ -40,19 +46,26 @@ impl <F: PrimeField> LagrangePolynomial<F> {
                     v: Box::from(denom.to_repr().as_ref()),
                 });
             }
-            coefficients.push(x * maybe_denom.unwrap())
+            coefficients.push(x * maybe_denom.unwrap());
         }
 
         Ok(Self { coefficients })
     }
 
-    pub fn len(&self) -> u8 {
+    #[must_use]
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn degree(&self) -> u8 {
         // SAFETY: new() does not allow constructing more than u8::MAX coefficients
         self.coefficients.len() as u8
     }
 
-    pub fn iter(&self) -> std::slice::Iter<'_, F> {
-        self.coefficients.iter()
+    // Evaluates P(0) using the first N points provided, ignoring extra points, if any.
+    fn evaluate_at_zero<T: Clone + Into<F>>(&self, points: &[T]) -> F {
+        points
+            .iter()
+            .zip(self.coefficients.iter())
+            .map(|(y, &x)| y.clone().into() * x)
+            .fold(F::zero(), |lhs, rhs| lhs + rhs)
     }
 }
 
@@ -69,8 +82,12 @@ pub struct SecretSharing {
 pub enum Error {
     #[error("Wrong or insecure secret sharing scheme. (expected {k} >= {n} > 1)")]
     BadSharingScheme { k: u8, n: u8 },
-    #[error("Not enough shares to reconstruct secret: need at least {actual}, got {required}")]
-    NotEnoughShares { required: u8, actual: u8 },
+    #[error("The degree of polynomial provided to reconstruct secret does not match number of points provided. \
+     Polynomial degree: {polynomial_degree}, number of points {points_count}")]
+    BadPolynomial {
+        polynomial_degree: u8,
+        points_count: u8,
+    },
     #[error("Prime field element inversion failed: {v:?}")]
     InvertError { v: Box<[u8]> },
 }
@@ -142,20 +159,23 @@ impl SecretSharing {
     /// # Errors
     /// Returns an error if there is no enough shares to reconstruct the secret or
     /// if an error occurred while evaluating the polynomial
-    pub fn reconstruct<F: PrimeField>(&self, shares: &[Share<F>], lagrange: &LagrangePolynomial<F>) -> Result<F, Error> {
-        if shares.len() < lagrange.len() as usize {
+    pub fn reconstruct<F>(shares: &[Share<F>], lagrange: &LagrangePolynomial<F>) -> Result<F, Error>
+    where
+        F: PrimeField + From<Share<F>>,
+    {
+        // this check is actually stricter than we need: shares.len() >= lagrange.len() is enough
+        // to reconstruct (ignore extra shares). The reason why it is here is specific to IPA protocol -
+        // we should not attempt to reconstruct secrets with more than n or less than n shares.
+        if shares.len() != lagrange.degree() as usize {
             // SAFETY: len() fits into u8
             #[allow(clippy::cast_possible_truncation)]
-            return Err(Error::NotEnoughShares {
-                required: lagrange.len(),
-                actual: shares.len() as u8,
+            return Err(Error::BadPolynomial {
+                polynomial_degree: lagrange.degree(),
+                points_count: shares.len() as u8,
             });
         }
 
-        let r = shares.iter()
-            .zip(lagrange.iter())
-            .map(|(share, &coeff)| share.y * coeff)
-            .fold(F::zero(), |lhs, rhs| lhs+rhs);
+        let r = lagrange.evaluate_at_zero(shares);
 
         Ok(r)
     }
@@ -166,7 +186,7 @@ impl SecretSharing {
     }
 }
 
-impl <F: Field> Add for &Share<F> {
+impl<F: Field> Add for &Share<F> {
     type Output = Share<F>;
 
     fn add(self, rhs: Self) -> Self::Output {
@@ -204,6 +224,12 @@ mod tests {
         }
     }
 
+    impl From<Share<Fp>> for Fp {
+        fn from(share: Share<Fp>) -> Self {
+            share.y
+        }
+    }
+
     #[test]
     fn can_share_8_byte_int() {
         let mut rng = StepRng::new(1, 1);
@@ -216,14 +242,19 @@ mod tests {
                 let secret = Fp::from(213);
                 let shares = shamir.split(secret, &mut rng);
 
-                assert_eq!(secret, shamir.reconstruct(&shares, &lc).unwrap());
+                assert_eq!(secret, SecretSharing::reconstruct(&shares, &lc).unwrap());
             }
         }
     }
 
     #[test]
     fn can_add_shares() {
-        fn check_addition(shamir: &SecretSharing, lc: &LagrangePolynomial<Fp>, lhs_secret: u64, rhs_secret: u64) {
+        fn check_addition(
+            shamir: &SecretSharing,
+            lc: &LagrangePolynomial<Fp>,
+            lhs_secret: u64,
+            rhs_secret: u64,
+        ) {
             let mut rng = StepRng::new(1, 1);
 
             let lhs_secret = Fp::from(lhs_secret);
@@ -238,16 +269,13 @@ mod tests {
                 .map(|(lhs, rhs)| lhs + rhs)
                 .collect::<Vec<_>>();
 
-            let sum_secret = shamir
-                .reconstruct(&sum_shares, lc)
-                .unwrap();
+            let sum_secret = SecretSharing::reconstruct(&sum_shares, lc).unwrap();
             assert_eq!(expected, sum_secret);
         }
 
         let k = NonZeroU8::new(2).unwrap();
         let n = NonZeroU8::new(3).unwrap();
-        let shamir =
-            SecretSharing::new(k, n).unwrap();
+        let shamir = SecretSharing::new(k, n).unwrap();
         let lc = LagrangePolynomial::<Fp>::new(n).unwrap();
 
         check_addition(&shamir, &lc, 42, 24);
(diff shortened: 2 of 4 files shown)


thurstonsand (commented):
so would this be a piece of what is needed for a generalized circuit solution? or is this more like a manual implementation of one possible circuit?

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+            let mut denom = F::one();
+            for share_j in shares {
+                if share_j.x != share_i.x {
+                    x *= F::from(u64::from(share_j.x));
+                    denom *= F::from(u64::from(share_j.x)) - F::from(u64::from(share_i.x));
+                }
+            }
+
+            let maybe_denom = denom.invert();
+            if maybe_denom.is_none().into() {
```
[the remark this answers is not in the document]
akoshelev: `CtOption` is a little clunky, in particular it is not an enum and it tries to make all operations to take constant time, so there is no easy way to get the value out of it

akoshelev:
> so would this be a piece of what is needed for a generalized circuit solution? or is this more like a manual implementation of one possible circuit?

I guess it depends whether we are going to use arithmetic circuits or boolean circuits for IPA. This part is not clear yet, but if we go with arithmetic circuits, we can build them on top of `ff`

src/shamir.rs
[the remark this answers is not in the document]
thurstonsand: I took a look at the [`CtOption` doc page](https://docs.rs/subtle/2.1.1/subtle/struct.CtOption.html) and it specifically mentions finite fields, so yeah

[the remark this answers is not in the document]
akoshelev: it implements `From<CtOption<T>>` for `Option<T>` but I am not sure it is the right thing to do because that will likely violate the constness of this API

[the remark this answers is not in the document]
akoshelev: We talked about it offline and the ideal solution would be to convert `CtOption` to `Result<T, E>` but this change has to be made in `CtOption` itself and this struct lives in the external crate.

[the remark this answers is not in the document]
martinthomson: You mean something like `Option::ok_or` ?

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+            })
+        }
+    }
+}
+
+/// Generate coefficients (sampled at random) for an arbitrary polynomial of infinite degree.
+fn gen_polynomial<R: RngCore, F: Field>(mut rng: R) -> impl Iterator<Item = F> {
+    repeat_with(move || {
+        let p: F = F::random(&mut rng);
+        p
```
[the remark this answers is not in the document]
akoshelev: that works, I think that was just a leftover from debugging

src/shamir.rs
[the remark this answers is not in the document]
akoshelev: yea, that was the method I was thinking about

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+            Err(ShareError::MisalignedShares {
+                lhs: self.x,
+                rhs: rhs.x,
+            })
+        }
+    }
+}
+
+/// Generate coefficients (sampled at random) for an arbitrary polynomial of infinite degree.
+fn gen_polynomial<R: RngCore, F: Field>(mut rng: R) -> impl Iterator<Item = F> {
```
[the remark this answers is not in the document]
akoshelev: ack!

martinthomson (approved):

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        let mut r = F::zero();
+
+        for share_i in shares {
+            let mut x = F::one();
+            let mut denom = F::one();
+            for share_j in shares {
+                if share_j.x != share_i.x {
+                    x *= F::from(u64::from(share_j.x));
```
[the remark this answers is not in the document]
martinthomson: Oh, nevermind.  It's a u8 input.

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+            let lhs_shares = shamir.split(lhs_secret, &mut rng);
+            let rhs_shares = shamir.split(rhs_secret, &mut rng);
+            let sum_shares = lhs_shares
+                .iter()
+                .zip(rhs_shares.iter())
+                .map(|(lhs, rhs)| lhs.add(rhs).unwrap())
+                .collect::<Vec<_>>();
+
+            let sum_secret = shamir
+                .reconstruct(&sum_shares[0..shamir.threshold() as usize])
```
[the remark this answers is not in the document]
akoshelev: I played with const generics a bit here and they seemed to me less ergonomic to use compared to variable `k` and `n` for the following reasons:

* any function that takes `SecretSharing` as an argument, now needs to be parametrized with const generic argument or make an assumption about the specific `K` used in IPA. 
* Dynamic creation of `SecretSharing` with const generic parameter does not seem to be possible (yet), so testing for K = 1,2,3,...255 gets tricky.
* `usize` is the only type Rust supports now to constrain array bounds (as of now) and it does not fit our use case well because there are a lot of invalid `K` values which we can only reject at runtime (0, 1, 256, ...). We need const expressions feature, otherwise someone can just simply crash our program by providing K = 1_000_000. 

Thinking more about it, there seem to be multiple directions we can go:


* Hardcode K = N = 3. Cache Lagrange coefficients and leverage the knowledge that K is small to leverage stack whenever possible.
* Assume K = N and implement secret sharing with no redundancy but support any K within [2, 255] range. Using const generics is still an option, but as described above it is probably less convenient than dynamic `K ` . Caching Lagrange coefficients is still beneficial and can be implemented, however there will be no compile time checks on `K`. 
* Implement Secret Sharing with maximum flexibility and assume K <= N <= 256 (this still feels reasonable). 

I am personally in favour of option 2, but curious to hear what you think

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+            .collect::<Vec<_>>();
+
+        let mut shares = Vec::with_capacity(self.n as usize);
+
+        // sample n points by evaluating the polynomial
+        for x in 1..=self.n {
+            let x_i = F::from(u64::from(x));
+            let mut y = secret;
+            for c in &coefficients {
+                y += (*c) * x_i;
```
benjaminsavage, line 91: I am almost certain you want:
  y += (*c) * x_i.pow(j)

Where j is the index of the loop

martinthomson: Maybe you can build in reverse and avoid the "pow" call.  The usual pattern here is to start from the non-constant coefficients and do something like

```rust
let mut y = 0;
for c in coefficients.iter().rev().chain([secret].iter()) {
  y *= x_i;
  y += *c;
}
```

martinthomson: Or more simply...

```rust
let mut y = 0;
for c in coefficients.iter().rev() {
  y += *c;
  y *= x_i;
}
y += secret;
```

akoshelev: yea, thanks @benjaminsavage for spotting a bug here - certainly a miss. The funny part that there is another bug at the line 81 where I take `k` coefficients while I actually need `k-1` and that issue finally appeared after I fixed polynomial degree.

@martinthomson I like what you suggested, in the future we may switch to use [`pow_vartime`](https://docs.rs/ff/0.12.0/ff/trait.Field.html#method.pow_vartime) as potentially more efficient

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+impl SecretSharing {
+    /// Constructs a new instance, returning `ShamirError` if input values do not allow
+    /// to construct a valid Shamir secret sharing scheme.
+    ///
+    /// # Errors
+    /// Rejects the following inputs:
+    /// * k == 1
+    /// * k > n
+    pub fn new(k: NonZeroU8, n: NonZeroU8) -> Result<Self, Error> {
+        if k.get() == 1 || k > n {
```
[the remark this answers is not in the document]
martinthomson: It depends on whether you think that k or n could ever result from input or if they are programmed.

Something that is programmed can assert.  Something that comes from input needs Result.  At some level, an assert here shifts responsibility for input checking to the callers of this function, but that might be better (they might be in a better position to provide useful feedback).

src/shamir.rs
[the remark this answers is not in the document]
akoshelev: I stand corrected here, we actually need k < n to be able to multiply secret shares and recover the product by using `n` shares, so I am inclined to keep the API as is, unless there is a strong objection against it. When (or rather if) we decide to cache LC, we can start rejecting inputs to reconstruct with number of shares < n.

src/shamir.rs
```diff
@@ -0,0 +1,364 @@
...
+        let shares = shamir.split(Fp::new(42), thread_rng());
+
+        assert!(matches!(
+            shares[0].add(&shares[1]),
+            Err(ShareError::MisalignedShares { .. })
+        ));
+    }
+
+    #[test]
+    fn fails_if_not_enough_shares() {
```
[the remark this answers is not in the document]
akoshelev: it is a good thing to cover I think, at least the fact that secret may not be recovered in this case. We can also potentially detect that and abort the reconstruction progress, but I am not sure how important it is to do right now

src/shamir.rs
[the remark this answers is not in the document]
akoshelev: Depending on the circumstances, function input values may be taken from user or hard-coded - I can see it both ways. Moreover, what is programmed today, may become a user input tomorrow. I would assert that usize is 8 bytes, but I would reject the function input by returning Result if I have a chance. What is particularly inconvenient about panics is lack of visibility and inverted control flow.  It seems to me that `Result` type in Rust is ergonomic enough to be a good substitute for most of the cases covered by panics/exceptions in other languages.

src/shamir.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        shares
+    }
+
+    /// Reconstructs a secret from a set of size at least k of shares.
+    ///
+    /// # Errors
+    /// Returns an error if there is no enough shares to reconstruct the secret or
+    /// if an error occurred while evaluating the polynomial
+    pub fn reconstruct<F: PrimeField>(&self, shares: &[Share<F>]) -> Result<F, Error> {
```
benjaminsavage, line 108: Honestly, I would consider taking in the Lagrange coefficients as an input, since there is really no need to re-calculate them every time.

src/shamir.rs
```diff
@@ -0,0 +1,398 @@
...
+    #[error("Not enough shares to reconstruct secret: need at least {actual}, got {required}")]
+    NotEnoughShares { required: u8, actual: u8 },
+    #[error("Prime field element inversion failed: {v:?}")]
+    InvertError { v: Box<[u8]> },
+}
+
+/// Represents a single share: (x, f(x)) point
+#[derive(Clone)]
+pub struct Share<F> {
+    x: u8,
```
benjaminsavage, line 38: Each helper will only ever have just one value of "x" for all of its shares. It seems like a waste of memory to copy it around so much. Maybe you can just store this value in a single place and reference it when you need it.

akoshelev: it is possible, but more expensive as we would need to keep an 8 byte pointer to 1 byte value. Also, if we want helper Y to only work with shares "x = Y" then we would need to validate that all shares it receives from outside have "x" set to Y and reject such inputs as soon as it discovers that, so `Share` object carrying `x` value makes it possible

src/shamir.rs
akoshelev: I think it is nice that the caller does not need to know how to compute them. Maybe this implementation is too generic for IPA, I don't see how we can cache them in the case when any `k` out of `n` of shares are provided to reconstruct a secret. I maybe reading https://en.wikipedia.org/wiki/Lagrange_polynomial wrong though - to me it looks like LC depend on specific points that are provided to compute the polynomial: i.e. if I have 3 points (1, 2, 3) and need to reconstruct a linear polynomial, LC would be different if I use points (1,2) or (2,3) for that.


Having said that, in IPA we probably don't need the generic implementation as we seem to always reconstruct from n. I could probably have this interface and leave callers responsible for correctness:

```rust
  pub fn reconstruct<F: PrimeField>(&self, shares: &[Share<F>], lagrange_coeffs: Option<&[F]>) -> Result<(F, Box<[F]>), Error)
```

so the first call would return LC back to the caller and they can reuse it later without knowing how exactly this module computed them. We could even do some rudimentary validation by checking that for any x included in shares, we have LC computed and stored inside `lagrange_coeff`. The only case where I think it won't work is trivial `k = 1`. 

any thoughts on that @benjaminsavage @martinthomson ?

akoshelev: so the best interface I could come up with is as follows (ignore the fact that it returns a tuple for a moment, this part can be improved easily)

```rust
pub fn reconstruct<'lcl, F: PrimeField>(&self, shares: &[Share<F>], lagrange: Option<&'lcl [F]>) -> Result<(F, Cow<'lcl, [F]>), Error>
``` 

This function fails to produce the correct result in this case which makes me a little uncomfortable

```rust
    #[test]
    fn wrong_result_if_lagrange_coefficients_provided_are_for_different_input() {
        let (k, n) = (2_u8, 3_u8);
        let rng = StepRng::new(1, 1);
        let secret = Fp::from(5);

        let sharing =
            SecretSharing::new(NonZeroU8::new(k).unwrap(), NonZeroU8::new(n).unwrap()).unwrap();

        let shares = sharing.split(secret, rng);
        // this call returns Lagrange coefficients for k = n 
        let (_, lc) = sharing.reconstruct(&shares, None).unwrap();

        // reconstruct will be able to find all the required coefficients, but fails to compute
        // the correct result because it only uses 2 out of 3 shares (we need different LC in this case)
        let actual = sharing.reconstruct(&shares[0..2], Some(&lc)).unwrap().0;
        assert_ne!(actual, secret)
    }
```


an alternative idea is to always require callers to provide `n` shares to reconstruct a secret, deviating from the most generic implementation but allows for efficient LC caching. I think I am leaning towards that

martinthomson: Why not create a container for the LC and have a method on that that does reconstruction?

src/shamir.rs
benjaminsavage: I still think it's un-necessary. Each helper should just know what its x-coordinate it. It doesn't need a pointer to it in every secret share.

When shares are received from other helpers for the purpose of reconstructing the secret - it just needs to know which batch of shares came from which helper.

But this is an optimization we can always add later. This version has more opportunities for validation which can assist with debugging.

src/shamir.rs
benjaminsavage: At the present time, I can only envision us reconstructing the secret within the context of secure multiplication. In that, you need all 3 shares because you're reconstructing from a quadratic secret sharing. So caching all 3 Lagrange coefficients seems fine to me.

I agree with @martinthomson that putting the reconstruction stuff into a separate container might make sense. I agree that the caller shouldn't need to know how to compute the coefficients.

src/shamir.rs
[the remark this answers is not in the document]
benjaminsavage: resolved?

src/shamir.rs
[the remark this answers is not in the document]
benjaminsavage: resolved?

src/shamir.rs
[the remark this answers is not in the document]
benjaminsavage: resolved?

src/shamir.rs
akoshelev: We discussed it yesterday in person and decided to create a wrapper struct that represents Lagrange polynomial and take it as an argument to `reconstruct` function

src/shamir.rs
akoshelev: it is done now

akoshelev:
@benjaminsavage @martinthomson I had to make some changes in `Field` trait to make it work with Shamir - in particular, implement `pow` and `invert`. Take a look when you have time and lmk what you think

src/field.rs
```diff
@@ -18,8 +18,21 @@ pub trait Field:
...
-    type Integer;
+    type Integer: Debug + Into<u128>;
+
     const PRIME: Self::Integer;
+    /// Additive identity element
+    const ZERO: Self;
+    /// Multiplicative identity element
+    const ONE: Self;
+
+    /// computes the multiplicative inverse of `self`. It is UB if `self` is 0.
```
benjaminsavage, line 29: I'm confused by this comment. What is "UB"?

src/field.rs
```diff
@@ -31,6 +44,28 @@ pub struct Fp31(<Self as Field>::Integer);
...
+        let mut t = Self::ONE;
+        for i in (0..Self::Integer::BITS - exp.leading_zeros()).rev() {
+            t *= t;
+            if (exp >> i) & 1 != 0 {
+                t *= *self;
+            }
+        }
+
+        t
+    }
```
benjaminsavage, line 68: Perhaps add a comment to explain what you're doing here? I'm not good at following bitwise logic like this.

src/field.rs
```diff
@@ -143,4 +179,39 @@ mod test {
         assert_eq!(Fp31(<Fp31 as Field>::PRIME - 1), Fp31(0) - Fp31(1));
         assert_eq!(Fp31(0), Fp31(0) * Fp31(1));
     }
+
+    #[test]
+    fn pow() {
```
benjaminsavage, line 184: These are great corner cases to test. Consider adding a basic test, e.g. 2^3 = 8, 3^3 = 27

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+            for i in 1..=n {
+                let mut x = F::ONE;
+                let mut denom = F::ONE;
+
+                let x_i = F::from(u128::from(i));
+                for j in 1..=n {
+                    if i != j {
+                        let x_j = F::from(u128::from(j));
+                        x *= x_j;
+                        denom *= x_j - x_i;
```
benjaminsavage, line 45: I'm never sure what the order of operations is going to be. I'd recommend adding parens around it to make it super explicit:
  denom *= (x_j - x_i);

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    // Evaluates P(0) using the first N points provided, ignoring extra points, if any.
+    fn evaluate_at_zero<T: Clone + Into<F>>(&self, points: &[T]) -> F {
+        points
+            .iter()
+            .zip(self.coefficients.iter())
+            .map(|(y, &x)| y.clone().into() * x)
+            .fold(F::ZERO, |lhs, rhs| lhs + rhs)
+    }
```
benjaminsavage, line 70: This is a bit tricky. The order of the points you provide must be ordered the same as the order of the helpers.

I suggest adding a comment about that maybe?

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+
+        shares
+    }
+
+    /// Reconstructs a secret from a set of size at least k of shares.
+    ///
+    /// # Errors
+    /// Returns an error if there is no enough shares to reconstruct the secret or
+    /// if an error occurred while evaluating the polynomial
+    pub fn reconstruct<F>(shares: &[Share<F>], lagrange: &LagrangePolynomial<F>) -> Result<F, Error>
```
benjaminsavage, line 174: Once again - perhaps add a comment explaining that the order of "shares" matters. It must match the order of the helpers.

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+    #[test]
+    fn fails_to_reconstruct_secret_with_forged_shares() {
+        let (k, n) = (NonZeroU8::new(2).unwrap(), NonZeroU8::new(3).unwrap());
+        let rng = StepRng::new(1, 1);
+        let secret = Fp31::from(5_u128);
+
+        let sharing = SecretSharing::new(k, n).unwrap();
+
+        let mut shares = sharing.split(secret, rng);
+        shares[2] = Share { y: shares[1].y };
```
benjaminsavage, line 347: What exactly is going on here? What do you mean by "forged"?

Also, if k=2, you should only need two shares to reconstruct I think.

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+        rng_seed: u64,
+        k: u8,
+        n: u8,
+    }
+
+    impl ShareReconstructInput {
+        fn gen() -> impl Strategy<Value = Self> {
+            (2u8..Fp31::PRIME)
+                .prop_flat_map(|v| {
+                    let k = Just(v);
```
benjaminsavage, line 382: What does "Just(v)" do?

benjaminsavage (approved):
This looks really good! I can't wait to land this!

src/field.rs
martinthomson: And something like 6^2 = 5.

martinthomson (commented):
I realized that I left a few comments unsubmitted.  A followup might be a good idea.

src/field.rs
```diff
@@ -31,6 +44,28 @@ pub struct Fp31(<Self as Field>::Integer);
 impl Field for Fp31 {
     type Integer = u8;
     const PRIME: Self::Integer = 31;
+    const ZERO: Self = Fp31(0);
+    const ONE: Self = Fp31(1);
+
+    fn invert(&self) -> Self {
```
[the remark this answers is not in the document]
akoshelev: that requires further constraining `Integer` type as we need to be able to subtract `2_u8` from `PRIME` and left shift to compute the `pow`.  I can either depend on `num-traits` to bound `Integer: PrimInt` or create a simple trait for it. I think the latter is slightly better as we don't need the full power of num-traits for now

src/shamir.rs
```diff
@@ -0,0 +1,419 @@
...
+                polynomial_degree: degree.get(),
+            })
+        } else {
+            let mut coefficients = Vec::with_capacity(degree.get() as usize);
+            let n = degree.get();
+            for i in 1..=n {
+                let mut x = F::ONE;
+                let mut denom = F::ONE;
+
+                let x_i = F::from(u128::from(i));
```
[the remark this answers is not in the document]
akoshelev: is it possible to define such bound without dyn? Adding `From<Into<u128>` bound causes a compile error for me

src/field.rs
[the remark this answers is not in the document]
martinthomson: Yeah, I get that.

Would it be easier to make `type PRIME : Field` instead?  Then you could do `Self::PRIME - Self::from(2_u128)`.  Of course, then you have the problem that Self::PRIME isn't in the field...

src/shamir.rs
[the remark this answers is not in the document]
martinthomson: Ok, it was just a theory.  I hadn't tried it.  And when implementing `From<Into<u128>>` was enough to count as implementing `From<u128>`, I thought it might be worth a shot.

src/field.rs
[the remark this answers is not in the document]
akoshelev: I have a workaround for `invert` that only requires `Field: Into<Self::Integer>` and `type Integer: Ord + Sub<Output = Self::Integer>` bounds

```rust
 /// computes the multiplicative inverse of `self`. It is UB if `self` is 0.
    #[must_use]
    fn invert(&self) -> Self {
        debug_assert!(
            self != &Self::ZERO,
            "Multiplicative inverse is not defined for 0"
        );

        self.pow(Self::PRIME - Self::ONE.into() - Self::ONE.into())
    }
```

for `pow` is a little more tricky and making `Prime: Field` is not helpful because we need right shift for `Self::Integer`:

```rust
pub trait Int :
    Sized
    + Copy
    + Debug
    + Ord
    + Sub<Output = Self>
    + Into<u128>
    + Shr<u32, Output = Self>
    + BitAnd<u8, Output = Self>
    + PartialEq
{
    const ZERO: Self;
    const BITS: u32;
}

impl Int for u8 {
    const ZERO: Self = 0_u8;
    const BITS: u32 = u8::BITS;
}

pub trait Field:
{
  type Integer: Int;
  ...
}
```

I don't see any issues here and I can raise a PR if this is acceptable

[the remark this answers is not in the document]
martinthomson: Sgtm. You can implement a default From implementation for the new Int type as well.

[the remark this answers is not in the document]
akoshelev: When implementing `From<Int>` for `Field` we would run into the same issue with dyn :(

src/field.rs
akoshelev: ack

src/shamir.rs
akoshelev: yes, this test ensures that reconstruction fails if we give it bad input such as S0, S1, S1

src/shamir.rs
akoshelev: it is basically a proptest strategy that always produces a single value (given as input). The reason why it is used here is because I wanted to enforce the invariant `n >= k` otherwise input does not make much sense and will be rejected by the secret sharing

src/field.rs
```diff
@@ -143,4 +179,39 @@ mod test {
...
+            let field_element = Fp31(i);
+            assert_eq!(
+                Fp31::ONE,
+                field_element.invert().mul(field_element),
+                "{field_element:?}*1/{field_element:?} != 1"
+            );
+        }
+    }
+
+    #[test]
```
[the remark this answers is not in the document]
akoshelev: Added it, but I don't understand why`cargo test --release` lets this test pass even without it

PR #33, opened 2022-06-20
IPA synthetic data generation for benchmarking IPA attribution accuracy

taikiy:
This command line tool generates synthetic source/trigger events based on real-life sample data. Generated data help us test IPA protocols to see if they are implemented correctly to generate attribution reports, as well as measure realistic benchmark.

I splitted the diff into 3 parts.

1. Command line structopts - 2a06f92
2. Sample data - 6b431ca
3. Data generation - a606358


src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+// "Ads" doesn't mean FB's L3 ads. It could be ads from different businesses.
+
+pub fn generate_events(
+    total_count: u32,
+    epoch: u8,
+    secret_share: bool,
+    seed: &Option<u64>,
+    out: &mut Box<dyn io::Write>,
+) -> (u32, u32) {
+    let mut rng = match seed {
```
akoshelev, line 55: we should probably inject rng here

src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        // TODO: 99.97% queries in ads manager account for L1-3 breakdown only. For now, we'll do 1 ad = 1 breakdown key
+        let ad_id: u32 = rng.gen();
+
+        // Number of unique people who saw the ad
+        let reach = sample.reach_per_ad(&mut rng);
+        debug!("reach: {}", reach);
+
+        // CVR for the ad
+        let cvr = sample.cvr_per_ad_account(&mut rng);
```
akoshelev, line 89: these parameters should be defined as input to the CLI imo.

src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+                epoch,
+                &ad_id.to_string(),
+                secret_share,
+                &sample,
+                &mut rng,
+                &mut ss_rng,
+            );
+
+            for e in events {
+                let json_string = match e {
```
akoshelev, line 121: I think serde supports untagged representation of enums if that is what you want here: https://serde.rs/enum-representations.html

src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+    mks
+}
+
+fn to_secret_share<R: RngCore + CryptoRng>(data: &[u8], rng: &mut R) -> SecretShare {
+    let mut ss: SecretShare = [Vec::new(), Vec::new(), Vec::new()];
+
+    for x in data {
+        let ss1 = rng.gen::<u8>();
+        let ss2 = rng.gen::<u8>();
+        let ss3 = ss1 ^ ss2 ^ x;
```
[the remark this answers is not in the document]
taikiy: Good point. While coding the second part of this task (clear-text attribution), I did some refactoring and moved this to trait. Sorry for the outdated code.

https://github.com/martinthomson/raw-ipa/pull/34/commits/0e7132108ca98243e577aed64cb0e4e1c53552d5#diff-c20549a7190ae4d9a0aa77dd6f73d738a8730543c4012b3b27c5bc8a43db7a90

src/bin/ipa_bench/gen.rs
taikiy: Do you mean passsing a fixed CVR to CLI input?

We'll likely to use this tool to generate millions of synthetic events that are similar to ones in real-life. Each ad have different performance results such as CVR, impressions, conversion values, timestamps, etc. hence we are taking samples during runtime from pre-defined distributions (sample.rs).

src/bin/ipa_bench/cmd.rs
```diff
@@ -0,0 +1,144 @@
...
+        )]
+        random_seed: Option<u64>,
+
+        #[structopt(
+            short,
+            long,
+            default_value = "0",
+            help = "Simulate ads created in this epoch. Impressions and conversions for a given ad may happen in the next epoch."
+        )]
+        epoch: u8,
```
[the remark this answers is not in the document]
taikiy: Epoch is a pre-defined time interval (currently 7 days) in IPA to rotate keys used to blind matchkeys.

src/bin/ipa_bench/cmd.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub common: CommonArgs,
+
+    #[structopt(subcommand)]
+    pub cmd: Command,
+}
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "init")]
+pub enum Command {
+    Init {
```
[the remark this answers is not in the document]
taikiy: Clear-text attribution and reporting will be parts of this CLI with "run" sub-command (#34). I was thinking "init" to prepare the test data which is used for IPA benchmarking. I actually named it "gen" at first as you suggest, but I thought "gen what?". gen-events would be better, but is it a bit long as a sub-command?

src/bin/ipa_bench/cmd.rs
```diff
@@ -0,0 +1,144 @@
...
+        #[structopt(
+            short,
+            long,
+            default_value = "1",
+            help = "Multiply the number of events generated by the scale factor. For example, --scale-factor=100 generates 10,000,000 synthetic events."
```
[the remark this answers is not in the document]
taikiy: It's simply to reduce the number of zeros we need to type. I imagine we at least need a million events to generate a decent report.

src/bin/ipa_bench/gen.rs
taikiy: I'm using the seed to generate two RNGs. I thought it also helps isolation of logics from 'cmd.rs'. Should I inject two RNGs instead?

src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+// TODO: Currently, users are mutually exclusive in each ad loop (i.e. User A in ad X will never appear in other ads).
+// We need to generate events from same users across ads (but how often should a user appear in different ads?)
+// "Ads" doesn't mean FB's L3 ads. It could be ads from different businesses.
+
+pub fn generate_events(
+    total_count: u32,
+    epoch: u8,
+    secret_share: bool,
+    seed: &Option<u64>,
+    out: &mut Box<dyn io::Write>,
```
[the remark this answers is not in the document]
taikiy: Sorry, what is DD?
I added <W: io::Write> in the function signature. Should I do the same for all other definitions (e.g. cmd.rs get_output)?

For my learning, could you explain what is the benefit of using the constraint over just a plain parameter type?

src/bin/ipa_bench/gen.rs
```diff
@@ -0,0 +1,428 @@
...
+    devices: u8,
+    impressions: u8,
+    conversions: u8,
+    epoch: u8,
+    breakdown_key: &str,
+    secret_share: bool,
+    sample: &Sample,
+    rng: &mut R,
+    ss_rng: &mut R,
+) -> Vec<EventType> {
```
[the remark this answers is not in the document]
taikiy: IIUC, returning an iterator from a function means I'll be returning something like .map() that implements std::iter, correct?
I wasn't sure how I can achieve this as this function iterates over 0..impressions and 0..conversions in two separate loops, and each event generation depends on the internal state (last_impression and last_conversion). Maybe I can create a struct and implement Iterator, but that sounds a bit overkill considering that this function only generates ~10,000 events as of now (impressions + conversions per ad < 10^3). Generators could be useful here, but it's in unstable rust. Is there a way to solve this?

taikiy:
@thurstonsand @akoshelev 
Thanks for reviewing! As discussed in the chat, I'll move hard-coded sample data out and create a function to read the config file as input. I'll keep this PR open for now for the sake of other comments, but I'll remove this once the newer version is ready.

src/bin/ipa_bench/gen.rs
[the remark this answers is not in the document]
akoshelev: yea, managing iterator state manually is not desirable unless there is a clear benefit for it.  If we have a hard limit on number of events generated by this function, then it is not necessary to move from `Vec` output.

just fyi, it is possible to create an iterator from function using [`std::iter::from_fn`](https://doc.rust-lang.org/std/iter/fn.from_fn.html), but maybe there is too much of the local state to manage in order to use it

src/bin/ipa_bench/gen.rs
akoshelev: I am curious why it is required to use two RNGs here? if they're seeded with the same initial value, both produce the same output. 

Injecting RNGs may make it easier to test and control the output of RNG as well as pick a different RNG implementation based on caller's need

src/bin/ipa_bench/cmd.rs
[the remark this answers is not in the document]
akoshelev: ah gotcha. as long as we don't need to generate less than base events or support arbitrary number of generated events, we should be fine then

src/bin/ipa_bench/cmd.rs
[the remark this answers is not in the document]
akoshelev: Given that we use a separate CLI to handle these, is it necessary to support subcommands? would something like this work instead?

```bash
  ipa_bench gen-events --output ...
  ipa_bench attribute --input ... --output ...
```

src/bin/ipa_bench/cmd.rs
```diff
@@ -124,6 +151,13 @@ impl Command {
     }
 }
 
+fn get_input(path: &Option<PathBuf>) -> Result<Box<dyn io::Read>, io::Error> {
```
thurstonsand, line 154: I believe @akoshelev mentioned we should not use `&Option` type. here's some [interesting info](https://stackoverflow.com/questions/16504643/what-is-the-overhead-of-rusts-option-type) on some compiler optimizations that can take place. that being said, any reason you couldn't just remove the `&` here?
```suggestion
fn get_input(path: Option<PathBuf>) -> Result<Box<dyn io::Read>, io::Error> {

Comment (thurstonsand) [inline src/bin/ipa_bench/cmd.rs:154]: i wonder if there's a way to avoid boxing the `dyn io::Read` here...

File: src/bin/ipa_bench/config.rs [hunk 11]
```diff
@@ -0,0 +1,134 @@
...
+
+    Config { config }
+}
+
+#[derive(Clone)]
+pub struct Config {
+    config: Value,
+}
+
+impl Config {
```
thurstonsand, line 21: typically for config, I like to extract all the values on init, so that I don't run into a "config missing" error half-way through executing the program. maybe not that big a deal in this case, but I've found that making sure your config is valid up front is generally helpful in the long run

src/bin/ipa_bench/gen_events.rs
```diff
@@ -251,38 +258,89 @@ fn gen_matchkeys<R: RngCore + CryptoRng>(count: u8, rng: &mut R) -> Vec<u64> {
...
     use std::env::temp_dir;
     use std::fs::{self, File};
     use std::io::prelude::*;
-    use std::io::{BufReader, Read, Write};
+    use std::io::{BufReader, Cursor, Read, Write};
     use std::path::PathBuf;
 
+    const DATA: &str = r#"
+      {
+        "devices_per_user": {
```
thurstonsand, line 276: i assume this is dummy data?

src/bin/ipa_bench/sample.rs
```diff
@@ -5,8 +5,8 @@ use std::time::Duration;
 
 use crate::config::Config;
 
-pub struct Sample {
-    config: Config,
+pub struct Sample<'a> {
+    config: &'a Config,
```
thurstonsand, line 9: will `Config` be used elsewhere? why make it a pointer?

src/bin/ipa_bench/cmd.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+fn get_input(path: &Option<PathBuf>) -> Result<Box<dyn io::Read>, io::Error> {
+    match path {
+        Some(ref path) => File::open(path).map(|f| Box::new(f) as Box<dyn io::Read>),
+        None => Ok(Box::new(io::stdin())),
+    }
+}
+
+fn get_output(path: &Option<PathBuf>, overwrite: bool) -> Result<Box<dyn io::Write>, io::Error> {
```
martinthomson, line 163: Can this be a method on CommonArgs?

src/bin/ipa_bench/config.rs
```diff
@@ -0,0 +1,127 @@
+use log::error;
+use rand_distr::num_traits::ToPrimitive;
+use serde_json::Value;
+use std::io::{BufReader, Read};
+use std::ops::Range;
+use std::process;
+
+#[derive(Clone)]
```
martinthomson, line 8: Why isn't this directly serializable?  I know that you would need to use named types rather than tuples, but that's probably a good thing, isn't it?

It seems like you get a lot of custom serialization code here, where this could be annotations on the types instead.

The only caveat is for bounds checking, which I don't see in your code.  Most of the f64 values should be positive (not zero or negative), but that isn't checked.  CVR should be a value between 0 and 1, but you don't check that.

It's OK if a tool like this crashes when being provided bad input, but it's something to note in documentation.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+pub enum EventType {
+    // Source event in clear
+    S(SourceEvent),
+    // Trigger event in clear
+    T(TriggerEvent),
+    // Source event in cipher if --secret-share option is enabled
+    ES(ESourceEvent),
+    // Trigger event in cipher if --secret-share option is enabled
+    ET(ETriggerEvent),
+}
```
martinthomson, line 50: I think that you want Event, and EncryptedEvent as discrete types.  Also, this isn't a *Type enum as it contains details of the event, not just the type.

Ultimately, there might be three values here:
1. The raw event, with values included directly.
2. A number of secret-shared events.
3. Encrypted secret-shared events.

Having explicit types for each is probably best.

This tool can concentrate on the first step: generating synthetic values for (1).

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+use std::io;
+use std::time::Duration;
+
+const DAYS_IN_EPOCH: u64 = 7;
+
+#[derive(Clone)]
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+pub struct Event {
+    // For this tool, we'll fix the length of a matchkey to u64
+    pub matchkeys: Vec<u64>,
```
martinthomson, line 22: Can you use a type alias for MatchKey?

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+    pub epoch: u8,
```
martinthomson, line 23: A newtype for epoch would be helpful, I think.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+pub struct SourceEvent {
+    pub event: Event,
+    pub breakdown_key: String,
+}
+
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+pub struct TriggerEvent {
+    pub event: Event,
+    pub value: u32,
+    pub zkp: String,
```
martinthomson, line 37: ?

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+    let mut rng = match seed {
+        None => ChaCha20Rng::from_entropy(),
+        Some(seed) => ChaCha20Rng::seed_from_u64(*seed),
+    };
+    debug!("seed: {:?}", rng.get_seed());
+
+    // Separate RNG for generating secret shares
+    let mut ss_rng = match seed {
+        None => ChaCha20Rng::from_entropy(),
+        Some(seed) => ChaCha20Rng::seed_from_u64(*seed),
```
martinthomson, line 81: Use rand::rngs::StdRng here.  No point in pulling in an explicit dependency.  That means that you don't get seed_from_u64, so you need to change how you invoke it.  You might need to print the seed in hex (and do some hex decoding) in that case.

Note that it is possible to have the hex conversion done by structopt: https://github.com/martinthomson/ohttp/blob/e27ee1af5c698091105c43cf86028ec86b2cbdb7/ohttp-client/src/main.rs#L12-L26

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+    out: &mut W,
+) -> (u32, u32) {
+    let mut rng = match seed {
+        None => ChaCha20Rng::from_entropy(),
+        Some(seed) => ChaCha20Rng::seed_from_u64(*seed),
+    };
+    debug!("seed: {:?}", rng.get_seed());
+
+    // Separate RNG for generating secret shares
+    let mut ss_rng = match seed {
```
martinthomson, line 79: Clippy will suggest that you do this instead:

```rust
let mut rng = if let Some(seed) = seed {
  StdRng::from_seed(seed)
} else {
  StdRng::new()
}
```

Also, I would move this code outside this function and have this function accept an RNG as input.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -0,0 +1,481 @@
...
+                sample,
+                &mut rng,
+                &mut ss_rng,
+            );
+
+            total_impressions += impressions.to_u32().unwrap();
+            total_conversions += conversions.to_u32().unwrap();
+
+            for e in events {
+                out.write_all(serde_json::to_string(&e).unwrap().as_bytes())
```
martinthomson, line 141: How are events delineated in the file?  It looks like you have an EOL on each, but that relies on serde_json not adding an EOL.  Try ASCII RS instead: https://datatracker.ietf.org/doc/html/rfc7464

src/bin/ipa_bench/gen_events.rs
thurstonsand: or, it's nice to use `map_or`:
```rust
let mut rng = seed.map_or(ChaCha20Rng::from_entropy(), ChaCha20Rng::seed_from_u64)

PR #35, opened 2022-06-24
Add MPC helper CLI

akoshelev:
## Overview

CLI starts a http or https server that currently only handles `/echo` requests. For TLS it uses self-signed certificate hidden behind a feature flag to prevent accidental usage in production configuration

We would use this endpoint in the future to communicate with MPC helpers.

## Testing

CLI picks up a random port if `--port` argument is omitted

```bash
☻  target/debug/helper -vvv                                                                                                                                                                             helper-service a52606f ✗
INFO - Logging setup at level 3
INFO - listening to http://127.0.0.1:52634
```

It understands `-p` argument

```bash
☻  target/debug/helper -vvv -p 1867                                                                                                                                                                     helper-service a52606f ✗
INFO - Logging setup at level 3
INFO - listening to http://127.0.0.1:1867
```

if compiled with `self-signed-certs` feature, CLI also supports HTTPS

```
☻  cargo build --features self-signed-certs && target/debug/helper -vvv -s https                                                                                                                        helper-service a52606f ✗
    Finished dev [unoptimized + debuginfo] target(s) in 0.29s
INFO - Logging setup at level 3
INFO - listening to https://127.0.0.1:52687
```

Server can echo the request back

```
☻  wget -q -nv "https://127.0.0.1:1867/echo?foo=bar" --no-check-certificate -O - | jq                                                                              helper-service 2e624c7 ✗
{
  "query_args": {
    "foo": "bar"
  },
  "headers": {
    "accept-encoding": "identity",
    "user-agent": "Wget/1.21.3",
    "accept": "*/*",
    "connection": "Keep-Alive",
    "host": "127.0.0.1:1867"
  }
}

```


src/net/server/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    BadQueryString(#[from] QueryRejection),
+}
+
+impl IntoResponse for MpcServerError {
+    fn into_response(self) -> Response {
+        let status_code = match &self {
+            MpcServerError::BadQueryString(_) => StatusCode::BAD_REQUEST,
+        };
+
+        (status_code, self.to_string()).into_response()
```
thurstonsand, line 28: curious if there's a "standard" way of converting between types in rust. should you prefer `into*` methods or `from*` methods? for instance, in this case does a `Response::from` exist? if so, why use `into_response()` over that?

src/net/server/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+/// Axum router definition for MPC helper endpoint
+#[must_use]
+pub fn router() -> Router {
+    Router::new().route("/echo", get(handlers::echo_handler))
```
thurstonsand, line 35: so the idea is that you would add a route here eventually for MPC comms?

src/net/server/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+                    .handle(handle)
+                    .serve(svc)
+                    .await
+                    .expect("Failed to serve");
+            }
+        }),
+    };
+
+    (
+        handle
```
thurstonsand, line 75: what does it mean for you to return `handle` and `task_handle` (which contains a `handle.clone()`) here?

src/net/server/mod.rs
akoshelev: I suspect that the reason for this custom trait to exist is to be able to do dynamic dispatch inside Axum as `From` trait is not object-safe.

in my experience `.into` is the preferred way to convert, but it often not possible because `into` method is not generic, so you fall back to `T::from(..)`

src/net/server/mod.rs
akoshelev: that's right, however we can also create many endpoints with different routers attached to them.

src/net/server/mod.rs
akoshelev: I return `SocketAddr` here in case if clients don't want to specify a port to bind to, in this case Kernel pick it up for us, `Handle` type is here to be able to get this information (it only becomes available after Kernel signals us which port is assigned). `JoinHandle` is the Tokio task handle that runs the HTTP(s) server, clients may want to abort it or (like CLI) just await it until user terminates it

thurstonsand: ok, think I got it.

so the `.listening()` function is async. i.e. it returns when listening has started, not ended.

what do you expect a caller of `bind` to do with the returned `SocketAddr`?

akoshelev: they should advertise it as the address they're listening to - for example, in unit tests I don't want to specify a port to bind to avoid "address already in use" error and let kernel to find any port that is available. After that, I can connect to this address using an HTTP client and make requests

thurstonsand: ok, gotcha. so mostly if you are going to subsequently connect a client to that server which has a randomly assigned port

PR #36, opened 2022-07-04
CLI for testing IPA scenarios on MPC helpers

taikiy:
## Note
This diff builds on top of PR #35. New changes are in commit 228783b277a54d77ac072a2b2d882750554f9229.

## Overview
CLI starts a http client that connects to "/echo" endpoint on a host specified in the command-line argument. This is the first iteration of creating a testing tool where we'll be able to execute multiple test scenarios of IPA protocols.

CLI assumes that a helper server is already up and running. In the next iteration, however, we'll add a feature which will start servers and communicate with them without needing to specify specific host addresses. At which point, we'll also add E2E test cases.

## CLI execution example
```
❯ ./target/debug/test_mpc -vvv -u "http://localhost:50951"
INFO - Logging setup at level 3
DEBUG - resolving host="localhost"
DEBUG - connecting to [::1]:50951
DEBUG - connecting to 127.0.0.1:50951
DEBUG - connected to 127.0.0.1:50951
DEBUG - flushed 55 bytes
DEBUG - parsed 3 headers
DEBUG - incoming body is content-length (67 bytes)
DEBUG - incoming body completed
DEBUG - pooling idle connection for ("http", localhost:50951)
Response: "{\"query_args\":{\"foo\":\"hello\"},\"headers\":{\"host\":\"localhost:50951\"}}"
```


src/net/client/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        match command {
+            Command::Echo(s) => self.echo(&s).await,
+        }
+    }
+}
+
+impl MpcHttpConnection {
+    #[must_use]
+    pub fn new(addr: &str) -> Self {
+        // this works for both http and https
```
akoshelev, line 40: I am curious how you made it work with self-signed certificate, I had to explicitly disable certificate validation because hyper client just refused to establish connection, so I had to do this

```
let conn = TlsConnector::builder()
            .danger_accept_invalid_certs(true)
            .build()
            .unwrap();
```

taikiy: For this PR, I assumed I'd only connect to CA-signed sites. I'll add a self-signed cert connection in the next PR.

PR #37, opened 2022-07-05
First look at implementing a step of the pipeline

thurstonsand:


martinthomson (commented):
Using streams looks promising.  Your example shows a stream in, but a singular value out though.  Is the goal to define steps in a graph and plumb inputs and outputs as streams where appropriate?

thurstonsand:
in this case, I was imagining the "steps" are simply the operations on a `Stream`: `map`, `filter,` `fold`, etc.

Thus, the entire pipeline would likely take a stream as input and produce an aggregate value as output. That being said, it very well could make sense to chain multiple pipelines back-to-back, where the output of one pipeline is a stream consumed as the input of the next pipeline. I'll see if I can write something up to see what that looks like

src/pipeline/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    where
+        Self: Sized,
+        Self::Item: Default,
+    {
+        IncrementerAdder::new(self)
+    }
+}
+
+/// magic so that any [`Stream`] implements [`PipelineExt`]
+impl<T: Stream + ?Sized> PipelineExt for T {}
```
akoshelev, line 35: We may just remove `?Sized` as all methods require `Self: Sized` anyway

src/pipeline/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+
+// this is the second of a pair of pipelines to process a stream of data.
+// in this stage, sum all values, and then to_string() the final sum.
+// the output of this pipeline is a string of the sum
+pub struct SumAndStringify {}
+impl SumAndStringify {
+    pub async fn run(&self, inp: impl Stream<Item = i32>) -> String {
+        // fold provides a way to accumulate all values at the end of computation.
+        // in this case, sum them
+        inp.fold(0, |acc, i| async move { acc + i })
```
akoshelev, line 66: it may be interesting to implement the summation as a stream transform and see how it looks like - it will be quite different from `IncrementerAdder`

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,60 @@
...
+impl<St: Stream> Stream for IncrementerAdder<St>
+where
+    St::Item: Copy + From<u8> + Add<St::Item, Output = St::Item>,
+{
+    type Item = St::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.project();
+        let res = ready!(this.stream.as_mut().poll_next(cx));
+        *this.inc = *this.inc + 1u8.into();
```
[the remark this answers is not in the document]
thurstonsand: true, except I need `Add` right below for the `Poll::Ready` statement

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,60 @@
...
+/// e.g.
+/// input stream: \[0, 0, 0, 0, 0\]
+/// output stream: \[1, 2, 3, 4, 5\]
+///
+/// This is not very ergonomic, and requires a deep understanding of types, as well as [Pin] and
+/// `pin_project`. The tradeoff is we can transform streams in a custom way; the primary benefit
+/// of a custom transform is that it can be stateful, as seen here with an incrementer.
+impl<St: Stream> Stream for IncrementerAdder<St>
+where
+    St::Item: Copy + From<u8> + Add<St::Item, Output = St::Item>,
```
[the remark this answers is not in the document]
thurstonsand: man seems like there's never ending features in rust haha

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// initialize with the default zero value of the item being incremented
+    pub fn new(stream: St) -> Self {
+        Self {
+            stream,
+            inc: St::Item::default(),
+        }
+    }
+}
+
+/// types get complicated here.
```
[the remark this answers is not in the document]
thurstonsand: the most complicated part of `poll_next` is `project`, all the machinery behind that, and why it's necessary. it's possible to write `poll_next` without fully understanding that, but you run the risk of hitting an edge case and having no idea what to do

src/pipeline/mod.rs
thurstonsand: done. they are actually pretty similar after all. I got to use `AddAssign` for this one, but i also had to include `Clone` (or `Copy`)

src/pipeline/incrementer_adder.rs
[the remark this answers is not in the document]
akoshelev: yep, these are not mutually exclusive, so it is possible to define both at the same time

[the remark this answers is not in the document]
thurstonsand: sure. I figured if I already had `Add`, not worth the trouble and added complexity to include `AddAssign` when I can just make it work in the code. but let me know if it's more idiomatic to include the `AddAssign`

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,60 @@
...
+impl<St: Stream> Stream for IncrementerAdder<St>
+where
+    St::Item: From<u8> + for<'a> Add<&'a St::Item, Output = St::Item>,
+{
+    type Item = St::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.project();
+        let res = ready!(this.stream.as_mut().poll_next(cx));
+        let one: St::Item = 1u8.into();
```
martinthomson, line 52: I prefer this if at all possible (sometimes you can't, but here you explicitly can):

```suggestion
        let one = St::Item::from(1u8);
```

Shouldn't the type be something with a bit more space for a count though?  `usize`, say.

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,60 @@
...
+        *this.inc = one + this.inc;
```
martinthomson, line 53:
```suggestion
        *this.inc += one;
```

You should be able to tweak the constraint.

martinthomson (approved):
Worth a try.  Let's see how this works when we have to use it in anger.  There, I have concerns about how you might need to structure code to use this.  But if we are building an assembly line, this seems to work nicely.

If we are going to try it, we might as well merge it.

src/pipeline/incrementer_adder.rs
thurstonsand: done

src/pipeline/incrementer_adder.rs
thurstonsand: also, I implemented the fix

thurstonsand:
i've addressed all your comments @martinthomson . think it's ready to merge (I don't have write access to do it myself)

src/pipeline/incrementer_adder.rs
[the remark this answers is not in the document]
akoshelev: It is certainly up to you here, it is very common for types that implement `Add` to implement `AddAssign` as well, so it shouldn't cost you anything to add this bound.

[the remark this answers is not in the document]
thurstonsand: went ahead and added it

src/pipeline/incrementer_adder.rs
```diff
@@ -0,0 +1,57 @@
+use futures::{ready, Stream};
+use pin_project::pin_project;
+use std::ops::{Add, AddAssign};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// implementation of how `incrementer_adder` works
+/// holds the stream being consumed, and a stateful incrementer that increases with every item
+#[pin_project]
+pub struct IncrementerAdder<St: Stream> {
```
akoshelev, line 10: should we hide these types behind a `#[cfg(test)]` attribute? They don't seem to be useful for actual work we'll be doing for IPA and probably serve just one purpose to show the idea how to work with streams, so we may as well hide them from global namespace

src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod error;
 pub mod field;
 pub mod helpers;
 pub mod net;
+pub mod pipeline;
```
martinthomson, line 9: Alex suggested that we keep the toy project as test-only.  I think that this is the place where you would do that.

```suggestion
#[cfg(test)]
pub mod pipeline;
```

martinthomson (approved):
I will add `#[cfg(test)]`.

PR #38, opened 2022-07-07
Generating pairs of replicated secret shares of 0 or 1

benjaminsavage:
In order to implement the sorting step, we need a secret sharing (in Z_p) of each bit of the match key.

The browser will emit XOR secret shares. We need to convert these from secret shares in Z_2 to secret shares in Z_p.

The way to do that is described in this paper: (https://eprint.iacr.org/2018/387.pdf). See the algorithm 5.2 for modulus conversion from Z_2 to Z_p. 

This algorithm relies on functionality "Frand", in which the helpers generate a random number r from the set {0, 1}, and produce two secret sharings of it:
1. One in Z_2
2. One in Z_p

The authors describe how to generate these secret sharings of r in the appendix. There are some errors in that text. The optimized algorithm implemented here is taken from Appendix F in the paper "Adam in Private: Secure and Fast Training of Deep Neural Networks with Adaptive Moment Estimation". 

As it says there:

"Note that in Protocol 14, the required communication complexity for computing [[r_1 ⊕ r_2 ⊕ r_3]] in Step 3 and 4 is only 3 |p| although two multiplications are used.
We use the IKHC multiplication protocol [22] in which
P_i on input [[a]]_i = (a_i, a_i+1) and [[b]]_i = (b_i, b_i+1) sends
d_i := a_i*b_i+1 + a_i+1*b_i − s_i+2,i to P_i+1, where s_i+2,i is a random value shared between P_i and P_i+2. 
At the start of Step 3, P_1 has [[r_1]]_1 = (r1, 0) and [[r_2]]_1 = (0, r2); similarly, P2 has (0, 0) and (r2, 0), and P3 has (0, r1) and (0, 0). 
Multiplying [[r_1]] with [[r_2]], d_i = −s_i+2,i for i = 2, 3, means that d_2 and d_3 are publicly known to all the parties. 
Therefore, these values can be constants, e.g., 0, and do not need to be sent. Similarly, when multiplying [[r1 ⊕ r2]] with [[r_3]], P_1 has [[r_3]]_1 = (0, 0) and d1 = −s3,1. 
By using the same technique, P_1 sends nothing. The elimination of these communications results in a total communication of 3 |p|, since the communication costs of the first and second XORs are |p| and 2 |p|."


src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    use crate::modulus_convert::modulus_conversion::mult_r1_r2_helper1;
+    use crate::modulus_convert::modulus_conversion::mult_r1_r2_helper2;
+    use crate::modulus_convert::modulus_conversion::mult_r1_r2_helper3;
+
+    use crate::modulus_convert::modulus_conversion::xor_r1_r2_r3_helper1;
+    use crate::modulus_convert::modulus_conversion::xor_r1_r2_r3_helper2;
+    use crate::modulus_convert::modulus_conversion::xor_r1_r2_r3_helper3;
+
+    use crate::modulus_convert::modulus_conversion::sub_mod_p;
```
benjaminsavage, line 219: I assume there is a much better way to do this =). What's the Rust best practice for this? Should I put all of these methods on some kind of object and then import that object? If so, what type of object should I use?

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
+pub mod modulus_conversion {
+
+    use rand::Rng;
+    use rand::RngCore;
+
+    pub struct ReplicatedBinarySecretSharing {
```
martinthomson, line 6: Would it make more sense to define this as:

```rust
pub struct ReplicatedBinarySecretShare((pub u8, pub u8));
```

Then you access each with `v.0` or `v.1`, which is equally obtuse, but more compact.

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    pub struct ReplicatedSecretSharing {
+        pub sh1: u64,
+        pub sh2: u64,
+    }
+
+    pub fn gen_r_binary_sharings<R: RngCore>(
+        mut rng_1: R,
+        mut rng_2: R,
```
martinthomson, line 18: Rather than passing in just the RNGs that you need, would it make more sense for there to be ambient information that had `rng_left` (for $\mathsf{P}_{i-1}$ and `rng_right` (for $\mathsf{P}_{i+1}$)? 

Edit: Gah, no math support on comments... :(

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+                    mult_mod_p(2, add_mod_p(d_2, s_2_3, p), p),
+                    p,
+                ),
+            });
+            values_to_share.push(d_2);
+        }
+        (output_shares, values_to_share)
+    }
+
+    pub fn xor_r1_r2_r3_helper3<R: RngCore>(
```
martinthomson, line 135:
```suggestion
    pub fn xor_r1_r2_r3_helper3<R: Rng>(
```

You are using Rng functions.

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+    pub fn add_mod_p(a: u64, b: u64, p: u64) -> u64 {
+        let offset: u64 = u64::MAX - p + 1;
+
+        let (sum, overlow) = a.overflowing_add(b);
+        if overlow {
+            return sum + offset;
+        } else if sum >= p {
+            return sum - p;
+        }
+        sum
```
martinthomson, line 175:
```suggestion
        if overlow {
            sum + offset
        } else if sum >= p {
            sum - p
        } else {
            sum
        }
```

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    // a and b should both be less than p
+    #[must_use]
+    pub fn sub_mod_p(a: u64, b: u64, p: u64) -> u64 {
+        let offset: u64 = u64::MAX - p + 1;
+
+        let (difference, overlow) = a.overflowing_sub(b);
+        if overlow {
+            return difference - offset;
```
martinthomson, line 185: As above.  Rust people don't like using `return`.  (It takes some getting used to)

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+        }
+        difference
+    }
+
+    // a and b should both be less than p
```
martinthomson, line 190: debug_assert!() is your friend.

martinthomson (commented):
As I mentioned in chat, we probably want a Field trait, with per-prime implementations.  We won't need many of these, but specializations will enable optimizations.  Let's look at the basic stuff here and merge this, after which we can iterate.

src/modulus_convert.rs
martinthomson: A lot of those effectively operate on bignums.  There's no reason for us to go beyond u64 (or much beyond it) for the base value.

I'm concerned at ff not providing constant-time code.  That is probably something we need here, unfortunately.

src/modulus_convert.rs
thurstonsand: is it intentionally a tuple of a tuple? aka the double parenthesis `((pub u8, pub u8))`. access would look like:

```rust
let share = ReplicatedBinarySecretShare((0, 1));
let first = share.0.0;
let second = share.0.1;
```

src/modulus_convert.rs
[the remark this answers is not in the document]
thurstonsand: or, I've seen `use super::*` in tests

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+                16_523_972_853_534_698_712_u64,
+                18_446_744_073_709_551_557_u64
+            ),
+            1_263_076_288_923_396_945_u64,
+        );
+    }
+
+    #[test]
+    fn test_three_helpers() {
+        // let p = 2147483647; // 2^31 - 1
```
thurstonsand, line 421: ?

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+        // let p = 2305843009213693951; // 2^61 - 1
+        let p = 18_446_744_073_709_551_557; // 2^64 - 59
+        let batch_size = 100;
+
+        // Helper 1 and 3 will share this seed
+        let seed_1: [u8; 32] = [1; 32];
```
thurstonsand, line 427: might consider slightly more compact
```suggestion
        let seed_1 = [1u8; 32];
```

src/modulus_convert.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        // Helper 2 and 1 will share this seed
+        let seed_2: [u8; 32] = [2; 32];
+
+        // Helper 3 and 2 will share this seed
+        let seed_3: [u8; 32] = [3; 32];
+
+        // Helper 1
+        let mut helper1_rng_1 = StdRng::from_seed(seed_1);
+        let mut helper1_rng_2 = StdRng::from_seed(seed_2);
```
thurstonsand, line 437: might consider using `clone` to make it super clear what your intent is. something like:
```rust
let seed_1 = [1u8; 32];
let mut helper1_rng_1 = StdRng::from_seed(seed_1);
let mut helper3_rng_1 = helper1_rng_1.clone();

let seed_2 = [2u8; 32];
let mut helper1_rng_2 = StdRng::from_seed(seed_2);
let mut helper2_rng_2 = helper1_rng_2.clone();

let seed_3 = [3u8; 32];
let mut helper2_rng_3 = StdRng::from_seed(seed_3);
let mut helper3_rng_3 = helper2_rng_3.clone();

benjaminsavage: I've switched over to using Martin's class now =)

martinthomson (commented):
I just want to check: did git eat the optimization or has this been moved out?

benjaminsavage:
Git didn't eat the optimization - It was never in this PR. I had tried (and failed) to make a stacked diff. I'm sorry - I think I'll abandon this PR and try to clean up my Git history to make things cleaner.

PR #39, opened 2022-07-08
Migrate to `tracing` crate

akoshelev:
## Overview

Tracing crate has all features we're using currently (print to stderr, verbosity, format) and is required to instrument MPC helpers with telemetry required to understand the performance of it and different protocols we'll be experimenting with.

This change also adds some pretty basic tracing of HTTP requests by using `tower-http` crate.

`stderrlog` crate is removed as we no longer need to use it. Besides that, it has a security advisory against it

```
Crate:     time
Version:   0.1.44
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.44
└── chrono 0.4.19
    └── stderrlog 0.5.3
        └── raw-ipa 0.1.0

error: 1 vulnerability found!
```

## Testing

terminal window 1

```bash
 cargo run --features self-signed-certs --bin helper -- -s https -p 3000 -vvv                                                                                                                                                           helper-service 6d5c2f5
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/helper -s https -p 3000 -vvv`
 INFO raw_ipa::cli::verbosity: Logging setup at level debug
 INFO helper: listening to https://127.0.0.1:3000
DEBUG rustls::server::hs: decided upon suite TLS13_AES_256_GCM_SHA384
DEBUG hyper::proto::h1::io: parsed 5 headers
DEBUG hyper::proto::h1::conn: incoming body is empty
DEBUG request{method=GET uri=/echo?foo=bar version=HTTP/1.1}: tower_http::trace::on_request: started processing request
DEBUG request{method=GET uri=/echo?foo=bar version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
DEBUG hyper::proto::h1::io: flushed 270 bytes
DEBUG hyper::proto::h1::conn: read eof
DEBUG rustls::conn: Sending warning alert CloseNotify
```

terminal window 2

```bash
wget -q -nv "https://127.0.0.1:3000/echo?foo=bar" --no-check-certificate -O -
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,10 +6,10 @@ edition = "2021"
 
 [features]
 default = ["debug", "cli"]
-cli = ["enable-serde", "stderrlog", "structopt", "web-app"]
+cli = ["enable-serde", "structopt", "web-app", "tracing-subscriber"]
 debug = ["hex"]
 enable-serde = ["serde", "serde_json", "rust-elgamal/enable-serde"]
-web-app = ["tokio", "axum", "axum-server", "hyper", "hyper-tls"]
+web-app = ["tokio", "axum", "axum-server", "hyper", "hyper-tls", "tower-http"]
 self-signed-certs = ["hyper-tls"]
 
 [dependencies]
@@ -27,14 +27,18 @@ serde = {version = "1.0", optional = true}
 serde_json = {version = "1.0", optional = true}
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, so pin this
 sha2 = "0.9"
-stderrlog = {version = "0.5", optional = true}
 structopt = {version = "0.3", optional = true}
 thiserror = "1.0"
+tracing = "0.1.35"
 tokio = { version = "1.19.2", optional = true, features = ["rt", "rt-multi-thread", "macros"] }
 axum = { version = "0.5.7", optional = true, features = ["http2"] }
 hyper = { version = "0.14.19", optional = true, features = ["client", "h2"] }
 hyper-tls = { version = "0.5.0", optional = true }
 axum-server = { version = "0.4.0", optional = true, features = ["rustls", "rustls-pemfile", "tls-rustls"] }
+tracing-subscriber = { version = "0.3.14", optional = true }
+tower-http = { version = "0.3.4", optional = true, features = ["trace"] }
+
+
 
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,33 +13,30 @@ web-app = ["tokio", "axum", "axum-server", "hyper", "hyper-tls", "tower-http"]
 self-signed-certs = ["hyper-tls"]
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -27,14 +27,18 @@ serde = {version = "1.0", optional = true}
...
 structopt = {version = "0.3", optional = true}
 thiserror = "1.0"
+tracing = "0.1.35"
 tokio = { version = "1.19.2", optional = true, features = ["rt", "rt-multi-thread", "macros"] }
 axum = { version = "0.5.7", optional = true, features = ["http2"] }
 hyper = { version = "0.14.19", optional = true, features = ["client", "h2"] }
 hyper-tls = { version = "0.5.0", optional = true }
 axum-server = { version = "0.4.0", optional = true, features = ["rustls", "rustls-pemfile", "tls-rustls"] }
+tracing-subscriber = { version = "0.3.14", optional = true }
+tower-http = { version = "0.3.4", optional = true, features = ["trace"] }
```
martinthomson, line 39: I think that we should continue to sort dependencies.  Since you added the out-of-order ones, maybe you can fix them now :)

martinthomson, line 39: I think that we should continue to sort dependencies.  Since you added the out-of-order ones, maybe you can fix them now :)

src/cli/verbosity.rs
```diff
@@ -1,5 +1,8 @@
-use log::info;
+use std::io;
```
martinthomson, line 1:
```suggestion
use std::io::stderr;
```

src/cli/verbosity.rs
```diff
@@ -12,18 +15,29 @@ pub struct Verbosity {
...
-            .timestamp(stderrlog::Timestamp::Off)
-            .init()
-            .unwrap_or_else(|e| {
-                if !self.quiet {
-                    eprintln!("unable to configure logging: {:?}", e);
-                }
-            });
-        info!("Logging setup at level {}", self.verbose);
+        let filter_layer = from_verbosity(self.quiet, self.verbose);
+        let fmt_layer = fmt::layer().without_time().with_writer(io::stderr);
```
martinthomson, line 35:
```suggestion
        let fmt_layer = fmt::layer().without_time().with_writer(stderr);
```

src/cli/verbosity.rs
```diff
@@ -12,18 +15,29 @@ pub struct Verbosity {
     verbose: usize,
 }
 
+fn from_verbosity(quiet: bool, verbose: usize) -> LevelFilter {
```
martinthomson, line 18: Move to Verbosity as a private method.

I'd suggest something, but GitHub doesn't let me.

akoshelev: oh, I don't know why I didn't do it in the first place. Thanks for catching it!

thurstonsand:
LGTM

PR #40, opened 2022-07-08
Outline of a PRSS framework

martinthomson:
This uses x25519 to exchange secrets, HKDF (SHA-256) to derive secrets,
and AES as the PRF.  I've only implemented the ability to generate a
u128 with random access and bits with sequential access so far.

The testing for the endpoint part is incomplete too.

@benjaminsavage let me know if this (or something like it) is in a usable shape.

We need to work out the field math still, but once I've done a bit more testing on this then I'll start to build out some basic fields.


Cargo.toml
```diff
@@ -13,12 +13,17 @@ web-app = ["tokio", "axum", "axum-server", "hyper", "hyper-tls"]
...
+byteorder = "1"
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, not 0.10
 digest = "0.9"
 hex = {version = "0.4", optional = true}
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, so pin this
 hkdf = "0.11"
 log = "0.4"
+# This is stupid, but we have packages that want this old interface
+# those have to use the same RNG as packages that want the new interface.
+old_rand_core = { package = "rand_core", version = "0.5", default-features = false }
```
akoshelev, line 26: should we use https://crates.io/crates/x25519-dalek/2.0.0-pre.1 that uses [0.6](https://github.com/dalek-cryptography/x25519-dalek/pull/64)?

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+    }
+}
+
+/// The basic generator.  This generates values based on an arbitrary index.
+pub struct Generator {
+    cipher: Aes256,
+}
+
+impl Generator {
+    /// Generate the value at the given index.
```
akoshelev, line 173: I don't have the full context here, how the values of `index` are picked by the clients of this API? do they set an initial value and then just monotonically increase it or pick at random using a PRNG?

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+/// This intermediate object exists so that multiple generators can be constructed,
+/// with each one dedicated to one purpose.
+pub struct GeneratorFactory {
+    kdf: Hkdf<Sha256>,
+}
+
+impl GeneratorFactory {
+    /// Create a new generator using the provided context string.
+    /// # Panics
+    /// Never: we don't let that happen.
```
thurstonsand, line 156: if it never panics, do you need to mention it? is it because `clippy` caught the `.unwrap()`?

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+use aes::{
+    cipher::{generic_array::GenericArray, BlockEncrypt, KeyInit},
+    Aes256,
+};
+use byteorder::{ByteOrder, NativeEndian};
+use hkdf::Hkdf;
+use rand::{CryptoRng, RngCore};
+use sha2::Sha256;
+use x25519_dalek::{EphemeralSecret, PublicKey};
+
```
thurstonsand, line 11: any paper or write-up you can link to describing the goal of the math being done here?

Cargo.toml
martinthomson: That might be easier than what I did, yeah.

src/prss.rs
martinthomson: Drawing at random is generally a poor choice. I was thinking that the number would be structured.  You have some number of bits dedicated to a record counter, some number of bits dedicated to a step number, or something like that.

There might need to be a layer on top that makes this part a bit easier to manage, I don't know.

src/prss.rs
martinthomson: It's because clippy found the unwrap without understanding the conditions.

src/prss.rs
martinthomson: I should do that write-up.  Let me get back to you on that though.

src/field.rs
```diff
@@ -0,0 +1,131 @@
...
+    use super::Fp31;
+
+    #[test]
+    fn fp31() {
+        let x = Fp31(24);
+        let y = Fp31(23);
+
+        assert_eq!(Fp31(16), x + y);
+        assert_eq!(Fp31(25), x * y);
+        assert_eq!(Fp31(1), x - y);
```
benjaminsavage, line 125: Suggest adding a few more test cases like 0 + 0 and 0 - 0 and 1 + 0, 0 - 1, etc.

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+    /// Generate two random field values, one that is known to the left helper
+    /// and one that is known to the right helper.
+    #[must_use]
+    pub fn generate_fields<F: Field>(&self, index: u128) -> (F, F) {
+        let (l, r) = self.generate_values(index);
+        (F::from(l), F::from(r))
+    }
+
+    /// Generate an additive share of zero.
```
benjaminsavage, line 36: This one is subtle. probably merits a comment explaining how it works. Took me a second to work it out =).

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    /// Generate additive shares of zero in a field.
+    #[must_use]
+    pub fn zero<F: Field>(&self, index: u128) -> F {
+        let (l, r): (F, F) = self.generate_fields(index);
+        l - r
+    }
+
+    /// Generate additive shares of zero in a field.
```
benjaminsavage, line 70: copy paste error for this comment I think

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+    /// Generate an additive shares of a random value.
+    #[must_use]
+    pub fn random_u128(&self, index: u128) -> u128 {
+        let (l, r) = self.generate_values(index);
+        l.wrapping_add(r)
+    }
+
+    /// Generate the next share in `ZZ_2`
+    #[must_use]
+    pub fn next_zero_bit_share(&mut self) -> bool {
```
benjaminsavage, line 59: Might we also want a bitwise share which isn't always zero?

src/prss.rs
```diff
@@ -0,0 +1,493 @@
...
+}
+
+/// A generator for a single bit.  Unlike the base generator, this is a stateful object.
+pub struct BitGenerator {
+    /// The underlying generator.
+    g: Generator,
+    /// The current index, shifted left by 7 bits, plus a 7 bit index.
+    i: u128,
+    /// The value we got from the current index (this is outdated if `i & 0x7f == 0`).
+    v: u128,
```
benjaminsavage, line 191: I don't get what is going on with the index =(

benjaminsavage (approved):
This looks great. Definitely usable. I'd love to get this landed so that I can build on top of it!

martinthomson (commented):
@thurstonsand asked for a write-up.  Here's a short one.

In summary, the process establishes pairwise shared secrets between all participants.  It then uses a KDF to extract entropy from that secret and produce uniformly random, but distinct outputs for different uses.  From those outputs it seeds a PRF.

The idea is that you want to generate a shared secret between each pair of participants.  For that, I'm using X25519.  Each participant generates a key pair and publishes its public key.  I'm having each participant publish two keys, one for the participant to the left, one for the participant to the right.

From those 6 key pairs, we match up each and produce 3 shared secrets.  

The output of this is not uniformly random (in X25519, the resulting value is from the large subgroup in the field).  So we extract entropy from that using the extract function from HKDF.  This produces a pseudorandom key.

We ask HKDF expand to produce enough bits for an AES key.  I'm using AES-256 here because I can.  Different HKDF labels allow us to create multiple unrelated shared secrets.  This is a well-established technique (TLS uses it for example).

I'm using AES here as a raw PRF.  Given an input (which I'm taking as u128), and the generated key, AES provides a pseudorandom permutation of that input.  AES is used to output a u128 value that is effectively random.

A PRP is a PRF with a weird additional property that each input has a distinct output.  That's probably not ideal, but this code destroys that property by taking the AES output modulo the field prime.

<details>
<summary>For more on taking a modulus...</summary>

Note that modulo is a terrible idea if the size of your field is close to the size of the random input.  For example, if you had random 7 bit value and you are taking a modulus into GF(127), the value 0 has twice the probability of being chosen than any other value. Here we're going to be choosing fields that are at most 64 bits wide, so the bias is minuscule.  (Rejection sampling might be possible, but then we'd need to reserve some number of bits from the input space for getting extra values.)  Biases like this can be exploited in some settings (this is why RC4 is no longer considered good), but that sort of attack is not applicable in this setting...to my knowledge.

Adding extra bits to the random value dilutes the bias, effectively by over-sampling.  An 8-bit value into GF(127) biases toward 0 and 1 50% more than other values, but a 16-bit value has just a 0.19% bias on 4 values.  For the example of Fp31, there are 31 values, of which 23 have 10976850545836724323181988441692831744 chances of being picked and the remaining 8 have 10976850545836724323181988441692831745 chances.  That is biased, but at 9e-36% or 2<sup>-123</sup> probability that bias is negligible.  As long as we choose primes that are small (i.e., less than 2<sup>64</sup>) that amount of bias should not provide any advantage to an adversary.

</details>

The way I anticipate this being used is that the functioning of each helper is synchronized.  Inputs are provided in the same order to each helper, allowing us to allocate a number to each.  Each call that is made of these functions can thereby be given a unique number that all helpers can agree on.  That input might be derived by allocating numbers to each record, to each step, and to each invocation of these functions.  That allocation can use some convention we invent.  The numbers we choose can be packed into a u128 (again using an agreed convention) and then used as an index to these functions.

PR #41, opened 2022-07-14
Radix sort

richajaindce:
Implementing Radix Sorting using the approach described in https://eprint.iacr.org/2019/695.pdf
Screenshot 2022-07-14 at 10 11 08 PM

This diff implements finding order of each input in the sorted list. Following will be handled in later diffs
1. Asynchronous calls and also helper communication
2. Optimization e.g. remove quadratic and linear randoms from king
3. Oblivious Shuffle
4. Stubs to calculate number of bytes exchanged between helpers


src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,290 @@
+use ff::PrimeField;
```
benjaminsavage, line 1: Let's use Fp31 instead of PrimeField now that it has landed.

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,290 @@
...
+use rand::thread_rng;
+
+const TOTAL_HELPERS: usize = 3;
+
+#[must_use]
+pub fn mpc_sort<F: PrimeField>(shared_secrets: &[Vec<F>]) -> Vec<Vec<F>> {
+    let (linear_randoms, quadratic_randoms) = get_random_nos_for_mult(shared_secrets.len());
```
benjaminsavage, line 8: For all of the Shamir secret sharing, we can now use @akoshelev 's code: https://github.com/martinthomson/raw-ipa/pull/32

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,290 @@
+use ff::PrimeField;
+use rand::thread_rng;
+
+const TOTAL_HELPERS: usize = 3;
+
+#[must_use]
+pub fn mpc_sort<F: PrimeField>(shared_secrets: &[Vec<F>]) -> Vec<Vec<F>> {
```
benjaminsavage, line 7: Currently this implementation seems really tied to Shamir Secret Sharing. 

We might want to keep it flexible, and use other types of secure multiplication. Can we stay generic for now? What I mean is, just refer to multiplication and addition, but not to any of the implementation details?

benjaminsavage (changes requested):

src/helpers/radix_sort.rs
richajaindce: Yes! I am working on this and push it in the next iteration

richajaindce:
Closing this request, since have created a separate one on top of Ben's implementation of ReplicatedSecretSharing

PR #44, opened 2022-07-15
Introduce metrics to the system

akoshelev:
## Overview

There are currently two metrics implementations that most of the crates use:
* [OpenTelemetry](https://docs.rs/opentelemetry/latest/opentelemetry/#metrics)
* [Metrics](https://docs.rs/metrics/0.19.0/metrics/)

Both provide similar API that allows creating counters, gauges and histograms.
However, there seem to be a warning that OpenTelemetry implementation is not stable.
Also, OpenTelemetry crate is more than just metrics, it also provides tracing API,
which may shadow `tracing` if pulled in scope.

Because of these reasons, this change suggests to use `Metrics` crate. However, it is
likely that we would need to start using OpenTelemetry at some point because it
provides integration with many open-source and third-party metric/trace collectors
and visualization tools. I don't believe we will have to rewrite IPA implementation
to do that, because it is possible to integrate tracing with OpenTelemetry using
[tracing-opentelemetry](https://docs.rs/tracing-opentelemetry/0.17.4/tracing_opentelemetry/)
crate and integrate `Metrics` by creating a `Recorder` that just dumps everything
to `OpenTelemetry` sink

## Description of the change

* Add `Metrics` crate as dependency
* Modify tower tracing layer to emit `requests.received` metric
* Unit tests for the above
* A toy implementation of metrics sinkkk for CLI that dumps everything to stderr

## Testing

```bash
cargo run --features self-signed-certs --bin helper -- -s https -p 3000 -vvv
```

```bash
wget -q -nv "https://127.0.0.1:3000/echo?foo=bar" --no-check-certificate -O -
wget -q -nv "https://127.0.0.1:3000/echo?anotherfoo=anotherbar" --no-check-certificate -O -
```

```bash
DEBUG rustls::server::hs: decided upon suite TLS13_AES_256_GCM_SHA384
DEBUG hyper::proto::h1::io: parsed 5 headers
DEBUG hyper::proto::h1::conn: incoming body is empty
DEBUG request{method=GET uri=/echo?foo=bar version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
DEBUG hyper::proto::h1::io: flushed 270 bytes
DEBUG hyper::proto::h1::conn: read eof
DEBUG rustls::conn: Sending warning alert CloseNotify
DEBUG rustls::server::hs: decided upon suite TLS13_AES_256_GCM_SHA384
DEBUG hyper::proto::h1::io: parsed 5 headers
DEBUG hyper::proto::h1::conn: incoming body is empty
DEBUG request{method=GET uri=/echo?anotherfoo=anotherbar version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
DEBUG hyper::proto::h1::io: flushed 284 bytes
DEBUG hyper::proto::h1::conn: read eof
DEBUG rustls::conn: Sending warning alert CloseNotify

+-------------------+-----------------------------------------------------+-------+---------------------------------+
| metric            | description                                         | value | dimensions                      |
+===================================================================================================================+
| requests.received | Total number of requests received by the web server | 2     | uri                             |
|                   |                                                     |       | /echo?foo=bar = 1               |
|                   |                                                     |       | /echo?anotherfoo=anotherbar = 1 |
|                   |                                                     |       | method                          |
|                   |                                                     |       | GET = 2                         |
|                   |                                                     |       | version                         |
|                   |                                                     |       | HTTP/1.1 = 2                    |
|                   |                                                     |       |                                 |
+-------------------+-----------------------------------------------------+-------+---------------------------------+
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,10 +6,10 @@ edition = "2021"
 
 [features]
 default = ["debug", "cli"]
-cli = ["enable-serde", "structopt", "web-app", "tracing-subscriber"]
+cli = ["comfy-table", "enable-serde", "metrics-tracing-context", "metrics-util", "structopt", "tracing-subscriber", "web-app"]
 debug = ["hex"]
 enable-serde = ["serde", "serde_json", "rust-elgamal/enable-serde"]
-web-app = ["tokio", "tokio-stream", "axum", "axum-server", "hyper", "hyper-tls", "tower-http"]
+web-app = ["tokio", "tokio-stream", "axum", "axum-server", "hyper", "hyper-tls", "tower", "tower-http"]
 self-signed-certs = ["hyper-tls"]
 test-fixture = []
 
@@ -19,6 +19,7 @@ async-trait = "0.1.56"
 axum = { version = "0.5.7", optional = true, features = ["http2"] }
 axum-server = { version = "0.4.0", optional = true, features = ["rustls", "rustls-pemfile", "tls-rustls"] }
 byteorder = "1"
+comfy-table = { version = "6.0.0", optional = true }
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, not 0.10
 digest = "0.9"
 futures = "0.3.21"
@@ -29,7 +30,9 @@ hex = { version = "0.4", optional = true }
 hkdf = "0.11"
 hyper = { version = "0.14.19", optional = true, features = ["client", "h2"] }
 hyper-tls = { version = "0.5.0", optional = true }
-log = "0.4"
+metrics = "0.19.0"
+metrics-tracing-context = { version = "0.11.0", optional = true }
+metrics-util = { version = "0.13.0", optional = true }
 pin-project = "1.0.11"
 rand = "0.8"
 rand_core = "0.6"
@@ -44,6 +47,7 @@ structopt = { version = "0.3", optional = true }
 thiserror = "1.0"
 tokio = { version = "1.19.2", optional = true, features = ["rt", "rt-multi-thread", "macros"] }
 tokio-stream = { version = "0.1.9", optional = true }
+tower = { version = "0.4.13", optional = true }
 tower-http = { version = "0.3.4", optional = true, features = ["trace"] }
 tracing = "0.1.35"
 tracing-subscriber = { version = "0.3.14", optional = true }
--- a/src/bin/helper.rs
+++ b/src/bin/helper.rs
@@ -1,9 +1,13 @@
 use std::error::Error;
 
 use hyper::http::uri::Scheme;
+
 use raw_ipa::cli::Verbosity;
 use raw_ipa::net::{bind_mpc_helper_server, BindTarget};
+
 use std::net::SocketAddr;
+use std::panic;
+
 use structopt::StructOpt;
 use tracing::info;
 
@@ -26,7 +30,7 @@ struct Args {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::from_args();
-    args.logging.setup_logging();
+    let _handle = args.logging.setup_logging();
 
     // decide what protocol we're going to use here
     let addr = SocketAddr::from(([127, 0, 0, 1], args.port.unwrap_or(0)));
@@ -44,8 +48,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     // start server
     let (addr, server_handle) = bind_mpc_helper_server(target).await;
-    info!("listening to {}://{}", args.scheme, addr);
-    server_handle.await?;
+    info!(
+        "listening to {}://{}, press Enter to quit",
+        args.scheme, addr
+    );
+    let _ = std::io::stdin().read_line(&mut String::new())?;
+    server_handle.abort();
 
     Ok(())
 }
--- a/src/cli/metric_collector.rs
+++ b/src/cli/metric_collector.rs
@@ -1,0 +1,44 @@
+use crate::cli::Metrics;
+use metrics_tracing_context::TracingContextLayer;
+use metrics_util::debugging::{DebuggingRecorder, Snapshotter};
+use metrics_util::layers::Layer;
+use std::io::stderr;
+use std::thread;
+
+/// Collects metrics using `DebuggingRecorder` and dumps them to `stderr` when dropped.
+pub struct CollectorHandle {
+    snapshotter: Snapshotter,
+}
+
+///
+/// Initializes this collector by installing `DebuggingRecorder` to keep track of metrics
+/// emitted from different parts of the app.
+///
+/// ## Panics
+/// Panics if metric recorder has already been set
+#[must_use]
+pub fn install_collector() -> CollectorHandle {
+    let recorder = DebuggingRecorder::new();
+    let snapshotter = recorder.snapshotter();
+
+    // use span fields as dimensions for metric
+    let recorder = TracingContextLayer::all().layer(recorder);
+    metrics::set_boxed_recorder(Box::new(recorder))
+        .expect("Metric recorder has been installed already");
+
+    // register metrics
+    crate::telemetry::metrics::register();
+
+    CollectorHandle { snapshotter }
+}
+
+impl Drop for CollectorHandle {
+    fn drop(&mut self) {
+        if !thread::panicking() {
+            let stats = Metrics::from_snapshot(self.snapshotter.snapshot());
+            stats
+                .print(&mut stderr())
+                .expect("Failed to dump metrics to stderr");
+        }
+    }
+}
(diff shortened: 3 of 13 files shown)


src/cli/stats.rs
```diff
@@ -0,0 +1,122 @@
...
+            _ => unreachable!(),
+        };
+        for label in key.key().labels() {
+            let (label_key, label_val) = label.clone().into_parts();
+            let dimension_values = self
+                .dimensions
+                .entry(label_key)
+                .or_insert_with(HashMap::new);
+
+            *dimension_values.entry(label_val).or_insert_with(|| 0) += val;
```
thurstonsand, line 50:
```suggestion
            *dimension_values.entry(label_val).or_insert(0) += val;
```

src/cli/stats.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Dumps the stats to the provided Write interface.
+    ///
+    /// ## Errors
+    /// returns an IO error if it fails to write to the provided writer.
+    pub fn print(&self, w: &mut impl Write) -> Result<(), std::io::Error> {
+        let mut metrics_table = Table::new();
+        metrics_table.set_header(vec!["metric", "description", "value", "dimensions"]);
+
+        for (key_name, counter_stats) in &self.counters {
+            let key_name = key_name;
```
thurstonsand, line 100: ?

src/cli/verbosity.rs
```diff
@@ -15,16 +19,34 @@ pub struct Verbosity {
...
         let fmt_layer = fmt::layer().without_time().with_writer(stderr);
 
         tracing_subscriber::registry()
             .with(self.level_filter())
             .with(fmt_layer)
+            .with(MetricsLayer::new())
             .init();
+
+        let handle = LoggingHandle {
+            metrics_handle: if self.quiet {
```
thurstonsand, line 40: nice shorthand:

```rust
metrics_handle: (!self.quiet).then(|| install_collector())
```

src/net/server/mod.rs
```diff
@@ -149,19 +156,22 @@ ShF2TD9MWOlghJSEC6+W3nModkc=
...
-        body,
-        client::HttpConnector,
-        header::{HeaderName, HeaderValue},
-        http::uri::Scheme,
-        Body, Request, Response, StatusCode,
+        body, client::HttpConnector, http::uri::Scheme, Body, Request, Response, StatusCode,
     };
     use hyper_tls::{native_tls::TlsConnector, HttpsConnector};
     use std::collections::HashMap;
     use std::str::FromStr;
```
thurstonsand, line 167: whitespace?

src/telemetry/mod.rs
```diff
@@ -0,0 +1,48 @@
+pub mod metrics {
```
thurstonsand, line 1: so why have mod inside mod like this? why not just create a top-level `metrics` mod?

src/telemetry/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    use metrics::Unit;
+
+    pub const REQUESTS_RECEIVED: &str = "requests.received";
+
+    /// Registers metrics used in the system with the metrics recorder.
+    ///
+    /// ## Panics
+    /// Panic if there is no recorder installed
+    pub fn register() {
+        assert!(
```
thurstonsand, line 12: is this equivalent to `assert!(metrics::try_recorder().is_some())`?

src/cli/metric_collector.rs
```diff
@@ -0,0 +1,44 @@
+use crate::cli::Metrics;
```
thurstonsand, line 1: curious what the reasoning is to include this here in the `cli` mod vs the `metrics` mod?

Cargo.toml
```diff
@@ -6,26 +6,28 @@ edition = "2021"
...
 digest = "0.9"
-
 hex = { version = "0.4", optional = true }
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, so pin this
 hkdf = "0.11"
 hyper = { version = "0.14.19", optional = true, features = ["client", "h2"] }
 hyper-tls = { version = "0.5.0", optional = true }
-log = "0.4"
+metrics = "0.19.0"
+metrics-tracing-context = { version = "0.11.0", optional = true }
```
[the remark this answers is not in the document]
akoshelev: `optional` means it is not available and won't be [compiled by default](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies). In this case I don't want traits from this crate to appear/be available everywhere - the only place where we need to use them is the MPC web server or CLI

src/bin/helper.rs
```diff
@@ -1,9 +1,13 @@
 use std::error::Error;
 
 use hyper::http::uri::Scheme;
+
```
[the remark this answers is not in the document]
akoshelev: seems like either rustfmt or my IDE insert these to group imports. Once set, neither of them removes blank lines. I am not sure whether fixing them is important given that rustfmt is happy

src/bin/helper.rs
```diff
@@ -26,7 +30,7 @@ struct Args {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::from_args();
-    args.logging.setup_logging();
+    let _handle = args.logging.setup_logging();
```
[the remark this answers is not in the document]
akoshelev: `let _ = ...` drops value immediately after the statement, while `let _handle  = ...`  injects drop at the end of the scope where `_handle` is visible. I need the latter here

src/bin/helper.rs
```diff
@@ -44,8 +48,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     // start server
     let (addr, server_handle) = bind_mpc_helper_server(target).await;
-    info!("listening to {}://{}", args.scheme, addr);
-    server_handle.await?;
+    info!(
+        "listening to {}://{}, press Enter to quit",
+        args.scheme, addr
+    );
+    let _ = std::io::stdin().read_line(&mut String::new())?;
```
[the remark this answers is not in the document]
akoshelev: I just want to read enter here

src/bin/ua.rs
```diff
@@ -132,7 +132,7 @@ impl Action {
 
 fn main() {
     let args = Args::from_args();
-    args.common.logging.setup_logging();
+    let _handle = args.common.logging.setup_logging();
```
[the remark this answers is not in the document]
akoshelev: see above

src/cli/metric_collector.rs
akoshelev: this collector is dumb and only works when running things from the CLI because it dumps everything to `stderr`

src/cli/stats.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl CounterDetails {
+    pub fn add(&mut self, key: &CompositeKey, val: &DebugValue) {
+        let val = match val {
+            DebugValue::Counter(v) => v,
+            _ => unreachable!(),
+        };
+        for label in key.key().labels() {
+            let (label_key, label_val) = label.clone().into_parts();
```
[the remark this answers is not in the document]
akoshelev: `into_parts` takes `self` (i.e. consumes it) while `label` is a reference to `Label` that does not implement `Copy`.

src/cli/stats.rs
akoshelev: 👍

src/cli/stats.rs
akoshelev: ugh, there used to be `Clone` here, good catch, removing

src/cli/verbosity.rs
akoshelev: 👍

src/telemetry/mod.rs
akoshelev: I could - my thinking was that this module is very small now and a file per module may make things too complicated. Don't mind either way though

src/telemetry/mod.rs
akoshelev: yep I think it is the same

src/bin/helper.rs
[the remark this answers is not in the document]
thurstonsand: that's pretty subtle...but got it

src/bin/helper.rs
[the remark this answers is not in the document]
thurstonsand: I believe they are not grouped -- ie if you put whitespace between every single `use` statement, rustfmt would honor that. not sure if that's the kind of behavior we want

akoshelev:
@martinthomson would you be able to take a look at this? I am working on instrumenting the messaging layer and having metrics would be quite useful to understand the performance of individual operations.

martinthomson (approved):
Again, free pass because I'm out for a few more weeks.

PR #46, opened 2022-07-19
Field: move `pow` and `invert` to the trait def.

akoshelev:
These implementations are generic enough to be defined for any field, but they require some additional bounds on `Self::Integer` to be able to subtract, right shift and determine the size in bits of the underlying type

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -1,7 +1,28 @@
+use std::ops::{BitAnd, Shr};
 use std::{
     fmt::Debug,
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
+
+// Trait for primitive integer types used to represent the underlying type for field values
+pub trait Int:
+    Sized
+    + Copy
+    + Debug
+    + Ord
+    + Sub<Output = Self>
+    + Into<u128>
+    + From<u8>
+    + Shr<u32, Output = Self>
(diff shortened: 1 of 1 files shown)


src/field.rs
```diff
@@ -1,8 +1,30 @@
...
+// Trait for primitive integer types used to represent the underlying type for field values
+pub trait Int:
+    Sized
+    + Copy
+    + Debug
+    + Ord
+    + Sub<Output = Self>
+    + Into<u128>
+    + Shr<u32, Output = Self>
+    + BitAnd<u8, Output = Self>
```
martinthomson, line 16: Doesn't this need to be:

```suggestion
    + BitAnd<Self, Output = Self>
```

martinthomson: I guess then you would need to define `const ONE: Self` as well.

You could bound with `From<u8>` and drop both `ZERO` and `ONE`.

I'm mostly just concerned that this won't work when the Field uses a different integer basis.

martinthomson (approved):
This looks like it might work pretty well.  I like the general shape of it.

src/field.rs
akoshelev: yea I was thinking about it too - don't have a good explanation why I decided against it. It should always be safe to upcast from `u8`.

thurstonsand:
LGTM

PR #47, opened 2022-07-19
Adding a ReplicatedSecretSharing implementation

benjaminsavage:
In this rev, only supports addition, subtraction and multiplication by a constant

### DIFF
--- a/src/replicated_secret_sharing.rs
+++ b/src/replicated_secret_sharing.rs
@@ -1,0 +1,191 @@
+use std::{
+    fmt::Debug,
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use crate::field::Field;
+use crate::prss::Participant;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct ReplicatedSecretSharing<T>(T, T);
+
+impl<T: Field> ReplicatedSecretSharing<T> {
+    #[must_use]
+    pub fn construct(a: T, b: T) -> Self {
+        Self(a, b)
+    }
+}
+
+impl<T: Field> Add for ReplicatedSecretSharing<T> {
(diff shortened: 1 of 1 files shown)


src/replicated_secret_sharing.rs
```diff
@@ -0,0 +1,191 @@
+use std::{
+    fmt::Debug,
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use crate::field::Field;
+use crate::prss::Participant;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct ReplicatedSecretSharing<T>(T, T);
```
akoshelev, line 10: does this represent a share in replicated secret sharing scheme?

src/replicated_secret_sharing.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::replicated_secret_sharing::ReplicatedSecretSharing;
+
+    use crate::field::Fp31;
+
+    fn secret_share(
```
akoshelev, line 49: the way I would suggest to structure this is to have a function `secret_share(secret: u8, mut rng: &Rng) -> [ReplicatedSecretSharing<Fp31>; 3]` that does the sharing `(secret -> (u8, u8, u8)` and then produces secret shares - it could eliminate the human error constructing the shares and make tests a bit easier to understand

akoshelev (approved):
lgtm, minor feedback

src/replicated_secret_sharing.rs
benjaminsavage: This represents one helper's data within a replicated secret sharing scheme

PR #48, opened 2022-07-20
Radix sort on replicated shares

richajaindce:
This is built on top of PR #45 but I am unable to stack it on top of it. 

Implementing Radix Sorting using the approach described in https://eprint.iacr.org/2019/695.pdf
image

This diff implements finding order of each input in the sorted list. Following will be handled in later diffs

Asynchronous calls and also helper communication
Oblivious Shuffle
Stubs to calculate number of bytes exchanged between helpers


thurstonsand:
can you resolve the conflicts with the main branch?

richajaindce:
Since this is dependent on code related to replicated shares, I will wait for that to finish before making this ready for shipping

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+    (0..mult_inputs[0].len()).for_each(|i| {
+        let a1 = mult_inputs[0][i].0;
+        let a2 = mult_inputs[1][i].0;
+        let a3 = mult_inputs[2][i].0;
+        let b1 = mult_inputs[0][i].1;
+        let b2 = mult_inputs[1][i].1;
+        let b3 = mult_inputs[2][i].1;
+        mult_outputs.push(multiply_secret_shares(
+            &h1, &h2, &h3, a1, a2, a3, b1, b2, b3,
+        ));
```
benjaminsavage, line 59: I feel like it would have been more clear to just do step 1 of the multiplication within the helpers_gen_mult_inputs function. Having that function just return a data structure with all the inputs, just to pass them into this function which has to re-assemble all the inputs out of that data structure doesn't add anything but complexity.

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+    });
+    ((h1, h2, h3), mult_outputs)
```
benjaminsavage, line 61: It looks like you never use the first return value... why not remove it?

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+fn helpers_gen_permutations(mult_output: &[ReplicatedShares]) -> Vec<ReplicatedShares> {
+    // Helpers compute sum
+    let mut gen_permutation: Vec<ReplicatedShares> = Vec::new();
```
benjaminsavage, line 66: I'd use a tuple. not a Vec. Also, why is the name of a vector called "gen_"?

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+    let mut shares1 = Vec::new();
+    let mut shares2 = Vec::new();
+    let mut shares3 = Vec::new();
+
+    for iter in mult_output {
```
benjaminsavage, line 71: "iter" isn't a great variable name =).

But I think the bigger issue is the way this is written in a way where all the data is together in one place, just to get separated into pieces like this. I'd rather see code that mimics more the way it'll really work - where each helper only ever has access to a subset of the data.

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+);
+
+type RandomShareGenerationHelpers = (
+    RandomShareGenerationHelper,
+    RandomShareGenerationHelper,
+    RandomShareGenerationHelper,
+);
+
+#[must_use]
+pub fn mpc_sort(shared_secrets: &[ReplicatedShares]) -> Vec<ReplicatedShares> {
```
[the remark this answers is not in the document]
benjaminsavage: This is also a misnomer. It doesn't actually sort the input. This just returns *secret shares* of a permutation corresponding to one bit.

src/modulus_convert.rs
```diff
@@ -394,90 +396,96 @@ impl RandomShareGenerationHelper {
...
+    let h2 = RandomShareGenerationHelper::init(p2, HelperIdentity::H2);
+
+    // Helper 3
+    let h3 = RandomShareGenerationHelper::init(p3, HelperIdentity::H3);
+
+    (h1, h2, h3)
+}
+
+#[must_use]
+pub fn secret_share(
```
benjaminsavage, line 459: This also belongs only in test modules.

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+                ReplicatedFp31SecretSharing::construct(
+                    Fp31::from(expected_sort_output[i]),
+                    Fp31::from(expected_sort_output[i])
+                )
+            );
+        });
+    }
+
+    #[test]
+    fn gen_mult_inputs() {
```
benjaminsavage, line 193: I don't think we need to separately test this one. The test above is probably enough. I'm also not sure this particular division into sub-functions is the one we will stick with - so adding tests for each one will just make it harder to refactor.

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+        let mult_output = output.1;
+
+        assert_eq!(3, mult_output.len());
+        assert_eq!(shares.len() * 2, mult_output[0].len());
+        assert_eq!(shares.len() * 2, mult_output[1].len());
+        assert_eq!(shares.len() * 2, mult_output[2].len());
+    }
+
+    #[test]
+    fn replicated_share_sort_test_individual_components_size_check() {
```
benjaminsavage, line 218: Same as above, rather than having 3, highly repetitive tests, I'd recommend just adding these assets to the one main test.

benjaminsavage (changes requested):

src/helpers/radix_sort.rs
richajaindce: This one is a vector of a tuple

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    // Fold the result
+    let mut permutation_share = Vec::new();
+    (0..len).for_each(|j| {
+        permutation_share.push(mult_output[j] + mult_output[j + len]);
+    });
+    permutation_share
+}
+
+fn get_helper_specific_shares(
```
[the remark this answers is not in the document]
richajaindce: Makes sense.. moving all "visible to all code" inside the tests

src/helpers/radix_sort.rs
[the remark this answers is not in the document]
richajaindce: Agreed, changed it to radix_mpc_radix_sort

src/helpers/radix_sort.rs
```diff
@@ -0,0 +1,247 @@
...
+    shares: &[ReplicatedFp31SecretSharing],
+    helper: &RandomShareGenerationHelper,
+) -> Vec<(ReplicatedFp31SecretSharing, ReplicatedFp31SecretSharing)> {
+    
+    // Step1: Calculate x and 1 - x
+    let mut x_and_one_minus_x: Vec<ReplicatedFp31SecretSharing> = shares
+        .iter()
+        .map(|x| helper.get_share_of_one() - *x)
+        .collect();
+    x_and_one_minus_x.append(&mut shares.to_vec());
```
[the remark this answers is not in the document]
richajaindce: Append function moves all values from other array, thats why it needs a mutable reference
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append

PR #51, opened 2022-07-23
add mp-spdz prototype, data generator, and readme

eriktaubeneck:
Currently only merging this in as a branch so we can work on it. cc @martinthomson @benjaminsavage @bmcase

### DIFF
--- a/research-prototype/Programs/Source/ipae2e.mpc
+++ b/research-prototype/Programs/Source/ipae2e.mpc
@@ -56,7 +56,7 @@ for row in rows:
         exec(cmd, {}, local_params)
 
 # Radix sort
-reports.sort((0,), n_bits=32)
+reports.sort((Columns.MATCHKEY,), n_bits=32)
 
 if verbose:
     for row in rows:
@@ -189,7 +189,8 @@ for breakdown_key in breakdown_keys:
     for row in rows:
         cmd = (
             f"breakdown_key_sums[{breakdown_key}] = breakdown_key_sums[{breakdown_key}] + "
-            f"(reports[{row}][3] == {breakdown_key}).if_else(final_credits[{row}], 0)"
+            f"(reports[{row}][Columns.BREAKDOWN_KEY] =="
+            f"{breakdown_key}).if_else(final_credits[{row}], 0)"
         )
         exec(cmd, {}, local_params)
 
--- a/research-prototype/Programs/Source/ipae2e.mpc
+++ b/research-prototype/Programs/Source/ipae2e.mpc
@@ -22,45 +22,45 @@ print_ln('********************  IPA E2E %s **********************', numrows)
 
 _list = Matrix(numrows, 4, sint)
 
+local_params = {
+    '_list': _list,
+    'sint': sint,
+    'print_ln': print_ln,
+}
 
-@for_range(numrows)
-def _(i):
-    _list[i][0] = sint.get_input_from(0)  # matchkey
-    _list[i][1] = sint.get_input_from(0)  # istrigger
-    _list[i][2] = sint.get_input_from(0)  # value
-    _list[i][3] = sint.get_input_from(0)  # breakdown
-
+for i in range(numrows):
+    for j in range(3):
+        cmd = f"_list[{i}][{j}] = sint.get_input_from(0)"
+        exec(cmd, {}, local_params)
 
 # Radix sort
 _list.sort((0,), n_bits=32)
 
 if verbose:
     for i in range(numrows):
-        print_ln(
-            '%s, %s, %s, %s',
-            _list[i][0].reveal(),
-            _list[i][1].reveal(),
-            _list[i][2].reveal(),
-            _list[i][3].reveal(),
-        )
+        cmd = (f"print_ln('%s, %s, %s, %s', "
+               f"_list[{i}][0].reveal(), _list[{i}][1].reveal(),"
+               f"_list[{i}][2].reveal(), _list[{i}][3].reveal(),)")
+        exec(cmd, {}, local_params)
 
 helperbit = Array(numrows, sint)
+local_params['helperbit'] = helperbit
 
-
-@for_range(numrows-1)
-def _(i):
-    helperbit[i] = (_list[i][0] == _list[i+1][0]).if_else(1, 0)
+for i in range(numrows-1):
+    cmd = f"helperbit[{i}] = (_list[{i}][0] == _list[{i}+1][0]).if_else(1, 0)"
+    exec(cmd, {}, local_params)
 
 
 helperbit_and_istrigger = Array(numrows, sint)
+local_params['helperbit_and_istrigger'] = helperbit_and_istrigger
 helperbit_xor_istrigger = Array(numrows, sint)
+local_params['helperbit_xor_istrigger'] = helperbit_xor_istrigger
 
-
-@for_range(numrows-1)
-def _(i):
-    helperbit_and_istrigger[i] = helperbit[i] * _list[i][1]
-    helperbit_xor_istrigger[i] = (helperbit[i] + _list[i][1]) % 2
-
+for i in range(numrows-1):
+    cmd = f"helperbit_and_istrigger[{i}] = helperbit[{i}] * _list[{i}][1]"
+    exec(cmd, {}, local_params)
+    cmd = f"helperbit_xor_istrigger[{i}] = (helperbit[{i}] + _list[{i}][1]) % 2"
+    exec(cmd, {}, local_params)
 
 if verbose:
     print_ln('%s', helperbit.reveal())
@@ -71,29 +71,33 @@ if verbose:
 stopbits = [Array(numrows, sint) for _ in range(numrows_power+1)]
 
 for stopbit in stopbits:
-    @for_range(numrows)
-    def __(i):
-        stopbit[i] = 1
+    local_params['stopbit'] = stopbit
+    for i in range(numrows):
+        cmd = f"stopbit[{i}] = 1"
+        exec(cmd, {}, local_params)
 
 credits = [Array(numrows, sint) for _ in range(numrows_power+1)]
-
+local_params['credits'] = credits
 
 for i, (stopbit, credit) in enumerate(zip(stopbits, credits)):
     stepsize = 2 ** i
+    local_params['stopbit'] = stopbit
+    local_params['credit'] = credit
+    local_params['stepsize'] = stepsize
 
-    @for_range(numrows - stepsize)
-    def set_flags(current):
+    for current in range(numrows - stepsize):
         successor = current + stepsize
         # can remove stopbit1[current] since will always be 1
         # another optimization could be to multiply one time all the
         # helperbit * istrigger bits since these are fixed.
-        flag = helperbit_and_istrigger[successor]
 
+        cmd = f"flag = helperbit_and_istrigger[{successor}];"
         # credit is initialized to value, _list[i][2] = value[i]
-        credit[current] = _list[current][2] + flag * _list[successor][2]
+        cmd += f"credit[{current}] = _list[{current}][2] + flag * _list[{successor}][2];"
 
         # * stopbit[successor]  stopbit[successor] here would always be 1 so can drop
-        stopbit[current] = flag
+        cmd += f"stopbit[{current}] = flag"
+        exec(cmd, {}, local_params)
 
     if verbose:
         print_ln('iteration: %s', i)
@@ -109,12 +113,11 @@ if verbose:
 # I can't sum all of credit, I need to sum only those that correspond to last touches
 # helperbit_xor_istrigger * credits[-1] will give this.
 creditfinal = Array(numrows, sint)
+local_params['creditfinal'] = creditfinal
 
-
-@for_range(numrows)
-def _(i):
-    creditfinal[i] = helperbit_xor_istrigger[i] * credits[-1][i]
-
+for i in range(numrows):
+    cmd = f"creditfinal[{i}] = helperbit_xor_istrigger[{i}] * credits[-1][{i}]"
+    exec(cmd, {}, local_params)
 
 if verbose:
     print_ln("creditfinal:")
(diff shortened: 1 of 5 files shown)


eriktaubeneck:
I should highlight that this is almost entirely @bmcase's work. I just wrote the readme and refactored a little bit.

bmcase:
Thanks @eriktaubeneck. I also uploaded my original code (https://github.com/martinthomson/raw-ipa/pull/53) as the refactoring changes you made increased the circuit depth by a factor of 10,000x.   I think circuit depth will be quite important especially as we need to plan for deployment across different geographic regions, but hopefully we can test out the affects of depth on latency soon.

eriktaubeneck:
@bmcase I'm now getting the same results, and the same efficiency in depth (i.e. for 2048 rows, I get 135 rounds.)

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+
+for bk in range(len(bksums)):
+    cmd = f"bksums[{bk}] = sint(0)"
+    exec(cmd, {}, local_params)
+    for i in range(numrows):
+        cmd = (
+            f"bksums[{bk}] = bksums[{bk}] + "
+            f"(_list[{i}][3] == {bk}).if_else(creditfinal[{i}], 0)"
+        )
+        exec(cmd, {}, local_params)
```
eriktaubeneck, line 142: @bmcase I'm 99% sure this does the same thing as before (but doesn't do the "tree" approach. Given that we have to do the `if_else` on every row anyways, I'm not sure it really added any efficiency, and this is definitely easier to follow. Let me know if that makes sense.

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+
+# Radix sort
+_list.sort((0,), n_bits=32)
+
+if verbose:
+    for i in range(numrows):
+        cmd = (f"print_ln('%s, %s, %s, %s', "
+               f"_list[{i}][0].reveal(), _list[{i}][1].reveal(),"
+               f"_list[{i}][2].reveal(), _list[{i}][3].reveal(),)")
+        exec(cmd, {}, local_params)
```
benjaminsavage, line 50: I assume you're using this cmd / exec paradigm as a part of some kind of optimization - but wow this is not readable. Is there any way to just write normal looking code?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
+import os
```
benjaminsavage, line 1: Is there some way to get syntax highlighting?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+
+if verbose and numrows_power > 5:
+    raise Exception(
+        f"You cannot compiling with VERBOSE on for {numrows} rows. "
+        f"This will result in quite a bit of output in your terminal."
+    )
+
+print_ln('********************  IPA E2E %s **********************', numrows)
+
+_list = Matrix(numrows, 4, sint)
```
benjaminsavage, line 29: Why is it called "_list"? Is the underscore important?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+    exec(cmd, {}, local_params)
+
+
+helperbit_and_istrigger = Array(numrows, sint)
+local_params['helperbit_and_istrigger'] = helperbit_and_istrigger
+helperbit_xor_istrigger = Array(numrows, sint)
+local_params['helperbit_xor_istrigger'] = helperbit_xor_istrigger
+
+for i in range(numrows-1):
+    cmd = f"helperbit_and_istrigger[{i}] = helperbit[{i}] * _list[{i}][1]"
```
benjaminsavage, line 66: _list[{i}][1] is like... not very readable. Could we make the rows of the list some kind of struct that we can index via a key? like:

_list[{i}].helper_bit

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+    exec(cmd, {}, local_params)
+    cmd = f"helperbit_xor_istrigger[{i}] = (helperbit[{i}] + _list[{i}][1]) % 2"
```
benjaminsavage, line 68: Why not use ^ instead?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+stopbits = [Array(numrows, sint) for _ in range(numrows_power)]
+local_params['stopbits'] = stopbits
+
+for stopbit in stopbits:
+    local_params['stopbit'] = stopbit
+    for i in range(numrows):
+        cmd = f"stopbit[{i}] = 1"
+        exec(cmd, {}, local_params)
+
+credits = [Array(numrows, sint) for _ in range(numrows_power)]
```
benjaminsavage, line 86: What is going on here?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+local_params['credits'] = credits
+
+for i, (stopbit, credit) in enumerate(zip(stopbits, credits)):
```
benjaminsavage, line 89: What does enumerate(zip(a, b)) do?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+    local_params['stepsize'] = stepsize
+
+    for current in range(numrows - stepsize):
+        successor = current + stepsize
+        # another optimization could be to multiply one time all the
+        # helperbit * istrigger bits since these are fixed.
+        if i == 0:
+            cmd = f"flag = helperbit_and_istrigger[{successor}];"
+            # credit is initialized to value, _list[i][2] = value[i]
+            cmd += f"credit[{current}] = _list[{current}][2] + flag * _list[{successor}][2];"
```
benjaminsavage, line 102: I'm not sure what _list[{i}][2] is. Once again, can we somehow indicate what index 2 means? If we cannot use a key-value mapping i.e. _list[{i}].value, then let's at least define an enum so we can do:

_list[{i}][VALUE]

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+            # * stopbit[successor]  stopbit[successor] here would always be 1 so can drop
+            cmd += f"stopbit[{current}] = flag"
+        else:
+            cmd = f"flag = stopbits[{i-1}][{current}] * helperbit_and_istrigger[{successor}];"
+            cmd += f"credit[{current}] = "
+            cmd += f"credits[{i-1}][{current}] + flag * credits[{i-1}][{successor}];"
+            cmd += f"stopbit[{current}] = flag * stopbits[{i-1}][{successor}]"
```
benjaminsavage, line 109: += being used to concatenate a bunch of strings, which sort of loosely correspond to lines of code... is just not awesome.

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,154 @@
...
+        print_ln('credit %s', credit.reveal())
+        print_ln('stopbit %s', stopbit.reveal())
+
+# I can't sum all of credit, I need to sum only those that correspond to last touches
+# helperbit_xor_istrigger * credits[-1] will give this.
+creditfinal = Array(numrows, sint)
+local_params['creditfinal'] = creditfinal
+
+for i in range(numrows):
+    cmd = f"creditfinal[{i}] = helperbit_xor_istrigger[{i}] * credits[-1][{i}]"
```
benjaminsavage, line 124: What does credits[-1] even do?

research-prototype/ipainput.py
```diff
@@ -0,0 +1,32 @@
...
+mkmod = 2**31
+datamod = 2**8
+breakdown_keys = 4
+
+player_data = Path("Player-Data")
+player_data.mkdir(parents=True, exist_ok=True)
+
+with open(player_data / "Input-P0-0", "w") as f:
+    for i in range(numrows):
+        mk = random.randint(0, numrows//approx_rows_per_mk)
```
benjaminsavage, line 23: What does the "//" mean?

research-prototype/ipainput.py
```diff
@@ -0,0 +1,32 @@
...
+        is_trigger = random.randint(0, 1)
+
+        if(is_trigger):
+            value, bk = random.randint(0, datamod-1), 0
+        else:
+            value, bk = 0, random.randint(0, breakdown_keys-1)
+        f.write(f"{mk}\n{is_trigger}\n{value}\n{bk}\n")  # data = value
```
benjaminsavage, line 30: Wait, you're using the same separator for between rows and between elements in a row?

benjaminsavage (changes requested):
The legibility is pretty poor. I'd love to make the code at least a BIT more accessible.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: I think I found a way!

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: it was originally `list`, which is a built-in. but it should actually be something meaningful like `reports`.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: I wish. It's not actually a list, it's an MP-SPDZ `Matrix`, which is just a matrix and can only be accessed by indices, unfortunately.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: Not sure. @bmcase? I'm guessing because we have additive secret shares in this setup?

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: Great idea!

research-prototype/ipainput.py
eriktaubeneck: Integer division.

research-prototype/ipainput.py
eriktaubeneck: Yes, the way that MP-SPDZ reads in is just one data point from one line at a time.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: I renamed this to `credit_arrays` so that it's hopefully more clear. The -1 index gives you the last element in the list (in this case, the last `Array` in the list of arrays.)

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: yea, it's what we're stuck with. I updated a bunch of the naming the hopefully make it clearer.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: creating a list of Arrays, one for each "iteration" of the protocol.

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: `enumerate(['a', 'b', 'c']) -> [(0, 'a'), (1, 'b'), (2, 'c')]`. See the [python docs.](https://docs.python.org/3/library/functions.html#enumerate).

`zip([10, 20, 30], [100, 200, 300]) -> [(10, 100), (20, 200), (30, 300)]`. See the [python docs.](https://docs.python.org/3/library/functions.html#zip).

research-prototype/Programs/Source/ipae2e.mpc
eriktaubeneck: fixed.

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,196 @@
...
+
+for breakdown_key in breakdown_keys:
+    cmd = f"breakdown_key_sums[{breakdown_key}] = sint(0)"
+    exec(cmd, {}, local_params)
+    for row in rows:
+        cmd = (
+            f"breakdown_key_sums[{breakdown_key}] = breakdown_key_sums[{breakdown_key}] + "
+            f"(reports[{row}][3] == {breakdown_key}).if_else(final_credits[{row}], 0)"
+        )
+        exec(cmd, {}, local_params)
```
eriktaubeneck, line 194: @bmcase I'm 99% sure this does the same thing as before (but doesn't do the "tree" approach. Given that we have to do the if_else on every row anyways, I'm not sure it really added any efficiency, and this is definitely easier to follow. Let me know if that makes sense.

eriktaubeneck:
@benjaminsavage thanks for the review, very helpful ideas! Code is much more readable now (though still full of some crazy hacks.)

research-prototype/Programs/Source/ipae2e.mpc
bmcase: The type sint does not supports ^.  These are arithmetic circuits but the values are assumed to be 0, 1 so we can get bit AND with just *  and XOR with + mod 2

research-prototype/Programs/Source/ipae2e.mpc
bmcase: breakdown_key_sums[{breakdown_key}] = breakdown_key_sums[{breakdown_key}] + (reports[{row}][3] == {breakdown_key}).if_else(final_credits[{row}], 0)

- @eriktaubeneck  this looks like it should be the same; as long as this doesn't increase the rounds for you it should be fine.  It might be that the circuit you get at though won't run multi-threaded, but maybe we can test that out later.   
-can we use beakdown_key instead of "3" here?

eriktaubeneck: good catch!

PR #53, opened 2022-07-25
IPA in MP-SPDZ

bmcase:
Uploading my MP-SPDZ implementation of Interoperable Private Attribution.  

@eriktaubeneck uploaded a refactoring of this code (https://github.com/martinthomson/raw-ipa/pull/51/files) but the simplifications cause the depth of the circuit to blowup by a factor of 10,000x.  

My code is more verbose (e.g. extra arrays so as to never read and write to the same memory in a loop, loop unrolling to avoid nested for loops), but these are all intentional optimizations to bring the circuit depth down to near optimal (scaling with log(numrows)).

If there is a way to improve the compiler to keep the simplifications to code but also get optimal depth, that would of course be great, but until then I don't think we should sacrifice so much on circuit depth.

cc @martinthomson, @benjaminsavage, @akoshelev


eriktaubeneck:
@bmcase let's rebase this on top of eriktaubeneck:mp-spdz-prototype, so that the diff is more clear.

PR #54, opened 2022-07-26
add mp-spdz prototype, data generator, and readme

eriktaubeneck:
@martinthomson any concerns merging this into main?

You can see #51 for the conversation on the original PR.


research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -0,0 +1,197 @@
...
+    'sint': sint,
+    'print_ln': print_ln,
+    'Columns': Columns,
+}
+
+
+# Load data
+for row in rows:
+    for column in Columns:
+        cmd = f"reports[{row}][{column}] = sint.get_input_from(0)"
```
eriktaubeneck, line 55: add comment that one party would never have all the input

eriktaubeneck:
merged with #59 to get formatting checks for .py and .mpc files

eriktaubeneck:
with a huge shout out to @algebravic, I refactored this to no longer use the ugly `exec` hack and instead use vector commands!

eriktaubeneck:
the only thing I'm unsure about here is using `argparse`. it ends up causing the compiled program to be named with the 1 `numrows_power` and `ipa_verbose` variable when used, e.g. compiling withL
```
../../MP-SPDZ/compile.py -C -R 32 ipae2e 5 ipa_verbose
```
requires you to run:
```
../../MP-SPDZ/Scripts/ring.sh -R 32 ipae2e-5-ipa_verbose
```

I think the ENV variables was probably a better hack... curious if anyone else has thoughts.

eriktaubeneck:
cc @benjaminsavage check it out, much better now!!

benjaminsavage (approved):
Awesome. This is a MAJOR improvement in readability. Thank you for this!

PR #55, opened 2022-07-26
Secure multiplication prototype using async/await model

akoshelev:
## Overview

This PR builds on top of https://github.com/martinthomson/raw-ipa/pull/49 and adds secure multiplication protocol using replicated secret sharing ([paper](https://eprint.iacr.org/2018/387.pdf)). 

The whole protocol is implemented inside `SecureMul` struct.

I wanted to make sure once we start using streams we would be able to adopt this approach, so this change also contains a stream implementation for secure multiplication where shares are fed to each helper using stream API. This looks mostly OK to me except that we need to pin stream to the heap because of `async` closure. 

Helper communication API deliberately kept small to make sure I don't accidentally add some features that would be hard to implement for real network communication. This change also brings `TestHelper` implementation that is used in unit tests.


src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
+pub enum HelperAddr {
+    Left,
+    Right,
+}
+
+/// Trait for MPC helpers to communicate with each other. Helpers can send messages and
+/// receive messages from a specific helper.
+#[async_trait]
+pub trait Ring {
```
benjaminsavage, line 32: I'll have to stop using the term "Ring" to refer to the concept in Mathematics: https://en.wikipedia.org/wiki/Ring_(mathematics)

That's fine - I think we don't need to draw any distinction with a field.

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+use serde::Serialize;
+use std::fmt::Debug;
+
+/// Trait for messages sent between helpers
+pub trait Message: Debug + Send + Serialize + DeserializeOwned + 'static {}
+
+impl<T> Message for T where T: Debug + Send + Serialize + DeserializeOwned + 'static {}
+
+/// Destination. Currently we only support Left and Right, but we could support the exact address
+/// too
```
benjaminsavage, line 22: Makes sense. I can imagine other cases where we could have a KING or some such thing (e.g. when all the helpers need to send data to a single place to be revealed).

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/// Trait for MPC helpers to communicate with each other. Helpers can send messages and
+/// receive messages from a specific helper.
+#[async_trait]
+pub trait Ring {
+    /// Send message to the destination. Implementations are free to choose whether it is required
+    /// to wait until `dest` acknowledges message or simply put it to a outgoing queue
+    async fn send<T: Message>(&self, dest: HelperAddr, msg: T) -> Result<(), Error>;
+    async fn receive<T: Message>(&self, source: HelperAddr) -> Result<T, Error>;
```
benjaminsavage, line 36: Nice and simple!

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+        payload: Box<[u8]>,
+    }
+
+    /// A mock implementation of `Ring` trait to be used in unit tests where all helpers are running
+    /// inside the same process. Provides simple and inefficient implementation by buffering all messages
+    /// on `send` and polling the buffer on `receive`. Message is determined to be the same if it has
+    /// the same `TypeId` and helper address matches the destination. For example, message `Foo` sent
+    /// to Helper 2 will be received and removed from the local buffer only when Helper 2 attempts to
+    /// receive it.
+    #[derive(Debug)]
```
benjaminsavage, line 69: Thank you for this comment. Super helpful and very clear.

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+    impl TestHelper {
+        /// Constructs a new instance of test helper using the specified `buf_capacity` buffer
+        /// capacity for the internally used channel.
+        ///
+        /// ## Panics
+        /// Panics if Mutex used internally for synchronization is poisoned or if there are more
+        /// than one message with the same type id and destination address arriving via `send` call.
+        #[must_use]
+        pub fn new(buf_capacity: usize) -> Self {
+            let (tx, mut rx) = channel::<MessageEnvelope>(buf_capacity);
```
benjaminsavage, line 93: What do "tx" and "rx" mean? These variable names aren't super obvious in their meaning (to me)

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+                        let buf = &mut *buf.lock().unwrap();
+                        match buf.entry((item.source, item.type_id)) {
+                            Entry::Occupied(_entry) => {
+                                panic!("Duplicated message {item:?}")
+                            }
+                            Entry::Vacant(entry) => entry.insert(item.payload),
+                        };
+                    }
+                }
+            });
```
benjaminsavage, line 109: This is all greek to me.

src/helpers/ring.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    #[async_trait]
+    impl Ring for TestHelper {
+        async fn send<T: Message>(&self, dest: HelperAddr, msg: T) -> Result<(), Error> {
+            assert!(self.left.is_some());
+            assert!(self.right.is_some());
+
+            let (target, source) = match dest {
+                HelperAddr::Left => (self.left.as_ref().unwrap(), HelperAddr::Right),
+                HelperAddr::Right => (self.right.as_ref().unwrap(), HelperAddr::Left),
```
benjaminsavage, line 136: WTF? Left => Right and Right => Left?

src/lib.rs
```diff
@@ -7,7 +7,9 @@ pub mod field;
 pub mod helpers;
 pub mod net;
 pub mod prss;
+mod replicated_secret_sharing;
```
benjaminsavage, line 10: I can't help noticing this is the only one without a "pub" - why is that?

src/replicated_secret_sharing.rs
```diff
@@ -1,40 +1,62 @@
...
+#[derive(Clone, Copy, PartialEq)]
 pub struct ReplicatedSecretSharing<T>(T, T);
 
+impl<T: Debug> Debug for ReplicatedSecretSharing<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "(")?;
+        self.0.fmt(f)?;
+        write!(f, ",")?;
+        self.1.fmt(f)?;
+        write!(f, ")")
```
benjaminsavage, line 18: Isn't there a way to put this on one line?

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+use crate::field::Field;
+use crate::helpers::ring::{HelperAddr, Ring};
+use crate::prss::Participant;
+use crate::replicated_secret_sharing::ReplicatedSecretSharing;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use thiserror::Error;
+
+/// Secure multiplication protocol using replicated secret sharing over some field `F`.
+/// * [paper](https://eprint.iacr.org/2018/387.pdf)
```
benjaminsavage, line 11: /// IKHC multiplication protocol
/// for use with replicated secret sharing over some field `F`.

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+    index: u128,
+    a_share: ReplicatedSecretSharing<F>,
+    b_share: ReplicatedSecretSharing<F>,
+}
+
+/// A message sent by each helper when they've multiplied their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct DValue {
+    index: u128,
+    d: u128,
```
benjaminsavage, line 23: Woah... I don't think we need anything that big. Honestly - even a u32 is probably going to be enough for our usecase.

benjaminsavage: It should be determined by `F`

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+        let DValue {
+            d: left_d,
+            index: left_index,
+        } = ctx.helper_ring.receive(HelperAddr::Left).await?;
+
+        // sanity check to make sure they've computed it using the same seed
+        if left_index == self.index {
+            // now we are ready to construct the result - 2/3 secret shares of a * b.
+            let lhs = a0 * b0 + F::from(left_d) + s0;
+            let rhs = a1 * b1 + s1 + F::from(right_d);
```
benjaminsavage, line 89: Just a nit - but can you put the terms in the same order?

That is, make the second line:

  let rhs = a1 * b1 + F::from(right_d) + s1;

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+
+impl<F: Field> SecureMul<F> {
+    /// Executes the secure multiplication on the MPC helper side. Each helper will proceed with
+    /// their part, eventually producing 2/3 shares of the product and that is what this function
+    /// returns.
+    ///
+    /// ## Errors
+    /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+    /// back via the error response
+    pub async fn execute<R: Ring>(
```
benjaminsavage, line 51: I see that we lost the Optimization here. that's fine - it keeps this diff simpler. We can add it later.

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+    impl<St, F, Fut> Stream for ChunkScan<St, F, Fut>
+    where
+        St: Stream,
+        St::Item: Clone,
+        F: FnMut(Vec<St::Item>) -> Fut,
+        Fut: Future<Output = Result<St::Item, BoxError>>,
+    {
+        type Item = St::Item;
+
+        fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
benjaminsavage, line 157: I'm surprised to see this living inside of securemul.rs - it seems generic and not scoped to just secure multiplication.

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub fn secure_multiply<'a, F, R, S>(
+        input_stream: S,
+        ctx: &'a ProtocolContext<'a, R>,
+        index: u128,
+    ) -> impl Stream<Item = ReplicatedSecretSharing<F>> + 'a
+    where
+        S: Stream<Item = ReplicatedSecretSharing<F>> + 'a,
+        F: Field + 'static,
+        R: Ring,
+    {
```
benjaminsavage, line 225: I don't really understand what is going on here, and how (if at all), I as a protocol author need to interact with Streams.

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+                        participant: &participant,
+                        helper_ring: &helper_ring,
+                    };
+                    let mut stream = secure_multiply(input, &ctx, start_index);
+
+                    // compute a*b
+                    let _ = stream.next().await.expect("Failed to compute a*b");
+
+                    // compute (a*b)*c and return it
+                    stream.next().await.expect("Failed to compute a*b*c")
```
benjaminsavage, line 328: Is the intention for protocol authors to write code like this? I don't find this very transparent. It's not clear that this computes a*b*c - could we make that shine through more?

src/securemul.rs
```diff
@@ -0,0 +1,430 @@
...
+        assert!(b < Fp31::PRIME);
+
+        let a = Fp31::from(u128::from(a));
+        let b = Fp31::from(u128::from(b));
+
+        thread_local! {
+            static INDEX: AtomicU64 = AtomicU64::default();
+        }
+
+        let index = u128::from(INDEX.with(|i| i.fetch_add(1, Ordering::Release)));
```
benjaminsavage, line 359: What on earth is going on here?

benjaminsavage (changes requested):
Just a few small changes requested.

src/securemul.rs
benjaminsavage: Fair enough.

src/securemul.rs
benjaminsavage: OK, I believe you.

src/securemul.rs
benjaminsavage: ...is it worth it?

benjaminsavage (approved):
It may not be perfect - but this is blocking other development.

I think if we get this out we start using the await? syntax immediately - leading to better protocol development. 

In the meantime - we can continue to iterate on all the internals without needing to significantly alter the syntax. This seems like a good compromise to me.

akoshelev:
merging it as @martinthomson is away, will address the feedback afterwards if any.

PR #57, opened 2022-07-28
vector

algebravic:
- add mp-spdz prototype, data generator, and readme
- add IPA_VERBOSE and IPA_NUMROWS_POWER as env variables to set at compile time
- remove using Public Input for row number, did not work
- use exec to unroll loops
- update readme to reflect updates to MP-SPDZ
- iron out final bugs
- add gitattributes for syntax highlighting for .mpc files
- rename _list to reports
- rename ambiguous variables, add Enum for Matrix columns
- Rewriting Erik's ipae2e.mpc using vectors instead of
- Rewrote Erik's code using vectors

### DIFF
--- a/research-prototype/Programs/Source/erik.mpc
+++ b/research-prototype/Programs/Source/erik.mpc
@@ -8,7 +8,9 @@ VALUE = 1
 IS_TRIGGER = 2
 BREAKDOWN_KEY = 3
 
-def get_args() -> argparse.Namespace:
+BREAKDOWN_VALUES = 4
+
+def get_args(args) -> argparse.Namespace:
 
     # It turns out that you *can* get arguments from the compile.py
     # command.  The arguments are everything after the name
@@ -18,15 +20,17 @@ def get_args() -> argparse.Namespace:
     arg_parser.add_argument("numrows_power", type=int, default = 10,
                             nargs = '?',
                             help="Power of 2 for number of Rows")
-    arg_parser.add_argument("verbose", type = bool, default = False,
+    arg_parser.add_argument("breakdown_values", type = int, default = 4,
+                            nargs = '?',
+                            help="Calculate Breakdown?")
+    arg_parser.add_argument("verbose", type = int, default = 0,
                             nargs = '?',
                             help="verbose?")
-    pargs = program.args[1: ]
-    return arg_parser.parse_args(pargs)
+    return arg_parser.parse_args(args)
 
 def main() -> None:
 
-    args = get_args()
+    args = get_args(program.args[1: ])
     print(f"args = {args}")
     if args.verbose and args.numrows_power > 5:
         raise ValueError("Cannot handle verbose with numrows = 2**{numrows_power}")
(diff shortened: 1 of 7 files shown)


bmcase:
Since this isn't doing a sort, it is really the create helperbits and oblivious attribution stages.   

The great news is the vector operations do seem to be really helping with the compile memory usage.  Comparing to just the oblivious attribution stage before this is using around 13x less memory.

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+    match_keys = Array(numrows, sint)
+    
+    match_bits = Matrix(numrows, args.n_bits, sint)
+    if args.sort_type == BIT_RADIX_SORT:
+        match_bits.assign_vector(
+            sint.get_input_from(0, size = numrows * args.n_bits))
+        match_keys.assign_vector(reconstruct(match_bits))
+    else:
+
+        match_keys.assign_vector(sint.get_input_from(0, size = numrows))
```
eriktaubeneck, line 208: these seem to consume different numbers of inputs

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+
+    reports = Matrix(numrows, 3, sint)
+    reports.assign_vector(sint.get_input_from(0, size = numrows * 3))
+
+    # Now do the sort
+    # Radix sort
+
+    if args.do_sort:
```
eriktaubeneck, line 216: is there any reason we wouldn't do the sort?

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+            mask = breakdown == breakdown_key
+            # Sum up the final_credits for this breakdown
+            # We can do this with sum, but tree_reduce is more efficient.
+            breakdown_key_sums[breakdown_key] = tree_reduce(lambda x,y: x + y,
+                                                            mask * final_credits)
+            return breakdown_key_sums
+
+        return None
+
+def main():
```
eriktaubeneck, line 185: not sure if this is particularly helpful? it just forces everything to be indented and doesn't actually change the functionality

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+from pathlib import Path
+from Compiler import types, library, instructions, sorting
+
+VALUE = 0
+IS_TRIGGER = 1
+BREAKDOWN_KEY = 2
+
+RADIX_SORT = 0
+BIT_RADIX_SORT = 1
+BATCHER_SORT = 2
```
eriktaubeneck, line 17: this would be much better as strings

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+import argparse
+import os
+import sys
+from typing import Optional
+from pathlib import Path
+from Compiler import types, library, instructions, sorting
+
+VALUE = 0
+IS_TRIGGER = 1
+BREAKDOWN_KEY = 2
```
eriktaubeneck, line 13: why not continue to use the Enum.IntEnum?

eriktaubeneck (changes requested):
Generally, there is A LOT going on here. It would be great if this was based off the existing `mp-spdz-prototype` branch, and edits were made directly to the ipae2e file, so that it was easy to see the differences. additionally, it would be great to have 1 PR for each incremental change, (i.e. one just focused on vectorization, and a different one with extra sort options, etc.)

research-prototype/Programs/Source/erik.mpc
```diff
@@ -0,0 +1,238 @@
...
+    return result
+
+def get_args(args) -> argparse.Namespace:
+    """
+    Get the arguments from the remainder of compile.py command line.
+    """
+
+    # It turns out that you *can* get arguments from the compile.py
+    # command.  The arguments are everything after the name
+    # of the file to be compiled.
```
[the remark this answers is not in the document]
eriktaubeneck: found another bug this this... calling `../../MP-SPDZ/compile.py -C -R 32 ipae2e 10` ends up naming the compiled programs `ipae2e-10`, which means you have to specify which version you're running...

PR #58, opened 2022-07-29
Modify event structures

taikiy:
## Overview

This diff changes source and trigger event structures in the following way:

1. Change "breakdown_key" and "zkp" fields' type from string to u32.
2. Duplicate fields that exist in trigger event to source event and vice versa. This ensures that two aribtrary data chunks are indistinguishable so that an observer cannot learn whether one is a source or trigger event.
3. Add a flag (boolean) to indicate whether the event type is source (false/0) or trigger(true/1).

### DIFF
--- a/src/bin/ipa_bench/gen_events.rs
+++ b/src/bin/ipa_bench/gen_events.rs
@@ -16,16 +16,25 @@ use std::time::Duration;
 const RECORD_SEPARATOR: u8 = 30;
 
 const DAYS_IN_EPOCH: u64 = 7;
-type MatchKey = Vec<u64>;
+type MatchKey = u64;
 type Epoch = u8;
+type Number = u32;
 
 #[derive(Clone)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct EventBase {
-    // For this tool, we'll fix the length of a matchkey to u64
-    pub matchkeys: MatchKey,
+    /// Secret shared and then encrypted match keys.
+    pub matchkeys: Vec<MatchKey>,
+
+    /// The epoch which this event is generated. Using an 8-bit value = 256 epochs > 4 years (assuming 1 epoch = 1 week).
+    /// This field is in the clear.
     pub epoch: Epoch,
-    pub timestamp: u32,
+
+    /// An offset in seconds into a given epoch. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
+    pub timestamp: Number,
+
+    /// 0 (false) if this is a source event. 1 (true) otherwise.
+    pub flag: bool,
 }
 
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
@@ -33,7 +42,33 @@ pub struct SourceEvent {
     pub event: EventBase,
 
     /// Ad breakdown key value
-    pub breakdown_key: String,
+    pub breakdown_key: Number,
+
+    /// Fields from TriggerEvent for padding
+    value: Number,
+    zkp: Number,
+}
+
+impl SourceEvent {
+    #[must_use]
+    pub fn new(
+        matchkeys: Vec<MatchKey>,
+        epoch: Epoch,
+        timestamp: Number,
+        breakdown_key: Number,
+    ) -> Self {
+        Self {
+            event: EventBase {
+                matchkeys,
+                epoch,
+                timestamp,
+                flag: false,
+            },
+            breakdown_key,
+            value: 0,
+            zkp: 0,
+        }
+    }
 }
 
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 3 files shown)


src/bin/ipa_bench/gen_events.rs
```diff
@@ -16,35 +16,96 @@ use std::time::Duration;
...
+    /// The epoch which this event is generated. Using an 8-bit value = 256 epochs > 4 years (assuming 1 epoch = 1 week).
+    /// This field is in the clear.
     pub epoch: Epoch,
-    pub timestamp: u32,
+
+    /// An offset in seconds into a given epoch. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
+    pub timestamp: Number,
+
+    /// 0 (false) if this is a source event. 1 (true) otherwise.
+    pub flag: bool,
```
benjaminsavage, line 37: In our recently published spec, we have a different set of fields with different names: https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md
```
struct {
    int match_key_share;
    int timestamp_share;
    int is_trigger_report_share;
    int breakdown_key_share;
    int trigger_value_share;
} generic_report_share
```

Perhaps we can make this look more similar to that. e.g. flag => is_trigger

src/bin/ipa_bench/gen_events.rs
```diff
@@ -16,35 +16,96 @@ use std::time::Duration;
...
 type Epoch = u8;
+type Number = u32;
 
 #[derive(Clone)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct EventBase {
-    // For this tool, we'll fix the length of a matchkey to u64
-    pub matchkeys: MatchKey,
+    /// Secret shared and then encrypted match keys.
+    pub matchkeys: Vec<MatchKey>,
```
benjaminsavage, line 27: At the moment, we only have a solution for one match key. Shall we simplify it down to just one match key for now?

src/bin/ipa_bench/gen_events.rs
```diff
@@ -16,35 +16,96 @@ use std::time::Duration;
...
 pub struct SourceEvent {
     pub event: EventBase,
 
     /// Ad breakdown key value
-    pub breakdown_key: String,
+    pub breakdown_key: Number,
+
+    /// Fields from TriggerEvent for padding
+    value: Number,
+    zkp: Number,
```
benjaminsavage, line 49: I suggest we remove this. It's not in our document, and we may not even need it. The "capping" thing should already protect pretty well against this type of attack.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -16,35 +16,96 @@ use std::time::Duration;
...
+}
+
+impl SourceEvent {
+    #[must_use]
+    pub fn new(
+        matchkeys: Vec<MatchKey>,
```
benjaminsavage, line 55: Again, perhaps we should remove the Vec and start with just one.

benjaminsavage (changes requested):

src/bin/ipa_bench/gen_events.rs
taikiy: Is one match key constraint permanent or temporary?
If we are going to support multiple matchkeys in the future, instead of changing the type, how about I change the data generation code to only generate one match key but in the vector. That way I can keep most of the current logics where I assume the data type is vector.

benjaminsavage: OK, that's fine. Yes, eventually I would love to support multiple match keys.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -105,7 +154,8 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
 
         for _ in 0..reach {
             // # of devices == # of matchkeys
-            let devices = sample.devices_per_user(rng);
+            // Hard code this number to 1 as we only have a solution for one match key.
+            let devices = 1;
```
benjaminsavage, line 158: This is a misunderstanding. Just one match key can certainly be useful across multiple devices. If the same (single) match key is set to the same value on multiple devices - the events can still be joined together.

Multiple match keys just gives a "boost" on match rates - that's all.

benjaminsavage (changes requested):
I'd like to keep the "multiple devices per user" logic.

src/bin/ipa_bench/gen_events.rs
taikiy: Agree. I should've said "at the moment, we only have ...".

benjaminsavage (approved):
LGTM

src/bin/ipa_bench/gen_events.rs
```diff
@@ -16,35 +16,84 @@ use std::time::Duration;
...
+    /// The epoch which this event is generated. Using an 8-bit value = 256 epochs > 4 years (assuming 1 epoch = 1 week).
+    /// This field is in the clear.
     pub epoch: Epoch,
-    pub timestamp: u32,
+
+    /// An offset in seconds into a given epoch. The clear is u32 (< 2^20 seconds), then encrypted and secret shared.
+    pub timestamp: Number,
+
+    /// 0 (false) if this is a source event. 1 (true) otherwise.
+    pub is_trigger: bool,
```
martinthomson, line 37: Don't you want to use an enum for this?  That is, something like:

```rust
pub enum Event {
  Source {
    event: EventBase,
    breakdown_key: Number,
  },
  Trigger {
    event: EventBase,
    value: Number,
  },
}
```
I realize that you want to have the binary value of source and trigger events put the `breakdown_key` and `value` fields in different positions, but for that you will want to implement `Serialize` and `Deserialize` manually anyway.

src/helpers/models.rs
```diff
@@ -126,7 +127,7 @@ pub struct SourceEvent {
     pub event: Event,
 
     /// A key to group sets of the events.
-    pub breakdown_key: PlainText,
+    pub breakdown_key: Number,
```
martinthomson, line 130: Why is this stuff duplicated?

taikiy: I implemented event structures in models.rs for production IPA, i.e. secret shared and encrypted.

Synthetic event generation are in clear, hence we have another definition in gen_events.rs.

src/bin/ipa_bench/gen_events.rs
taikiy: Makes sense. Let me work on implementing Serialize/Deserialize manually in a separate diff. I think the work would be large enough that it should be reviewed separately. Also, I'll be on the leave for a month so I want this PR to land before I'm gone if possible.

PR #59, opened 2022-07-31
add pre-commit and github action for python formatting. includes .mpc files

eriktaubeneck:
adding in python formatting tools, supporting #54

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -48,14 +48,14 @@ cargo clippy --tests -- -D warnings --D clippy::pedantic || error "Clippy errors
 
 cargo test || error "Test failures"
 
-which flake518 || error "flake518 not found. Install with `python3 -m pip install flake518` in the desired python environment."
+which flake518 || error "flake518 not found. Install with \`python3 -m pip install flake518\` in the desired python environment."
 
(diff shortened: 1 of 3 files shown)


eriktaubeneck:
@richajaindce looks like you set this up originally for rust. Can you double check me here for the python setup? thanks!

pre-commit
```diff
@@ -47,3 +47,15 @@ error() {
 cargo clippy --tests -- -D warnings --D clippy::pedantic || error "Clippy errors"
 
 cargo test || error "Test failures"
+
+which flake518 || error "flake518 not found. Install with \`python3 -m pip install flake518\` in the desired python environment."
```
martinthomson, line 51: These might do better with a venv instead.

martinthomson (approved):

pre-commit
eriktaubeneck: Yea, I thought about that. I figured leaving it up to the user to figure out if they want to use a venv, but we can also go that route.

eriktaubeneck: @martinthomson  any objection to merging like this, and setting that up if/when we need it? (trying to unblock while i'm on vacation)

PR #60, opened 2022-08-02
Update a flaky test

martinthomson:
The output here is too low entropy to avoid collisions, so run it a few
times.


benjaminsavage (approved):
Thank you. This was bothering me =).

thurstonsand (approved):
yeah this was bothering me too. thanks!

PR #63, opened 2022-08-03
Implementing random share generation

benjaminsavage:
In order to convert from an XOR secret sharing of the match key to a list of Replicated, additive secret shares, we need pairs of sharings of a random value "r", which is either 0 or 1, and none of the helpers should know the value of "r".

Now that @akoshelev has landed his secure multiplication struct - with async/await this code is MUCH more legible. Thank you so much!

This is NOT the optimized logic - but I figure I'll start with this simpler diff first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ aes = "0.8"
 async-trait = "0.1.56"
 axum = { version = "0.5.7", optional = true, features = ["http2"] }
 axum-server = { version = "0.4.0", optional = true, features = ["rustls", "rustls-pemfile", "tls-rustls"] }
+bit-vec = "0.6.3"
 byteorder = "1"
 # rust-elgamal (via curve25519-dalek-ng) only works with digest 0.9, not 0.10
 digest = "0.9"
--- a/src/modulus_convert.rs
+++ b/src/modulus_convert.rs
@@ -1,0 +1,187 @@
+use crate::error::BoxError;
+use crate::field::Field;
+use crate::helpers::ring::Ring;
+use crate::prss::Participant;
+use crate::replicated_secret_sharing::ReplicatedSecretSharing;
+use crate::securemul::{ProtocolContext, SecureMul};
+use bit_vec::BitVec;
+
+pub enum HelperIdentity {
+    H1,
+    H2,
+    H3,
+}
+
+pub struct RandomShareGenerationHelper<'a> {
+    rng: &'a Participant,
+    identity: HelperIdentity,
+}
+
+impl<'a> RandomShareGenerationHelper<'a> {
+    #[must_use]
+    pub fn new(rng: &'a Participant, identity: HelperIdentity) -> Self {
+        Self { rng, identity }
+    }
+
+    #[must_use]
+    pub fn split_binary<T: Field>(
+        &self,
+        left: bool,
+        right: bool,
+    ) -> (
+        ReplicatedSecretSharing<T>,
+        ReplicatedSecretSharing<T>,
+        ReplicatedSecretSharing<T>,
+    ) {
+        let left = u128::from(left);
+        let right = u128::from(right);
+        match self.identity {
+            HelperIdentity::H1 => (
+                ReplicatedSecretSharing::new(T::from(left), T::ZERO),
+                ReplicatedSecretSharing::new(T::ZERO, T::from(right)),
+                ReplicatedSecretSharing::new(T::ZERO, T::ZERO),
+            ),
+            HelperIdentity::H2 => (
+                ReplicatedSecretSharing::new(T::ZERO, T::ZERO),
+                ReplicatedSecretSharing::new(T::from(left), T::ZERO),
+                ReplicatedSecretSharing::new(T::ZERO, T::from(right)),
+            ),
+            HelperIdentity::H3 => (
+                ReplicatedSecretSharing::new(T::ZERO, T::from(right)),
+                ReplicatedSecretSharing::new(T::ZERO, T::ZERO),
+                ReplicatedSecretSharing::new(T::from(left), T::ZERO),
+            ),
+        }
+    }
+
+    fn get_share_of_one<T: Field>(&self) -> ReplicatedSecretSharing<T> {
+        match self.identity {
+            HelperIdentity::H1 => ReplicatedSecretSharing::new(T::ONE, T::ZERO),
+            HelperIdentity::H2 => ReplicatedSecretSharing::new(T::ZERO, T::ZERO),
+            HelperIdentity::H3 => ReplicatedSecretSharing::new(T::ZERO, T::ONE),
+        }
+    }
+
+    #[must_use]
+    pub fn convert_shares_step_1<T: Field>(
+        input: u64,
+        r_shares: &[(bool, ReplicatedSecretSharing<T>)],
+    ) -> (Vec<ReplicatedSecretSharing<T>>, BitVec) {
+        let mut bit_vec = BitVec::from_bytes(&input.to_be_bytes());
+        for i in 0..64 {
+            let r_binary_share = r_shares[i].0;
+            bit_vec.set(i, bit_vec[i] ^ r_binary_share);
+        }
+        (r_shares.iter().map(|x| x.1).collect(), bit_vec)
+    }
+
+    #[must_use]
+    pub fn convert_shares_step_2<T: Field>(
+        &self,
+        input_xor_r: &BitVec,
+        r_shares: &[ReplicatedSecretSharing<T>],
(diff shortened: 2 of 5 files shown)


src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+        idx: u128,
+        ctx: &ProtocolContext<'_, R>,
+    ) -> Result<Vec<(bool, ReplicatedSecretSharing<T>)>, BoxError> {
+        let mut idx = idx;
+
+        let (left, right) = self.rng.generate_values(idx);
+        let mut output = Vec::with_capacity(128);
+        let left_bits = BitVec::from_bytes(&left.to_be_bytes());
+        let right_bits = BitVec::from_bytes(&right.to_be_bytes());
+        for i in 0..128 {
```
benjaminsavage, line 74: I am not confident in the parallelisation here. 

Ideally, I would like all 128 multiplications to happen in parallel. I'm worried that the way I've written it - they'll be run in sequence. I would appreciate reviewers help in figuring out how to ensure there's just a single batch of communication for all of them together.

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+            let h1_share = result_shares.0[i];
+            let h2_share = result_shares.1[i];
+            let h3_share = result_shares.2[i];
+
+            let expected_value: u128 = u128::from(h1_share.0 ^ h2_share.0 ^ h3_share.0);
+
+            crate::replicated_secret_sharing::tests::assert_valid_secret_sharing(
+                h1_share.1, h2_share.1, h3_share.1,
+            );
+            crate::replicated_secret_sharing::tests::assert_secret_shared_value(
```
martinthomson, line 138: I would add a "use" for this rather than spell it all out.

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    use crate::field::Fp31;
+    use crate::helpers;
+    use crate::helpers::ring::mock::TestHelper;
+
+    #[tokio::test]
+    async fn gen_pairs() -> Result<(), BoxError> {
+        let ring = helpers::ring::mock::make_three();
+        let participants = crate::prss::test::make_three();
+        let context = crate::securemul::tests::make_context(&ring, &participants);
```
martinthomson, line 114: These all need a "use" declaration.  You'll need "use ... as ..." for make_three though.

src/modulus_convert.rs
martinthomson, line 74:
```suggestion
        for i in 0..u128::BITS {
```

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// If there is a timeout (or other error) it should be signalled in the response
+    pub async fn gen_r_pairs<T: Field, R: Ring>(
+        &self,
+        idx: u128,
+        ctx: &ProtocolContext<'_, R>,
+    ) -> Result<Vec<(bool, ReplicatedSecretSharing<T>)>, BoxError> {
+        let mut idx = idx;
+
+        let (left, right) = self.rng.generate_values(idx);
+        let mut output = Vec::with_capacity(128);
```
martinthomson, line 71:
```suggestion
        let mut output = Vec::with_capacity(u128::BITS);
```

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+
+pub struct RandomShareGenerationHelper<'a> {
+    rng: &'a Participant,
+    identity: HelperIdentity,
+}
+
+impl<'a> RandomShareGenerationHelper<'a> {
+    #[must_use]
+    pub fn new(rng: &'a Participant, identity: HelperIdentity) -> Self {
+        Self { rng, identity }
```
martinthomson, line 23: I think that maybe we need to revise how Participant is created and used.  What we really want here is the ability to create multiple streams of shared state from a single key exchange.  I'll take a look at this today.

martinthomson (commented):
This is looking much better.

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Lots of things may go wrong here, as this method calls the "Secure Mult" implementation
+    /// If there is a timeout (or other error) it should be signalled in the response
+    pub async fn gen_r_pairs<T: Field, R: Ring>(
+        &self,
+        idx: u128,
+        ctx: &ProtocolContext<'_, R>,
+    ) -> Result<Vec<(bool, ReplicatedSecretSharing<T>)>, BoxError> {
+        let mut idx = idx;
+
+        let (left, right) = self.rng.generate_values(idx);
```
akoshelev, line 70: looking at examples in `Modulus conversion` doc and the documentation to `Participant` struct I am not sure I fully understand this. In the doc, each helper wants to generate (Xi, Xi+1) share: H1 -> (x1, x2), H2 -> (x2, x3) and H3 -> (x3, x1) using helper's own seed and seed of the helper on the right side.

I could be misreading this, but `Participant` struct has a note about seeds being provided for left and right parties for each helper. 

```
    /// Generate an additive share of zero.
    /// Each party generates two values, one that is shared with the party to their left,
    /// one with the party to their right.
```

Am I reading this correctly?

src/modulus_convert.rs
```diff
@@ -0,0 +1,148 @@
...
+        right: bool,
+    ) -> (
+        ReplicatedSecretSharing<T>,
+        ReplicatedSecretSharing<T>,
+        ReplicatedSecretSharing<T>,
+    ) {
+        let left = u128::from(left);
+        let right = u128::from(right);
+        match self.identity {
+            HelperIdentity::H1 => (
```
akoshelev, line 39: I am not sure if it is possible, but it would be easier if each helper does not have an identity but only knows about its left and right peer. In that case it could only generate share for party on the left: always (0, 0), its own share (X_i, 0) and share for helper on the right (0, X_i+1).

benjaminsavage: Unfortunately, it *does* need to know its identity. This particular function here is a perfect example of a time when each of the three helpers needs to do something different, and the optimization which reduces the communication overhead from 6 to 3 is another.

src/modulus_convert.rs
benjaminsavage: So this line generates 2 random u128 values, the first is the same random number as the helper to the left will generate (for this index) and the second is the same random number as the helper to the right will generate (for this index).

Next, we will interpret these u128s as a sequence of just 128 bits. So now we have 128 binary secret sharings of the form:

H1 -> (x1, x2), H2 -> (x2, x3), H3 -> (x3, x1)

So each of these is a secret sharing in Z_2, which is either 0 or 1. No helper knows all 3 shares, so none of them know if it's a secret sharing of 0 or 1. Each one just has two bits.

The rest of this protocol is there to generate a *matching* secret sharing of the same value (either 0 or 1), but in a different modulus (i.e. Fp31).

Cargo.toml
```diff
@@ -17,6 +17,7 @@ aes = "0.8"
 async-trait = "0.1.56"
 axum = { version = "0.5.7", optional = true, features = ["http2"] }
 axum-server = { version = "0.4.0", optional = true, features = ["rustls", "rustls-pemfile", "tls-rustls"] }
+bit-vec = "0.6.3"
```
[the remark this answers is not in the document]
benjaminsavage: I'm not sure how to use `bitfield`, can you share some sample code? I couldn't find any good documentation.

Alternatively I can just do without and use something like

`left_bit = (left >> i) & 1_u128`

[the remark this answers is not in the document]
akoshelev: yea this is basically what `bitfield` does. 

```rust
use bitfield::Bit;

...
        let (left, right) = self.rng.generate_values(idx);
        let size: usize = u128::BITS.try_into().unwrap();
        let mut output = Vec::with_capacity(size);
        let mut mult_1_futures = Vec::with_capacity(size);
        for i in 0..size {
            // bit is a method on the trait implemented for u128 by the `bitfield` crate
            let left_bit = left.bit(i);
            let right_bit = right.bit(i);
```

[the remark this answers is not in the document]
martinthomson: I don't see how (x>>i)&1 is so hard to understand. Fewer dependencies is good.

src/modulus_convert.rs
martinthomson: Put // in front of that and add it to the code?

src/modulus_convert.rs
richajaindce: This section would be needed for sort also to calculate 1 - x, can we extract this in a function like before?

richajaindce: One more example of this would be to get a share of 1 for each helper -- this is needed in sort protocol

PR #64, opened 2022-08-03
Python linting fix

thurstonsand:
for some reason, the main branch has a failing github workflow around python linting. this resolves that issue, as well as breaks up the workflow for rust/python CI


thurstonsand:
I noticed that I was failing tests after rebasing, so this should resolve that issue

eriktaubeneck:
thanks!

PR #65, opened 2022-08-05
PATCG benchmarks

bmcase:
does not seem there is any significant change to performance


bmcase:
@eriktaubeneck, @benjaminsavage  I added code for capping to both versions of the code.  Plus some other fixes from last week to be merged.

bmcase:
Also there is a strange bug I haven't been able to find yet in `ipae2e.mpc` that when you run it on `ipatestinput2.py`  it works when the input has size less than or equal to 2^6 but not for 2^7 or larger (gives all zeros).  I'm not sure why yet.  `vectorized.mpc` handles this test case for any input size as you'd expect.  Maybe you can keep an eye out for what might be causing as you review.

research-prototype/Programs/Source/justsort.mpc
```diff
@@ -0,0 +1,73 @@
+"""
+Just soring, with Libary Radix sort
```
thurstonsand, line 2:
```suggestion
Just sorting, with Library Radix sort
```

research-prototype/Programs/Source/justsort.mpc
```diff
@@ -0,0 +1,73 @@
...
+from typing import Optional, Tuple
+import operator
+from pathlib import Path
+from Compiler import types, library, instructions, sorting
+
+numrows = 2 ** 15
+
+# load the data
+
+reports = Matrix(numrows, 4, sint)
```
thurstonsand, line 16: where is this used? I see `reports` is redefined on line 57

bmcase: ah, that was from a different version. I mean to remove those lines not in main.

research-prototype/Programs/Source/vectorized.mpc
```diff
@@ -61,6 +62,9 @@ def get_args(args) -> argparse.Namespace:
     # It turns out that you *can* get arguments from the compile.py
     # command.  The arguments are everything after the name
     # of the file to be compiled.
+    # Note that these arguements get added to the of the compiled file,
```
thurstonsand, line 65:
```suggestion
    # Note that these arguements get added to the end of the compiled file, 
```
?

research-prototype/Programs/Source/vectorized.mpc
```diff
@@ -0,0 +1,320 @@
...
+                              & stopbit.get_vector(base = stepsize,
+                                                   size = new_size))
+
+        stopbit.assign_vector(zeros.get_vector(size = stepsize),
+                              base = new_size)
+
+        # Replace the first new_size elements, leaving the others alone
+        credit.assign_vector(new_credit)
+
+    # Calculate final_credits of source events by zering out the values of trigger rows.
```
thurstonsand, line 172:
```suggestion
    # Calculate final_credits of source events by zeroing out the values of trigger rows. 
```

research-prototype/PATCG_2022.8.11_README.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Scaling # rows
+
+
+## 2^12 4
+
+
+
+* `/usr/bin/time --verbose ../../../MP-SPDZ/compile.py -C -R 32 vectorized 12`
+* P0:  `/home/ec2-user/workspace/MP-SPDZ/Scripts/../sy-rep-ring-party.x --player 0 vectorized-12 --hostname ec2-35-161-7-220.us-west-2.compute.amazonaws.com`
```
eriktaubeneck, line 20: we should stub out these hostnames and put them into a environmental variable so that folks can replicate.

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -4,7 +4,7 @@ from enum import IntEnum
 from Compiler.library import print_ln, tree_reduce
 from Compiler.types import Array, Matrix, sint
 
-numrows_power = int(os.getenv("IPA_NUMROWS_POWER", 10))
+numrows_power = int(os.getenv("IPA_NUMROWS_POWER", 4))
```
eriktaubeneck, line 7: any reason why we're changing the default?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -18,7 +18,7 @@ if verbose:
     else:
         raise Exception(f"Unrecognized verbose value: {verbose}")
 
-if verbose and numrows_power > 5:
+if verbose and numrows_power > 9:
```
eriktaubeneck, line 21: 2^9 is *a lot* of rows for verbose output. do we really need this?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -62,34 +62,21 @@ helperbits.assign_vector(
 
 helperbit_and_istrigger_flags = Array(numrows, sint)
 helperbit_and_istrigger_flags.assign_vector(
-    helperbits.get_vector(size=numrows - 1)
-    * reports.get_column(Columns.IS_TRIGGER).get_vector(size=numrows - 1),
```
eriktaubeneck, line 66: isn't the `helperbit` supposed to tell you if the preceding row is a is a trigger event?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -123,17 +110,25 @@ for exponent, (stopbit, credit) in enumerate(zip(stopbits, credit_arrays)):
             )
         )
 
+        credit.assign_vector(credit_arrays[exponent - 1].get_vector(base=numrows - stepsize, size=stepsize),
```
eriktaubeneck, line 113: why are we assigning this vector twice?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -123,17 +110,25 @@ for exponent, (stopbit, credit) in enumerate(zip(stopbits, credit_arrays)):
...
 
 # I can't sum all of credit, I need to sum only those that correspond to last
-# touches helperbit_xor_istrigger * credit_arrays[-1] will give this.
+# touches NOT_is_trigger * credit_arrays[-1] will give this.
 final_credits = Array(numrows, sint)
+
+
 final_credits.assign_vector(
-    helperbit_xor_istrigger_flags.get_vector() * credit_arrays[-1].get_vector()
+    (
```
eriktaubeneck, line 128: double parens?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -146,6 +141,39 @@ if verbose:
...
+current_contribution.assign_vector(0)
+cap = 10
+for row in rows:
+
+    current_contribution[row] = (
+                                    current_contribution[row] 
+                                    * helperbits[row]
+                                )
+
+    min = (
```
eriktaubeneck, line 155: double parens?

research-prototype/Programs/Source/ipae2e.mpc
```diff
@@ -146,6 +141,39 @@ if verbose:
             final_credits[row].reveal(),
         )
 
+############  CAPPING ############
+current_contribution = Array(numrows,sint)
+current_contribution.assign_vector(0)
+cap = 10
+for row in rows:
```
eriktaubeneck, line 148: shouldn't we vectorize this too?

research-prototype/Programs/Source/vectorized.mpc
```diff
@@ -0,0 +1,320 @@
+"""
```
eriktaubeneck, line 1: we incorporated all the vectorization into `ipae2e.mpc`. I don't think it makes sense to have two different implementations in the repo.

research-prototype/ipatestinput1.py
```diff
@@ -0,0 +1,43 @@
+import argparse
```
eriktaubeneck, line 1: why do we need 2 new input files? can we instead refactor the original one and add a descriptive option to the command line interface to choose which version to use? it's unclear to me what the differences are.

eriktaubeneck (commented):
The sorting files don't seem to be used in `ipae2e.mpc`. Should we be moving them into a different PR?

research-prototype/Programs/Source/ipae2e.mpc
bmcase: there is a bug in ipae2e.mpc when you go from 2^6 to 2^7 on `ipatestcase2.py`  I wasn't able to find it but was trying to print more to do so.

martinthomson:
@bmcase, this should not be opened against the [other repo](https://github.com/private-attribution/research-prototype), probably manually given the complexity of the move.

Issue #62, opened 2022-08-03
Ring feedback


martinthomson:
Just looking at the code in ring.rs and I'm liking it a fair bit.

There are some things that I think need some consideration though.  Some of this only affects the test implementation.

1. For `Message: Send + Serialize`: I don't see why you would need `Send` (and `'static` for that matter).  If you implement `Serialize`, you can serialize on the same thread as the `send()` call is invoked and then you only need to transfer a buffer of bytes.

2. `Box<[u8]>` == `Vec<u8>`.

3. Buffering.  There is buffering in the `mpsc::channel`.  Why do you need to hold messages in a separate one-item buffer (the `HashMap`)?  Given that the receiver is single-threaded (though it is `Send`), you can just pull from that channel synchronously when calling `receive()`.  The single item buffer already essentially assumes that there is no intermixed messages, so you don't even need to keep the type annotation on each message, just have the serialization fail.

4. Using `Option<Sender>` for peers.  Better to have a setup object instead; see how I did PRSS.  That will allow you to lose the input queue also.

akoshelev:
1. I believe `Message` still needs to be `Send` because we send it across the await point (method `send` on `Ring` trait is `async`). we can get around that by moving serialization to the clients of `send/receive` method but it looks worse to me than the existing implementation.

2. Is there a particular reason why you would prefer `Vec` over `Box`? The latter is not a growable container and takes only 66% of the space on the stack compared to `Vec`. None of these matter for the test implementation, I did it mostly just to educate folks less familiar with Rust that it is also possible to have an exclusive pointer to slice that does not support adding/removing elements from it.

3. I was afraid that FIFO without buffering may not cover all of our use cases. For multiplication it certainly works out of the box, would it be enough to cover cases where H1 and H3 need to send two different messages to H2 that can process them in any order? I am not convinced that we would need that ever, so I don't mind changing it but would be great if you can confirm that this is entirely redundant in this case

4. Makes sense, I would probably use `Builder` pattern as the most common and because pretty much everyone is familiar with it

akoshelev:
@martinthomson does this sound reasonable?

martinthomson:
Not sure about 1.  If the `send()` call turns the message into bytes immediately, then it never has to cross between threads, only the bytes do.

[...]

PR #66, opened 2022-08-07
Add support for running protocols concurrently

akoshelev:
## Overview

This PR partially addresses #62 and also adds support for running multiple protocols of the same type concurrently. For example, now it is possible to schedule as many multiplications as possible as long as all of them have unique protocol id. 

## Protocol id

Currently it is an opaque struct and secure multiplication uses seed index which may not be what we would want in the future, but it works for now

## Communication gateway

The fact that there can be many protocols running at the same time requires another layer of abstraction - starting protocols and dispatching messages to the receiving protocol correctly requires an orchestration struct that `CommunicationGateway` provides an abstraction for.  

`ProtocolContext` now carries a reference to the gateway instead of `Ring` and on API level makes it impossible to call `send` and `receive` without creating a channel first by calling `ring_channel`. See `SecureMul` for an example how this is now done. 

## Using channels instead of hashmap

It's been discussed in #62, but now per-protocol buffering is implemented using channels rather than storing messages inside a hashmap. That adds an additional requirement for messages to be sent in order, otherwise deserialization may fail.

### DIFF
--- a/src/helpers/ring.rs
+++ b/src/helpers/ring.rs
@@ -45,7 +45,7 @@ pub mod mock {
     use std::collections::hash_map::Entry;
     use std::collections::HashMap;
     use std::sync::{Arc, Mutex};
-    use tokio::sync::mpsc::{channel, Sender};
+    use tokio::sync::mpsc::{channel, Receiver, Sender};
 
     /// Internally we represent all messages to be a sequence of bytes and store them inside
     /// a hashmap where each element is addressable by message type id and destination (i.e. who
@@ -68,29 +68,74 @@ pub mod mock {
     /// receive it.
     #[derive(Debug)]
     pub struct TestHelper {
-        // A handle to send message to this helper
-        input_queue: Sender<MessageEnvelope>,
-
         // Reference to helper channel on the left side
-        left: Option<Sender<MessageEnvelope>>,
+        left: Sender<MessageEnvelope>,
 
         // Reference to helper channel on the right side
-        right: Option<Sender<MessageEnvelope>>,
+        right: Sender<MessageEnvelope>,
 
         // buffer for messages sent to this helper
         buf: Arc<Mutex<MessageBuf>>,
     }
 
-    impl TestHelper {
-        /// Constructs a new instance of test helper using the specified `buf_capacity` buffer
-        /// capacity for the internally used channel.
+    #[derive(Debug)]
+    pub struct TestHelperSetup {
+        input: Sender<MessageEnvelope>,
+        rx: Receiver<MessageEnvelope>,
+        left: Option<Sender<MessageEnvelope>>,
+        right: Option<Sender<MessageEnvelope>>,
+    }
+
+    impl TestHelperSetup {
+        /// Constructs a new instance of test helper builder using the specified `buf_capacity`
+        /// buffer capacity for the internally used channel.
+        ///
+        /// Every test helper instance to set up correctly requires both right and left side channels
+        /// to be provided via this builder by calling `set_left` and `set_right` methods. `setup`
+        /// method will consume this builder and produce a new instance of `TestHelper`.
         ///
         /// ## Panics
         /// Panics if Mutex used internally for synchronization is poisoned or if there are more
         /// than one message with the same type id and destination address arriving via `send` call.
         #[must_use]
         pub fn new(buf_capacity: usize) -> Self {
-            let (tx, mut rx) = channel::<MessageEnvelope>(buf_capacity);
+            let (tx, rx) = channel(buf_capacity);
+            Self {
+                input: tx,
+                rx,
+                left: None,
+                right: None,
+            }
+        }
+
+        fn set_left(&mut self, left: Sender<MessageEnvelope>) {
+            self.left = Some(left);
+        }
+
+        fn set_right(&mut self, right: Sender<MessageEnvelope>) {
+            self.right = Some(right);
+        }
+
+        fn get_input(&self) -> Sender<MessageEnvelope> {
+            self.input.clone()
+        }
+
+        #[must_use]
+        pub fn setup(self) -> TestHelper {
+            TestHelper::new(
+                self.rx,
+                self.left.expect("left helper must be set"),
+                self.right.expect("right helper must be set"),
+            )
+        }
+    }
+
+    impl TestHelper {
+        fn new(
+            mut rx: Receiver<MessageEnvelope>,
+            left: Sender<MessageEnvelope>,
+            right: Sender<MessageEnvelope>,
+        ) -> Self {
             let buf = Arc::new(Mutex::new(HashMap::new()));
 
             tokio::spawn({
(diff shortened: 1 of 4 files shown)


src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
+        /// is set to `None` once test helper is connected to this instance
+        inbound: Option<HashMap<HelperAddr, Receiver<MessageEnvelope>>>,
+    }
 
-        // buffer for messages sent to this helper
-        buf: Arc<Mutex<MessageBuf>>,
+    /// Keeps track of active protocols
+    #[derive(Debug, Default)]
+    struct RoutingTable {
+        active_routes: HashMap<ProtocolId, Route>,
```
martinthomson, line 111: Ideally, the protocol ID is a very small number, with this being a `Vec<Route>` or similar.

If we assume that the protocol ID corresponds to a step in the overall process, we could make this a EnumMap and the protocol ID could be a short enum.  The problem is that you would need to make this type generic (which flows on upstream).

src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
             // but this method accepts the destination. To obtain source from destination
             // we invert it - message send to the left helper is originated from helper on the
             // right side.
-            let (target, source) = match dest {
-                HelperAddr::Left => (self.left.as_ref().unwrap(), HelperAddr::Right),
-                HelperAddr::Right => (self.right.as_ref().unwrap(), HelperAddr::Left),
+            let source = match dest {
+                HelperAddr::Left => HelperAddr::Right,
+                HelperAddr::Right => HelperAddr::Left,
             };
```
martinthomson, line 149: This is better as a method on HelperAddr.

src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
 
             let bytes = serde_json::to_vec(&msg).unwrap().into_boxed_slice();
             let envelope = MessageEnvelope {
-                type_id: TypeId::of::<T>(),
                 source,
```
martinthomson, line 153: Do you really need to signal the source?  I guess that - for a test - it's OK to have a little redundancy, but won't the receiver know where the channel comes from?  (If it doesn't, maybe that can be the first message off each new channel...)

src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
+
+    /// Receive a given message from the `source`
+    async fn receive<T: Message>(&mut self, source: HelperAddr) -> Result<T, Error>;
 }
 
 #[cfg(test)]
 pub mod mock {
     use crate::helpers::error::Error;
-    use crate::helpers::ring::{HelperAddr, Message, Ring};
+    use crate::helpers::error::Error::{ReceiveError, SendError};
```
martinthomson, line 60: use crate::helpers::error::Error::{self, ReceiveError, SendError};

src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
                 dest,
-                inner: Box::new(e) as _,
+                inner: "No sender for this destination".into(),
             })?;
-            Ok(())
+
+            sender.send(envelope).await.map_err(|e| Error::SendError {
+                dest,
+                inner: Box::new(e) as _,
+            })
```
martinthomson, line 166: Inconsistent use of SendError vs Error::SendError here.

(I can't follow what the .into() is doing for the former.)

src/helpers/ring.rs
```diff
@@ -26,183 +27,270 @@ pub enum HelperAddr {
...
+        senders: HashMap<HelperAddr, Sender<MessageEnvelope>>,
+        receivers: HashMap<HelperAddr, Receiver<MessageEnvelope>>,
+    }
 
-    /// Each message is packed inside an envelope with some meta information about it.
     #[derive(Debug)]
     struct MessageEnvelope {
         source: HelperAddr,
-        type_id: TypeId,
+        protocol_id: ProtocolId,
```
martinthomson, line 92: As I mentioned elsewhere regarding `source`, I don't think that you need to include the message type/protocol ID on every message.  It's probably OK to have a little redundancy in tests, but that is what it is.

src/helpers/ring.rs
benjaminsavage: For my own education, what does adding "self" into a list of imports like this do?

martinthomson: This:

```rust
use crate::helpers::error::Error::{self, ReceiveError, SendError};
```

...is equivalent to:

```rust
use crate::helpers::error::Error;
use crate::helpers::error::Error::{ReceiveError, SendError};
```

benjaminsavage: Thanks!

benjaminsavage:
This is great! Thanks @akoshelev! I merged these changes into my local copy of #63 and it resolves the "Duplicated message MessageEnvelope" error that is causing the tests to fail right now. It looks like this improvement makes it so that the multiplications can run in parallel, and it is also compatible with the optimisation where only some helpers need to communicate when multiplying shares that all parties know to contain a lot of zeroes.

Looking forward to getting this landed to unblock #63.

src/securemul.rs
```diff
@@ -49,10 +50,11 @@ impl<F: Field> SecureMul<F> {
...
     /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
     /// back via the error response
-    pub async fn execute<R: Ring>(
+    pub async fn execute<R: CommunicationGateway>(
         self,
         ctx: &ProtocolContext<'_, R>,
     ) -> Result<ReplicatedSecretSharing<F>, BoxError> {
+        let mut helper_ring = ctx.helper_ring.ring_channel(ProtocolId::from(self.index));
         // generate shared randomness.
         let (s0, s1) = ctx.participant.generate_fields(self.index);
```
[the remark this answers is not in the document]
akoshelev: Thanks @martinthomson I was looking at #68 and I think I got the idea here, but want to clarify my understanding of it. Ben's use case in #63 for modulus conversion requires `256` multiplications to be performed (2 per bit). Does that mean that we would generate a participant with `256` spaces and conversion protocol will do the following (just a sketch to demonstrate the idea)

```rust
   for bit in 0..u128::BITS {
         // participant needs to be renamed to `prss_space`
         let ctx = ProtocolContext { participant: &participant[bit], share_a: ..., share_b: ... }
         securemul.execute(ctx)


   } 
    
```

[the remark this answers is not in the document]
martinthomson: The modulus conversion would occur within a single space, one that is reserved for that purpose.

Thinking holistically, a complete run will involve multiple multiplications.  Each multiplication can be assigned a PRSS space.

(If you don't know how many you need, then make a large space and pick spaces sequentially as you go through the code.)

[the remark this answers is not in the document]
akoshelev: in this case , would we give modulus conversion a participant with `N` spaces so it can borrow one for itself and leave N-1 for secure multiplications or we need another struct capable of producing `Participant` instances with reserved space (similar to `setup` but without public keys)?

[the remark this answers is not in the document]
martinthomson: If the multiplications can be given indices from the same space, then you might reasonably use a single space for those multiplications.  It's going to depend on how you drive the multiplications, I think.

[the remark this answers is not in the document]
akoshelev: Is my understanding correct that we would want `ProtocolId` to be some sort of composite id that includes both "step" and "index". So the space (step) needs to be exposed to `SecureMul` somehow so it can construct it from these two pieces

src/helpers/ring.rs
akoshelev: ack, I would still hide it behind `cfg(test)` because it is only used in one place but more importantly I am not sure whether it makes sense in the long term to have it - likely we would have communication patterns that require sending messages to a specific helper, determining source by destination won't work in this case

src/helpers/ring.rs
akoshelev: I can probably remove `source` too but that would make helper setup a bit more complicated. Currently there is one loop in the `TestHelper` that receives all messages sent towards that helper and then dispatches them based on the protocol id value encoded inside the message to `TestChannel` that picks the appropriate receiver. While it is possible to teach `TestHelper` to spawn two tasks and listen to multiple receiver ends and apply the same logic receiving a message, having `source` looks a bit easier to understand

src/helpers/ring.rs
akoshelev: nice!

src/helpers/ring.rs
akoshelev: ack, better to use just `SendError` imo.

Any string can be converted to a boxed error apparently which is convenient in tests: https://doc.rust-lang.org/src/std/error.rs.html#243-283

src/helpers/ring.rs
akoshelev: the std implementation of hashmap is pretty good imo. It uses linear probing so it is cache friendly and allow sparse keys which is a clear advantage over pure `Vec`. it is also more ergonomic to use as I don't need to worry about growth

PR #68, opened 2022-08-08
Split the PRSS code into multiple spaces

martinthomson:
The comments I had on #66 might be easier to resolve with this change in place.  There are a few things that need to be merged (ProtocolId from #66 and SpaceIndex from this one) to make this all come together, but I wanted to get the pieces in place first.

### DIFF
--- a/src/prss.rs
+++ b/src/prss.rs
@@ -7,18 +7,35 @@ use byteorder::{ByteOrder, LittleEndian};
 use hkdf::Hkdf;
 use rand::{CryptoRng, RngCore};
 use sha2::Sha256;
+use std::ops::Index;
+use std::{marker::PhantomData, mem::size_of};
 use x25519_dalek::{EphemeralSecret, PublicKey};
 
-/// A participant in a 2-of-3 replicated secret sharing.
-#[derive(Debug)] // TODO custom debug implementation
-pub struct Participant {
-    left: Generator,
-    left_bits: BitGenerator,
-    right: Generator,
-    right_bits: BitGenerator,
+pub trait SpaceIndex {
+    const MAX: usize;
+
+    #[must_use]
+    fn as_usize(&self) -> usize;
+
+    #[must_use]
+    fn to_bytes(&self) -> [u8; 8] {
+        let mut buf = [0; 8];
+        for (i, &v) in self.as_usize().to_le_bytes().iter().enumerate() {
+            buf[i] = v;
+        }
+        buf
+    }
 }
 
-impl Participant {
+/// A participant in a 2-of-N replicated secret sharing.
+#[derive(Debug)] // TODO custom debug implementation
+#[allow(clippy::module_name_repetitions)]
+pub struct PrssSpace {
+    left: Generator,
+    right: Generator,
+}
+
+impl PrssSpace {
     /// Generate two random values, one that is known to the left helper
     /// and one that is known to the right helper.
     #[must_use]
@@ -63,12 +80,6 @@ impl Participant {
         l.wrapping_add(r)
     }
 
-    /// Generate the next share in `ZZ_2`
-    #[must_use]
-    pub fn next_zero_bit_share(&mut self) -> bool {
-        self.left_bits.next_bit() ^ self.right_bits.next_bit()
-    }
-
     /// Generate additive shares of zero in a field.
     #[must_use]
     pub fn zero<F: Field>(&self, index: u128) -> F {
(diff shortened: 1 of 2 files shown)


martinthomson:
Usage here is fairly simple.  If you are running a multi-step protocol with 5 steps, you implement `SpaceIndex` with MAX = 5 and give each step a number from 0 to 4 (inclusive).  You create a `Participant` passing this type to the `setup() function.  Then, when you get a `p: Participant`, you access values by invoking `let (l, r) = p[step].generate_fields(index)`.

This means using two indices for everything, but I think that being explicit is both useful and necessary.

src/prss.rs
```diff
@@ -7,18 +7,35 @@ use byteorder::{ByteOrder, LittleEndian};
 use hkdf::Hkdf;
 use rand::{CryptoRng, RngCore};
 use sha2::Sha256;
+use std::ops::Index;
+use std::{marker::PhantomData, mem::size_of};
 use x25519_dalek::{EphemeralSecret, PublicKey};
 
-/// A participant in a 2-of-3 replicated secret sharing.
+pub trait SpaceIndex {
+    const MAX: usize;
```
akoshelev, line 15: what do you think about using a fixed-size integer type for it or a static assert that enforces a minimum size for `usize` for this program to work correctly?

src/prss.rs
```diff
@@ -471,7 +435,7 @@ pub mod test {
         // There isn't enough entropy in this field to be sure that the test will pass.
         // So run a few rounds looking for a mismatch.
         let mut v2 = Fp31::from(0_u8);
-        for _ in 0..12 {
+        for _ in 0..21 {
```
akoshelev, line 438: just fyi - I am still seeing sporadic failures so maybe we can make it an infinite loop or restructure the test to make it stable

```bash
thread 'prss::test::three_party_random' panicked at 'assertion failed: `(left != right)`
  left: `11_mod31`,
 right: `11_mod31`', src/prss.rs:448:9
```

src/prss.rs
```diff
@@ -189,59 +228,36 @@ impl Generator {
...
-        assert!(index.leading_zeros() >= 7, "indices >= 2^121 not supported");
-        let i = index << 7;
-        Self { g, i, v: 0 }
+    use crate::field::Fp31;
+
+    use super::{Generator, KeyExchange, Participant, ParticipantSetup, PrssSpace, SpaceIndex};
+
+    /// In testing, having a single space is easiest to use.
+    /// This provides an implementation of `PrssSpace`.
+    pub struct SingleSpace(());
```
akoshelev, line 241:
```suggestion
    pub struct SingleSpace;
```

src/prss.rs
```diff
@@ -189,59 +228,36 @@ impl Generator {
...
+
+    impl SingleSpace {
+        pub const ONE: Self = Self(());
     }
```
akoshelev, line 245:
```suggestion
```

src/prss.rs
```diff
@@ -189,59 +228,36 @@ impl Generator {
...
-}
 
-impl From<Generator> for BitGenerator {
-    fn from(g: Generator) -> Self {
-        Self { g, i: 0, v: 0 }
+    // This makes it easier to use the single space.
+    impl std::ops::Deref for Participant<SingleSpace> {
+        type Target = PrssSpace;
+        fn deref(&self) -> &Self::Target {
+            &self[SingleSpace::ONE]
```
akoshelev, line 258:
```suggestion
            &self[SingleSpace]
```

src/prss.rs
martinthomson: It's harder to use the value as an index without that.  I think that it's OK to just deal with the weirdness here, which the compiler should be able to optimize out completely.

src/prss.rs
martinthomson: That's concerning.  Maybe the problem is the birthday paradox is causing this to be a much higher rate of collision.  I had this pegged at 1/31 odds of a collision each run, so (1/31)^21 odds, which is close to 2^-100.  If we get collisions on that, maybe there is a bug.

src/prss.rs
akoshelev: Yep it looks like we don't bake any assumptions here about `usize` being 8 byte long, so it is probably fine not to introduce static compilation checks here

akoshelev (approved):

src/securemul.rs
```diff
@@ -210,23 +210,26 @@ mod tests {
...
-                })
-            },
-        );
+        let handles =
+            input
+                .into_iter()
+                .zip(participants)
+                .zip(ring)
+                .map(|((input, prss), helper_ring)| {
+                    tokio::spawn(async move {
```
thurstonsand, line 219: just so i understand, what's the reason for using `tokio::spawn` here vs just directly using `async move`?

akoshelev: There is no technical need to spawn a task, I just wanted to model helpers as "root" tasks spawning futures and also wanted to make sure "spawn" constraints on references (`static + Send + Sync ) do not make it impossible to send contexts across spawn boundaries

PR #69, opened 2022-08-08
research-prototype: update readme

akoshelev:
* Format code blocks as bash scripts
* Add instructions how to run MPC on multiple hosts

### DIFF
--- a/research-prototype/README.md
+++ b/research-prototype/README.md
@@ -99,7 +99,7 @@ rsync -e "ssh -i <ssh cert>" -Pav Player-Data $USER@$HOST:$MP_SPDZ_DIR
 Pick one host to be the coordinator (and player 0 by convention used by MP_SPDZ).
 
 ```bash
-export $COORDINATOR=<host IP or DNS>
+COORDINATOR=<host IP or DNS>
 ```
 
 Run this command on the coordinator host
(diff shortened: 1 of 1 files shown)


eriktaubeneck:
👌 thank you for adding this!

research-prototype/README.md
```diff
@@ -75,6 +75,47 @@ Note that you should also generate random data accordingly. Also, to avoid dumpi
...
+* `MP_SDPZ_DIR` must be set to the directory where IPA is installed, for example `/home/raw-ipa/research-prototype`.
+
+```bash
+rsync -e "ssh -i <ssh cert>" -Pav Player-Data $USER@$HOST:$MP_SPDZ_DIR
+```
+
+Pick one host to be the coordinator (and player 0 by convention used by MP_SPDZ).
+
+```bash
+export $COORDINATOR=<host IP or DNS>
```
martinthomson, line 102:
```suggestion
COORDINATOR=<host IP or DNS>
```

You don't need to export, and using \$ might not work well.

martinthomson (approved):

research-prototype/README.md
akoshelev: oh good catch, that is certainly not going to work. Thanks for fixing it!

PR #70, opened 2022-08-10
Field: add as_u128 method

akoshelev:
It is convenient to have a common implementation without requiring all field implementations to define their own. Unfortunately I couldn't use (or I don't know how) `From<Field> for u128` because `Field` is a trait so I ended up defining a method on the trait itself.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -77,6 +77,13 @@ pub trait Field:
 
         term
     }
+
+    /// Blanket implementation to represent the instance of this trait as 16 byte integer.
+    /// Uses the fact that such conversion already exists via `Self` -> `Self::Integer` -> `Into<u128>`
+    fn as_u128(&self) -> u128 {
+        let int: Self::Integer = (*self).into();
+        int.into()
+    }
 }
 
 // TODO(mt) - this code defining fields can be turned into a macro if we ever
--- a/src/securemul.rs
+++ b/src/securemul.rs
@@ -59,12 +59,7 @@ impl<F: Field> SecureMul<F> {
         // compute the value (d_i) we want to send to the right helper (i+1)
         let (a0, a1) = self.a_share.as_tuple();
         let (b0, b1) = self.b_share.as_tuple();
-        let right_d: F = a0 * b1 + a1 * b0 - s0;
-
-        // this ugliness is needed just to convert Field to u128. There are better ways to do it
-        // and there is a PR open to make it easier
-        let right_d: <F as Field>::Integer = right_d.into();
-        let right_d: u128 = right_d.into();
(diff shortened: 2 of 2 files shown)


src/field.rs
```diff
@@ -77,6 +77,13 @@ pub trait Field:
 
         term
     }
+
+    /// Blanket implementation to represent the instance of this trait as 16 byte integer.
+    /// Uses the fact that such conversion already exists via `Self` -> `Self::Integer` -> `Into<u128>`
+    fn as_u128(&self) -> u128 {
+        let int: Self::Integer = (*self).into();
```
martinthomson, line 84: Why not

```suggestion
        let int: Self::Integer = self.into();
```

Or for the entire thing:

```rust
<self as Into<Self::Integer>>.into().into()

**Review by martinthomson — COMMENTED** [review]
Can we not provide a blanket implementation of Into<u128> instead? Or is this easier to manage this way?

Ideally, I would like the code that you are fixing to look like:

```rust
let right_d = u128::from(a0 * b1 + a1 * b0 - s0);
```

Which I find clearer than the alternatives.

akoshelev:
> Can we not provide a blanket implementation of Into instead? Or is this easier to manage this way?
> 
> Ideally, I would like the code that you are fixing to look like:
> 
> ```rust
> let right_d = u128::from(a0 * b1 + a1 * b0 - s0);
> ```
> 
> Which I find clearer than the alternatives.

I'd love to, but the problem is (if I understand it correctly) that we are trying to define `From/Into` implementation for a trait object, so the only thing I am able to do inside the conversion is to call a method on that trait

```rust
impl From<dyn Field> for u128 {
    fn from(v: &dyn Field) -> Self {
        ???
    }
}
```

src/field.rs
akoshelev: I believe compiler is getting confused because of `&self` which implies a reference to self while `Into` is defined for `Field` and not for `&Field`. I am not sure why it does not attempt to dereference it (it is `Clone` and `Copy`) so that was the only way I could convince it that conversion works

wrt to your second suggestion to convert it to a trait object (Into is a trait) - I don't think that would work because `Into` is not an object-safe trait

```rust
error[E0038]: the trait `Into` cannot be made into an object
  --> src/field.rs:86:18
   |
86 |         (self as &dyn Into<Self::Integer>).into().into()
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^ `Into` cannot be made into an object
```

martinthomson:
Ah, and I see that putting `Into<u128>` on Field doesn't work for other reasons.  Oh well, this works, I guess.

PR #71, opened 2022-08-10
Fix the three_party_random test properly

martinthomson:
Some clown forgot to feed a fresh index into the second call to
random(), so we were getting a collision with 1/31 probability.

Fix that.


akoshelev:
Thanks for fixing it!

akoshelev (approved):

PR #74, opened 2022-08-11
Use the MMO^pi function for PRSS

martinthomson:
I'm not entirely clear on the relative benefits of using this over the
fixed-key model of AES - particularly as we are reducing the `u128`
modulo $p << 2^{128}$ - but this seems like a relatively cheap way to
build the function.  This function provides correlation robustness,
whereas the fixed-key mode used previously does not.

Paper: https://eprint.iacr.org/2019/074
Function: $\mathsf{MMO}^{\pi}$

### DIFF
--- a/src/prss.rs
+++ b/src/prss.rs
@@ -179,13 +179,14 @@ pub struct Generator {
 
 impl Generator {
     /// Generate the value at the given index.
+    /// This uses the MMO^{\pi} function described in <https://eprint.iacr.org/2019/074>.
     #[must_use]
     pub fn generate(&self, index: u128) -> u128 {
         let mut buf = [0_u8; 16];
         LittleEndian::write_u128(&mut buf, index);
         self.cipher
             .encrypt_block(GenericArray::from_mut_slice(&mut buf));
-        LittleEndian::read_u128(&buf)
+        LittleEndian::read_u128(&buf) ^ index
     }
 }
 


src/prss.rs
```diff
@@ -179,13 +179,14 @@ pub struct Generator {
...
     /// Generate the value at the given index.
+    /// This uses the MMO^{\pi} function described in <https://eprint.iacr.org/2019/074>.
     #[must_use]
     pub fn generate(&self, index: u128) -> u128 {
         let mut buf = [0_u8; 16];
         LittleEndian::write_u128(&mut buf, index);
         self.cipher
             .encrypt_block(GenericArray::from_mut_slice(&mut buf));
-        LittleEndian::read_u128(&buf)
+        LittleEndian::read_u128(&buf) ^ index
```
akoshelev, line 189: Paper claims that $H(x, i) = π(x) ⊕ x$ does not have correlation robustness while $π(x)= π(x) ⊕ x$ does. What is `i` in our context here?

Also I wonder if our index values would be monotonically increasing over time from very small numbers, so it could be easy to see the bits that got flipped by just looking at bits set to 1 in `index`? Curious if it has any effect on the correlation robustness property

martinthomson: So the $H(x,i)$ version of this is one where there is a two-dimensional process (as used in OT extension) with two inputs.  As the $i$ input is not used, there is no correlation robustness in that construction.  However, we're looking at a one-dimensional input here, for which the construction is sufficient (or at least that is my understanding, limited as it is).

martinthomson: The PRP definition should ensure that any change, no matter how small, propagates throughout the entire output space.  The value of the inputs should be effectively hidden.  (Otherwise, all of your traffic is busted.)

akoshelev: yea I see how AES output can be a PRP, just it is not clear how XORIng with input makes it more secure especially with index not being a random number. feels like i can just XOR it again with index to get the same output as previous implementation produced

akoshelev (approved):

src/prss.rs
martinthomson: I don't really understand how this works, especially the $\hat{\mathsf{MMO}}^\pi$ variant, which does an rotate and xor operation as well, but this is probably OK for now.

PR #76, opened 2022-08-12
[WIP] [Blocked] Secure Sort implementation 

richajaindce:
[WIP] 
Implementing Radix Sort using the approach described in https://eprint.iacr.org/2019/695.pdf

image

This diff implements finding order of each input in the sorted list. 

For now this diff is blocked on executing multiple secure multiplication in parallel. I am parking this PR for using later once those changes are made. The test in this PR fails due to parallelization not handled

### DIFF
--- a/src/find_sort_location.rs
+++ b/src/find_sort_location.rs
@@ -1,0 +1,142 @@
+#![allow(dead_code)]
+
+use futures::future::join_all;
+
+use crate::{
+    field::Field,
+    helpers::ring::Ring,
+    replicated_secret_sharing::ReplicatedSecretSharing,
+    securemul::{ProtocolContext, SecureMul},
+};
+
+pub struct SecureSort<'a, R, F> {
+    context: &'a ProtocolContext<'a, R>,
+    share_of_one: ReplicatedSecretSharing<F>,
+}
+
+impl<F: Field, R: Ring> SecureSort<'_, R, F> {
+    pub async fn execute(
+        self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<ReplicatedSecretSharing<F>> {
+        let gen_permutation_share = |mult_output: Vec<ReplicatedSecretSharing<F>>| {
+            let len = mult_output.len() / 2;
+
+            // Fold the result
+            let mut permutation_share = Vec::new();
+            (0..len).for_each(|j| {
+                permutation_share.push(mult_output[j] + mult_output[j + len]);
+            });
+            permutation_share
+        };
+        gen_permutation_share(self.secure_multiply(self.prepare_mult_inputs(shares)).await)
+    }
+
+    fn prepare_mult_inputs(
+        &self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)> {
+        // Step1: Calculate x and 1 - x
+        let mut x_and_one_minus_x: Vec<ReplicatedSecretSharing<F>> =
+            shares.iter().map(|x| self.share_of_one - *x).collect();
+        x_and_one_minus_x.append(&mut shares.to_owned());
+
+        // Step2: Calculate cumulative sum
+        let length = x_and_one_minus_x.len();
+        let mut cumulative_sum = Vec::with_capacity(length);
+        cumulative_sum.push(x_and_one_minus_x[0]);
+        for i in 1..length {
+            cumulative_sum.push(cumulative_sum[i - 1] + x_and_one_minus_x[i]);
+        }
+
+        // Step3: Consolidate step 1 and step 2 output
+        x_and_one_minus_x
+            .iter()
+            .zip(cumulative_sum.iter())
+            .map(|(x_and_one_minus, cumulative_sum)| (*x_and_one_minus, *cumulative_sum))
+            .collect()
+    }
+
+    async fn secure_multiply(
+        &self,
+        mult_inputs: Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)>,
+    ) -> Vec<ReplicatedSecretSharing<F>> {
+        // Spawn all multiplication, wait for them to finish in parallel and then collect the results
+        let async_multiply = mult_inputs.iter().enumerate().map(|(index, input)| {
+            let output = SecureMul {
+                a_share: input.0,
+                b_share: input.1,
+                index: index as u128,
+            };
+            output.execute(self.context)
+        });
+        join_all(
+            async_multiply
+                .into_iter()
+                .map(|handle| async { handle.await.unwrap() }),
+        )
+        .await
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        field::Fp31,
+        helpers::{self, ring::mock::TestHelper},
+        replicated_secret_sharing::ReplicatedSecretSharing,
+    };
+
+    use super::SecureSort;
+
+    #[tokio::test]
+    async fn find_sorted_position() {
+        let ring = helpers::ring::mock::make_three();
+        let participants = crate::prss::test::make_three();
+        let context = crate::securemul::tests::make_context(&ring, &participants);
+        let mut secure_sort_output = Vec::new();
+
+        let input = [
+            (1, 0, 0),
+            (0, 0, 0),
+            (0, 1, 0),
+            (0, 0, 0),
+            (0, 0, 0),
+            (0, 0, 0),
+        ]
+        .to_vec();
+        let mut shares = [Vec::new(), Vec::new(), Vec::new()];
+
+        for iter in input {
+            let replicated_ss =
+                crate::replicated_secret_sharing::tests::secret_share(iter.0, iter.1, iter.2);
+            shares[0].push(replicated_ss.0);
+            shares[1].push(replicated_ss.1);
+            shares[2].push(replicated_ss.2);
+        }
+
+        for i in 0..3 {
+            let secure_sort = SecureSort::<TestHelper, Fp31> {
+                context: &context[i],
+                share_of_one: match i {
+                    0 => ReplicatedSecretSharing::new(Fp31::from(1_u8), Fp31::from(0_u8)),
+                    1 => ReplicatedSecretSharing::new(Fp31::from(0_u8), Fp31::from(0_u8)),
+                    _ => ReplicatedSecretSharing::new(Fp31::from(0_u8), Fp31::from(1_u8)),
+                },
+            };
+            secure_sort_output.push(secure_sort.execute(&shares[i]).await);
+        }
+
+        let expected_sort_output = [5_u128, 1, 6, 2, 3, 4].as_ref();
+
+        (0..secure_sort_output[0].len()).for_each(|i| {
+            assert_eq!(
+                secure_sort_output[0][i] + secure_sort_output[1][i] + secure_sort_output[2][i],
+                ReplicatedSecretSharing::new(
+                    Fp31::from(expected_sort_output[i]),
+                    Fp31::from(expected_sort_output[i])
+                )
+            );
+        });
+    }
+}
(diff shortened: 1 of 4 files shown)


src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+            permutation_share
+        };
+        gen_permutation_share(self.secure_multiply(self.prepare_mult_inputs(shares)).await)
+    }
+
+    fn prepare_mult_inputs(
+        &self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)> {
+        // Step1: Calculate x and 1 - x
```
thurstonsand, line 39: there's a different way you can do this function, using `scan`:

```rust
    shares
        .iter()
        .map(|x| self.share_of_one - *x)
        .chain(shares.iter())
        .scan(
            ReplicatedSecretSharing::<F>::new(F::ZERO, F::ZERO),
            |sum, n| {
                *sum += n;
                Some((n, *sum))
            },
        )
        .collect()
```

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+            let output = SecureMul {
+                a_share: input.0,
+                b_share: input.1,
+                index: index as u128,
+            };
+            output.execute(self.context)
+        });
+        join_all(
+            async_multiply
+                .into_iter()
```
thurstonsand, line 75: im curious -- what happens if you don't have this `into_iter`?

thurstonsand (changes requested):

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+    };
+
+    use super::SecureSort;
+
+    #[tokio::test]
+    async fn find_sorted_position() {
+        let ring = helpers::ring::mock::make_three();
+        let participants = crate::prss::test::make_three();
+        let context = crate::securemul::tests::make_context(&ring, &participants);
+        let mut secure_sort_output = Vec::new();
```
thurstonsand, line 97:
```suggestion
        let mut secure_sort_output = Vec::with_capacity(input.len());
```

tho this would require moving `let input` above this line

thurstonsand (changes requested):

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub async fn execute(
+        self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<ReplicatedSecretSharing<F>> {
+        let gen_permutation_share = |mult_output: Vec<ReplicatedSecretSharing<F>>| {
+            let len = mult_output.len() / 2;
+
+            // Fold the result
+            let mut permutation_share = Vec::new();
+            (0..len).for_each(|j| {
```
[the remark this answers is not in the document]
richajaindce: Thanks for the suggestion!!

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+    context: &'a ProtocolContext<'a, R>,
+    share_of_one: ReplicatedSecretSharing<F>,
+}
+
+impl<F: Field, R: Ring> SecureSort<'_, R, F> {
+    pub async fn execute(
+        self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<ReplicatedSecretSharing<F>> {
+        let gen_permutation_share = |mult_output: Vec<ReplicatedSecretSharing<F>>| {
```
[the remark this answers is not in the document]
richajaindce: I had done that more for the sake of readability.

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+        gen_permutation_share(self.secure_multiply(self.prepare_mult_inputs(shares)).await)
+    }
+
+    fn prepare_mult_inputs(
+        &self,
+        shares: &[ReplicatedSecretSharing<F>],
+    ) -> Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)> {
+        // Step1: Calculate x and 1 - x
+        let mut x_and_one_minus_x: Vec<ReplicatedSecretSharing<F>> =
+            shares.iter().map(|x| self.share_of_one - *x).collect();
```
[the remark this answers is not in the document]
richajaindce: Does this provide any benefits since the compiler auto detects the type?

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+        mult_inputs: Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)>,
+    ) -> Vec<ReplicatedSecretSharing<F>> {
+        // Spawn all multiplication, wait for them to finish in parallel and then collect the results
+        let async_multiply = mult_inputs.iter().enumerate().map(|(index, input)| {
+            let output = SecureMul {
+                a_share: input.0,
+                b_share: input.1,
+                index: index as u128,
+            };
+            output.execute(self.context)
```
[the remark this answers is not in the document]
richajaindce: We want to spawn all multiplications in parallel and ideally wait for all of them to finish in the end. Will change this to try_join_all

src/find_sort_location.rs
akoshelev, line 22: the final addition operation can be done in-place with no extra allocation by truncating the input vector after addition

akoshelev, line 22: do you think it is worth adding `debug_assert` to make sure `mult_output` length is even? seems that `gen_permutation_share` will be broken if that invariant does not hold

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+        // Step2: Calculate cumulative sum
+        let length = x_and_one_minus_x.len();
+        let mut cumulative_sum = Vec::with_capacity(length);
+        cumulative_sum.push(x_and_one_minus_x[0]);
+        for i in 1..length {
+            cumulative_sum.push(cumulative_sum[i - 1] + x_and_one_minus_x[i]);
+        }
+
+        // Step3: Consolidate step 1 and step 2 output
+        x_and_one_minus_x
```
akoshelev, line 53: nit - you can consume `x_and_minus_x` at this moment so using `.into_iter()` will save an extra copy here

However, given that `x_and_one_minus_x.len() %2 == 0` you can also avoid an extra vector allocation here by chaining iterator calls


```rust
 fn prepare_mult_inputs(
        &self,
        shares: &[ReplicatedSecretSharing<F>],
    ) -> Vec<(ReplicatedSecretSharing<F>, ReplicatedSecretSharing<F>)> {
        // Step1: Calculate x and 1 - x
        let mut x_and_one_minus_x: Vec<ReplicatedSecretSharing<F>> =
            shares.iter().map(|x| self.share_of_one - *x).collect();
        x_and_one_minus_x.extend_from_slice(shares);

        x_and_one_minus_x
            // Step2: Calculate cumulative sum
            .windows(2)
            // SAFETY: input size is at least 2 so it is safe to assume slice always has 2 elements
            .map(|slice| slice[0] + slice[1])
            // Step3: Consolidate step 1 and step 2 output
            .zip(x_and_one_minus_x.iter())
            .map(|(cumulative_sum, x_and_one_minus)| (*x_and_one_minus, cumulative_sum))
            .collect()
    }

```

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+        join_all(
+            async_multiply
+                .into_iter()
+                .map(|handle| async { handle.await.unwrap() }),
+        )
+        .await
+    }
+}
+
+#[cfg(test)]
```
akoshelev, line 82: would be great to test this algorithm on 0 sized input, 1 element input etc.

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+
+pub struct SecureSort<'a, R, F> {
+    context: &'a ProtocolContext<'a, R>,
+    share_of_one: ReplicatedSecretSharing<F>,
+}
+
+impl<F: Field, R: Ring> SecureSort<'_, R, F> {
+    pub async fn execute(
+        self,
+        shares: &[ReplicatedSecretSharing<F>],
```
akoshelev, line 20: what is our strategy for dealing with associated data here? I assume we would have it as we care about sort stability

src/find_sort_location.rs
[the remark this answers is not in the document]
thurstonsand: i was only talking about readability of the code. they are functionally the same

src/find_sort_location.rs
thurstonsand: what do you think of using `scan` to do this? https://github.com/martinthomson/raw-ipa/pull/76#discussion_r944655032

akoshelev: that code is strictly better. we should use it instead

src/find_sort_location.rs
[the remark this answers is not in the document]
akoshelev: curious about what makes one variant more readable than another.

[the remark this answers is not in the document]
thurstonsand: it's pretty minor, but when i see the `collect` fn, it's easier to just see the type right there with it, vs hunting for the assignment and seeing the type there

src/find_sort_location.rs
martinthomson: It's not perfect.  You don't need to extend_from_slice if you are iterating, you can use Iterator::chain() instead.  Also, don't collect x_and_one_minus_x, just feed it into zip().

src/find_sort_location.rs
```diff
@@ -0,0 +1,142 @@
...
+                crate::replicated_secret_sharing::tests::secret_share(iter.0, iter.1, iter.2);
+            shares[0].push(replicated_ss.0);
+            shares[1].push(replicated_ss.1);
+            shares[2].push(replicated_ss.2);
+        }
+
+        for i in 0..3 {
+            let secure_sort = SecureSort::<TestHelper, Fp31> {
+                context: &context[i],
+                share_of_one: match i {
```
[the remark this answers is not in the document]
martinthomson: I think that for optimizations like this, we probably want a function to do this.

Also, I would like to see this use the new HelperRole enum, so that there is no `_` arm on the match.

src/securemul.rs
```diff
@@ -12,9 +12,9 @@ use thiserror::Error;
 /// K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, and B. Pinkas. High-throughput secure AES computation. In WAHC@CCS 2018, pp. 13–24, 2018
 #[derive(Debug)]
 pub struct SecureMul<F> {
```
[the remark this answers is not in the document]
martinthomson: Yeah, make a new() function.  This makes fields mutable to anyone outside of this module (if they hold a mutable reference, etc...).

src/securemul.rs
```diff
@@ -283,7 +283,7 @@ mod tests {
         Ok(validate_and_reconstruct(result_shares).into())
     }
 
-    fn make_context<'a>(
+    pub(crate) fn make_context<'a>(
```
martinthomson, line 286: I think that it's OK to make this an unconditional pub with `pub(crate)` on the test module.

src/find_sort_location.rs
thurstonsand: @martinthomson what about my alternative? https://github.com/martinthomson/raw-ipa/pull/76#discussion_r944655032

martinthomson: Yeah, that is better.  Still not sure about the use of .collect() on the end of that chain though.  I guess that this is a decision imposed by the API to the function though.

src/find_sort_location.rs
richajaindce: This is really cool! Thanks for the suggestion!

src/find_sort_location.rs
richajaindce: We dont need to worry about the associated data within the sort protocol since the protocol will spit out a permutation which will be applied to the associated data.

richajaindce:
This is incorporated in #90 . Closing this one

PR #78, opened 2022-08-15
Add Helper Identity enum

akoshelev:
There seem to be more and more protocols we need to implement that require helpers to know their exact position inside the MPC ring. For instance, modulus conversion needs to construct binary secret shares differently depending on which helper performs this operation (#63).

Secure sort (#66) seem to need that as well as it requires helpers to agree on constructing a secret share of `1`.

This change adds necessary plumbing to make helper identity available for protocol writers - `ctx.helper_ring.identity()` should return the exact helper location

### DIFF
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -17,7 +17,7 @@ pub use event::{
 use crate::error::{Error, Res};
 use crate::threshold::EncryptionKey as ThresholdEncryptionKey;
 #[cfg(feature = "enable-serde")]
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Display, Error as FmtError, Formatter};
 #[cfg(feature = "enable-serde")]
 use std::fs;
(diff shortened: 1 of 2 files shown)


src/helpers/mod.rs
```diff
@@ -27,6 +27,15 @@ use std::ops::Index;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
+/// Represents a unique identity of each helper running MPC computation.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+pub enum Identity {
```
martinthomson, line 33: Maybe for later, but it would be good to have functions either like this:

```rust
fn left(self) -> Self { ... }
fn right(self) -> Self { ... }
```
or like this:

```rust
fn peer(self, dir: Direction) -> Self { ... }
```

where:

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
enum Direction { Left, Right }
```

martinthomson (approved):

src/helpers/mod.rs
thurstonsand: might be good to add the `left()`/`right()` at same place as `identity()`, on `Ring`

akoshelev: hmm, I kinda like this interface now

```rust
   async fn send<T>(peer: Identity, msg: T)
```

and calling it from multiplication like this

```rust
   ring.send(ctx.identity.peer(Direction::Right), d_value)
```

PR #79, opened 2022-08-18
MPC network component

thurstonsand:
This PR covers how the network component of MPC is done. Namely, how the MPC helpers communicate with each other. This will happen exclusively via multiplication, so that is the piece set to be implemented.


src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+use std::fmt::Formatter;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Step {
+    Example(ExampleCircuit),
+    ModulusConversion(ModulusConversionDetails),
+}
+
+impl Step {
+    pub fn to_path(&self) -> String {
```
thurstonsand, line 16: @akoshelev this is pretty cumbersome. got any better ideas of how to serialize `Step`s?

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Step {
+    Example(ExampleCircuit),
+    ModulusConversion(ModulusConversionDetails),
+}
+
+impl Step {
+    #[must_use]
+    pub fn to_path(&self) -> String {
```
thurstonsand, line 17: @akoshelev this is pretty cumbersome. got any better ideas of how to serialize/deserialize `Step`s?

src/net/server/mod.rs
```diff
@@ -33,7 +37,12 @@ impl IntoResponse for MpcServerError {
 /// Axum router definition for MPC helper endpoint
 #[must_use]
 pub fn router() -> Router {
-    Router::new().route("/echo", get(handlers::echo_handler))
+    Router::new()
+        .route("/echo", get(handlers::echo_handler))
+        .route(
+            "/mul/query_id/:query_id/step/*step",
```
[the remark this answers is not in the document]
akoshelev: bikeshedding continues 

```suggestion
            "/mul/steps/:step?query_id=:query_id",
```

mostly I feel like it better adheres to the api design principles proposed here, curious about @martinthomson thoughts on this

https://cloud.google.com/apis/design/resource_names

src/helpers/ring/mod.rs
```diff
@@ -0,0 +1,43 @@
+//! This module contains implementations and traits that enable MPC helpers to communicate with
```
akoshelev, line 1: lets rebase this PR on top of `main` to exclude things that don't belong here

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
+use crate::framework::Error;
+#[cfg(feature = "enable-serde")]
+use serde::{
+    de::{Error as SerdeError, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+use std::fmt::Formatter;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Step {
```
akoshelev, line 10: I feel that I'd like to see a top level structure like this

```rust
pub enum Step {
    Test(TestStep),
     IPA(IPAStep)
}

pub enum IPAStep {
   ConvertShares(ModulusConversion),
   Sort(...),
   Attribute(...),
   ApplyDifferentialPrivacy(...)
}
```

thoughts?

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+            "mul2" => Ok(ExampleCircuit::Mul2()),
+            _ => Err(Error::StepParseError),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ModulusConversionDetails {
+    Mul1 { bit_num: u32 },
+    Mul2 { bit_num: u32 },
```
akoshelev, line 64:
```suggestion
    Mul1 { bit_num: u8 },
    Mul2 { bit_num: u8 },
```

akoshelev: or even

```rust
   BitMul1(u8),
   BitMul2(u8)
```

we would need some documentation here anyway to explain what these operations are

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+impl Step {
+    #[must_use]
+    pub fn to_path(&self) -> String {
+        match self {
+            Self::Example(circuit) => format!("example/{}", circuit.to_path()),
+            Self::ModulusConversion(details) => format!("modulus_conversion/{}", details.to_path()),
+        }
+    }
+
+    pub fn from_path(path_str: &str) -> Result<Step, Error> {
```
akoshelev, line 24: that is probably belong to `TryFrom<&str> for Step` implementation.

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        let path_str = path_str.strip_prefix('/').unwrap_or(path_str);
+        let (step, next) = path_str.split_once('/').ok_or(Error::StepParseError)?;
+        match step {
+            "example" => Ok(Self::Example(ExampleCircuit::from_path(next)?)),
```
akoshelev, line 28: I like this traversal pattern here

src/framework/step/mod.rs
akoshelev, line 17: I wonder why it is not enough to just implement `Serialize, Deserialize` on `Step` and force the same for enclosed structs/enums?
https://docs.rs/axum/0.5.15/axum/extract/struct.Path.html

src/framework/mod.rs
```diff
@@ -0,0 +1,24 @@
+use crate::net::MpcServerError;
+use axum::response::{IntoResponse, Response};
+use thiserror::Error;
+pub mod step;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("invalid step")]
```
akoshelev, line 8: it is a good practice to include the invalid value into error description (unless it is a secret of course)

src/framework/step/mod.rs
martinthomson: I don't think that we want that.  What we want is an implementation of a trait that can take an arbitrary input and produce a path (component) that we can use to route.  What @thurstonsand has is entirely fixed, which is a fine starting point, but we probably need to parameterize.

Then we would have test code with 

```rust
enum TestStep {
  Step1,
  Whatever { something more, or something, }
}
impl Step for TestStep { 
  fn to_path(&self) -> ...
```

And IPA, with its own structure, much like how Alex did it.

Those can coexist on the server, but on different base paths.

src/net/server/mod.rs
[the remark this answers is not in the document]
martinthomson: I prefer what Thurston did.

It's totally OK to create a resource to represent the query as a whole.  Or, in this case, a collection of resources that represent the internal pieces of the query.  Having them in a hierarchical structure makes more sense than using query parameters.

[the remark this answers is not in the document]
thurstonsand: at any rate, i can switch to hyphens over underscores

src/helpers/ring/mod.rs
thurstonsand: actually, this is latest with main. I moved some stuff around.

`ring.rs` used to have all the code in one file, including the mock module. i brought that into a folder and separated things out, so i can add an `http` implementation

src/framework/step/mod.rs
thurstonsand: i can create a trait with `to_path` and `from_path` and have all steps implement that

src/framework/step/mod.rs
thurstonsand: agreed. this isn't actually implemented yet, and honestly might be better to give these more descriptive names as well

src/framework/step/mod.rs
thurstonsand: my only concern with doing this is that it's not particularly descriptive. obviously, we are trying to parse a string into a `Step`, but it's not just "any" string, it's specifically a url path string. what are your thoughts?

src/framework/step/mod.rs
thurstonsand: let's talk about that. the implementation of `Serialize`/`Deserialize` here was a lot of boilerplate, and i wanted to only have to do that once (and rely on string parsing after that).

we could do `serialize_with` and `deserialize_with` to get around that tho. but even that is more boilerplate than what i've got here

src/framework/mod.rs
thurstonsand: i agree. i'll think about how best to represent that

src/net/server/mod.rs
```diff
@@ -33,7 +37,12 @@ impl IntoResponse for MpcServerError {
...
+            post(handlers::mul_handler),
+        )
```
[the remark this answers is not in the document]
thurstonsand: `:step` matches a single segment, while `*step` matches all remaining segments

src/framework/step/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+            )),
+            _ => Err(Error::StepParseError),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ExampleCircuit {
+    Mul1(),
+    Mul2(),
```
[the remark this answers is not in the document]
thurstonsand: yes, i was thinking of doing something more with the example, but ended up not.

src/helpers/ring/mod.rs
akoshelev: gotcha. It is totally fine to do but generally I'd recommend to defer moving things around so we can just focus on reviewing new code. Once we finalize and agree it, we can submit another PR that better organizes the content - usually such changes do not require careful review compared to functional PRs

src/net/server/mod.rs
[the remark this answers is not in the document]
akoshelev: yea it did not seem obvious to me that `mul` has a collection of queries, but I don't have a strong argument against it

src/framework/mod.rs
akoshelev: I think it could be enough to say this

```rust
pub enum Error {
    #[error("invalid step: {0}")]
    InvalidStep(String)
```

src/framework/step/mod.rs
martinthomson: We can think of this as just a string... That also happens to conform to the stair of a URI path component.

PR #80, opened 2022-08-22
Add `Step` trait

akoshelev:
## Overview

As discussed offline with @martinthomson and @thurstonsand, we need to model protocol step as an enum. 

Effectively, in the long term we will have a composite id that uniquely identifies every multiplication happening in the system that consists of:

* Query id (not in this PR). Every query has its own circuit created for their own purposes. 
* Record id (not in this PR) - monotonically increasing integer that uniquely identifies every record (secret share) in the input. 
* Step (this PR). Unique transformation on a given record. Effectively used to uniquely identify multiply operation as it is the only one currently that requires communication. 

Steps form a tree where only leaf nodes are used in the communication and intermediate nodes are classifiers.

A combination of these three identifiers allows each helper to correctly interpret every value sent towards it and therefore let us create steps/protocols that require some of the underlying protocol implementations to run in parallel. 

Right now our mental model about communication includes every helper to create a new request per query + step

```
https://<helper_endpoint>/api/query/<query_id>/<step>

[values_stream]
```

record id is the index in the values array. In this case helpers will be able to understand each value that is sent towards them by looking at request path that includes `query` and `step` and at the index of the item inside the value stream/array. 

## Ongoing work

This PR duplicates some work in #79  to unblock @benjaminsavage and @richajaindce efforts to build modulus conversion and sorting protocols as they require multiplications to happen concurrently. In particular it defines a `Step` trait that is also used or will be used in #79. 

This PR also conflicts with #66 so if we decide to push this version, that change will be dropped


src/securemul.rs
```diff
@@ -241,8 +260,63 @@ mod tests {
...
+
+        let mut multiplications = Vec::new();
+        for i in 1..10_u128 {
+            // there is something weird going on the compiler's side. I don't see why we need
+            // to use async move as `i` is Copy + Clone, but compiler complains about it not living
+            // long enough
+            let ctx = &context;
+            let f = async move {
+                let h1_future = SecureMul {
+                    step: TestStep::Mul1(i as u8),
```
martinthomson, line 285: Please don't use the index as an argument to the step.  That will mean that you have separation at the step AND index layer, but based on the same value.  It's not redundant, but aligned, which isn't what we are looking for.

What you have here is a case where there are three helpers acting, so each should just use TestStep::Mul(0) or some other fixed step identifier.  They are each operating on the same data, so they will have the same index too.


I don't know if this test will pass if you make this change, because I'm not seeing anywhere that guarantees that the test infrastructure maintains order based on index.  What you have done might be hiding that as a problem.

src/securemul.rs
```diff
@@ -241,8 +260,63 @@ mod tests {
...
+                    index: i,
+                    a_share: a[1],
+                    b_share: b[1],
+                }
+                .execute(&ctx[1]);
+                let h3_future = SecureMul {
+                    step: TestStep::Mul1(i as u8),
+                    index: i,
+                    a_share: a[2],
+                    b_share: b[2],
```
martinthomson, line 302: I thought that we decided to add a SecureMul::new() method for this?  I note that you are placing arguments in a different order in this test and the other.  FWIW, I think that this ordering makes sense.

src/helpers/ring.rs
```diff
@@ -127,18 +130,18 @@ pub mod mock {
...
+        fn set_right(&mut self, right: Sender<MessageEnvelope<S>>) {
             self.right = Some(right);
         }
     }
 
     #[async_trait]
-    impl Ring for TestHelper {
-        async fn send<T: Message>(&self, dest: HelperAddr, msg: T) -> Result<(), Error> {
+    impl<S: Step> Ring<S> for TestHelper<S> {
+        async fn send<T: Message>(&self, dest: HelperAddr, step: S, msg: T) -> Result<(), Error> {
```
martinthomson, line 144: IMPORTANT: I don't see anything here that ensures that sending occurs in order.

martinthomson (commented):
I like the API changes.  I think we need to address the ordering issue though.

src/helpers/ring.rs
akoshelev: I thought the idea was to let actual implementation of this trait to reorder events as they arrive according to their position (query + record + step). Is there a way to impose ordering on caller side without creating a custom scheduler?

src/securemul.rs
akoshelev: I agree that it is better to semantically separate index and step. However, we seem to require "dynamic" steps (where step is essentially composed of Step name + index, for example in share conversion we use `Mul(bit_id)` . That is what I tried to reproduce here to make sure we would be able to do it with this change.

So I am trying to understand whether just logically representing share "bit" `i` and `index` as independent counters would suffice or there is a bigger issue that I don't see

src/securemul.rs
akoshelev: fair point, I was trying to make a change with as few changed lines as possible to make review easier, but I could fix this too.

src/helpers/ring.rs
martinthomson: This is the implementation, so this is going to need to deal with that somehow.

What we discussed was a unique stream of items for each (query + step), but that each item would have a counter (or index) that would place it in sequence on that stream.

src/securemul.rs
martinthomson: Yeah, I think that what you want is a bit number and a record number.  You pass the bit number to `TestStep::Mul1(bit_number)` and the record number (or index) goes in `index`.

src/helpers/ring.rs
akoshelev: gotcha. I thought you had questions about `send` interface. 

I was hoping to push it separately to unblock folks faster, but I can do it as part of this change too

martinthomson: Your call.  Though I think that your tests were hiding any potential problems that might come about from not having implemented it, so I wanted to ensure that people wouldn't use it correctly and then be surprised when they got garbage sometimes (it should be rare that things get reordered if you aren't working very hard, but it could happen).

richajaindce:
@akoshelev Trying the understand the terminology. As an example
Query id : Would this be unique per query made by a report collector i.e. all events will have same query ID?
Record id : Is this referring to a single event (source/trigger) in the query? If so, do we expect record ID to be maintained all across the protocol (i.e. e.g. even after shuffle, record ids will point to the initial input?) If not, what is the purpose of this field? 
Step : We would have communication for other protocols e.g. reshare, shuffle, reveal, etc. Would they all also have a "Step". As a whole, what does a step represent? is it a mini protocol (like multiply) or a unit step which any mini protocol would execute on a single helper or something else?

src/securemul.rs
```diff
@@ -76,7 +79,7 @@ impl<F: Field> SecureMul<F> {
         let DValue {
             d: left_d,
             index: left_index,
-        } = ctx.helper_ring.receive(HelperAddr::Left).await?;
+        } = ctx.helper_ring.receive(HelperAddr::Left, self.step).await?;
```
richajaindce, line 82: Do we expect the records to always come in order for this to work or, we plan to handle ordering in the Ring communication itself? (referring to IndexMismatch in the code on line 92 below)

martinthomson: My expectation is that there will be a total ordering of items that are sent out that the communication infrastructure only needs to preserve.  You will determine (by setting a record number) what the order should be.  The changes we're talking about here should remove the IndexMismatch error ultimately.

akoshelev:
Will close it in favor of #82 if approved

akoshelev:
Closing as #82 replaces it

PR #81, opened 2022-08-23
Cleanup, removing old files...

benjaminsavage:
that are not up to date with our newest version of the IPA proposal.


akoshelev (approved):
Thanks for cleaning this up!

martinthomson (approved):
🔥

benjaminsavage:
I think I might have missed some more.

For example, 
src/helpers/aggregation.rs

Is that still in use?

PR #82, opened 2022-08-26
Overhaul of TestHelper and SecureMul 

akoshelev:
## Overview

Following up the offline conversation with @martinthomson, the test
framework, secure multiplication have been completely rewritten to
follow the new network layer API. See the main motivation in https://github.com/martinthomson/raw-ipa/pull/80,
this PR tries to finalize the shape of `QueryId`, `RecordId` and
`Step` as well as their use at the borderline between protocols and
infra. All concepts presented https://github.com/martinthomson/raw-ipa/pull/80 largely remain the same.

## List of changes compared to #80 

* Test helper network now uses bi-directional channels (implemented as two physical channels) per **query id** and **step** with a handshake protocol to establish those connections
* QueryId is just a unit struct for the moment as we don't have a use case for it.
* Multiplication is now available via `context` and can be requested by providing record id and step. The reason for it is to be able to implement backpressure later
* `HelperAddr` is dropped in favor of `Identity` 
*  Streaming multiply test has been moved to  the `stream` mod


src/helpers/mock.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        loop {
+            // Depending on connection status, request a new connection, spin and wait for
+            // connection acknowledgment from peer or return the sender end of connection
+            // if it is ready
+            let control_message = {
+                let mut connections = self.connections.lock().unwrap();
+                let conn_state = connections
+                    .entry((peer, step))
+                    .or_insert_with(|| ConnectionState::Listen);
```
thurstonsand, line 206:
```suggestion
                    .or_insert(ConnectionState::Listen);
```

src/helpers/mesh.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Receive a message that is associated with the given record id.
+    async fn receive<T: Message>(&mut self, source: Identity, record: RecordId)
+        -> Result<T, Error>;
+
+    /// Returns the unique identity of this helper.
+    fn identity(&self) -> Identity;
+}
+
+/// This is the entry point for protocols to request communication when they require it.
+pub trait Gateway<M: Mesh, S: Step> {
```
thurstonsand, line 44: so to understand the workflow here --

protocol writers will have a long-lived `Gateway` they hold onto, and for each `Step` they take, they use the `Gateway` to generate a `Mesh` with which to send/receive? and this `Mesh` will be valid for the duration of a `Step`?

src/helpers/mock.rs
```diff
@@ -0,0 +1,286 @@
...
+
+                        Some(ControlMessage::ConnectionRequest(self.identity, step, rx))
+                    }
+                    ConnectionState::Established(sender) => {
+                        return sender.clone();
+                    }
+                }
+            };
+
+            if let Some(msg) = control_message {
```
thurstonsand, line 221: is this not always a `Some(msg)`?

src/helpers/mock.rs
```diff
@@ -0,0 +1,286 @@
...
+        // spin and wait until message with the same record id appears in the buffer
+        // when it happens, pop it out, try to reinterpret its bytes as `T` and return
+        loop {
+            {
+                let mut buf = self.buf.lock().unwrap();
+                if let Some(msgs) = buf.get_mut(&(peer, step)) {
+                    let l = msgs.len();
+                    for i in 0..l {
+                        if msgs[i].record_id == record_id {
+                            msgs.swap(i, l - 1);
```
thurstonsand, line 243: this works, but seems like it'd be easier to understand as a `HashMap<RecordId, MessageEnvelope>`

src/helpers/mod.rs
```diff
@@ -1,11 +1,58 @@
...
+
+#[derive(Debug, Copy, Clone)]
+pub enum Direction {
+    Left,
+    Right,
+}
+
+impl Identity {
+    #[must_use]
+    pub fn all_variants() -> &'static [Identity; 3] {
```
thurstonsand, line 23: what's the difference between this and something like

```rust
const all_variants: &'static [Identity; 3] = &[Identity::H1, Identity::H2, Identity::H3];
```

src/lib.rs
```diff
@@ -1,12 +1,15 @@
 #![deny(clippy::clone_on_ref_ptr)]
 
+extern crate core;
```
thurstonsand, line 3: what does this line do?

src/helpers/mock.rs
akoshelev: yea, initially I thought that I should just have a linear buffer of objects but ended up implementing a lookup by `record_id` which is what hashmap is for. I'll probably keep it around as there are other ideas circulating about the internals so this method may be changed entirely

src/helpers/mock.rs
akoshelev: Does anybody know why this suggestion is not visible from the files tab??

thurstonsand: if you've made a change to that line, it'll only show up on commits before the change. so you'd have to pick an earlier commit in the files tab to see it

src/helpers/mod.rs
thurstonsand: oh you know i had no idea godbolt supported rust. cool!

src/helpers/mock.rs
thurstonsand: 👍

src/helpers/mod.rs
thurstonsand: you know what, there actually is a difference. there isn't a way to reference `const all_variants` without a reference to a particular instance, i.e. `self.all_variants`. while the `fn all_variants()` can be referenced via `Identity::all_variants()`

akoshelev: hmm, does the following cover this use case? 

```rust
impl Identity {
    pub const ALL_VARIANTS: &'static [Identity; 3] = &[Identity::H1, Identity::H2, Identity::H3];
   
    pub fn l(&self) -> usize {
        Self::ALL_VARIANTS.len()
    }
}

fn l() -> usize {
    Identity::ALL_VARIANTS.len()
}


```

martinthomson:
I realize that we really need to get this moving.

src/helpers/mod.rs
thurstonsand: oh, yeah. so it does work. my IDE was telling me that was invalid, but i didn't actually test it on compile. thanks CLion

PR #84, opened 2022-09-01
Remove spin and wait from receive

### LINKED ISSUE #62: Ring feedback

akoshelev:
## Overview

Previously `receive` was running a tight loop where on each iteration it checked whether the message is available or not. This change makes all the hard work happening inside one task that holds onto a local buffer of messages, so we don't need any synchronization there. The communication between the event loop and `receive` function is done via oneshot channels. 

We discussed this optimization [here](https://github.com/akoshelev/raw-ipa/commit/5f50e270eeb20e86db0687b1de61fb13b89c0815#r82460624) and this change should fix #62

### DIFF
--- a/src/helpers/mock.rs
+++ b/src/helpers/mock.rs
@@ -296,7 +296,7 @@ impl<S: Step> Controller<S> {
         if let Some(rx) = rx {
             self.peers
                 .get(&peer)
-                .unwrap_or_else(|| panic!("No peer with id {peer:?}"))
+                .expect("peer with id {peer:?} should exist")
                 .send(ControlMessage::ConnectionRequest(self.identity, step, rx))
                 .await
                 .unwrap();
--- a/src/helpers/mock.rs
+++ b/src/helpers/mock.rs
@@ -262,7 +262,7 @@ impl<S: Step> Controller<S> {
                         }
                     }
                     Some(receive_request) = receive_rx.recv() => {
-                        buf.entry(receive_request.channel())
+                        buf.entry(receive_request.channel_id())
                            .or_default()
                            .receive_request(receive_request.record_id, receive_request.sender);
                     }
--- a/src/helpers/mock.rs
+++ b/src/helpers/mock.rs
@@ -1,10 +1,10 @@
 ///! Provides an implementation of `Gateway` and `Mesh` suitable for unit tests.
-use crate::chunkscan::prepend;
 use crate::helpers::error::Error;
 use crate::helpers::mesh::{Gateway, Mesh, Message};
 use crate::helpers::Identity;
 use crate::protocol::{QueryId, RecordId, Step};
 use async_trait::async_trait;
+use futures::Stream;
 use futures_util::stream::SelectAll;
 use futures_util::StreamExt;
 use std::collections::hash_map::Entry;
@@ -347,4 +347,8 @@ fn make_controllers<S: Step>() -> [Controller<S>; 3] {
 
         Controller::launch(identity, peer_senders, rx)
     })
+}
+
+pub fn prepend<T: Copy + Clone, S: Stream>(id: T, stream: S) -> impl Stream<Item = (T, S::Item)> {
+    stream.map(move |item| (id, item))
 }
--- a/src/helpers/mock.rs
+++ b/src/helpers/mock.rs
@@ -1,14 +1,17 @@
-/// Provides an implementation of `Gateway` and `Mesh` suitable for unit tests.
-use std::collections::HashMap;
-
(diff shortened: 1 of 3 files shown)


src/helpers/mock.rs
```diff
@@ -136,117 +213,106 @@ impl<S: Step> Mesh for TestMesh<S> {
...
-                        return sender.clone();
-                    }
-                }
-            };
+        };
 
+        if let Some(rx) = rx {
             self.peers
                 .get(&peer)
                 .unwrap_or_else(|| panic!("No peer with id {peer:?}"))
```
martinthomson, line 299:
```suggestion
                .expect("peer with id {peer:?} should exist")
```

src/helpers/mock.rs
```diff
@@ -136,117 +213,106 @@ impl<S: Step> Mesh for TestMesh<S> {
...
         assert_ne!(self.identity, peer);
 
-        // start listening for incoming messages and move them from channel to the buffer
-        tokio::spawn({
-            let buf = Arc::clone(&self.buf);
-            async move {
-                while let Some(msg) = rx.recv().await {
-                    let mut buf = buf.lock().unwrap();
-                    buf.entry((peer, step)).or_insert_with(Vec::new).push(msg);
+        let (result, rx) = {
```
martinthomson, line 285:
```suggestion
        let (sender, rx) = {
```

src/helpers/mock.rs
```diff
@@ -136,117 +213,106 @@ impl<S: Step> Mesh for TestMesh<S> {
...
                 .unwrap_or_else(|| panic!("No peer with id {peer:?}"))
-                .send(control_message)
+                .send(ControlMessage::ConnectionRequest(self.identity, step, rx))
                 .await
                 .unwrap();
-
-            tokio::task::yield_now().await;
         }
+
+        result
```
martinthomson, line 305:
```suggestion
        sender
```

src/helpers/mock.rs
```diff
@@ -136,117 +213,106 @@ impl<S: Step> Mesh for TestMesh<S> {
...
+                        match control_message {
+                            ControlMessage::ConnectionRequest(peer, step, peer_connection) => {
+                                channels.push(prepend((peer, step), ReceiverStream::new(peer_connection)));
+                            }
                         }
                     }
+                    Some(receive_request) = receive_rx.recv() => {
+                        buf.entry(receive_request.channel())
+                           .or_default()
+                           .receive_request(receive_request.record_id, receive_request.sender);
```
martinthomson, line 267: Do you want to use ChannelId for this?  It seems like you could save some typing at least.

src/helpers/mock.rs
```diff
@@ -136,117 +213,106 @@ impl<S: Step> Mesh for TestMesh<S> {
...
+            loop {
+                // Make a random choice what to process next:
+                // * Receive and process a control message
+                // * Receive a message from another helper
+                // * Handle the request to receive a message from another helper
+                tokio::select! {
+                    Some(control_message) = control_rx.recv() => {
+                        match control_message {
+                            ControlMessage::ConnectionRequest(peer, step, peer_connection) => {
+                                channels.push(prepend((peer, step), ReceiverStream::new(peer_connection)));
```
martinthomson, line 260: Do you need to implement prepend in the other module?  It seems like a fairly specialized tweak, so maybe it belongs here (at least for now).

martinthomson (approved):
This is a distinct improvement; from here, I think we'll need a more rigorous testing regime before we make more changes.

src/helpers/mock.rs
akoshelev: did you mean using `ReceiveRequest` here?

PR #86, opened 2022-09-13
Refactoring code to move the code in appropriate folders

richajaindce:


richajaindce:
- Lets also move field.rs to some folder - any suggestions where?
- I wasn't able to move net/ due to compile error due to an error - DeadThread which should ideally sit inside ipa_bench @taikiy can you help on this

src/secret_sharing_schemes/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod replicated_secret_sharing;
```
akoshelev, line 1: I don't think we need to make the whole module visible, just re-exporting secret sharing structs is enough

src/protocol/securemul.rs
```diff
@@ -247,19 +226,19 @@ mod tests {
     use rand::Rng;
     use rand_core::RngCore;
 
-    use crate::replicated_secret_sharing::ReplicatedSecretSharing;
+    use crate::secret_sharing_schemes::replicated_secret_sharing::ReplicatedSecretSharing;
```
akoshelev, line 229: secret sharing is repeated 3 times here, maybe we can have `secret_sharing` module that exports `Replicated` and `Shamir`?

src/protocol/mod.rs
```diff
@@ -1,3 +1,6 @@
+pub mod context;
+pub mod securemul;
```
akoshelev, line 2: ditto: lets just export `SecureMul` struct so code can import it as `use crate::protocol::SecureMul`

akoshelev:
`chunkscan` module does not technically belong to `helpers` as it is just a generic stream ext, so maybe we can keep it in the root folder for now

akoshelev (commented):
Thanks for making this change @richajaindce. It's been bugging me for a while too and I am glad that you took the lead here

richajaindce:
> `chunkscan` module does not technically belong to `helpers` as it is just a generic stream ext, so maybe we can keep it in the root folder for now
I was struggling to put this as well as field.rs in a module, how about we move both to another folder named common?

akoshelev:
> > `chunkscan` module does not technically belong to `helpers` as it is just a generic stream ext, so maybe we can keep it in the root folder for now
> > I was struggling to put this as well as field.rs in a module, how about we move both to another folder named common? @akoshelev

I think it is fine if we cannot find a new home for every file in the root folder. there is a high chance that `chunkscan` will go away and be deleted very soon and `field` currently is just one implementation so imo it is fine to keep them where they are

Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 
 [features]
 default = ["debug", "cli"]
+bin = []
```
akoshelev, line 9: don't see this feature being used inside the code. am I missing it?

src/bin/ipa_bench/net/client/mod.rs
```diff
@@ -6,6 +6,7 @@ use thiserror::Error as ThisError;
 
 use super::Command;
 
+#[allow(dead_code)]
```
akoshelev, line 9: it is not ideal to sprinkle these suppressions everywhere, in my experience it usually happens when I forget to declare module public. is it the same here?

richajaindce: @taikiy would you intend to use this code in MpcHandle and MpcHttpConnection or is it safe to purge that code for now?

Cargo.toml
richajaindce: some of the constructs in ipa_bench folder i.e test_mpc, helper and ipa_bench are using it. CC: @taikiy

akoshelev: can you folks point me to one of those constructs? I don't seem to be able to find it

akoshelev (approved):
Thanks for cleaning this up!

PR #87, opened 2022-09-14
Update source/trigger events structure for synthetic data generation

taikiy:
## Overview
This diff updates source and trigger structs used for synthetic data generation. Synthetic data will be used to either create an attribution report benchmark (executed in clear), or transform the data into encrypted and secret shared format to be fed to IPA to generate reproducible test results.
Data structure is (almost) the same as the one specified in the [end-to-end doc](https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#gathering-events). Just not encrypted nor secret shared.

```
struct {
    int match_key_share;
    int timestamp_share;
    int is_trigger_report_share;
    int breakdown_key_share;
    int trigger_value_share;
} generic_report_share
```

## What have changed?
* Merged source and trigger events into `GenericReport`
* Report only contains 1 match key instead of multiple
* Added `attribution_constraint_id`
* Removed --secret-share option. If the test data needs to be fed into MPC, we can create another module which encrypts and secret shares the data (not part of this diff).

Since this tool's purpose is to generate human-readable test data, resulting output contains fields in clear and not secret shared.

### DIFF
--- a/src/bin/ipa_bench/gen_events.rs
+++ b/src/bin/ipa_bench/gen_events.rs
@@ -4,10 +4,6 @@ use log::{debug, info, trace};
 use rand::{CryptoRng, Rng, RngCore};
 use rand_distr::num_traits::ToPrimitive;
 use rand_distr::{Bernoulli, Distribution};
-use raw_ipa::helpers::models::{
-    Event as EEvent, SecretSharable, SecretShare, SourceEvent as ESourceEvent,
-    TriggerEvent as ETriggerEvent,
-};
 use serde::{Deserialize, Serialize};
 use std::io;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


src/bin/ipa_bench/gen_events.rs
```diff
@@ -34,99 +37,47 @@ pub struct EventBase {
     pub timestamp: Number,
 
     /// 0 (false) if this is a source event. 1 (true) otherwise.
-    pub is_trigger: bool,
-}
+    pub is_trigger_report: bool,
```
akoshelev, line 40: I wonder if source/trigger events can be better represented as enum

```
  enum Event {
     Source { timestamp, ... },
     Trigger { timestamp, value }
  }
```

code that serializes this enum should make sure both events have the same size on disk/network but at least in the code we will never be able to create a source event with a non-null (0) value

src/bin/ipa_bench/gen_events.rs
```diff
@@ -258,54 +176,44 @@ fn gen_events<R: RngCore + CryptoRng>(
...
-                matchkeys.clone(),
-                //TODO: Carry to next epoch if timestamp > DAYS_IN_EPOCH
-                params.epoch,
-                u32::try_from(t.as_secs()).unwrap(),
-                conversion_value,
-            )));
-        }
+        // days_in_epoch + sample.impression_time_diff < 4 years < u8::MAX (assuming 1 epoch = 1 week)
+        #[allow(clippy::cast_possible_truncation)]
+        let new_epoch = epoch + (t.as_secs() / SECONDS_IN_A_WEEK) as u8;
```
akoshelev, line 185: you may want to encapsulate the epoch into a struct that handles addition/subtraction and backed by a u8

thurstonsand: i strongly agree. especially for cases like below, so you can hide such things as `u32::try_from(...).unwrap()`

taikiy:
Thanks for reviewing, @akoshelev and @thurstonsand. I'm going to close this pull request and open a new one to resolve merge conflicts.

PR #88, opened 2022-09-15
Add benchmarks for arithmetic circuits

akoshelev:
This change adds both Criterion and iai benchmark frameworks as they have different properties that are useful for us.
It adds a simple benchmark to execute an arithmetic circuit with the given width and depth to stress the messaging layer and identify weak spots there. 

In order for code to be shared across these benchmarks, I moved basic building blocks for tests, such as `make_world`, `make_participants` to a shared module `test-fixture` only available for tests and benchmarks. 

## Benchmark results

Criterion (circuits vary from 5k to 64M multiplications):

IAI (500k multiplications):

```
iai_benchmark
  Instructions:         23079054098
  L1 Accesses:          33928001201
  L2 Accesses:            418313391
  RAM Accesses:            35847108
  Estimated Cycles:     37274216936
```

Flamegraph (64M multiplications):

flamegraph

Notes/observations: 

* it takes too long to run 64M multiplications as part of Criterion or IAI, so I think we should reduce the number of 
multiplications to the total of 1M for benchmarks.
* It scales linearly with number of multiplications. 
* From the flamegraphs it looks that the system operates at approximately 80% of total capacity. 20% of the time threads are parked and I suspect it indicates contention on `Controller.connections` field. 

## Next steps

* See if using bucket-level locking (dashmap) can help
* Instrument messaging layer with traces


martinthomson:
Based on the quality of the output and the fact that I'm out on leave for the next little while, I'm going to merge this, sight unseen.

🐿️

PR #90, opened 2022-09-19
[Semi honest sort] Computing bit permutations

richajaindce:
[Semi honest sort] Computing bit permutations
    Executes sorting of a bit column on mpc helpers. Each helper receives their input shares and do following steps
    1. local computation by `prepare_mult_inputs` which outputs 2 vectors [x,y]
    2. multiply each row of previous output individually (i.e. x*y) across mpc helpers.
    3. add ith column by i+len to obtain helper's share of sorted location, where len is same as input shares length

PS: Have stacked this over the "refactor" branch changes so you would see more changes than needed for this commit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,8 +48,9 @@ tokio = { version = "1.19.2", optional = true, features = ["rt", "rt-multi-threa
 tokio-stream = { version = "0.1.9", optional = true }
 tower = { version = "0.4.13", optional = true }
 tower-http = { version = "0.3.4", optional = true, features = ["trace"] }
-tracing = "0.1.35"
-tracing-subscriber = { version = "0.3.14", optional = true }
+# disable debug traces and below for release builds and keep everything for debug build
+tracing = { version = "0.1.35", features = ["max_level_trace", "release_max_level_info"] }
+tracing-subscriber = { version = "0.3.14", optional = true, features = ["env-filter"] }
 x25519-dalek = "2.0.0-pre.1"
 
 [dev-dependencies]
--- a/src/helpers/mesh.rs
+++ b/src/helpers/mesh.rs
@@ -6,9 +6,11 @@
 //! corresponding helper without needing to know the exact location - this is what this module
 //! enables MPC helper service to do.
 //!
+use crate::field::Field;
 use crate::helpers::error::Error;
 use crate::helpers::Identity;
 use crate::protocol::{RecordId, Step};
+use crate::secret_sharing::Replicated;
 use async_trait::async_trait;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
@@ -38,6 +40,15 @@ pub trait Mesh {
 
     /// Returns the unique identity of this helper.
     fn identity(&self) -> Identity;
+
+    /// Returns share of value one for this helper.
+    fn share_of_one<F: Field>(&self) -> Replicated<F> {
+        match self.identity() {
+            Identity::H1 => Replicated::new(F::ONE, F::ZERO),
+            Identity::H2 => Replicated::new(F::ZERO, F::ZERO),
+            Identity::H3 => Replicated::new(F::ZERO, F::ONE),
+        }
+    }
 }
 
 /// This is the entry point for protocols to request communication when they require it.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,5 @@
 #![deny(clippy::clone_on_ref_ptr)]
 
-#[cfg(feature = "cli")]
 pub mod chunkscan;
 pub mod cli;
 pub mod error;
--- a/src/protocol/sort/bit_permutations.rs
+++ b/src/protocol/sort/bit_permutations.rs
@@ -1,0 +1,159 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::mesh::{Gateway, Mesh},
+    protocol::{context::ProtocolContext, IPAProtocolStep, RecordId, SortStep},
+    secret_sharing::Replicated,
+};
+
+use futures::future::try_join_all;
+
+#[derive(Debug)]
+pub struct BitPermutations<'a, F> {
+    input: &'a [Replicated<F>],
+}
+
+impl<'a, F: Field> BitPermutations<'a, F> {
+    /// Create an object to generate bit permutations for a given bit column of query. This is GENBITPERM(Algorithm 3) from the paper
+    #[allow(dead_code)]
+    pub fn new(input: &'a [Replicated<F>]) -> BitPermutations<'a, F> {
+        Self { input }
+    }
+
+    /// In this step, multiplication inputs are generated locally at each helper from input share, x in following steps
+    /// 1. calculate 1 - x, x and concatenate them
+    /// 2. calculate cumulative sum at each vector row
+    /// 3. return back tuple of step 1 and step 2 output
+    fn prepare_mult_inputs<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
+        &self,
+        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
+    ) -> Vec<(Replicated<F>, Replicated<F>)>
+    where
+        F: Field,
+    {
+        let channel = ctx
+            .gateway
+            .get_channel(IPAProtocolStep::Sort(SortStep::BitPermutations));
+
+        self.input
+            .iter()
+            .map(|x| channel.share_of_one() - *x)
+            .chain(self.input.iter().copied())
+            .scan(Replicated::<F>::new(F::ZERO, F::ZERO), |sum, n| {
+                *sum = *sum + n;
+                Some((n, *sum))
+            })
+            .collect()
+    }
+
+    /// multiplies the input vector pairs across helpers and returns result
+    /// For this, it spawns all multiplication, wait for them to finish in parallel and then collect the results
+
+    #[allow(clippy::cast_possible_truncation)]
+    async fn secure_multiply<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
+        &self,
+        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
+        mult_inputs: Vec<(Replicated<F>, Replicated<F>)>,
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        let async_multiply = mult_inputs
+            .iter()
+            .enumerate()
+            .map(|(index, input)| async move {
+                ctx.multiply(
+                    RecordId::from(index as u32),
+                    IPAProtocolStep::Sort(SortStep::BitPermutations),
+                )
+                .await
+                .execute(input.0, input.1)
+                .await
+            });
+
+        try_join_all(async_multiply).await
+    }
+
+    /// Executes sorting of a bit column on mpc helpers. Each helper receives their input shares and do following steps
+    /// 1. local computation by `prepare_mult_inputs` which outputs 2 vectors [x,y]
+    /// 2. multiply each row of previous output individually (i.e. x*y) across mpc helpers.
+    /// 3. add ith column by i+len to obtain helper's share of sorted location, where len is same as input shares length
+    #[allow(dead_code)]
+    pub async fn execute<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
+        self,
+        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        let mult_input = self.prepare_mult_inputs(ctx);
+        assert_eq!(mult_input.len(), self.input.len() * 2);
+
+        let mut mult_output = self.secure_multiply(ctx, mult_input).await?;
+        assert_eq!(mult_output.len(), self.input.len() * 2);
+        // Generate permutation location
+        let len = mult_output.len() / 2;
+        for i in 0..len {
+            mult_output[i] = mult_output[i] + mult_output[i + len];
+        }
+        mult_output.truncate(len);
+
+        Ok(mult_output)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use rand::rngs::mock::StepRng;
+    use tokio::try_join;
+
+    use crate::{
+        field::Fp31,
+        protocol::{securemul, sort::bit_permutations::BitPermutations, IPAProtocolStep, QueryId}, test_fixture::{make_world,validate_and_reconstruct, TestWorld, make_contexts, share},
+    };
+
+    #[tokio::test]
+    pub async fn bit_permutations() {
+        let world: TestWorld<IPAProtocolStep> = make_world(QueryId);
+        let context = make_contexts(&world);
+        let mut rand = StepRng::new(100, 1);
+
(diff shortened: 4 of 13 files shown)


src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+                shares[i].push(share[i]);
+            }
+        }
+
+        let h0_future = generate_bit_permutations(&context[0]).execute(&shares[0]);
+        let h1_future = generate_bit_permutations(&context[1]).execute(&shares[1]);
+        let h2_future = generate_bit_permutations(&context[2]).execute(&shares[2]);
+
+        let result = try_join!(h0_future, h1_future, h2_future).unwrap();
+        let expected_sort_output = [5_u128, 1, 6, 2, 3, 7, 4].as_ref();
```
akoshelev, line 138: can you add an explanation why this output is expected? 

```suggestion
        let expected_sort_output = [5_u128, 1, 6, 2, 3, 7, 4];
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /// multiplies the input vector pairs across helpers and returns result
+    /// For this, it spawns all multiplication, wait for them to finish in parallel and then collect the results
+
+    #[allow(clippy::cast_possible_truncation)]
+    async fn secure_multiply<F>(
+        &self,
+        mult_inputs: Vec<(Replicated<F>, Replicated<F>)>,
+    ) -> Vec<Replicated<F>>
```
akoshelev, line 60: I think as it involves communication, it is fallible, therefore `Result<Vec<_>, E>` is more appropriate

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+                    .multiply(
+                        RecordId::from(index as u32),
+                        IPAProtocolStep::Sort(SortStep::BitPermutations),
+                    )
+                    .await
+                    .execute(input.0, input.1)
+                    .await
+            });
+
+        join_all(async_multiply.map(|handle| async { handle.await.unwrap() })).await
```
akoshelev, line 78:
```suggestion
        try_join_all(async_multiply).await
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        let result = try_join!(h0_future, h1_future, h2_future).unwrap();
+        let expected_sort_output = [5_u128, 1, 6, 2, 3, 7, 4].as_ref();
+
+        assert_eq!(result.0.len(), input_len);
+        assert_eq!(result.1.len(), input_len);
+        assert_eq!(result.2.len(), input_len);
+
+        (0..result.0.len()).for_each(|i| {
+            assert_eq!(
```
akoshelev, line 145: I am curious if `validate_and_reconstruct` is what you want here. It will reconstruct the value from all shares, perform some validation checks and then you can compare the value with expected output

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+    protocol::{context::ProtocolContext, IPAProtocolStep, RecordId, SortStep, Step},
+    secret_sharing::Replicated,
+};
+
+use futures::future::join_all;
+
+#[derive(Debug)]
+pub struct BitPermutations<'a, G, M, S: Step + SpaceIndex> {
+    pub(crate) context: &'a ProtocolContext<'a, G, S>,
+    pub(crate) phantom: PhantomData<M>,
```
akoshelev, line 19: I think we can make it a bit more ergonomic to use if we invert the state and keep the input here instead of `context`.

```rust
#[derive(Debug)]
pub struct BitPermutations<'a, F> {
    input: &'a [Replicated<F>]
}
```

I have a sketch ready if you want to see how it would look like https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4359f076c0dfccc632e67e4ccf422166

let me know what you think

src/lib.rs
```diff
@@ -1,14 +1,10 @@
 #![deny(clippy::clone_on_ref_ptr)]
 
-mod chunkscan;
 #[cfg(feature = "cli")]
```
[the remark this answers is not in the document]
richajaindce: yeah looks like it.. removing it

src/protocol/sort/bit_permutations.rs
richajaindce: Yes that looks like a better function. However, the order in which we get shares is different since each helper would give out one share, hence we need to handle it differently

src/protocol/sort/bit_permutations.rs
richajaindce: Looks good! Lot of boilerplate code got removed e.g. PhantomData and interface to call BitPermutations looks much cleaner!

richajaindce: However, this pattern is opposite of what is done in securemul i.e. it passes inputs to execute vs this passes inputs to struct declaration. Should we have similar approach in both declarations?

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub async fn execute<F>(self, shares: &[Replicated<F>]) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        let mult_output = self.secure_multiply(self.prepare_mult_inputs(shares)).await;
+
+        assert_eq!(mult_output.len(), shares.len() * 2);
+        // Generate permutation location
+        let len = mult_output.len() / 2;
+        Ok((0..len)
```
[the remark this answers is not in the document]
richajaindce: not sure why we would do that since output of multiplication is 2x size and we need to add ith index to len + ith index element.

src/helpers/mesh.rs
```diff
@@ -38,6 +40,9 @@ pub trait Mesh {
 
     /// Returns the unique identity of this helper.
     fn identity(&self) -> Identity;
+
+    /// Returns share of value one for this helper.
+    fn share_of_one<F: Field>(&self) -> Replicated<F>;
```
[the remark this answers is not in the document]
richajaindce: This is a state which each helper needs to remember which is controlled by Mesh. mock.rs seems to be the correct place to implement this. Let me know otherwise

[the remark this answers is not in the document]
akoshelev: My thinking is that `share_of_one` implementation is not `Mesh` specific. For example, if we have 3 different implementations of `Mesh` (test,prod,e2e), all of them will implement `share_of_one` in exactly the same way, copy-pasting the same code.


There are two ways to deal with it. First - you can implement `share_of_one` inside `Mesh` trait itself, providing a blanket implementation for everything that implements `Mesh`. Alternatively, you can move it to `Replicated`:

```
   impl Replicated<F> {
         pub fn share_of_one(identity: Identity) -> Self
   }
```

src/protocol/sort/bit_permutations.rs
akoshelev: I believe it is ok for them to be different - secure mul is just a very important primitive to be used everywhere, so we may implement it differently to keep it ergonomic to use

src/helpers/mesh.rs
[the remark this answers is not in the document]
richajaindce: I will move it to the trait. That makes more sense to me

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// 3. add ith column by i+len to obtain helper's share of sorted location, where len is same as input shares length
+    #[allow(dead_code)]
+    pub async fn execute<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
+        self,
+        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        let mult_input = self.prepare_mult_inputs(ctx);
```
akoshelev, line 89: I stared at this code again and I believe there is a way to avoid buffering. What happens here is `mult_input` is a vector of size `2*N` and `mult_output` is a vector as well (same size). so in line 93 we have 4*N items in memory. we can do better by returning an iterator from `prepare_mult_inputs` like this


```rust
 fn prepare_mult_inputs<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
        &self,
        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
    ) -> impl Iterator<Item=(Replicated<F>, Replicated<F>)> + 'a
    where
        F: Field,
    {
        let share_of_one = ctx.gateway
            .get_channel(IPAProtocolStep::Sort(SortStep::BitPermutations))
            .share_of_one();

        self.input
            .iter()
            .map(move |x| share_of_one - *x)
            .chain(self.input.iter().copied())
            .scan(Replicated::<F>::new(F::ZERO, F::ZERO), |sum, n| {
                *sum = *sum + n;
                Some((n, *sum))
            })
    }
```

`secure_multiply` just need to take an iterator instead of `Vec`

```rust
async fn secure_multiply<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
        &self,
        ctx: &ProtocolContext<'_, G, IPAProtocolStep>,
        mult_inputs: impl Iterator<Item=(Replicated<F>, Replicated<F>)>,
    ) -> Result<Vec<Replicated<F>>, BoxError>
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        let async_multiply = mult_inputs
+            .iter()
+            .enumerate()
+            .map(|(index, input)| async move {
+                ctx.multiply(
+                    RecordId::from(index as u32),
```
akoshelev, line 66: I think record id should be provided from outside, i.e. our input is `(RecordId, Replicated<F>)`

src/protocol/sort/bit_permutations.rs
richajaindce: Have made the change such that at no given time we have 4N memory. 
prepare_mult_inputs returns an iterator. We run a map on top of it to call secure_multiply for each value (which will also return in an iterator). This returns an async handle which we collect in a vector. In the end, join_all is run on all async handles on multiplications. LMK if this looks fine

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn new(input: &'a [Replicated<F>]) -> BitPermutations<'a, F> {
+        Self { input }
+    }
+
+    /// In this step, multiplication inputs are generated locally at each helper from input share, x in following steps
+    /// 1. calculate 1 - x, x and concatenate them
+    /// 2. calculate cumulative sum at each vector row
+    /// 3. return back tuple of step 1 and step 2 output
+    #[allow(clippy::cast_possible_truncation)]
+    fn prepare_mult_inputs<M: Mesh + 'a, G: Gateway<M, IPAProtocolStep>>(
```
akoshelev, line 29:
```suggestion
    fn prepare_mult_inputs<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+        &self,
+        ctx: &ProtocolContext<'a, G, IPAProtocolStep>,
+    ) -> impl Iterator<Item = (RecordId, (Replicated<F>, Replicated<F>))> + 'a
+    where
+        F: Field,
+    {
+        let channel = ctx
+            .gateway
+            .get_channel(IPAProtocolStep::Sort(SortStep::BitPermutations));
```
akoshelev, line 38:
```suggestion
        let share_of_one = ctx
            .gateway
            .get_channel(IPAProtocolStep::Sort(SortStep::BitPermutations))
            .share_of_one();
```

you don't need to keep a reference to `Mesh` for the whole iteration, that drops the lifetime `'a` requirement for it

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        self.input
+            .iter()
+            .map(move |x: &Replicated<F>| channel.share_of_one() - *x)
```
akoshelev, line 42:
```suggestion
            .map(move |x: &Replicated<F>| share_of_one - *x)
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+                *sum += n;
+                Some((RecordId::from(index as u32), (n, *sum)))
+            })
+    }
+
+    /// multiplies the input vector pairs across helpers and returns result
+    /// For this, it spawns all multiplication, wait for them to finish in parallel and then collect the results
+
+    #[allow(clippy::cast_possible_truncation)]
+    async fn secure_multiply<M: Mesh + 'a, G: Gateway<M, IPAProtocolStep>>(
```
akoshelev, line 55:
```suggestion
    async fn secure_multiply<M: Mesh, G: Gateway<M, IPAProtocolStep>>(
```

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+            .get_channel(IPAProtocolStep::Sort(SortStep::BitPermutations));
+
+        self.input
+            .iter()
+            .map(move |x: &Replicated<F>| channel.share_of_one() - *x)
+            .chain(self.input.iter().copied())
+            .enumerate()
+            .scan(Replicated::<F>::new(F::ZERO, F::ZERO), |sum, (index, n)| {
+                *sum += n;
+                Some((RecordId::from(index as u32), (n, *sum)))
```
akoshelev, line 47: `RecordId` should be provided from the outside of sort imo, but we can push this first and follow up on that

src/protocol/sort/bit_permutations.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<Replicated<F>, BoxError>
+    where
+        F: Field,
+    {
+        ctx.multiply(
+            mult_input.0,
+            IPAProtocolStep::Sort(SortStep::BitPermutations),
+        )
+        .await
+        .execute(mult_input.1 .0, mult_input.1 .1)
```
akoshelev, line 68:
```suggestion
        .execute(mult_input.1.0, mult_input.1.1)
```

Cargo.toml
```diff
@@ -37,6 +37,7 @@ pin-project = "1.0.11"
 rand = "0.8"
 rand_core = "0.6"
 rand_distr = "0.4.3"
+rhs_first_assign = "0.1.0"
```
akoshelev, line 40: eh, I guess I was wrong claiming it is possible to use `AddAssign` in the iteration. Borrow checker would prevent that. I don't think it is worth a new dependency to be able to write `+=` instead of `+`. Shall we revert back to it?

akoshelev (approved):
gtg after applying suggestions

src/protocol/sort/bit_permutations.rs
richajaindce: makes sense!

Cargo.toml
richajaindce: we can also use an intermediate variable to addassign, i will do that instead

akoshelev:
Merging it myself while Martin is away.

PR #93, opened 2022-09-21
Update source/trigger events structure for synthetic data (in clear) generation

taikiy:
## Overview
This diff updates source and trigger structs used for synthetic data generation. Synthetic data will be used to either create an attribution report benchmark (executed in clear), or transform the data into encrypted and secret shared format to be fed to IPA to generate reproducible test results.
Data structure is (almost) the same as the one specified in the [end-to-end doc](https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#gathering-events). Just not encrypted nor secret shared. (the struct in end-to-end doc is missing `epoch` and that's a known issue)

```
struct {
    int match_key_share;
    int timestamp_share;
    int is_trigger_report_share;
    int breakdown_key_share;
    int trigger_value_share;
} generic_report_share
```

## What have changed?
* Removed old event structs, and replaced with the new ones. Now, we have a common data model across all the files in the repository.
* Merged source and trigger events into `GenericReport`
* Report only contains 1 match key instead of multiple
* Added `attribution_constraint_id`
* Removed --secret-share option. If the test data needs to be fed into MPC, we can create another module which encrypts and secret shares the data (not part of this diff).

Since this tool's purpose is to generate human-readable test data, resulting output contains fields in clear and not secret shared.


src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
-    pub timestamp: SecretShare,
+    fn epoch(&self) -> Self::Epoch;
+    fn offset(&self) -> Self::Offset;
 }
 
-#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
-pub struct SourceEvent {
-    pub event: Event,
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct EventTimestamp(u64);
```
akoshelev, line 127: +1 on hiding the internal representation of the timestamp and defining `From`!
also curious why we need `u64` if we seem to be able to keep everything inside a `u32`?

src/bin/ipa_bench/cmd.rs
```diff
@@ -158,15 +146,12 @@ impl Command {
...
 
         let mut rng = random_seed.map_or(StdRng::from_entropy(), StdRng::seed_from_u64);
-        let mut ss_rng = random_seed.map_or(StdRng::from_entropy(), StdRng::seed_from_u64);
 
         let (s_count, t_count) = generate_events(
             &sample,
             DEFAULT_EVENT_GEN_COUNT * scale_factor,
-            epoch,
-            secret_share,
+            &epoch,
```
thurstonsand, line 153: why are `sample` and `epoch` passed in as shared values?

src/bin/ipa_bench/gen_events.rs
```diff
@@ -203,109 +79,71 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
...
-            events.push(Event::Source(SourceEvent::new(
-                matchkeys.clone(),
-                //TODO: Carry to next epoch if timestamp > DAYS_IN_EPOCH
-                params.epoch,
-                u32::try_from(t.as_secs()).unwrap(),
-                params.breakdown_key,
-            )));
-        }
+    let matchkey = gen_matchkey(rng);
+    let base_epoch: EventTimestamp = (u64::from(*epoch) * EventTimestamp::SECONDS_IN_EPOCH).into();
```
thurstonsand, line 93: as @akoshelev had mentioned in a diff PR, it would be nice to wrap this timestamp arithmetic in a struct to hide all these kinds of conversions

src/bin/ipa_bench/gen_events.rs
```diff
@@ -203,109 +79,71 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
...
 }
 
-fn gen_matchkeys<R: RngCore + CryptoRng>(count: u8, rng: &mut R) -> Vec<MatchKey> {
-    let mut mks = Vec::new();
-
-    for _ in 0..count {
-        mks.push(rng.gen::<u64>());
-    }
-    mks
+fn gen_matchkey<R: RngCore + CryptoRng>(rng: &mut R) -> MatchKey {
```
thurstonsand, line 133: seems like you could just remove this fn and use `rng.gen` directly

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
+                event,
+                breakdown_key,
+            } => {
+                let mut state = serializer.serialize_struct("GenericReport", 7)?;
+                state.serialize_field("matchkey", &event.matchkey)?;
+                state.serialize_field(
+                    "attribution_constraint_id",
+                    &event.attribution_constraint_id,
+                )?;
+                state.serialize_field("epoch", &event.timestamp.epoch())?;
```
akoshelev, line 249: can we implement `Serialize` for `EpochTimestamp` and use it here? at the very least we can fit both epoch and offset in a single `u32` or `[u8; 3]`

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct Event {
+    // An identifier, set in the user agent, which identifies an individual person. This must never be released (beyond
+    /// the match key provider) to any party in unencrypted form. For the purpose of this tool, however, the value is in
+    /// clear and not secret shared.
+    pub matchkey: MatchKey,
+
+    /// An identifier, specified by the report collector, to denote if a given pair of source and trigger events can be
```
thurstonsand, line 201: huh, what is this identifier for? specifically, in what scenarios would it be `None`, and what is the significance of the `Number` when it is there?

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum GenericReport {
+    /// An event produced on websites/apps when a user interacts with an ad (i.e. impression, click).
+    Source {
+        event: Event,
+
+        /// A key, specified by the report collector, which allows for producing aggregates across many groups (or breakdowns.)
+        /// This field is only valid if is_trigger_report is [false].
```
thurstonsand, line 216: `is_trigger_report` is no longer valid, since this is an enum now, right?

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
+        breakdown_key: Number,
+    },
+
+    /// An event produced on websites/apps when a user takes an action (i.e. product view, purchase).
+    Trigger {
+        event: Event,
+
+        /// The value of the trigger report to be aggregated. This field is only valid if is_trigger_report is [true].
```
thurstonsand, line 224: same here for `is_trigger_report`

src/bin/ipa_bench/models.rs
taikiy: This is in our [e2e doc](https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#types-of-queries), which is used together with match keys for attribution.

We only know that it will be an `int`. This identifier is also optional (only use match keys in attribution process). Using `Option` complicates the logic because we have to deal with `None` but I couldn't use `0` to indicate that the value is non-existent since `0` is a valid integer.

src/bin/ipa_bench/models.rs
taikiy: Thanks for catching!

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
     }
 }
 
-#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
-pub struct Event {
-    /// Secret shared and then encrypted match keys.
-    pub matchkeys: Vec<SecretShare>,
+pub trait EpochDuration: Sized + Add<Output = Self> + AddAssign + PartialOrd + From<u64> {
```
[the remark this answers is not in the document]
taikiy: I wanted to create type aliases `Epoch` and `Offset` within `EventTimestamp`, but it seems that the only possible way to define type aliases are within traits. Also, I think there will be secret-shared version of `EventTimestamp`, but I'm not sure how that will look like yet. Do you think we should just use `EventTimestamp` without this trait?

src/bin/ipa_bench/models.rs
```diff
@@ -109,57 +111,169 @@ impl SecretSharable for u64 {
...
 }
 
-#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
-pub struct SourceEvent {
-    pub event: Event,
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct EventTimestamp(u64);
+
+impl EpochDuration for EventTimestamp {
+    type Epoch = u8;
```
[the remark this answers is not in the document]
taikiy: Yes, let me open an issue in the repo. We mentioned cross-epoch attribution during our meetup and I think Benjamin and Daniel are investigating. Nevertheless, I'll open an issue just to make sure we are covering all cases like key rotation, MAC, DP and attribution.

src/bin/ipa_bench/models.rs
taikiy: I chose `u64` because rust's `Duration` struct uses it to represent seconds like below. It eliminates extra casts and can directly call `from`/`into`.

That said, the parameter is nameless and private, hence the only way to instantiate `EventTimestamp` outside `models` is `From`, which is defined @161 in this file.

```
pub struct Duration {
    secs: u64,
    nanos: u32,
}
```

src/bin/ipa_bench/models.rs
taikiy: Yes, let me work on improving `Serialize` and `Deserialize` implementation in the next diff.

I think we need multiple `EventTimestamp` (or `GenericReport` for that matter) variants. One that is used in MPC, which should implement `Serialize` that mixes epoch and offset to `[u8; 3]', and another for outside MPC such as synthetic data generation (e2e test input). For the latter, I would like to keep the serialization simple.

src/bin/ipa_bench/gen_events.rs
taikiy: I'll create `add_epoch()`.

taikiy: Thanks! I created `From<Epoch>'.

src/bin/ipa_bench/models.rs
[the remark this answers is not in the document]
taikiy: https://github.com/patcg-individual-drafts/ipa/issues/22

src/bin/ipa_bench/models.rs
akoshelev: to me it makes sense to use the smallest container value possible - that helps a lot with CPU cache efficiency, memory usage etc. usually implementing `From<Duration>`, `From<u32>`, `From<u64>` is enough to make API easy to use

src/bin/ipa_bench/models.rs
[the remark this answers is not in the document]
akoshelev: yep, I think trait is probably redundant here. for epoch and offset you could define a type alias because `EventTimestamp` is relatively simple

src/bin/ipa_bench/cmd.rs
akoshelev: I had the same Clippy warning and my conclusion is that we should suppress it because it makes our API look weird for arguments that implement `Copy`

src/bin/ipa_bench/gen_events.rs
```diff
@@ -203,109 +86,66 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
...
-    if secret_share {
-        for mk in &matchkeys {
-            // Currently, all geneerated match keys are set in all source events from the same user. This is an ideal
-            // scenario where all devices are used equally. In reality, however, that isn't the case. Should we pick
-            // a few match keys out from the events?
-            ss_mks.push(mk.xor_split(ss_rng));
-        }
-    }
+) -> Vec<GenericReport> {
+    let mut reports: Vec<GenericReport> = Vec::new();
```
akoshelev, line 97:
```suggestion
    let mut reports: Vec<GenericReport> = Vec::with_capacity(impressions + conversions);
```

akoshelev (commented):
so my understanding is that we still need to work out serialization, correct?

src/bin/ipa_bench/models.rs
taikiy: That was another reason that led me to use trait so that we can have two different EventTimestamp structs. Let me implement `Serialize` based on your suggestion first and see if that works for all use cases including an E2E test. If not, we can make modifications then.

taikiy:
@akoshelev 

> so my understanding is that we still need to work out serialization, correct?

Yes. I also need to implement `Deserialize` which will be more complex than `Serialize`. I think that'll make this diff too large so I'd rather implement that in another diff.

taikiy:
Rust check failure is addressed by #96.

src/bin/ipa_bench/models.rs
akoshelev: how about taking the serialization out of this diff? I think apart from it, it is in a good shape.

src/bin/ipa_bench/gen_events.rs
```diff
@@ -1,132 +1,25 @@
...
-    conversions: u8,
-    epoch: Epoch,
-    breakdown_key: Number,
-}
-
 // TODO: Currently, users are mutually exclusive in each ad loop (i.e. User A in ad X will never appear in other ads).
 // We need to generate events from same users across ads (but how often should a user appear in different ads?)
-// "Ads" doesn't mean FB's L3 ads. It could be ads from different businesses.
 
+#[allow(clippy::needless_pass_by_value)]
```
akoshelev, line 17: are you sure you need it? I was able to run clippy locally without this

src/bin/ipa_bench/gen_events.rs
```diff
@@ -203,109 +80,104 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
...
-
-    for _ in 0..count {
-        mks.push(rng.gen::<u64>());
+/// Adds two `EventTimestamp` instances.
+/// `offset` overflow will carry to `epoch`.
+/// `epoch` overflow will wrap and has no effect on `offset`.
+///
+/// # Exmaples
+///
+/// ```no_run
```
akoshelev, line 142: I don't think it is discoverable by rust engine anyway :( I wasn't able to run it even with `no_run` removed

src/bin/ipa_bench/gen_events.rs
```diff
@@ -203,109 +80,104 @@ pub fn generate_events<R: RngCore + CryptoRng, W: io::Write>(
...
+///
+/// let ts = add_event_timestamps(EventTimestamp::new(0, 1), EventTimestamp::new(0, OFFSET_MAX));
+/// assert_eq!(ts.epoch(), 1);
+/// assert_eq!(ts.offset(), 0);
+///
+/// let ts = add_event_timestamps(EventTimestamp::new(1, 1), EventTimestamp::new(Epoch::MAX, 0));
+/// assert_eq!(ts.epoch(), 0);
+/// assert_eq!(ts.offset(), 1);
+/// ```
+fn add_event_timestamps(rhs: EventTimestamp, lhs: EventTimestamp) -> EventTimestamp {
```
akoshelev, line 153: you can also add a method to `EventTimestamp` to do that

```rust

impl EventTimestamp {
    /// private 
    fn add(rhs: Self, lhs: Self) -> Self { ... }
}
```

akoshelev (approved):
some suggestions but otherwise lgtm

src/bin/ipa_bench/gen_events.rs
taikiy: Let's keep this function outside `EventTimestamp` for now because we might not need to support Add operation for timestamps. This is currently only needed for synthetic data generation.

PR #94, opened 2022-09-21
implement Server-Side Receive Logic for Network Layer

thurstonsand:
Implements the server-side receive logic for the network layer. handles the `recv_stream` half of the [`Network`](https://github.com/private-attribution/ipa/blob/main/src/helpers/fabric.rs#L25) interface.


src/protocol/mod.rs
```diff
@@ -21,7 +22,10 @@ use std::hash::Hash;
 /// See `IPAProtocolStep` for a canonical implementation of this trait. Every time we switch to
 /// use a new circuit, there will be an additional struct/enum that implements `Step`, but eventually
 /// it should converge to a single implementation.
-pub trait Step: Copy + Clone + Debug + Eq + Hash + Send + 'static {}
+pub trait Step:
+    Copy + Clone + Debug + Eq + Hash + Send + TryFrom<String> + ToString + 'static
```
akoshelev, line 26: the official guidance is to implement `Display` instead: https://doc.rust-lang.org/std/string/trait.ToString.html

src/cli/net/client/mod.rs
```diff
@@ -38,25 +48,63 @@ impl MpcHandle for MpcHttpConnection {
...
         let result = hyper::body::to_bytes(response.into_body()).await?;
 
         Ok(result.to_vec())
     }
+
+    async fn mul<S: Step, M: Message>(
+        &self,
+        query_id: QueryId,
+        step: S,
+        messages: &[MessageEnvelope<M>],
```
akoshelev, line 94: I am not sure this client API is useful for us. as a client I want to multiply `a`\*`b`. `/mul/query-id/{}/step/{}` is an endpoint for helpers to send intermediate results of multiplication before they can produce some result. as such, I think we should provide more high level API to multiply two values. For each helper client, that would mean taking two secret shares (a and b) and sending back a secret share (a*b). 

We should also keep in mind that we don't want this API to be available in prod because it reveals the result of multiplication

src/cli/net/client/mod.rs
```diff
@@ -38,25 +48,63 @@ impl MpcHandle for MpcHttpConnection {
...
-    pub fn new(addr: &str) -> Self {
+    pub fn new(addr: Uri) -> Self {
         // this works for both http and https
         let https = HttpsConnector::new();
         let client = Client::builder().build::<_, Body>(https);
 
         Self {
             client,
-            addr: addr.parse().expect("Cannot parse the URI"),
+            addr_parts: addr.into_parts(),
```
akoshelev, line 58: I'd rather keep scheme and authority in the state: `scheme: Scheme, authority: Authority`. you wouldn't need to unwrap these values each time you want to access them

src/cli/net/client/mod.rs
thurstonsand: I'm curious -- what will the mechanism be to communicate the intermediate results of the multiplication then?

src/protocol/mod.rs
thurstonsand: fixed

src/cli/net/client/mod.rs
thurstonsand: done

src/cli/net/client/mod.rs
akoshelev: I agree that helper endpoint should accept `mul/query-id/...` requests, so server implementation for this endpoint is important. I would not expect any client application to call this API directly though

akoshelev: eh, I guess I missed that change from the refactoring PR: https://github.com/martinthomson/raw-ipa/tree/9adbe555be149c0b3e1cc3b0afd9f6e72b6181a6/src/net. 

`net` module does not belong to `cli`, it should reside in  the `src` folder. would you be able to move it back as part of this change? I guess that removes the confusion about the client API as well

src/cli/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,106 @@
...
+        let step = S::try_from(step).map_err::<serde_json::Error, _>(serde::de::Error::custom)?;
+        Ok(Path(query_id, step))
+    }
+}
+
+/// accepts all the relevant information from the request, and push all of it onto the `outgoing`
+/// channel
+/// TODO: implement the receiving end of `outgoing`
+pub async fn handler<S: Step>(
+    outgoing: mpsc::Sender<BufferedMessages<S>>,
```
akoshelev, line 26: I think we should have a stateful router that injects the entry point to networking layer through the extensions mechanism. Injecting individual channels like this works too, but it can quickly grow unmaintainable as we would be managing individual channels for every handler (or share a single channel between some of them and use unique channels for others).

I agree that handlers should do almost no work and after parsing and validating the request, pass the payload to the networking layer, but I think we should make a "gateway" to it available to handlers through extensions

src/cli/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,106 @@
...
+        data_size: usize,
+        body: &'static [u8],
+    ) -> Request<Body> {
+        assert_eq!(
+            body.len() % data_size,
+            0,
+            "body len must align with data_size"
+        );
+        let uri = format!(
+            "http://localhost:3000/mul/query-id/{}/step/{}",
```
akoshelev, line 66: please make sure you test https too - there are `echo` tests that do that, you can probably lift some methods/functions from that module to re-use them for these tests too

src/cli/net/mod.rs
```diff
@@ -1,12 +1,78 @@
...
+use async_trait::async_trait;
+use axum::body::Bytes;
+use axum::extract::{FromRequest, RequestParts};
+use axum::http::header::HeaderName;
+
+static OFFSET_HEADER_NAME: HeaderName = HeaderName::from_static("offset");
+static DATA_SIZE_HEADER_NAME: HeaderName = HeaderName::from_static("data-size");
+
+pub struct RecordHeaders {
+    offset: usize,
```
akoshelev, line 26: what is the valid range for offset? is it possible to use u32 here? in general I am not a big fan of using `usize` for application data because it is platform dependent. It is very rare to actually hit that issue and we could use static assertions to enforce minimum length for usize, but imo it is a good practice to use fixed length integers even if it costs inconveniences while using these as indices

src/cli/net/mod.rs
```diff
@@ -1,12 +1,78 @@
...
+    data_size: usize,
```
akoshelev, line 27: would it be ever different from [`content-length`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header?

src/cli/net/server/handlers/mul.rs
akoshelev: btw converting bytes to a vector of messages is probably a good use case for `spawn_blocking` because it may block scheduler for longer that we would want.

src/cli/net/server/mod.rs
```diff
@@ -21,23 +26,69 @@ pub mod handlers;
...
+    #[error("header not found: {0}")]
+    MissingHeader(String),
+    #[error("invalid header: {0}")]
+    InvalidHeader(BoxError),
+    #[error(transparent)]
+    BadPathString(#[from] PathRejection),
+    #[error(transparent)]
+    HttpError(#[from] hyper::Error),
+    #[error("parse error: {0}")]
+    SerdeError(#[from] serde_json::Error),
```
akoshelev, line 38: is this used anywhere?

src/cli/net/server/mod.rs
```diff
@@ -21,23 +26,69 @@ pub mod handlers;
...
+    #[error("could not forward messages: {0}")]
+    SendError(BoxError),
+}
+
+/// [`From`] implementation for [`MpcServerError::SendError`]
+impl<T: Debug + Send + Sync + 'static> From<mpsc::error::SendError<T>> for MpcServerError {
```
akoshelev, line 44: nit: it may be better to drop these requirements for `T` as `SendError` is literally just "channel closed" error.  You could implement this for any `T` if you convert this string to `Error` (there is a From implementation for that)

thurstonsand: probably for the best, because this is also driving `Step` to be `Sync`, since `T` here must be `Sync`

src/cli/net/server/mod.rs
thurstonsand: not anymore. removed

src/cli/net/server/handlers/mul.rs
thurstonsand: i'll keep `spawn_blocking` in mind when I get to that part of the implementation.

i also agree with your gateway idea. I think, for the purposes of this PR, it might make sense to just leave this as a channel here, with the idea that i'll swap it out for a gateway, for effectively the same functionality. However, I don't have a problem switching to an `Extensions` implementation. that's an interesting idea, actually, because it could parse the `query_id` and `step`, then inject the correct gateway from a `HashMap` of all existing gateways

src/protocol/mod.rs
```diff
@@ -24,7 +25,10 @@ use crate::helpers::prss::SpaceIndex;
 /// See `IPAProtocolStep` for a canonical implementation of this trait. Every time we switch to
 /// use a new circuit, there will be an additional struct/enum that implements `Step`, but eventually
 /// it should converge to a single implementation.
-pub trait Step: Copy + Clone + Debug + Eq + Hash + Send + 'static {}
+pub trait Step:
+    Copy + Clone + Debug + Display + Eq + Hash + Send + Sync + TryFrom<String, Error = Error> + 'static
```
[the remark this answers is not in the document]
thurstonsand: i mentioned below, but maybe not if we can simplify handling of `SendError`

src/test_fixture/world.rs
```diff
@@ -44,6 +54,24 @@ impl Debug for TestStep {
     }
 }
 
+impl TryFrom<String> for TestStep {
+    type Error = Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let value = value.strip_prefix('/').unwrap_or(&value);
+        if value == "mul2" {
```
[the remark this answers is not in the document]
thurstonsand: sure, we can. technically, this should strictly be machine talking to machine, so urls should be exact. but doesn't hurt

src/test_fixture/world.rs
```diff
@@ -44,6 +54,24 @@ impl Debug for TestStep {
...
+            Ok(Self::Mul2)
+        } else {
+            value
+                .split_once('/')
+                .and_then(|(pre, suf)| (pre == "mul1").then_some(suf))
+                .and_then(|suf| suf.parse::<u8>().ok())
+                .map(Self::Mul1)
+                .ok_or(Error::InvalidId)
```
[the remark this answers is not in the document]
thurstonsand: i don't mind doing that. I come from a background of security-conscious user-facing APIs, where even revealing details of the nature of the error can be bad security practice. context here is different tho, so including that detail is probably good.

src/protocol/mod.rs
```diff
@@ -75,6 +95,43 @@ impl SpaceIndex for IPAProtocolStep {
...
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub enum SortStep {
+    BitPermutations,
+}
+
+impl SortStep {
+    const BIT_PERMUTATIONS_STR: &'static str = "bit-permutations";
+}
+
+impl Debug for SortStep {
```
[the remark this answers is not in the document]
thurstonsand: this comes from merging `main`. there's a lot of these custom `Debug`s for all the `Step`s. I agree that we should probably just use the derived definition. I'll go through and replace

src/protocol/mod.rs
```diff
@@ -35,9 +39,24 @@ pub enum IPAProtocolStep {
...
+    type Error = Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let value = value.strip_prefix('/').unwrap_or(&value);
+        if value == Self::CONVERT_SHARES_STR {
+            Ok(Self::ConvertShares)
+        } else if let Some(rem) = value.strip_prefix(Self::SORT_STR) {
+            Ok(Self::Sort(String::from(rem).try_into()?))
+        } else {
+            Err(Error::NotFound)
```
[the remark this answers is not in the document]
thurstonsand: in this case we're parsing the path. if it's an invalid path, should that not be a `NotFound` error?

src/cli/net/mod.rs
thurstonsand: ok, I had no strong feelings of using `usize`. i can switch to u32. i would imagine max `offset` == max query size, which probably won't exceed a u32

src/cli/net/mod.rs
thurstonsand: i thought `content-length` refers to the size of the entire message? this should be the length of a single piece of data within the message

src/error.rs
```diff
@@ -20,6 +20,8 @@ pub enum Error {
     #[error("too many helpers")]
     TooManyHelpers,
 
+    #[error("failed to parse int")]
+    ParseError(#[from] std::num::ParseIntError),
```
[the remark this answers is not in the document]
thurstonsand: actually, nevermind. it is needed [here](https://github.com/martinthomson/raw-ipa/pull/94/files#diff-2bb5f35b32641587c1cae31f2b3db2dc028d3606820ce9fda87da4841ec9752cR17)

src/cli/net/server/mod.rs
thurstonsand: sorry, this is also used [here](https://github.com/martinthomson/raw-ipa/pull/94/files#diff-2bb5f35b32641587c1cae31f2b3db2dc028d3606820ce9fda87da4841ec9752cR17)

src/protocol/mod.rs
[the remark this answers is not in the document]
thurstonsand: yep, was able to remove it after modifying `From for SendError`

src/cli/net/mod.rs
thurstonsand: leave docs to clarify

thurstonsand: make a GH issue to alias offset type, so we can change max size of data at later date

src/cli/net/server/handlers/mul.rs
thurstonsand: done

src/net/client/mod.rs
```diff
@@ -0,0 +1,213 @@
+use super::Command;
```
thurstonsand, line 1: unfortunately git has decided this is a brand new file... i'll try to point to the new pieces of code

src/net/client/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    async fn echo(&self, s: &str) -> Result<Vec<u8>, MpcClientError> {
+        let uri = self.build_uri(format!("/echo?foo={}", s))?;
+
+        let response = self.client.get(uri).await?;
+        let result = hyper::body::to_bytes(response.into_body()).await?;
+
+        Ok(result.to_vec())
+    }
+
+    async fn mul<S: Step>(
```
thurstonsand, line 97: mul implementation

src/net/client/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        let resp_status = response.status();
+        resp_status
+            .is_success()
+            .then_some(())
+            .ok_or(MpcClientError::FailedRequest(resp_status))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
thurstonsand, line 119: added tests

src/net/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::str::FromStr;
+
+use crate::net::server::MpcServerError;
+use crate::protocol::{QueryId, RecordId};
+use async_trait::async_trait;
+use axum::body::Bytes;
+use axum::extract::{FromRequest, RequestParts};
+use axum::http::header::HeaderName;
+
+static OFFSET_HEADER_NAME: HeaderName = HeaderName::from_static("offset");
```
thurstonsand, line 23: standardize headers

src/net/server/mod.rs
```diff
@@ -21,23 +27,71 @@ pub mod handlers;
...
+            | Self::MissingHeader(_)
+            | Self::InvalidHeader(_) => StatusCode::BAD_REQUEST,
+            Self::HttpError(_) | Self::SendError(_) => StatusCode::INTERNAL_SERVER_ERROR,
         };
 
         (status_code, self.to_string()).into_response()
     }
 }
 
+/// TODO: how to remove <S, M> from router definition? It should not be there
```
akoshelev, line 81: curious why do you need to get rid of `S`?

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,186 @@
...
+/// TODO: value of map may change to different implementation
+pub type GatewayMap<S> = Arc<Mutex<HashMap<(QueryId, S), mpsc::Sender<BufferedMessages<S>>>>>;
+
+/// Injects the appropriate gateway/mesh into the Axum request, so that downstream handlers have
+/// simple access to the correct value
+///
+/// For now, stub out the gateway/mesh logic with just a channel
+/// TODO: if request asks for gateway that does not exist, should it add the gateway, or should
+///       it fail? who is responsible for managing the existence of gateways?
+pub async fn gateway_middleware_fn<B: Send, S: Step>(
```
akoshelev, line 41: this is a step in the right direction - is there a particular reason to inject it as middleware rather than shared state?
https://docs.rs/axum/latest/axum/struct.Extension.html

I would imagine IPA application gateway to exist for all requests - that way we can add backpressure, so it would need to be present as a struct and the entry point from axum handler to IPA app

src/net/server/mod.rs
thurstonsand: ah, that was an old comment. let me remove the TODO

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,186 @@
...
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    Ok(next.run(req).await)
+}
+
+/// accepts all the relevant information from the request, and push all of it onto the gateway
+/// TODO: implement the receiving end of `outgoing`
+pub async fn handler<S: Step>(
+    Extension(outgoing): Extension<mpsc::Sender<BufferedMessages<S>>>,
+    Path(query_id, step): Path<S>,
```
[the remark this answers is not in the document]
thurstonsand: added that in

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,166 @@
...
+
+/// accepts all the relevant information from the request, and push all of it onto the gateway
+#[allow(clippy::unused_async)] // handler is expected to be async
+#[allow(clippy::cast_possible_truncation)] // length of envelopes array known to be less u32
+pub async fn handler(
+    Extension(mut permit): Extension<PollSender<MessageChunks<IPAProtocolStep>>>,
+    Path(_query_id, step): Path<IPAProtocolStep>,
+    Query(identity): Query<Identity>,
+    RecordHeaders { offset, data_size }: RecordHeaders,
+    body: Bytes,
```
akoshelev, line 55: we should eventually replace this with `BodyStream` as this is the mechanism in hyper to provide http2 friendly backpressure (window is adjusted after body future is polled).

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,79 @@
...
+                let req = req_parts
+                    .try_into_request()
+                    .expect("body should not be used");
+                ResponseFuture::Inner(self.inner.call(req))
+            }
+        }
+    }
+}
+
+/// When cloning, the `PollSender` does not transfer any permits it has acquired.
```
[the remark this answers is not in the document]
thurstonsand: there are a few `clone`s that happen before the first `poll_ready` is called, so we do need to make sure `sender` is valid. If you look at my latest push, I changed how this worked to be more explicitly a state machine that is hopefully clearer

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,79 @@
+use super::future::ResponseFuture;
+use axum::extract::RequestParts;
+use axum::http::{Request, Response};
+use std::task::{ready, Context, Poll};
+use tokio::sync::mpsc;
+use tokio_util::sync::PollSender;
+use tower::Service;
+
+/// `poll_ready` will first grab a permit via the `PollSender`. Once ready, the sender is injected
+/// into the Request as an Extension. This can only be done once
```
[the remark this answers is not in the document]
thurstonsand: I believe that wouldn't be too difficult to backport. Instead of a single `Sender` here, you could have a map of `Sender`s keyed on `ChannelId`, which we will be able to extract from the `Request`

src/net/server/handlers/message_stream_layer/future.rs
```diff
@@ -0,0 +1,33 @@
...
+    F: Future<Output = Result<Response<B>, E>>,
+    B: Default,
+{
+    type Output = Result<Response<B>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project() {
+            ProjResp::Inner(f) => Poll::Ready(Ok(ready!(f.poll(cx))?)),
+            ProjResp::PollErr => {
+                let mut res = Response::new(B::default());
```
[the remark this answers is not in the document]
thurstonsand: it's an empty body with the status code. What would be the best approach to fill the body? I don't want to necessarily force a concrete type here

src/net/server/handlers/message_stream_layer/future.rs
```diff
@@ -0,0 +1,33 @@
...
+                *res.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
```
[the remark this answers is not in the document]
thurstonsand: this error will only be possible if it's been forwarded from the `MessageStream`. and in that case, the only time when an error occurs is if the sender has been closed before a call to `poll_reserve`. so I believe that would be a 500

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,79 @@
...
+            // when `call` is called, will respond with error
+            drop(self.sender.take());
+            Poll::Ready(Ok(()))
+        }
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        match self.sender.take() {
+            // error in poll_ready, or `call` called more than once
+            None => ResponseFuture::PollErr,
```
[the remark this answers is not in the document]
thurstonsand: I have an alternative implementation that is hopefully more clear. take a look

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,79 @@
...
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if ready!(self.sender.as_mut().unwrap().poll_reserve(cx)).is_ok() {
+            self.inner.poll_ready(cx)
+        } else {
+            // if poll_reserve returns error, drop the sender since it is no longer usable
+            // when `call` is called, will respond with error
```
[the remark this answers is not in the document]
thurstonsand: axum router enforces `Infallible`

src/net/server/handlers/mul.rs
thurstonsand: hm...unfortunately this will not be easy to do. `BodyStream` does not implement [`HttpBody`](https://docs.rs/hyper/latest/hyper/body/trait.HttpBody.html) which is required for the `Request<T>`'s T (in Axum). it has a `FromRequest` implementation, but we can't use that because we are already using `Request<T>`

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        match self.sender.take() {
+            // error in poll_ready, or `call` called more than once
+            None => ResponseFuture::PollErr,
+            Some(sender) => {
+                let mut req_parts = RequestParts::new(req);
+                req_parts.extensions_mut().insert(sender);
+                let req = req_parts
+                    .try_into_request()
+                    .expect("body should not be used");
```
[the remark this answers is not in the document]
thurstonsand: oh, of course.

I implemented a wrapper around `PollSender`. let me know if that's what you intended.

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+/// When cloning, the `PollSender` does not transfer any permits it has acquired.
+/// Thus, `clone` is equivalent to initializing a new [`UpstreamMiddleware`] with the same `sender`
+impl<S: Clone, T> Clone for MessageStream<S, T> {
+    // manually implement Clone to get around requirements that `T` must be [`Clone`]
+    fn clone(&self) -> Self {
+        // Enforces that `.clone()` is never called after `.poll_ready()` has been called
+        #[cfg(debug_assertions)]
```
akoshelev, line 127: Don't we need this check everywhere? it is cheap

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        match &self.sender {
+            SenderStatus::NotReady(_) => panic!("poll_ready not called"),
+            SenderStatus::Ready(_) => {
+                if let SenderStatus::Ready(sender) =
+                    std::mem::replace(&mut self.sender, SenderStatus::Spent)
+                {
+                    // must insert value wrapped in `Some` due to `Extension`'s `FromRequest` impl
```
akoshelev, line 104: You don't need that because you can just borrow the body (`body_mut`) inside the handler, convert it to owned vec of envelopes and only then borrow the extension mutably and send.

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,186 @@
...
+    gateway_map: GatewayMap<S>,
+    req: Request<B>,
+    next: Next<B>,
+) -> Result<Response, StatusCode> {
+    let mut req_parts = RequestParts::new(req);
+    let Path(query_id, step) = Path::<S>::from_request(&mut req_parts)
+        .await
+        .map_err(|err| err.into_response().status())?;
+
+    let gateway = gateway_map
```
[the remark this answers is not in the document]
thurstonsand: obsolete code

src/net/server/handlers/mul.rs
thurstonsand: obsolete code

src/net/server/mod.rs
```diff
@@ -50,9 +104,11 @@ pub enum BindTarget {
 
 /// Starts a new instance of MPC helper and binds it to a given target.
 /// Returns a socket it is listening to and the join handle of the web server running.
-#[allow(dead_code)]
-pub async fn bind(target: BindTarget) -> (SocketAddr, JoinHandle<()>) {
-    let svc = router()
+pub async fn bind<S: Step>(
```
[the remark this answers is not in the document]
thurstonsand: it is now wrapped in a struct, so this is done

src/cli/net/server/handlers/mul.rs
thurstonsand: obsolete code

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,134 @@
...
+                        std::mem::replace(&mut self.sender, SenderStatus::Spent)
+                    {
+                        self.sender = SenderStatus::Ready(sender);
+                        self.inner.poll_ready(cx)
+                    } else {
+                        panic!("SenderStatus was not NotReady")
+                    }
+                } else {
+                    // if `poll_reserve` returns error, forward error to response
+                    self.sender = SenderStatus::Err;
```
akoshelev, line 88: it would be great if we include what exactly went wrong in the error. Note that in release builds you don't have to return this back to the clients, but in debug mode it is extremely valuable to have. Also, this handling logic (include error description in http body for debug, remove in release) can be done in one place

src/net/server/handlers/message_stream_layer/future.rs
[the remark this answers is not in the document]
akoshelev: Would it be better if we include a generic error description in release and detailed error information (as raw bytes or owned string) in debug build? Later we may want to \implement `Body` trait for IPA errors and cut off debug information in that implementation

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+pub struct ReservedPermit<T>(PollSender<T>);
+impl<T: Send + 'static> ReservedPermit<T> {
+    pub fn send(&mut self, item: T) -> Result<(), MpcServerError> {
+        Ok(self.0.send_item(item)?)
+    }
+}
+
+/// `poll_ready` will first grab a permit via the `PollSender`. Once ready, the sender is injected
```
akoshelev, line 40: lets work out a better name and description for it

src/net/server/handlers/message_stream_layer/future.rs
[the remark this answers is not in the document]
akoshelev: I think we may want to be careful with this assumption. For now, it is true but you tightly coupled `ResponseFuture` to its usage inside `MessageStream`, but there is no enforcement for that rule. If someone uses it in other context or changes how `MessageStream` operates, this assumption may not be true anymore. We may have an easy fix for it which is just to carry the status code inside the error itself

src/net/server/handlers/message_stream_layer/service.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        match &mut self.sender {
+            SenderStatus::NotReady(sender) => {
+                if ready!(sender.poll_reserve(cx)).is_ok() {
+                    // swap `NotReady` for `Ready`
+                    if let SenderStatus::NotReady(sender) =
+                        std::mem::replace(&mut self.sender, SenderStatus::Spent)
+                    {
+                        self.sender = SenderStatus::Ready(sender);
```
akoshelev, line 81: I don't quite get why we need to perform two step change `NotReady` -> `Spent` -> `Ready` here. Can we just go straight from `NotReady` to `Ready`?

src/net/server/handlers/message_stream_layer/future.rs
```diff
@@ -0,0 +1,33 @@
...
+    Inner(#[pin] F),
+    PollErr,
+}
+
+impl<F, B, E> Future for ResponseFuture<F>
+where
+    F: Future<Output = Result<Response<B>, E>>,
+    B: Default,
+{
+    type Output = Result<Response<B>, E>;
```
akoshelev, line 21:
```suggestion
    type Output = F::Output;
```

src/net/client/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    async fn mul(
+        &self,
+        query_id: QueryId,
+        step: UniqueStepId,
+        identity: Identity,
+        offset: u32,
+        data_size: u32,
+        messages: Bytes,
```
akoshelev, line 105: nit: shall we create a wrapper struct for it? this is at the edge of clippy yelling at us for having too many arguments

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,169 @@
...
+/// `permit`. If we try to extract the `permit` via the `Extension`'s `FromRequest` implementation,
+/// it will call `.clone()` on it, which will remove the `OwnedPermit`. Thus, we must access the
+/// `permit` via `Request::extensions_mut`, which returns [`Extensions`] without cloning.
+/// Unfortunately, this also means we must wrap the `permit` in an `Option` such that we can
+/// `take()` the `permit`, freeing up the `req` for `Request::into_body`
+#[allow(clippy::unused_async)] // handler is expected to be async
+#[allow(clippy::cast_possible_truncation)] // length of envelopes array known to be less u32
+pub async fn handler(
+    Path(_query_id, step): Path,
+    Query(IdentityQuery { identity }): Query<IdentityQuery>,
```
akoshelev, line 40: can you please add a todo here to fix this once we figure out discovery and handshake stuff? Trusting client to tell us their identity is a big security hole and we shouldn't forget to fix it later. It is fine imo to have it right now, but would be great to have a list of things we need to fix

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,169 @@
...
+            payload: chunk.to_vec().into_boxed_slice(),
+        })
+        .collect::<Vec<_>>();
+
+    permit.send((channel_id, envelopes))?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
akoshelev, line 70: can you also test that malformed requests are rejected (bad headers,query,etc). Also I wonder if it would be easy to test the backpressure as well? It may be hard to test it end-to-end, but should be fairly straightforward to test the service wrapper

src/net/server/handlers/message_stream_layer/layer.rs
```diff
@@ -0,0 +1,23 @@
+use super::MessageStream;
+use tokio::sync::mpsc;
+use tower::Layer;
+
+/// Wraps a [`MessageStream`] so that it can be used in a [`Layer`]
```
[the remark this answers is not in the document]
thurstonsand: yes. this is a very simple wrapper around the `MessageStream` service

src/net/server/handlers/message_stream_layer/service.rs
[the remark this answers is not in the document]
thurstonsand: we do. we can get the `Request` in `poll_ready`, so we'll have the path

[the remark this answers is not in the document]
thurstonsand: no, just checked, and you're right. we only get request in `call`... potentially we could move back to an implementation that uses a `MiddlewareFn` and move the logic into that. we should be ok to do that since we can leave the body unserialized

src/net/server/handlers/message_stream_layer/service.rs
thurstonsand: sure. removed the cfg

src/net/server/handlers/message_stream_layer/service.rs
thurstonsand: oh good. much better...

src/net/client/mod.rs
thurstonsand: sure, done

src/net/server/handlers/message_stream_layer/service.rs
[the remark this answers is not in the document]
thurstonsand: ok, i deleted the entire `message_stream_layer` folder and implemented an async fn middleware. take a look, it's much simpler and gives us access to `Request`

src/net/server/handlers/mul.rs
thurstonsand: done

src/net/server/handlers/mul.rs
thurstonsand: will work on this. good ideas

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,209 @@
...
+    req.extensions_mut().insert(ReservedPermit::new(permit));
+    Ok(next.run(req).await)
+}
+
+/// extracts the [`MessageChunks`] from the request and forwards it to the Message layer via the
+/// `permit`. If we try to extract the `permit` via the `Extension`'s `FromRequest` implementation,
+/// it will call `.clone()` on it, which will remove the `OwnedPermit`. Thus, we must access the
+/// `permit` via `Request::extensions_mut`, which returns [`Extensions`] without cloning.
+/// Unfortunately, this also means we must wrap the `permit` in an `Option` such that we can
+/// `take()` the `permit`, freeing up the `req` for `Request::into_body`
```
akoshelev, line 75: remove the part about cloning?

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,209 @@
...
+impl<T> Clone for ReservedPermit<T> {
+    /// # Panics
+    /// if a permit exists
+    fn clone(&self) -> Self {
+        assert!(self.0.is_none());
+        Self(None)
+    }
+}
+
+pub async fn message_stream_middleware<T: Send + 'static, B>(
```
akoshelev, line 60: yea that is much better. I would also rename it to something like `obtain_permit_mw` because `message_stream` is not even a trait in the code.

src/net/server/handlers/message_stream_layer/service.rs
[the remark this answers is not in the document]
akoshelev: nice, so we can inject our global state into the middleware later, when we construct the router.

src/net/server/handlers/mul.rs
thurstonsand: It's probably important that we document the messiness with `FromRequest`. where do you recommend I have that? should it just be a regular comment within the fn?

src/net/server/handlers/mul.rs
thurstonsand: sounds good to me

src/net/server/handlers/mul.rs
akoshelev: sorry I meant the part about `Option.take`

thurstonsand: gotcha. removed that part

src/net/server/handlers/message_stream_layer/future.rs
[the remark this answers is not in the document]
thurstonsand: this code has been deleted, but the `MpcServerError` already [implements `IntoResponse`](https://github.com/thurstonsand/raw-ipa/blob/client-server/src/net/server/mod.rs#L74), where we could define a body if we wish. actually, as of now, the body is populated with the full error message

src/net/server/handlers/message_stream_layer/future.rs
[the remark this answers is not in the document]
thurstonsand: yeah, that does make sense. this code has been removed, but will keep this advice in mind

src/net/server/handlers/message_stream_layer/service.rs
thurstonsand: at the time, the only error that could occur was that the sender was closed. so that information was ultimately conveyed. that being said, this code has been removed now

src/net/server/handlers/message_stream_layer/service.rs
thurstonsand: code is now obsolete

src/net/server/handlers/message_stream_layer/service.rs
thurstonsand: code is obsolete

src/net/server/handlers/message_stream_layer/future.rs
thurstonsand: code is obsolete

src/net/server/handlers/mul.rs
thurstonsand: done

src/net/server/handlers/mul.rs
```diff
@@ -192,16 +197,243 @@ mod tests {
...
+        assert_eq!(resp.status(), expected_status);
+    }
+
+    #[tokio::test]
+    #[allow(clippy::mutable_key_type)] // [`HeaderName`] is known good key
+    #[allow(clippy::too_many_lines)] // testing all permutations of request
+    async fn malformed_req_fails() {
+        const MALFORMED_DATA_SIZE: usize = 7;
+        let (port, rx) = init_server().await;
+        tokio::spawn(async move {
```
akoshelev, line 264: I don't think this task does anything useful?

src/net/server/handlers/mul.rs
```diff
@@ -192,16 +197,243 @@ mod tests {
...
+        let resp = Client::default()
+            .request(req)
+            .await
+            .expect("request should complete successfully");
+        assert_eq!(resp.status(), expected_status);
+    }
+
+    #[tokio::test]
+    #[allow(clippy::mutable_key_type)] // [`HeaderName`] is known good key
+    #[allow(clippy::too_many_lines)] // testing all permutations of request
```
akoshelev, line 260: I agree with clippy here - you could split this test into multiple by creating a helper function that prepares a valid request. then each test will replace one part of that request with the malformed entity and execute it

akoshelev: In addition to be more readable, this will also be faster because rust executes tests in parallel

src/net/server/handlers/mul.rs
```diff
@@ -192,16 +197,243 @@ mod tests {
...
+            malformed_body,
+        );
+        expect_res(req, StatusCode::BAD_REQUEST).await;
+    }
+
+    fn poll<F, T>(f: &mut F) -> Poll<T>
+    where
+        F: Future<Output = T> + Unpin,
+    {
+        Pin::new(f).poll(&mut Context::from_waker(futures::task::noop_waker_ref()))
```
akoshelev, line 387: you could use `poll_unpin` here because `F` is constrained to be `Unpin` and there is no need to pin it.

also this function could probably have a more appropriate name like `poll_once` maybe?

src/net/server/handlers/mul.rs
```diff
@@ -192,16 +197,243 @@ mod tests {
...
+                .collect::<Vec<_>>();
+
+            assert_eq!(status, StatusCode::OK, "{}", resp_body_str);
+            let messages = rx.try_recv().expect("should have already received value");
+            assert_eq!(messages, (channel_id, envs));
+        }
+    }
+
+    #[allow(clippy::mutable_key_type)] // `HeaderName` is known good key
+    fn build_malformed_req<Q: Into<String>, S: Into<String>, I: Into<String>, B: Into<Body>>(
```
akoshelev, line 226: this function has the ability to construct any request, not just malformed, so maybe just `post_request`?

src/net/mod.rs
```diff
@@ -57,6 +57,17 @@ impl RecordHeaders {
         req.header(OFFSET_HEADER_NAME.clone(), self.offset)
             .header(DATA_SIZE_HEADER_NAME.clone(), self.data_size)
     }
+
+    /// # Errors
+    /// Returns an error if the body's length does not align with the data-size header.
+    /// Namely, checks if `body.len() % data-size == 0`
+    pub fn matches_body(&self, body_len: usize) -> Result<(), MpcServerError> {
+        (body_len % self.data_size as usize == 0)
```
akoshelev, line 65: please please never trust your inputs to be within the valid range. This will crash the server if `data_size` is 0. We could argue whether we should return $\bot$ or detailed explanation (I strongly advise to be as detailed as possible in debug builds and tell the caller what exactly went wrong) in this case, but everything you receive must be sanitized properly.

martinthomson (commented):
I'm not sure about this in its current form.

Can we break it down a little?

Things that might be able to be built easily:

1. A server that does nothing, except maybe echo requests.
2. A client that can connect and make arbitrary requests to the server (with the caller providing all inputs).
3. URI logic for identifying helper->helper channels.
4. A server that accepts requests for a helper->helper channel and just produces a stream of bytes.  That is, like this patch does, with a query param or header to identify the client.
5. A client that initiates a helper->helper channel and just accepts a stream of bytes.
6. A means of configuring helpers with the identity (including addresses and certificates) for their peers.
7. A way to authenticate where requests are coming from using those certificates so that you can know who to attribute messages to.
8. A way to request that a new MPC circuit be instantiated on each helper.
9. A means of coordinating shared secrets (for PRSS) across helpers.  (Tip: treat this like a channel, but use parallel infrastructure because it will need to send a single small blob of data.)
10. A simpler way to coordinate making an MPC circuit across all helpers.
11. A way to submit inputs for the MPC circuit you created.  Unencrypted.
12. Key configuration for helpers.
13. Decrypting inputs on each helper.


I'm happy enough to land this and iterate, the code is mostly OK, but maybe the next one can be more modest in size.

src/net/server/handlers/mul.rs
thurstonsand: oh, @akoshelev is right. since all the requests are malformed, nothing ever gets pushed onto the channel. i removed the task

src/net/server/handlers/mul.rs
thurstonsand: ok, i came up with a much cleaner version of this test.

src/net/server/handlers/mul.rs
thurstonsand: i think `poll_once` specifically implies that it will consume the future, right? whereas `poll` acts in the same way that `Future::poll` acts.

src/net/server/handlers/mul.rs
thurstonsand: you are right, but I reworked this test, and this fn was removed

src/net/mod.rs
thurstonsand: of course...adding an extra check on `data_size`

src/net/mod.rs
```diff
@@ -57,6 +57,17 @@ impl RecordHeaders {
         req.header(OFFSET_HEADER_NAME.clone(), self.offset)
             .header(DATA_SIZE_HEADER_NAME.clone(), self.data_size)
     }
+
+    /// # Errors
+    /// Returns an error if the body's length does not align with the data-size header.
+    /// Namely, checks if `body.len() % data-size == 0`
+    pub fn matches_body(&self, body_len: usize) -> Result<(), MpcServerError> {
```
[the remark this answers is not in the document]
thurstonsand: sure, i can do that

src/protocol/sort/mod.rs
```diff
@@ -29,6 +31,35 @@ pub enum ShuffleStep {
...
+    const STEP3_STR: &'static str = "step3";
+}
+
+impl Display for ShuffleStep {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ShuffleStep::Step1 => write!(f, "{}", Self::STEP1_STR),
+            ShuffleStep::Step2 => write!(f, "{}", Self::STEP2_STR),
+            ShuffleStep::Step3 => write!(f, "{}", Self::STEP3_STR),
+        }
```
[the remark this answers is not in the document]
thurstonsand: ah, that code is from before the Step changes you made. I just removed it all

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+/// extracts the [`MessageChunks`] from the request and forwards it to the Message layer via the
+/// `permit`. If we try to extract the `permit` via the `Extension`'s `FromRequest` implementation,
+/// it will call `.clone()` on it, which will remove the `OwnedPermit`. Thus, we must access the
+/// `permit` via `Request::extensions_mut`, which returns [`Extensions`] without cloning.
+#[allow(clippy::unused_async)] // handler is expected to be async
+#[allow(clippy::cast_possible_truncation)] // length of envelopes array known to be less u32
+pub async fn handler(
+    Path(_query_id, step): Path,
```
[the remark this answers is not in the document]
thurstonsand: gotcha. I was following axum documentation here, but I have no problem simplifying the fn signature

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,439 @@
...
+    mut req: Request<Body>,
+) -> Result<(), MpcServerError> {
+    // prepare data
+    let channel_id = ChannelId { identity, step };
+
+    let body = hyper::body::to_bytes(req.body_mut()).await?;
+    headers.matches_body(body.len())?;
+    let envelopes = body
+        .as_ref()
+        .chunks(headers.data_size as usize)
```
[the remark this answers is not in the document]
thurstonsand: the assumption i'm making is that it's possible for different requests to have records of different sizes, but that within a request they will all be the same size.

I'm willing to assume a hard-coded size for now. I know some of the work that ben is doing for uplift to malicious might double the size of a record, so we should consider that.

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,439 @@
...
+        .enumerate()
+        .map(|(record_id, chunk)| MessageEnvelope {
+            record_id: RecordId::from(headers.offset + record_id as u32),
+            payload: chunk.to_vec().into_boxed_slice(),
+        })
+        .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
thurstonsand: I'm welcome to an alternative suggestion. What would we send to messaging layer if it's not a vec (which is what is [defined in `MessageChunks` now](https://github.com/private-attribution/ipa/blob/cea96f47a0d3d74e1bf4efdf03e190b70097efe9/src/helpers/fabric.rs#L21). The `.send` necessitates `Send + 'static`

src/net/server/handlers/mul.rs
```diff
@@ -0,0 +1,439 @@
...
+impl<T> Clone for ReservedPermit<T> {
+    /// # Panics
+    /// if a permit exists
+    fn clone(&self) -> Self {
+        assert!(self.0.is_none());
+        Self(None)
+    }
+}
+
+pub async fn obtain_permit_mw<T: Send + 'static, B>(
```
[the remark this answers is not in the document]
thurstonsand: right, sorry I missed this fn. will add docs

src/net/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        (body_len % self.data_size as usize == 0)
+            .then_some(())
+            .ok_or_else(|| {
+                MpcServerError::InvalidHeader("data-size header does not align with body".into())
+            })
```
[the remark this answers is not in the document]
thurstonsand: reworked it to be more clear. I also took Alex's advice and moved this logic to the `from_request` fn

src/net/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+/// # `data_size`
+/// the batch will be transmitted as a single `Bytes` block, and the receiver will need to know how
+/// to divide up the block into individual records. `data_size` represents the number of bytes each
+/// record consists of
+pub struct RecordHeaders {
+    offset: u32,
+    data_size: u32,
+}
+
+impl RecordHeaders {
```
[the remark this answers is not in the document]
thurstonsand: Might be good to have a quick design discussion on this topic. As I understood the networking/messaging layer interface, the network layer will receive streams in discrete chunks that start and end at a some arbitrary record. so for instance, messaging layer might have 2 Channels of data to send to a different helper:

H1, step1 -> records[0..100]
H2, step2 -> records[0..100]

and it might decide to split those up into 20-record chunks. so it would pass those chunks to the network layer, which would make unique POST reqs of contiguous bytes representing:

H1, step1 -> records[0,20]
H2, step1 -> records[0,20]
H1, step1 -> records[20,40]
H2, step1 -> records[20,40]
etc...

the receiving helpers' network layers would receive those contiguous chunks, break them up into [`MessageEnvelope`s](https://github.com/private-attribution/ipa/blob/cea96f47a0d3d74e1bf4efdf03e190b70097efe9/src/helpers/fabric.rs#L16), and pass them up to their messaging layers.

It sounds like you're considering a different solution. one that would have a single POST stream all records through it?

src/net/mod.rs
```diff
@@ -0,0 +1,100 @@
+// need this to allow MessageEnvelope to have custom serde bounds
```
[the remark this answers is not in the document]
thurstonsand: I can make a GH issue for that and follow up with a PR

src/protocol/mod.rs
```diff
@@ -128,6 +138,25 @@ impl AsRef<str> for UniqueStepId {
     }
 }
 
+impl From<String> for UniqueStepId {
+    fn from(id: String) -> Self {
+        let id = id
+            .strip_prefix('/')
+            .map_or_else(|| id.clone(), str::to_owned);
```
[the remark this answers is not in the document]
thurstonsand: alright. I can make a GH issue and follow up in a later PR

src/net/client/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        Ok(result.to_vec())
+    }
+
+    async fn mul(&self, args: HttpMulArgs) -> Result<(), MpcClientError> {
+        let uri = self.build_uri(format!(
+            "/mul/query-id/{}/step/{}?identity={}",
+            String::from(args.query_id),
+            String::from(args.step.clone()),
+            String::from(args.identity),
```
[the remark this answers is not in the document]
thurstonsand: yes, `as_ref()` makes a lot of more sense. done

thurstonsand:
I will definitely target a smaller PR next time

src/net/server/handlers/mul.rs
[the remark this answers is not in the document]
martinthomson: Also add it to the type above, I think.  It doesn't need to float free.

src/net/server/handlers/mul.rs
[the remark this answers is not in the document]
martinthomson: Hmm, yeah, that is awkward.  Perhaps we should track this in an issue.

@akoshelev do you think that an interface to the network that is more along the lines of `send(&self, channel: ChannelId) -> Stream<Message>` would be acceptable?  (The same for receiving.)  That might be less work for Thurston to manage even as many networking libraries try hard to make stream-level interactions independent of each other.

src/net/mod.rs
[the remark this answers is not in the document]
martinthomson: At the networking layer, what I would expect is that each stream is a contiguous, undifferentiated sequence of records, starting from index 0 and increasing monotonically without gaps.  Different streams will be used for records from different steps or to/from different helpers.  Though your interface will accept records in chunks, you should be able to assert that each chunk you receive for a stream is contiguous with the previous one.  You shouldn't care - except to the extent that it affects performance - how big those chunks are.

martinthomson (approved):

src/net/server/handlers/mul.rs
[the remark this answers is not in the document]
akoshelev: yea that was the alternative I considered when working on #130. I thought it might be easier for us to implement it this way, but I am not so sure right now. The thing that tripped me off was the fact that infra cannot pull streams it never received requests to receive in this case. I also wasn't sure whether I can make it efficient by letting infra to guess which channel is ready. We could probably mitigate that by adding another method to `Network` that can hint which stream has the stream of bytes ready. 

any thoughts?

src/cli/net/mod.rs
thurstonsand: https://github.com/private-attribution/ipa/issues/166

src/net/mod.rs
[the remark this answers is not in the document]
thurstonsand: https://github.com/private-attribution/ipa/issues/167

src/protocol/mod.rs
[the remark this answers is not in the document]
thurstonsand: given we're moving away from axum, I'm not too sure how any new framework will act. with your suggestion on switching to `as_ref`, i got rid of all the `clone`s in this impl. given this `strip_prefix` will only need to happen once, in the implementation of `From<&str> for UniqueStepId`, should we leave it as-is for now?

thurstonsand:
@martinthomson if it makes sense to you, the plan is to move forward with this implementation that does the discrete request/response model I described here: https://github.com/private-attribution/ipa/pull/94#discussion_r1007300375 and focus on getting an end-to-end solution for now. Then, we can revisit your proposed streaming solution later. I've captured some thoughts and discussions @akoshelev and I had about that in an issue here: https://github.com/private-attribution/ipa/issues/181

akoshelev (approved):

martinthomson:
To be clear, I don't like that model, but if the goal is to work toward something, I can live with making progress.

akoshelev:
@martinthomson just want to make sure we all have correct understanding of the ideal model of communication for IPA. Does #181 proposal captures it? Stream all records for every step (i.e. HTTP request body is a stream) directly to the infrastructure layer

martinthomson:
Pretty much.  Step "/foo/bar" will post to `/q/<queryid>/foo/bar` and send all of the records in the body.  No reason to break that up into multiple requests or chunks.

akoshelev:
landing this to unblock the remaining work

PR #95, opened 2022-09-22
[Semi honest sort] reshare(i, [x])

richajaindce:
This implements reshare algorithm of "Efficient Secure Three-Party Sorting Protocol with an Honest Majority" at communication cost of 2R.
Input: Pi-1 and Pi+1 knows their secret shares
Output: At the end of the protocol, [x] receive renewed shares of x across all 3 helpers with randomness added

Screenshot 2022-09-22 at 11 05 10 PM


src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,146 @@
...
+
+            // Sleep until `to_helper.right` sends us their part1 value
+
+            let DValue { d: part1 }: DValue<F> = channel
+                .receive(channel.identity().peer(Direction::Right), self.record_id)
+                .await?;
+
+            Ok(Replicated::new(part1 + part2, r0))
+        } else {
+            Ok(Replicated::new(r1, r0))
```
akoshelev, line 102: I don't understand why it is (r1, r0) and not the opposite. It does not match what is shown on the picture in the description - it seems everything is just reversed, what is the reason for that?

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,146 @@
...
+                    to_helper.peer(Direction::Right),
+                    self.record_id,
+                    DValue { d: part1 },
+                )
+                .await?;
+
+            // Sleep until `to_helper.left` sends us their part2 value
+
+            let DValue { d: part2 } = channel
+                .receive(channel.identity().peer(Direction::Left), self.record_id)
```
akoshelev, line 79: @thurstonsand we need to decide whether we should have a generic API to send and receive
`api/query/../step/../` or specialized API for multiplication/resharing/etc. both have pros and cons

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,146 @@
...
+            channel
+                .send(
+                    to_helper.peer(Direction::Right),
+                    self.record_id,
+                    DValue { d: part1 },
+                )
+                .await?;
+
+            // Sleep until `to_helper.left` sends us their part2 value
+
```
akoshelev, line 77:
```suggestion

```

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,146 @@
...
+            channel
+                .send(
+                    to_helper.peer(Direction::Left),
+                    self.record_id,
+                    DValue { d: part2 },
+                )
+                .await?;
+
+            // Sleep until `to_helper.right` sends us their part1 value
+
```
akoshelev, line 95:
```suggestion

```

src/protocol/sort/reshare.rs
richajaindce: I got messed up while writing this code. (The picture description is correct). Surprisingly validate_and_reconstruct did not catch this as well. Have added an extra check in that function to ensure Replicated secret shares are same across helpers.

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,146 @@
...
+        let inputs = input.as_tuple();
+        // `to_helper.left` calculates part1 = (input.0 + input.1) - r1 and sends part1 to `to_helper.right`
+        if channel.identity() == to_helper.peer(Direction::Left) {
+            let part1 = inputs.0 + inputs.1 - r1;
+            channel
+                .send(
+                    to_helper.peer(Direction::Right),
+                    self.record_id,
+                    DValue { d: part1 },
```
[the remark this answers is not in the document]
richajaindce: What is the purpose of DValue here? is it for handling backpressure later on? Is it ok to reuse the same name or would it be better to choose a new name for each sub-protocol?

src/protocol/sort/reshare.rs
akoshelev: I think it worked because **everything** was reversed so math was the same

src/protocol/sort/reshare.rs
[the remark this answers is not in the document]
akoshelev: `DValue` is multiplication-specific message type that indicates helper's intent to send its multiplication intermediate value to peer. We can invent another message type for reshare - that is better because allows us not to confuse two different intents (to multiply and to reshare)

src/protocol/sort/reshare.rs
richajaindce: Everything was reversed BUT the replicated secret share construct i.e. P1.left = P3.right etc was not being met. Hence added that check. Such things can become very easy to mess up

src/protocol/sort/reshare.rs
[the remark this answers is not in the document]
richajaindce: sounds good. Renaming this to PartValue

src/protocol/context.rs
```diff
@@ -20,6 +20,10 @@ impl<'a, G, S: Step + SpaceIndex> ProtocolContext<'a, G, S> {
         }
     }
 
+    pub fn reshare(&'a self, record_id: RecordId, step: S) -> Reshare<'a, G, S> {
```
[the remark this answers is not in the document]
richajaindce: Based on our discussion today, I am going to move out reshare form the context. As a template for all the protocols, inputs will go in the struct and context, step, etc will go in the execute function (to maintain consistency in code)

src/protocol/sort/reshare.rs
thurstonsand: do you have an example of how such APIs would differ, from the client/server perspective? or would they all basically act like "send buffer of data to helper X" and "receive data from helper Y and forward to protocol layer", with no further complexity?

src/protocol/context.rs
```diff
@@ -20,6 +20,10 @@ impl<'a, G, S: Step + SpaceIndex> ProtocolContext<'a, G, S> {
         }
     }
 
+    // pub fn reshare(&'a self, record_id: RecordId, step: S) -> Reshare<'a, G, S> {
+    //     Reshare::new(&self.participant[step], self.gateway, record_id, step)
+    // }
+
```
akoshelev, line 26:
```suggestion
    
```

akoshelev (approved):

PR #97, opened 2022-09-29
[Shuffle] Part1 : Building blocks needed for implementing shuffle

richajaindce:
This PR contains building blocks for implementing Shuffle.

This includes primitive functions i.e. 
1. apply, 
2. apply_inv and 
3. algorithm to generate a random permutation of a given sequence


src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,46 @@
+use embed_doc_image::embed_doc_image;
+/// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
+/// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
+#[embed_doc_image("apply", "images/sort/apply.png")]
+#[allow(dead_code)]
+pub fn apply<T: Copy + Default>(sigma: &[usize], values: &[T]) -> Vec<T> {
```
akoshelev, line 6: sigma is more scientific notation, in this context this seems to be permutation source indices and permutation dest indices for `apply_inv`

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,46 @@
...
+    let mut applied_vec: Vec<T> = (0..sigma.len()).map(|_| Default::default()).collect();
```
akoshelev, line 7: can we do it in place by taking `&mut [T]`

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
+use crate::helpers::prss::PrssSpace;
+
+#[allow(dead_code)]
+
```
akoshelev, line 4:
```suggestion

```

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// drawing an element from the hat until no elements remain. The algorithm produces an unbiased
+    /// permutation: every permutation is equally likely. This algorithm takes time proportional to
+    /// the number of items being shuffled and shuffles them in place.
+    #[allow(dead_code)]
+    fn generate_random_sequence(
+        batchsize: usize,
+        random_value: u128,
+        left_permute: bool,
+        prss: &PrssSpace,
+    ) -> Vec<usize> {
```
akoshelev, line 20: I am pretty sure we can make it efficient by returning an iterator instead as there is no input array that must be shuffled, we basically generate a random permutation of (1, N)

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+        let mut permutation: Vec<usize> = (0..batchsize).collect();
+        (1..batchsize).rev().for_each(|i| {
+            let location = if left_permute {
+                (prss.generate_values(random_value).0 as usize) % i
```
akoshelev, line 24: there is no reason to call this in a loop, it returns the same value over and over. Moreover this call is compute-heavy as it uses a block cipher to encrypt the index. 

But the bigger question is whether it is secure. This proposal uses a single pseudo-random value to derive the whole permutation, so we are not drawing from a "hat" n times

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// The Fisher–Yates shuffle is an algorithm for generating a random permutation of a finite
+    /// sequence—in plain terms, the algorithm shuffles the sequence. The algorithm effectively
+    /// puts all the elements into a hat; it continually determines the next element by randomly
+    /// drawing an element from the hat until no elements remain. The algorithm produces an unbiased
+    /// permutation: every permutation is equally likely. This algorithm takes time proportional to
+    /// the number of items being shuffled and shuffles them in place.
+    #[allow(dead_code)]
+    fn generate_random_sequence(
+        batchsize: usize,
+        random_value: u128,
```
akoshelev, line 17: lets be consistent and call it `index`

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
+use crate::helpers::prss::PrssSpace;
+
+#[allow(dead_code)]
+
+pub struct Shuffle {}
```
akoshelev, line 5: do we expect state to be added later or we can just replace this struct by a single function that implements FY shuffle? btw, I think Rust std does have an implementation of FY shuffle

src/protocol/sort/apply.rs
richajaindce: Thanks for suggesting a better name!

src/protocol/sort/shuffle.rs
richajaindce: Great to know.. I will use the Rust std library instead. Yes this struct will be filled up in the next diff

src/protocol/sort/shuffle.rs
richajaindce: I have replaced everything with a call to shuffle crate which returns the permutation. However, that makes it not return an iterator but an array of permutation.

src/protocol/sort/shuffle.rs
richajaindce: Replacing it completely with a rust library function.

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
+use embed_doc_image::embed_doc_image;
+/// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
+/// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
+#[embed_doc_image("apply", "images/sort/apply.png")]
+#[allow(dead_code)]
+pub fn apply<T: Copy + Default>(source_indices: &[usize], values: &[T]) -> Vec<T> {
```
benjaminsavage, line 6: I'd suggest renaming "source_indices" to "permutation"

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut applied_vec: Vec<T> = (0..source_indices.len())
+        .map(|_| Default::default())
+        .collect();
+    (0..source_indices.len()).for_each(|i| {
+        let idx = source_indices[i];
+        applied_vec[i] = values[idx];
+    });
```
benjaminsavage, line 13: Can this be combined into just one loop? This appears to be looping over the permutation just to assign a default value, then looping a second time to assign things.

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+    applied_vec
+}
+
+/// To compute `apply_inv` on values, destination_indices(i) can be regarded as the destination of i, i.e., the i-th item
+/// is moved by `apply_inv` to be the σ(i)-th item. Therefore, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is
+/// reordered into (D, C, A, B).
+#[embed_doc_image("apply_inv", "images/sort/apply_inv.png")]
+#[allow(clippy::module_name_repetitions, dead_code)]
+pub fn apply_inv<T: Copy + Default>(destination_indices: &[usize], values: &[T]) -> Vec<T> {
```
benjaminsavage, line 22: Again, I'd recommend renaming "destination_indices" to "permutation"

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut applied_vec: Vec<T> = (0..destination_indices.len())
+        .map(|_| Default::default())
+        .collect();
+    (0..destination_indices.len()).for_each(|i| {
+        let idx = destination_indices[i];
+        applied_vec[idx] = values[i];
+    });
```
benjaminsavage, line 29: Same question about: "Can this be combined into just one operation?"

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+    applied_vec
+}
+
+/// Applying inverse using heapsort which applies destination in-place and in a fixed time
+#[allow(clippy::module_name_repetitions, dead_code)]
+pub fn apply_inv_inplace<T: Copy + Default>(destination_indices: &mut [usize], values: &mut [T]) {
```
benjaminsavage, line 35: It occurs to me that in all of these methods, we might not want to use "usize". According to the [docs](https://doc.rust-lang.org/std/primitive.usize.html), this is 32-bit or 64-bit depending on the architecture we are using. In our case, if we are sorting less than 2^32 elements (over 4 billion) 32-bits is sufficient. We probably never need a 64-bit number. I imagine we will run this code on modern boxes that support 64-bit pointers though. That seems like a big waste.

It would even be cool to use a u16 if you're sorting less than 65,000 items - but that's an optimisation we can probably ignore.

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+        .map(|_| Default::default())
+        .collect();
+    (0..destination_indices.len()).for_each(|i| {
+        let idx = destination_indices[i];
+        applied_vec[idx] = values[i];
+    });
+    applied_vec
+}
+
+/// Applying inverse using heapsort which applies destination in-place and in a fixed time
```
benjaminsavage, line 33: Can we similarly link out to some kind of documentation that explains the algorithm going on here?

src/protocol/sort/apply.rs
```diff
@@ -0,0 +1,103 @@
...
+        let output = apply(&indices, &values);
+        assert_eq!(output, expected_output_apply);
+
+        let expected_output_apply_inv = ["D", "C", "A", "B"];
+        let output = apply_inv(&indices, &values);
+        assert_eq!(output, expected_output_apply_inv);
+    }
+
+    #[test]
+    fn test_inplace() {
```
benjaminsavage, line 96: I think we need more tests for this one, to push the limits a bit. Specifically I'd recommend unit tests for:
- Sorting a number of elements that's not a power of 2 (e.g. 7 elements)
- A few layers deeper heap, e.g. 32 elements

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+#[allow(dead_code)]
+pub struct Shuffle {}
+
+impl Shuffle {
+    /// The Fisher–Yates shuffle is an algorithm for generating a random permutation of a finite
+    /// sequence—in plain terms, the algorithm shuffles the sequence. The algorithm effectively
+    /// puts all the elements into a hat; it continually determines the next element by randomly
+    /// drawing an element from the hat until no elements remain. The algorithm produces an unbiased
+    /// permutation: every permutation is equally likely. This algorithm takes time proportional to
+    /// the number of items being shuffled and shuffles them in place.
```
benjaminsavage, line 14: I'd recommend linking to the Wikipedia page from which you're taking this text: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[allow(clippy::cast_possible_truncation, dead_code)]
+    fn generate_random_sequence(
```
benjaminsavage, line 16: I'd recommend naming it "..._permutation" instead of "..._sequence"

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+        prss: &PrssSpace,
+    ) -> Vec<usize> {
+        let mut permutation: Vec<usize> = (0..batchsize).collect();
+        let rand = prss.generate_values(index.into());
+        let mut rng = if left_permute {
+            StepRng::new(rand.0 as u64, 1)
+        } else {
+            StepRng::new(rand.1 as u64, 1)
+        };
+        permutation.shuffle(&mut rng);
```
benjaminsavage, line 29: Is this really an implementation of Fisher-Yates?

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,99 @@
...
+            batchsize,
+            random_value,
+            false,
+            &p3[TestStep::Shuffle],
+        );
+
+        let expected_numbers: Vec<usize> = (0..batchsize).collect();
+        assert_eq!(sequence1right, sequence2left);
+        assert_eq!(sequence2right, sequence3left);
+        assert_eq!(sequence3right, sequence1left);
```
benjaminsavage, line 89: I'd suggest also checking that the sequences we *don't* expect to be equal are indeed not equal. If the batchsize is large enough, the random chance of them being equal should be very small.

src/protocol/sort/apply.rs
richajaindce: Replacing this function by a rust crate permutation : https://docs.rs/permutation/latest/permutation/permutation/struct.Permutation.html
it has functions to apply permutation and inverse in place. However it uses cycle decomposition (hence can be open to a potential timely attack). Confirming on that with Ryo.

src/protocol/sort/apply.rs
richajaindce: we basically wanted to initialize a vector before filling it. For now replacing this function with https://docs.rs/permutation/latest/permutation/permutation/struct.Permutation.html

src/protocol/sort/apply.rs
richajaindce: Removing this for now, will improve this based on Ryo's response

src/protocol/sort/apply.rs
richajaindce: makes sense.. Will include this if we go ahead with heap approach

src/protocol/sort/apply.rs
richajaindce: Let me change this to Permutation uses a usize to keep indices. For now, I will prefer to use existing libraries to implement these functions but leaving a TODO to change this to use a u32 and/or u16 later as an optimization.

src/test_fixture/world.rs
```diff
@@ -42,20 +43,22 @@ impl Debug for TestStep {
             TestStep::Mul1(v) => write!(f, "TestStep/Mul1[{}]", v),
             TestStep::Mul2 => write!(f, "TestStep/Mul2"),
             TestStep::Reshare(v) => write!(f, "TestStep/Reshare[{}]", v),
+            TestStep::Shuffle => write!(f, "TestStep/Shuffle"),
```
richajaindce, line 46: This will move to some other place based on review feedback. Just leaving it as a placeholder code.

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,100 @@
...
+        prss: &PrssSpace,
+    ) -> Vec<usize> {
+        let mut permutation: Vec<usize> = (0..batchsize).collect();
+        let rand = prss.generate_values(index.into());
+        let mut rng = if left_permute {
+            StepRng::new(rand.0 as u64, 1)
+        } else {
+            StepRng::new(rand.1 as u64, 1)
+        };
+        permutation.shuffle(&mut rng);
```
richajaindce, line 24: The test seems to hang for 5-10secs while executing this function of  permutation.shuffle! Not sure whats causing this. Do you think using a better random generator might help here? If so, which one?
@akoshelev @thurstonsand @taikiy

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,100 @@
...
+#[allow(dead_code)]
+pub struct Shuffle {}
+
+impl Shuffle {
+    /// This implements Fisher Yates shuffle described here <https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle>
+    #[allow(clippy::cast_possible_truncation, dead_code)]
+    fn generate_random_permutation(
+        batchsize: usize,
+        index: u64,
+        left_permute: bool,
```
benjaminsavage, line 14: I left this comment in another diff, but I should have left it here. I don't like using a boolean for this, because when you're reading the call site you don't know what the "true" or "false" you're seeing means. Also, the name "left_permute" means nothing.

I'd prefer an enum argument with only two options, like: "LEFT_SHARED_RANDOMNESS" and "RIGHT_SHARED_RANDOMNESS"

Or perhaps make two functions with slightly different names, like:

generate_random_permutation_left_randomness()
generate_random_permutation_right_randomness()

src/protocol/sort/shuffle.rs
benjaminsavage: That's concerning. Randomly shuffling 10000 elements shouldn't take 5-10s.

benjaminsavage (approved):
This looks good to me! I just have one small nit about the "left_permute" argument.

As for the efficiency concern with "shuffle!" it's hard to say if this is some artefact of the test runner, or if its a real thing. Either way, I'd like to land this. It's a good foundation to keep building upon. We can change the implementation if needs be.

src/protocol/sort/shuffle.rs
richajaindce: I actually made the changes in the other diff to use your 1st suggestion of enum. We have an enum named Direction which is Left or Right. I have used that instead. Link : https://github.com/martinthomson/raw-ipa/pull/98/files#diff-894d6d54e26cbe3155f95c21f6bd6b557778f2705f8371fc2bc1d31b2292cb74R45
let me incorporate this in the relevant code here

src/protocol/sort/shuffle.rs
akoshelev: [`StepRng`](https://docs.rs/rand/latest/rand/rngs/mock/struct.StepRng.html) is not intended to be used in prod code, it's main purpose is to provide predictable rng for testing purpose.

the issue you're observing is caused by uniform sampler - because `StepRng` returns back numbers that are not uniformly distributed across the range, it causes more rejections by the uniform sampler. [this loop](https://docs.rs/rand/latest/src/rand/distributions/uniform.rs.html#547-553) is the hot spot.

I am curious whether we can let each helper to generate random permutation without any coordination with other helpers? In this case using a crypto rng here should suffice

richajaindce: We need the seed to be the same for the peers so that they generate the exact same permutation. This would need prss

akoshelev: can you generate a seed using prss and then initialize crypto rng with that seed?

src/protocol/sort/reshare.rs
```diff
@@ -67,14 +68,15 @@ impl<F: Field> Reshare<F> {
                 )
                 .await?;
 
-            // Sleep until `to_helper.left` sends us their part2 value
+            // Sleep until `to_helper.right` sends us their part2 value
             let PartValue { part: part2 } = channel
-                .receive(channel.identity().peer(Direction::Left), record_id)
+                .receive(to_helper.peer(Direction::Right), record_id)
```
benjaminsavage, line 73: Thank you =)

src/protocol/sort/shuffle.rs
benjaminsavage: I think that should work.

On line 21 would it just be:
    CryptoRng::new(rand.0 as u64, 1)

And on line 23 just:
    CryptoRng::new(rand.1 as u64, 1)

akoshelev: yep just want remark that we should use all 16 bytes of randomness provided by PRSS to initialize it (we talked with @richajaindce about it - PRSS uses AES256 internally so ideally we should be using 32 bytes, but that would require adding more methods to PRSS itself) - when @martinthomson is back, we can get some clarity on whether this approach is viable or not (using CryptoRng seeded by PRSS)

martinthomson: So it seems like the PRNG is drawn on pretty heavily here, which means that any quality issues will turn into performance issues.  Fisher-Yates draws progressively smaller random values, which means that you draw at least n-1 times, starting at a draw of n, followed by n-1, n-2, etc...  The rand implementation uses rejection sampling, meaning that when you get to 2^x-1 you have a 50% chance of a rejection and another draw.

With `StepRng`, you are stepping one at a time through the entire output space, starting from the initial value.  Thus, once you exceed the current value, you need to go right around and then back to the start.  Little surprise then that this is slow as it takes almost 2^64 operations for each draw.

The PRSS implementation uses modulo.  This might be faster for this use case (I don't know).  It will at least be more deterministic, though it might not be constant time.

Seeding AES-256 requires 32 bytes, but you should look to do that from the shared entropy (i.e., spawn a new space) rather than seeding from its own output.

richajaindce: I was thinking of using a Chacha8Rng which needs a seed of size [u8;32]. However, PRSS gives a seed size u128 (equivalent to [u8;16]. In this case, 
1. we can copy [u8;16] twice or 
2. generate two randoms from prss (second random generated from the returned index hence not enough entropy) and append them
I am implementing first approach for now. What do you think?

richajaindce: Chach8Rng expects a [u8;32] seed. This is also the recommended size from the docs. Since PRSS is currently giving [u8;16], we can do following
1. Use u8;16 twice
2. have two PrssSpace to give us two random seeds.
For 2nd, ideally we should have a Space which returns a [u8;32] (which internally instantiates two spaces). Which approach are we comfortable taking? @martinthomson @akoshelev

src/protocol/sort/reshare.rs
richajaindce: Apologies for the confusion caused :/

src/protocol/sort/shuffle.rs
richajaindce: Also implemented 1st approach since that looks the cleanest for now. Based on discussion with @akoshelev there is an option of creating fake steps for protocols which need to use prss to generate seeds but will make the code un-necessarily complicated

benjaminsavage (approved):
Thanks @richajaindce for the great work!

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl Shuffle {
+    /// This implements Fisher Yates shuffle described here <https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle>
+    #[allow(dead_code)]
+    fn generate_random_permutation(
+        batchsize: usize,
+        direction: Direction,
+        index: u128,
+        prss: &PrssSpace,
+    ) -> Vec<usize> {
```
akoshelev, line 18: It is a little concerning that `Permutation` crate requirements start to influence our code design decisions here - the fact that we return `Vec<usize>` is mainly driven by `Permutation::from_vec` reqs right? It would be great if we could address this soon

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,113 @@
...
+        let mut permutation: Vec<usize> = (0..batchsize).collect();
+        let seed = prss.generate_values(index);
+
+        let seed_bytes = if direction == Direction::Left {
+            seed.0.to_le_bytes()
+        } else {
+            seed.1.to_le_bytes()
+        };
+        // Chacha8Rng expects a [u8;32] seed whereas prss returns a u128 number.
+        // We are using two spaces to generate indexes and concatenating them
```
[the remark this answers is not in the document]
richajaindce: Addressed this in #103

src/protocol/sort/shuffle.rs
richajaindce: Thats right. However, this function is returns a permutation in the range of usize. It doesn't have anything to do with Field. What do you think should be the intended behaviour here? @akoshelev

PR #98, opened 2022-10-03
[Shuffle] Part 2 : Shuffle implementation

richajaindce:
Executes shuffling the inputs within mpc such that the output is indistinguishable from the input to any of the helpers.


src/protocol/sort/bit_permutations.rs
```diff
@@ -2,19 +2,20 @@ use crate::{
...
+use embed_doc_image::embed_doc_image;
 use futures::future::try_join_all;
 
+use super::SortStep;
 #[derive(Debug)]
 pub struct BitPermutations<'a, F> {
     input: &'a [Replicated<F>],
 }
 
+/// Create an object to generate bit permutations for a given bit column of query. This is GENBITPERM(Algorithm 3) from the paper
```
benjaminsavage, line 17: Future readers of your code will not know what "the paper" refers to. And are you sure it's still "Algorithm 3" is the new one?

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've reshared their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct PartValue<F> {
+    part: F,
+}
+/// Reshare(i, [x])
+// This implements reshare algorithm of "Efficient Secure Three-Party Sorting Protocol with an Honest Majority" at communication cost of 2R.
+// Input: Pi-1 and Pi+1 knows their secret shares
```
benjaminsavage, line 22: "knows" should not be plural

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+// Output: At the end of the protocol, [all 3 helpers receive renewed shares of x with randomness added
```
benjaminsavage, line 23: spurious "[", please remove

benjaminsavage: Also, I'm not sure what you mean by "with randomness added". That sounds like we added a random number to the secret-shared value, which isn't what's going on. I'd say: "...all 3 helpers receive their shares of a new, random secret sharing of the secret value".

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl<F: Field> Reshare<F> {
+    #[allow(dead_code)]
+    pub fn new(input: Replicated<F>) -> Self {
+        Self { input }
+    }
+
+    /// Steps
+    #[embed_doc_image("reshare", "images/sort/reshare.png")]
+    /// 1. While calculating for helper, we call PRSS to get randoms for helper who needs to receive reshares (say `rand_left`, `rand_right`)
```
benjaminsavage, line 37: I'd write out what PRSS means at least once, and replace the phrase "get randoms for helper" with "get random values that will match the random values generated by the other helpers" or similar.

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+    ///    `to_helper.left` knows `rand_left` (named r1) and `to_helper.right` knows `rand_right` (named r0)
+    /// 2. `to_helper.left` calculates part1 = (input.0 + input.1) - r1
+    ///    `to_helper.right` calculates part2 = (input.0 - r0)
```
benjaminsavage, line 40: based on the image, I assume you mean input.2 and not input.0

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+                .send(
+                    to_helper.peer(Direction::Right),
+                    record_id,
+                    PartValue { part: part1 },
+                )
+                .await?;
+
+            // Sleep until `to_helper.left` sends us their part2 value
+            let PartValue { part: part2 } = channel
+                .receive(channel.identity().peer(Direction::Left), record_id)
```
benjaminsavage, line 72: Wait a minute... don't you mean RIGHT? In this fork of the code, we *are* the "left" and we are now waiting to get part2 from the *right*, correct?

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+                .await?;
+
+            Ok(Replicated::new(part1 + part2, r1))
+        } else if channel.identity() == to_helper.peer(Direction::Right) {
+            //  `to_helper.right` calculates part2 = (input.0 - r0) and sends it to `to_helper.left`
+            let part2 = inputs.0 - r0;
```
benjaminsavage, line 78: Pretty sure you mean "input.2" on both the commend and the next line.

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+                    to_helper.peer(Direction::Left),
+                    record_id,
+                    PartValue { part: part2 },
+                )
+                .await?;
+
+            // Sleep until `to_helper.right` sends us their part1 value
+            let PartValue::<F> { part: part1 } = channel
+                .receive(channel.identity().peer(Direction::Right), record_id)
+                .await?;
```
benjaminsavage, line 90: We *are* the right in this fork... aren't we waiting for the *left* to send us part1?

src/protocol/sort/reshare.rs
```diff
@@ -0,0 +1,138 @@
...
+        let reshare1 = Reshare::new(share[1]);
+        let reshare2 = Reshare::new(share[2]);
+
+        let h0_future = reshare0.execute(&context[0], record_id, step, Identity::H2);
+        let h1_future = reshare1.execute(&context[1], record_id, step, Identity::H2);
+        let h2_future = reshare2.execute(&context[2], record_id, step, Identity::H2);
+
+        let f = try_join!(h0_future, h1_future, h2_future).unwrap();
+        let output_share = validate_and_reconstruct(f);
+        assert_eq!(output_share, input);
```
benjaminsavage, line 136: Where do we check that this is a different sharing than the original sharing?

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,280 @@
...
+    ShuffleStep::{self, Step1, Step2, Step3},
+};
+
+#[allow(dead_code)]
+pub struct Shuffle<F, S> {
+    input: Vec<Replicated<F>>,
+    step_fn: fn(ShuffleStep) -> S,
+}
+
+/// This is SHUFFLE(Algorithm 1) described in <https://eprint.iacr.org/2019/695.pdf>.
```
benjaminsavage, line 32: Do we want to update the linked paper to the new one?

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,280 @@
...
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let step = (self.step_fn)(which_step);
+        let prss = &ctx.participant[step];
+        let channel = ctx.gateway.get_channel(step);
+
+        if to_helper != channel.identity() {
+            let mut permute = if to_helper.peer(Direction::Left) == channel.identity() {
+                self.generate_random_sequences(true, prss, seed)
+            } else {
+                self.generate_random_sequences(false, prss, seed)
```
benjaminsavage, line 122: I do not like using a boolean for this first argument. Can we use either an enum value, or make two different functions? From the caller's perspective, reading this code it's not clear what this bool means. I have to consult the function definition.

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,280 @@
...
+            };
+            apply_inv_inplace(&mut permute, &mut self.input);
+        }
+        self.reshare_all_shares(ctx, to_helper, step).await
+    }
+
+    /// Executes shuffling the inputs within mpc.
+    /// Shuffle calls `single_shuffle` three times with 2/3 helpers shuffling the shares each time.
+    /// Since each shuffle needs to be unique for the infrastructure to work, we have three steps to indicate that.
+    /// Due to this, the Shuffle object receives a step function and appends a `ShuffleStep` to form a concrete step.
```
benjaminsavage, line 132: This comment doesn't make sense to me. Can you rephrase this more clearly?

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,280 @@
...
+            shares2.push(share[2]);
+        });
+
+        let mut sh0 = Shuffle::new(shares0, TestStep::Shuffle);
+        let mut sh1 = Shuffle::new(shares1, TestStep::Shuffle);
+        let mut sh2 = Shuffle::new(shares2, TestStep::Shuffle);
+
+        let shuffle0 = sh0.execute(&context[0]);
+        let shuffle1 = sh1.execute(&context[1]);
+        let shuffle2 = sh2.execute(&context[2]);
```
benjaminsavage, line 264: These variable names could be better. What's the difference between sh0 and shuffle0? Can you please use more descriptive names?

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,280 @@
...
+        assert_eq!(result.1.len(), input_len);
+        assert_eq!(result.2.len(), input_len);
+
+        let mut output = HashSet::new();
+        (0..result.0.len()).for_each(|i| {
+            let val = validate_and_reconstruct((result.0[i], result.1[i], result.2[i]));
+            output.insert(u8::from(val));
+        });
+
+        assert_eq!(output, input);
```
benjaminsavage, line 278: This only asserts that the same elements exist in the shuffled list. We need additional checks to validate that the order of the elements is different, and each element has been re-shared in a way that is different from the original.

src/protocol/sort/bit_permutations.rs
richajaindce: I will link to the paper. The new paper implements malicious whereas we are implementing semi honest for now. Once we get to malicous, will review the comments to ensure they are updated

src/protocol/sort/reshare.rs
richajaindce: Thats right but from to_helper.right POV, his share corresponds to input.0. This comment is to ensure it matches with what is coded below

src/protocol/sort/reshare.rs
richajaindce: good point! let me add that check

src/protocol/sort/shuffle.rs
richajaindce: No, all these algorithms are being implemented from the older semi honest paper. We will just pickup the malicious algorithms from the new paper when we reach that point.

src/protocol/sort/reshare.rs
richajaindce: Nope, the protocol does not communicate at all with the to_helper node. eg if we are computing for H2, H1 sends its parts to H3 and H3 sends parts to H1! H2 just assigns the prss random numbers as its' shares.

src/protocol/sort/reshare.rs
richajaindce: From helper POV, there is nothing called inputs.2 since it has two shares inputs.0 and inputs.1
This comment is to align with what is coded. Let me try to make it clearer

src/protocol/sort/reshare.rs
richajaindce: No, as mentioned above to_helper receives nothing from its peers. Only the other helpers do send and receive!

src/protocol/sort/shuffle.rs
richajaindce: Changed this to use Direction::Left and Direction::Right

src/protocol/sort/shuffle.rs
```diff
@@ -0,0 +1,302 @@
...
+    #[allow(dead_code)]
+    pub async fn execute<M: Mesh, G: Gateway<M, S>>(
+        &mut self,
+        ctx: &ProtocolContext<'_, G, S>,
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        // TODO : Need to find a way to have a better seed here. Can we use QueryID + SortStep combination to generate a seed?
+        let seed = self.input.len().try_into().unwrap();
```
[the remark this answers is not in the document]
benjaminsavage: @martinthomson - what should we do here?

benjaminsavage (changes requested):

src/protocol/sort/shuffle.rs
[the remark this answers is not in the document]
richajaindce: As an update: we are moving away from providing a seed to prss to generate shared random values across helpers. Instead, we would pass an index which hasn't been used by the current step. Shuffle calls reshares for all its records which needs random numbers (=number of records = batchsize). 

So we will use index = batchsize and batchsize+1 to generate randoms needed for Crypto Rnd, Chacha8Rng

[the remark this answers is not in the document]
martinthomson: Ideally our PRSS would generate an RNG that can be used by the permutation crate.  Then you could just feed the PrssSpace to the permutation.  In the meantime, I think that we need to consider what we discussed.  Generate some values from the PrssSpace and feed them into an Rng.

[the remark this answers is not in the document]
richajaindce: Sounds good.

src/protocol/sort/shuffle.rs
```diff
@@ -1,97 +1,226 @@
...
+        ctx: &ProtocolContext<'_, G, S>,
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        F: Field,
+    {
+        self.input = self.single_shuffle(ctx, Step1).await.unwrap();
+        self.input = self.single_shuffle(ctx, Step2).await.unwrap();
+        self.input = self.single_shuffle(ctx, Step3).await.unwrap();
+
+        Ok(self.input.clone())
```
richajaindce, line 163: @akoshelev is there a way to get rid of this clone? I tried making input mutable but since it has a lifetime attached to it, when we assign a new value to the input, the code doesn't compile

richajaindce: Thanks for the offline suggestions! Fixed this

src/protocol/sort/shuffle.rs
```diff
@@ -1,99 +1,226 @@
...
         batchsize: usize,
         direction: Direction,
-        index: u128,
         prss: &PrssSpace,
-    ) -> Vec<usize> {
-        let mut permutation: Vec<usize> = (0..batchsize).collect();
-        let seed = prss.generate_values(index);
+    ) -> Permutation {
+        /// This implements Fisher Yates shuffle described here <https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle>
+        fn generate_random_permutation_internal(
```
benjaminsavage, line 49: I really feel like defining this inline function makes this code much less readable. It took me a while to figure out what is going on.

Given that this function is called just once, I'd recommend NOT defining it this way, and instead just inlining the code. I think that'll save you the hassle of dealing with the two "seed" values. Right now there is a lot of code to just sort of write and read from these two values.

benjaminsavage (changes requested):
Looks pretty good. Just one last comment - that one function is very hard to read. Let's make it simpler then land.

src/protocol/sort/shuffle.rs
richajaindce: We do need 2 seed values (since we get a seed of u128 from prss while chacha8rng needs a u256). This will change once @martinthomson finishes changes for PrssRng to support it. However, agreed on inlining this code to make it readable.

richajaindce: Let me know if this looks better

PR #99, opened 2022-10-03
Reveal

richajaindce:
Revealing is the protocol that on input a pair of shares, outputs a secret. For simplicity, we consider a simple revealing in which each Pi sends [a]i to Pi+1 and then reconstructs a from [a]i and [a]i−1.


richajaindce:
@akoshelev this branch is forked on top of reshare. This should be quite a straightforward PR since each helper just sends one share to right peer. Feel free to review this first.

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+    },
+    protocol::{context::ProtocolContext, RecordId, Step},
+    secret_sharing::Replicated,
+};
+use embed_doc_image::embed_doc_image;
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've revealed their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct AValue<F> {
```
akoshelev, line 17:
```suggestion
pub struct RevealValue<F> {
```

maybe?

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn new(input: Replicated<F>) -> Self {
+        Self { input }
+    }
+
+    /// Steps
+    /// Each helper sends their left share to the right helper. The helper then reconstructs their secret by adding the three shares
+    /// i.e. their own shares and received share.
+    #[embed_doc_image("reveal", "images/sort/reveal.png")]
+    #[allow(dead_code)]
+    pub async fn execute<M: Mesh, G: Gateway<M, S>, S: Step + SpaceIndex>(
```
akoshelev, line 43: I need to fix this Gateway definition, let me do it right now

akoshelev: #100

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+impl<F: Field> Reveal<F> {
+    #[allow(dead_code)]
+    pub fn new(input: Replicated<F>) -> Self {
+        Self { input }
+    }
+
+    /// Steps
+    /// Each helper sends their left share to the right helper. The helper then reconstructs their secret by adding the three shares
+    /// i.e. their own shares and received share.
+    #[embed_doc_image("reveal", "images/sort/reveal.png")]
```
akoshelev, line 41: I think `H3` initial state is wrong - it should know (`a3`, `a1`) instead of (`a2`, `a3`)

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/// This implements reveal algorithm of "Efficient Secure Three-Party Sorting Protocol with an Honest Majority" at communication cost of 3R.
+/// Revealing is the protocol that on input a pair of shares, outputs a secret. For simplicity, we
+/// consider a simple revealing in which each Pi sends [a]i to Pi+1 and then reconstructs a from [a]i and
+/// [a]i−1.
+// Input: Each helpers know their own secret shares
+// Output: At the end of the protocol, all 3 helpers know a revealed (or opened) secret
+#[derive(Debug)]
+pub struct Reveal<F> {
```
akoshelev, line 28: I wonder if we should make this protocol sort (or even apply/applyinv) specific. i.e. impossible to apply outside of the allowed boundaries of sort. I worry that someone may instruct helpers to reveal their secret shares they've received from report collectors using this (because it takes just a replicated secret share). Maybe wrapping a replicated secret share into its own type is sufficient, so taking `PermutedShare<F>` instead of `Replicated<F>` may be enough.

this is just an idea, curious if we should discuss it further or if the reasoning here makes sense

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+use embed_doc_image::embed_doc_image;
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've revealed their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct AValue<F> {
+    a: F,
+}
+
+/// This implements reveal algorithm of "Efficient Secure Three-Party Sorting Protocol with an Honest Majority" at communication cost of 3R.
```
benjaminsavage, line 21: "...implements the reveal algorithm..."

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+/// Revealing is the protocol that on input a pair of shares, outputs a secret. For simplicity, we
```
benjaminsavage, line 22: Why "a pair of shares"? Doesn't it take all 3 secret-shares of just one sharing?

src/protocol/sort/reveal.rs
benjaminsavage: I don't think we need to worry about this. The helpers will not be doing anything but running this Rust crate. They aren't writing any code. If we call this at the wrong time, we are in trouble. I don't think we need to try to hide this function from ourselves.

The helpers don't just respond to commands right? That would be a problem.

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub async fn execute<M: Mesh, G: Gateway<M, S>, S: Step + SpaceIndex>(
+        self,
+        ctx: &ProtocolContext<'_, G, S>,
+        record_id: RecordId,
+        step: S,
+    ) -> Result<F, BoxError> {
+        let mut channel = ctx.gateway.get_channel(step);
+
+        let inputs = self.input.as_tuple();
+        // `to_helper.left` calculates part1 = (input.0 + input.1) - r1 and sends part1 to `to_helper.right`
```
benjaminsavage, line 52: This is a copy-pasta error =). Please remove this comment.

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,107 @@
...
+        protocol::{sort::reveal::Reveal, QueryId, RecordId},
+        test_fixture::{make_contexts, make_world, share, TestStep, TestWorld},
+    };
+
+    #[tokio::test]
+    pub async fn reveal() {
+        let mut rand = StepRng::new(100, 1);
+
+        let input = Fp31::from(23_u128);
+        let share = share(input, &mut rand);
```
benjaminsavage, line 86: As crazy as this sounds, what if there was a bug that only affected some numbers, but not 23. To avoid that we should generate a few random numbers in a loop and "reveal" all of them. Same for "reshare".

src/protocol/sort/reveal.rs
akoshelev: artificially making our life harder by trying to hide it would not help at all, I agree. I maybe overindexing on this because I've seen cases where something that wasn't intended to be used on real data actually sneaked into production code. Because it looked legit at the review phase,  this bug did not manifest itself until it was too late. 

While making this `Reveal` input "incompatible" with the input for regular `Reveal` function that would be used in tests does not guarantee correctness on its own, it may provide some defense in depth, because it won't be easily pluggable into the pipeline of transforming secret shares. Compiler will actually reject this. Of course someone may still try to use it, but hopefully it will trigger a conversation at the PR stage.

akoshelev: actually, it is probably enough to make sure 'Reveal' is not accessible from outside the sort module...

src/protocol/sort/reveal.rs
richajaindce: tbh I have no preference of one name over the other since the paper just defines this protocol in 3 lines (and doesnt have a formal definition for it to correlate). Its mostly for the infrastructure to distinguish one protocol from another.

src/protocol/sort/reveal.rs
richajaindce: My bad. Fixed now

src/protocol/sort/reveal.rs
richajaindce: it already has 2 shares of its own (which is obtained as input). While running the protocol, it receives the 3rd share and computes the secret. Edited to make it clearer

src/protocol/sort/reveal.rs
richajaindce: apologies. removed it now

src/protocol/sort/reveal.rs
```diff
@@ -0,0 +1,114 @@
...
+
+            let input = Fp31::from(secret);
+            let share = share(input, &mut rand);
+            let record_id = RecordId::from(1);
+
+            let world: TestWorld<SortStep> = make_world(QueryId);
+            let context = make_contexts(&world);
+
+            // TODO : change this step name to a test step
+            let step = SortStep::BitPermutations;
```
richajaindce, line 98: @thurstonsand @akoshelev whats the best way to use a SortStep in a test environment? Ideally, would like to introduce a TestStep of type SortStep and then use it here. But SortStep is an enum and not a crate.

src/protocol/mod.rs
```diff
@@ -38,6 +38,8 @@ pub enum IPAProtocolStep {
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
 pub enum SortStep {
     BitPermutations,
+    #[cfg(test)]
+    TestReveal,
```
richajaindce, line 42: Not sure if this is an appropriate use of test config. But it seems to be working

src/protocol/sort/reveal.rs
richajaindce: I tried to include a TestReveal under cfg(test) - based on definition, would this code compile only during test run?

benjaminsavage: > actually, it is probably enough to make sure 'Reveal' is not accessible from outside the sort module...

But we need a "reveal" for other things as well, such as the CheckZero functionality. I'd like to avoid duplicating that code.

richajaindce: Oh ya I totally forgot about that being used outside sort under malicious setting. Let me revert this change and bring it back to original form

akoshelev: hmm, that makes me feel that we may want to create `reveal` method on `ProtocolContext` that returns `Reveal` similar to what currently `multiply` does. We would want `Reveal` constructors to be hidden from IPA code (using pub(super) or pub(in path) https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself). The idea is that only `ProtocolContext` should be able to instantiate it and we can verify that the call site is allowed to reveal by checking the step variable.

I don't see a better way to check the step rather than downcast, it is ugly but it should work.

richajaindce: Made these changes. Makes sense to keep reveal in context since same as multiply, this is not scoped to only sort but will be used in other part of protocols as well. (e.g. when we implement malicious setting and need CheckZero protocol).

benjaminsavage (approved):
Looks good to me!

PR #100, opened 2022-10-03
Make `Mesh` associated type for `Gateway`

akoshelev:
Every `Gateway` implementation is specific to one `Mesh` so it makes sense to have `Mesh` as an associated type. `Gateway` clients do not need to pull `Mesh` generic argument anymore and it makes client code simpler


richajaindce:
Lets merge this also since this would make some of the protocol code simpler.

benjaminsavage (approved):
Nice.

akoshelev:
cancelling it as #102 addresses these issues too but also changes the `Gateway` definition

PR #104, opened 2022-10-07
Turn PrssSpace into an Rng

martinthomson:
This version probably doesn't address @richajaindce's issues because it requires that a mutable reference to the RNG be obtained.  I'll follow up with a less efficient version that might be easier to use.

### DIFF
--- a/src/helpers/prss.rs
+++ b/src/helpers/prss.rs
@@ -12,7 +12,7 @@ use std::ops::Index;
 use std::{marker::PhantomData, mem::size_of};
 use x25519_dalek::{EphemeralSecret, PublicKey};
 
-pub trait SpaceIndex {
+pub trait SpaceIndex: Copy {
     const MAX: usize;
 
     #[must_use]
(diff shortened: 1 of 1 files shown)


src/helpers/prss.rs
```diff
@@ -94,14 +94,68 @@ impl PrssSpace {
         let (l, r): (F, F) = self.generate_fields(index);
         l + r
     }
+
+    /// Turn this space into a pair of random number generators, one that is shared
+    /// with the left and one that is shared with the right.
+    /// Nothing prevents this from being used after calls to generate values,
+    /// but it should not be used in both ways.
+    #[must_use]
+    pub fn as_rngs(&self) -> (PrssRng, PrssRng) {
```
akoshelev, line 103: would it be better if this method sets the counter to u32::MAX and also consumes self?

src/helpers/prss.rs
```diff
@@ -94,14 +94,68 @@ impl PrssSpace {
...
+    counter: u128,
 }
 
+impl RngCore for PrssRng {
+    #[allow(clippy::cast_possible_truncation)]
+    fn next_u32(&mut self) -> u32 {
+        self.next_u64() as u32
+    }
+
+    // This implementation wastes half of the bits that are generated.
```
akoshelev, line 132: I guess we could keep `Option<u64>` in `PrssRng`, store it as 64 MSB from `v` left after truncation and yield it when `next_u64` is called if we really need to

src/helpers/prss.rs
```diff
@@ -94,14 +94,68 @@ impl PrssSpace {
...
+    // That is OK for the same reason that we use in converting a `u128` to a small `Field`.
+    #[allow(clippy::cast_possible_truncation)]
+    fn next_u64(&mut self) -> u64 {
+        let v = self.generator.generate(self.counter);
+        self.counter += 1;
```
akoshelev, line 137: we can probably use atomics here to make it thread-safe?

src/helpers/prss.rs
```diff
@@ -218,7 +272,7 @@ impl GeneratorFactory {
 }
 
 /// The basic generator.  This generates values based on an arbitrary index.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
akoshelev, line 275: wondering if clone causes secret key bytes to be copied as well and how bad it is if it is the case

src/helpers/prss.rs
martinthomson: Take a look at the first commit.

I would love to have this consume self, but that is not compatible with the way in which this object is used throughout.

src/helpers/prss.rs
martinthomson: Yeah, as noted, I don't think that this is a real concern, absent performance metrics that suggest otherwise.

src/helpers/prss.rs
martinthomson: I looked into this and it's not ideal, but nor is it particularly bad.  The clone duplicates the expanded keys, which for most implementations is 11 `__mi128` variables.  That's large enough to be noticeable, but I don't think that it is a problem for what this would be used for (the permutation).

akoshelev (approved):
I also wonder if we should just implement our own trait for `Rng` with methods that take `&self` rather than `&mut self` if we really don't want exclusive references. My naive understanding is that we could use atomics to keep the index/ctr and feed it to the cipher after increment

martinthomson:
Using atomics would allow for concurrent access, but it wouldn't guarantee that the values were consistent between helpers; that's something that we need for the shuffle.

PR #106, opened 2022-10-07
[APPLYINV] This protocol will apply a secret shared permutation on secret shared input across three helpers

richajaindce:


richajaindce:
I am working on a separate PR to do protocol operations in place as compared to cloning objects across the sort protocols. Putting this out to avoid rebasing many more times. Please review the other parts of the code

richajaindce:
Found an issue in the implementation,  will get back after fixing it

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,172 @@
...
+    ) -> Result<(), BoxError> {
+        // 1. Shuffle permutations
+        let shuffle_perm_fn = compose_ipa_step!(ShufflePermutation, ApplyInv, Sort);
+        let mut shuffle_permutation = Shuffle::new(self.permutation, shuffle_perm_fn);
+        let h0_future = shuffle_permutation.execute(ctx);
+
+        // 2. Shuffle inputs
+        let shuffle_input_fn = compose_ipa_step!(ShuffleInputs, ApplyInv, Sort);
+        let mut shuffle_input = Shuffle::new(self.input, shuffle_input_fn);
+        let h1_future = shuffle_input.execute(ctx);
```
richajaindce, line 61: ApplyInv needs to call shuffle two times once for shuffling inputs and another for shuffling permutation. They both need to use the same sequence combination to calculate the shuffle. However, the infra needs a unique step for each communication requested. With each unique step comes a new prss space which will generate a new sequence not shared between input and permutation.
We would need the exact functionality for COMPOSE as well which will share the sequence between shuffle and unshuffle.
What is the best way to do this? @martinthomson @akoshelev

martinthomson:
```rust
enum ApplyInvStep {
  ShuffleInput(ShuffleStep),
  ShufflePermutation(ShuffleStep),
  /* plus any other */
}
```

Then you need to work through the integration of the pieces into a single numbering space:

```rust
impl SpaceIndex for ApplyInvStep { 
  const MAX: usize = ShuffleStep::MAX * 2;
  fn as_usize(&self) -> usize {
    match self {
      ShuffleInput(s) => s.as_usize(),
      ShufflePermutation(s) => ShuffleStep::MAX + s.as_usize(),
    }
  }
}
```

richajaindce: I am not 100% sure i understand Space completely. How would changing ShuffleInput(s) => 0 to s.usize() help here? 
Also, do the types in a Step (e.g. ShuffleInput and ShufflePermutation in ApplyInvStep) share same prss random seed? i.e. will generate_value(INDEX) called for ShuffleInput give same output as generate_value(INDEX) called for ShufflePermutation? @martinthomson

martinthomson: The goal is to ensure that every step that is executed during a query gets a different seed.  For that, you need a unique value for each, even if it is doing the same work as another step.  For instance, we multiply many times, but each multiplication needs a different identifier.

What this does is - at a local level - give each discrete step a different number.  It also lets you compose simple processes (say you just want to shuffle on its own) into more complex processes (applyinv with two shuffles).

richajaindce: Is it intentional that each step in the IPAProtocol should share the same prss? For now, all of the steps seem to be having the same prss (initialized by same random number).
For shuffle, we would need to have different random numbers across 3 steps. One way to handle that can be to call generate_values with different index for each step.

martinthomson: The idea here is to have a single Participant (naming likely needs work).  From that, multiple PrssSpace instances are created, one for each Step that is executed.  The way that we ensure that each PrssSpace is different is by providing a unique identifier and that is the responsibility of the SpaceIndex trait that IPAProtocolStep implements.

In a full execution of the protocol, there will be multiple places where we need a PRSS instance (PrssSpace).  There will be multiple shuffles and multiple multiplications.  Each of these will need a unique identifier so that the PrssSpace is different.

Ultimately, we need to provide unique random values for every single invocation of the PRSS (or PrssSpace).  For that, we need a unique identifier for the step and a unique identifier for the record that is being processed in that step.

(Note that in testing, you can use a smaller or more narrowly scoped SpaceIndex implementation that provides unique identifiers for each step.  Our current implementation here is lacking as it doesn't provide unique identifiers for each.)

richajaindce:
Making changes to this and will send off a new PR for it

PR #107, opened 2022-10-08
Cargo.toml: remove unused deps, bump versions, etc

akoshelev:
* Upgrade to clap v4 from structopt. Structopt is in maintenance mode and has been ported to clap_derive. 
* Bump the deps
* Bump Rust compiler to 1.64
* Remove unused dependencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,25 +15,24 @@ test-fixture = []
 
 [dependencies]
 aes = "0.8"
-async-trait = "0.1.56"
+async-trait = "0.1.57"
 axum = { version = "0.5.16", optional = true, features = ["http2"] }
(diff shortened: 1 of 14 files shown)


martinthomson (approved):

martinthomson:
Will dropping `rand_core` affect implementing `RngCore`?

akoshelev:
> RngCore

I believe these traits are exactly the same

https://docs.rs/rand/0.8.5/rand/trait.RngCore.html
https://docs.rs/rand_core/0.6.4/rand_core/trait.RngCore.html

If I am reading it correctly, `rand_core` is just a set of core `rand` traits. `rand` re-exports those and we can easily switch back to `rand_core` if we need to (i.e. if we decide to make raw-ipa a workspace for multiple crates and isolate core logic that shouldn't depend on the specific RNG implementation)

PR #108, opened 2022-10-10
Oblivious attribution protocol [part 1 - attribution]

taikiy:
# Overview
This diff implements the first part (out of two - "attribution" and "aggregation") of the last touch oblivious attribution in a secure multiparty computation. Currently it supports "Last Touch" attribution only. Produced result will be used in the second part - "aggregation", which requires sorting of the list.

# Details
Last touch attribution algorithm is described in details here:
https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#oblivious-last-touch-attribution

First part of the protocol accesses and accumulates data of each node in a sorted list, and produces an intermediate result that has accumulated credits for each breakdown key. The document also describes how the logic can be optimized as follows:
```
flag = current.stop_bit * successor.helper_bit * successor.trigger_bit
current.credit = current.credit + flag * successor.credit
current.stop_bit = flag * successor.stop_bit
```
This is what's implemented in `last_touch_attribution()` with multiplications done on a MPC. I believe there could be more optimizations if we put more constrains on the input, but for this round this basic/general implementation should be enough for us to measure baseline benchmarks.


taikiy:
Rust check workflow failing because of a clippy error in a code comment. I'll fix this in the next commit when I address feedbacks.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::mesh::{Gateway, Mesh},
+    protocol::context::ProtocolContext,
+    secret_sharing::Replicated,
+};
+use futures::future::try_join_all;
+
+/// Obilivious attribution protocol
```
martinthomson, line 11:
```suggestion
/// Oblivious attribution protocol
```

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+        //     credit = credit + b * successor.credit;
+        //     stop_bit = b * successor.stop_bit;
+        let mut step_size = 1;
+        while step_size < num_rows {
+            let end = num_rows - step_size;
+
+            // first, calculate [successor.helper_bit * successor.trigger_bit]
+            let mut b = self
+                .multiply_all(
+                    &input[Self::_HELPER_BIT][step_size..],
```
martinthomson, line 81: I find this ordering of the arrays surprising.  Why do we have an array of arrays where the outer array is indexed by the field type?  Why are we using array indices to access shares/fields anyway?

I had expected that we would be operating on a sequence of structs, where each field in the struct is a replicated share.  This looks like a sequence of sequences, where the outer sequence has some sort of implicit structure.  That leads to some odd potential failure modes that need to be guarded against, like where the Vec for the helper bit is a different length to the Vec for the stop bit.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+};
+use futures::future::try_join_all;
+
+/// Obilivious attribution protocol
+///
+/// Obliviously attribute trigger events to corresponding source events for different breakdowns in a MPC.
+/// There are several common attribution models such as Last Touch, Even Credit and Exponential Decay.
+/// https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#oblivious-last-touch-attribution
+#[allow(dead_code)]
+pub struct Attribute<'a, G> {
```
martinthomson, line 17: "Attribute" is maybe not the best choice of name for this, given the potential for it to be confused.  "Attribution" is probably cleaner, even if it doesn't have the right tense.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+                .enumerate()
+                .for_each(|(i, x)| stop_bits[i] = x);
+
+            step_size *= 2;
+        }
+
+        Ok(credits)
+    }
+
+    async fn multiply_all<M, F>(
```
martinthomson, line 120: Is this a function that we should be providing on the Gateway?

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+    where
+        F: Field,
+        M: Mesh,
+        G: Gateway<M, AttributionModel>,
+    {
+        assert_eq!(a.len(), b.len());
+
+        try_join_all((0..a.len()).map(|i| async move {
+            let x = a[i];
+            let y = b[i];
```
martinthomson, line 134: I think that you want to use zip() here instead:

```rust
a.zip(b).map(async move |(a, b)| self.ctx.multiply(RecordId::from(i), self.step).await.execute(a, b).await)
```

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// Obilivious attribution protocol
+///
+/// Obliviously attribute trigger events to corresponding source events for different breakdowns in a MPC.
+/// There are several common attribution models such as Last Touch, Even Credit and Exponential Decay.
+/// https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#oblivious-last-touch-attribution
+#[allow(dead_code)]
+pub struct Attribute<'a, G> {
+    ctx: &'a ProtocolContext<'a, G, AttributionModel>,
+    step: AttributionModel,
```
martinthomson, line 19: This isn't going to work, I don't think.  The step needs to be taken from the same global enum.  So in order to have this work, you need to take the top-level enum we're using.  Which means parameterizing on `S: Step` and taking an argument of that type.

I also think that we will probably want to make the implementation of attribution a generic parameter.  That is, you might define a generic attribution step as a trait, then a concrete implementation of that would be last touch.

martinthomson (changes requested):
Just a short review, based on a few things that I noticed.

src/protocol/attribute.rs
benjaminsavage: Yeah, I'd recommend just focusing on this one attribution algorithm in isolation for now. Once we have it running, we can think about how to support other stuff. I'd call this module "LastTouchAttribution"

src/protocol/attribute.rs
taikiy: I was just discussing this with Richa and others. I agree that we expect to receive a sequence of structs. When it comes to the actual protocol executions, we realized we were always processing a sequence of replicated shares for a certain field (i.e. match key). For this PR, I chose this input format for the sake of simplicity, assuming the input had been transposed. Should all top-level protocol steps expect that the input format would be a sequence of structs?

src/protocol/attribute.rs
taikiy: Makes sense. Many protocols seem to need this, so let's put it in the gateway.

src/protocol/attribute.rs
benjaminsavage: I'm confused by your reply taiki.

I think Martin is suggesting something like:

struct AttributionInputs {
    helper_bit: Replicated<F>,
    is_trigger_bit: Replicated<F>,
    value: Replicated<F>,
    breakdown_key: Replicated<F>,
}

And the "execute" function would accept a Vec of these things.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            // [b * successor.credit]
+            let b_cred = self.multiply_all(&b[..end], &credits[step_size..]).await?;
+
+            // [credit] = [credit + b * successor.credit]
+            b_cred
+                .into_iter()
+                .enumerate()
+                .for_each(|(i, x)| credits[i] += x);
```
benjaminsavage, line 98: OK, I see what you're doing here. It's very subtle. The index "i" refers to the correct successor here because the b_cred vector is shifted compared to the original input.

That is very clever, but I'm worried its a bit too subtle. It's harder to understand and I worry will lead to bugs.

Honestly, I would rather see something like:

b_cred = foo[i] + bar[i + step_size];

I feel like that makes it easier to see what's going on.

src/protocol/attribute.rs
benjaminsavage: @richajaindce could have used this in GenBitPerm

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut rng = StepRng::new(100, 1);
+
+        let input: [[u128; 9]; 6] = [
+            [1, 1, 1, 1, 1, 1, 1, 2, 3],
+            [2, 5, 8, 9, 10, 13, 27, 11, 14],
+            [0, 0, 1, 1, 1, 1, 1, 0, 1],
+            [0, 1, 1, 1, 1, 1, 1, 0, 0],
+            [3, 4, 0, 0, 0, 0, 0, 1, 0],
+            [0, 0, 10, 2, 1, 5, 1, 0, 10],
+        ];
```
benjaminsavage, line 171: Yeah, I have no idea what this means. I think the structs will help.

src/protocol/attribute.rs
taikiy: The code was bit more readable to write

```
.multiply_all(
    &self.input[Self::_HELPER_BIT][step_size..],
    &self.input[Self::_IS_TRIGGER_BIT][step_size..],
)
```

than

```
impl Index for AttributionInputs { ... }

.multiply_all(
    &self.input[step_size..],
    SELF::_HELPER_BIT,
    SELF::_IS_TRIGGER_BIT,
)
```

that's all. I'm okay with the latter.

As for the sort protocol, I believe it creates a vector for each sort key. @richajaindce, your thoughts?

richajaindce: It looks like we always operate on same type. This is true for running attribution and also if we want to execute a shuffle data attached to matchkey before sort (to hide information).  Would keeping all of data in memory be an issue? Using Taiki's approach, not everything needs to be in memory vs event approach needs all data to be in memory.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+        M: Mesh,
+        G: Gateway<M, AttributionModel>,
+    {
+        assert_eq!(a.len(), b.len());
+
+        try_join_all((0..a.len()).map(|i| async move {
+            let x = a[i];
+            let y = b[i];
+
+            // assume # of records < 2^32
```
akoshelev, line 136: lets add `debug_assert` for that maybe?

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+    async fn last_touch_attribution<M, F>(
+        &self,
+        input: &'a [Vec<Replicated<F>>],
+    ) -> Result<Vec<Replicated<F>>, BoxError>
+    where
+        M: Mesh,
+        G: Gateway<M, AttributionModel>,
+        F: Field,
+    {
+        // TOOD: input length range check. Can we assume `assert!(num_rows < u32::MAX)`?
```
akoshelev, line 58: I would much rather prefer if we turn this assumption into a type, i.e. use u32 here and below. In this case we don't need to keep being cautious about pieces of code that assume range fits into u32 but does not validate it. 

This unfortunately leads to major inconveniences working with vector indices and vector capacity as we have to use usizes for it. 

@martinthomson @benjaminsavage I wonder if we should create a wrapper around Vec and implement `Index<u32>/IndexMut<u32>` for it? We will enforce that this vector never grows over $2^{32}-1$ elements but working with it will be much more ergonomic and type safe than standard vec.

src/protocol/attribute.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut stop_bits: Vec<Replicated<F>> = vec![one; num_rows];
+
+        // 2. Accumulate
+        //
+        // Last touch accumulation algorithm:
+        //     b = stop_bit * successor.helper_bit * successor.trigger_bit;
+        //     credit = credit + b * successor.credit;
+        //     stop_bit = b * successor.stop_bit;
+        let mut step_size = 1;
+        while step_size < num_rows {
```
akoshelev, line 75: you can replace these three lines (init, while, *= 2) with a one liner that creates an iterator over powers of 2. 

```rust
        // generate powers of 2 that fit into input len. If num_rows is 15, this will produce [1, 2, 4, 8]
        for step_size in std::iter::successors(Some(1u32), |prev| prev.checked_mul(2)).take_while(|&v| v < num_rows) {
}
```

src/protocol/attribute.rs
martinthomson: Yes, I think that a wrapper would do a lot to help.

martinthomson: Oh, I should say: the index type should probably be a typedef, like `type RecordIndex = u32` or something along those lines.

src/protocol/attribute.rs
taikiy: I am going to fix this part to take a slice of structs, but just wanted to clarify what's going on in the code.
So, the [attribution algorithm](https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#oblivious-last-touch-attribution) would look like this:

```
// input: Vec<GenericReport>
let mut credit: Vec<Replicated<F>> = input.map(|x| x.value).collect();
let mut stop_bit: Vec<Replicated<F>> = [F::ONE; input.len()];

while step_size < input.len()
    let mut new_credit = credit.clone();
    let mut new_stop_bit = stop_bit.clone();

    for i in 0..input.len() {
        if i+step_size > input.len() break;

        let current = i;
        let successor = i+step_size;

        // in reality these will be lengthy async multiply calls
        let flag = stop_bit[i] * input[successor].helper_bit * input[successor].trigger_bit;
        new_credit[current] = credit[current] + flag * credit[successor];
        new_stop_bit[current] = flag * stop_bit[successor];
    }

    // `new_credit` is (input.len() - step_size) long, but `credit` must be input.len() length
    let end = input.len() - step_size;
    credit = new_credit[0..end].append(credit[end..]);
    // same for stop_bit
    stop_bit = new_stop_bit[0..end].append(stop_bit[end..]);
    
    step_size *= 2;
}
...
// we drop unnecessary fields. only need to return aggregated credits
Ok([{breakdown_key, credit}; unique(breakdown_key).len()])
```

Note that a `for` loop iteration doesn't depend on the previous iterations. Since they are independent we can eliminate the `for` loop and execute multiplications on slices of shares concurrently, which is what `multiply_all` does. It also eliminates awkward boundary checks, reassign of variables, padding vectors, etc.

src/protocol/attribute.rs
taikiy: Yeah, this is my fault. I didn't add any comments to this and some arrays are irrelevant. I'll clarify in this thread. https://github.com/martinthomson/raw-ipa/pull/108#pullrequestreview-1136724098

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+        context::ProtocolContext,
+        RecordId, Step,
+    },
+    secret_sharing::Replicated,
+};
+use futures::future::{try_join, try_join_all};
+
+/// Accumulation step for Oblivious Attribution protocol.
+#[allow(dead_code)]
+pub struct Accumulate<'a, F> {
```
taikiy, line 17: Renamed from `Attribution` to `Accumulate` to better reflect what it is actually doing. There are three steps in the oblivious attribution. 1. accumulation (this step), 2. sorting, 3. aggregation. I will create Attribution trait in future PR when I implement the rest of the steps.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        let expected_credit_output = vec![0_u128, 19, 19, 9, 7, 6, 1, 0, 10];
+
+        let steps = [
+            AccumulateTestStep::Step(0),
+            AccumulateTestStep::Step(1),
+            AccumulateTestStep::Step(2),
+            AccumulateTestStep::Step(3),
+            AccumulateTestStep::Step(4),
+        ];
```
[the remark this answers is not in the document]
martinthomson: Try `(0..=4).map(|i| AccumulateTestStep::Step(i))` if you want to type less.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+        //
+        // Each list element interacts with exactly one other element in each iteration, and the interaction do not
+        // depend on the calculation results other elements, allowing the algorithm to be executed in parallel.
+
+        // generate powers of 2 that fit into input len. If num_rows is 15, this will produce [1, 2, 4, 8]
+        for step_size in std::iter::successors(Some(1u32), |prev| prev.checked_mul(2))
+            .take_while(|&v| v < num_rows)
+        {
+            let end = num_rows - step_size;
+            let mut accumulation_futures = Vec::with_capacity(end as usize);
```
benjaminsavage, line 78: There is a bit of a tricky thing with the rows that are near the end. They don't need any more accumulation, but you need to pass them along somehow.

e.g. if there are 31 rows, then on iteration 5 you're going to have step_size = 16, and end = 15. So there will be just 15 objects in the "accumulation_futures" vector. Those have accumulated credit from the first 4 iterations that needs to not get lost.

src/protocol/attribution/accumulate.rs
benjaminsavage: Maybe AccumulateCredit? Since "accumulate" is kind of a general term.

src/protocol/attribution/accumulate.rs
taikiy: Following your example, do you mean at step_size = 16, end = 15 and accumulating credit at row i = 0, I need to pass credit[16] so that I can calculate `credit[0] + b * credit[16]`?

There is a separate `credits` vector that keeps track of the accumulated credits (line 110). When I create `current` and `successor` structs, I copy values from this `credits` (line 84 and 89) which have accumulated credits from the previous iteration.

benjaminsavage: Oh, I see. You're absolutely correct. The "credits" mutable vector should have everything in it.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+        // number of elements to the output won't be a problem. Adding u32::MAX elements will be 1B + u32::MAX, which
+        // exceeds our current assumption of `input.len() < 1B`.
+
+        Ok(output)
+    }
+
+    async fn get_accumulated_credit<S, N>(
+        &self,
+        ctx: &ProtocolContext<'_, S, N>,
+        step_size: u32,
```
benjaminsavage, line 139: Rather than pass in the step size (which it doesn't really need to know), perhaps just pass in a boolean indicating if this is the first step or not.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+                    report: self.input[i + step_size],
+                };
+
+                accumulation_futures.push(self.get_accumulated_credit(
+                    ctx,
+                    step_size,
+                    current,
+                    successor,
+                    RecordId::from(i),
+                    [steps[1], steps[2], steps[3], steps[4]],
```
benjaminsavage, line 99: This is really not awesome. Fortunately this is probably all going away as a result of @martinthomson's latest PR.

src/protocol/attribution/accumulate.rs
taikiy: Yes, I was wondering what would be the most non-absurd way to pass this. I thought `if step_size > 1` was a tiny bit more readable than `if !first_iteration`, but I'm okay with a boolean, too.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+                ));
+            }
+
+            let results = try_join_all(accumulation_futures).await?;
+
+            // save the calculation results to these support vectors for use in the next iteration
```
benjaminsavage, line 105: I get thrown off by the term "support vectors" since it makes me think of https://en.wikipedia.org/wiki/Support_vector_machine

I'd suggest:

// accumulate the credit from this iteration into the accumulation vectors

src/protocol/attribution/accumulate.rs
taikiy: Agree 100%. I'll probably wait for Martin's PR to merge and improve this part. I think I actually need different steps for each iteration (step_size). I do `try_join_all` before I go to the next iteration but that's not sync'ed across the helpers, so things might get tangled up.

src/protocol/attribution/accumulate.rs
benjaminsavage: You could do something similar to what @richajaindce did here: https://github.com/martinthomson/raw-ipa/pull/97/files#diff-894d6d54e26cbe3155f95c21f6bd6b557778f2705f8371fc2bc1d31b2292cb74

She made functions accept an enum - which only had two values. Direction::LEFT and Direction::RIGHT.

This makes reading the call site easier to understand than seeing the function accept "true" or "false" for the nth argument. You could have FooBar::FIRST_ITERATION and FooBar::SUBSEQUENT_ITERATION or something.

src/protocol/attribution/accumulate.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        let credit_future = ctx
+            .multiply(record_id, steps[2])
+            .await
+            .execute(b, successor.credit);
+
+        // for the same reason as calculating [b], we skip the multiplication in the first iteration
+        let stop_bit_future = if step_size > 1 {
+            futures::future::Either::Left(
```
benjaminsavage, line 172: What does this do?

src/protocol/attribution/accumulate.rs
taikiy: Hmm, that might be an overkill for this case. Rust doesn't support ternary conditional operator so the call site would then look like this.

```
let iteration = if step_size > 1 { FooBar::FIRST_ITERATION } else { FooBar::SUBSEQUENT_ITERATION }; // need the brackets, too

get_accumulated_credit(
    ctx,
    iteration,
    ...
```

benjaminsavage: You could have a mutable local variable initialized to FooBar::FIRST_ITERATION, which is updated to FooBar::SUBSEQUENT_ITERATION at the end of the loop. Then you don't need a ternary.

src/protocol/attribution/accumulate.rs
taikiy: `Either::Left` and `Either::Right` let you return futures conditionally. For this case, I want to skip the multiplication in the first iteration, so I return a future that resolves to `b` immediately when awaited.

src/protocol/attribution/accumulate.rs
taikiy: Or maybe I could use the `steps` to infer which iteration I'm in? As I mentioned in a separate comment, I think I need 4 * log(n) steps rather than just 4. I need to look into Martin's PR to see if that's feasible.

benjaminsavage (approved):
This looks good to me. I recommend you coordinate with @martinthomson to see if you should land this first, or wait until he lands his first.

src/protocol/batch.rs
```diff
@@ -0,0 +1,75 @@
+use std::ops::{Deref, DerefMut, Index, IndexMut};
+
+pub type RecordIndex = u32;
+
+/// A wrapper around Vec<T> ensuring that the wrapped vector has max length <= `RecordIndex::MAX`
+#[derive(Debug, Clone)]
+pub struct Batch<T>(Vec<T>);
```
taikiy, line 7: This wrapper ensures that the length of the inner vector is less than RecordIndex::MAX upon index references, but doesn't prevent it from growing larger at the time items are added (e.g., push(), append()). Is there a way to monitor changes in its size? @martinthomson @akoshelev

martinthomson: It might be better to build this without exposing the internal collection, so that you only need to cover a few different interfaces for manipulation.

I imagine that you want to support `Iterator::collect()`, but maybe you don't need push, extend_from_slice() and so forth.

taikiy:
Rebase onto @martinthomson's renewed Step scheme. The most recent commit to be reviewed: https://github.com/martinthomson/raw-ipa/pull/108/commits/9e4d5378e13bd9424340876804e8dbe1b0603580

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -0,0 +1,300 @@
...
+
+                accumulation_futures.push(Self::get_accumulated_credit(
+                    ctx.narrow(multiply_step.next()),
+                    current,
+                    successor,
+                    iteration_step.is_first_iteration(),
+                ));
+            }
+
+            let results = try_join_all(accumulation_futures).await?;
```
akoshelev, line 102: @martinthomson in many protocols there seem to be a pattern of generating large number of futures and then polling them all. Do you see it as a problem, given that no work is done until the first poll?

src/protocol/attribution/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        Self {
+            name,
+            count: 0,
+            id: String::from(name),
+        }
+    }
+
+    fn next(&mut self) -> &Self {
+        self.count += 1;
+        self.id = format!("{}_{}", self.name, self.count);
```
akoshelev, line 45: there will be $O(N)$ allocations here, we probably want to mark this as todo to fix. I think @martinthomson was planning to add a cheap way to create a unique step id for release builds

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -0,0 +1,300 @@
...
+        RecordId,
+    },
+    secret_sharing::Replicated,
+};
+use futures::future::{try_join, try_join_all};
+
+/// Accumulation step for Oblivious Attribution protocol.
+#[allow(dead_code)]
+pub struct AccumulateCredit<'a, F> {
+    input: &'a Batch<AttributionInputRow<F>>,
```
akoshelev, line 18: this input format implies that somebody after doing the sort must convert the regular batch of sorted secret shares into this "expanded" format with helper bits. Is the plan to insert an "intermediate" step between sort and attribution to convert shares format or you want `AccumulateCredit` to eventually implement the conversion?

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -0,0 +1,300 @@
...
+        //
+        //   b = current.stop_bit * successor.helper_bit * successor.trigger_bit;
+        //   new_credit[current_index] = current.credit + b * successor.credit;
+        //   new_stop_bit[current_index] = b * successor.stop_bit;
+        //
+        // Each list element interacts with exactly one other element in each
+        // iteration, and the interaction do not depend on the calculation results
+        // of other elements, allowing the algorithm to be executed in parallel.
+
+        let mut iteration_step = IterStep::new("iteration");
```
akoshelev, line 69: Please correct me if I am wrong, but it seems we are using `record_id` as step and `iteration_step` as record_id. Is it intended?

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -0,0 +1,300 @@
...
+        &self,
+        ctx: ProtocolContext<'_, N>,
+    ) -> Result<Batch<AccumulateCreditOutputRow<F>>, BoxError> {
+        #[allow(clippy::cast_possible_truncation)]
+        let num_rows = self.input.len() as RecordIndex;
+
+        // 1. Create credit and stop_bit vectors
+        // These vectors are updated in each iteration to help accumulate values and determine when to stop accumulating.
+
+        let one = Replicated::one(ctx.mesh().identity());
```
akoshelev, line 45: I wonder if it is possible to make credit and stop bit part of the input?

src/protocol/attribution/accumulate_credit.rs
taikiy: Yes, that is correct. This is the issue @benjaminsavage and I were talking about during the meeting. Since this is a sub-protocol that executes 4 multiplications on a single row, I wanted to pass a reference to the ctx, set 4 multiplication steps like Step::Mult1, Step::Mult2, ..., and each row (parallelized) have different record IDs (i.e. row i = RecordId::from(i)). With this set up, in the sub-protocol, the step should look like `protocol/iteration_1/mult_1` and `recrod_id=i` where i in `(0..input.len())`. But when I try to run that, I get this error:

```
thread 'protocol::attribution::accumulate_credit::tests::accumulate' panicked at 'Refined 'protocol' with step 'mult_1' twice', src/protocol/mod.rs:94:13
```

So, I ended up with this diff, which calls another `ctx.narrow()` for each row to give them unique steps (i.e. "row_i"), and each multiplication as separate records.

src/protocol/attribution/mod.rs
taikiy: This should go away once we decide to move away from `AsRef<str>`. I understand AsRef<str> for steps is temporary for this prototype?

src/protocol/attribution/accumulate_credit.rs
taikiy: Helper bits will be a part of the sort output. IIUC, the sort protocol will output a vector or vectors of index which corresponds to the sorted locations of the input list. I think there will be an intermediate step where we might remove unnecessary fields like match keys and timestamps, but that is still unknown.

src/protocol/attribution/accumulate_credit.rs
taikiy: I can see them being part of the input if we have a parent protocol that calls `AccumulateCredit` after `Sort`, but we don't have that yet. Credits and stop bits are the states/results generated by running this protocol, so it seems reasonable to create them inside this protocol.

src/protocol/attribution/accumulate_credit.rs
benjaminsavage: I have this strong suspicion that generating the "helper bits" *within* the MPC will be more efficient than trying to generate them in a separate step.

src/protocol/attribution/accumulate_credit.rs
benjaminsavage: I agree with @taikiy - these are basically local variables to this protocol. They're all initialized to Replicated::one(), or one of the other inputs at the beginning of the loop (and over-written in each iteration) so there's no value in passing them in.

src/protocol/attribution/accumulate_credit.rs
benjaminsavage: We shouldn't have to do this. This is an issue with where that "check" happens in the code. Looking forward to an infra fix that helps avoid the need to do this!

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -0,0 +1,300 @@
...
+        first_iteration: bool,
+    ) -> Result<(Replicated<F>, Replicated<F>), BoxError> {
+        // For each input row, we execute the accumulation logic in this method
+        // `log2(input.len())` times. Each accumulation logic is executed with
+        // the unique iteration/row pair sub-context. There are 2~4 multiplications
+        // in this accumulation logic, and each is tagged with a unique `RecordID`.
+
+        // first, calculate [successor.helper_bit * successor.trigger_bit]
+        let mut b = ctx
+            .multiply(RecordId::from(0))
```
benjaminsavage, line 148: This seems wrong. This should be a proper record_id, not just 0 or 1. This is related to the same infra issue from above - we need the ability to run N multiplications in parallel, with the same step, and only the record_id differing.

benjaminsavage (changes requested):
This is all looking really good, with the exception of needing to work around that infra issue that prevents us from using the same step name on different records.

PR #109, opened 2022-10-11
Unique random number per shuffle step

richajaindce:
Currently, PRSS is generating same random numbers across steps. For shuffle, each step expects a different random number to act as a seed.

Not sure if having same prss across the protocol is an expected behaviour or should each step have a unique prss random number generator.


richajaindce:
FINALLY understood why prss was being shared across steps. It was due to using the same spaceindex accidentally. This change is not needed anymore so closing it. Another PR with some added tests is coming up

PR #111, opened 2022-10-11
Fix build

akoshelev:
Somehow one more use of deprecated method `Permutation::from_vec` was still in the code after upgrading to 0.4


akoshelev:
Ah, I think #98 got into main first and brought this usage

richajaindce:
Looks good! Please merge this change

PR #112, opened 2022-10-12
rand_core is back!

martinthomson:
Yep, [it turns out we needed it](https://github.com/martinthomson/raw-ipa/pull/107#issuecomment-1272649289).


akoshelev:
long live `rand_core`. I found a way to import these from `rand_chacha` but that's a hack. 
I am genuinely surprised that these are not re-exported https://github.com/rust-random/rand/blob/master/src/lib.rs#L94

akoshelev (approved):

PR #114, opened 2022-10-12
Fixing test and adding asserts

richajaindce:
1. Fixing reshare test to avoid flaky behavior
2. Adding a test to validate a new space is generated for each step
3. Added assert in shuffle to ensure steps do not share prss.

### DIFF
--- a/src/protocol/sort/reshare.rs
+++ b/src/protocol/sort/reshare.rs
@@ -114,7 +114,7 @@ mod tests {
     pub async fn reshare() {
         let mut rand = StepRng::new(100, 1);
         let mut rng = rand::thread_rng();
-
+        let mut new_reshares_atleast_once = false;
         for _ in 0..10 {
             let secret = rng.gen::<u128>();
 
@@ -139,9 +139,11 @@ mod tests {
             let output_share = validate_and_reconstruct(f);
             assert_eq!(output_share, input);
 
-            assert_ne!(share[0], f.0);
-            assert_ne!(share[1], f.1);
-            assert_ne!(share[2], f.2);
+            if share[0] != f.0 && share[1] != f.1 && share[2] != f.2 {
+                new_reshares_atleast_once = true;
+                break;
+            }
         }
+        assert!(new_reshares_atleast_once);
     }
 }
--- a/src/test_fixture/world.rs
+++ b/src/test_fixture/world.rs
@@ -69,15 +69,15 @@ impl Debug for TestStep {
 impl Step for TestStep {}
 
 impl SpaceIndex for TestStep {
-    const MAX: usize = 5;
-
+    const MAX: usize = u8::BITS as usize * 3 + ShuffleStep::MAX + 1;
     fn as_usize(&self) -> usize {
+        let u8_size = u8::BITS as usize;
         match self {
-            TestStep::Mul1(_) => 0,
-            TestStep::Mul2 => 1,
-            TestStep::Reshare(_) => 2,
-            TestStep::Reveal(_) => 3,
-            TestStep::Shuffle(_) => 4,
+            TestStep::Mul1(s) => *s as usize,
+            TestStep::Mul2 => u8_size,
+            TestStep::Reshare(s) => u8_size + 1 + *s as usize,
+            TestStep::Reveal(s) => u8_size * 2 + 1 + *s as usize,
+            TestStep::Shuffle(s) => u8_size * 3 + 1 + s.as_usize(),
         }
     }
 }


src/protocol/sort/reshare.rs
```diff
@@ -142,9 +142,11 @@ mod tests {
...
 
-            assert_ne!(share[0], f.0);
-            assert_ne!(share[1], f.1);
-            assert_ne!(share[2], f.2);
+            if share[0] != f.0 && share[1] != f.1 && share[2] != f.2 {
+                new_reshares_atleast_once = true;
+                break;
+            }
         }
+        assert!(new_reshares_atleast_once);
```
benjaminsavage, line 150: I mean, this isn't really what we want to test. We want to test that they ALL got re-shared. This is a stop-gap solution to fix the test failure, but I think longer-term you want to test with a bigger field.

src/helpers/prss.rs
```diff
@@ -36,6 +36,12 @@ pub struct PrssSpace {
     right: Generator,
 }
 
+impl PartialEq for PrssSpace {
+    fn eq(&self, other: &Self) -> bool {
+        self.generate_values(1) == other.generate_values(1)
```
benjaminsavage, line 41: Hmmm... I guess this is one way to test it. Can't we compare the seeds instead? That seems better to me.

src/helpers/prss.rs
```diff
@@ -503,4 +513,18 @@ pub mod test {
...
+
+    #[tokio::test]
+    async fn different_prss() {
+        // each time we create a step, it should generate a different prss
+        let world: TestWorld<TestStep> = make_world(QueryId);
+        let context = make_contexts(&world);
+
+        let prss_mul10 = &context[0].participant[TestStep::Mul1(0)];
+        let prss_mul11 = &context[0].participant[TestStep::Mul1(1)];
+        let prss_mul2 = &context[0].participant[TestStep::Mul2];
```
benjaminsavage, line 525: What's Mul1 and Mul2? Why are there two "Mul" operations in "TestStep"? Could we use something named less similarly instead?

src/protocol/sort/shuffle.rs
```diff
@@ -149,6 +149,15 @@ impl<'a, F: Field, S: Step + SpaceIndex> Shuffle<'a, F, S> {
         *self.input = self.single_shuffle(ctx, Step2).await.unwrap();
         *self.input = self.single_shuffle(ctx, Step3).await.unwrap();
 
+        assert_ne!(
+            &ctx.participant[(self.step_fn)(Step1)],
+            &ctx.participant[(self.step_fn)(Step2)]
```
benjaminsavage, line 154: I'm not super familiar with this bit of the infra. Is the "participant" just a PRSS concept?

src/test_fixture/world.rs
```diff
@@ -53,15 +53,15 @@ impl Debug for TestStep {
...
-            TestStep::Mul1(_) => 0,
-            TestStep::Mul2 => 1,
-            TestStep::Reshare(_) => 2,
-            TestStep::Reveal(_) => 3,
-            TestStep::Shuffle(_) => 4,
+            TestStep::Mul1(s) => *s as usize,
+            TestStep::Mul2 => u8_size,
+            TestStep::Reshare(s) => u8_size + 1 + *s as usize,
+            TestStep::Reveal(s) => u8_size * 2 + 1 + *s as usize,
+            TestStep::Shuffle(s) => u8_size * 3 + 1 + s.as_usize(),
```
benjaminsavage, line 64: Yeah, I'm just lost here. No idea what's going on.

benjaminsavage (changes requested):

src/helpers/prss.rs
richajaindce: That would have been ideal :D BUT we dont store the seed anywhere and Aes256 is not comparable. Hence, just compared the generated random values which returns (u128, u128) and chances of their being equal wrongly is very low.

src/protocol/sort/shuffle.rs
richajaindce: yes it is, participant is vector of prss (i.e. for all the steps) whereas prss is for a single step. I think @akoshelev is looking at changing the name of participant to something more obvious.

src/protocol/sort/reshare.rs
richajaindce: Yes lets fix it longer term to use a larger field instead

src/helpers/prss.rs
richajaindce: Using a better example i.e. reshare and mul2

martinthomson (commented):
As we discussed, a lot of this is going to be obsolete with the changes that I'm implementing to our Step interface.  But in the interest of making progress, let's go ahead and do this.  I can rebase my code and remove the affected stuff.

src/test_fixture/world.rs
richajaindce: Leaving it like this since the new changes will remove all complexity from the step implementations!

PR #115, opened 2022-10-12
Unshuffle

richajaindce:


richajaindce:
This also moves generation of random permutation out single shuffle. Permutation will be passed on by the step who calls shuffle/unshuffle

The tests look quite ugly, I will try to handle them in next PR

src/protocol/sort/shuffle.rs
```diff
@@ -31,6 +31,52 @@ pub struct Shuffle<'a, F, S> {
...
+        ((0..batchsize).collect(), (0..batchsize).collect());
+    // shuffle 0..N based on seed
+    permutations
+        .0
+        .shuffle(&mut ChaCha8Rng::from_seed(seed_left.try_into().unwrap()));
+    permutations
+        .1
+        .shuffle(&mut ChaCha8Rng::from_seed(seed_right.try_into().unwrap()));
+
+    //take inverse and return
```
benjaminsavage, line 73: Why do we need to take the inverse? If it's just a random shuffle, does it matter?

src/protocol/sort/shuffle.rs
```diff
@@ -103,70 +119,121 @@ impl<'a, F: Field, S: Step + SpaceIndex> Shuffle<'a, F, S> {
...
+            let permute = if to_helper.peer(Direction::Left) == channel.identity() {
+                &permutations.0
             } else {
-                Direction::Right
+                &permutations.1
             };
-            let mut permute = Self::generate_random_permutation(self.input.len(), direction, prss);
-            apply_inv(&mut permute, &mut self.input);
+
+            if matches!(shuffle_or_unshuffle, ShuffleOrUnshuffle::Shuffle) {
```
benjaminsavage, line 144: Is this the canonical way to do this in Rust? Is it not to do a two line match expression with => operators?

src/protocol/sort/shuffle.rs
```diff
@@ -103,70 +119,121 @@ impl<'a, F: Field, S: Step + SpaceIndex> Shuffle<'a, F, S> {
...
+            .unwrap();
+        *self.input = self
+            .single(ShuffleOrUnshuffle::Shuffle, ctx, Step3, permutations)
+            .await
+            .unwrap();
+
+        Ok(())
+    }
+
+    #[embed_doc_image("unshuffle", "images/sort/unshuffle.png")]
```
benjaminsavage, line 186: This unshuffle diagram confused me more. I'm not sure why only one helper needs to "get permutation" each step. Isn't it two?

src/protocol/sort/shuffle.rs
```diff
@@ -103,70 +119,121 @@ impl<'a, F: Field, S: Step + SpaceIndex> Shuffle<'a, F, S> {
...
+            .unwrap();
+
+        assert_ne!(
+            &ctx.participant[(self.step_fn)(Step1)],
+            &ctx.participant[(self.step_fn)(Step2)]
+        );
+        assert_ne!(
+            &ctx.participant[(self.step_fn)(Step2)],
+            &ctx.participant[(self.step_fn)(Step3)]
+        );
```
benjaminsavage, line 219: Wait... are we in a test? Why are there asserts in production code?

src/protocol/sort/shuffle.rs
```diff
@@ -103,70 +119,121 @@ impl<'a, F: Field, S: Step + SpaceIndex> Shuffle<'a, F, S> {
...
 
-    /// `single_shuffle` is called for the helpers
-    /// i)   2 helpers generate random sequence using shared prss random number
+    /// `single` is called for the helpers
+    /// i)   2 helpers receive permutation pair and choose the permutation to be applied
     /// ii)  2 helpers apply the permutation to their shares
     /// iii) reshare to `to_helper`
-    #[allow(clippy::cast_possible_truncation)]
-    async fn single_shuffle<M: Mesh, G: Gateway<M, S>>(
+    async fn single<M: Mesh, G: Gateway<M, S>>(
```
benjaminsavage, line 126: select a different name. Perhaps: "shuffle_or_unshuffle_once"

src/protocol/sort/shuffle.rs
richajaindce: Removing it for now

src/protocol/sort/shuffle.rs
richajaindce: thanks for the suggestion!

src/protocol/sort/shuffle.rs
richajaindce: made it better based on suggestions

src/protocol/sort/shuffle.rs
richajaindce: doesnt matter, removing it

benjaminsavage (approved):
Thanks Richa!

PR #117, opened 2022-10-13
Secure apply inv : This protocol will apply a secret shared permutation on secret shared input across three helpers

richajaindce:


richajaindce:
FYI although the step structure will change and functions like compose_ipa_step will be not used, still putting this PR out since I am out for next 2 weeks and everyone can see an example (if needed) about whats needed for an intermediate stage of the sort.

richajaindce:
Closing and opening a new one

PR #102, opened 2022-10-06
Fabric interface and its in-memory implementation

akoshelev:
## Overview

This change brings the separation of the existing `mock` module into two parts: 

* `messaging` that takes care of handling messages and communication with protocols (serde, receive requests etc)
* `fabric` that handles connection management and byte movement from helper to helper. 

The core of this change is two traits: `Fabric` and `CommunicationChannel`. The former opens connections to other helpers and provides a stream of messages received from others. `CommunicationChannel` has only one method `send` that is used to send messages to peers.

Previously `mock` module was taking care of all of these tasks in a single tokio task. With the introduction of in-memory implementation of `Fabric` interface (which is `InMemoryEndpoint`), messaging layer event loop which was extracted from `mock` module and renamed to `messaging` now only processes two types of tasks:

* requests to receive a message from given peer
* pull messages from the `Fabric`

The actual networking layer that @thurstonsand is working on will eventually provide the implementation of `Fabric` interface that uses Hyper/Axum to send bytes to helper peers. 

## Goals

The main goal was to provide the same implementation of message handling on top of abstract communication layer that uses in-memory structures to send "messages" between helpers and actual network when MPC ring is formed by separate processes running on the same or different hosts. 

## Performance

I am observing significant improvement (20%+) for circuits with the depth > 1. Other circuits regressed ( <= 500k multiplications) or stayed largely within +- 5%. 

Before

image

After
image

for 64M multiplications (depth 64, width 1M)

Before
image

After
image

It would be great to fully understand why the overall throughput and latency is better which I did not have a chance to do

<details>
<summary>Raw results if someone interested</summary>

```bash
arithmetic/circuit/5000:1
                        time:   [52.211 ms 53.434 ms 54.828 ms]
                        thrpt:  [91.194 Kelem/s 93.573 Kelem/s 95.765 Kelem/s]
                 change:
                        time:   [+10.720% +13.432% +16.843%] (p = 0.00 < 0.05)
                        thrpt:  [-14.415% -11.842% -9.6819%]
                        Performance has regressed.
arithmetic/circuit/5000:10
                        time:   [391.98 ms 395.38 ms 398.66 ms]
                        thrpt:  [125.42 Kelem/s 126.46 Kelem/s 127.56 Kelem/s]
                 change:
                        time:   [-9.9960% -9.1350% -8.2551%] (p = 0.00 < 0.05)
                        thrpt:  [+8.9979% +10.053% +11.106%]
                        Performance has improved.
[...]

PR #122, opened 2022-10-13
Hierarchical step identification

martinthomson:
This works, but needs rebasing and tuning.  And maybe a few tests.


src/protocol/prss.rs
```diff
@@ -151,24 +133,69 @@ impl RngCore for PrssRng {
...
+                    left: self.left.generator(k.as_bytes()),
+                    right: self.right.generator(k.as_bytes()),
+                })
+            })
+        };
+        // Safety: we never drop a PRSS instance while the outer object lives.
+        // As each instance is pinned, it is safe to return a pointer to
+        // each once they are created as long as the pointer is not referenced
+        // past the lifetime the container (see above).
+        NonNull::from(unsafe { Pin::into_inner_unchecked(p.as_ref()) })
```
akoshelev, line 198: I wonder if `Arc` is a safer option. Are you worried about performance implications here?

src/protocol/prss.rs
```diff
@@ -151,24 +133,69 @@ impl RngCore for PrssRng {
...
+            right: KeyExchange::new(r),
+        }
+    }
+
+    /// Get the identified PRSS instance.
+    ///
+    /// # Panics
+    /// When used incorrectly.  For instance, if you ask for an RNG and then ask
+    /// for a PRSS using the same key.
+    pub fn prss(&self, key: impl AsRef<str>) -> &PrssSpace {
```
akoshelev, line 156: wondering if we should wrap `str` into an enum with a member for debug builds (just a string) and maybe an integer for release. Would this API allow non-string keys without additional overhead?

src/helpers/mesh.rs
```diff
@@ -52,12 +53,15 @@ pub trait Mesh {
 }
 
 /// This is the entry point for protocols to request communication when they require it.
-pub trait Gateway<M: Mesh, S: Step> {
-    /// Create or return an existing channel for a given step. Protocols can send messages to
-    /// any helper through this channel (see `Mesh` interface for details).
+pub trait Gateway {
```
akoshelev, line 56: rebase will be painful here, #102 changed a lot here

src/protocol/mod.rs
```diff
@@ -1,65 +1,145 @@
...
+}
 
-/// Set of steps that define the IPA protocol.
-#[derive(Copy, Clone, PartialEq, Eq, Hash)]
-pub enum IPAProtocolStep {
-    /// Convert from XOR shares to Replicated shares
-    ConvertShares,
-    /// Sort shares by the match key
-    Sort(SortStep),
+unsafe impl Send for UniqueStepId {}
```
akoshelev, line 66: hmm, do you really need it? everything looks Send to me

src/protocol/mod.rs
```diff
@@ -1,65 +1,145 @@
...
+                "Refined '{}' with step '{}' twice",
+                self.id,
+                step.as_ref(),
+            );
+        }
+
+        Self {
+            id: self.id.clone() + "/" + step.as_ref(),
+            #[cfg(debug_assertions)]
+            used: Arc::new(Mutex::new(HashSet::new())),
```
akoshelev, line 101: You probably wanted to clone it from `self`

src/protocol/securemul.rs
```diff
@@ -18,15 +21,20 @@ pub struct DValue<F> {
 /// for use with replicated secret sharing over some field F.
 /// K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, and B. Pinkas. High-throughput secure AES computation. In WAHC@CCS 2018, pp. 13–24, 2018
 #[derive(Debug)]
-pub struct SecureMul<'a, G, S> {
+pub struct SecureMul<'a, G> {
     prss: &'a PrssSpace,
     gateway: &'a G,
-    step: S,
+    step: &'a UniqueStepId,
```
akoshelev, line 27: we use step and step id interchangeably throughout the code for `UniqueStepId`. Is it going to cause some confusion and should we use `step_id` for it instead?

akoshelev (commented):
I like this change a lot - getting rid of complex `Step` hierarchy makes the protocol code simpler

src/protocol/mod.rs
```diff
@@ -1,65 +1,145 @@
...
-/// it should converge to a single implementation.
-pub trait Step: Copy + Clone + Debug + Eq + Hash + Send + 'static {}
+/// Note that the implementation of this context might change to use a different
+/// (possible more efficient) representation.  It is probably not particularly efficient
+/// to be cloning this object all over the place.  Of course, a string is pretty useful
+/// from a debugging perspective.
+#[derive(Clone, Debug)]
+pub struct UniqueStepId {
+    id: String,
+    /// This tracks the different values that have been provided to `refine()`.
```
richajaindce, line 61: same, this should be narrow()

src/protocol/securemul.rs
martinthomson: We've used "step" far more frequently.  I think that it is probably better to change the name we use for the `Step` type instead.

src/protocol/mod.rs
martinthomson: No, this was deliberate.  Once this is refined down to the next level, you want a clean slate.

src/protocol/prss.rs
martinthomson: Yes, I think that is the next step, where we work out how to take the unique ID and turn it into a flat number (for which copying should be very cheap) for release builds.

src/protocol/prss.rs
martinthomson: Let me see if I can make that work.  It seems like it might now that I think on it some more, even if I don't think that we really want the reference counting that comes with `Arc` (as the lifetime of the reference we give out is still bounded by the lifetime on `&self`), but it might avoid the ugliness.

src/protocol/mod.rs
thurstonsand: meaning you might want to reuse steps at different levels?

src/protocol/securemul.rs
thurstonsand: Should we also change `UniqueStepId` to `Step` since that is the more frequently used type? that would also align with the `step` var name

src/protocol/securemul.rs
```diff
@@ -270,48 +260,55 @@ pub mod tests {
...
-        let record_id = RecordId::from(1);
 
-        for i in 1..10_u8 {
+        for step in 1..10_u8 {
+            let a = share(Fp31::from(4_u128), &mut rand);
+            let b = share(Fp31::from(3_u128), &mut rand);
+
             // there is something weird going on the compiler's side. I don't see why we need
             // to use async move as `i` is Copy + Clone, but compiler complains about it not living
             // long enough
```
thurstonsand, line 289:
```suggestion
```

src/test_fixture/circuit.rs
```diff
@@ -29,21 +29,56 @@ pub async fn arithmetic<F: Field>(width: u32, depth: u8) {
...
+            "b186", "b187", "b188", "b189", "b190", "b191", "b192", "b193", "b194", "b195", "b196",
+            "b197", "b198", "b199", "b200", "b201", "b202", "b203", "b204", "b205", "b206", "b207",
+            "b208", "b209", "b210", "b211", "b212", "b213", "b214", "b215", "b216", "b217", "b218",
+            "b219", "b220", "b221", "b222", "b223", "b224", "b225", "b226", "b227", "b228", "b229",
+            "b230", "b231", "b232", "b233", "b234", "b235", "b236", "b237", "b238", "b239", "b240",
+            "b241", "b242", "b243", "b244", "b245", "b246", "b247", "b248", "b249", "b250", "b251",
+            "b252", "b253", "b254", "b255",
+        ];
+        let i = usize::from(self.0);
+        assert!(i < BIT_NAMES.len());
```
thurstonsand, line 62: if you change `BIT_NAMES` to a `[&str; 256]` type, couldn't you remove this assert?

src/protocol/prss.rs
akoshelev: I was mostly concerned about use-after-free, this code is not ugly.

akoshelev (approved):
lgtm, but I haven't thoroughly reviewed protocol changes. I think we can push this and iterate if needed

src/protocol/mod.rs
```diff
@@ -1,67 +1,147 @@
...
+            #[cfg(debug_assertions)]
+            used: Arc::new(Mutex::new(HashSet::new())),
+        }
+    }
+}
+
+impl Default for UniqueStepId {
+    // TODO(mt): this should might be better if it were to be constructed from
+    // a QueryId rather than using a default.
+    fn default() -> Self {
```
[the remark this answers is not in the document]
martinthomson: If you are only doing one thing (i.e., at most one communication or one lot of shared randomness), then you can avoid using `narrow`.  You will see this in a few of the tests.

src/protocol/context.rs
```diff
@@ -1,46 +1,85 @@
...
+    pub async fn multiply(&self, record_id: RecordId) -> SecureMul<'_, N> {
+        SecureMul::new(self.prss(), self.gateway, &self.step, record_id)
     }
 
     /// Request reveal for a given record.
     #[allow(clippy::unused_async)] // eventually there will be await b/c of backpressure implementation
-    pub fn reveal(&'a self, record_id: RecordId, step: S) -> Reveal<'a, F, S> {
-        Reveal::new(self.gateway, step, record_id)
+    #[must_use]
+    pub fn reveal(&self, record_id: RecordId) -> Reveal<'_, '_, N> {
```
benjaminsavage, line 82: Should this similarly be made async for the same back-pressure mechanism as stated above?

Not really related to this diff in particular, but just something that occurs to me.

src/protocol/mod.rs
```diff
@@ -1,67 +1,147 @@
...
+///
+/// Any stage of the protocol execution will involve multiple steps.  Each of these steps
+/// then might involve executing a process that can be broken down into further steps.
+/// Ultimately, there will be processes that need to invoke functions on a PRSS or send
+/// data to another helper that needs to be uniquely identified.
+///
+/// Steps are therefore composed into a hierarchy where top-level steps describe major
+/// building blocks for a protocol (such as sort shares, convert shares, apply DP, etc...),
+/// intermediate processes describe reusable processes (like shuffling), and steps at the
+/// lowest level unique identify multiplications.
```
benjaminsavage, line 25: %s/unique/uniquely

benjaminsavage (approved):
Thanks @martinthomson - this is a huge quality of life improvement! 

There are a few outstanding comments, but I'd like to merge this in so that @taikiy and I can rebase on top of this. We can continue additional improvements in other diffs. This is a big one though, and I think it's probably good to get it landed before more changes come along and force martin to rebase yet again!

PR #125, opened 2022-10-13
Moving Reveal protocol out of sort

benjaminsavage:
Reveal is also used in a variety of other protocols, including modulus conversion.


richajaindce:
Looks good! This should have been outside from the start :/ also, similar to multiplication having multiply_all for entire vector, we would also have reveal_all which might be useful for other protocols.

richajaindce (approved):
Looks good! This should have been outside from the start :/ also, similar to multiplication having multiply_all for entire vector, we would also have reveal_all which might be useful for other protocols.

richajaindce (approved):

akoshelev (approved):

PR #126, opened 2022-10-14
Pass out `Arc<Prss>` instead of `&Prss`

martinthomson:
This avoids using unsafe code, but I'm not sure if it is better as it adds a bunch of overhead.  I'm going to run some tests to see if this has any real effect in practice.

### DIFF
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -7,6 +7,7 @@ use crate::{
     },
     protocol::{prss::Participant, reveal::Reveal},
 };
+use std::sync::Arc;
 
 /// Context used by each helper to perform computation. Currently they need access to shared
 /// randomness generator (see `Participant`) and communication trait to send messages to each other.
@@ -55,7 +56,7 @@ impl<'a, N> ProtocolContext<'a, N> {
 
     /// Get the PRSS instance for this step.
     #[must_use]
-    pub fn prss(&self) -> &PrssSpace {
+    pub fn prss(&self) -> Arc<PrssSpace> {
         self.participant.prss(&self.step)
     }
 }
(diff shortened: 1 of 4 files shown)


akoshelev (approved):
would definitely be interesting to compare these two implementations

martinthomson:
I'm getting so much noise in testing that I can't tell if my original patch is altering performance characteristics.  That leaves me at a loss to determine what this one does.  My ancient CPU clearly doesn't like this code very much.

akoshelev:
> I'm getting so much noise in testing that I can't tell if my original patch is altering performance characteristics. That leaves me at a loss to determine what this one does. My ancient CPU clearly doesn't like this code very much.

Let me do this - we have 3 AWS hosts that we can use to benchmark

akoshelev:
I am getting an error trying to compile this branch - looks like it is not fully ready to be merged

```bash
error[E0277]: the trait bound `std::string::String: protocol::Step` is not satisfied
  --> src/test_fixture/circuit.rs:42:49
   |
42 |         let bit_ctx = narrow_contexts(&top_ctx, &format!("bit{}", bit));
   |                       ---------------           ^^^^^^^^^^^^^^^^^^^^^^ the trait `protocol::Step` is not implemented for `std::string::String`
   |                       |
   |                       required by a bound introduced by this call
   |
   = help: the following other types implement trait `protocol::Step`:
             IpaProtocolStep
             ShuffleStep
             SortStep
             StreamingStep
note: required by a bound in `narrow_contexts`
  --> src/test_fixture/mod.rs:45:17
   |
43 | pub fn narrow_contexts<'a>(
   |        --------------- required by a bound in this
44 |     contexts: &[ProtocolContext<'a, Arc<InMemoryEndpoint>>; 3],
45 |     step: &impl Step,
   |                 ^^^^ required by this bound in `narrow_contexts`
```

martinthomson:
Right.  I just need to rebase onto main.  Which is on my TODO list.  The hack I have for the circuit is in place.

martinthomson:
Lumping this in with the other one now to save on conflict management (this was already a bad merge and I had to do another hard merge on #127).

akoshelev:
after #128 is merged I can run the benchmarks again. For some reason criterion wasn't showing me the diff between two runs :(

PR #127, opened 2022-10-14
PRSS cleanup

martinthomson:
I think that these names are better.  Outside of the module, there isn't much that changes.

### DIFF
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -48,7 +48,7 @@ impl<'a, N> ProtocolContext<'a, N> {
     /// Make a sub-context.
     /// Note that each invocation of this should use a unique value of `step`.
     #[must_use]
-    pub fn narrow<S: Step>(&self, step: &S) -> Self {
+    pub fn narrow<S: Step + ?Sized>(&self, step: &S) -> Self {
         ProtocolContext {
             role: self.role,
             step: self.step.narrow(step),
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use super::{
     prss::{IndexedSharedRandomness, SequentialSharedRandomness},
     securemul::SecureMul,
@@ -64,7 +66,7 @@ impl<'a, N> ProtocolContext<'a, N> {
     /// If `prss_rng()` is invoked for the same context, this will panic.  Use of
     /// these two functions are mutually exclusive.
     #[must_use]
-    pub fn prss(&self) -> &IndexedSharedRandomness {
+    pub fn prss(&self) -> Arc<IndexedSharedRandomness> {
         self.prss.indexed(&self.step)
     }
 
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -1,11 +1,15 @@
-use super::{prss::PrssSpace, securemul::SecureMul, RecordId, Step, UniqueStepId};
+use super::{
+    prss::{IndexedSharedRandomness, SequentialSharedRandomness},
+    securemul::SecureMul,
+    RecordId, Step, UniqueStepId,
+};
 use crate::{
     helpers::{
         fabric::Network,
         messaging::{Gateway, Mesh},
         Identity,
     },
-    protocol::{prss::Participant, reveal::Reveal},
+    protocol::{prss::Endpoint as PrssEndpoint, reveal::Reveal},
 };
 
 /// Context used by each helper to perform computation. Currently they need access to shared
@@ -15,16 +19,16 @@ use crate::{
 pub struct ProtocolContext<'a, N> {
     role: Identity,
     step: UniqueStepId,
-    participant: &'a Participant,
+    prss: &'a PrssEndpoint,
     gateway: &'a Gateway<N>,
 }
 
 impl<'a, N> ProtocolContext<'a, N> {
-    pub fn new(role: Identity, participant: &'a Participant, gateway: &'a Gateway<N>) -> Self {
+    pub fn new(role: Identity, participant: &'a PrssEndpoint, gateway: &'a Gateway<N>) -> Self {
         Self {
             role,
             step: UniqueStepId::default(),
-            participant,
+            prss: participant,
             gateway,
         }
     }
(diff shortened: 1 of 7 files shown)


src/protocol/context.rs
```diff
@@ -55,7 +55,7 @@ impl<'a, N> ProtocolContext<'a, N> {
 
     /// Get the PRSS instance for this step.
     #[must_use]
-    pub fn prss(&self) -> &PrssSpace {
+    pub fn prss(&self) -> &SharedRandomness {
```
akoshelev, line 58: do you plan to rename method name as well?

akoshelev (approved):

src/protocol/context.rs
martinthomson: Hmm, what we're really looking at is IndexedSharedRandomness (as opposed to SequentialSharedRandomness for the rand::Rng implementation).  Maybe I should revisit this again.

src/protocol/prss.rs
```diff
@@ -219,14 +247,14 @@ impl ParticipantSetup {
     /// This is generic over `SpaceIndex` so that it can construct the
     /// appropriate number of spaces for use by a protocol participant.
```
akoshelev, line 248:
```suggestion
```

akoshelev: maybe we should rename `ParticipantSetup` to `PrssEndpointSetup` to be consistent with naming?

src/protocol/prss.rs
```diff
@@ -153,49 +146,84 @@ impl Participant {
...
         // This makes the lookup perform an allocation, which is very much suboptimal.
-        let p = if let Some(value) = self.spaces.get(key) {
-            value
+        let item = if let Some(item) = self.items.get(key) {
+            item
         } else {
-            self.spaces.entry(key.to_owned()).or_insert_with_key(|k| {
-                Box::pin(PrssSpace {
+            self.items.entry(key.to_owned()).or_insert_with_key(|k| {
+                EndpointItem::Indexed(Box::pin(IndexedSharedRandomness {
```
akoshelev, line 197: probably not for this PR, but I wonder if we should return a drop guard instead of reference to the value? I worry that we would keep all prss around for the duration of the query which may be billions if we have unique step per record. If the value returned is dropped, we can follow the backpointer to `Endpoint` and remove it from the map

src/protocol/prss.rs
martinthomson: Hmm, didn't I do that?  Maybe I did that in the other PR.  I lose track.

martinthomson:
Updated the name of the PR to reflect the changed scope.

akoshelev:
Benchmarking is complete and `Arc<Prss>` is a clear winner (`raw-ptr` is the version with references, `new` is `arc`)

```bash
[ec2-user@ip-172-31-57-95 raw-ipa]$ critcmp raw-ptr new
group                            new                                    raw-ptr
-----                            ---                                    -------
arithmetic/circuit/1000000:1     1.00      12.5±0.19s 77.8 KElem/sec    1.00      12.5±0.17s 78.0 KElem/sec
arithmetic/circuit/1000000:10    1.00     118.3±0.57s 82.5 KElem/sec    1.02     120.4±0.54s 81.1 KElem/sec
arithmetic/circuit/1000000:64    1.00     706.7±3.81s 88.4 KElem/sec    1.02     723.6±5.21s 86.4 KElem/sec
arithmetic/circuit/500000:1      1.00       6.1±0.06s 80.1 KElem/sec    1.00       6.1±0.05s 79.8 KElem/sec
arithmetic/circuit/500000:10     1.00      58.3±1.25s 83.8 KElem/sec    1.01      58.9±0.42s 82.9 KElem/sec
arithmetic/circuit/500000:64     1.00     350.7±1.36s 89.1 KElem/sec    1.03     362.7±1.88s 86.2 KElem/sec
arithmetic/circuit/50000:1       1.00    595.3±9.35ms 82.0 KElem/sec    1.01   603.6±12.34ms 80.9 KElem/sec
arithmetic/circuit/50000:10      1.00       5.6±0.03s 87.4 KElem/sec    1.02       5.7±0.04s 85.4 KElem/sec
arithmetic/circuit/50000:64      1.00      34.2±0.16s 91.5 KElem/sec    1.02      34.8±0.18s 89.7 KElem/sec
arithmetic/circuit/5000:1        1.00     56.0±1.32ms 87.3 KElem/sec    1.04     58.2±0.94ms 83.9 KElem/sec
arithmetic/circuit/5000:10       1.00    514.5±3.71ms 94.9 KElem/sec    1.05    541.7±2.80ms 90.1 KElem/sec
arithmetic/circuit/5000:64       1.00       3.2±0.01s 98.8 KElem/sec    1.03       3.3±0.02s 96.0 KElem/sec
```

akoshelev:
Screen Shot 2022-10-19 at 5 37 23 PM

src/protocol/prss.rs
```diff
@@ -153,59 +150,91 @@ impl Participant {
...
-                Box::pin(PrssSpace {
+            self.items.entry(key.to_owned()).or_insert_with_key(|k| {
+                EndpointItem::Indexed(Arc::new(IndexedSharedRandomness {
                     left: self.left.generator(k.as_bytes()),
                     right: self.right.generator(k.as_bytes()),
-                })
+                }))
             })
         };
         // Safety: we never drop a PRSS instance while the outer object lives.
```
akoshelev, line 204: this comment seems to be outdated now

src/protocol/prss.rs
```diff
@@ -153,59 +150,91 @@ impl Participant {
...
         // This makes the lookup perform an allocation, which is very much suboptimal.
-        let p = if let Some(value) = self.spaces.get(key) {
-            value
+        let item = if let Some(item) = self.items.get(key) {
+            item
         } else {
-            self.spaces.entry(key.to_owned()).or_insert_with_key(|k| {
-                Box::pin(PrssSpace {
+            self.items.entry(key.to_owned()).or_insert_with_key(|k| {
+                EndpointItem::Indexed(Arc::new(IndexedSharedRandomness {
```
akoshelev, line 198: i still feel a little nervous about keeping all prss instances forever, but I guess we can change it later?

akoshelev (approved):

src/protocol/prss.rs
martinthomson: They will be reclaimed once the query is over, so it's not so bad.  We don't otherwise have any signal we can use to remove them, but if we ever get one, then we can do something about it.

src/protocol/prss.rs
martinthomson:
```suggestion
```

PR #128, opened 2022-10-14
Implemented modulus conversion

benjaminsavage:
Assume that the browser will secret share a match key into XOR secret shares, with each share encrypted towards a different helper.

Upon decryption, each helper will possess just one share (i.e. not a replicated sharing).

Assume we interpret this as a bit-packed array of secret sharings in Z_2. Basically, for each of the bits of the match key, there are three shares, each one boolean, and each helper possesses just one of them.

This ConvertShares protocol takes in a u64 (which is interpreted as a bit-packing of up to 64 of these Z_2 secret sharings) and converts this to a vector of secret-sharings in a larger field, where each element of the vector is a sharing of either zero or one.

Example:

### Match key: 
`0b11001`
### Output:
```
vec![ 
    Replicated<F>, // a random sharing of `1` 
    Replicated<F>, // a random sharing of `1` 
    Replicated<F>, // a random sharing of `0`
    Replicated<F>, // a random sharing of `0`
    Replicated<F>, // a random sharing of `1`
]
```

### DIFF
--- a/src/protocol/modulus_conversion/convert_shares.rs
+++ b/src/protocol/modulus_conversion/convert_shares.rs
@@ -1,0 +1,180 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::{fabric::Network, prss::SpaceIndex},
+    protocol::{
+        context::ProtocolContext,
+        modulus_conversion::gen_random::{GenRandom, ReplicatedBinary},
+        reveal_additive_binary::RevealAdditiveBinary,
+        RecordId, Step,
+    },
+    secret_sharing::Replicated,
+};
+use futures::future::try_join_all;
+
+pub struct XorShares {
+    num_bits: u8,
+    packed_bits: u64,
+}
+
+pub struct ConvertShares {
+    input: XorShares,
+}
+
+///
+/// This is an implementation of
+/// Protocol 5.2 Modulus-conversion protocol from `Z_2^u` to `Z_p`
+/// from the paper <https://eprint.iacr.org/2018/387.pdf>
+///
+/// It works by generating two secret-sharings of a random number `r`,
+/// one in `Z_2`, the other in `Z_p`. The sharing in `Z_2` is subtracted
+/// from the input and the result is revealed.
+///
+/// If the revealed result is `0`, that indicates that `r` had the same value
+/// as the secret input, so the sharing in `Z_p` is returned.
+/// If the revealed result is a `1`, that indicate that `r` was different than
+/// the secret nput, so the sharing 1 - the sharing in `Z_p` is returned.
+impl ConvertShares {
+    #[allow(dead_code)]
+    pub fn new(input: XorShares) -> Self {
+        Self { input }
+    }
+
+    #[allow(dead_code)]
+    pub async fn execute<F: Field, S: Step + SpaceIndex, N: Network<S>>(
+        &self,
+        ctx: &ProtocolContext<'_, S, N>,
+        record_id: RecordId,
+        step0: S,
+        step1: S,
+        step2: S,
+        step3: S,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let prss = &ctx.participant[step0];
+        let (left, right) = prss.generate_values(record_id.into());
+
+        let futures = (0..self.input.num_bits).into_iter().map(|i| async move {
+            let inner_record_id = RecordId::from(
+                u32::from(record_id) * u32::from(self.input.num_bits) + u32::from(i),
+            );
+            let b0 = left & (1 << i) != 0;
+            let b1 = right & (1 << i) != 0;
+
+            let input = self.input.packed_bits & (1 << i) != 0;
+
+            let input_xor_r = input ^ b0;
+
+            let r_binary = ReplicatedBinary::new(b0, b1);
+
+            let r_big_field: Replicated<F> = GenRandom::new(r_binary)
+                .execute(ctx, inner_record_id, step1, step2)
+                .await?;
+
+            let revealed_output =
+                RevealAdditiveBinary::execute(ctx, step3, inner_record_id, input_xor_r).await?;
+
+            if revealed_output {
+                Ok(Replicated::<F>::one(ctx.identity) - r_big_field)
+            } else {
+                Ok(r_big_field)
+            }
+        });
+        try_join_all(futures).await
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        field::{Field, Fp31},
+        protocol::{
+            modulus_conversion::convert_shares::{ConvertShares, XorShares},
+            QueryId, RecordId, SpaceIndex, Step,
+        },
+        test_fixture::{make_contexts, make_world, validate_and_reconstruct, TestWorld},
+    };
+    use futures::future::try_join_all;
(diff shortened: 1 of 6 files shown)


src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -0,0 +1,180 @@
...
+        step0: S,
+        step1: S,
+        step2: S,
+        step3: S,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let prss = &ctx.participant[step0];
+        let (left, right) = prss.generate_values(record_id.into());
+
+        let futures = (0..self.input.num_bits).into_iter().map(|i| async move {
+            let inner_record_id = RecordId::from(
```
akoshelev, line 57: This looks hacky to me - is it possible that the upcoming `Step` change fixes this? It still makes more sense to me to move bit number to the step - that avoids u32 overflow here (the consequence of it would be two conversions generating the same `inner_record_id` )

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -0,0 +1,180 @@
...
+
+            let input_xor_r = input ^ b0;
+
+            let r_binary = ReplicatedBinary::new(b0, b1);
+
+            let r_big_field: Replicated<F> = GenRandom::new(r_binary)
+                .execute(ctx, inner_record_id, step1, step2)
+                .await?;
+
+            let revealed_output =
```
akoshelev, line 73: can we run this in parallel with `GenRandom`?

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage: Yeah, I wasn't sure what was the best way to use the infra here. I'm open to whatever you and @martinthomson think best here.

Honestly, there's no dependency between the record being processed and the production of this "double random". In fact, if we wanted to, we could save query-latency by just pre-generating a gigantic batch of "double randoms" and saving them to disk. We could run that at non-peak times. Then, when we need to process a query we could just take them from the bank.

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage: Yes. I'll make it parallel for now, but long-term we should contemplate pre-generating them at non-peak times.

src/protocol/reveal_additive_binary.rs
```diff
@@ -0,0 +1,103 @@
...
+    #[allow(dead_code)]
+    pub async fn execute<S: Step + SpaceIndex, N: Network<S>>(
+        ctx: &ProtocolContext<'_, S, N>,
+        step: S,
+        record_id: RecordId,
+        input: bool,
+    ) -> Result<bool, BoxError> {
+        let mut channel = ctx.gateway.get_channel(step);
+
+        channel
```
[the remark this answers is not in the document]
benjaminsavage: We definitely should. I was just being lazy =). I'll fix it.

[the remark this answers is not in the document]
benjaminsavage: No can do.
```
error[E0499]: cannot borrow `channel` as mutable more than once at a time
  --> src/protocol/reveal_additive_binary.rs:44:33
   |
37 |           let send_left_future = channel
   |  ________________________________-
38 | |             .send(
39 | |                 channel.identity().peer(Direction::Left),
40 | |                 record_id,
41 | |                 RevealValue { share: input },
42 | |             );
   | |_____________- first mutable borrow occurs here
43 |
44 |           let send_right_future = channel
   |  _________________________________^
45 | |             .send(
46 | |                 channel.identity().peer(Direction::Right),
47 | |                 record_id,
48 | |                 RevealValue { share: input },
49 | |             );
   | |_____________^ second mutable borrow occurs here
50 |
51 |           try_join(send_left_future, send_right_future).await?;
   |                    ---------------- first borrow later used here
```

help... is this something I can work around? Or is this a limitation in the infra we will need to address in a latter diff?

src/protocol/reveal_additive_binary.rs
```diff
@@ -0,0 +1,102 @@
...
+            RevealValue { share: input },
+        );
+
+        let send_right_future = channel.send(
+            channel.identity().peer(Direction::Right),
+            record_id,
+            RevealValue { share: input },
+        );
+
+        try_join(send_left_future, send_right_future).await?;
```
benjaminsavage, line 49: This *does not work*. 

The error is:

```
error[E0499]: cannot borrow `channel` as mutable more than once at a time
  --> src/protocol/reveal_additive_binary.rs:43:33
   |
37 |           let send_left_future = channel.send(
   |  ________________________________-
38 | |             channel.identity().peer(Direction::Left),
39 | |             record_id,
40 | |             RevealValue { share: input },
41 | |         );
   | |_________- first mutable borrow occurs here
42 |
43 |           let send_right_future = channel.send(
   |  _________________________________^
44 | |             channel.identity().peer(Direction::Right),
45 | |             record_id,
46 | |             RevealValue { share: input },
47 | |         );
   | |_________^ second mutable borrow occurs here
48 |
49 |           try_join(send_left_future, send_right_future).await?;
   |                    ---------------- first borrow later used here
```

Is this something I can work around? Or does this require an infra-layer change?

src/protocol/reveal_additive_binary.rs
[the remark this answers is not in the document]
benjaminsavage: OK, 
So I tried to create two different channels, so that I could send in parallel. It does not work. The subsequent "receive" never finishes. It just gets stuck forever.

I'm giving up on this... I hope someone else can help me make this parallel in a follow-up diff.

src/protocol/reveal_additive_binary.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        let mut bools: Vec<bool> = Vec::with_capacity(40);
+
+        let inputs = (0..10).into_iter().map(|i| {
+            let b0 = rng.gen::<bool>();
+            let b1 = rng.gen::<bool>();
+            let b2 = rng.gen::<bool>();
+            bools.push((b0 ^ b1) ^ b2);
+
+            (i, b0, b1, b2, c0.clone(), c1.clone(), c2.clone())
```
benjaminsavage, line 93: It is a little funky that we need to clone the contexts for each record...

src/protocol/reveal_additive_binary.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        // Sleep until `helper's right` sends their share
+        let share_from_right: RevealValue = channel
+            .receive(channel.identity().peer(Direction::Right), record_id)
+            .await?;
+
+        // Sleep until `helper's left` sends their share
+        let share_from_left: RevealValue = channel
+            .receive(channel.identity().peer(Direction::Left), record_id)
+            .await?;
```
benjaminsavage, line 61: It also really feels like this should be able to be done in parallel. As I mentioned above, I attempted to do so by using two channels (channel_left and channel_right), but it didn't work.

src/protocol/reveal_additive_binary.rs
akoshelev: oh, that's my bad - infra does not support it yet. #132 should fix it

PR #130, opened 2022-10-15
Move send buffers from network to infra layer

akoshelev:
As discussed offline with Martin, we want to keep the network layer simple. That means that all the buffering logic on send and receive should be handled by the infrastructure layer. Receive buffers already exist there, this change lifts the send buffers to it.

Right now the implementation is dumb and just accepts whatever the protocols send to it. There is no accepted record id interval or any other mechanism to block the protocol if it advanced too far. That will be the next step.

The main purpose of this change is to get us to agree on the API interface between infra and network. It is modelled as `Network` trait that provides two methods:

* `sink` - returns the sending side of the channel. All messages directed towards the other helper parties must be written into it.`Sink` interface is a quite power abstraction with flow control and flush mechanisms built-in. There is no use of them yet, but we may decide to just buffer everything inside the `Sink`.
* `stream` - the incoming flow of messages sent by other helpers.

Note that it is possible and maybe even desirable to require callsites to provide `channel_id` argument to these methods. That would mean keeping a logical "channel" interface between infra and network. This would complicate the implementation of both layers, but may be deemed necessary to implement as it better represents the real world of IPA where multiple channels are used at the same time and it is preferred to avoid contention between different channels.


src/helpers/fabric.rs
```diff
@@ -24,23 +24,15 @@ pub type MessageChunks = (ChannelId, Vec<MessageEnvelope>);
...
-    type Channel: CommunicationChannel;
+    type Sink: futures::Sink<MessageChunks, Error = Error> + Send + Unpin + 'static;
     type MessageStream: Stream<Item = MessageChunks> + Send + Unpin + 'static;
 
-    /// Returns a new connection to be open. `channel_id` indicates the parameters of this
-    /// connection (destination peer and step). Once the connection is returned it is immediately
-    /// ready for sending messages.
-    async fn get_connection(&self, channel_id: ChannelId) -> Self::Channel;
+    /// Returns a sink that accepts data to be sent to other helper parties.
+    fn sink(&self) -> Self::Sink;
```
martinthomson, line 31: I think that we want this name to be a bit clearer in terms of its purpose.  send_stream(), perhaps?

src/helpers/fabric.rs
```diff
@@ -24,23 +24,15 @@ pub type MessageChunks = (ChannelId, Vec<MessageEnvelope>);
...
     /// Returns a stream to receive messages that have arrived from other helpers. Note that
     /// some implementations may panic if this method is called more than once.
-    fn message_stream(&self) -> Self::MessageStream;
-}
-
-#[async_trait]
-pub trait CommunicationChannel {
-    /// Send a given message
-    async fn send(&self, msg: MessageEnvelope) -> Result<(), Error>;
+    fn stream(&self) -> Self::MessageStream;
```
martinthomson, line 35: recv_stream() to match?

src/helpers/buffers.rs
```diff
@@ -0,0 +1,139 @@
...
+        (channel_id, data)
+    }
+}
+
+impl ReceiveBuffer {
+    pub fn new() -> Self {
+        Self {
+            inner: HashMap::default(),
+        }
+    }
```
martinthomson, line 88: derive(Default) ?

src/helpers/messaging.rs
```diff
@@ -143,46 +121,80 @@ impl<N: Network> Mesh<'_, '_, N> {
...
                     }
                     Some((channel_id, messages)) = message_stream.next() => {
                         tracing::trace!("received {} message(s) from {:?}", messages.len(), channel_id);
-                        buf.entry(channel_id)
-                           .or_default()
-                           .receive_messages(messages);
+                        receive_buf.receive_messages(&channel_id, messages);
+                    }
+                    Some((channel_id, msg)) = envelope_rx.recv() => {
+                        if let Some(buf_to_send) = send_buf.push(channel_id.clone(), msg) {
```
martinthomson, line 168: So I take it that reordering is a TBD?  From what I can see, you are sending things in arbitrary order on each channel.

src/helpers/buffers.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Process message that has been received
+    pub fn receive_messages(&mut self, channel_id: &ChannelId, messages: Vec<MessageEnvelope>) {
+        for msg in messages {
+            match self
+                .inner
+                .entry(channel_id.clone())
+                .or_default()
+                .entry(msg.record_id)
```
martinthomson, line 121: This bothers me a little.  We might want to consider adding new structs rather than having this one work its way down through two hashmaps (do this later).

src/helpers/messaging.rs
```diff
@@ -143,46 +121,80 @@ impl<N: Network> Mesh<'_, '_, N> {
...
+        let mut message_stream = network.stream();
+        let mut network_sink = network.sink();
+
+        let control_handle = tokio::spawn(async move {
+            // to make forward progress, we periodically check if the system is stalled
+            // if nothing happens for long period of time, we try to unblock it by flushing
+            // the data that remains inside buffers. Note that the interval picked here is somewhat
+            // random - waiting for too long will result in elevated latencies. On the other hand,
+            // sending buffers that are half-full will lead to underutilizing the network
+            const INTERVAL: Duration = Duration::from_millis(200);
```
martinthomson, line 145: How likely do you believe this to be?  Maybe instead add a flush instruction for a channel that you can use in those cases where everything is done (but you might not have a full batch to send).

Right now, it looks like you are going to need this for every test execution, which I don't like.

martinthomson (approved):
Starting to take shape.  Go ahead and land once you are happy with it.

I see some major pieces that you might want to track using issues: reordering on the send side being fairly important, but also the `_network` thing.

src/helpers/messaging.rs
akoshelev: yea the current mitigation is to set `batch size = 1` in unit tests (happens in `Default` implementation for test config). If batch size is 1, we don't buffer anything and therefore don't rely on this

src/helpers/messaging.rs
akoshelev: correct, `send_buf` is dumb and just accumulates up to `batch_size` records before sending them down to network layer. I hope we don't need to change this particular piece of code though because send buffers should take care of ordering.

There is another important piece that is missing here - window size update. We should probably slow down producers that went too far ahead by looking in `record_id` and blocking their `send` if `record_id` is outside of accepted range. I think we need some cooperation between send buffers and the gateway to make this happen.

I will add a TODO so we don't forget about it

src/helpers/buffers.rs
akoshelev: eh, I've added it but never used it :(

src/helpers/buffers.rs
akoshelev: same, I don't like nested maps as well, but hopefully we can get rid of the second map and replace it with a vec indexed by record_id + offset

src/helpers/messaging.rs
martinthomson: A batch size of 1 works pretty nicely, yeah.

PR #131, opened 2022-10-16
Remove identity method from `Mesh`

akoshelev:
Given the conversation in #110 [comment](https://github.com/martinthomson/raw-ipa/pull/110/files#r992706026) we don't seem to need to dynamically change the identities of helpers dynamically. Therefore we don't need to force client code to obtain a reference to the channel first.

The standard way to access helper identity should be via `ProtocolContext` and, ~in some cases, via `Gateway`~. No access through `Gateway` is required anymore


src/protocol/securemul.rs
```diff
@@ -69,15 +69,18 @@ impl<'a, S: Step, N: Network<S>> SecureMul<'a, S, N> {
         // notify helper on the right that we've computed our value
         channel
             .send(
-                channel.identity().peer(Direction::Right),
+                self.gateway.identity().peer(Direction::Right),
```
benjaminsavage, line 72: Honestly, I think the better solution here is to just change the internal state that SecureMul keeps around. Nothing else keeps a special pointer to the gateway and the prss, everything else just gets what it needs directly from the ProtocolContext. I'd recommend updating SecureMul to do the same.

benjaminsavage (approved):

src/protocol/securemul.rs
akoshelev: I am leaning towards the same conclusion, but want to postpone that change until I talk to Martin about it - we had some considerations for this struct that led to this API. For now, I removed `identity` method from `Gateway` and just keep the role inside `SecureMul`

martinthomson (approved):

PR #132, opened 2022-10-17
`Mesh` methods take shared reference

akoshelev:
There is currently no need to take an exclusive reference to self. Given that `Mesh` is no longer a trait, this change is fairly easy to make.

Taking a shared reference instead of `&mut self` enables protocols to run multiple concurrent multiplications at the same time.

The use case for it is presented in #128


martinthomson:
So we need to be careful about how we end up identifying those channels when it comes to the networking layer.  Simultaneous access can fail in surprising ways.  That said, if all problems could be solved by removing `mut` strategically, programming would be much nicer.

PR #134, opened 2022-10-18
Rename step types to simplify some more

martinthomson:
This touches a bunch of code, but it is an automated rename so you can repeat it on your branch to rebase (then use `git merge -s ours rename-step` to create a commit that integrates this change).

### DIFF
--- a/src/helpers/error.rs
+++ b/src/helpers/error.rs
@@ -1,6 +1,6 @@
 use crate::error::BoxError;
 use crate::helpers::Identity;
-use crate::protocol::{RecordId, UniqueStepId};
+use crate::protocol::{RecordId, Step};
 use thiserror::Error;
 use tokio::sync::mpsc::error::SendError;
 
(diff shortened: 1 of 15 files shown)


martinthomson:
Note that this requires rebase, but it is easier to just redo this.  Renames are mechanical.  Review with that in mind.

Change is
* UniqueStepId -> Step (this is what most people will be dealing with
* Step -> SubStep (you might be minting these, but they won't appear in the algorithms anywhere)

benjaminsavage (approved):
Sure, why not =)

akoshelev (approved):

martinthomson:
Clippy doesn't like this.  I'll need to take another look later.

benjaminsavage:
Do we still want to do this? Or should we just abandon this diff?

martinthomson:
I abandoned it, but made the change manually instead.  These rename changes are a nuisance.

PR #136, opened 2022-10-18
Secure multiplication of vectors

taikiy:
Many protocols require a large number of iterations that executes multiplications on two shares.

Instead of having to do `for i in len() { ctx.multiply.execute(a, b) }`, it would be useful to be able to pass two vectors and execute multiplications row by row.

Unlike `multiply()`, `multiply_all()` method does not take record_id as a parameter. Instead, consecutive record_id's from the vector index will be generated.

### DIFF
--- a/src/protocol/securemul.rs
+++ b/src/protocol/securemul.rs
@@ -1,9 +1,11 @@
+use super::batch::{Batch, RecordIndex};
 use super::UniqueStepId;
 use crate::error::BoxError;
 use crate::field::Field;
 use crate::helpers::{fabric::Network, messaging::Gateway, Direction};
 use crate::protocol::{prss::PrssSpace, RecordId};
 use crate::secret_sharing::Replicated;
+use futures::future::try_join_all;
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
 use thiserror::Error;
@@ -12,6 +14,53 @@ use thiserror::Error;
 #[derive(Debug, Serialize, Deserialize, Default)]
 pub struct DValue<F> {
     d: F,
+}
+
+#[derive(Debug)]
+pub struct SecureMulAll<'a, N> {
+    prss: &'a PrssSpace,
+    gateway: &'a Gateway<N>,
+    step: &'a UniqueStepId,
+}
+
+impl<'a, N: Network> SecureMulAll<'a, N> {
+    pub fn new(prss: &'a PrssSpace, gateway: &'a Gateway<N>, step: &'a UniqueStepId) -> Self {
+        Self {
+            prss,
+            gateway,
+            step,
+        }
+    }
+
+    pub async fn execute<F: Field>(
+        self,
+        a: Batch<Replicated<F>>,
+        b: Batch<Replicated<F>>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        debug_assert_eq!(a.len(), b.len(), "lengths of the vectors must be the same");
+
+        let a: Vec<_> = a.into();
+        let b: Vec<_> = b.into();
+
+        try_join_all(
+            a.into_iter()
+                .zip(b)
+                .enumerate()
+                .map(|(i, (x, y))| async move {
+                    // Batch<> ensures that `i` is safe to cast
+                    #[allow(clippy::cast_possible_truncation)]
+                    SecureMul::new(
+                        self.prss,
+                        self.gateway,
+                        self.step,
+                        RecordId::from(i as RecordIndex),
+                    )
+                    .execute(x, y)
+                    .await
+                }),
+        )
+        .await
+    }
 }
 
 /// IKHC multiplication protocol
@@ -220,6 +269,7 @@ pub mod tests {
     use crate::error::BoxError;
     use crate::field::{Field, Fp31};
     use crate::helpers::fabric::Network;
+    use crate::protocol::batch::Batch;
     use crate::protocol::{context::ProtocolContext, QueryId, RecordId};
     use crate::secret_sharing::Replicated;
     use crate::test_fixture::{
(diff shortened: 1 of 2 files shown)


src/protocol/securemul.rs
```diff
@@ -14,6 +16,53 @@ pub struct DValue<F> {
...
+            prss,
+            gateway,
+            step,
+        }
+    }
+
+    pub async fn execute<F: Field>(
+        self,
+        a: Batch<Replicated<F>>,
+        b: Batch<Replicated<F>>,
```
taikiy, line 38: Should I take slices instead of clones to avoid extra memory allocations?

src/protocol/securemul.rs
```diff
@@ -14,6 +16,53 @@ pub struct DValue<F> {
...
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        debug_assert_eq!(a.len(), b.len(), "lengths of the vectors must be the same");
```
akoshelev, line 40: this is probably cheap enough to validate in release too

src/protocol/securemul.rs
akoshelev: this code moves `a` and `b`, so there is no extra allocations here. On the caller's side there might be a need to clone `a` and `b` though. I like your suggestion of taking a reference here because it is more appropriate here. 

There might be an implementation that takes `a` by value (or exclusive reference) and borrows `b` to then re-use `a` to store the result of `a` * `b`. This makes the whole operation  $2 \times N$ in space compared to $3 \times N$

src/protocol/securemul.rs
```diff
@@ -14,6 +16,53 @@ pub struct DValue<F> {
...
+
+        let a: Vec<_> = a.into();
```
akoshelev, line 42: You could also implement `IntoIterator` for `Batch<T>` to avoid this cast

src/protocol/securemul.rs
```diff
@@ -14,6 +16,53 @@ pub struct DValue<F> {
...
+            gateway,
+            step,
+        }
+    }
+
+    pub async fn execute<F: Field>(
+        self,
+        a: Batch<Replicated<F>>,
+        b: Batch<Replicated<F>>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
```
akoshelev, line 39: would it be better to return `Batch<F>` ?

src/protocol/securemul.rs
martinthomson: Shouldn't this take `impl IntoIterator<Item = Replicated<F>>` or maybe just a slice?  For the code that you are working on this will always be a finite set of inputs, likely a slice.

src/protocol/securemul.rs
martinthomson: Or `impl Iterator<Item = Replicated<F>>`

src/protocol/securemul.rs
```diff
@@ -14,6 +16,53 @@ pub struct DValue<F> {
...
 
+#[derive(Debug)]
+pub struct SecureMulAll<'a, N> {
+    prss: &'a PrssSpace,
+    gateway: &'a Gateway<N>,
+    step: &'a UniqueStepId,
+}
+
+impl<'a, N: Network> SecureMulAll<'a, N> {
+    pub fn new(prss: &'a PrssSpace, gateway: &'a Gateway<N>, step: &'a UniqueStepId) -> Self {
```
martinthomson, line 27: Is there ever a case where you are going to execute this multiple times in parts?  I want to ensure that starting from 0 each time is going to work properly.

If it doesn't, then this could take a starting index.

martinthomson (commented):
I question whether using `Batch` is appropriate at this stage.

That structure exists to support those cases where we might need to exchange indices, but once we are past that need, then it might be easier just to work on regular Rust data structures.  This example is very much one that would benefit from the use of `[Into]Iterator` rather than taking references (or moves) of data structures.

src/protocol/securemul.rs
taikiy: For now, I think we are okay with starting from 0 each time.

Just out of curiosity, if we were to fix this, instead of taking a starting index, could we `narrow()` the context? Running this under a sub-context is unintended for the caller, but does it result in a bad state?

src/protocol/securemul.rs
taikiy: I changed the parameter to take a slice, to allow the call site to continue using the referenced vector.

akoshelev:
> I am not convinced we need this. Happy to chat about it - and I'm open to being convinced.
> 
> I wonder if we should instead wait a while, to see how the whole batching / streaming thing plays out before we start trying to build abstractions like this. On the other hand, once we DO figure that part out, this might might provide a single, centralised place to go and update. I'm not sure. @akoshelev - what do you think?

I thought that we need this in more than one place, so having a dedicated fn/struct for it makes sense to me.

taikiy:
In the capping protocol, there are several places where we do multiplications on two vectors which are likely to be created as single vectors (instead of being the input struct) within the protocol. e.g.

`d = cap - next.helper_bit * (cap + next.compare_bit * (cap - next.current_contribution))`

Here, `compare_bit` and `current_contribution` are vectors that are unique to this protocol, and I could use `multiply_all`. I thought Bit Permutation also have a similar situation?

This is indeed not a "protocol" but just wrapping iterative multiplications in a function. I'd love to improve on this with the streaming multiplication, that would make this function more useful.

src/protocol/securemul.rs
```diff
@@ -14,6 +15,48 @@ pub struct DValue<F> {
...
+    /// Executes the secure multiplications on the given slices. This is a wrapper
+    /// on `SecureMul` that iteratively performs multiplications on elements in the
+    /// given lists positioned at the same index.
+    ///
+    /// Each multiplication is assigned a `RecordId` starting from 0. Calling this
+    /// method multiple times in parts could result in a bad state.
+    pub async fn execute<F: Field>(
+        self,
+        a: &[Replicated<F>],
+        b: &[Replicated<F>],
```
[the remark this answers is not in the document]
taikiy: I chose slices because the input vectors lengths are finite, and don't need to take the ownership of them. I could do `Iterator<Item = &Replicated<F>>` (note the &) then the caller can call this with `iter()` instead of `into_iter()`.

If the input vectors are generated with the streaming multiplication, then iterators would be more appropriate I think.

src/protocol/securemul.rs
taikiy: I think using the index part is okay, because the caller also uses the index from `for i in ...` alike to generate record IDs. 

The part I'm not sure of is whether there is a case where we execute this multiple times in parts, as you and Martin said. Would there be a case where we do something like this?

```
let v1, v2: Vec<Replicated<F>> = ...
multiply_all(v1[..x], v2[..x]);
...
multiply_all(v1[x+1..], v2[x+1..]);
```

I think narrow() would cover this case nicely. That doesn't force the caller to keep track and pass a start index each time.

src/protocol/securemul.rs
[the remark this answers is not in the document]
benjaminsavage: Unfortunately, I'm too much of Rust n00b to know what we should do here. This seems like a find place to start and we can always modify it later if we need / want to.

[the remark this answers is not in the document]
martinthomson: So when you say finite, do you mean finite in the sense that it is always a small number that is independent of the number of records?  In that case, provided that the number of records that the inputs span (I know that you are processing in increasingly large jumps) is sufficiently small (number: TBD) then a slice is fine.  On the other hand, if it could be a very large span of the total set of records or dependent on the number of records, an iterator is better.

A stream would be even better, I suspect, but then you would also be producing a stream.

taikiy:
Closing this PR for now. We can bring this back if any protocols have strong need for it.

PR #140, opened 2022-10-19
Optimizing the multiplication protocol used in share conversion

benjaminsavage:
We can use a specialized multiplication protocol to save half the communication bandwidth, because we know that a large number of shares are zero.

### DIFF
--- a/src/protocol/modulus_conversion/specialized_mul.rs
+++ b/src/protocol/modulus_conversion/specialized_mul.rs
@@ -1,0 +1,424 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::{fabric::Network, Direction, Identity},
+    protocol::{context::ProtocolContext, RecordId},
+    secret_sharing::Replicated,
+};
+
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've multiplied their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct DValue<F> {
+    d: F,
+}
+
+/// A variant of the IKHC multiplication protocol
+/// for use in special cases where many of the shares are known to be zero.
+/// Original IKHC multiplication protocol from:
+/// K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, and B. Pinkas. High-throughput secure AES computation. In WAHC@CCS 2018, pp. 13–24, 2018
+/// Optimizations from: "Adam in Private: Secure and Fast Training of Deep Neural Networks with Adaptive Moment Estimation"
+/// by Nuttapong Attrapadung, Koki Hamada, Dai Ikarashi, Ryo Kikuchi*, Takahiro Matsuda, Ibuki Mishina, Hiraku Morita, and Jacob C. N. Schuldt
+#[derive(Debug)]
+pub struct SpecializedMul {}
+
+impl SpecializedMul {
+    /// A highly specialized variant of the secure multiplication protocol which is only valid
+    /// in the case where 4 of the 6 shares are zero.
+    ///
+    /// This is taken from Appendix F: "Conversion Protocols" from the paper:
+    /// "Adam in Private: Secure and Fast Training of Deep Neural Networks with Adaptive Moment Estimation"
+    /// by Nuttapong Attrapadung, Koki Hamada, Dai Ikarashi, Ryo Kikuchi*, Takahiro Matsuda,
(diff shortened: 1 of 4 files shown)


src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// need to be sent.
+    ///
+    /// ## Errors
+    /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+    /// back via the error response
+    #[allow(dead_code)]
+    pub async fn execute_specialized_1<F: Field, N: Network>(
+        ctx: ProtocolContext<'_, N>,
+        record_id: RecordId,
+        a: Replicated<F>,
```
akoshelev, line 52: we could also keep `a` and `b`  inside the struct and make this method consume self. alternatively this can just be an `async fn` because there is no state

src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -0,0 +1,424 @@
...
+/// for use in special cases where many of the shares are known to be zero.
+/// Original IKHC multiplication protocol from:
+/// K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, and B. Pinkas. High-throughput secure AES computation. In WAHC@CCS 2018, pp. 13–24, 2018
+/// Optimizations from: "Adam in Private: Secure and Fast Training of Deep Neural Networks with Adaptive Moment Estimation"
+/// by Nuttapong Attrapadung, Koki Hamada, Dai Ikarashi, Ryo Kikuchi*, Takahiro Matsuda, Ibuki Mishina, Hiraku Morita, and Jacob C. N. Schuldt
+#[derive(Debug)]
+pub struct SpecializedMul {}
+
+impl SpecializedMul {
+    /// A highly specialized variant of the secure multiplication protocol which is only valid
```
akoshelev, line 27: I love the explanation here

src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// In the IKHC multiplication protocol, each helper computes `d_i` and
+    /// sends it to the next helper. But in this case, `d_1` is publicly known to all
+    /// the helper parties and can be replaced with a constant, e.g. 0. Therefore, it does not
+    /// need to be sent.
+    ///
+    /// ## Errors
+    /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+    /// back via the error response
+    #[allow(dead_code)]
+    pub async fn execute_specialized_2<F: Field, N: Network>(
```
akoshelev, line 130: naming is hard...

akoshelev (approved):

src/protocol/modulus_conversion/specialized_mul.rs
benjaminsavage: I prefer pure functions and no state, so I'm going to try to keep all my protocols just accepting all inputs.

benjaminsavage: Are you suggesting I should just eliminate the struct entirely and simply export a function? That makes sense! I'll do that.

src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -0,0 +1,424 @@
...
+        b: Replicated<F>,
+    ) -> Result<Replicated<F>, BoxError> {
+        match ctx.role() {
+            Identity::H1 => {
+                let prss = &ctx.prss();
+                let (s_3_1, _) = prss.generate_fields(record_id.into());
+
+                // d_1 = a_1 * b_2 + a_2 * b_1 - s_3,1
+                // d_1 = a_1 * b_2 + 0 * 0 - s_3,1
+                let (a_1, _) = a.as_tuple();
```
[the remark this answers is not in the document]
benjaminsavage: That's not a bad idea, especially for a debug build. Maybe in a release build you could skip that, working under the assumption that such bugs would only come from incorrectly written protocols, and that those should be discovered when running a debug build in testing.

src/protocol/modulus_conversion/specialized_mul.rs
benjaminsavage: indeed =(

src/protocol/modulus_conversion/specialized_mul.rs
benjaminsavage: Thanks!

PR #142, opened 2022-10-20
Update the README and add CONTRIBUTING

martinthomson:
This is all very drafty and I'm definitely not a lawyer, but hopefully this is enough to start with.


benjaminsavage (approved):
Thanks @martinthomson - this is a great start!

PR #144, opened 2022-10-20
Implementing the Check Zero protocol

benjaminsavage:
This is needed later on to upgrade to malicious security

### DIFF
--- a/src/protocol/check_zero.rs
+++ b/src/protocol/check_zero.rs
@@ -1,0 +1,149 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::fabric::Network,
+    protocol::{context::ProtocolContext, reveal::reveal, RecordId},
+    secret_sharing::Replicated,
+};
+
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've multiplied their own shares
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct DValue<F> {
+    d: F,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum Step {
+    MultiplyWithR,
+    RevealR,
+}
+
+impl crate::protocol::Step for Step {}
+
+impl AsRef<str> for Step {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::MultiplyWithR => "multiply_with_r",
+            Self::RevealR => "reveal_r",
+        }
+    }
+}
+
+/// A very simple protocol to check if a replicated secret sharing is in fact a sharing of zero
+///
+/// NOTE: this protocol leaks information about `v` the helpers. Please only use this in cases where
+/// this type of information leakage is acceptable, such as where `v` is the product of a secret value
+/// and a random, unknown value.
+///
+/// This is an implementation of PROTOCOL 3.7 from the paper:
+/// Fast Large-Scale Honest-Majority MPC for Malicious Adversaries
+/// <https://link.springer.com/content/pdf/10.1007/978-3-319-96878-0_2.pdf>
+///
+/// The parties start out holding a replicated secret sharing of a value `v`, which they would like to check for equality to zero.
+/// First, the parties generate a secret sharing of a random value `r`, which is not known to any of them.
+/// Next, the parties compute a secret sharing of `r * v` using a multiplication protocol that is secure up to an additive attack.
+/// Then, the parties reveal (sometimes called "open") the secret sharing of `r * v`
+/// If `v` was a secret sharing of zero, this revealed value will also be zero.
+/// On the other hand, if `v` was NOT a secret sharing of zero, then there are two possibilities:
+/// 1.) If the randomly chosen value `r` just so happenned to be zero, then the revealed value will be zero.
+/// This will happen with probability `1/|F|` (where `|F|` denotes the cardinality of the field)
+/// 2.) If the randomly chosen value `r` is any other value in the field aside from zero, then the revealed value will NOT be zero.
+///
+/// Clearly, the accuracy of this protocol is highly dependent on the field that is used.
+/// In a large field, like the integers modulo 2^31 - 1, the odds of `check_zero` returning "true"
+/// when `v` is NOT actually a sharing of zero is extrmely small; it is less than one two billion odds.
+/// In a silly field, like our test field of the integers modulo 31, the odds are very good. It'll incorrectly return "true"
+/// about 3.2% of the time.
+///
+/// ## Errors
+/// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+/// back via the error response
+#[allow(dead_code)]
+pub async fn check_zero<F: Field, N: Network>(
+    ctx: ProtocolContext<'_, N>,
+    record_id: RecordId,
+    v: Replicated<F>,
+) -> Result<bool, BoxError> {
+    let prss = &ctx.prss();
+    let (left, right) = prss.generate_fields(record_id.into());
+    let r_sharing = Replicated::new(left, right);
+
+    let r_times_v = ctx
+        .narrow(&Step::MultiplyWithR)
+        .multiply(record_id)
+        .await
+        .execute(r_sharing, v)
+        .await?;
+    let r_revealed = reveal(ctx.narrow(&Step::RevealR), record_id, r_times_v).await?;
+
+    Ok(r_revealed == F::ZERO)
+}
+
+#[cfg(test)]
+pub mod tests {
+    use crate::error::BoxError;
+    use crate::field::{Field, Fp31};
+    use crate::protocol::{check_zero::check_zero, QueryId, RecordId};
+    use crate::test_fixture::{logging, make_contexts, make_world, share, TestWorld};
+
+    #[tokio::test]
+    async fn basic() -> Result<(), BoxError> {
+        logging::setup();
+
+        let world: TestWorld = make_world(QueryId);
+        let context = make_contexts(&world);
+        let mut rng = rand::thread_rng();
+
+        for i in 0..u32::from(Fp31::PRIME) {
+            let a = Fp31::from(i);
+            let mut num_false_positives = 0;
+            for j in 0..10 {
+                let a_shares = share(a, &mut rng);
+                let row = i * u32::from(Fp31::PRIME) + j;
+                let record_id = RecordId::from(row);
+
+                let iteration = format!("{}", row);
+
+                let protocol_output = tokio::try_join!(
+                    check_zero(context[0].narrow(&iteration), record_id, a_shares[0],),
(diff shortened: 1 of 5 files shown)


src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Then, the parties reveal (sometimes called "open") the secret sharing of `r * v`
+/// If `v` was a secret sharing of zero, this revealed value will also be zero.
+/// On the other hand, if `v` was NOT a secret sharing of zero, then there are two possibilities:
+/// 1.) If the randomly chosen value `r` just so happenned to be zero, then the revealed value will be zero.
+/// This will happen with probability `1/|F|` (where `|F|` denotes the cardinality of the field)
+/// 2.) If the randomly chosen value `r` is any other value in the field aside from zero, then the revealed value will NOT be zero.
+///
+/// Clearly, the accuracy of this protocol is highly dependent on the field that is used.
+/// In a large field, like the integers modulo 2^31 - 1, the odds of `check_zero` returning "true"
+/// when `v` is NOT actually a sharing of zero is extrmely small; it is less than one two billion odds.
```
martinthomson, line 56:
```suggestion
/// when `v` is NOT actually a sharing of zero is extremely small; it is less than one two billion odds.
```

This is not quite right.  The probability that this returns true, based on the above given $1-(1-\frac{1}{|F|})^2$.  For $|F|=2^{31}-1$ I think that is one in a billion, which is basically $1:2^{30}$.  That's still pretty good, to be clear, but you are off by a factor of two.

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+/// In a silly field, like our test field of the integers modulo 31, the odds are very good. It'll incorrectly return "true"
+/// about 3.2% of the time.
```
martinthomson, line 58: By the above this is $6.\dot{5}$%

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+/// back via the error response
+#[allow(dead_code)]
+pub async fn check_zero<F: Field, N: Network>(
+    ctx: ProtocolContext<'_, N>,
+    record_id: RecordId,
+    v: Replicated<F>,
+) -> Result<bool, BoxError> {
+    let prss = &ctx.prss();
+    let (left, right) = prss.generate_fields(record_id.into());
```
martinthomson, line 70: I see this a lot.  Should we make the functions on the PRSS take `impl Into<u128>` or `RecordId`?

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    let r_times_v = ctx
+        .narrow(&Step::MultiplyWithR)
+        .multiply(record_id)
+        .await
+        .execute(r_sharing, v)
+        .await?;
+    let r_revealed = reveal(ctx.narrow(&Step::RevealR), record_id, r_times_v).await?;
+
+    Ok(r_revealed == F::ZERO)
```
martinthomson, line 81: I think you want this:

```suggestion
    let rv = reveal(ctx.narrow(&Step::RevealR), record_id, r_times_v).await?;

    Ok(rv == F::ZERO)
```

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+pub async fn check_zero<F: Field, N: Network>(
+    ctx: ProtocolContext<'_, N>,
+    record_id: RecordId,
+    v: Replicated<F>,
+) -> Result<bool, BoxError> {
+    let prss = &ctx.prss();
+    let (left, right) = prss.generate_fields(record_id.into());
+    let r_sharing = Replicated::new(left, right);
+
+    let r_times_v = ctx
```
martinthomson, line 73:
```suggestion
    let rv_share = ctx
```

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+/// back via the error response
+#[allow(dead_code)]
+pub async fn check_zero<F: Field, N: Network>(
+    ctx: ProtocolContext<'_, N>,
+    record_id: RecordId,
+    v: Replicated<F>,
+) -> Result<bool, BoxError> {
+    let prss = &ctx.prss();
+    let (left, right) = prss.generate_fields(record_id.into());
+    let r_sharing = Replicated::new(left, right);
```
martinthomson, line 71: Do we need PRSS to generate `Replicated`?

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+    #[tokio::test]
+    async fn basic() -> Result<(), BoxError> {
+        logging::setup();
+
+        let world: TestWorld = make_world(QueryId);
+        let context = make_contexts(&world);
+        let mut rng = rand::thread_rng();
+
+        for i in 0..u32::from(Fp31::PRIME) {
+            let a = Fp31::from(i);
```
martinthomson, line 100: Check variable names here.  I think that you can shadow the variable on the loop and do:

```rust
for v in 0..u32::from(Fp31::PRIME) {
    let v = Fp31::from(v);
```

I wonder what we need to implement so that we can just do:

```rust
for v in Fp31::ZERO..Fp31::PRIME {
}
```

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+                            num_false_positives += 1;
+                        }
+                    }
+                }
+            }
+
+            // Fp31 is just too small of a field.
+            // Through random chance, it'll incorrectly return "true"
+            // one time in 31. The odds of incorrectly returning "true"
+            // 5 times or more is... small...
```
martinthomson, line 143: This one is correct, because you aren't counting when v==1 as a false positive :)

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+impl AsRef<str> for Step {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::MultiplyWithR => "multiply_with_r",
+            Self::RevealR => "reveal_r",
+        }
+    }
+}
+
+/// A very simple protocol to check if a replicated secret sharing is in fact a sharing of zero
```
martinthomson, line 34:
```suggestion
/// A very simple protocol to check if a replicated secret sharing is a sharing of zero.
```

martinthomson (approved):
Pretty straightforward.

src/protocol/check_zero.rs
benjaminsavage: I don't think so. Why are you squaring it?

The only way it goes wrong is if r == 0. The odds of choosing that is just one out of the size of the field.

Here's the relevant snippet from the paper:

"`F_checkZero` receives the honest parties shares, and use them to reconstruct the
shared value. Then, if it is 0, then the simulator decides whether to send accept
to the honest parties or reject. Otherwise, `F_checkZero` tosses a coin to decide what
to send to the parties (i.e., in this case, the simulator is not given the opportunity
to modify the output). In particular, when the checked value does not equal to 0,
the output will still be accept with probability 1 / |F| . This captures the event in
Protocol 3.7 where v != 0 but T = 0, which happens also with probability 1/F 
since r is uniformly distributed over F."

src/protocol/check_zero.rs
benjaminsavage: That's a good idea. I think we should make them also accept RecordId directly. I'll put that in a follow-up diff to keep things logically separated.

src/protocol/check_zero.rs
benjaminsavage: We could do that, and it would essentially be the "functionality" named "`F_rand` in the papers, so maybe we could give it a name like that.

src/protocol/check_zero.rs
martinthomson: I'm thinking about both v and r being random, which leads to the squaring.  But I guess that your point would be that - in this specific application of `F_checkZero` - the value of `v` cannot be zero.  I didn't see how that was possible in the case where you are doing a dot product.  Though each individual value is non-zero, the sum is not guaranteed to be.

src/protocol/check_zero.rs
martinthomson: Note that taking RecordId means that you can't feed it a bare number, which could be annoying for testing purposes.  Making the function generic avoids that issue, but also ensures that when you compile the real system you only pay for what you need.

src/protocol/check_zero.rs
benjaminsavage: I can do this, but I was using the loop variable "i" below in my funky calculation of record_id.

src/protocol/check_zero.rs
```diff
@@ -0,0 +1,149 @@
...
+        let context = make_contexts(&world);
+        let mut rng = rand::thread_rng();
+
+        for i in 0..u32::from(Fp31::PRIME) {
+            let a = Fp31::from(i);
+            let mut num_false_positives = 0;
+            for j in 0..10 {
+                let a_shares = share(a, &mut rng);
+                let row = i * u32::from(Fp31::PRIME) + j;
+                let record_id = RecordId::from(row);
```
martinthomson, line 105: Can you instead just maintain a counter for this? i.e.,

```rust
let mut record_id = 0;
for i in ... {
    for j in ... {
        record_id += 1;
        let record_id = RecordId::from(record_id); // or name it something else because this shadowing is super sketchy...
    }
}
```

src/protocol/check_zero.rs
benjaminsavage: Ah, I think you're conflating two separate things:

This protocol is just "check zero". For this particular functionality, if "v" is indeed zero, then this protocol will always return the correct answer. Put another way, there are no "false negatives".
If "v" is NOT zero, this protocol MIGHT return an incorrect answer. The odds of that are just the odds that r == 0.

When we later on get to the application of this protocol to upgrading a circuit to malicious security, THEN, indeed, the odds are different, because we are talking about something else. Then, the odds that the protocol says "Everything is all good!" when everything is NOT all good are indeed 2/ |F|.

Quoting from the paper: 
"Overall, we therefore have that the adversary can cheat with probability at most 2/|F|. We prove this formally in Lemma 4.2. A full specification appears in Protocol 4.1"

...but this will be in a different comment in a different file. If you're just reading the "check_zero.rs" file, the current comment is accurate. I will definitely have another comment, in whatever that other file is eventually called, explaining the odds of successfully cheating.

src/protocol/check_zero.rs
benjaminsavage: Can we just implement two versions of this function which accept different arguments?

martinthomson: Rust doesn't do overloading like C++, so you would need two names.

src/protocol/check_zero.rs
benjaminsavage: yup, I'll just use a counter of some sort.

PR #145, opened 2022-10-20
Adding a new dot-product protocol

benjaminsavage:
If you have two vectors of secret shared numbers, and you'd like to compute the dot-product of these two vectors (that is, the sum of the pairwise products), this can be accomplished with a communication cost that is equivalent to just one multiplication. That's right, it's irrelevant how long the vectors are. Pretty cool huh?

### DIFF
--- a/src/protocol/dot_product.rs
+++ b/src/protocol/dot_product.rs
@@ -1,0 +1,161 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::{fabric::Network, Direction},
+    protocol::{context::ProtocolContext, RecordId},
+    secret_sharing::Replicated,
+};
+
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've computed one share of the result
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct UValue<F> {
+    u: F,
+}
+
+///
+/// This protocol is an extension of the CHIKP multiplication protocol described in the paper:
+/// "High-Throughput Secure AES Computation" (High-throughput secure AES computation. In WAHC@CCS 2018, pp. 13–24, 2018)
+/// by Koji Chida, Koki Hamada, Dai Ikarashi, Ryo Kikuchi and Benny Pinkas
+/// <https://dl.acm.org/doi/10.1145/3267973.3267977>
+///
+/// The extension allows for the computation of the sum of multiple multiplications (in other words, a dot-product operation)
+/// at the communication cost of a single multiplication.
+///
+/// This extension is summarized in the paper:
+/// "Fast Large-Scale Honest-Majority MPC for Malicious Adversaries"
+/// by Koji Chida, Daniel Genkin, Koki Hamada, Dai Ikarashi, Ryo Kikuchi, Yehuda Lindell, and Ariel Nof
+/// <https://link.springer.com/content/pdf/10.1007/978-3-319-96878-0_2.pdf>
+///
+/// In that paper, this functionality is called `F_product` and
+/// an instantiation for Replciated Secret Sharing is provided in section 6.1 on page 25
+///
+/// To summarize the CHIKP protocol:
+/// each helper computes `u_i = (s_i + s_i+1) * (t_i + t_i+1) - s_i+1 * t_i+1`
+/// It then randomizes this share with correlated randomness, and sends it to the helper to its right
+/// Each helper then defines the pair `(u_i, u_i+1)` as its shares of the output.
+///
+/// (Note it is equivalent to compute `u_i = s_i * t_i + s_i+1 * t_i + s_i * t_i+1`,
+/// but this requires three multiplications instead of two. Assuming multiplying field elements is
+/// significantly more costly (in CPU terms) that addition, this is less efficient to compute)
+///
+/// The key observation, that makes this protocol work, is that each helper can compute many `u_i` values
+/// from many multiplications, sum them all together, randomize them with correlated randomness, and can
+/// send this single field value to the next helper. The result will be a replicated secret sharing of the
+/// dot-product of the two vectors.
+///
+/// ## Errors
+/// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+/// back via the error response
+#[allow(dead_code)]
+pub async fn dot_product<F: Field, N: Network>(
+    ctx: ProtocolContext<'_, N>,
+    record_id: RecordId,
+    a: &[Replicated<F>],
+    b: &[Replicated<F>],
+) -> Result<Replicated<F>, BoxError> {
+    assert_eq!(a.len(), b.len(), "lengths of the lists must be the same");
+
+    let prss = &ctx.prss();
+    let (s_left, s_right): (F, F) = prss.generate_fields(record_id.into());
+
+    let mut u_right = s_right - s_left;
+    for i in 0..a.len() {
+        let (a_left, a_right) = a[i].as_tuple();
+        let (b_left, b_right) = b[i].as_tuple();
+        u_right += (a_left + a_right) * (b_left + b_right) - a_right * b_right;
+    }
+
+    // send our `u_i+1` value to the helper on the right
+    let channel = ctx.mesh();
+    channel
+        .send(
+            channel.identity().peer(Direction::Right),
+            record_id,
+            UValue { u: u_right },
+        )
+        .await?;
+
+    // receive `u_i` value from helper to the left
+    let UValue { u: u_left } = channel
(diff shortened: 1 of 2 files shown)


src/protocol/dot_product.rs
```diff
@@ -0,0 +1,162 @@
...
+        u_right += (a_left + a_right) * (b_left + b_right) - a_right * b_right;
+    }
+
+    // send our `u_i+1` value to the helper on the right
+    let channel = ctx.mesh();
+    channel
+        .send(
+            channel.identity().peer(Direction::Right),
+            record_id,
+            UValue { u: u_right },
```
martinthomson, line 77: This bit where we wrap values in a new type is not ergonomic.  We should just be able to send the field value.

martinthomson (approved):

src/protocol/dot_product.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Each helper then defines the pair `(u_i, u_i+1)` as its shares of the output.
+///
+/// (Note it is equivalent to compute `u_i+1 = s_i * t_i + s_i+1 * t_i + s_i * t_i+1`,
+/// but this requires three multiplications instead of two. Assuming multiplying field elements is
+/// significantly more costly (in CPU terms) than addition, this is less efficient to compute)
+///
+/// The key observation, that makes this protocol work, is that each helper can compute many `u_i+1` values
+/// from many multiplications, sum them all together, randomize them with correlated randomness, and can
+/// send this single field value to the next helper. The result will be a replicated secret sharing of the
+/// dot-product of the two vectors.
```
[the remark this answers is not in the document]
benjaminsavage: This is just the dot-product protocol. But if you're talking about how it'll eventually get used in the malicious security checker, it will compare:
$[u]$ and $r[w]$
Where:
$[u] = \sum [α_k] * [r*z_k]$
and: 
$r[w] = r\sum [α_k] * [z_k]$?

Where all of the `[α_k]` are just random, secret-shared values we get from prss.generate_replicated() and all of the `[z_k]` are the secret-shared outputs of all the multiplications / re-shares / inputs.

I don't *think* this is the same as double-entry book-keeping. It's more of an information theoretic MAC. Here's how Chida et. all describe it:

"Our protocol works by running multiple circuit computations in parallel: one
that computes the circuit over the real inputs and others which compute the
circuit over randomized inputs. The outputs of the randomized circuits are then
used to verify the correctness of the “original” circuit computation, thereby
constituting a SPDZ-like MAC"

src/protocol/dot_product.rs
benjaminsavage: I agree. We should make that change. @akoshelev - is there a specific reason why we are doing this? Can we just stop using these structs?

src/protocol/dot_product.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// randomness it shares with the helper to the left, and sends to the right.
+    /// It receives a randomized sum from the helper to its left, and is now in
+    /// possession of a replicated secret sharing of the dot-product of the two
+    /// streaming vectors of replicated secret shares.
+    ///
+    /// ## Errors
+    /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+    /// back via the error response
+    #[allow(dead_code)]
+    pub async fn accumulate_dot_product<F: Field, N: Network>(
```
thurstonsand, line 69: i think there is a way to use stream without `Pin<Box<dyn Stream>>`. roughly speaking, it would look something like:

```rust
pub async fn accumulate_dot_product<F: Field, N: Network, S: Stream<Item = PairOfShares<F>>(
    ctx: ProtocolContext<'_, N>,
    record_id: RecordId,
    input_stream: S,
) -> Result<Replicated<F>, BoxError> {
    ...

    // instead of `while let Some(item) = input_stream.next()`
    // use `fold`: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.fold
    let u_right = input_stream.fold(s_right - s_left, |acc, (a_share, b_share)| async move { 
        let (a_left, a_right) = a_share.as_tuple();
        let (b_left, b_right) = b_share.as_tuple();
        acc + (a_left + a_right) * (b_left + b_right) - a_right * b_right
    }).await
    ...
```
`Pin` is a pain to work with directly, and `.fold` will hide that complexity away.

src/protocol/dot_product.rs
```diff
@@ -0,0 +1,173 @@
...
+            let world = make_world(QueryId);
+            let context = make_contexts(&world);
+            let record_id = RecordId::from(0);
+            let step = "some_step".to_string();
+
+            let result_shares = tokio::try_join!(
+                accumulate_dot_product(
+                    context[0].narrow(&step),
+                    record_id,
+                    Box::pin(stream::iter(vec1)),
```
thurstonsand, line 151: if you move away from `Pin`, you can drop the `Box::pin` here and pass the `Stream` directly

thurstonsand (changes requested):

src/protocol/dot_product.rs
```diff
@@ -0,0 +1,173 @@
...
+        ctx: ProtocolContext<'_, N>,
+        record_id: RecordId,
+        mut input_stream: Pin<Box<dyn Stream<Item = PairOfShares<F>>>>,
+    ) -> Result<Replicated<F>, BoxError> {
+        let prss = &ctx.prss();
+        let (s_left, s_right): (F, F) = prss.generate_fields(record_id);
+
+        let mut u_right = s_right - s_left;
+
+        while let Some(item) = input_stream.next().await {
```
akoshelev, line 79: how do you plan to implement check-pointing? I would imagine with this model you need to open a new stream inside every maliciously secure protocol and send everything to that stream. There is only one place where you can do the verification which is after this function returns. I wonder if you would want intermediate checkpoints as well - that may require incorporating verification and dot product together

src/protocol/dot_product.rs
akoshelev: the original idea was to use the type information to resolve the ambiguity in cases where protocols may want to send two or more messages in the same step (and record). Infra theoretically can support this by resolving `record_id` to multiple messages. Having said that I am yet to see a use case for it, so I don't have a good argument against getting rid of wrappers.

src/protocol/dot_product.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    use serde::{Deserialize, Serialize};
+
+    /// A message sent by each helper when they've computed one share of the result
+    #[derive(Debug, Serialize, Deserialize, Default)]
+    pub struct UValue<F> {
+        u: F,
+    }
+
+    type PairOfShares<F> = (Replicated<F>, Replicated<F>);
```
akoshelev, line 54: it may be worth creating a wrapper here that explicitly defines which share is $z_k$ and which is $r \times z_k$. 
Looks like if I reorder these shares by sending $z_{k_1}$, $r \times z_{k_1}$ and then $r \times z_{k_2}$  $z_{k_2}$ that will lead to incorrect result so it is worth trying to prevent that

benjaminsavage: Oh, this will only be for just one. We will have two dot-product accumulator string thingies. One for the dot product:
$[u] = \sum [α_k] * [r*z_k]$
and another one for the dot product:
$r[w] = r\sum [α_k] * [z_k]$?

For the first stream, we will pass in pairs of $([α_k], [r*z_k])$.
For the second stream, we will pass in pairs of $([α_k], [z_k])$

And actually, the order will not matter at all because dot_product(a, b) == dot_product(b, a)

src/protocol/dot_product.rs
benjaminsavage: Oh my gosh, thank you so much! This is a million times better. I am really happy to not have to touch this pin / box stuff =).

src/protocol/dot_product.rs
benjaminsavage: I imagine that you'll need to say:

let v = MaliciousSecurityValidator::new();

And then you'll need to tell it about all of the inputs that go into the circuit, and every time you encounter a multiplication, or a reshare, you'll need to tell it about that. Then, at the end of some period of time (e.g. before we reveal a permutation within our sort protocol) you'll say:

v.validate();

It'll close out the two streams, which will finalise the computation of these two dot products ( $[w]$ and $[u]$), reveal $r$, and finally call `CheckZero` on $r[w] - [u]$. 

If it's zero, all good. If it's NOT zero, abort the entire protocol and file an incident report.

src/protocol/dot_product.rs
benjaminsavage: That's awesome. Thank you so much @thurstonsand !

PR #146, opened 2022-10-20
HTTP2/HTTP 1.1 metrics for IPA server

akoshelev:
This change makes IPA web server to emit a metric that shows which HTTP protocol was used for every request made to the helper. The goal is to be aware if helpers start using HTTP 1.1. Restricting server code to accept HTTP 2.0 only is not great because HTTP 1.1 is way easier to debug and therefore valuable to support.

Alternatively we could just use a config that instructs the server which protocol it should exclusively support. Right now it feels that having this visibility is enough, but we may reconsider this in the future.

Note that it may be a bit misleading if connection starts as HTTP/1.1 and then gets upgraded to HTTP/2.0 but in our setting helper clients should probably initiate HTTP/2 from the beginning.

### DIFF
--- a/src/telemetry/mod.rs
+++ b/src/telemetry/mod.rs
@@ -24,14 +24,11 @@ pub mod metrics {
             const HTTP3: &str = "request.protocol.HTTP/3";
             const UNKNOWN: &str = "request.protocol.HTTP/UNKNOWN";
 
-            KeyName::from_const_str(if v.0 == Version::HTTP_11 {
-                HTTP11
-            } else if v.0 == Version::HTTP_2 {
-                HTTP2
-            } else if v.0 == Version::HTTP_3 {
-                HTTP3
-            } else {
-                UNKNOWN
+            KeyName::from_const_str(match v.0 {
+                Version::HTTP_11 => HTTP11,
+                Version::HTTP_2 => HTTP2,
+                Version::HTTP_3 => HTTP3,
+                _ => UNKNOWN,
             })
         }
     }
--- a/src/telemetry/mod.rs
+++ b/src/telemetry/mod.rs
@@ -20,8 +20,8 @@ pub mod metrics {
     impl From<RequestProtocolVersion> for KeyName {
         fn from(v: RequestProtocolVersion) -> Self {
             const HTTP11: &str = "request.protocol.HTTP/1.1";
-            const HTTP2: &str = "request.protocol.HTTP/2.0";
-            const HTTP3: &str = "request.protocol.HTTP/3.0";
+            const HTTP2: &str = "request.protocol.HTTP/2";
+            const HTTP3: &str = "request.protocol.HTTP/3";
             const UNKNOWN: &str = "request.protocol.HTTP/UNKNOWN";
 
             KeyName::from_const_str(if v.0 == Version::HTTP_11 {
--- a/src/telemetry/mod.rs
+++ b/src/telemetry/mod.rs
@@ -1,8 +1,40 @@
 pub mod metrics {
-    use metrics::describe_counter;
+    use axum::http::Version;
     use metrics::Unit;
+    use metrics::{describe_counter, KeyName};
 
     pub const REQUESTS_RECEIVED: &str = "requests.received";
+
+    /// Metric that records the version of HTTP protocol used for a particular request.
+    #[cfg(feature = "web-app")]
+    pub struct RequestProtocolVersion(Version);
+
+    #[cfg(feature = "web-app")]
+    impl From<Version> for RequestProtocolVersion {
+        fn from(v: Version) -> Self {
+            RequestProtocolVersion(v)
+        }
+    }
+
+    #[cfg(feature = "web-app")]
+    impl From<RequestProtocolVersion> for KeyName {
+        fn from(v: RequestProtocolVersion) -> Self {
+            const HTTP11: &str = "request.protocol.HTTP/1.1";
+            const HTTP2: &str = "request.protocol.HTTP/2.0";
+            const HTTP3: &str = "request.protocol.HTTP/3.0";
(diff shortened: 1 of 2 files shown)


src/telemetry/mod.rs
```diff
@@ -1,9 +1,41 @@
...
+            RequestProtocolVersion(v)
+        }
+    }
+
+    #[cfg(feature = "web-app")]
+    impl From<RequestProtocolVersion> for KeyName {
+        fn from(v: RequestProtocolVersion) -> Self {
+            const HTTP11: &str = "request.protocol.HTTP/1.1";
+            const HTTP2: &str = "request.protocol.HTTP/2.0";
+            const HTTP3: &str = "request.protocol.HTTP/3.0";
```
martinthomson, line 24:
```suggestion
            const HTTP2: &str = "request.protocol.HTTP/2";
            const HTTP3: &str = "request.protocol.HTTP/3";
```

martinthomson (approved):

src/telemetry/mod.rs
```diff
@@ -1,9 +1,41 @@
...
+
+            KeyName::from_const_str(if v.0 == Version::HTTP_11 {
+                HTTP11
+            } else if v.0 == Version::HTTP_2 {
+                HTTP2
+            } else if v.0 == Version::HTTP_3 {
+                HTTP3
+            } else {
+                UNKNOWN
+            })
```
thurstonsand, line 35: would a match statement be an improvement here?
```suggestion
            KeyName::from_const_str(match v.0 {
                Version::HTTP_11 => HTTP11,
                Version::HTTP_2 => HTTP2,
                Version::HTTP_3 => HTTP3,
                _ => UNKNOWN,
            })
```

src/telemetry/mod.rs
```diff
@@ -19,23 +51,34 @@ pub mod metrics {
...
+    pub fn get_counter_value<K: Into<KeyName>>(
         snapshot: metrics_util::debugging::Snapshot,
-        metric_name: &'static str,
+        metric_name: K,
     ) -> Option<u64> {
         use metrics_util::debugging::DebugValue;
         use metrics_util::MetricKind;
 
         let snapshot = snapshot.into_vec();
+        let metric_name = metric_name.into();
```
thurstonsand, line 78: confusing that this works... is it looking at `as_str()` and somehow realizing that `KeyName` is the exclusive conversion that has this function? what if there were 2 possible `Into<_>` that implemented `as_str()`?

src/telemetry/mod.rs
akoshelev: I don't think you can do that because `Version` is not an enum and these values are just constants :(

src/telemetry/mod.rs
akoshelev: I think compiler uses the knowledge that `K: Into<KeyName>` to resolve the ambiguity here

akoshelev:
verified with CLI that metrics are indeed there

server:

```bash
cargo run --features self-signed-certs --bin helper -- -s http -p 3000

+---------------------------+-----------------------------------------------------+-------+------------+
| metric                    | description                                         | value | dimensions |
+======================================================================================================+
| request.protocol.HTTP/1.1 | Total number of HTTP/1.1 requests received          | 1     |            |
|---------------------------+-----------------------------------------------------+-------+------------|
| requests.received         | Total number of requests received by the web server | 2     |            |
|---------------------------+-----------------------------------------------------+-------+------------|
| request.protocol.HTTP/2.0 | Total number of HTTP/2 requests received            | 1     |            |
+---------------------------+-----------------------------------------------------+-------+------------+
```

client

```bash
curl "http://127.0.0.1:3000/echo?anotherfoo=anotherbar"                                                                                                                           infra-send-buffers 3edb69e ✗
{"query_args":{"anotherfoo":"anotherbar"},"headers":{"host":"127.0.0.1:3000","accept":"*/*","user-agent":"curl/7.79.1"}}

curl "http://127.0.0.1:3000/echo?anotherfoo=anotherbar" --http2-prior-knowledge                                                                                                   infra-send-buffers 3edb69e ✗
{"query_args":{"anotherfoo":"anotherbar"},"headers":{"user-agent":"curl/7.79.1","accept":"*/*"}}
```

src/telemetry/mod.rs
martinthomson: Try it.  Also, why is this doing a string lookup rather than taking an enum?

akoshelev: This is the definition of `Version` struct - it is not an enum and `Http` is not accessible from it

```rust
/// Represents a version of the HTTP spec.
#[derive(PartialEq, PartialOrd, Copy, Clone, Eq, Ord, Hash)]
pub struct Version(Http);

impl Version {
    /// `HTTP/0.9`
    pub const HTTP_09: Version = Version(Http::Http09);

    /// `HTTP/1.0`
    pub const HTTP_10: Version = Version(Http::Http10);

    /// `HTTP/1.1`
    pub const HTTP_11: Version = Version(Http::Http11);

    /// `HTTP/2.0`
    pub const HTTP_2: Version = Version(Http::H2);

    /// `HTTP/3.0`
    pub const HTTP_3: Version = Version(Http::H3);
}

```

martinthomson: The match should work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=503992d87137d897f95361095e3ac546

akoshelev: nice! changed it to the proposed match

src/telemetry/mod.rs
thurstonsand: huh so it conditionally compiles based on how many implicit conversions are in scope

thurstonsand (approved):

PR #147, opened 2022-10-21
Accept anything that converts into u128 as an index

martinthomson:
Because we really don't care and this makes writing the code easier.


benjaminsavage (approved):
A thing of beauty =)

PR #149, opened 2022-10-23
Implementing the malicious security validator

benjaminsavage:
This class must be informed about all the inputs to an arithmetic circuit, informed about each multiplication (or reshare) that happens, and asked to `validate()` at the end of the circuit.

### DIFF
--- a/src/protocol/malicious.rs
+++ b/src/protocol/malicious.rs
@@ -3,7 +3,7 @@ use crate::{
     field::Field,
     helpers::{fabric::Network, Direction},
     protocol::{check_zero::check_zero, context::ProtocolContext, reveal::reveal, RecordId},
-    secret_sharing::Replicated,
+    secret_sharing::{MaliciousReplicated, Replicated},
 };
 use futures::future::try_join;
 
@@ -114,8 +114,7 @@ impl<F: Field> SecurityValidator<F> {
         &mut self,
         ctx: ProtocolContext<'_, N>,
         record_id: RecordId,
-        x: Replicated<F>,
-        rx: Replicated<F>,
+        input: MaliciousReplicated<F>,
     ) {
         // The helpers need to use the same shared randomness to generate the random constant used to validate a given multiplication.
         // This is a bit tricky, because we cannot count on the multiplications being executed in the same order across all the helpers.
@@ -124,8 +123,8 @@ impl<F: Field> SecurityValidator<F> {
         // That way, we don't need to worry about the order in which the multiplications are executed.
         let random_constant = ctx.prss().generate_replicated(record_id);
 
-        self.u += Self::compute_dot_product_contribution(random_constant, rx);
-        self.w += Self::compute_dot_product_contribution(random_constant, x);
+        self.u += Self::compute_dot_product_contribution(random_constant, input.rx());
+        self.w += Self::compute_dot_product_contribution(random_constant, input.x());
     }
 
     /// ## Errors
@@ -183,6 +182,7 @@ pub mod tests {
         malicious::{SecurityValidator, Step},
         QueryId, RecordId,
     };
+    use crate::secret_sharing::MaliciousReplicated;
     use crate::test_fixture::{logging, make_contexts, make_world, share, TestWorld};
     use futures::future::{try_join, try_join_all};
     use proptest::prelude::Rng;
@@ -234,17 +234,18 @@ pub mod tests {
             )
             .await?;
 
+            let a_malicious = MaliciousReplicated::new(a_shares[i], ra);
+            let b_malicious = MaliciousReplicated::new(b_shares[i], rb);
+
             v.accumulate_macs(
                 a_ctx.narrow(&Step::ValidateInput),
                 RecordId::from(0),
-                a_shares[i],
-                ra,
+                a_malicious,
             );
             v.accumulate_macs(
                 b_ctx.narrow(&Step::ValidateInput),
                 RecordId::from(1),
-                b_shares[i],
-                rb,
+                b_malicious,
             );
 
             let (ab, rab) = try_join(
@@ -264,8 +265,7 @@ pub mod tests {
             v.accumulate_macs(
                 a_ctx.narrow(&Step::ValidateMultiplySubstep),
                 RecordId::from(0),
-                ab,
-                rab,
+                MaliciousReplicated::new(ab, rab),
             );
 
             v.validate(ctx.narrow(&"SecurityValidatorValidate".to_string()))
@@ -354,8 +354,7 @@ pub mod tests {
                         v.accumulate_macs(
                             narrowed_ctx.narrow(&Step::ValidateInput),
                             *record_id,
-                            *x,
-                            *rx,
+                            MaliciousReplicated::new(*x, *rx),
                         );
                     }
 
@@ -400,8 +399,7 @@ pub mod tests {
                         v.accumulate_macs(
                             narrowed_ctx.narrow(&Step::ValidateMultiplySubstep),
                             *record_id,
-                            ab,
-                            rab,
+                            MaliciousReplicated::new(ab, rab),
                         );
                     }
 
--- a/src/protocol/malicious.rs
+++ b/src/protocol/malicious.rs
@@ -1,0 +1,306 @@
+use crate::{
+    error::BoxError,
+    field::Field,
+    helpers::{fabric::Network, Direction},
+    protocol::{check_zero::check_zero, context::ProtocolContext, reveal::reveal, RecordId},
+    secret_sharing::Replicated,
+};
+use futures::future::try_join;
+
+use serde::{Deserialize, Serialize};
+
+/// A message sent by each helper when they've computed one share of the result
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct UValue<F> {
+    payload: F,
+}
+
+#[allow(dead_code)]
+pub struct SecurityValidator<'a, F, N> {
+    internally_indexed_ctx: ProtocolContext<'a, N>,
+    r_share: Replicated<F>,
+    u: F,
+    w: F,
+}
+
+impl<'a, F: Field, N: Network> SecurityValidator<'a, F, N> {
+    #[allow(dead_code)]
+    pub fn new(ctx: &ProtocolContext<'a, N>) -> SecurityValidator<'a, F, N> {
+        let internally_indexed_ctx = ctx.narrow(&"ValidationInternalIndexed".to_string());
+        let prss = internally_indexed_ctx.prss();
+
+        let r_share = prss.generate_replicated(RecordId::from(0));
+        let (u_left, u_right): (F, F) = prss.generate_fields(RecordId::from(1));
+        let (w_left, w_right): (F, F) = prss.generate_fields(RecordId::from(2));
+
+        SecurityValidator {
+            internally_indexed_ctx,
+            r_share,
+            u: u_right - u_left,
+            w: w_right - w_left,
+        }
+    }
+
+    #[allow(dead_code)]
+    fn compute_dot_product_contribution(a: Replicated<F>, b: Replicated<F>) -> F {
+        let (a_left, a_right) = a.as_tuple();
+        let (b_left, b_right) = b.as_tuple();
+        (a_left + a_right) * (b_left + b_right) - a_right * b_right
+    }
+
+    #[allow(dead_code)]
+    fn accumulate_macs(
+        &mut self,
+        ctx: &ProtocolContext<'_, N>,
+        record_id: RecordId,
+        x: Replicated<F>,
+        rx: Replicated<F>,
+    ) {
+        // The helpers need to use the same shared randomness to generate the random constant used to validate a given multiplication.
+        // This is a bit tricky, because we cannot count on the multiplications being executed in the same order across all the helpers.
+        // The easiest way is to just narrow the context used to perform the multiplication, and then re-use the same record_id.
+        // This ensures that when the helpers all go to validate the multiplication: "1/foo/bar/baz", they all use the prss from "1/foo/bar/baz/validate".
+        // That way, we don't need to worry about the order in which the multiplications are executed.
+        let prss = ctx.narrow(&"validate".to_string()).prss();
+        let random_constant = prss.generate_replicated(record_id);
+
+        self.u += Self::compute_dot_product_contribution(random_constant, rx);
+        self.w += Self::compute_dot_product_contribution(random_constant, x);
+    }
+
+    ///
+    /// For each input, `x` of the arithmetic circuit, call this method to obtain
+    /// a secret-sharing of `r * x`.
+    /// This will also affect the information theoretic MAC that is eventually validated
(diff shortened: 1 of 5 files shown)


src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+        ctx: &ProtocolContext<'_, N>,
+        record_id: RecordId,
+        z: Replicated<F>,
+        rz: Replicated<F>,
+    ) {
+        self.accumulate_macs(ctx, record_id, z, rz);
+    }
+
+    #[allow(dead_code)]
+    pub async fn validate(&self) -> Result<bool, BoxError> {
```
akoshelev, line 110: btw, I just noticed that this function returns `bool`, does that mean that protocols will have to validate the value to be `true`? I guess the behavior will be the same (abort the whole process) so I wonder if we can make `validate` implement that (i.e. return `bool` and panic instead)? 

Btw it is important to ensure that we can reliably abort the whole process, by default I believe `panic` just terminates the thread so in some cases we may end up running query even after validator/protocol reported an issue.

We should probably talk to Martin or Andy about this, I believe we can either call `process::abort` (which is not possible to test imo) or put `panic=abort` in `Cargo.toml`. It would be great to understand how we are going to deal with it, regardless of the exact place we decide to put the validation check

martinthomson: I think that this shouldn't panic, but should instead return `Result<(), BoxError>` with a validation failure producing the `Err` arm.

Generally speaking, we should ensure that `Result::Err` causes processing to cease.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+        SecurityValidator {
+            internally_indexed_ctx,
+            r_share,
+            u: u_right - u_left,
+            w: w_right - w_left,
+        }
+    }
+
+    #[allow(dead_code)]
+    fn compute_dot_product_contribution(a: Replicated<F>, b: Replicated<F>) -> F {
```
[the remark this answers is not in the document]
benjaminsavage: I don't think so. As discussed, we probably want to make a new `MaliciousReplicated<F>` struct, and a `MaliciousProtocolContext` and put it into those.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+    ) {
+        // The helpers need to use the same shared randomness to generate the random constant used to validate a given multiplication.
+        // This is a bit tricky, because we cannot count on the multiplications being executed in the same order across all the helpers.
+        // The easiest way is to just narrow the context used to perform the multiplication, and then re-use the same record_id.
+        // This ensures that when the helpers all go to validate the multiplication: "1/foo/bar/baz", they all use the prss from "1/foo/bar/baz/validate".
+        // That way, we don't need to worry about the order in which the multiplications are executed.
+        let prss = ctx.narrow(&"validate".to_string()).prss();
+        let random_constant = prss.generate_replicated(record_id);
+
+        self.u += Self::compute_dot_product_contribution(random_constant, rx);
```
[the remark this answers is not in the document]
benjaminsavage: As we discussed today, probably the best answer is to merge these two and use streams for this.

src/protocol/malicious.rs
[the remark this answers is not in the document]
benjaminsavage: Great question. Minimally, we need to generate a new random secret-shared `r`. We can't keep using the old one as validate() causes it to be revealed. Also, all the new inputs need to be entered.

So probably it's easies to just make the `SecurityValidator` useless / throw errors if used again. It's easy enough to just create a new one.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        self.accumulate_macs(ctx, record_id, x, rx);
+        Ok(rx)
+    }
+
+    ///
+    /// Each time a multiplication is performed, it must be performed twice, in order to compute both the desired result: `z`,
+    /// as well as the result `r * z`. Once these two values are obtained, they should be passed to this function.
+    #[allow(dead_code)]
+    pub fn validate_multiplication_output(
```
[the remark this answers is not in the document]
benjaminsavage: The reason I made the different was to try to hide `r` from everything else aside from the `SecurityValidator`. This interface allows you to get the _dual_ for all your inputs. Everything else - you just perform operations with those _duals_ to keep getting _dual_ versions of all the intermediate stuff.

If you want to make the interfaces the same, that would seem to require exposing `r`. We can do that, but we just need to be really careful about what people do with it (i.e. they better not `reveal` it).

src/protocol/malicious.rs
benjaminsavage: OK, makes sense. I'll update the return type.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+            u: u_right - u_left,
+            w: w_right - w_left,
+        }
+    }
+
+    #[allow(dead_code)]
+    fn compute_dot_product_contribution(a: Replicated<F>, b: Replicated<F>) -> F {
+        let (a_left, a_right) = a.as_tuple();
+        let (b_left, b_right) = b.as_tuple();
+        (a_left + a_right) * (b_left + b_right) - a_right * b_right
```
martinthomson, line 48: Would this be easier?

```rust
(a.left() + a.right()) * (b.left() + b.right()) - a.right() * b.right()
```

It seems like most applications of "as_tuple()" would be more neatly expressed with "left()" and "right()" accessors.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+        record_id: RecordId,
+        x: Replicated<F>,
+        rx: Replicated<F>,
+    ) {
+        // The helpers need to use the same shared randomness to generate the random constant used to validate a given multiplication.
+        // This is a bit tricky, because we cannot count on the multiplications being executed in the same order across all the helpers.
+        // The easiest way is to just narrow the context used to perform the multiplication, and then re-use the same record_id.
+        // This ensures that when the helpers all go to validate the multiplication: "1/foo/bar/baz", they all use the prss from "1/foo/bar/baz/validate".
+        // That way, we don't need to worry about the order in which the multiplications are executed.
+        let prss = ctx.narrow(&"validate".to_string()).prss();
```
martinthomson, line 64: Ultimately, we'll need an enum for this, even though that will contain just a single value.  I guess you can do

```rust
struct ValidateMultiplySubstep;
impl AsRef<str> for ValidateMultiplySubstep {
   fn as_ref(&self) -> &str { "validate" }
}
```

src/protocol/malicious.rs
```diff
@@ -0,0 +1,306 @@
...
+    ///
+    #[allow(dead_code)]
+    pub async fn input(
+        &mut self,
+        ctx: &ProtocolContext<'_, N>,
+        record_id: RecordId,
+        x: Replicated<F>,
+    ) -> Result<Replicated<F>, BoxError> {
+        let rx = ctx
+            .narrow(&"validate_input".to_string())
```
martinthomson, line 85: Again, this should be an enum rather than a string.  Strings are for debugging.

martinthomson (commented):
I'm not sure about the interface to this one.

My expectation is that you present the same interface as the HBC multiply, but you use some global state to manage validation and only release the result when it is validated.

I would expect that you would have global state keyed by the step that takes some fixed number of multiplies (set in config, I think), accumulates values (x, rx, z, and rz), and then produces a future that resolves when validation on a batch is complete.  Because multiplies for a step might happen in any order on different peers, though you don't need to take values in strict order, you would need to ensure that they all finalize with the same set of records.

Much of the work that Alex is doing on #150 applies here then.  You need a structure that tracks which values have been received and which ones you are waiting on.  You can do that with a BitVec that starts out all zero or all one and you can toggle the values off/on as they come in.  When all the bits you care about are flipped, you are done and can finalize everything.

Having a more involved API will just make it hard to adapt existing code that uses multiplication.  Ideally, we make the choice of multiply part of the context, chosen with config.

src/protocol/malicious.rs
benjaminsavage: Yes, this would be better. I'll make that change.

src/protocol/malicious.rs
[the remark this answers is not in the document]
benjaminsavage: ...after some discussion with @martinthomson - we might not need streams at all. It could be that Arc<Mutex<>> is just as performant and less complex. I might wind up eliminating the other dot-product protocol entirely (but I would do that in another diff)

src/protocol/malicious.rs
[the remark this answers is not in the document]
benjaminsavage: OK, I went with your suggestion to just expose a single function in the public interface, and not treat inputs in a special way. This led to a huge simplification of the code since the SecurityValidator no longer needs to be shared across threads - only the share of "r" needs to be available across threads, and it can be safely cloned.

If we still need to access the SecurityValidator across threads, we can use Arc<Mutex<>>.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+        // This ensures that when the helpers all go to validate the multiplication: "1/foo/bar/baz", they all use the prss from "1/foo/bar/baz/validate".
+        // That way, we don't need to worry about the order in which the multiplications are executed.
+        let random_constant = ctx.prss().generate_replicated(record_id);
+
+        self.u += Self::compute_dot_product_contribution(random_constant, rx);
+        self.w += Self::compute_dot_product_contribution(random_constant, x);
+    }
+
+    #[allow(dead_code)]
+    pub async fn validate(&self) -> Result<(), BoxError> {
```
benjaminsavage, line 97: This method is the only place I use `self.ctx`. An alternative would be to NOT save that on the struct, and instead let the caller just pass it into the `validate()` function.

martinthomson: I think that I might prefer that, though you need to ensure that it is the right one.  How hard is that likely to be after running a circuit?

I can imagine that you might have a process that runs `input()` then does the body of the protocol, then runs `validate()` in the same span, but is that something that you can rely on when validation might be indirectly invoked via a call to `Reveal` or other sub-protocols?

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub fn accumulate_macs(
+        &mut self,
+        ctx: ProtocolContext<'_, N>,
+        record_id: RecordId,
+        x: Replicated<F>,
+        rx: Replicated<F>,
+    ) {
+        // The helpers need to use the same shared randomness to generate the random constant used to validate a given multiplication.
+        // This is a bit tricky, because we cannot count on the multiplications being executed in the same order across all the helpers.
+        // The easiest way is to just narrow the context used to perform the multiplication, and then re-use the same record_id.
```
martinthomson, line 87: You are no longer narrowing here.

src/protocol/malicious.rs
benjaminsavage: I think it's totally fine if it's a different context than was used at the outset. All that matters is that it's not re-using anything. So as long as you give it a brand new PRSS-Space, where it can use a couple of RecordIds (e.g. 0, 1, 2, 3, 4) you should be fine.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+        // This ensures that when the helpers all go to validate the multiplication: "1/foo/bar/baz", they all use the prss from "1/foo/bar/baz/validate".
+        // That way, we don't need to worry about the order in which the multiplications are executed.
+        let random_constant = ctx.prss().generate_replicated(record_id);
+
+        self.u += Self::compute_dot_product_contribution(random_constant, rx);
+        self.w += Self::compute_dot_product_contribution(random_constant, x);
+    }
+
+    #[allow(dead_code)]
```
martinthomson, line 96: Given how many of these you have, it might be easier to put `#![allow(dead_code)]` at the top of the module.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+                record_3,
+                UValue { payload: self.u },
+            ),
+            channel.send(
+                self.ctx.role().peer(Direction::Right),
+                record_4,
+                UValue { payload: self.w },
+            ),
+        )
+        .await?;
```
martinthomson, line 117: I want to flag this await here.  You don't use the value, nor do you depend on this operation completing in any way.  However, I believe that this is the correct approach anyway, because this is how we will ultimately manage back pressure.  Inability to send will then block further processing at this point.

This is not obvious and there is a risk that someone decides to get smart and defer the await on this promise until the very end.

The same does not apply to the use of `try_join()` here, though I think that you could just do two `channel.send(...).await?` calls in sequence.  The two records are being sent on the same channel, so you don't need to allow for them arriving out of order.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+        let record_3 = RecordId::from(3);
+        let record_4 = RecordId::from(4);
+        let record_5 = RecordId::from(5);
+        let record_6 = RecordId::from(6);
+
+        // send our `u_i+1` value to the helper on the right
+        let channel = self.ctx.mesh();
+        try_join(
+            channel.send(
+                self.ctx.role().peer(Direction::Right),
```
martinthomson, line 107: Maybe factor out this expression; it's causing your lines to be wrapped.

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+    use crate::field::Fp31;
+    use crate::protocol::{
+        malicious::{SecurityValidator, Step},
+        QueryId, RecordId,
+    };
+    use crate::test_fixture::{logging, make_contexts, make_world, share, TestWorld};
+    use futures::future::{try_join, try_join_all};
+    use proptest::prelude::Rng;
+
+    #[tokio::test]
```
martinthomson, line 158: Some commentary here might help establish what the goal of the test is.  It's easy to follow, but...

src/protocol/malicious.rs
```diff
@@ -0,0 +1,350 @@
...
+        let a_shares = share(a, &mut rng);
+        let b_shares = share(b, &mut rng);
+
+        let futures = (0..3).into_iter().zip(context).map(|(i, ctx)| async move {
+            let mut v =
+                SecurityValidator::new(ctx.narrow(&"SecurityValidatorInternal".to_string()));
+            let r_share = v.r_share();
+
+            let a_record_narrowed_context = ctx.narrow(&"record_1_hack".to_string());
+            let b_record_narrowed_context = ctx.narrow(&"record_2_hack".to_string());
```
martinthomson, line 178: I think that you just want:

```rust
a_ctx = ctx.narrow("multiply");
b_ctx = ctx.clone();
```

Rather than having each narrow separately.

Then you use `narrow()` each of those for each of:

* a*r or b*r
* accumulate_macs
* the multiply of a*b
* the multiply of a*(b*r) or b*(a*r)
* the multiply of r*(a*b)

martinthomson (commented):
A solid start on building a useful engine.  I'm too tired to leave good feedback, so this is what you get.

PR #151, opened 2022-10-25
Making SecureMul just take a context

benjaminsavage:
This one protocol looked different from all the others. That seems weird. Let's make them more similar.


martinthomson (approved):
So much cleaner, thanks.

PR #152, opened 2022-10-26
Adding a new struct to help with upgrading to malicious security

benjaminsavage:
To upgrade to malicious security, we need to run every arithmetic circuit twice; one with the original numbers, and once with those numbers times a random, secret-shared value `r`. All of the local operations (like addition, subtraction, negation, and multiplication with a constant) can be done without communication between the helpers. We just need to do them both locally. This struct should help with that. It overrides the basic math operators to ensure all of the accounting is done twice.

### DIFF
--- a/src/secret_sharing/malicious_replicated.rs
+++ b/src/secret_sharing/malicious_replicated.rs
@@ -1,0 +1,186 @@
+use std::fmt::Formatter;
+use std::ops::AddAssign;
+use std::ops::SubAssign;
+use std::{
+    fmt::Debug,
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use crate::field::Field;
+use crate::helpers::Identity;
+use crate::secret_sharing::Replicated;
+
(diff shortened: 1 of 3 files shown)


src/secret_sharing/malicious_replicated.rs
```diff
@@ -0,0 +1,186 @@
...
+    fmt::Debug,
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use crate::field::Field;
+use crate::helpers::Identity;
+use crate::secret_sharing::Replicated;
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct MaliciousReplicated<T>(Replicated<T>, Replicated<T>);
```
martinthomson, line 14: I would prefer 
```rust
pub struct MaliciousReplicated<F> {
  x: Replicated<F>,
  rx: Replicated<F>,
}
```

benjaminsavage: Done!

PR #154, opened 2022-10-26
A Variant of the secure multiplication protocol used to provide malic…

benjaminsavage:
When upgrading our protocols to Malicious Security, multiplications are a bit different. For this reason, I've developed another protocol. Multiplications basically have to be doubled, twice as much communication. They also affect the information-theoretic MACs. To avoid the risk of someone forgetting to do this update, I've put that update into the multiplication protocol itself. Since many multiplications could be happening in parallel across multiple threads, it's clear that these information-theoretic MACs are shared state. I've chosen to approach this problem by using Arc<Mutex<>>


src/protocol/maliciously_secure_mul.rs
```diff
@@ -0,0 +1,346 @@
...
+        let duplicate_ctx = self.ctx.narrow(&Step::DuplicateMultiply);
+        let (ab, rab) = try_join(
+            SecureMul::new(self.ctx, self.record_id).execute(a.x(), b.x()),
+            SecureMul::new(&duplicate_ctx, self.record_id).execute(a.rx(), b.x()),
+        )
+        .await?;
+
+        let malicious_ab = MaliciousReplicated::new(ab, rab);
+
+        self.security_validator.lock().unwrap().accumulate_macs(
```
akoshelev, line 93: this will make all multiplications to execute serially one after the other. I wonder if `SecurityValidator` should just be `Send` and provide `accumulate_macs` method that takes `&self`. We could back this implementation by a channel. I wonder how crazy it sounds to implement fields backed by atomic value rather than just an integer...

martinthomson: I don't think that this serializes multiplications if this action is short.  It does block, but a short bit of non-concurrent code might be OK.  I might prefer to hide the implementation details (mutex vs. channel) under an API though.  That means creating a wrapper that either holds the producer end of an mpsc channel or the `Arc<Mutex<X>>`.

benjaminsavage:
Abandoning this. I split it into 3 smaller PRs:

https://github.com/private-attribution/ipa/pull/163
https://github.com/private-attribution/ipa/pull/164
https://github.com/private-attribution/ipa/pull/169

PR #155, opened 2022-10-27
Update the CI checks

martinthomson:
The actions-rs/cargo action isn't needed and the toolchain install is a little out of date now.

I've switched to one that seems better, but now I can't change organization settings to remove the others and enable these, so I guess it will have to wait.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,40 +6,27 @@ jobs:
     env:
       CARGO_TERM_COLOR: always
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v3
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: stable
+          components: clippy,rustfmt
 
-      - uses: actions-rs/cargo@v1
-        name: Build
-        with:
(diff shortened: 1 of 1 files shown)


martinthomson:
https://github.com/private-attribution/ipa/settings/actions is where the setting is, but it is disabled for me right now.

akoshelev (approved):
👍

akoshelev:
I can't enable it either (I don't seem to have permissions for it), will wait for @martinthomson to enable it manually. 

Right now, actions are not running on new PRs
https://github.com/private-attribution/ipa/actions/runs/3341296915
```
dtolnay/rust-toolchain@stable is not allowed to be used in private-attribution/ipa. Actions in this workflow must be: within a repository owned by private-attribution, created by GitHub, or matching the following: actions-rs/toolchain@v1, actions-rs/cargo@v1.
```

akoshelev:
we also have some weird clippy failures that neither Thurston nor I can reproduce locally, so hoping that this will fix them
https://github.com/akoshelev/raw-ipa/actions/runs/3340348137/jobs/5530341427

akoshelev:
we need to revert this as there is something wrong with GitHub that does not allow Martin to change the actions engine

PR #157, opened 2022-10-27
Refactoring: move field.rs to ff/

taikiy:
I'm planning to add new trait/struct to represent binary shares, which will implement some of the traits `Field` implements. I'm moving field.rs under secret_sharing/ before making those changes.


akoshelev:
How about moving it to a new module `ff` (finite fields) or `algebra` instead? Secret sharing module on its own may be large

taikiy:
My intention for this refactoring was to add the `Binary` trait, which is another secret shared value type but for Boolean circuits. I wanted to put both `Field` and `Binary` traits under the same folder since they are like siblings, but I couldn't think of a good name that group them together. Are there good names for secret sharing "schemes" and secret shared "values"?

akoshelev:
I am probably asking silly questions, isn't it a binary field $Z_2$ definition? so rather than a trait, we could just implement `Field` for it?

taikiy:
Yes, the binary field is basically $Z_2$. That said, we should probably create a separate trait for boolean circuit protocols to be explicit about the input/output type constraints.

`Field` also has `type Integer: Int` alias which works great with integer fields. For bit shares, we want to use `bool` to reduce the size. We could change the definition of `Int` to support `bool`, but together with the reason above, I thought it'd be better to create a separate trait than using `Field` to represent bits.

akoshelev:
> Yes, the binary field is basically . That said, we should probably create a separate trait for boolean circuit protocols to be explicit about the input/output type constraints. e.g. Some arithmetic protocols may accept fields with different primes except , vice versa.

In this case I think such protocols should accept the concrete field implementation (i.e. $Z_2$ or $Z_p$) rather than be generic over `F`? 


> `Field` also has `type Integer: Int` alias which works great with integer fields. For bit shares, we want to use `bool` to reduce the size. We could change the definition of `Int` to support `bool`, but together with the reason above, I thought it'd be better to create a separate trait than using `Field` to represent bits.

`Int` is a trait so it does not imply how much space field needs to occupy in memory. It should be totally doable to create binary field backed by a `u8`.

taikiy:
> In this case I think such protocols should accept the concrete field implementation (i.e. $Z_2$ or $Z_p$ ) rather than be generic over F?

Wait, I thought $Z_p$ is the `Field` and a concrete implementations are `Fp2`, `Fp31`, `Fp127`, etc.. Modulus conversion, for example, takes $Z_2$ as the input and outputs any $Z_p$. Shouldn't the output be generic over `Field`, but restrict the output from being $Z_2$?

akoshelev:
we can narrow down the field trait even further by defining `PrimeField : Field` marker trait (for now?).  In this case your conversion may take `PrimeField` as input and produce values in $Z_2$.  We could also take `F: Field` and panic if F is $Z_2$ I guess...

taikiy:
OK, I moved the file under `ff/` folder. Let's continue the discussion in another PR where I add Fp2 struct.

akoshelev (approved):

benjaminsavage (approved):

PR #158, opened 2022-10-27
Binary trait for single-bit secret shared values

taikiy:
Similar to the `Field` trait for arithmetic circuits, this diff adds the `Binary` trait for boolean circuits. `Replicated<B: Binary>` will be used as inputs/outputs of boolean circuit protocols, such as bit-decomposition and bit-wise comparison.

Secret sharing structs such as `Replicated<T>` need to be able to take either `Field` or `Binary` trait as its generic type. For that, I've extracted common traits that both Field and Binary should implement in order for us to send them over the network, and created the `Share` trait.


taikiy:
I will make a substantial change based on the discussion in #157. Closing this one for now.

PR #159, opened 2022-10-27
Byte serialization for field values

akoshelev:
As I am preparing to start using contiguous buffers on infra side, serde serialization of messages and fields needs to get out of the way. Fields need to be able to transform themselves into a sequence of bytes and back because that is how infra will store and interpret them. 

It is not the most efficient implementation because it still heap allocate all the time, but paves the path to have long byte buffers and array-backed payloads.

This also removes the annoying constraint that field values need to be wrapped before protocols can send them, so this change takes care of that too.

This was previously part of #150 but it grew too large and no longer reviewable in my opinion. So I am submitting it piece by piece.


src/field.rs
```diff
@@ -91,6 +91,52 @@ pub trait Field:
...
+    /// Generic implementation to serialize fields into a buffer. Callers need to make sure
+    /// there is enough capacity to store the value of this field.
+    /// It is less efficient because it operates with generic representation of fields as 16 byte
+    /// integers, so consider overriding it for actual field implementations
+    ///
+    /// ## Errors
+    /// Returns an error if buffer did not have enough capacity to store this field value
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        let bytes = Self::Integer::BITS as usize / 8;
+        let raw_value = &self.as_u128().to_le_bytes()[..bytes];
```
martinthomson, line 104: This is the allocation. Probably harmless and can be optimised out automatically

src/field.rs
```diff
@@ -91,6 +91,52 @@ pub trait Field:
...
+    /// Generic implementation to deserialize fields from buffer.
+    /// It is less efficient because it allocates 16 bytes on the stack to accommodate for all
+    /// possible field implementations, so consider overriding it for actual field implementations
+    ///
+    /// In the bright future when we have const generic expressions, this can be changed to provide
+    /// zero-cost generic implementation
+    ///
+    /// ## Errors
+    /// Returns an error if buffer did not have enough capacity left to read the field value.
+    fn deserialize<R: Read>(reader: &mut R) -> io::Result<Self> {
```
martinthomson, line 130: Is there a reason you are using this specific api? You aren’t implementing a trait so this could be as simple as taking a slice (mut for write) and panicking if something goes wrong ( as it really shouldn’t)

src/helpers/messaging.rs
```diff
@@ -78,14 +109,13 @@ impl<N: Network> Mesh<'_, '_, N> {
         record_id: RecordId,
         msg: T,
     ) -> Result<(), Error> {
-        let bytes = serde_json::to_vec(&msg)
-            .map_err(|e| Error::serialization_error(record_id, self.step, e))?
-            .into_boxed_slice();
+        let mut c = Cursor::new(Vec::with_capacity(T::BYTES as usize));
```
martinthomson, line 112: Is the other allocation. Can we use smallvec here This

martinthomson: If we want to cut an allocation we can do that later

martinthomson (approved):

src/field.rs
akoshelev: that's right and I started with taking `&mut [u8]` (in fact #150 has it). The thing that I didn't like (and maybe I am wrong) is that writer needs to know the offset to write the value. If we were to re-use a single slice to write two values for instance, something needs to take care of providing two different slices or make `offset` an argument to the API. 

but if our API has the shape of `&mut [u8], offset: usize`, then a typical `Write`/`Read` implementation takes care of offset tracking nicely.

I don't have strong feelings about it, if you think there is a good reason to take `&mut [u8]` instead, I can do that

src/helpers/messaging.rs
akoshelev: totally, that will be the end goal. did not want to change it here because `SendBuffer` will be affected as well as `Network` trait.

src/field.rs
martinthomson: I'm just concerned that we will need to do out-of-order writes, which aren't easy to do with a writer.

(For byte-level placements, we don't need an offset argument as we can just pass in the slice that we want the field to write into.  We only need an offset if there is a need to do bit-level placement.  Avoid that for now.)

src/field.rs
akoshelev: oh wow, I didn't know that. I tried to find it but failed, can you hint me where it happens?

martinthomson: Maybe it doesn't heap allocate, but `to_le_bytes()` at least makes a copy of the `u128` as `[u8; 16]`.  It could be aliased to the original value (though not portably) without a copy, but that's not how they did it.

akoshelev: not that it matters at this point, but I thought it should be possible to just reinterpret u128 as `[u8; 16]` in LE order? surely for BE we need to make a copy, but I was hopeful that it is not happening here...

src/field.rs
akoshelev: let me make that change then - I'd love to learn more about out-of-order writes later. Is it because we need to convert BE to LE?

akoshelev: done

src/field.rs
martinthomson: Let's not compromise portability that way.

akoshelev:
@martinthomson I think I addressed most of your comments, do you think it can be merged?

benjaminsavage:
I'm merging now. We can always continue to improve.

PR #163, opened 2022-10-28
Make context.multiply() consume the context

benjaminsavage:
This was @martinthomson's recommendation. I think it helps us utilize the "step" stuff better.

### DIFF
--- a/src/protocol/sort/bit_permutations.rs
+++ b/src/protocol/sort/bit_permutations.rs
@@ -8,8 +8,11 @@ use crate::{
 use crate::helpers::fabric::Network;
 use embed_doc_image::embed_doc_image;
 use futures::future::try_join_all;
-/// Generate bit permutations for a given bit column of query.
-/// This is GENBITPERM(Algorithm 3) described in <https://eprint.iacr.org/2019/695.pdf>.
+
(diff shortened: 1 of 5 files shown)


src/protocol/sort/bit_permutations.rs
```diff
@@ -74,9 +47,28 @@ impl<'a, F: Field> BitPermutations<'a, F> {
...
+        let mult_input = self
+            .input
+            .iter()
+            .map(move |x: &Replicated<F>| share_of_one - *x)
+            .chain(self.input.iter().copied())
+            .enumerate()
+            .scan(Replicated::<F>::new(F::ZERO, F::ZERO), |sum, (index, x)| {
+                *sum += x;
+                Some((
+                    ctx.narrow(&format!("record_{}", index)),
```
martinthomson, line 61: This looks like a mistake to me.  Any reason not to clone the context as here?

As in std::iter::repeat()

benjaminsavage: I'm not sure exactly what you're suggesting I should replace this with.

If I swap this line with `ctx.clone()` it does not work. It tells me: 
`cannot move out of `*ctx` which is behind a shared reference`

I'm not sure how you're suggesting I use `std::iter::repeat()`, could you share a snippet of code?

PR #165, opened 2022-10-28
Secure apply inv

richajaindce:
This protocol will apply a secret shared permutation on secret shared input across three helpers


src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+        &self,
+        ctx: &ProtocolContext<'_, N>,
+    ) -> Result<Vec<F>, BoxError> {
+        let reveals = self
+            .permutation
+            .iter()
+            .enumerate()
+            .map(|(index, input)| async move {
+                reveal(
+                    ctx.narrow(&index.to_string()),
```
richajaindce, line 96: Not sure if this is correct? I am calling a narrow per record ID for reveal. But removing this gives :"'Refined 'protocol' with step 'reveal_permutation' twice'"

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        try_join!(h0_future, h1_future).unwrap();
+
+        // 4. Reveal permutation to all helpers
+        let permutation = self
+            .reveal_permutations(&ctx.narrow(&RevealPermutation))
+            .await
+            .unwrap();
+        let mut perms = Vec::new();
+        // TODO(richa) : This is ugly, converted F -> u128 -> usize needed for getting an appropriate Permutation to apply.
```
richajaindce, line 71: Is there a better way to write this? @akoshelev @thurstonsand

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+    #[allow(dead_code)]
+    pub async fn execute<N: Network>(
+        &mut self,
+        ctx: &ProtocolContext<'_, N>,
+    ) -> Result<(), BoxError> {
+        // 1. Generate permutation
+        let permutations = generate_random_permutation(self.input.len(), &ctx.prss());
+        // 2. Shuffle permutations
+        let mut shuffle_permutation = Shuffle::new(self.permutation);
+        let h0_future = shuffle_permutation.execute(ctx.narrow(&ShufflePermutation), &permutations);
```
akoshelev, line 57: There is a very subtle clone that I am almost certain is happening here under the hood. `Permutation` when applied must mutate its own state, so you have to pass it by exclusive reference (`&mut`). The fact that shared reference is used makes me suspicious that in order to actually apply the permutation, you clone it to obtain a mutable ref.


I think a better design would be to take `&mut Permutation` and clone where you need to apply it twice in parallel (right here I presume)

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// This is APPLYINV(Algorithm 4) described in <https://eprint.iacr.org/2019/695.pdf>.
+/// This is a protocol that applies the inverse of a shared permutation to a shared-vector representation.
+/// Input: Each helpers know their own secret shares of input and permutation
+/// Output: At the end of the protocol, all helpers receive inputs after the permutation is applied
+#[derive(Debug)]
+#[embed_doc_image("secureapplyinv", "images/sort/secureapplyinv.png")]
+pub struct SecureApplyInv<'a, F> {
+    input: &'a mut Vec<Replicated<F>>,
+    permutation: &'a mut Vec<Replicated<F>>,
```
akoshelev, line 31: qq just to clarify my understanding - does permutation need to be in the same field as input? I assume that is just secret shared indices, so is it possible that we may decide to use $p = 2^{61}$ for element field and $p = 2^{31}$ for permutation? Does it make sense or is it even possible to take two different fields here `F1` and `F2`?

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+            let h1_future = secure_applyinv1.execute(&context[1]);
+            let h2_future = secure_applyinv2.execute(&context[2]);
+
+            try_join!(h0_future, h1_future, h2_future).unwrap();
+
+            assert_eq!(input_shares.0.len(), BATCHSIZE);
+            assert_eq!(input_shares.1.len(), BATCHSIZE);
+            assert_eq!(input_shares.2.len(), BATCHSIZE);
+
+            // We should get the same result of applying inverse as what we get when applying in clear
```
akoshelev, line 158: the same code has been used in bit permutations, shall we define a helper function for it?

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+        let mut perms = Vec::new();
+        // TODO(richa) : This is ugly, converted F -> u128 -> usize needed for getting an appropriate Permutation to apply.
+        // Besides this makes the code dependent on Permutation struct needs which is not a good design choice.
+        for i in &permutation {
+            perms.push(i.as_u128() as usize);
+        }
+        let mut permutation = Permutation::oneline(perms);
+
+        // 5. apply the permutation on the shuffled input
+        // The paper suggests taking apply_inv. However, Permutation::oneline returns an inverted permutation, so we are negating
```
akoshelev, line 79: is it worth deviating from the paper here? would `Permutation::oneline().inverse()` do what you want?

src/protocol/sort/secureapplyinv.rs
richajaindce: Makes sense.. let me change this

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+    permutation: &'a mut Vec<Replicated<F>>,
+}
+
+impl<'a, F: Field> SecureApplyInv<'a, F> {
+    #[allow(dead_code)]
+    pub fn new(input: &'a mut Vec<Replicated<F>>, permutation: &'a mut Vec<Replicated<F>>) -> Self {
+        Self { input, permutation }
+    }
+    /// This algorithm applies a permutation on the inputs. The permutation is not available in clear and
+    /// we do not want to reveal the actual permutation to any of the helpers while applying it.
```
benjaminsavage, line 40: /// 
/// This algorithm applies a permutation to the `input` vector. The permutation is secret-shared, 
/// and none of the helpers should learn it through this protocol.
/// To keep the permutation secret, it (and the inputs) are first randomly securely shuffled.
/// After this shuffle, the permutation can be revealed. 
/// An adversary can only obtain a shuffled permutation, which is just a random permutation.

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// 3. Secret shared value is shuffled using the same permutation as 1
+    /// 4. The permutation is revealed
+    /// 5. All helpers call `apply` to apply the permutation locally.
+    #[allow(dead_code)]
+    pub async fn execute<N: Network>(
+        &mut self,
+        ctx: &ProtocolContext<'_, N>,
+    ) -> Result<(), BoxError> {
+        // 1. Generate permutation
+        let permutations = generate_random_permutation(self.input.len(), &ctx.prss());
```
benjaminsavage, line 54: Is it just one permutation, or multiple? The function name is singular but the variable is plural.

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+        // 2. Shuffle permutations
+        let mut shuffle_permutation = Shuffle::new(self.permutation);
+        let h0_future = shuffle_permutation.execute(ctx.narrow(&ShufflePermutation), &permutations);
+
+        // 3. Shuffle inputs
+        let mut shuffle_input = Shuffle::new(self.input);
+        let h1_future = shuffle_input.execute(ctx.narrow(&ShuffleInputs), &permutations);
+
+        try_join!(h0_future, h1_future).unwrap();
```
benjaminsavage, line 63: I think it'll be more readable to compress this into just 4 lines of code with a lot less local variable names.
It also starts to show just how confusing it is to have both a local variable `permutations` which is actually very different from `self.permutation`. So I suggest a slightly different naming scheme.

```
let (shuffled_inputs, shuffled_permutation) = try_join(
    Shuffle::new(self.input).execute(ctx.narrow(&ShuffleInputs), &random_permutations),
    Shuffle::new(self.sort_permutation).execute(ctx.narrow(&ShufflePermutation), &random_permutation),
).await?;

File: src/protocol/sort/secureapplyinv.rs (hunk 6 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs:79]: I'm totally confused by this comment.

File: src/protocol/sort/secureapplyinv.rs (hunk 1 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs:96]: Honestly, I can never get this right either. I suggest just land whatever hack works, and we will do a cleanup pass over the entire "step" thing in all the protocols later.

As soon as we have an end-to-end testable protocol and we can profile it, I'm sure we will find a million crazy things that are all wrong with the way we are using "narrow()".

File: src/protocol/sort/secureapplyinv.rs [hunk 10]
```diff
@@ -0,0 +1,171 @@
...
+
+        // 5. apply the permutation on the shuffled input
+        // The paper suggests taking apply_inv. However, Permutation::oneline returns an inverted permutation, so we are negating
+        // the effect by just calling apply instead of apply_inv here
+        apply(&mut permutation, &mut self.input);
+        Ok(())
+    }
+
+    #[allow(clippy::cast_possible_truncation)]
+    async fn reveal_permutations<N: Network>(
```
benjaminsavage, line 86: I don't like that this is plural, yet it only reveals a single permutation.

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    use super::SecureApplyInv;
+
+    #[tokio::test]
+    pub async fn secureapplyinv() {
+        const BATCHSIZE: usize = 25;
+        for _ in 0..10 {
+            let mut rng = rand::thread_rng();
+            let mut input = [0u128; BATCHSIZE];
+            random_number::random_fill!(input, 0..24);
```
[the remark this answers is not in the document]
benjaminsavage: Agreed. Let's not include yet another crate just for this.

benjaminsavage (changes requested):

src/protocol/sort/secureapplyinv.rs
benjaminsavage: Actually, I think the confusing thing about this is actually a much deeper issue.

I don't think it makes sense for the caller of the `Shuffle` protocol to call into the `shuffle` module and invoke the `generate_random_permutation` function themselves. This should be a private function invoked internally by the `Shuffle` protocol. 

I'd recommend changing the implementation of the `Shuffle.execute()` method to just internally `narrow` the context that is passed it, to then generate two out of the three random permutations, and then apply them.

benjaminsavage: Oh wait... nevermind... but then how do we apply the same exact random permutation to both the inputs and the shuffle_permutation... I see why you did it this way...

OK, let's at least make it somehow more clear (via the name?) that you're generating two out of the three permutations that will eventually be applied. This is (in some sense) a secret-share of the full permutation. Perhaps call it: "random_permutation_share".

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,171 @@
...
+
+            let mut secure_applyinv0 = SecureApplyInv::new(&mut input_shares.0, &mut perm_shares.0);
+            let mut secure_applyinv1 = SecureApplyInv::new(&mut input_shares.1, &mut perm_shares.1);
+            let mut secure_applyinv2 = SecureApplyInv::new(&mut input_shares.2, &mut perm_shares.2);
+
+            let h0_future = secure_applyinv0.execute(&context[0]);
+            let h1_future = secure_applyinv1.execute(&context[1]);
+            let h2_future = secure_applyinv2.execute(&context[2]);
+
+            try_join!(h0_future, h1_future, h2_future).unwrap();
```
benjaminsavage, line 152: I'm using this syntax (that I like better). It avoids macros and so many local variables:

```
let results = try_join_all(vec![
    SecureApplyInv::new(...)
        .execute(...),
    SecureApplyInv::new(...)
        .execute(...),
    SecureApplyInv::new(...)
        .execute(...),
]).await?;

File: src/protocol/sort/secureapplyinv.rs (hunk 2 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:71]: I would leave it as is, but move this inside the reveal_all_permutations function

File: src/protocol/sort/secureapplyinv.rs (hunk 4 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:31]: For now, I will let the protocol take in two fields F1 and F2 for input and permutation respectively. SecretSharing change can be handled on top of it?

File: src/protocol/sort/secureapplyinv.rs (hunk 7 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:40]: Thanks for a better comment message!

File: src/protocol/sort/secureapplyinv.rs (hunk 8 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:54]: Yes, the reason is exactly that we want same permutation to be used for both input and permutation. 
It generates two permutations per helper (left and right) that's why has a "s" in the name. Both the permutations are used while executing the shuffle protocol.
For naming, worried "share" can cause more confusion, so renaming it to "generate_random_permutations_using_prss" to reflect the role of prss hence common permutations across helpers?

File: src/protocol/sort/secureapplyinv.rs (hunk 10 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:86]: Changing it to reveal_a_permutation

File: src/protocol/sort/secureapplyinv.rs (hunk 9 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:63]: I had tried to do that but rust wasnt happy to create a variable and call a function in the same line :( temporary value dropped while borrowed
consider using a `let` binding to create a longer lived value

File: src/protocol/sort/secureapplyinv.rs (hunk 6 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:79]: Removing this comment.

File: src/protocol/sort/secureapplyinv.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:126]: random_fill is generating a random permutation. This is optional to test the functionality so removing it

File: src/protocol/sort/secureapplyinv.rs (hunk 12 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:152]: Love it! I am going to change all the protocols to have this syntax. For now, only changing SecureApplyInv.

File: src/protocol/sort/secureapplyinv.rs (hunk 5 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs:158]: Can!

File: src/protocol/sort/secureapplyinv.rs (hunk 4 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs:31]: that's fine. If you want to do that now - OK. If you want to wait until later, I'm fine with that too.

File: src/protocol/sort/secureapplyinv.rs (hunk 12 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs]: Before we batch change all of them - let's discuss. I too suggested this and @akoshelev had the idea that we would use the ::new() method to pass all of the *unique* arguments, while keeping `execute()` constant, i.e. always accepting the ProtocolContext and RecordId.

Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs]: I started making the changes in Shuffle (acted too fast). Reverting them for now

File: src/protocol/sort/secureapplyinv.rs (hunk 4 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs]: ACTUALLY, I cant do this. With the new changes in SecurityValidatorAccumulator, it expects a single type of F whereas for validating shuffles we would need two F's (one for input and one for permutation). Lets discuss this offline.
For this PR, leaving it as F.

Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs]: Ok, let’s leave it for now - and discuss this later.

File: src/protocol/sort/secureapplyinv.rs (hunk 8 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs]: That should help. Let’s pick a different local variable name as well. Something like “two_of_three_permutations”

File: src/protocol/sort/secureapplyinv.rs (hunk 12 above)
Comment (benjaminsavage) [reply src/protocol/sort/secureapplyinv.rs]: :). Let’s land this PR first. It’s really close.

File: src/protocol/sort/secureapplyinv.rs (hunk 1 above)
Comment (richajaindce) [reply src/protocol/sort/secureapplyinv.rs]: PR #177 makes it possible to reuse same ctx and we dont need to use string any more. So, I am rebasing on top of that PR

File: src/protocol/reveal.rs [hunk 13]
```diff
@@ -36,6 +40,27 @@ pub async fn reveal<F: Field>(
     Ok(input.left() + input.right() + share)
 }
 
+/// Given a vector containing secret shares of a permutation, this returns a revealed permutation.
+/// This executes `reveal` protocol on each row of the vector and then constructs a `Permutation` object
+/// from the revealed rows.
+#[allow(clippy::cast_possible_truncation, clippy::module_name_repetitions)]
+pub async fn reveal_a_permutation<F: Field>(
+    ctx: ProtocolContext<'_, F>,
+    permutations: &mut [Replicated<F>],
```
benjaminsavage, line 49: Why is this plural? Are there more than one permutation? It appears to be a list of secret shared numbers, which I assume are positions from 1..n.

src/protocol/reveal.rs
```diff
@@ -36,6 +40,27 @@ pub async fn reveal<F: Field>(
...
+) -> Result<Permutation, BoxError> {
+    let reveals =
+        zip(repeat(ctx), permutations)
+            .enumerate()
+            .map(|(index, (ctx, input))| async move {
+                reveal(ctx, RecordId::from(index as u32), *input).await
```
benjaminsavage, line 55: you shouldn't need the `index as u32` now. Looking at the example on line 67 of bit_permutation.rs - there is no cast.

src/protocol/reveal.rs
```diff
@@ -36,6 +40,27 @@ pub async fn reveal<F: Field>(
...
+            });
+    let permutation = try_join_all(reveals).await?;
```
benjaminsavage, line 57: I'd recommend just combining this with the part above. No need for a local variable `reveals`. Let's just inline it.

src/protocol/reveal.rs
```diff
@@ -36,6 +40,27 @@ pub async fn reveal<F: Field>(
...
+    let mut perms = Vec::new();
+    for i in permutation {
+        perms.push(i.as_u128() as usize);
+    }
+    Ok(Permutation::oneline(perms))
```
benjaminsavage, line 62: I'm pretty sure we will eventually get rid of this permutations crate if it makes us do all these extra operations and use all this extra memory.

src/test_fixture/sharing.rs
```diff
@@ -34,3 +36,16 @@ pub fn validate_and_reconstruct<T: Field>(
 
     input.0.left() + input.1.left() + input.2.left()
 }
+
+/// Validates expected result from the secret shares obtained.
+///
+/// # Panics
+/// Panics if the expected result is not same as obtained result. Also panics if `validate_and_reconstruct` fails
+pub fn validate_result_from_shares(expected_result: &[u128], result: &ReplicatedShares) {
```
benjaminsavage, line 44: The name of this function in no way indicates that this is only to be used to validate slices / arrays / lists. Let's call it something like `validate_list_of_shares`

src/test_fixture/sharing.rs
```diff
@@ -34,3 +36,16 @@ pub fn validate_and_reconstruct<T: Field>(
...
+    (0..result.0.len()).for_each(|i| {
```
benjaminsavage, line 45: Worth adding an assert that the lengths are the same?

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,124 @@
...
+        ctx: &ProtocolContext<'_, F>,
+        input: &'_ mut Vec<Replicated<F>>,
+        sort_permutation: &'_ mut Vec<Replicated<F>>,
+    ) -> Result<(), BoxError> {
+        let mut random_permutations =
+            get_two_of_three_random_permutations(input.len(), &ctx.prss());
+        let mut random_permutations_copy = random_permutations.clone();
+
+        let mut shuffle_inputs = Shuffle::new(input);
+        let mut shuffle_permutation = Shuffle::new(sort_permutation);
```
benjaminsavage, line 54: I'm surprised that these need to be mutable. I understand that input and sort_permutation are both mutated, is that why?

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let (_, _) = try_join!(
```
benjaminsavage, line 56: You don't need to use the macro version. You can just use `try_join` from the futures crate:
```
use futures::future::try_join; 
```

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,124 @@
...
+            shuffle_inputs.execute(ctx.narrow(&ShuffleInputs), &mut random_permutations,),
+            shuffle_permutation.execute(
```
benjaminsavage, line 58: I'd prefer to just inline it, as the local variable names aren't adding anything (if anything, they confuse me a bit):
```
Shuffle::new(input).execute(...),
Shuffle::new(sort_permutation).execute(...),
```

src/protocol/sort/secureapplyinv.rs
```diff
@@ -0,0 +1,124 @@
...
+            let mut rng = rand::thread_rng();
+            let input: Vec<u128> = (0..(BATCHSIZE as u128)).collect();
+
+            let mut permutation: Vec<usize> = (0..BATCHSIZE).collect();
+            permutation.shuffle(&mut rng);
+
+            let mut expected_result = input.clone();
+            let mut cloned_perm = Permutation::oneline(permutation.clone()).inverse();
+
+            apply_inv(&mut cloned_perm, &mut expected_result);
```
benjaminsavage, line 97: I find this extremely difficult to follow.

It's really strange that `expected_result` is mutable. Maybe some comments would help? All of the cloning definitely leaves me saying "huh? what just happenned?" And this apply_inv of the inverse of something... all of these "inverses" confuse me a lot.

benjaminsavage (approved):
Looks good. Just a few little comments - then let's merge this!

src/protocol/sort/secureapplyinv.rs
akoshelev: Yep, I added this topic to our agenda for Nov offsite

src/protocol/reveal.rs
richajaindce: Yes this trait impl was added in Martin's latest PR. Changing this

src/protocol/reveal.rs
richajaindce: :(

src/protocol/sort/secureapplyinv.rs
richajaindce: Yes and shuffle changes the datamembers i.e. input and sort_permutation (to save on memory)

src/protocol/sort/secureapplyinv.rs
richajaindce: Oh try_join! didnt let me do this but try_join was ok. Changing it

src/protocol/sort/secureapplyinv.rs
richajaindce: Since SecureApplyInv changes the inputs sent to it, ended up cloning them to validate the results.
The actual paper expects us to apply an inverse on the inverted Permutation (i.e. apply_inv(perm.inverse(), input)).
Since this is same as apply(perm, input), we are doing this instead now both in the code and in the test.
Added this comment to reflect correctly

PR #164, opened 2022-10-28
Creating the accumulator state for the eventual malicious validation

benjaminsavage:
Next step along the way towards eventually creating maliciously secure multiplication.

PR #169, opened 2022-10-30
Adding a Maliciously Secure multiplication protocol

benjaminsavage:
The 3rd step in a stack of 3 diffs to upgrade our arithmetic circuits to malicious security.

This introduces a new "Maliciously Secure Multiplication" protocol, which performs a multiplication twice, once to compute `a * b`, and once to produce `r * a * b`. This protocol automatically updates the information theoretic MACs that will eventually be checked at the conclusion of the protocol.

### DIFF
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -95,9 +95,7 @@ impl<'a, F: Field> ProtocolContext<'a, F> {
     pub fn prss_rng(&self) -> (SequentialSharedRandomness, SequentialSharedRandomness) {
         self.prss.sequential(&self.step)
     }
-}
 
-impl<'a, F: Field> ProtocolContext<'a, F> {
     /// Get a set of communications channels to different peers.
     #[must_use]
     pub fn mesh(&self) -> Mesh<'_, '_> {
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use super::{
+    maliciously_secure_mul::MaliciouslySecureMul,
     prss::{IndexedSharedRandomness, SequentialSharedRandomness},
     securemul::SecureMul,
     RecordId, Step, UniqueStepId,
@@ -26,7 +27,7 @@ pub struct ProtocolContext<'a, F> {
     accumulator: Option<SecurityValidatorAccumulator<F>>,
 }
 
-impl<'a, F> ProtocolContext<'a, F> {
+impl<'a, F: Field> ProtocolContext<'a, F> {
     pub fn new(role: Identity, participant: &'a PrssEndpoint, gateway: &'a Gateway) -> Self {
         Self {
             role,
@@ -69,9 +70,7 @@ impl<'a, F> ProtocolContext<'a, F> {
             step: self.step.narrow(step),
             prss: self.prss,
             gateway: self.gateway,
-            // TODO: make this work
-            // accumulator: self.accumulator, //TODO: make this work
-            accumulator: None, // God help me, I just can't make this work
+            accumulator: self.accumulator.clone(),
         }
     }
 
@@ -117,10 +116,8 @@ impl<'a, F: Field> ProtocolContext<'a, F> {
     /// ## Panics
     /// If you failed to upgrade to malicious protocol context
     #[allow(clippy::unused_async)] // eventually there will be await b/c of backpressure implementation
-    pub async fn malicious_multiply(self, _record_id: RecordId) {
-        // -> MaliciouslySecureMul<'a, N, F> {
-        let _accumulator = self.accumulator.as_ref().unwrap().clone();
-        // TODO: next diff!
-        // MaliciouslySecureMul::new(self, record_id, accumulator)
+    pub async fn malicious_multiply(self, record_id: RecordId) -> MaliciouslySecureMul<'a, F> {
+        let accumulator = self.accumulator.as_ref().unwrap().clone();
+        MaliciouslySecureMul::new(self, record_id, accumulator)
     }
 }
--- a/src/protocol/malicious.rs
+++ b/src/protocol/malicious.rs
@@ -202,7 +202,9 @@ pub mod tests {
         QueryId, RecordId,
     };
     use crate::secret_sharing::{MaliciousReplicated, Replicated};
-    use crate::test_fixture::{logging, make_contexts, make_world, share, TestWorld};
+    use crate::test_fixture::{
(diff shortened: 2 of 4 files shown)


src/protocol/context.rs
```diff
@@ -70,9 +71,7 @@ impl<'a, N, F> ProtocolContext<'a, N, F> {
             step: self.step.narrow(step),
             prss: self.prss,
             gateway: self.gateway,
-            // TODO: make this work
-            // accumulator: self.accumulator, //TODO: make this work
-            accumulator: None, // God help me, I just can't make this work
+            accumulator: self.accumulator.clone(),
```
benjaminsavage, line 74: It was introducing that bound on `F` that helped me finally resolve this.

src/protocol/context.rs
```diff
@@ -118,10 +117,8 @@ impl<'a, N: Network, F: Field> ProtocolContext<'a, N, F> {
     /// ## Panics
     /// If you failed to upgrade to malicious protocol context
     #[allow(clippy::unused_async)] // eventually there will be await b/c of backpressure implementation
-    pub async fn malicious_multiply(self, _record_id: RecordId) {
-        // -> MaliciouslySecureMul<'a, N, F> {
-        let _accumulator = self.accumulator.as_ref().unwrap().clone();
-        // TODO: next diff!
-        // MaliciouslySecureMul::new(self, record_id, accumulator)
+    pub async fn malicious_multiply(self, record_id: RecordId) -> MaliciouslySecureMul<'a, N, F> {
+        let accumulator = self.accumulator.as_ref().unwrap().clone();
```
benjaminsavage, line 121: This line is... a real mouthful. I have to admit I don't really follow all of these steps.

src/protocol/malicious.rs
```diff
@@ -284,30 +286,31 @@ pub mod tests {
...
+                .malicious_multiply(RecordId::from(0))
+                .await
+                .execute(a_malicious, b_malicious)
+                .await?;
 
-            acc.accumulate_macs(
-                &a_ctx.narrow(&Step::ValidateMultiplySubstep).prss(),
-                RecordId::from(0),
-                MaliciousReplicated::new(ab, rab),
-            );
```
benjaminsavage, line 305: This happens within the Maliciously Secure Mult protocol now.

src/protocol/malicious.rs
```diff
@@ -284,30 +286,31 @@ pub mod tests {
...
+            v.validate(ctx.narrow("SecurityValidatorValidate")).await?;
 
-            v.validate(ctx.narrow("SecurityValidatorValidate")).await
+            Ok::<(MaliciousReplicated<Fp31>, Replicated<Fp31>), BoxError>((mult_result, r_share))
```
benjaminsavage, line 298: It's super annoying that it can't figure out what kind of Error this OK should use, and thus I have to include this entire long line of garbage.

src/protocol/context.rs
martinthomson: Right.  You should now merge the two impls of ProtocolContext, as the second already had that bound.

src/protocol/context.rs
martinthomson: You could try `self.accumulator.cloned().unwrap();` instead, I guess.

src/protocol/malicious.rs
martinthomson: You might be able to replace pieces of this with `_` or put an explicit return type on the closure above (like `.map(|(i, ctx)| -> Result<(MaliciousReplicated<_>, Replicated<_>), BoxError> async move {`

src/protocol/malicious.rs
```diff
@@ -333,7 +336,8 @@ pub mod tests {
     /// MACs are compared. If any helper deviated from the protocol, chances are that the MACs will not match up.
     /// There is a small chance of failure which is `2 / |F|`, where `|F|` is the cardinality of the prime field.
     #[tokio::test]
-    async fn test_complex_circuit() -> Result<(), BoxError> {
+    #[allow(clippy::too_many_lines)]
```
martinthomson, line 339: It's a sign!

src/protocol/malicious.rs
```diff
@@ -356,91 +360,112 @@ pub mod tests {
         let futures =
             context
                 .into_iter()
-                .zip(shared_inputs)
+                .zip(shared_inputs.clone())
                 .map(|(ctx, input_shares)| async move {
```
martinthomson, line 364: I think that I would factor out the body of this circuit.

martinthomson (approved):

src/protocol/context.rs
```diff
@@ -27,7 +28,7 @@ pub struct ProtocolContext<'a, N, F> {
     accumulator: Option<SecurityValidatorAccumulator<F>>,
 }
 
-impl<'a, N, F> ProtocolContext<'a, N, F> {
+impl<'a, N, F: Field> ProtocolContext<'a, N, F> {
```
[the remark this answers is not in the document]
taikiy: I should have commented on #164, but should `ProtocolContext` be generic over `F`? We have one protocol, `ConvertShares`, which takes `XorShares` as its input. It isn't causing any issue now because the context in ConvertShares is only used to identify the helper, but using a context that takes an `F` to execute on shares that are not fields feels odd. It also forces `make_contexts` to return contexts with `Fp31`, but the code that generates the contexts is agnostic about this.

Would it be possible to pass `SecurityValidatorAccumulator` and specify an `F` when calling `MaliciouslySecureMul::execute()`?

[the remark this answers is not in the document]
taikiy: I created #180, which removes `F` from `ProtocolContext`, just to illustrate as an example. If we actually need `F`, please disregard. Otherwise, I'll be happy to rebase on top of this once merged, make more edits if necessary, and push again.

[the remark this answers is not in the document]
benjaminsavage: As discussed on #180, I think we will need <F> for now, since we will not have access to the `SecurityValidatorAccumulator` within all the protocols which invoke multiplications. As such, it's easiest to just get it from the ProtocolContext. And the ProtocolContext will be validating entire protocols over a single field (one that is used for information-theoretic MACs).

src/protocol/context.rs
benjaminsavage: It's fine. I'll leave it like this for now.

src/protocol/malicious.rs
benjaminsavage: OK! I can replace everything except the `BoxError` part with `_`!

PR #170, opened 2022-10-31
Fp2 field for boolean circuits

taikiy:
This is a stack diff on top of #157. 7f42b8a is the new diff introduced by this PR.

## Overview
This diff adds the new struct `Fp2` which represents a binary field $Z_2$. `Fp2` is compatible with the existing field `Fp31` as it implements `Field`, and it automatically works with `Replicated`. `Fp2` also implements the new trait `BinaryField`, which is derived from `Field` and defines bit operations `&`, `|`, `^` and `!`. Protocols that input/output bits can either specify a generic type (i.e. `Replicated<B: BinaryField>`) or the concrete implementation (i.e. `execute(input: Fp2)`).

## Note
Implementations of the field arithmetic operations (e.g. `Add`, `Mul`, etc.) are exactly the same for `Fp2` and `Fp31`, hence I created a macro that generates `Field` implementations. One thing that has changed from the original code was upcasting to `u128` instead of `u16`. Since `Fp31`'s (and `Fp2`) internal value uses `u8`, upcasting to `u16` was sufficient. `u128` will cover all future implementations of fields as long as the prime is kept at most 64 bits, which should be long enough for the real-world use case.

### DIFF
--- a/src/ff/prime_field.rs
+++ b/src/ff/prime_field.rs
@@ -1,16 +1,10 @@
 use super::{field::BinaryField, Field};
 use serde::{Deserialize, Serialize};
-use std::{
-    fmt::Debug,
-    ops::{
-        Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Mul,
-        MulAssign, Neg, Not, Sub, SubAssign,
-    },
-};
+use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
 
 macro_rules! field_impl {
-    ( $( $field:ty ),* ) => { $(
-        impl Add for $field {
+    ( $field:ty ) => {
+        impl std::ops::Add for $field {
             type Output = Self;
 
             fn add(self, rhs: Self) -> Self::Output {
@@ -18,14 +12,14 @@ macro_rules! field_impl {
             }
         }
 
-        impl AddAssign for $field {
+        impl std::ops::AddAssign for $field {
             #[allow(clippy::assign_op_pattern)]
             fn add_assign(&mut self, rhs: Self) {
                 *self = *self + rhs;
             }
         }
 
-        impl Neg for $field {
+        impl std::ops::Neg for $field {
             type Output = Self;
 
             fn neg(self) -> Self::Output {
@@ -33,7 +27,7 @@ macro_rules! field_impl {
             }
         }
 
-        impl Sub for $field {
+        impl std::ops::Sub for $field {
             type Output = Self;
 
             fn sub(self, rhs: Self) -> Self::Output {
@@ -46,14 +40,14 @@ macro_rules! field_impl {
             }
         }
 
-        impl SubAssign for $field {
+        impl std::ops::SubAssign for $field {
             #[allow(clippy::assign_op_pattern)]
             fn sub_assign(&mut self, rhs: Self) {
                 *self = *self - rhs;
             }
         }
 
-        impl Mul for $field {
+        impl std::ops::Mul for $field {
             type Output = Self;
 
             fn mul(self, rhs: Self) -> Self::Output {
@@ -64,7 +58,7 @@ macro_rules! field_impl {
             }
         }
 
-        impl MulAssign for $field {
+        impl std::ops::MulAssign for $field {
             #[allow(clippy::assign_op_pattern)]
             fn mul_assign(&mut self, rhs: Self) {
                 *self = *self * rhs;
@@ -90,17 +84,19 @@ macro_rules! field_impl {
             }
         }
 
-        impl Debug for $field {
+        impl std::fmt::Debug for $field {
             fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                 write!(f, "{}_mod{}", self.0, Self::PRIME)
             }
         }
-    )* };
+    };
 }
 
 #[derive(Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct Fp2(<Self as Field>::Integer);
+
+field_impl! { Fp2 }
 
 impl Field for Fp2 {
     type Integer = u8;
@@ -115,7 +111,7 @@ impl BitAnd for Fp2 {
     type Output = Self;
 
     fn bitand(self, rhs: Self) -> Self::Output {
-        Self((self.0 & rhs.0) & 1)
+        Self(self.0 & rhs.0)
     }
 }
 
@@ -129,7 +125,7 @@ impl BitOr for Fp2 {
     type Output = Self;
 
     fn bitor(self, rhs: Self) -> Self::Output {
-        Self((self.0 | rhs.0) & 1)
+        Self(self.0 | rhs.0)
     }
 }
 
@@ -143,7 +139,7 @@ impl BitXor for Fp2 {
     type Output = Self;
 
     fn bitxor(self, rhs: Self) -> Self::Output {
-        Self((self.0 ^ rhs.0) & 1)
+        Self(self.0 ^ rhs.0)
     }
 }
 
@@ -157,7 +153,10 @@ impl Not for Fp2 {
     type Output = Self;
 
     fn not(self) -> Self::Output {
-        Self((self.0 + 1) & 1)
+        // Using `::from()` makes sure that the internal value is always 0 or 1, but since
+        // we use `u8` to represent a binary value, `!0` and `!1` will result in 255 and
+        // 254 respectively. Add `& 1` at the end to mask the LSB.
+        Self(!self.0 & 1)
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn fp2_binary_op() {
+        let zero = Fp2::ZERO;
+        let one = Fp2::ONE;
+
+        assert_eq!(one, one & one);
+        assert_eq!(zero, zero & one);
+        assert_eq!(zero, one & zero);
+        assert_eq!(zero, zero & zero);
+        assert_eq!(zero, Fp2(31) & Fp2(32));
+        assert_eq!(one, Fp2(31) & Fp2(63));
```
benjaminsavage, line 206: That's interesting. Is this the behavior we want? Do we want it to assert that the value is either 0 or 1 in the constructor? I don't see a ::new() method at all. Does that mean that if you do not do any operation (like this `&` operation here) that it would remain as 31?

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn zero() {
+        macro_rules! gen_zero_test {
+            ( $( $field:ident ),* ) => { $(
+                assert_eq!(
+                    $field(0),
+                    $field::from(<$field as Field>::PRIME),
+                    "from takes a modulus",
+                );
+                assert_eq!($field(0), $field(0) + $field(0));
+                assert_eq!($field(0), $field(0) + $field(0));
```
benjaminsavage, line 252: These lines are the same. Did you intend for one to be addition and one subtraction?

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -87,21 +85,21 @@ impl ConvertShares {
         let futures = bits
             .into_iter()
             .map(|(ctx, b0, b1, input_xor_r)| async move {
-                let r_binary = ReplicatedBinary::new(b0, b1);
+                let r_binary = Replicated::new(Fp2::from(b0), Fp2::from(b1));
```
benjaminsavage, line 88: Very cool. I like this.

src/protocol/reveal_additive_binary.rs
```diff
@@ -30,23 +22,15 @@ impl RevealAdditiveBinary {
...
+    ) -> Result<Fp2, BoxError> {
         let channel = ctx.mesh();
 
         // Send share to helper to the left
-        let future_left = channel.send(
-            ctx.role().peer(Direction::Left),
-            record_id,
-            RevealValue { share: input },
-        );
+        let future_left = channel.send(ctx.role().peer(Direction::Left), record_id, input);
```
benjaminsavage, line 30: That's much nicer, thank you!

benjaminsavage (approved):
I love this. Great stuff.

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+    const ONE: Self = Fp2(1);
+}
+
+impl BinaryField for Fp2 {}
+
+impl BitAnd for Fp2 {
+    type Output = Self;
+
+    fn bitand(self, rhs: Self) -> Self::Output {
+        Self((self.0 & rhs.0) & 1)
```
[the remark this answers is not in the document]
taikiy: This is because Fp2 is backed up by `u8` and implements `Field` with all arithmetic operations like Add, Mul. `& 1` is equivalent to `% Fp2::PRIME` and I thought `&` might perform slightly better. I'm not sure.

An example where this would be a problem is `Fp2(2) | Fp2(1)`, which will output Fp2(3) without `& 1` and fails this equality check: `assert_eq!(Fp2::ONE, Fp2(2) | Fp2(1))`.

src/ff/prime_field.rs
taikiy: Yes, that is correct. Since Fp2 implements the `Field` trait, internal value is `u8` and it supports `From<u128>`. Implementing `::new()` doesn't help much because one could still use `Fp2::from()`. That said, this comes in handy when a boolean circuit protocol executes e.g. Sum on $Z_2$. We can do something like `Fp2::from(b_1 + ... + b_n)`.

benjaminsavage: Sure, but shouldn’t the Fp2::from() implement some kind of range checking and/or casting?

src/ff/prime_field.rs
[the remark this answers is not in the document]
benjaminsavage: This is another symptom of the same root issue of not doing range checking in the Fp2::from method

src/ff/prime_field.rs
taikiy: `From<u128>` implementation is inside the macro and common across `Field` (lines 86-91). We could remove it from the macro and force each `Fp` structs to do custom implementations.

I think it makes sense to not do range checking, though. We want to accept any range of numbers to `FpX::from()` so that we don't have to worry about the input going out of the range, but internally modulo takes care of overflows. It may turn out that we always want to force p or 1 for `Fp2`, and if so, we can do custom `From<u128>` implementations later?

src/ff/prime_field.rs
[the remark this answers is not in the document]
martinthomson: Fp31 clamps in it’s From impl

src/ff/prime_field.rs
benjaminsavage: As @martinthomson points out, "Fp31 clamps in it’s From impl". Indeed, on lines 86-91 there is a "` mod self::PRIME`. That should work just fine as a clamping mechanism. I think that's a perfectly reasonable way to ensure that the internally stored byte is always either a 0 or a 1. Let's do that, so that we do not need a `& 1` after every operation.

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+    fmt::Debug,
+    ops::{
+        Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Mul,
+        MulAssign, Neg, Not, Sub, SubAssign,
+    },
+};
+
+macro_rules! field_impl {
+    ( $( $field:ty ),* ) => { $(
+        impl Add for $field {
```
martinthomson, line 13: To make this macro usable more generally (that is, without having to add `use` statements for all of the types the macro refers to), you should make this:

```suggestion
        impl std::ops::Add for $field {
```
Basically, use a fully-qualified name for everything where possible.

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn bitxor_assign(&mut self, rhs: Self) {
+        *self = *self ^ rhs;
+    }
+}
+
+impl Not for Fp2 {
+    type Output = Self;
+
+    fn not(self) -> Self::Output {
+        Self((self.0 + 1) & 1)
```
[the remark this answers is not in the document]
martinthomson: Or `!self.0 & 1`.

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+use std::{
+    fmt::Debug,
+    ops::{
+        Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Mul,
+        MulAssign, Neg, Not, Sub, SubAssign,
+    },
+};
+
+macro_rules! field_impl {
+    ( $( $field:ty ),* ) => { $(
```
martinthomson, line 12: I would not make this a repeated thing.  We're not implementing a bunch of fields, so repeating this for each type is fine.

src/ff/prime_field.rs
martinthomson: To that end, I would not use `Fp2(x)` in tests as that creates an instance of the object that violates the invariant that Fp2 carries only 0 or 1.  I know that this is intended to break that deliberately, but that's not really a fair test.

src/ff/prime_field.rs
```diff
@@ -0,0 +1,307 @@
...
+        assert_eq!(Fp31(1), x - y);
+
+        let mut x = Fp31(1);
+        x += Fp31(2);
+        assert_eq!(Fp31(3), x);
+    }
+
+    #[test]
+    fn zero() {
+        macro_rules! gen_zero_test {
```
martinthomson, line 244: Don't use a macro for this, use a generic function.  And invoke that function from two `#[test]` functions.

martinthomson (commented):
Yeah, this is good stuff.

src/ff/prime_field.rs
benjaminsavage: I agree with @martinthomson

benjaminsavage (changes requested):
Lots of good suggestions here. Marking as "request changes" just to move it back to your queue @taikiy

src/ff/prime_field.rs
taikiy: Oh, I missed that clamp part. Let me remove `& 1` and use `::from()` everywhere.

benjaminsavage (approved):
Looks good!

PR #171, opened 2022-10-31
Removing dead code

benjaminsavage:
#159 made all of this dead code - but somehow it was not detected as such. This PR just removes it all.


richajaindce (approved):
Makes sense to remove these!

PR #175, opened 2022-10-31
Remove Shamir Secret Sharing

benjaminsavage:
I'm sorry @akoshelev. It's a wonderful implementation of Shamir Secret Sharing. But as this code is not currently in use, I feel like we should remove it for now. In the future, if we have need for it, we can always come and retrieve it from version control.


martinthomson:
@benjaminsavage did you mean to target the main branch here?

benjaminsavage:
What the hell is even this PR? It looks like a duplicate... sorry... closing.

PR #176, opened 2022-10-31
Update CI checks

martinthomson:
Reverts private-attribution/ipa#161 (which in turn reverted this as an original change).

This simplifies the dependencies we have on actions for rust.  Now that I've worked out how to enable the action, we can see how this works.


martinthomson:
Note to self, once this lands, drop actions-rs from [the organization settings](https://github.com/organizations/private-attribution/settings/actions)

PR #177, opened 2022-10-31
Three changes

martinthomson:
1. remove network from gateway, which reduces the complexity of the context declaration
2. add RecordId::from(usize), which complicates usage in some places, but simplifies others
3. fix the bit permutation code

I'll split this up later if necessary, but I need to attend to other things for now.

### DIFF
--- a/src/protocol/securemul.rs
+++ b/src/protocol/securemul.rs
@@ -1,6 +1,6 @@
 use crate::error::BoxError;
 use crate::ff::Field;
-use crate::helpers::{fabric::Network, Direction};
+use crate::helpers::Direction;
 use crate::protocol::{context::ProtocolContext, RecordId};
 use crate::secret_sharing::Replicated;
 use std::fmt::Debug;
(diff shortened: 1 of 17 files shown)


src/protocol/securemul.rs
```diff
@@ -108,11 +103,9 @@ pub mod tests {
     #[allow(clippy::cast_possible_truncation)]
     pub async fn concurrent_mul() {
         type MulArgs<F> = (Replicated<F>, Replicated<F>);
-        async fn mul<F: Field>(
-            v: (ProtocolContext<'_, Arc<InMemoryEndpoint>, F>, MulArgs<F>),
-        ) -> Replicated<F> {
+        async fn mul<F: Field>(v: (ProtocolContext<'_, F>, MulArgs<F>)) -> Replicated<F> {
```
benjaminsavage, line 106: thank you for removing the `Arc<InMemoryEndpoint>`... it scares me to see that every time =).

benjaminsavage (approved):
This is a really great PR. I love that you were able to remove the `N` from everywhere. Really simplifies things. And I like the way you resolved that issue with the GenBitPerm. I'm definitely going to apply some of those ideas in other files.

PR #179, opened 2022-10-31
Compose

richajaindce:
This protocol composes two permutations by applying the first secret-shared permutation(sigma) to another secret-shared permutation(rho)


src/protocol/reveal.rs
```diff
@@ -36,6 +40,27 @@ pub async fn reveal<F: Field>(
...
+    let reveals =
+        zip(repeat(ctx), permutations)
+            .enumerate()
+            .map(|(index, (ctx, input))| async move {
+                reveal(ctx, RecordId::from(index as u32), *input).await
+            });
+    let permutation = try_join_all(reveals).await?;
+    let mut perms = Vec::new();
+    for i in permutation {
+        perms.push(i.as_u128() as usize);
```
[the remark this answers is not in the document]
richajaindce: It works.

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+    sigma: &'a mut Vec<Replicated<F>>,
+    rho: &'a mut Vec<Replicated<F>>,
+}
+
+impl<'a, F: Field> Compose<'a, F> {
+    #[allow(dead_code)]
+    pub fn new(sigma: &'a mut Vec<Replicated<F>>, rho: &'a mut Vec<Replicated<F>>) -> Self {
+        Self { sigma, rho }
+    }
+    #[embed_doc_image("compose", "images/sort/compose.png")]
```
benjaminsavage, line 33: I have a suggestion for a different diagram that I think would be more clear: https://docs.google.com/drawings/d/143VUS1-vs2j2SMwKluPMdQk7xco6yx-8JKEsWi36h9Q/edit?usp=sharing

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// This algorithm composes two permutations (`rho` and `sigma`). Both permutations are secret-shared,
+    /// and none of the helpers should learn it through this protocol.
+    /// Steps
+    /// ![Compose steps][compose]
+    /// 1. Generate random permutations using prss
+    /// 2. First permutation (sigma) is shuffled with random permutations
+    /// 3. Reveal the permutation
+    /// 4. Revealed permutation is applied locally on another permutation shares (rho)
+    /// 5. Unshuffle the permutation with random permutations to negate the effect of shuffling
```
benjaminsavage, line 42: /// Unshuffle the permutation with the same random permutations used in step 2, to undo the effect of the shuffling

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub async fn execute(&mut self, ctx: ProtocolContext<'_, F>) -> Result<(), BoxError> {
+        let mut random_permutations =
+            get_two_of_three_random_permutations(self.rho.len(), &ctx.prss());
+        let mut random_permutations_copy = random_permutations.clone();
+
+        let mut shuffle_sigma = Shuffle::new(self.sigma);
+        shuffle_sigma
+            .execute(ctx.narrow(&ShuffleSigma), &mut random_permutations)
+            .await?;
+
```
benjaminsavage, line 53: Let's just inline it:
```
let _ = Shuffle::new(self.sigma).execute(ctx.narrow(&ShuffleSigma), &mut random_permutations).await?;
```

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut perms = reveal_a_permutation(ctx.narrow(&RevealPermutation), self.sigma).await?;
+
+        apply_inv(&mut perms, &mut self.rho);
+
+        let mut unshuffle_rho = Shuffle::new(self.rho);
+        unshuffle_rho
+            .execute_unshuffle(ctx.narrow(&UnshuffleRho), &mut random_permutations_copy)
+            .await?;
```
benjaminsavage, line 61:
```
let _ = Shuffle::new(self.rho).execute_unshuffle(ctx.narrow(&UnshuffleRho), &mut random_permutations_copy).await?;
```

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+            generate_shares, make_contexts, make_world, validate_result_from_shares, TestWorld,
+        },
+    };
+
+    #[tokio::test]
+    pub async fn compose() {
+        const BATCHSIZE: usize = 25;
+        for _ in 0..10 {
+            let mut rng_sigma = rand::thread_rng();
+            let mut rng_rho = rand::thread_rng();
```
benjaminsavage, line 88: Why do you need two RNGs?

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+            let sigma_u128: Vec<u128> = sigma.iter().map(|x| *x as u128).collect();
+
+            let mut rho: Vec<usize> = (0..BATCHSIZE).collect();
+            rho.shuffle(&mut rng_rho);
+            let rho_u128: Vec<u128> = rho.iter().map(|x| *x as u128).collect();
+
+            let mut rho_composed = rho_u128.clone();
+
+            let mut sigma_copy = Permutation::oneline(sigma.clone());
+            apply_inv(&mut sigma_copy, &mut rho_composed);
```
benjaminsavage, line 102: I think the less locals the better, because all the names just get confusing.
If it'll let you do this, I'd rather something like this:
```
apply_inv(
    &mut Permutation::oneline(sigma.clone()), 
    rho_u128.clone(),
);
```

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+
+            let mut compose0 = Compose::new(&mut sigma_shares.0, &mut rho_shares.0);
+            let mut compose1 = Compose::new(&mut sigma_shares.1, &mut rho_shares.1);
+            let mut compose2 = Compose::new(&mut sigma_shares.2, &mut rho_shares.2);
+
+            let h0_future = compose0.execute(ctx0);
+            let h1_future = compose1.execute(ctx1);
+            let h2_future = compose2.execute(ctx2);
+
+            try_join!(h0_future, h1_future, h2_future).unwrap();
```
benjaminsavage, line 117: If you make the test return a `Result<(), BoxError>` you can replace the `.unwrap()` with `?`

benjaminsavage (approved):
This looks really good!

src/protocol/sort/compose.rs
richajaindce: I was using two rngs so that the two obtained permutations are totally independent of each other.

src/protocol/sort/compose.rs
richajaindce: I still need the rho_composed to validate the results. So keeping that

src/protocol/sort/compose.rs
richajaindce: Nice! Hadnt done this before

src/protocol/sort/compose.rs
richajaindce: Thanks for the suggestion, using that instead. However, the template of this is different from other images. Once done with sort, will try to revisit all the diagrams to ensure correctness and consistent template

src/protocol/sort/compose.rs
```diff
@@ -0,0 +1,127 @@
...
+    ComposeStep::{RevealPermutation, ShuffleSigma, UnshuffleRho},
+};
+
+/// This is an implementation of Compose (Algorithm 5) found in the paper:
+/// "An Efficient Secure Three-Party Sorting Protocol with an Honest Majority"
+/// by K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, N. Kiribuchi, and B. Pinkas
+/// <https://eprint.iacr.org/2019/695.pdf>
+/// This protocol composes two permutations by applying one secret-shared permutation(sigma) to another secret-shared permutation(rho)
+/// Input: First permutation(sigma) i.e. permutation that sorts all i-1th bits and other permutation(rho) i.e. sort permutation for ith bit
+/// Output: All helpers receive secret shares of permutation which sort inputs until ith bits.
```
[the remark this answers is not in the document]
richajaindce: apply_inv(rho, apply_inv(sigma, x)) == apply(sigma, rho) ! 
Here, apply(sigma, rho) is called compose in the paper.

Havent proved it with math but this works seamlessly. Adding a testcase for verification also.

PR #180, opened 2022-10-31
Remove `F` from ProtocolContext

taikiy:
I created this PR as a recommendation regarding my comment on #169:

> I should have commented on #164, but should `ProtocolContext` be generic over `F`? We have one protocol, `ConvertShares`, which takes `XorShares` as its input. It isn't causing any issue now because the context in ConvertShares is only used to identify the helper, but using a context that takes an `F` to execute on shares that are not fields feels odd. It also forces `make_contexts` to return contexts with `Fp31`, but the code that generates the contexts is agnostic about this.
> 
> Would it be possible to pass `SecurityValidatorAccumulator` and specify an `F` when calling `MaliciouslySecureMul::execute()`?

In this PR, I moved `SecurityValidatorAccumulator` as a `malicous_multiply()` parameter. But once we have `MaliciouslySecureMul`, I think it makes more sense for `execute` to be generic over `F` and inside it creates an instance of `SecurityValidatorAccumulator`.

### DIFF
--- a/src/protocol/attribution/accumulate_credit.rs
+++ b/src/protocol/attribution/accumulate_credit.rs
@@ -52,7 +52,7 @@ impl<'a, F: Field> AccumulateCredit<'a, F> {
     #[allow(dead_code)]
     pub async fn execute(
         &self,
-        ctx: ProtocolContext<'_, F>,
+        ctx: ProtocolContext<'_>,
     ) -> Result<Batch<AccumulateCreditOutputRow<F>>, BoxError> {
         #[allow(clippy::cast_possible_truncation)]
         let num_rows = self.input.len() as RecordIndex;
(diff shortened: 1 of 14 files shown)


martinthomson:
I share Ben's view on this.  I think that we'll want to use a single configuration for most of the work that the protocol does.  Though we might use Fp2 in some places, those are not generic in the sense that they can be switched out for another field.  But the overall protocol will need something here.

If things get more complex (and we need more security for the malicious checks, say), then we can indirect through a type that captures the full set of field types that are used, but let's burn that bridge when we come to it.

taikiy:
OK, thanks for the comments! I'm going to close this PR.
Meanwhile, I created #184 to make `Fp2` protocols work with the current state.

Issue #135, opened 2022-10-18
New step model rejects attempts to use the same step on multiple records


akoshelev:
After we landed #122 , our previous model of interaction between steps no longer works. We thought that using finite number of steps that define the circuit and executing this circuit for many records is optimal, but debug assertions added in the mentioned PR make it no longer possible.

Demo program to demonstrate the issue

```rust
#[cfg(test)]
mod toy_demo {
    use crate::field::{Field, Fp31};
    use crate::helpers::fabric::Network;
    use crate::protocol::context::ProtocolContext;
    use crate::protocol::{QueryId, RecordId, Step};
    use crate::secret_sharing::Replicated;
    use futures::StreamExt;
    use futures_util::stream::FuturesOrdered;
    use rand::thread_rng;
    use crate::test_fixture::{make_contexts, make_world, share};


    /// Takes a collection as input and executes `MultiplyTwice` protocol on every input row.
    #[derive(Debug)]
    struct ToyProtocol<'a, F> {
        input: &'a [(Replicated<F>, Replicated<F>)]
    }

    /// This protocol runs per row and just multiplies a*b*a
    struct MultiplyTwice<F> {
        a: Replicated<F>,
        b: Replicated<F>,
        record_id: RecordId
    }

    enum ToyStep {
        Step1,
        Step2
    }

    impl <'a, F: Field> ToyProtocol<'a, F> {
        async fn execute<N: Network>(self, ctx: &ProtocolContext<'a, N>) -> Vec<Replicated<F>> {

            let mut futures = FuturesOrdered::new();

            // for every input row, execute `MultiplyTwice`
            for (record_id, r) in self.input.iter().enumerate() {
                let record_id = RecordId::from(record_id as u32);
                let sub_protocol = MultiplyTwice { a: r.0, b: r.1, record_id };
                futures.push_back(sub_protocol.execute(&ctx));
            }

            let res = futures.collect::<Vec<_>>().await;
            res
        }
    }

    impl <F: Field> MultiplyTwice<F> {
        async fn execute<N: Network>(self, ctx: &ProtocolContext<'_, N>) -> Replicated<F> {
            // error occurs here because we narrow the same context more than once (once per record)
            let ctx = ctx.narrow(&ToyStep::Step1);
            let r = ctx.multiply(self.record_id).await.execute(self.a, self.b).await.unwrap();

            let ctx = ctx.narrow(&ToyStep::Step2);
[...]

PR #183, opened 2022-10-31
Range of accepted records inside the send buffer

akoshelev:
## Overview

This change continues work introduced in #150 to provide strict ordering for messages sent between helpers. As fields are serializable (#159) to a slice, storing them inside a contiguous buffer is now possible. `FixedSizeByteVec` is the base structure for it that keeps track of elements inserted and indicates when it is ready to be drained.

`SendBuffer` uses this structure under the hood to provide ordering and decide whether the range is valid or not by simply counting the number of elements drained from that buffer (that defines the lower bound) and total capacity of the buffer (upper bound). If record is outside of that range it is rejected (for now).

Things that are still missing and will be dealt with if this is a step in the right direction

* Use tinyvec or smallvec for message payload
* Use `Vec<u8>` inside and between network and infra layers
* Don't reject records greater than the maximum acceptable record id, block the caller until the window is moved instead.


akoshelev:
@martinthomson would you able to take a look? The next change will be the final one and will bring the tiny vec and raw bytes. That should conclude infra changes until we move to streaming API

src/helpers/buffers/fsv.rs
```diff
@@ -0,0 +1,220 @@
...
+            added: bitvec![0; region_size * region_count],
+            elements_drained: 0,
+            region_size,
+        }
+    }
+
+    /// Inserts a new element to the specified position, returning the previous element at this `index`.
+    /// ## Panics
+    /// * Panics if `index` is out of bounds.
+    pub fn insert(&mut self, index: usize, elem: [u8; N]) -> Option<[u8; N]> {
```
martinthomson, line 50: Why are you returning what was present at that index?

Also, it seems like you are taking an index into your structure, rather than a record number.  This code should only track a window starting from the index just past the last data that was sent, but accept a RecordId as the index.  Then you get something like:

```rust
fn insert(&mut self, i: RecordId, data: &[u8; N]) {
  let offset = usize::from(i - self.sent);
  assert!(!self.added[offset]);
  self.data[offset * N..offset*(N+1)].copy_from_slice(data);
  self.added[offset] = true;
}
```

src/helpers/buffers/fsv.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// elements have been inserted, returns them in order. `self.ready()` should indicate that.
+    ///
+    /// ## Panic
+    /// Panics if `ready()` is false before calling `drain()`.
+    pub fn drain(&mut self) -> Vec<u8> {
+        let all_ones = self.added.drain(..self.region_size).all(|bit| bit);
+        assert!(all_ones, "drain() called when ready() was false");
+
+        let r = self.data.drain(..self.region_size * N).collect();
+        self.elements_drained += self.region_size;
```
martinthomson, line 85: split_off() is your friend.

src/helpers/buffers/fsv.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub fn ready(&self) -> bool {
+        self.added[..self.region_size].all()
+    }
+
+    /// Drains the elements from the beginning of the vector. If all of the first `self.region_size`
+    /// elements have been inserted, returns them in order. `self.ready()` should indicate that.
+    ///
+    /// ## Panic
+    /// Panics if `ready()` is false before calling `drain()`.
+    pub fn drain(&mut self) -> Vec<u8> {
```
martinthomson, line 80: Maybe you could write this as `take(&mut self) -> Option<Vec<u8>>` and avoid separate ready and drain calls at the caller side.  Then you don't need to worry about assertions in this one.

src/helpers/buffers/fsv.rs
```diff
@@ -0,0 +1,220 @@
...
+        let all_ones = self.added.drain(..self.region_size).all(|bit| bit);
+        assert!(all_ones, "drain() called when ready() was false");
```
martinthomson, line 82: See above.

src/helpers/buffers/fsv.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl<const N: usize> FixedSizeByteVec<N> {
+    pub const ELEMENT_SIZE_BYTES: usize = N;
+
+    pub fn new(region_count: usize, region_size: usize) -> Self {
+        Self {
+            data: vec![0_u8; N * region_size * region_count],
+            added: bitvec![0; region_size * region_count],
+            elements_drained: 0,
+            region_size,
```
martinthomson, line 43: It might be neater in some ways if you allow the caller to take an arbitrary number of elements; then the drain/take could accept an argument that says how many records you want to return.  Nothing in this design requires that this object remember how many records are returned in each pull.  That means that - if you want to force progress, you can take a small slice - and on every pull you might pull more if more is available.

src/helpers/buffers/send.rs
```diff
@@ -1,61 +1,314 @@
...
             inner: HashMap::default(),
         }
     }
 
+    /// TODO: change the output to Vec<u8> - we no longer need a wrapper. The raw byte vector
+    /// will be communicated down to the network layer.
+    #[allow(clippy::needless_pass_by_value)] // will be fixed when tiny/smallvec is used
     pub fn push(
         &mut self,
         channel_id: ChannelId,
```
martinthomson, line 76: I think that this should take `Borrow<ChannelId>`, because you don't need to create a clone unless you are adding a new item to the map.

martinthomson (commented):
Gotta do a talk in a few minutes and I have no idea when I'll next get time, so a partial review.

src/helpers/buffers/fsv.rs
akoshelev: that is used to check that nobody inserts a record with the same record id. Taking an index rather than `RecordId`  helps to keep this struct generic and not tied up to the send buffer implementation.

src/helpers/buffers/fsv.rs
akoshelev: true, but I wonder what would be the use case for it? For now, setting send buffer capacity to `N*<drain_threshold>` as you suggested before worked perfectly to maintain the concurrency and keeping unit tests fast (when drain threshold is set to 1).

src/protocol/reveal.rs
```diff
@@ -63,7 +63,7 @@ mod tests {
             let input = Fp31::from(secret);
             let share = share(input, &mut rand);
 
-            let record_id = RecordId::from(i);
+            let record_id = RecordId::from(0_u32);
```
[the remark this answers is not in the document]
akoshelev: the good thing is that tracing helps a lot to catch these issues. Of course you probably need to be aware how infra works and sends data

src/protocol/securemul.rs
```diff
@@ -81,16 +81,36 @@ pub mod tests {
...
-        assert_eq!(30, multiply_sync(&context, "1", 6, 5, &mut rand).await?);
-        assert_eq!(25, multiply_sync(&context, "2", 5, 5, &mut rand).await?);
-        assert_eq!(7, multiply_sync(&context, "3", 7, 1, &mut rand).await?);
-        assert_eq!(0, multiply_sync(&context, "4", 0, 14, &mut rand).await?);
-        assert_eq!(8, multiply_sync(&context, "5", 7, 10, &mut rand).await?);
-        assert_eq!(4, multiply_sync(&context, "6", 5, 7, &mut rand).await?);
-        assert_eq!(1, multiply_sync(&context, "7", 16, 2, &mut rand).await?);
+        assert_eq!(
+            30,
+            multiply_sync(make_contexts(&world), 6, 5, &mut rand).await?
```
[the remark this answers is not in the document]
akoshelev: just lazy me, changed it back to clone()

src/protocol/malicious.rs
```diff
@@ -385,10 +382,9 @@ pub mod tests {
                                 .iter()
                                 .zip(input_shares.iter().skip(1))
                                 .zip(row_narrowed_contexts.iter())
-                                .enumerate()
-                                .map(|(i, ((a, b), ctx))| async move {
+                                .map(|((a, b), ctx)| async move {
                                     ctx.narrow("SingleMult")
-                                        .multiply(RecordId::from(u32::try_from(i).unwrap()))
+                                        .multiply(RecordId::from(0_u32))
```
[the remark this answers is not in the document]
akoshelev: so there is a trick that folks use to get things moving - they narrow the step to record (row_narrowed_context does that). In this case you must start records from 0 and this change reveals that issue.

we should fix #135 asap imo

src/helpers/buffers/fsv.rs
martinthomson: Does it have to involve a copy?  It probably does, but it doesn't have to.

src/helpers/buffers/fsv.rs
martinthomson: You had some code in some place that tried to unblock stuff elsewhere by pulling in something less than a full slice.  Either way, it seems like this might be more general (and this has less state as a result).

src/helpers/buffers/fsv.rs
martinthomson: I don't care about the argument type (use `impl Into<usize>` if you prefer).  I was more concerned about this returning what was already in place, when the caller should never be putting stuff in at the same index.

src/helpers/buffers/fsv.rs
akoshelev: `split_off` does the opposite to what I need. It leaves [0, mid) elements inside the vec and returns [mid, len). Is there a way to avoid a copy?

src/helpers/buffers/fsv.rs
akoshelev: yea I think that code is not compatible anymore with contiguous buffers. I removed it entirely

src/helpers/buffers/fsv.rs
akoshelev: I was basically mimicking HashMap behavior here https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert. For generic structs it seems ok to accept values put at the same index more than once. SendBuffer handles that already so I am curious what is the reason of having this check here?

> The comment suggests that the only time Some() is returned instead of None is in the case of an error. One assumes this is to help assist with debugging? If that's the case - should we return an error with this payload instead?

I think that is basically the question here. Do we consider inserting something to this data structure at the same index more than once as an error? I know for our existing use case (send buffers) it is an error but I validate it at the callsite. My assumption is that this is a generic data structure and callers need to decide what to do. 

It is easy to go either way though. I just want to make sure it is clear what I meant here and I am totally OK changing it if this argument is not strong enough

benjaminsavage:
This looks pretty good to me. This PR has been in review a long time - and I'd like to get it landed to unblock your next few diffs (and also help you avoid needing to continue rebasing!).

@martinthomson - when you are back from London - let's do a review of this whole part of the infra together and see if there's any additional changes we should make to it.

benjaminsavage (approved):

PR #184, opened 2022-10-31
`make_contexts` returns a `ProtocolContext<'_, F>`

taikiy:
Per discussion in #180, `ProtocolContext` is generic over an `F`. To support protocols that take `BinaryField` (i.e. `Fp2`) as inputs, change `make_contexts` to also become generic over `F` instead of returning a concrete implementation.

### DIFF
--- a/src/protocol/attribution/accumulate_credit.rs
+++ b/src/protocol/attribution/accumulate_credit.rs
@@ -265,7 +265,7 @@ mod tests {
     #[tokio::test]
     pub async fn accumulate() {
         let world = make_world(QueryId);
-        let context = make_contexts(&world);
+        let context = make_contexts::<Fp31>(&world);
         let mut rng = StepRng::new(100, 1);
 
         let raw_input: [[u128; 4]; 9] = [
--- a/src/protocol/check_zero.rs
+++ b/src/protocol/check_zero.rs
@@ -91,7 +91,7 @@ pub mod tests {
         logging::setup();
 
         let world: TestWorld = make_world(QueryId);
-        let context = make_contexts(&world);
+        let context = make_contexts::<Fp31>(&world);
         let mut rng = rand::thread_rng();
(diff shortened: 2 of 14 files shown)


benjaminsavage (approved):
Works for me.

martinthomson:
I am not a fan of this. We made the field a generic param because we wanted to switch it out, but we don’t need that for Fp2.

those places we use Fp2 know that that field is needed and don’t need to also support FpWhatever.

benjaminsavage:
I'm not strongly opinionated on this one. Would you rather revert it @martinthomson ?

taikiy:
When I try to call `ctx.multiply().await.execute(a, b)`, where a and b are `Fp2`, the code won’t compile because the context now expects `Fp31`, which is the concrete type `make_contexts` returns. Is there a work around to this? @martinthomson

martinthomson:
Oh, right.  This is probably something we need to take great care with.

Let's look at this hollistically.  In a real protocol execution, we'll primarily be operating in a set field (Fp31 maybe in tests, but obviously something larger in reality).  But, sometimes, we will need to do some operations in Fp2 specifically, or maybe we need to upgrade to a larger field for some reason.  For that, I think that we can (and maybe should) offer a function that can create `Context<...Fp2...>` from a `Context<...Fp31...>`.  For those tests that use Fp2, this is a tiny bit more work, but it should keep things clearer.

One thing I've been thinking of doing is conditionally compiling tests to use a different field implementation so that we can be confident that the different fields are working correctly.

That is, something like this:

```rust
#[cfg(feature = "test-fpm31")]
type TestField = FpM31;
#[cfg(not(all(feature = "test-fpm31")))]
type TestField = Fp31;
```
With all mentions of Fp31 in tests replaced with TestField.

akoshelev:
I love this idea, maybe we can even have an approval step (or several) that runs tests against all field implementations

PR #185, opened 2022-11-01
Custom debug implementation for UniqueStepId

akoshelev:
Derive implementation prints the guts of the Mutex used inside `used`

    ```
    sending 1 message(s) to channel[peer=H2,step=UniqueStepId { id: "protocol/0", used: Mutex { data: {}, poisoned: false, .. } }]
    ```

this change also adds `logging::setup` to the tests that don't have it


thurstonsand (approved):

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -264,6 +265,8 @@ mod tests {
 
     #[tokio::test]
     pub async fn accumulate() {
+        logging::setup();
+
```
martinthomson, line 269: Crazy idea: we could move this to make_world().  I assume that there is no harm in doing it more than once (?)

src/protocol/mod.rs
```diff
@@ -151,6 +152,12 @@ impl AsRef<str> for IpaProtocolStep {
     }
 }
 
+impl Debug for UniqueStepId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "UniqueStepId(id={})", self.id)
```
martinthomson, line 157:
```suggestion
        write!(f, "step={}", self.id)
```

src/protocol/attribution/accumulate_credit.rs
akoshelev: I love this suggestion. yes `logging::setup()` is safe to call in parallel, more than once, etc.

PR #186, opened 2022-11-01
Sort

richajaindce:
Still WIP but opening for discussion


src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub async fn execute<F: Field>(
+        &self,
+        ctx: ProtocolContext<'_, F>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let mut bits = convert_all(ctx.narrow(&ModulusConversion), self.input).await?;
+        let mut sigma = BitPermutations::new(&bits[0])
+            .execute(ctx.narrow(&BitPermutation))
+            .await?;
+        for (i, ith_bit) in bits.iter_mut().enumerate().skip(1) {
+            SecureApplyInv::execute(
```
richajaindce, line 40: @akoshelev @martinthomson needed some help in what context needs to be called here for applyinv, bit permutation and compose. These three protocols will be called for each bit in the match key.

src/protocol/sort/mod.rs
```diff
@@ -4,22 +4,55 @@ use std::fmt::Debug;
...
+    BitPermutation(u8),
+    ApplyInv(u8),
+    ComposeStep(u8),
 }
 
 impl Step for SortStep {}
 
 impl AsRef<str> for SortStep {
     fn as_ref(&self) -> &str {
+        const BITPERMUTATIONS: [&str; 64] = [
```
richajaindce, line 25: This is quite ugly and repetitive. Since records are used in lower layers, I just created these steps per bit. Happy to incorporate a better way of doing this.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +107,35 @@ impl ConvertShares {
     }
 }
 
+pub async fn convert_all_shares<F: Field>(
+    ctx: ProtocolContext<'_, F>,
+    input: &[u64],
+) -> Result<Vec<Vec<Replicated<F>>>, BoxError> {
+    let converted_shares = try_join_all(zip(repeat(ctx), input).enumerate().map(
+        |(record_id, (ctx, row))| async move {
+            let bit_number = format!("record_id{}", record_id);
```
benjaminsavage, line 116: You seem to be confusing "bit number" with "record ID".

In this code `record_id` ranges from 0 to the number of rows you're converting shares for. Assuming one u64 corresponds to one match key, this could range into the millions. Indeed, we will have record IDs going into the millions. But the match key is < 64 bits. The `ConvertShares` protocol you're invoking below on line 118 already iterates over all of the bits and converts them all.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +107,35 @@ impl ConvertShares {
...
+
+            ConvertShares::new(XorShares {
+                num_bits: 40,
```
benjaminsavage, line 119: I don't think you should hard-code in this number. This should probably be a param.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +108,36 @@ impl ConvertShares {
...
 }
 
+pub async fn convert_all_shares<F: Field>(
+    ctx: &ProtocolContext<'_, F>,
+    input: &[u64],
+) -> Result<Vec<Vec<Replicated<F>>>, BoxError> {
+    let converted_shares = try_join_all(zip(repeat(ctx), input).enumerate().map(
+        |(record_id, (ctx, row))| async move {
+            ConvertShares::new(XorShares {
+                num_bits: 64,
```
benjaminsavage, line 118: Let's make this a param - not a hardcoded value. 

For example, if we are only using 40-bit match keys, then we only need to convert 40 bits of this thing.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +108,36 @@ impl ConvertShares {
...
+                ctx.narrow(&ModulusConversion(record_id.try_into().unwrap())),
+                RecordId::from(record_id),
+            )
+            .await
+        },
+    ))
+    .await?;
+
+    //Transpose the output to return bit-wise results
+    let cols = vec![Replicated::new(F::ZERO, F::ZERO); converted_shares.len()];
```
benjaminsavage, line 131: I don't understand what's going on here. If we are sorting 1 million rows, then converted_shares.len() will take the value 1 million. How are there 1 million "columns"? I think this is an error.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +108,36 @@ impl ConvertShares {
...
+    let mut output = vec![cols; converted_shares[0].len()];
```
benjaminsavage, line 132: This length is that number you've specified on for 118, `num_bits`. Let's use the same param for this rather than this approach which will fail if there is no element zero.

BTW, *this* seems like the number of columns - the number of bits.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +108,36 @@ impl ConvertShares {
...
+    for (row_idx, row) in converted_shares.into_iter().enumerate() {
+        for (col_idx, col) in row.into_iter().enumerate() {
+            output[col_idx][row_idx] = col;
+        }
+    }
```
benjaminsavage, line 137: If you'd just like to get a `Vec<Replicated<F>>' representing the nth bit of all the rows, you can do:

```
let nth_bit:Vec<Replicated<F>> = converted_shares.iter().map(|bitwise_shares| {
    bitwise_shares[n]
}).collect();
```

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use super::{compose::Compose, secureapplyinv::SecureApplyInv};
+use embed_doc_image::embed_doc_image;
+
+/// This is an implementation of `GenPerm` (Algorithm 6) described in:
+/// "An Efficient Secure Three-Party Sorting Protocol with an Honest Majority"
+/// by K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, N. Kiribuchi, and B. Pinkas
+/// <https://eprint.iacr.org/2019/695.pdf>.
+#[derive(Debug)]
+pub struct GenerateSortPermutations<'a> {
```
benjaminsavage, line 21: Why’s it plural? It just generates a single permutation does it not?

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// 3i.  Apply inverse of i-1th bit permutation on ith input bits
+    /// 3ii. Obtain bit permutation to sort ith input bit
+    /// 3iii.Compose ith bit permutation on i-1th composition. We have now obtained ith composition
+    /// 4. Return nth composition as the permutation to sort the given inputs
+    pub async fn execute<F: Field>(
+        &self,
+        ctx: ProtocolContext<'_, F>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let mut bits = convert_all_shares(&ctx, self.input).await?;
+        let mut sigma = BitPermutations::new(&bits[0])
```
benjaminsavage, line 48: I see. You wanted the data in this format which is why you did that whole "transpose" thing at the end of the `convert_all_shares` function.

I'd like to avoid all of that copying bytes around. Is there a way that we can just keep things in the original format?

i.e.
```
      | bit n | bit n-1 | bit n-2 | ... | bit 3 | bit 2 | bit 1 |
row 1 |       |         |         | ... |       |       |       | 
row 2 |       |         |         | ... |       |       |       | 
 ...  |       |         |         | ... |       |       |       | 
row m |       |         |         | ... |       |       |       | 
```

You could get the shares of the nth bit by doing:
```
foo.iter().map(|bit_shares| { bit_shares[n] })
```
I feel like there should be some way to make that an "iterator" that just loops over things and doesn't need to deep-copy them. I think if you do not include a "collect()" then it is a lazy iterator?

benjaminsavage: I don't think the variable name "sigma" is adding anything to comprehensibility here. Perhaps a name like "bit_zero_permutation" would be better.

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+            .execute(ctx.narrow(&BitPermutation(0)))
+            .await?;
+        for (i, ith_bit) in bits.iter_mut().enumerate().skip(1) {
```
benjaminsavage, line 51: We can use a more meaningful name than `i` like
```
for (bit_number, shares_of_bit_value) in ...
```

src/protocol/sort/mod.rs
```diff
@@ -4,22 +4,63 @@ use std::fmt::Debug;
...
 mod shuffle;
 
 #[allow(clippy::module_name_repetitions)]
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 pub enum SortStep {
-    BitPermutations,
+    ModulusConversion(u8),
+    BitPermutation(u8),
+    ApplyInv(u8),
+    ComposeStep(u8),
```
benjaminsavage, line 18: I think this is great! It is simultaneously clear that we are doing the same type of step over and over for each bit, but we can generate a unique step ID. A u8 should be sufficient as we will never need to sort 256-bit numbers =).

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+            SecureApplyInv::execute(
+                &ctx.narrow(&ApplyInv(i.try_into().unwrap())),
```
benjaminsavage, line 53: I think this approach to using narrow is *great*. We definitely need a new prss context for each bit, and this way we get that while still being clear about what logical step is going on. I just wish we could eliminate the big messy cast. You can probably do that by making the loop variable a u8.

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+                ith_bit,
+                &mut sigma.clone(),
+            )
+            .await?;
+            let mut rho = BitPermutations::new(ith_bit)
```
benjaminsavage, line 58: As with above, instead of "rho", I think "permutation_for_bit" could be a better name.

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+                .execute(ctx.narrow(&BitPermutation(i.try_into().unwrap())))
+                .await?;
+            Compose::new(&mut sigma, &mut rho)
+                .execute(ctx.narrow(&ComposeStep(i.try_into().unwrap())))
+                .await?;
+            sigma = rho;
```
benjaminsavage, line 64: OK, so if "sigma" is being used to mean: "cumulative permutation for all the bits thus far", then perhaps we could call it something like "composed_permutation".

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+    };
+
+    #[tokio::test]
+    pub async fn generate_sort_permutations() {
+        let world = make_world(QueryId);
+        let [ctx0, ctx1, ctx2] = make_contexts(&world);
+        let mut rng = rand::thread_rng();
+        let mask = (1_u64 << 63) - 1;
+
+        let batchsize = 30;
```
benjaminsavage, line 87: After reading the code, I'm worried that there is some confusion between the number of bits and the number of rows. To ensure that we are definitely not mixing up the two, let's make this number MUCH larger than the number of bits. Let's say 1000.

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+        protocol::{sort::generate_sort_permutations::GenerateSortPermutations, QueryId},
+        test_fixture::{make_contexts, make_world, validate_list_of_shares},
+    };
+
+    #[tokio::test]
+    pub async fn generate_sort_permutations() {
+        let world = make_world(QueryId);
+        let [ctx0, ctx1, ctx2] = make_contexts(&world);
+        let mut rng = rand::thread_rng();
+        let mask = (1_u64 << 63) - 1;
```
benjaminsavage, line 85: What is the purpose of this mask? I think we can easily live without it since the "ConvertShares" just lets you specify how many bits are meaningful and it'll just ignore the rest.

src/protocol/sort/generate_sort_permutations.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let sortperms0 = GenerateSortPermutations::new(&shares[0]);
+        let sortperms1 = GenerateSortPermutations::new(&shares[1]);
+        let sortperms2 = GenerateSortPermutations::new(&shares[2]);
+
+        let h0_future = sortperms0.execute(ctx0);
+        let h1_future = sortperms1.execute(ctx1);
+        let h2_future = sortperms2.execute(ctx2);
+
+        let result = try_join!(h0_future, h1_future, h2_future).unwrap();
```
benjaminsavage, line 123: Let's inline this:

```
    GenerateSortPermutations::new(&shares[0]).execute(ctx0),
    GenerateSortPermutations::new(&shares[1]).execute(ctx1),
    GenerateSortPermutations::new(&shares[2]).execute(ctx2),
]).await?;

File: src/protocol/modulus_conversion/convert_shares.rs [hunk 19]
```diff
@@ -108,6 +108,36 @@ impl ConvertShares {
...
+    input: &[u64],
+) -> Result<Vec<Vec<Replicated<F>>>, BoxError> {
+    let converted_shares = try_join_all(zip(repeat(ctx), input).enumerate().map(
+        |(record_id, (ctx, row))| async move {
+            ConvertShares::new(XorShares {
+                num_bits: 64,
+                packed_bits: *row,
+            })
+            .execute(
+                ctx.narrow(&ModulusConversion(record_id.try_into().unwrap())),
```
benjaminsavage, line 122: This is definitely wrong. This enum only supports values up to 64... we should be able to sort millions.

I *think* there is this hack where if you have NEVER called `narrow()` then everything just works. Like, you can clone the context and invoke the same thing multiple times for the different record IDs. You can use this trick exactly once though. As soon as you've called "narrow()" it doesn't work anymore.

benjaminsavage (changes requested):
This is so exciting to see it all coming together! Can't wait to get this landed!

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Done!

src/protocol/sort/generate_sort_permutations.rs
richajaindce: True.. Let me change it

src/protocol/sort/generate_sort_permutations.rs
richajaindce: I need help on this from @akoshelev. The issue is that we return a &str which expects us to allocate some memory. Wasnt able to think of a way to code this

src/protocol/sort/generate_sort_permutations.rs
richajaindce: Changed them to bit_i_permutation and bit_i_minus_1_permutation instead

src/protocol/sort/generate_sort_permutations.rs
richajaindce: Changing it!

src/protocol/sort/generate_sort_permutations.rs
richajaindce: Sounds good

src/protocol/sort/generate_sort_permutations.rs
richajaindce: Ahh we cant since maximum number of rows can be = Field size (since we secret share permutation in the field). If we are using Fp31 for now, it can be 31 at max for now.

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: This is the number of columns which will be created in the transpose (which is same as number of rows = 1 million in the example). Let me make it more accessible. However, agree it would be better to not create extra memory for doing transpose. Working on it

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Yes these are the columns in convert share output which need to become rows in the transpose

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Doesnt this get called per bit? Then we can run Modulus conversion for upto 64 bits?

src/protocol/sort/mod.rs
akoshelev: macro?

src/protocol/sort/generate_sort_permutations.rs
akoshelev: I think it would be tricky to get `i` as `u8` because `enumerate()` returns `usize`. casting `usize` to `u8` is fallible, hence `try_into` here.

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage: What do you mean? This is still not a parameter to the function. It's still hard-coded.

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage:
```
ConvertShares::new(XorShares {
    num_bits: 64,
    packed_bits: row,
}).execute(ctx, record_id).await?
```

That will convert one input `row` into a `Vec` of `Replicated<F>` having length `64`. It converts all the bits at once with one invocation of this protocol.

src/protocol/sort/generate_sort_permutations.rs
benjaminsavage: Oh that's right... I forgot about that.... Let me create an Fp(Some large prime) class for you to test with!

src/protocol/sort/mod.rs
martinthomson: Might I suggest that you construct something more like:

SortBit(u8) : Step >> enum { ModulusConversion, BitPermutation, ApplyInv, Compose }

!  That is, nest this so that you get "bit45/modconv" or "bit7/bitperm".  Then we only have to do the `AsRef<str>` implementation that is indexed by a u8 once.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -108,6 +108,37 @@ impl ConvertShares {
...
+                ctx.narrow(&ModulusConversion(record_id)),
+                RecordId::from(record_id),
+            )
+            .await
+        },
+    ))
+    .await?;
+
+    //Transpose the output to return bit-wise results
+    let cols = vec![Replicated::new(F::ZERO, F::ZERO); converted_shares.len()];
```
martinthomson, line 132: An allocation on this scale is not ideal.  I would instead use std::iter::repeat()

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -0,0 +1,135 @@
...
+        &self,
+        ctx: ProtocolContext<'_, F>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
+        let mut bits = convert_all_shares(&ctx, self.input, self.num_bits).await?;
+        let mut bit_i_minus_1_compose = BitPermutation::new(&bits[0])
+            .execute(ctx.narrow(&BitPermutationStep(0)))
+            .await?;
+        for (bit_num, bit_value_share) in bits.iter_mut().enumerate().skip(1) {
+            SecureApplyInv::execute(
+                &ctx.narrow(&ApplyInv(bit_num.try_into().unwrap())),
```
martinthomson, line 54: We need to start passing by value for these.

src/protocol/sort/mod.rs
richajaindce: Just realized that modconv would run per record (vs all other protocols running per bit) so its size would be usize. How should we be handling that?

martinthomson: Don't you just convert each bit as a value in Fp2 into a value in F?  And each record is indexed with a RecordId.

src/protocol/sort/generate_sort_permutations.rs
benjaminsavage: #190

src/protocol/sort/mod.rs
richajaindce: Yes thats what we need. However, we have been needing to narrow the context for record else step complains of duplicate context. Example here of narrowing:  https://github.com/private-attribution/ipa/blob/main/src/protocol/modulus_conversion/convert_shares.rs#L245
@martinthomson @benjaminsavage

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: not needed anymore

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -156,7 +156,7 @@ pub async fn convert_shares_for_a_bit<F: Field>(
                 packed_bits: *row,
             })
             .execute_one_bit(
-                ctx.narrow(&ModulusConversion(record_id)),
+                ctx,
```
richajaindce, line 159: @martinthomson made this commit to demo ModulusConversion step is failing when called for different records. Let us know how to best resolve this. The sort test fails with assert "Fails with assert : panicked at 'Refined 'protocol/sort0/mod_conv' with step 'double_random' twice'"

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -0,0 +1,145 @@
...
+            let mut bit_value_share = convert_shares_for_a_bit(
+                ctx.narrow(&ModulusConversion),
+                self.input,
+                self.num_bits,
+                bit_num,
+            )
+            .await?;
+            if bit_num != 0 {
+                SecureApplyInv::execute(
+                    &ctx.narrow(&ApplyInv),
```
martinthomson, line 60:
```suggestion
                    ctx.narrow(&ApplyInv),
```

The last pass-by-reference.

src/protocol/modulus_conversion/convert_shares.rs
martinthomson: Oh.  Looking at your code, it is basically perfect in this regard, so I'll have to look at the modulus conversion stuff now...

src/protocol/sort/generate_sort_permutation.rs
richajaindce: Apologies missed this one. Changing it

src/protocol/context.rs
```diff
@@ -63,6 +65,31 @@ impl<'a, F: Field, SS: SecretSharing<F>> ProtocolContext<'a, SS, F> {
...
+            record_id: self.record_id,
+            _marker: PhantomData::default(),
+        }
+    }
+
+    #[must_use]
+    /// Make a sub-context which is bound to a record in case the same step is bound to a different `record_id`
+    /// # Panics
+    /// Panics in case the context is already bound to the same `record_id`
+    pub fn bind(&self, record_id: RecordId) -> Self {
```
richajaindce, line 77: Thanks @akoshelev for helping with a sketch to introduce a bind to bind a step to a record in case it is already bound to a different record. @martinthomson would this work for now while we find a better solution?

src/protocol/mod.rs
```diff
@@ -93,6 +93,15 @@ impl PartialEq for UniqueStepId {
...
 
 impl UniqueStepId {
+    #[must_use]
+    pub fn from_step_id(step: &Self) -> Self {
+        Self {
+            id: step.id.clone(),
+            #[cfg(debug_assertions)]
+            used: Arc::new(Mutex::new(HashSet::new())),
+        }
+    }
```
benjaminsavage, line 103: Wait... am I understanding this correctly?

It looks like every time you call `ctx.bind(record_id)` it creates a brand-new, empty hash set, with associated wrapping Mutex. That seems wrong to me. Isn't the idea that all of the contexts would share a pointer to the *same* hash set?

richajaindce: For this implementation yes it is creating a hash set per record ID. However, since opt-level=3 doesnt look at debug_assertions, do you think it is ok to leave it like this for now until we get to a better solution?

akoshelev: that's my code and it looks funky I agree. We need to create a brand new hash set to make sure we can narrow to the same step more than once. One alternative to that is to keep record id in the hashset which is bad. This code is just to unblock testing sort and we need a better long term solution for sure #135

src/protocol/context.rs
benjaminsavage: @martinthomson is travelling to London this week. This looks like an acceptable compromise to me (for now), although as we have discussed it does not provide us with sufficient protection against re-use of the same PRSS context the generate shared randomness multiple times. We can address that in another PR, e.g. #208

src/protocol/mod.rs
benjaminsavage: Per our VC discussion today I get it: each record-bound-context is going to be a totally independent entity, only validating that it isn't narrowed to the same exact step multiple times. This provides no protection against PRSS re-use - and that will be addressed elsewhere, e.g. #208

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -90,6 +90,31 @@ impl ConvertShares {
...
+) -> Result<Vec<Replicated<F>>, BoxError> {
+    let converted_shares = try_join_all(zip(repeat(ctx), input).enumerate().map(
+        |(record_id, (ctx, row))| async move {
+            let record_id = RecordId::from(record_id);
+            ConvertShares::new(XorShares {
+                num_bits,
+                packed_bits_left: row.0,
+                packed_bits_right: row.1,
+            })
+            .execute_one_bit(ctx.bind(record_id), record_id, bit_index)
```
benjaminsavage, line 110: One downside of this `bind()` approach, is that calling `bind()` within a protocol like this means we will have to call `bind()` again in the next protocol. We can't re-use the contexts.

So after we convert these shares, if the next step is to `sort()`, we will need to call `bind()` once per row, once again.

I can live with this for now, but we should think about how we can potentially avoid this (or if this is a problem we are concerned with. `bind()` is a pretty cheap operation...

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            let mut bit_i_permutation = BitPermutation::new(&bit_value_share)
+                .execute(ctx.narrow(&BitPermutationStep))
+                .await?;
+
+            if bit_num != 0 {
+                Compose::new(&mut i_minus_1_compose, &mut bit_i_permutation)
+                    .execute(ctx.narrow(&ComposeStep))
+                    .await?;
+            }
```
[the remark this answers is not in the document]
richajaindce: Originally the paper is written like the way you mention i.e. for bit0 and then a loop from 1..bit_size. I thought this was was easier to understand. Let me restructure and then we can decide. 
Agree on the mutable, checking if it can be done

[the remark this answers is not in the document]
richajaindce: Addressed clear update/modification in #210

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -0,0 +1,159 @@
...
+        for bit_num in 1..self.num_bits {
+            let ctx_bit = ctx.narrow(&Sort(bit_num));
+            let bit_i = convert_shares_for_a_bit(
+                ctx_bit.narrow(&ModulusConversion),
+                self.input,
+                self.num_bits,
+                bit_num,
+            )
+            .await?;
+            let applied_bit = SecureApplyInv::execute(
```
benjaminsavage, line 73: Naming is so hard... I'm really struggling to suggest a better name here... We aren't applying the _bit_ so much as applying the current "composed permutation", which is a sorting based on all the preceding bits... 

I'm leaning towards: "bit_i_sorted_by_less_significant_bits"... which I know is stupidly long... but at least it explains what this protocol is doing =).

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -0,0 +1,159 @@
...
+
+            let bit_i_permutation = BitPermutation::new(&applied_bit)
+                .execute(ctx_bit.narrow(&BitPermutationStep))
+                .await?;
+
+            let new_composed_permutation = Compose::new(composed_permutation, bit_i_permutation)
+                .execute(ctx_bit.narrow(&ComposeStep))
+                .await?;
+
+            composed_permutation = new_composed_permutation;
```
benjaminsavage, line 88:
```
composed_permutation = Compose::new(composed_permutation, bit_i_permutation)
    .execute(ctx_bit.narrow(&ComposeStep))
    .await?;
```

benjaminsavage (approved):
This is looking much better now! I think the changes in #210 _really_ help a lot.

Once we sort out all the comments on #210 and get it landed - I only have extremely minor comments left on this diff. Then let's land it!

src/protocol/sort/generate_sort_permutation.rs
richajaindce: trying to keep new composed different from older one, so renaming composed_less_significant_bits_permutation and composed_i_permutation

benjaminsavage (approved):
Looks good!

PR #187, opened 2022-11-01
Make multiply and malicious_multiply non-async functions on `ProtocolContext`

akoshelev:
The original idea of using `async` was to provide backpressure to slow down protocols that are moving too fast. However, after doing some benchmarking, creating the multiplication struct is almost zero cost and protocol will be blocked after computing `right_d` anyway.

There is no reason to provide backpressure at the creation time - it carries no benefit compared to providing backpressure on `execute` call.


thurstonsand (approved):
nice simplification

martinthomson (approved):
WFM

PR #188, opened 2022-11-02
Add new action to check deps against advisory DB

akoshelev:
Two new actions have been added. One runs the security audit (cargo-audit) if any of the cargo config files has changed. Another one runs periodically and scans Cargo.lock against advisory db. If a new CVE is posted there, new issue will be created in IPA repo (at least I hope it will, I haven't tested that part)

## Testing

Tested it locally to confirm that action runs if `Cargo.*` files changed: https://github.com/akoshelev/raw-ipa/actions/runs/3375031039

### DIFF
--- a/.github/workflows/audit-check.yml
+++ b/.github/workflows/audit-check.yml
@@ -1,0 +1,24 @@
+name: Check dependencies against Advisory DB
+on:
+  push:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+  pull_request:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+jobs:
+  audit:
+    name: Security audit
+    env:
+      CARGO_TERM_COLOR: always
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+      contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/audit-check.yml
```diff
@@ -0,0 +1,24 @@
...
+      - '**/Cargo.lock'
+  pull_request:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+jobs:
+  audit:
+    name: Security audit
+    env:
+      CARGO_TERM_COLOR: always
```
thurstonsand, line 15: only used here, and not in audit-daily?

thurstonsand (approved):

.github/workflows/audit-check.yml
akoshelev: daily should create an issue if there is a violation so I was hoping that nobody would ever look into the CLI output of it. Don't have a strong opinion about it though....

thurstonsand: ah that makes sense, as long as all relevant info is captured in the issue

thurstonsand (approved):

PR #189, opened 2022-11-03
adding method to convert just one bit

benjaminsavage:
@richajaindce - this might be helpful to you in #186

### DIFF
--- a/src/protocol/modulus_conversion/convert_shares.rs
+++ b/src/protocol/modulus_conversion/convert_shares.rs
@@ -106,11 +106,44 @@ impl ConvertShares {
             });
         try_join_all(futures).await
     }
+
+    #[allow(dead_code)]
+    pub async fn execute_one_bit<F: Field>(
+        &self,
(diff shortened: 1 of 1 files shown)


richajaindce (approved):
Looks good! Although there is code duplication between execute and execute_one_bit, we might end up not using execute. Lets land this and then we can revisit that later.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -106,11 +106,44 @@ impl ConvertShares {
             });
         try_join_all(futures).await
     }
+
+    #[allow(dead_code)]
```
martinthomson, line 110: This doesn't get called from the other code.  That seems like a real problem.

richajaindce: Not sure if I follow. For now, we have left the execute function here in case it would be needed. We plan to purge it if finally it is not needed.

PR #190, opened 2022-11-03
Creating a prime field with the largest 32-bit integer

benjaminsavage:
In #186, @richajaindce wants to write a test for sorting N rows of data. Sadly, she cannot test sorting more than 30 rows of data, because 30 is the maximum value we can represent with our silly test field Fp31! 

So I created a much bigger field, using the prime number `2^32 - 5`, which is the largest prime number less than 2^32.

### DIFF
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -27,6 +27,10 @@ impl Int for u8 {
     const BITS: u32 = u8::BITS;
 }
 
+impl Int for u32 {
+    const BITS: u32 = u32::BITS;
+}
+
 pub trait Field:
     Add<Output = Self>
     + AddAssign
@@ -48,6 +52,7 @@ pub trait Field:
     type Integer: Int;
 
     const PRIME: Self::Integer;
+    const PRIME_TO_INTMAX_DELTA: Self::Integer;
     /// Additive identity element
     const ZERO: Self;
     /// Multiplicative identity element
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -51,7 +51,6 @@ pub trait Field:
     type Integer: Int;
 
     const PRIME: Self::Integer;
-    const PRIME_TO_INTMAX_DELTA: Self::Integer;
     /// Additive identity element
     const ZERO: Self;
     /// Multiplicative identity element
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -15,7 +15,6 @@ pub trait Int:
     + Ord
     + Sub<Output = Self>
     + Into<u128>
-    + From<u8>
     + Shr<u32, Output = Self>
     + BitAnd<Self, Output = Self>
     + PartialEq
@@ -60,36 +59,6 @@ pub trait Field:
     /// Derived from the size of the backing field, this constant indicates how much
     /// space is required to store this field value
     const SIZE_IN_BYTES: u32 = Self::Integer::BITS / 8;
-
-    /// computes the multiplicative inverse of `self`. It is UB if `self` is 0.
-    #[must_use]
-    fn invert(&self) -> Self {
-        debug_assert!(
-            self != &Self::ZERO,
-            "Multiplicative inverse is not defined for 0"
-        );
-
-        self.pow(Self::PRIME - Self::ONE.into() - Self::ONE.into())
-    }
-
-    /// Computes modular exponentiation, self^exp (mod PRIME)
-    #[must_use]
-    fn pow(&self, exp: Self::Integer) -> Self {
-        let mut term = Self::ONE;
-        // compute: x^exp
-        // binary representation of exp: a_k*2^k + a_k-1*2^k-1 + ... + a_0*2^0
-        // x^exp = x^a_0 + (x^2)^a_1 + ... + (x^(2^k))^a_k
-        // * start at 0 and each time square the term.
-        // * term contributes to the result iff a_k bit is 1 (as (x^a)^0 == 1)
-        for i in (0..Self::Integer::BITS).rev() {
-            term *= term;
-            if (exp >> i) & Self::Integer::from(1) != Self::Integer::from(0) {
-                term *= *self;
-            }
-        }
-
-        term
-    }
 
     /// Blanket implementation to represent the instance of this trait as 16 byte integer.
     /// Uses the fact that such conversion already exists via `Self` -> `Self::Integer` -> `Into<u128>`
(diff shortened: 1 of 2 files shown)


src/ff/prime_field.rs
```diff
@@ -167,15 +180,30 @@ pub struct Fp31(<Self as Field>::Integer);
...
     const PRIME: Self::Integer = 31;
+    const PRIME_TO_INTMAX_DELTA: Self::Integer = 225;
     const ZERO: Self = Fp31(0);
     const ONE: Self = Fp31(1);
 }
 
 field_impl! { Fp31 }
 
+#[derive(Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
```
akoshelev, line 191: I believe we don't need this anymore. There are two special methods to serialize and deserialize a field. I forgot to remove it from `Fp31` :(

src/ff/prime_field.rs
```diff
@@ -67,7 +73,13 @@ macro_rules! field_impl {
 
         impl From<$field> for u8 {
```
akoshelev, line 74: eh I don't think we should have those conversions defined for every field. In general, the only thing we could guarantee is conversion from field to the largest int (u128). Everything else is fallible in general (like values from $F_{2^{32}-5}$ field > 255 cannot be converted to `u8`. The convenience of doing just `u8::from($field)` comes with a steep price of runtime errors that are hard to debug and investigate. 

we could implement `TryInto` for them if we really want .

src/ff/prime_field.rs
```diff
@@ -8,7 +8,13 @@ macro_rules! field_impl {
             type Output = Self;
 
             fn add(self, rhs: Self) -> Self::Output {
-                Self((self.0 + rhs.0) % Self::PRIME)
+                let (result, did_overflow) = self.0.overflowing_add(rhs.0);
```
akoshelev, line 11: this looks dubious - can we just use regular overflow check? given that we branch here anyway

```rust
fn modulo_add(a: u32, b: u32) -> u32 {
    if a < PRIME - b {
        // no overflow
        a + b
    } else {
        // overflow, wrap around PRIME
        let d = PRIME - b;
        a - d
    }
}
```

benjaminsavage: That logic is incorrect. In the `else` case you're returning
```
a + PRIME - b
```
when what we want is:
```
a + b - PRIME
```

I think the code I have *is* what we want. The problem is that we just have to handle integer overflow when you add two very large u32 values. We can fortify this with a bunch of tests (I already have a few that helped me catch issues).

If we want to eliminate the branch, we can do so like this:
```
let (result, did_overflow) = self.0.overflowing_add(rhs.0);
Self((result + u32::from(did_overflow) * Self::PRIME_TO_INTMAX_DELTA) % Self::PRIME)
```

benjaminsavage: @martinthomson - how would you suggest we implement this?

martinthomson: Promoting to the next larger integer type might be easier.

Self::Integer::try_from((Self::OverflowInt::from(a) + Self::OverflowInt::from(b)) % Self::OverflowInt::from(Self::PRIME)).unwrap()

src/ff/prime_field.rs
martinthomson: I agree with Alex.  Though we could probably change this to u64 on the basis that we don't ever want something larger than that.

src/ff/field.rs
```diff
@@ -56,36 +60,6 @@ pub trait Field:
...
-        for i in (0..Self::Integer::BITS).rev() {
-            term *= term;
-            if (exp >> i) & Self::Integer::from(1) != Self::Integer::from(0) {
-                term *= *self;
-            }
-        }
-
-        term
-    }
-
```
benjaminsavage, line 88: I think it was only the Shamir secret sharing code that had any dependency on these functions. Now that has been removed, we no longer need to maintain this. Can always bring it back from version control in the future if we need it again.

src/ff/prime_field.rs
benjaminsavage: Both approaches seem rather wasteful. In the one case, that's a lot of extra copy stuff around. In the other, we are losing another bit of statistical security.

martinthomson:
FpM31 has a fast implementation of modular reduction: https://speakerdeck.com/gtank/engineering-prime-numbers?slide=21  (fast reduction for $2^{32}-5$ on slide 23).

src/ff/prime_field.rs
martinthomson: Upconversion to u64 is free in practice as long as you are in a register (I believe).

benjaminsavage: OK, free sounds good =). I'll go with that.

akoshelev: agree with Martin that upcasting to u64 is free, however I am not sure if we can make a blanket implementation for every field that does that. It can be perfectly fine and optimal for 32 bit field, but for 64 bit and/or default implementation I'd rather go with manual overflow check - apart from my buggy code, I am not sure I understand what is the issue there

akoshelev: This would work for additions, but don't we have the same issue for multiplications? right now everything gets upcasted to `u128` to multply, I wonder if it is going to offset all the gains we get by using `u64` in additions

src/ff/prime_field.rs
```diff
@@ -1,14 +1,15 @@
...
 macro_rules! field_impl {
-    ( $field:ty ) => {
+    ( $field:ty, $int:ty ) => {
         impl std::ops::Add for $field {
             type Output = Self;
 
             fn add(self, rhs: Self) -> Self::Output {
-                Self((self.0 + rhs.0) % Self::PRIME)
+                let c = u64::from;
+                debug_assert!(c(Self::PRIME) < (u64::MAX >> 1));
```
benjaminsavage, line 11: I tried to use `const_assert!` from the `static_assertions` crate [docs](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.const_assert.html) but it didn't work.

It basically objected to the use of `Self::` within that macro.

src/ff/prime_field.rs
```diff
@@ -84,6 +81,12 @@ macro_rules! field_impl {
             }
         }
 
+        impl From<$field> for $int {
+            fn from(v: $field) -> Self {
+                v.0
+            }
+        }
```
benjaminsavage, line 88: This is used in the `as_u128` function on the `Field` trait.
```
fn as_u128(&self) -> u128 {
    let int: Self::Integer = (*self).into();
    int.into()
}
```

akoshelev (approved):

Issue #192, opened 2022-11-03
Upgrading IPA infrastructure to support malicious and semi-honest protocols


akoshelev:
This issue exists to solicit some early feedback on upgrading the infrastructure to seamlessly support malicious and semi-honest protocols. The current approach is that protocols are generic over secret share types and behave the same regardless of whether we use regular replicated secret sharing or malicious one. That model has a lot of benefits: the protocol code is the same, so it only needs to be written once. Verification happens behind the scenes and protocols don't know about it. 

I've been working on this for some time and here is the model I propose we use. I am not 100% confident that I will be able to implement this (this change is WIP) but would love some feedback if someone thinks this is not going to work anyway @benjaminsavage , @martinthomson 

```rust
/// New trait that all secret sharing models must implement. So far it seems the only thing we 
/// can do wit them is addition and multiplication
trait SecretShare<F: Field> : Add<Output = Self> + Mul<Output = Self> + Sized {}

/// Protocol context is now generic over share type and field
struct ProtocolContext<S: SecretShare<F>, F: Field> {
    s: PhantomData<S>,
    f: PhantomData<F>,
    // other fields like prss go here
}

/// Both semi-honest and malicious replicated sharings implement the secret share trait.
impl <F: Field> SecretShare<F> for Replicated<F> {}
impl <F: Field> SecretShare<F> for MaliciousReplicated<F> {}

// Generic multiplication now looks different. It requires to provide everything at once, i.e.
// record id, a, b
#[async_trait::async_trait]
trait Multiply<F: Field> {
    type Share: SecretShare<F>;

    async fn multiply(self, record_id: RecordId, a: Self::Share, b: Self::Share) -> Result<Self::Share, BoxError>;
}


/// For protocols that use replicated secret sharing we use semi-honest multiplication
#[async_trait::async_trait]
impl <F: Field> Multiply<F> for ProtocolContext<Replicated<F>, F> {
    type Share = Replicated<F>;

    async fn multiply(self, record_id: RecordId, a: Self::Share, b: Self::Share) -> Result<Self::Share, BoxError> {
        todo!()
    }
}

/// For protocols that use malicious secret sharing we use malicious multiplication
#[async_trait::async_trait]
impl <F: Field> Multiply<F> for ProtocolContext<MaliciousReplicated<F>, F> {
    type Share = MaliciousReplicated<F>;

[...]

PR #193, opened 2022-11-04
Make ProtocolContext generic over secret sharing. Add secret sharing trait

akoshelev:
This change:

* Introduces a new trait: `SecretShare` that all the existing secret sharing implement.
* Introduces a new trait: `SecureMul` that is implemented by both semi-honest and malicious multiplications that already exists
* Moves multiplication module to its own folder
* Changes `ProtocolContext` to be generic over `SecretShare`
* Implement `SecureMul` for malicious and semi-honest contexts. Note that all multiplications now have to go through the context in one pass (i.e. instead of `ctx.multiply(...).execute(a, b)`, now it is `ctx.multiply(..., a, b)`
* Updates the usages of new protocol context in sort protocols and unit tests
* Refactors malicious context to prohibit calling `upgrade_to_malicious` on contexts that are already malicious
* addresses #192 

This change does not:

* make protocols generic over secret sharing type. Everything is implemented over `Replicated<F>` for now, but this paves the path to make them generic

## Notes

there is a bunch of noise inside this large PR, this is how I would suggest  to review it

* Start with `src/secret_sharing/mod.rs`  - this is where the new trait `SecretShare` is defined. See if supertraits for it make sense.
* then look at `src/protocol/mul/mod.rs` - new `SecureMul` trait lives there. Also malicious and semi-honest multiplications for the corresponding contexts are defined there (just a hop to the existing implementations)
* `src/protocol/context.rs` contains changes made against `ProtocolContext`. 
* Check how `to_semi_honest(self)` method on the context is used inside malicious multiplication `src/protocol/mul/malicious.rs` 

Everything else I believe is a change to conform to this new interface


src/protocol/reveal_additive_binary.rs
```diff
@@ -19,8 +20,8 @@ pub struct RevealAdditiveBinary {}
 
 impl RevealAdditiveBinary {
     #[allow(dead_code)]
-    pub async fn execute<F: Field>(
-        ctx: ProtocolContext<'_, F>,
+    pub async fn execute<S: SecretShare<F>, F: Field>(
+        ctx: ProtocolContext<'_, S, F>,
         record_id: RecordId,
         input: Fp2,
     ) -> Result<Fp2, BoxError> {
```
benjaminsavage, line 27: OK, it's kind of weird that this is the one protocol that appears to be "generic over any kind of secret sharing", and it's a bit concerning because logically speaking, this protocol is _only_ applicable for an *additive secret sharing* (one we don't even have a class for) and because `input` is typed as Fp2.

So everything is a bit weird here. A few ideas:

1. This approach has no path to upgrade to malicious, so I'll probably need to change it to use replicated secret sharing instead of additive.
2. Given that Fields are generic and we have support for Fp2, we should probably just have one, single Reveal protocol and use it for both Fp2 and other cases.
3. This protocol just accepts an Fp2 argument that has no compile-time (or run-time) relationship to the SecretSharing of the protocol context. I'm not sure if that's a problem. I guess it's not... I have to think about this. I think that the SecretSharing (and Field) of the protocol context relate to the malicious security information-theoretic MACs being accumulated. In which case, this is OK. In our protocol we wind up revealing a bool, and that's fine. It's different than the field we are working in. I just need to think about how (if at all) it should affect that MAC.

src/protocol/reveal_additive_binary.rs
```diff
@@ -63,7 +64,7 @@ mod tests {
         let mut rng = rand::thread_rng();
 
         let world: TestWorld = make_world(QueryId);
-        let ctx = make_contexts::<Fp31>(&world);
+        let ctx = make_contexts::<Fp2>(&world);
```
benjaminsavage, line 67: It's weird that this doesn't matter =). Honestly, let's keep it at Fp31 for the reasons in my comment above.

src/protocol/sort/bit_permutations.rs
```diff
@@ -46,7 +47,7 @@ impl<'a, F: Field> BitPermutations<'a, F> {
     #[allow(dead_code)]
     pub async fn execute(
         &self,
-        ctx: ProtocolContext<'_, F>,
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
```
benjaminsavage, line 50: This is the first one I want to modify to use SecretShare<F> instead of Replicated.

src/secret_sharing/mod.rs
```diff
@@ -1,5 +1,14 @@
...
 mod replicated;
 
+use crate::ff::Field;
 pub use malicious_replicated::MaliciousReplicated;
 pub use replicated::Replicated;
+use std::fmt::Debug;
+use std::ops::{Add, AddAssign, Mul, Neg, Sub, SubAssign};
+
+/// Secret share of a secret has additive and multiplicative properties.
+pub trait SecretShare<F>: Add + AddAssign + Neg + Sub + SubAssign + Mul<F> + Debug + Sized {}
```
benjaminsavage, line 11: I would prefer to call it: `SecretSharing<F>` since that sounds more conversational.

benjaminsavage: Can we somehow add "possible to get a sharing of ONE", (or really of any constant) to this?

benjaminsavage (approved):

src/protocol/malicious.rs
```diff
@@ -156,7 +156,10 @@ impl<F: Field> SecurityValidator<F> {
     /// ## Panics
     /// Will panic if the mutex is poisoned
     #[allow(clippy::await_holding_lock)]
-    pub async fn validate(self, ctx: ProtocolContext<'_, F>) -> Result<(), BoxError> {
+    pub async fn validate(
+        self,
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
```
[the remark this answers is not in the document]
akoshelev: It is because it calls `reveal` and `check_zero` protocols that currently support `Replicated` only.
once they are generic over any secret share type, we can change this code

src/protocol/mul/malicious.rs
```diff
@@ -83,11 +84,19 @@ impl<'a, F: Field> MaliciouslySecureMul<'a, F> {
...
         let duplicate_multiply_ctx = self.ctx.narrow(&Step::DuplicateMultiply);
         let random_constant_prss = self.ctx.narrow(&Step::RandomnessForValidation).prss();
-        let (ab, rab) = try_join(
-            SecureMul::new(self.ctx, self.record_id).execute(a.x(), b.x()),
-            SecureMul::new(duplicate_multiply_ctx, self.record_id).execute(a.rx(), b.x()),
-        )
-        .await?;
+        let (ab, rab) = {
+            // Convince compiler that neither a nor b will be used across the await point
+            // to relax the requirement for either of them to be Sync
```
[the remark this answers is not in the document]
akoshelev: We call this function from `SecureMul` trait implementation. The trait definition uses `async_trait` crate because Rust does not support async functions inside traits (yet, 1.65 is going to change it). `async_trait` requires return to be `Send` (we can opt out but I don't think we should): https://docs.rs/async-trait/latest/async_trait/#explanation

future returned by `MaliciouslySecureMul` wasn't `Send` and this change makes it `Send`.

src/protocol/reveal_additive_binary.rs
akoshelev: yea this is mostly mechanical change from me to make everything compile and pass the tests. 

> This approach has no path to upgrade to malicious, so I'll probably need to change it to use replicated secret sharing instead of additive.

Yea it is not immediately obvious to me how to protect against cheating helpers inside mod conversion

> Given that Fields are generic and we have support for Fp2, we should probably just have one, single Reveal protocol and use it for both Fp2 and other cases.

I agree, we should make `Reveal` generic over any secret sharing.

src/secret_sharing/mod.rs
akoshelev: I could add something like `fn one(role: Identity) -> Self` but this does not seem to be enough to construct maliciously secure share? we may also have `fn one(ctx: ProtocolContext<_, Self, F)` but that would require r_share to be provided from the context and I don't know what index to use to generate randomness at this point.

src/protocol/context.rs
```diff
@@ -15,37 +14,30 @@ use crate::{
...
     role: Identity,
     step: UniqueStepId,
     prss: &'a PrssEndpoint,
     gateway: &'a Gateway,
     accumulator: Option<SecurityValidatorAccumulator<F>>,
+    _marker: PhantomData<S>,
 }
 
-impl<'a, F: Field> ProtocolContext<'a, F> {
+impl<'a, F: Field, SS: SecretSharing<F>> ProtocolContext<'a, SS, F> {
```
martinthomson, line 32: Can we be consistent about S vs SS?

src/protocol/context.rs
```diff
@@ -101,18 +94,58 @@ impl<'a, F: Field> ProtocolContext<'a, F> {
...
+    /// `ProtocolContext<'a, S: SecretShare<F>, F: Field>` as the context. If that is not possible,
+    /// this implementation makes it easier to reinterpret the context as semi-honest.
+    ///
+    /// The context received will be an exact copy of malicious, so it will be tied up to the same step
+    /// and prss.
     #[must_use]
-    pub fn malicious_multiply(self, record_id: RecordId) -> MaliciouslySecureMul<'a, F> {
-        let accumulator = self.accumulator.as_ref().unwrap().clone();
-        MaliciouslySecureMul::new(self, record_id, accumulator)
+    pub fn to_semi_honest(self) -> ProtocolContext<'a, Replicated<F>, F> {
```
martinthomson, line 141: Question for @benjaminsavage here: what sort of safeguards do you want to have on the context regarding the validation step?  Do you want to have a way to encourage/force someone to do the malicious validation before doing this downgrade?

src/protocol/malicious.rs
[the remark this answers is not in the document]
martinthomson: Open an issue for that?

martinthomson (approved):

src/protocol/context.rs
benjaminsavage: I vote for just `S`

src/protocol/context.rs
benjaminsavage: I am not convinced this method will stick around. I understand that we need it right now - but lets see how things play out. I suspect it'll just dissapear.

src/protocol/malicious.rs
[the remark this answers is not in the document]
akoshelev: #201

akoshelev:
folks I am going to merge it, but would happily follow up on `SS` generic parameter - we can continue iterating on it on this PR.

PR #195, opened 2022-11-04
Bitwise PrefixOr protocol

taikiy:
# Overview
This diff implements `PrefixOr`, one of sub-protocols to perform inequality checks. This is an implementation of "5.2 Prefix-Or" described in this [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf), [12].

# Description
`PrefixOr` takes inputs $[a\_1]\_p,...,[a\_l]\_p$ where $a_1,...,a\_l ∈ \\{0,1\\} ⊆ F\_p$ and computes $[b\_1]\_p,...,[b\_l]\_{p}$ where $b\_i = \vee\_{j=0}^{i} a\_j$.

In other words, PrefixOr( $a\_B$ ) will output $b\_B$ where b's i'th bit is 1 iff a's i'th bit is the first bit to have 1.

# Example
`PrefixOr(0 0 0 1 0 0 1 0) -> 0 0 0 1 1 1 1 1`


src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// ```
+    async fn step8(
+        f: &[Replicated<B>],
+        b: &[Replicated<B>],
+        ctx: ProtocolContext<'_, B>,
+    ) -> Result<Vec<Replicated<B>>, BoxError> {
+        let lambda = f.len();
+        let mut mul = Vec::new();
+        for (i, &f_bit) in f.iter().enumerate().take(lambda) {
+            for (j, &b_bit) in b.iter().enumerate().take(lambda) {
```
benjaminsavage, line 198: I don't love the use of `.take(lambda)` when we *know* that these vectors are indeed this length.

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+                let c = ctx.clone();
```
benjaminsavage, line 199: I think that these lambda^2 multiplications are currently all using the same PRSS randomness - which is not good. They probably need different contexts, possibly narrowed to the "bit" you're dealing with?

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+            })
+            .collect::<Vec<_>>()
+    }
+
+    /// Execute `PrefixOr`
+    #[allow(dead_code)]
+    #[allow(clippy::many_single_char_names)]
+    pub async fn execute(
+        &self,
+        ctx: ProtocolContext<'_, B>,
```
benjaminsavage, line 234: I suspect we need to pass in a record_id

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+        b: &[Replicated<B>],
+        ctx: ProtocolContext<'_, B>,
+    ) -> Result<Vec<Replicated<B>>, BoxError> {
+        let lambda = f.len();
+        let mut mul = Vec::new();
+        for (i, &f_bit) in f.iter().enumerate().take(lambda) {
+            for (j, &b_bit) in b.iter().enumerate().take(lambda) {
+                let c = ctx.clone();
+                mul.push(
+                    c.multiply(RecordId::from(lambda * i + j))
```
benjaminsavage, line 201: Here I *think* you want to keep the record_id that was passed in, and instead use the context to give yourself different randomness for each step.

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+        // regardless of the dummy bits, but may affect the performance if
+        // `λ^2 - l` becomes large. We should revisit this protocol once the
+        // prototype is complete, and optimize if necessary.
+        // TODO(taikiy): Prefix-Or optimization
+
+        // It should be safe to cast from `usize` to `u32` and back cast since
+        // the input we expect to operate on are 40-bit < `u8` << `u32`.
+        #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+        let lambda = f64::from(self.input.len() as u32).sqrt().ceil() as usize;
+        let dummy = vec![Replicated::new(B::ZERO, B::ZERO); lambda * lambda - self.input.len()];
```
benjaminsavage, line 253: I guess we need this as we plan to utilize u32... so we will pad out 4 dummy values to get to 36?

But we don't really want to compute a square root each time we invoke this protocol do we? So maybe we should special case it? like match on input length and fatal if it's not coded for?

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+        let x = Self::step1(&a, lambda, ctx.narrow(&Step::BlockOr1)).await?;
+        let y = Self::step2(&x, ctx.narrow(&Step::BlockOr2)).await?;
+        let f = Self::step3_4(&x, &y);
+        let g = Self::step5(&f, &a, ctx.narrow(&Step::Mult1)).await?;
+        let c = Self::step6(&g, lambda);
+        let b = Self::step7(&c, ctx.narrow(&Step::BlockOr3)).await?;
+        let s = Self::step8(&f, &b, ctx.narrow(&Step::Mult2)).await?;
+        let b = Self::step9(&s, &y, &f);
+
+        Ok(b[0..self.input.len()].to_vec())
```
benjaminsavage, line 265: Does `to_vec()` do a deep copy? That would make me sad. I'd like to avoid that.

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,365 @@
...
+impl crate::protocol::Step for Step {}
+
+impl AsRef<str> for Step {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::BlockOr1 => "block_or_1",
+            Self::BlockOr2 => "block_or_2",
+            Self::BlockOr3 => "block_or_3",
+            Self::Mult1 => "mult_1",
+            Self::Mult2 => "mult_2",
```
benjaminsavage, line 287: These names could be better =).

benjaminsavage (commented):
I'll continue reviewing the tests - but here are my comments for now.

src/protocol/boolean/prefix_or.rs
taikiy: Thanks for catching. I changed the code to narrow for each bit.

src/protocol/boolean/prefix_or.rs
taikiy: I changed to code to assert if the input length is not either 8, 16, or 32. If the length is one of these, I return 3, 4 or 6 as lambda. Anything else, the protocol panics.

src/protocol/boolean/prefix_or.rs
taikiy: I believe it creates a clone. I couldn't think of other way to make this work. This isn't the only part we clone a vector though. All the steps return `Vec<Replicated<B>>` now, which I think I can do hacky optimizations like passing `&mut` vectors and reusing them. I can address them at the optimization phase?

src/protocol/boolean/prefix_or.rs
taikiy: OK, namings are hard... I changed them to (hopefully) more self-explanatory names.

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    /// Securely computes `∨ [a_1],...[a_n]`
+    async fn block_or(
+        a: &[Replicated<B>],
+        k: usize,
+        ctx: ProtocolContext<'_, Replicated<B>, B>,
+        record_id: RecordId,
+    ) -> Result<Replicated<B>, BoxError> {
+        #[allow(clippy::cast_possible_truncation)]
+        let mut step = IterStep::new("bit", k as u32);
```
benjaminsavage, line 52: I don't think we need a u32 to represent the "bit". There should be at most 64 bits for any of these protocols.

Is this because you're trying to re-use this `IterStep` in a bunch of different places? I don't think we want that coupling. Feel free to make another step abstraction for each protocol.

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,400 @@
...
+        let mut v = Replicated::new(B::ZERO, B::ZERO);
+        for &bit in a {
+            v = Self::bit_or(v, bit, ctx.narrow(step.next()), record_id).await?;
```
benjaminsavage, line 55: This does one extra round of communication - because you're actually invoking the secure multiply operation to multiply with a known zero value. Instead of initialising v to Replicated::default(), why not just initialise to the first element of a?

src/protocol/boolean/prefix_or.rs
```diff
@@ -0,0 +1,400 @@
...
+        for i in 0..TEST_TRIES {
+            let len = BITS;
+            let input: Vec<Fp2> = (0..len).map(|_| Fp2::from(rng.gen::<bool>())).collect();
+            let mut expected: Vec<Fp2> = Vec::with_capacity(len);
+
+            // Calculate Prefix-Or of the secret number
+            input.iter().fold(Fp2::ZERO, |acc, &x| {
+                expected.push(acc | x);
+                acc | x
+            });
```
benjaminsavage, line 362: Is there a way to generate this in one pass, instead of mutating the vector in each iteration? I think there is some `fold` like method for that.

src/protocol/boolean/prefix_or.rs
taikiy: I created `IterStep` originally for the Attribution protocol, where it executes log(n) rounds of multiplications, hence the inner counter is u32. I agree that we don't need u32 to represent the bit for this case, but `IterStep` uses u32 to support other use cases.

src/protocol/boolean/prefix_or.rs
benjaminsavage: Yeah... I can't figure it out either. Whatever, this is just test code =).

benjaminsavage (changes requested):
Requesting changes for the removal of the extra multiplication with zero, and the re-use of IterStep.

src/protocol/boolean/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+/// For each record, we decompose a value into bits (i.e. credits in the
+/// Attribution protocol), and execute some binary operations like OR'ing each
+/// bit. For each bitwise secure computation, we need to "narrow" the context
+/// with a new step to make sure we are using an unique PRSS.
+///
+/// This is a temporary solution for narrowing contexts until the infra is
+/// updated with a new step scheme.
+struct BitOpStep {
+    count: u8,
+    id: String,
```
benjaminsavage, line 14: I don’t think this needs to actually instantiate a string. I think the other ones all avoid this.

benjaminsavage (approved):

PR #197, opened 2022-11-04
Peer discovery

thurstonsand:
implements MPC helper peer discovery as a trait to allow for multiple methods of discovery. includes PRSS config.

Initial implementation uses a config file to describe peers, which should be enough to get us off the ground. We'll consider more sophisticated methods in the future, such as leader/follower style or C&C style.


src/net/discovery/config.rs
```diff
@@ -0,0 +1,204 @@
...
+{
+    "peers": [
+        {
+            "origin": "http://localhost:3000",
+            "public_key": [19,204,244,38,60,236,188,48,245,14,106,139,156,135,67,148,61,221,230,32,
+                121,88,11,192,185,1,155,5,186,143,233,36]
+        },
+        {
+            "origin": "http://localhost:3001",
+            "public_key": [146,91,249,130,67,207,112,183,41,222,29,117,191,79,230,190,152,169,134,
```
akoshelev, line 158: lets use hex encoding in config files instead

src/net/discovery/config.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Reads config from `file_location`. Expects file to be json format
+    /// # Errors
+    /// if the file does not exist, or is in an invalid format
+    #[allow(dead_code)] // TODO: will use in upcoming PR
+    pub fn new(file_location: PathBuf) -> Result<Self, Error> {
+        use std::{fs::File, io::BufReader};
+
+        let contents = File::open(file_location)?;
+        let buffered = BufReader::new(contents);
+        Ok(serde_json::from_reader(buffered)?)
```
akoshelev, line 94: @martinthomson I know that you are very cautious about adding new dependencies, what do you think about using toml-based configs? Json technically works fine, but imo it is bad for configuration because of nesting and braces. Bringing toml would require us to import `config` crate...

src/net/discovery/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Provides a set of peer helpers for an MPC computation. Also includes the client pointing to the
+/// running server. Since the running server is aware of which [`Identity`] it is (`H1`, `H2`, or
+/// `H3`), it should be able to use only the references to other servers. However, it's possible for
+/// a server to send data to itself.
+///
+/// Any potential failures should be captured in the initialization of the implementer.
+#[allow(clippy::module_name_repetitions)] // following standard naming convention
+pub trait PeerDiscovery {
+    fn peers(&self) -> [MpcHelperClient; 3];
+    fn prss(&self, role: Identity) -> prss::Endpoint;
```
akoshelev, line 24: I think it couples too much with the implementation of PRSS and HTTP. It is no longer solves one problem of discovering peers, it functions as helper configuration that is capable of constructing things that are unrelated (http client and prss). I worry that this may lead to "god object".

I would  suggest to make this a peer discovery trait only 

```rust
   // or peer(&self, role: Identity) 
   fn peers(&self) -> [<Information that is necessary to construct http client and prss. PRSS public key, http endpoint, etc>; 3]
```

src/net/discovery/config.rs
```diff
@@ -0,0 +1,204 @@
...
+mod tests {
+    use super::*;
+    use rand::distributions::Alphanumeric;
+    use rand::{thread_rng, Rng};
+    use std::fs::File;
+    use std::io::Write;
+
+    #[allow(unused)] // TODO: how to test prss seed equality?
+    const ROLE: Identity = Identity::H2;
+    #[allow(unused)] // TODO: how to test prss seed equality?
```
akoshelev, line 130: just generate values at the same index - two different prss instances with the same seed will yield the same value

src/net/discovery/config.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn peers(&self) -> [MpcHelperClient; 3] {
+        self.peers
+            .iter()
+            .map(|peer| MpcHelperClient::new(peer.origin.clone().into()))
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|_| panic!("could not convert vec into array"))
+    }
+
+    // TODO: do we need different public key for HTTP and PRSS, or can they share?
```
akoshelev, line 109: I thought about it again and I think the answer is no. Security is one issue, but also TLS may use RSA with public keys being 10x larger than ECC. Martin knows better whether we should use RSA or ECC to encrypt the traffic (for some reason RSA is very popular and I don't have a good explanation why it is preferred over ECC), but there are technical limitations on public key crypto as well. So imo we should use different public keys

src/net/discovery/config.rs
martinthomson: toml is fine (and better than JSON, yes)

src/net/discovery/config.rs
martinthomson: We should have a different key for the TLS server side of things, definitely.  TLS and PRSS are two separate purposes and using the same key for both isn't possible anyway.

We'll want to use ECC, but probably ECDSA with P-256.  RSA signing is maybe OK, but it's unnecessarily inefficient.

src/net/discovery/mod.rs
thurstonsand: gotcha. so return the same information, but let something else (some HTTP entity/PRSS entity) use that information to actually construct things. i think that makes sense to me

src/net/discovery/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    #[derive(Clone, Serialize, Deserialize)]
+    pub struct HttpConfig {
+        pub origin: Origin,
+        pub public_key: PublicKey,
+    }
+
+    #[derive(Clone, Serialize, Deserialize)]
+    pub struct PrssConfig {
+        pub public_key: PublicKey,
```
thurstonsand, line 149: should I be using a different type than `x25519_dalek::PublicKey` here?

src/net/discovery/conf.rs
```diff
@@ -0,0 +1,118 @@
...
+        public_key = "12c09881a1c7a92d1c70d9ea619d7ae0684b9cb45ecc207b98ef30ec2160a074"
+"#;
+
+    fn origin_to_string(origin: &peer::Origin) -> String {
+        let uri = Uri::from(origin.clone());
+        format!("{}://{}", uri.scheme().unwrap(), uri.authority().unwrap())
+    }
+
+    #[test]
+    fn config_from_file() {
```
thurstonsand, line 93: since there's no way to feed `config` crate a string literal, the only test here will be read from file

src/net/discovery/config.rs
thurstonsand: now using toml

src/net/discovery/mod.rs
akoshelev: hex vends serde serializers: https://docs.rs/hex/latest/src/hex/serde.rs.html#12

src/net/discovery/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+pub mod peer {
+    use axum::http::{
+        uri::{Authority, Scheme},
+        Uri,
+    };
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+    use std::fmt::Formatter;
+    use std::ops::{Deref, DerefMut};
+
+    /// Describes just the origin of a url, i.e.: "http\[s\]://\[authority\]", minus the path and
```
akoshelev, line 20: I look at this again and maybe we should just assume that helper addr is a url? that will remove all of this complexity of serializing from here

src/net/discovery/mod.rs
thurstonsand: ok. should I just be returning a byte array for the public_key then?

src/net/discovery/conf.rs
thurstonsand: ah, yeah that does indeed work. thanks

src/net/discovery/mod.rs
thurstonsand: what type are you proposing we use? `Uri` itself is not serializable, so we would still need to handle that. we could do string, and maybe just check that it's valid after init?

src/net/discovery/conf.rs
```diff
@@ -0,0 +1,118 @@
...
+    const H3_PUBLIC_KEY: [u8; 32] = [
+        18, 192, 152, 129, 161, 199, 169, 45, 28, 112, 217, 234, 97, 157, 122, 224, 104, 75, 156,
+        180, 94, 204, 32, 123, 152, 239, 48, 236, 33, 96, 160, 116,
+    ];
+    const H1_URI: &str = "http://localhost:3000";
+    const H2_URI: &str = "http://localhost:3001";
+    const H3_URI: &str = "http://localhost:3002";
+    const EXAMPLE_CONFIG: &str = r#"
+# H1
+[[peers]]
```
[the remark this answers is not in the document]
thurstonsand: i agree this is kinda ugly in toml. it might look better as yaml, but also, explicitly having `H1`/`H2`/`H3` is fine. i can switch

src/net/discovery/conf.rs
```diff
@@ -0,0 +1,118 @@
...
+    use crate::net::discovery::peer::PublicKey;
+    use hyper::Uri;
+    use rand::{
+        distributions::Alphanumeric,
+        {thread_rng, Rng},
+    };
+    use std::fs::File;
+    use std::io::Write;
+
+    const H1_PUBLIC_KEY: [u8; 32] = [
```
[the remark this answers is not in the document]
thurstonsand: sure, done

src/net/discovery/mod.rs
akoshelev: yea I like using str here. less code is better. Parsing a url and validating it should be 2 lines or code

thurstonsand: my only concern at that point is that all implementers of `PeerDiscovery` will be responsible for validating the string. so you'd have 2 lines of code per implementer, as opposed to having this code exist in one place

src/net/discovery/conf.rs
```diff
@@ -0,0 +1,213 @@
...
+    [h3.http]
+        origin = "http://localhost:3002"
+        public_key = "12c09881a1c7a92d1c70d9ea619d7ae0684b9cb45ecc207b98ef30ec2160a074"
+    [h3.prss]
+        public_key = "12c09881a1c7a92d1c70d9ea619d7ae0684b9cb45ecc207b98ef30ec2160a074"
+"#;
+
+    fn origin_to_string(uri: &Uri) -> String {
+        format!("{}://{}", uri.scheme().unwrap(), uri.authority().unwrap())
+    }
```
akoshelev, line 125: you don't need this function, just use `uri.to_string()`,

akoshelev (commented):
I think it is almost ready - one important change I think we should make is to return a reference to `[config; 3]` instead of copying it. Everything else is minor

src/net/discovery/conf.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct Conf {
+    peers: [peer::Config; 3],
+}
+
+impl Conf {
+    /// Reads config from `file_location`. Expects file to be json format
+    /// # Errors
+    /// if the file does not exist, or is in an invalid format
+    #[allow(dead_code)] // TODO: will use in upcoming PR
+    pub fn from_file(file_location: &str) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
thurstonsand: ok, I can remove this, tho I was imagining we may want to potentially override certain values with env vars or possibly stack config files, or other complexities. and i would want to contain that complexity within a function.

but, we don't need that now, so i can remove it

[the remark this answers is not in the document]
thurstonsand: are you saying you would prefer to keep the function, but change its argument to accept the string contents of the file?

src/net/discovery/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Provides a set of peer helpers for an MPC computation. Also includes the client pointing to the
+/// running server. Since the running server is aware of which [`Identity`] it is (`H1`, `H2`, or
+/// `H3`), it should be able to use only the references to other servers. However, it's possible for
+/// a server to send data to itself.
+///
+/// Any potential failures should be captured in the initialization of the implementer.
+#[allow(clippy::module_name_repetitions)] // following standard naming convention
+pub trait PeerDiscovery {
+    fn peers(&self) -> [peer::Config; 3];
```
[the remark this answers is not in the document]
thurstonsand: sure, we can do that

src/net/discovery/conf.rs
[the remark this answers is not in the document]
akoshelev: right, I think we should only have constructor that takes a string slice and constructs the valid config from it.

akoshelev (approved):

src/net/discovery/conf.rs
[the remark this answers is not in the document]
thurstonsand: alright. we can do that

akoshelev (approved):
LGTM. Thanks for making it more ergonomic to use!

PR #198, opened 2022-11-04
Small optimizations for reveal_permutation fn

akoshelev:
Stumbled across this code while looking for patterns how to use `narrow`, decided to send a quick fix

* we don't need to allocate a vector to hold the results of reveal. `try_join_all` returns a `Vec` so we can re-use it if it stores `usize` elements
* no need to take an exclusive reference to permutation for reveal operation
* naming suggestion

### DIFF
--- a/src/protocol/reveal.rs
+++ b/src/protocol/reveal.rs
@@ -43,20 +43,23 @@ pub async fn reveal<F: Field>(
 /// Given a vector containing secret shares of a permutation, this returns a revealed permutation.
 /// This executes `reveal` protocol on each row of the vector and then constructs a `Permutation` object
 /// from the revealed rows.
-#[allow(clippy::cast_possible_truncation, clippy::module_name_repetitions)]
-pub async fn reveal_a_permutation<F: Field>(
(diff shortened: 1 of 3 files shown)


src/protocol/reveal.rs
```diff
@@ -43,20 +43,23 @@ pub async fn reveal<F: Field>(
 /// Given a vector containing secret shares of a permutation, this returns a revealed permutation.
 /// This executes `reveal` protocol on each row of the vector and then constructs a `Permutation` object
 /// from the revealed rows.
-#[allow(clippy::cast_possible_truncation, clippy::module_name_repetitions)]
-pub async fn reveal_a_permutation<F: Field>(
+#[allow(clippy::module_name_repetitions)]
+pub async fn reveal_permutation<F: Field>(
     ctx: ProtocolContext<'_, F>,
-    permutation: &mut [Replicated<F>],
+    permutation: &[Replicated<F>],
```
martinthomson, line 49: 👍

martinthomson (approved):

PR #199, opened 2022-11-05
Adding a Malicious Reveal protocol

benjaminsavage:
...and tests for both successful and unsuccessful cases

### DIFF
--- a/src/protocol/reveal.rs
+++ b/src/protocol/reveal.rs
@@ -99,7 +99,6 @@ pub async fn reveal_permutation<F: Field>(
 mod tests {
     use futures::future::{try_join, try_join_all};
     use proptest::prelude::Rng;
-    use rand::rngs::mock::StepRng;
     use tokio::try_join;
 
     use crate::{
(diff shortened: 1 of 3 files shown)


akoshelev:
this partially addresses #201

src/protocol/reveal.rs
```diff
@@ -40,6 +44,35 @@ pub async fn reveal<F: Field>(
     Ok(input.left() + input.right() + share)
 }
 
+#[allow(dead_code)]
+#[allow(clippy::module_name_repetitions)]
+pub async fn reveal_malicious<F: Field>(
```
akoshelev, line 49: would you expect to eventually use `Malicious<F>` here? I was assuming that we would want to seamlessly plugin malicious reveal into semi-honest protocols by changing the secret share type and using malicious reveal + validation? 

Adopting malicious reveal may need us to have a `Reveal` trait that protocols would execute not actually knowing that they are executing maliciously secure circuit. Does this match your understanding or you know a different way of doing it?

benjaminsavage: I'm not really sure about this.

I *think* that we never actually need to utilize the "randomized" values that are carried around in the MaliciousReplicated shares. It is possible that before we can call "reveal" that we need to valid .validate() - and only if the context passes the validation - then call reveal.

In that case, we could have the impact of calling "validate()" update some kind of state within the ProtocolContext. For example, we could have some bit that say: "currently valid" which would be set to true, and immediately set back to false the next time anything happens.

We could potentially tie this in with the whole "downgrade_to_semi_honest" thing, and introduce a third state.

akoshelev: I think we could use a drop guard to make validation opaque to semi-honest protocols. But at some point we need to decide which reveal protocol we call - semi honest with a cost of 1 communication or malicious with a cost of 2. It would not be great to have this branching inside protocols, so we probably need to leverage protocol context for that that will expose `reveal` method similar to `multiply`. If it is parameterized with malicious shares then we will use malicious reveal logic inside it, otherwise it will be semi-honest

PR #202, opened 2022-11-06
Add `InfraError` member to error type.

akoshelev:
I noticed that we use boxed errors inside protocols in cases where we don't have to because infrastructure errors cannot be converted to `Error` object and therefore `?` operator cannot be used.

This change was inspired by #199 where reveal protocol could've just returned `Result<(), Error>` and test for correct error type inside unit tests.


benjaminsavage (approved):
sure, why not

PR #204, opened 2022-11-06
Suppress similar_names and module_name_repetitions lints

akoshelev:
Discussed this on Wire. We mostly ignore them anyway


benjaminsavage (approved):
Love it.

PR #206, opened 2022-11-07
Bitwise Less-Than Protocol

taikiy:
## Overview
This diff implements `BitwiseLessThan` protocol. It takes two bitwise-shared inputs in $\mathbb{Z}_p$ and returns 1 if and only if the first parameter is less than the second parameter. This is an implementation of "5.3 Bitwise Less-Than" described in this [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf), [13].

## Description
`BitwiseLessThan` takes inputs $[a]\_B = ([a\_1]\_p,...,[a\_l]\_p)$ where $a\_1,...,a\_l ∈ {0,1} ⊆ F\_p$ and $[b]\_B = ([b\_1]\_p,...,[b\_l]\_p)$ where $b\_1,...,b\_l ∈ {0,1} ⊆ F\_p$, then computes $h ∈ {0, 1} := a <\{?} b$ where $h = 1$ iff $a$ is less than $b$.

`BitwiseLessThan` invokes one sub-protocol, `PrefixOr`, and `2l` mult's (`l` = bit length, `l` xor's + `l` mult's).

### DIFF
--- a/src/protocol/boolean/bitwise_lt.rs
+++ b/src/protocol/boolean/bitwise_lt.rs
@@ -1,0 +1,258 @@
+use super::prefix_or::PrefixOr;
+use super::BitOpStep;
+use crate::error::BoxError;
+use crate::ff::BinaryField;
+use crate::protocol::{context::ProtocolContext, mul::SecureMul, RecordId};
+use crate::secret_sharing::Replicated;
+use futures::future::try_join_all;
+use std::iter::{repeat, zip};
+
+/// This is an implementation of Bitwise Less-Than on bitwise-shared numbers.
+///
+/// `BitwiseLessThan` takes inputs `[a]_B = ([a_1]_p,...,[a_l]_p)` where
+/// `a1,...,a_l ∈ {0,1} ⊆ F_p` and `[b]_B = ([b_1]_p,...,[b_l]_p)` where
+/// `b1,...,b_l ∈ {0,1} ⊆ F_p`, then computes `h ∈ {0, 1} <- a <? b` where
+/// `h = 1` iff `a` is less than `b`.
+///
+/// Note that `[a]_B` can be converted to `[a]_p` by `Σ (2^i * a_i), i=0..l`. In
+/// other words, if comparing two integers, the protocol expects inputs to be in
+/// the little-endian; the least-significant byte at the smallest address (0'th
+/// element).
+///
+/// 5.3 Bitwise Less-Than
+/// "Unconditionally Secure Constant-Rounds Multi-party Computation for Equality, Comparison, Bits, and Exponentiation"
+/// I. Damgård et al.
+pub struct BitwiseLessThan<'a, B: BinaryField> {
+    a: &'a [Replicated<B>],
+    b: &'a [Replicated<B>],
+}
+
+impl<'a, B: BinaryField> BitwiseLessThan<'a, B> {
+    #[allow(dead_code)]
+    pub fn new(a: &'a [Replicated<B>], b: &'a [Replicated<B>]) -> Self {
+        debug_assert_eq!(a.len(), b.len(), "Length of the input bits must be equal");
+        Self { a, b }
+    }
+
+    /// Step 1. `for i=0..l-1, [e_i] = XOR([a_i], [b_i])`
+    ///
+    /// We can compute `[a] ⊕ [b], a,b ∈ Z_2` as `[a + b]` locally.
+    ///
+    /// # Example
+    /// ```ignore
+    ///   //  bit-0         bit-7
+    ///   //    v             v
+    ///   [a] = 1 0 1 0 1 0 0 0   // 21 in little-endian
+    ///   [b] = 0 1 1 1 1 0 0 0   // 30 in little-endian
+    ///   [e] = 1 1 0 1 0 0 0 0
+    /// ```
+    fn step1(a: &[Replicated<B>], b: &[Replicated<B>]) -> Vec<Replicated<B>> {
+        zip(a, b)
+            .map(|(&a_bit, &b_bit)| a_bit + b_bit)
+            .collect::<Vec<_>>()
+    }
+
+    /// Step 2. `([f_(l-1)]..[f_0]) = PrefixOr([e_(l-1)]..[e_0])`
+    ///
+    /// We compute `PrefixOr` of [e] in the reverse order. Remember that the
+    /// inputs are in little-endian format. In this step, we try to find the
+    /// smallest `i` (or MSB since `e` is reversed) where `a_i != b_i`. The
+    /// output is in big-endian, note that the ordering of `[f]` in the notation
+    /// above is also reversed as in `([f_(l-1)]..[f_0])`, hence we reverse the
+    /// vector once again before returning.
+    ///
+    /// # Example
+    /// ```ignore
+    ///   //  bit-0         bit-7
+    ///   //    v             v
+    ///   [e] = 1 1 0 1 0 0 0 0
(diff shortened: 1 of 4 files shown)


benjaminsavage (changes requested):
I think that actually all of the secret shares are meant to be in a large field, not binary fields.

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -0,0 +1,258 @@
...
+///
+/// Note that `[a]_B` can be converted to `[a]_p` by `Σ (2^i * a_i), i=0..l`. In
+/// other words, if comparing two integers, the protocol expects inputs to be in
+/// the little-endian; the least-significant byte at the smallest address (0'th
+/// element).
+///
+/// 5.3 Bitwise Less-Than
+/// "Unconditionally Secure Constant-Rounds Multi-party Computation for Equality, Comparison, Bits, and Exponentiation"
+/// I. Damgård et al.
+pub struct BitwiseLessThan<'a, B: BinaryField> {
```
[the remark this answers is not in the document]
taikiy: > As this comment says: the inputs are all elements of a larger prime field: ⊆ F_p. This implementation is assuming a binary field: B: BinaryField.

The paper says `a1,...,a_l ∈ {0,1} ⊆ F_p`. I interpreted this as; `{0, 1} ⊆ F_p` then `{0,1} ∈ Z_2 ⊆ F_p` is also true. Therefore, I thought I could use `BinaryField` to simplify algebraic operations rather than using `Field` with `mod 2` operations. If so, then the code actually matches the comment since `BinaryField ⊆ Field`. Maybe I'm missing to see cases where using `BinaryField` becomes a problem?

> All of this work is only necessary if you're operating in a larger field. Otherwise there are much, much simpler ways to compute "less than".

Yes. we want to check `a_p <? b_p`, and not `[a]_p <? [b]_p` or `[a]_B <? [b]_B`.

Let me clarify for to make sure my understanding is correct. The inputs to `BitwiseLessThan` are bitwise-shared secrets `[a]_B` and `[b]_B`. That is, we have a secret integer `a ⊆ F_p`, and transform it into bits `a_B = (a_0,...a_l) where a_i {0,1} ⊆ Z_2`. Then we XOR secret-share each bit `a_i` to `[a_i]` (equivalent to additive sharing in `Z_2`), which forms `[a]_B = ([a_1],...[a_l])`. Hence, we cannot reconstruct an additive sharing of `[a]_p` from `[a]_B`.

So, this comment was incorrect:

> Note that `[a]_B` can be converted to `[a]_p` by `Σ (2^i * a_i), i=0..l`

It should be:

> Note that `a_B` can be converted to `a_p` by `Σ (2^i * a_i), i=0..l`.

My intention here was to just say that a binary representation of `a` needs to be in little-endian format.

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// ```ignore
+    ///   //  bit-0         bit-7
+    ///   //    v             v
+    ///   [a] = 1 0 1 0 1 0 0 0   // 21 in little-endian
+    ///   [b] = 0 1 1 1 1 0 0 0   // 30 in little-endian
+    ///   [e] = 1 1 0 1 0 0 0 0
+    /// ```
+    fn step1(a: &[Replicated<B>], b: &[Replicated<B>]) -> Vec<Replicated<B>> {
+        zip(a, b)
+            .map(|(&a_bit, &b_bit)| a_bit + b_bit)
```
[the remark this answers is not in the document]
benjaminsavage: I think this will fail in any field other than F2. The paper suggests computing XOR as follows:

```
[d_i]_p = [a_i]_p - [b_i]_p
[e_i]_p = MULT([d_i]_p, [d_i]_p)
```

This logic is currently missing.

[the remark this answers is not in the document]
martinthomson: Don't we already have an XOR protocol?  The modulus reduction does: $[d_i]_p = [a_i]_p + [b_i]_p - 2.MULT([a_i]_p,[b_i]_p)$.

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -0,0 +1,263 @@
...
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp2>(&world);
+        let mut rand = StepRng::new(1, 1);
+
+        let a_bits = fp31_to_bits(a);
+        let b_bits = fp31_to_bits(b);
+
+        // Generate secret shares
+        #[allow(clippy::type_complexity)]
+        let ((a0, (a1, a2)), (b0, (b1, b2))): (BitwiseShares, BitwiseShares) = zip(a_bits, b_bits)
```
benjaminsavage, line 207: This is still testing it in Fp2. Can you modify this to test the bits as shares of 0, 1 in a larger field?

src/protocol/boolean/xor.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct Xor<F: Field> {
+    a: Replicated<F>,
+    b: Replicated<F>,
+}
+
+impl<'a, F: Field> Xor<F> {
+    #[allow(dead_code)]
+    pub fn new(a: Replicated<F>, b: Replicated<F>) -> Self {
+        Self { a, b }
+    }
```
benjaminsavage, line 17: As mentioned above, I see no gain to instantiating this struct. Let's just make it a funtion.

benjaminsavage (changes requested):

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -0,0 +1,257 @@
...
+    use std::iter::zip;
+
+    type BitwiseShares = (
+        Vec<Replicated<Fp31>>,
+        (Vec<Replicated<Fp31>>, Vec<Replicated<Fp31>>),
+    );
+
+    fn fp31_to_bits(x: Fp31) -> Vec<Fp31> {
+        let x = u8::from(x);
+        (0..u8::BITS).map(|i| Fp31::from(x >> i & 1)).collect()
```
benjaminsavage, line 188: It would also be great to test this on `Fp32BitPrime`

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -0,0 +1,257 @@
...
+    async fn bitwise_lt(a: Fp31, b: Fp31) -> Result<Fp31, BoxError> {
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp31>(&world);
+        let mut rand = StepRng::new(1, 1);
+
+        let a_bits = fp31_to_bits(a);
+        let b_bits = fp31_to_bits(b);
+
+        // Generate secret shares
+        #[allow(clippy::type_complexity)]
```
benjaminsavage, line 200: ngl, clippy ain't wrong about this...

I get that you want helper-wise vectors of stuff, and not a single vector of triples of stuff... but this is kind of gnarly.

You could produce a simpler data structure, (e.g. make the `fp31_to_bits()` function just return `Vec<(Replicated<F>, Replicated<F>, Replicated<F>)>`) and then when you want to invoke the BitwiseLessThan::execute() function map this into what it is expecting, i.e.
```
BitwiseLessThan::execute(..., a_bits.map(|bit| bit.0), b_bits.map(|bit| bit.0)),
...
```

benjaminsavage (changes requested):

benjaminsavage (approved):
We can address the outstanding comments in a follow up PR

Issue #118, opened 2022-10-13
s/Identity/Role


martinthomson:
The term "identity" is pretty heavily loaded already, but what we really want this type to represent is the *role* that the helper is assuming in the execution.

This might reflect the fact that a helper might get assigned different roles for different queries.

Besides, it's a shorter string to type.

PR #207, opened 2022-11-08
s/Identity/Role

akoshelev:
while it is not too late. The usage of the word "identity" is growing

Only refactoring, no functional changes. 
addresses #118

### DIFF
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -6,18 +6,22 @@ pub mod fabric;
 pub mod messaging;
 
 use crate::helpers::Direction::{Left, Right};
-use crate::helpers::Identity::{H1, H2, H3};
+use crate::helpers::Role::{H1, H2, H3};
 pub use error::Error;
 pub use error::Result;
 
-/// Represents a unique identity of each helper running MPC computation.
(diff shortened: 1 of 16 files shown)


src/helpers/mod.rs
```diff
@@ -29,21 +33,21 @@ pub enum Direction {
...
     #[must_use]
-    pub fn all_variants() -> &'static [Identity; 3] {
-        static VARIANTS: &[Identity; 3] = &[Identity::H1, Identity::H2, Identity::H3];
+    pub fn all() -> &'static [Role; 3] {
+        static VARIANTS: &[Role; 3] = &[Role::H1, Role::H2, Role::H3];
 
         VARIANTS
     }
 
     /// Returns the identity of a peer that is located at the specified direction
```
richajaindce, line 48: nit: role here also

richajaindce (approved):
Pretty straightforward. Lets make this change!

richajaindce (approved):

src/helpers/mod.rs
akoshelev: good catch!

benjaminsavage (approved):

PR #208, opened 2022-11-08
Validate indices used in PRSS

akoshelev:
Similarly to validating steps while narrowing contexts, this change makes PRSS validate the indices passed on to it when generating randomness.

It should be a mistake to use the same index twice within the same PRSS instance. This change enables PRSS to panic and abort in such cases.

Interestingly it already helped to catch one bug inside the malicious implementation.

This should help with some issues presented in #135 as well. The cost we pay is another HashSet in memory. It can grow up to $O(N)$ where $N$ is the  size of the input. It may be acceptable for debug builds. There is also an optimization we could use to reduce the footprint https://docs.rs/tinyset/latest/tinyset/setu64/struct.SetU64.html

### DIFF
--- a/src/protocol/prss.rs
+++ b/src/protocol/prss.rs
@@ -9,10 +9,12 @@ use hkdf::Hkdf;
 use rand::{CryptoRng, RngCore};
 use sha2::Sha256;
 use std::{
-    collections::{HashMap, HashSet},
-    fmt::{Debug, Formatter},
+    collections::HashMap,
+    fmt::Debug,
     sync::{Arc, Mutex},
 };
+#[cfg(debug_assertions)]
+use std::{collections::HashSet, fmt::Formatter};
 use x25519_dalek::{EphemeralSecret, PublicKey};
 
 use super::UniqueStepId;
@@ -27,6 +29,7 @@ struct UsedSet {
     used: Arc<Mutex<HashSet<usize>>>,
 }
 
+#[cfg(debug_assertions)]
 impl UsedSet {
     fn new(key: String) -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


benjaminsavage (approved):
I love this! And you already caught a bug! That's great.

I mean - this isn't run at all except in a debug build right? I think that should catch logic errors and help us avoid stuff that could pose a real security risk. Big thumbs up from me.

src/protocol/prss.rs
```diff
@@ -35,6 +85,11 @@ impl IndexedSharedRandomness {
     #[must_use]
     pub fn generate_values<I: Into<u128>>(&self, index: I) -> (u128, u128) {
         let index = index.into();
+        #[cfg(debug_assertions)]
+        {
+            self.used.insert(index);
+        }
```
martinthomson, line 91:
```suggestion
        if cfg!(debug_assertions) {
            self.used.insert(index);
        }
```

src/protocol/prss.rs
```diff
@@ -582,4 +640,27 @@ pub mod test {
...
+        indices.shuffle(&mut thread_rng());
+        let indexed_prss = p2.indexed(&step);
+
+        for index in indices {
+            let _ = indexed_prss.random_u128(index);
+        }
+    }
+
+    #[test]
+    #[should_panic]
```
martinthomson, line 658:
```suggestion
    #[cfg(debug_assertions)]
    #[should_panic]
```

martinthomson (approved):
Some small tweaks, but this is good.

(I was going to be smarter, but I realize we don't need that.)

akoshelev:
> Some small tweaks, but this is good.
> 
> (I was going to be smarter, but I realize we don't need that.)

just curious what was your idea (maybe I can reiterate and implement it)?

src/protocol/prss.rs
akoshelev: that does not work for release builds as `used` field is in debug only. we have to take out the whole block for release

martinthomson:
I was only going to track the last 128 values in a `u128` bitfield.  That would have been more work even if it was more efficient. This is better, even if it is slower, perhaps.

PR #209, opened 2022-11-08
Bitwise-represented random values

taikiy:
This is a stacked diff. The commit specific to this PR is 2822c56f09bc4aa181d75b872615f7c9243a894c.

## Overview
Similar to `generate_fields::<F>`, this diff implements `generate_bitwise_fields::<F: Field, B: BinaryField>`, which generates two random field values in base2 (i.e. two vectors of `B` values).

## Description
`generate_bitwise_fields::<F, B>` is generic over `F` and `B` because we transform two random values in `F` to two vectors of binary fields `B`. One of protocols that uses this function is the bit-decomposition. Bit-decomposition transforms $[a]\_p$ to $[a]\_B$. During the process, it uses random bits of the size $\lceil log(p) \rceil$. Here, `p` is the prime of a field `F`, therefore we need `F` to know from which field we are generating the random values.

This is essentially the same as the caller calling `generate_fields` and changing the output values into bits themselves. This is a helper function to do just that.

In Damgård's [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf), this is described as $Ran\_2$ and `SOLVED-BITS` protocols [9-11]. Since we are already using PRSS, we'll just use that to generate random bits.

### DIFF
--- a/src/protocol/prss.rs
+++ b/src/protocol/prss.rs
@@ -1,4 +1,4 @@
-use crate::ff::Field;
+use crate::ff::{BinaryField, Field, Int};
 use crate::secret_sharing::Replicated;
 use aes::{
     cipher::{generic_array::GenericArray, BlockEncrypt, KeyInit},
@@ -8,6 +8,7 @@ use byteorder::{ByteOrder, LittleEndian};
 use hkdf::Hkdf;
 use rand::{CryptoRng, RngCore};
 use sha2::Sha256;
+use std::iter::zip;
 use std::{
     collections::HashMap,
     fmt::Debug,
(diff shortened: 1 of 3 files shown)


src/protocol/prss.rs
```diff
@@ -533,6 +568,42 @@ pub mod test {
...
+        let (r2_l, r2_r) = (b2_to_b10(&rb2_l), b2_to_b10(&rb2_r));
+        let (rb3_l, rb3_r) = p3
+            .indexed(&step)
+            .generate_bitwise_fields::<Fp31, Fp2, u128>(IDX);
+        let (r3_l, r3_r) = (b2_to_b10(&rb3_l), b2_to_b10(&rb3_r));
+
+        // Verify for each party i, l_i == r_(i-1)
+        assert_eq!(r1_l, r3_r);
+        assert_eq!(r2_l, r1_r);
+        assert_eq!(r3_l, r2_r);
```
benjaminsavage, line 604: You're just checking that this is a valid replicated secret sharing - but we should also validate that $a ∈ {0, 1}$

taikiy:
Per discussion offline. I will implement SOLVED-BITS.

PR #210, opened 2022-11-08
Restructuring code in shuffle/applyinv/compose to consume inputs

richajaindce:
Restructuring code in shuffle/applyinv/compose to consume data which is input to them and return the transformed output rather than doing an in-place change. This will help in better code readability.
It is a part of comment on #186 : https://github.com/private-attribution/ipa/pull/186#discussion_r1016252919 "What I really don't like about these 23 lines of code, is that it's not obvious which arguments are updated / modified by these Protocols. "

### DIFF
--- a/src/protocol/sort/apply.rs
+++ b/src/protocol/sort/apply.rs
@@ -14,10 +14,11 @@ use permutation::Permutation;
 /// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
 /// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
 /// ![Apply steps][apply]
-pub fn apply<T: Copy + Default, S>(permutation: &mut Permutation, values: &mut S)
+pub fn apply<T: Copy + Default, S>(permutation: Permutation, values: &mut S)
 where
     S: AsMut<[T]>,
 {
+    let mut permutation = permutation;
     permutation.apply_slice_in_place(values);
 }
 
@@ -25,10 +26,11 @@ where
 /// is moved by `apply_inv` to be the σ(i)-th item. Therefore, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is
 /// reordered into (D, C, A, B).
 /// ![Apply inv steps][apply_inv]
-pub fn apply_inv<T: Copy + Default, S>(permutation: &mut Permutation, values: &mut S)
+pub fn apply_inv<T: Copy + Default, S>(permutation: Permutation, values: &mut S)
 where
     S: AsMut<[T]>,
 {
+    let mut permutation = permutation;
     permutation.apply_inv_slice_in_place(values);
 }
 
@@ -41,15 +43,15 @@ mod tests {
     #[test]
     fn apply_shares() {
         let mut values = ["A", "B", "C", "D"];
-        let mut indices = Permutation::oneline([2, 3, 1, 0]).inverse();
+        let indices = Permutation::oneline([2, 3, 1, 0]).inverse();
         let expected_output_apply = ["C", "D", "B", "A"];
-        apply(&mut indices, &mut values);
+        apply(indices, &mut values);
         assert_eq!(values, expected_output_apply);
 
         let mut values = ["A", "B", "C", "D"];
-        let mut indices = Permutation::oneline([2, 3, 1, 0]).inverse();
+        let indices = Permutation::oneline([2, 3, 1, 0]).inverse();
         let expected_output_apply_inv = ["D", "C", "A", "B"];
-        apply_inv(&mut indices, &mut values);
+        apply_inv(indices, &mut values);
         assert_eq!(values, expected_output_apply_inv);
     }
 
@@ -58,7 +60,7 @@ mod tests {
         let sigma = vec![4, 2, 0, 5, 1, 3];
         let mut rho = vec![3, 4, 0, 5, 1, 2];
         // Applying sigma on rho
-        apply_inv(&mut Permutation::oneline(sigma), &mut rho);
+        apply_inv(Permutation::oneline(sigma), &mut rho);
         assert_eq!(rho, vec![1, 0, 3, 2, 4, 5]);
     }
 
@@ -74,10 +76,10 @@ mod tests {
         let mut values: Vec<_> = (0..batchsize).collect();
         let mut values_copy = values.clone();
 
-        apply(&mut Permutation::oneline(permutation.clone()), &mut values);
+        apply(Permutation::oneline(permutation.clone()), &mut values);
 
         apply_inv(
-            &mut Permutation::oneline(permutation).inverse(),
+            Permutation::oneline(permutation).inverse(),
             &mut values_copy,
         );
 
--- a/src/protocol/sort/compose.rs
+++ b/src/protocol/sort/compose.rs
@@ -1,5 +1,3 @@
-use std::mem::take;
-
 use crate::{
     error::BoxError,
     ff::Field,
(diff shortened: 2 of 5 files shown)


src/protocol/sort/compose.rs
```diff
@@ -44,24 +46,24 @@ impl<'a, F: Field> Compose<'a, F> {
...
         &mut self,
         ctx: ProtocolContext<'_, Replicated<F>, F>,
-    ) -> Result<(), BoxError> {
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
         let mut random_permutations =
             get_two_of_three_random_permutations(self.rho.len(), &ctx.prss());
         let mut random_permutations_copy = random_permutations.clone();
 
-        Shuffle::new(self.sigma)
+        self.sigma = Shuffle::new(take(&mut self.sigma))
```
benjaminsavage, line 54: This is cool. I learned about `std::mem::take()` today!

But if I understand it correctly, this will actually make a deep copy of the vector, and fully zero out the old one. That's a lot of wasted work. 

I think there is a better way:

Let's just make this protocol into a single function invocation. So instead of:
```
Compose::new(sigma, rho).execute(ctx).await?
```

Let's just do:
```
use crate::protocol::sort::compose::compose_permutations;

compose_permutations(ctx, sigma, rho).await?
```

This is the way I did it with `DoubleRandom::execute()` and `check_zero` and `reveal`. I just don't think creating the struct, assigning member variables, and referencing them from the `execute` method is adding anything. If anything, it's creating this added complexity for us where we have to call `std::mem::take()`.

thurstonsand, line 54: the compiler is usually smart enough to figure out if you do
```suggestion
        self.sigma = Shuffle::new(self.sigma)
```

but do you really need to assign it back to `self`? seems like it might be more clear if you do
```suggestion
        let sigma = Shuffle::new(self.sigma)
```
last thing: it's also an option to explicitly extract the values from self:
```rust
let Self { sigma, rho } = self;

File: src/protocol/sort/compose.rs [hunk 2]
```diff
@@ -102,20 +104,20 @@ mod tests {
...
+            let mut compose0 = Compose::new(sigma_shares.0, rho_shares.0);
+            let mut compose1 = Compose::new(sigma_shares.1, rho_shares.1);
+            let mut compose2 = Compose::new(sigma_shares.2, rho_shares.2);
 
             let h0_future = compose0.execute(ctx0);
             let h1_future = compose1.execute(ctx1);
             let h2_future = compose2.execute(ctx2);
 
-            try_join!(h0_future, h1_future, h2_future)?;
+            rho_shares = try_join!(h0_future, h1_future, h2_future)?;
```
thurstonsand, line 120: IMO you shouldn't reuse the same variable unless there's a specific reason to (for instance, if you want to modify a borrowed value). in this case, would it be more clear if you did something like
```suggestion
            let composed_rho_shares = try_join!(h0_future, h1_future, h2_future)?;
```

src/protocol/sort/secureapplyinv.rs
```diff
@@ -43,25 +43,25 @@ impl SecureApplyInv {
...
         )
         .await?;
         let mut permutation =
-            reveal_permutation(ctx.narrow(&RevealPermutation), sort_permutation).await?;
+            reveal_permutation(ctx.narrow(&RevealPermutation), &sort_permutation).await?;
         // The paper expects us to apply an inverse on the inverted Permutation (i.e. apply_inv(permutation.inverse(), input))
         // Since this is same as apply(permutation, input), we are doing that instead to save on compute.
-        apply(&mut permutation, input);
-        Ok(())
+        apply(&mut permutation, &mut shuffled_input);
```
thurstonsand, line 63: is it possible for `apply` to consume `permutation`? such that this would look like
```suggestion
        apply(permutation, &mut shuffled_input);
```
?

src/protocol/sort/shuffle.rs
```diff
@@ -156,21 +158,21 @@ impl<'a, F: Field> Shuffle<'a, F> {
...
             .shuffle_or_unshuffle_once(ShuffleOrUnshuffle::Shuffle, &ctx, Step1, permutations)
             .await?;
-        *self.input = self
+        self.input = self
             .shuffle_or_unshuffle_once(ShuffleOrUnshuffle::Shuffle, &ctx, Step2, permutations)
             .await?;
-        *self.input = self
+        self.input = self
             .shuffle_or_unshuffle_once(ShuffleOrUnshuffle::Shuffle, &ctx, Step3, permutations)
             .await?;
```
thurstonsand, line 174: given you will immediately `take` this input on the next line, maybe just skip setting it to `self`?
```suggestion
        let input = self
            .shuffle_or_unshuffle_once(ShuffleOrUnshuffle::Shuffle, &ctx, Step3, permutations)
            .await?;
        Ok(input)
```
or simply
```suggestion
        Ok(self
            .shuffle_or_unshuffle_once(ShuffleOrUnshuffle::Shuffle, &ctx, Step3, permutations)
            .await?)
```

thurstonsand (changes requested):

src/protocol/sort/secureapplyinv.rs
```diff
@@ -43,25 +43,25 @@ impl SecureApplyInv {
...
-        sort_permutation: &'_ mut Vec<Replicated<F>>,
-    ) -> Result<(), BoxError> {
+        input: Vec<Replicated<F>>,
+        sort_permutation: Vec<Replicated<F>>,
+    ) -> Result<Vec<Replicated<F>>, BoxError> {
         let mut random_permutations =
             get_two_of_three_random_permutations(input.len(), &ctx.prss());
 
-        let (_, _) = try_join(
+        let (mut shuffled_input, sort_permutation) = try_join(
```
benjaminsavage, line 52: I already like this better. Much more clear what each one is doing.

src/protocol/sort/shuffle.rs
```diff
@@ -156,21 +158,21 @@ impl<'a, F: Field> Shuffle<'a, F> {
...
 
-        Ok(())
+        Ok(take(&mut self.input))
```
benjaminsavage, line 175: I'd suggest doing the same thing I suggested above - just eliminate the struct and member variables, and make `shuffle_permutation` just a function that takes ownership over the vector you pass in.

benjaminsavage (changes requested):
I like where this is going! But I think we can eliminate the use of `std::mem::take()`

src/protocol/sort/compose.rs
richajaindce: Ended up removing these variables from members to send as function parameters

src/protocol/sort/shuffle.rs
richajaindce: Shuffle is doing a lot more than just taking this as an input. `shuffle_or_unshuffle_once` modifies the input and needs a mut ref for it. Have changed the other ones for now.

src/protocol/sort/compose.rs
richajaindce: Actually mem::take or mem::swap dont do a deep copy but only copy the contents of the variable (i.e. copy location of vector data and length and, do not copy the actual data stored on the heap): https://stackoverflow.com/questions/39327859/how-to-swap-variables-pointing-to-data-without-swapping-their-content
I have also tried to consume the permutation variable in apply and apply_inv

src/protocol/sort/apply.rs
```diff
@@ -14,21 +14,23 @@ use permutation::Permutation;
 /// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
 /// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
 /// ![Apply steps][apply]
-pub fn apply<T: Copy + Default, S>(permutation: &mut Permutation, values: &mut S)
+pub fn apply<T: Copy + Default, S>(permutation: Permutation, values: &mut S)
 where
     S: AsMut<[T]>,
 {
+    let mut permutation = permutation;
```
richajaindce, line 21: This will consume the permutation received from the caller

src/protocol/sort/compose.rs
```diff
@@ -41,27 +34,27 @@ impl<'a, F: Field> Compose<'a, F> {
...
-        Shuffle::new(self.sigma)
-            .execute(ctx.narrow(&ShuffleSigma), &mut random_permutations)
+        let shuffled_sigma = Shuffle::new(sigma, random_permutations.clone())
+            .execute(ctx.narrow(&ShuffleSigma))
             .await?;
 
-        let mut perms = reveal_permutation(ctx.narrow(&RevealPermutation), self.sigma).await?;
+        let revealed_permutation =
+            reveal_permutation(ctx.narrow(&RevealPermutation), &shuffled_sigma).await?;
+        let mut applied_rho = rho;
```
benjaminsavage, line 50: What does this do?

benjaminsavage (approved):
Let’s do this!

src/protocol/sort/secureapplyinv.rs
richajaindce: Took a stab at doing this

src/protocol/sort/compose.rs
richajaindce: it converts a immutable variable to a mutable one

PR #211, opened 2022-11-09
Small size optimization for message payload

akoshelev:
Given that we have a limit on maximum message size and it is quite small (8 bytes), there is no reason to heap allocate every time a message is sent and kept inside the send buffer. Send buffer is already a contiguous slice of memory, so we can copy data from the stack there. This change brings `tinyvec` crate to IPA and uses the fixed-size array to store message payload 

## Why tinyvec

Basically I picked `tinyvec` over `smallvec` by reading [this](https://www.reddit.com/r/rust/comments/n2429h/arrayvec_vs_smallvec_vs_tinyvec/). Dealing with unsoundness is not fun, so I thought using a 100% safe code is better


src/helpers/buffers/send.rs
```diff
@@ -157,21 +136,21 @@ impl Config {
...
     use crate::protocol::{RecordId, UniqueStepId};
-    use rand::seq::SliceRandom;
-    use rand::thread_rng;
-    use std::cmp::Ordering;
+
+    use tinyvec::array_vec;
 
     use crate::helpers::fabric::{ChannelId, MessageEnvelope};
 
     impl Clone for MessageEnvelope {
```
thurstonsand, line 148: could you not just `#[derive(Clone, Copy)]` for `MessageEnvelope`?

src/net/server/handlers/mul.rs
```diff
@@ -92,33 +93,22 @@ pub async fn handler(
         step,
     };
 
-    let body = hyper::body::to_bytes(req.body_mut()).await?;
-    let envelopes = body
-        .as_ref()
-        .chunks(headers.data_size as usize)
```
thurstonsand, line 98: so we should remove `data_size` no? will still need `offset` to handle out-of-order messages until we get streaming solution

src/protocol/mod.rs
```diff
@@ -243,3 +244,15 @@ impl From<RecordId> for u32 {
...
     }
 }
+
+impl From<RecordId> for usize {
+    fn from(r: RecordId) -> Self {
+        r.0 as usize
+    }
+}
+
+impl AddAssign<usize> for RecordId {
```
thurstonsand, line 254: i saw you're doing `*offset += 1`, but would it be better to `impl AddAssign<u32> for RecordId` and change that line to `*offset += 1u32`?

thurstonsand (changes requested):

src/helpers/buffers/send.rs
akoshelev: I did not want to do that to avoid making it cloneable by accident. It it a mistake to try to clone it. On the other hand, it is convenient to clone it in these unit tests. Note that I could've done this `#[cfg_attr(test, Clone)]`  instead. For now it is only used in one place and maybe it will be gone soon

src/protocol/mod.rs
akoshelev: just curious why it might be better - you need to type more for something that is already convertible to `usize`?

thurstonsand: only to avoid the conversion + `unwrap`. seems like that is unnecessary work

thurstonsand (approved):
this will be a nice optimization win. looks great

PR #212, opened 2022-11-09
Revert "Bitwise PrefixOr protocol"

taikiy:
Reverts private-attribution/ipa#195

Unit test is in some kind of an infinite loop and not finishing.


akoshelev:
don't need to - just apply this fix: https://github.com/private-attribution/ipa/pull/211/files#diff-d6bbb189e78fe83156a19fd5d5f24b4e9caba747df1dc055873043e05f83837d

it is because we pushed windowing change to infra, now we need to be careful picking up record id value

akoshelev:
caused by #183

akoshelev:
it is concerning however that test runs for 4 seconds - that is a lot for a unit test :(

taikiy:
Thanks Alex. Closing this one.

taikiy:
> it is concerning however that test runs for 4 seconds - that is a lot for a unit test :(

I reduced the iterations from 100 to 16. It now runs under a second but has less chance to catch errors. Let's see how that works out.

PR #215, opened 2022-11-10
Network impl

thurstonsand:
renames filenames `fabric.rs` -> `network.rs`

Implements the `Network` trait for HTTP server/client. Still need to hook up the Server and the send event loop, but this exposes the channels needed for that work.


src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+        Poll::Ready(Ok(()))
+    }
+}
+
+/// Http implementation of a [`Network`]. Uses channels for both [`Sink`] and [`Stream`]
+/// implementations.
+/// # Panics
+/// if `recv_stream` or `recv_messages` called more than once.
+#[allow(dead_code)] // TODO: WIP
+pub struct HttpNetwork<'a> {
```
akoshelev, line 71: is it going to stay for streaming model or just temporary to get discrete working?

src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+    peers: &'a [peer::Config; 3],
+    query_id: QueryId,
+    sink_sender: mpsc::Sender<MessageChunks>,
```
akoshelev, line 74: do you plan to resolve the role of this helper somewhere else? I believe currently helper's role is sent in query string

src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+use std::task::{Context, Poll};
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+use tokio_util::sync::{PollSendError, PollSender};
+
+/// Wrapper around a [`PollSender`] to modify the error message to match what the [`Network`] trait
+/// requires. The only error that [`PollSender`] will generate is "channel closed", and thus is the
+/// only error message forwarded from this [`Sink`].
+#[pin_project]
+pub struct HttpMessageSink<T> {
```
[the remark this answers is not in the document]
thurstonsand: yeah. let me take a stab at that

src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct HttpNetwork<'a> {
+    peers: &'a [peer::Config; 3],
+    query_id: QueryId,
+    sink_sender: mpsc::Sender<MessageChunks>,
+    sink_receiver: Arc<Mutex<Option<mpsc::Receiver<MessageChunks>>>>,
+    message_stream_sender: mpsc::Sender<MessageChunks>,
+    message_stream_receiver: Arc<Mutex<Option<mpsc::Receiver<MessageChunks>>>>,
+}
+
+impl<'a, 'b: 'a> HttpNetwork<'a> {
```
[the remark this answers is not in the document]
thurstonsand: oh, yeah this is better

src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[must_use]
+    #[allow(unused)]
+    pub fn new<D: PeerDiscovery>(
+        peer_discovery: &'b D,
+        sink_queue_depth: usize,
+        message_stream_queue_depth: usize,
+        query_id: QueryId,
+    ) -> Self {
+        let (stx, srx) = mpsc::channel(sink_queue_depth);
```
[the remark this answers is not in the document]
thurstonsand: there isn't an inherent reason why we do. I just thought it might eventually be something we might modify with config, for tuning purposes. but i can hard-code it for now and we can reintroduce later if needed

src/net/http_network.rs
thurstonsand: even under streaming model we will need an implementation of the `Network` trait, right? or are you talking about something else?

src/net/http_network.rs
thurstonsand: that's true, and it'll be pulled off the query string and used for `ChannelId` inside `MessageChunks`. is that what you mean?

src/net/http_network.rs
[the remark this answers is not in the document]
thurstonsand: ok, it's in one place now

src/net/http_network.rs
akoshelev: Yea receiver will get the role from query string, I wasn't sure where exactly we are going to add it to the request on the sender side

thurstonsand: the client's `send_messages` fn takes it as arg: https://github.com/private-attribution/ipa/blob/main/src/net/client/mod.rs#L23

if the network is aware what role it is, it can just pass it in there. tho that will have to be added to the `HttpNetwork` struct. i have a PR that implements the send, but I just checked and definitely used the wrong role there. whoops, will fix that: https://github.com/private-attribution/ipa/pull/221/files#diff-08bc7005bd05cd19c26aa4d50474446db8892cd720a69a482a523c2099561bb3R72

src/net/http_network.rs
akoshelev: we would need `Network` implementation for sure. I wonder if this implementation can be used in streaming model. The reason I am asking is because again this looks very similar to `InMemoryEndpoint`. If we need it in the long term (which does not seem to be the case, but lmk if I am wrong), we could invest in keeping one implementation

akoshelev: and also name is a bit suspicious - nothing that is related to HTTP is here

src/net/http_network.rs
thurstonsand: ok, I fixed that PR to use the correct role

src/net/http_network.rs
thurstonsand: if you look at this PR: https://github.com/private-attribution/ipa/pull/221

it starts using the HTTP clients

thurstonsand: to answer your question about moving to Streaming, the network trait itself will change to accommodate the new api that includes `ChannelId`, so I figure at that time we can modify this `HttpNetwork` to adapt to new API, and to handle streams

src/net/http_network.rs
```diff
@@ -0,0 +1,155 @@
...
+            .iter()
+            .map(|peer_conf| {
+                // no https for now
+                MpcHelperClient::new(peer_conf.http.origin.clone())
+            })
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|arr: Vec<_>| {
+                panic!("unexpected change in array length: {}", arr.len())
+            })
```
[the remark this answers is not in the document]
akoshelev: it shouldn't require `Display`. `Debug` is enough and should be derived for `MpcHelperClient`.

src/helpers/error.rs
```diff
@@ -16,6 +20,11 @@ pub enum Error {
         #[source]
         inner: BoxError,
     },
+    #[error("An error occurred while sending data to unknown helper")]
```
akoshelev, line 23: Lets maybe not intrigue the readers of this error message - "unknown helper" sparks interest as I'd love to know what helper was destined to receive a message from us but I won't be able to :)

this problem of actually adding *context* to error messages is a bit tricky to address in Rust. I don't want to solve it in this PR, but we should keep track of it, so maybe create an issue for it.

src/net/http_network.rs
akoshelev: ack, now it makes sense

akoshelev (approved):

src/helpers/error.rs
thurstonsand: haha i certainly don't want to intrigue the readers. here's an issue to add context to this error specifically, and more broadly to have a way to add context: https://github.com/private-attribution/ipa/issues/226

src/net/http_network.rs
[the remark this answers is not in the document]
thurstonsand: yep, that did it

Issue #120, opened 2022-10-13
Replicated probably shouldn't implement Copy


martinthomson:
Though we are currently implementing `Copy` on `Field`, we really don't want to be copying collections of them all over the place.  A single field needs to be sent over channels and all of that business, but we shouldn't be need to send `Replicated` instances anywhere.

I found a couple of places where we relied on having `Copy` implemented, so this might not be trivial to fix, so I'm opening an issue.

PR #216, opened 2022-11-10
Replacing permutations crate with home-rolled code

benjaminsavage:
The permutations crate has a few issues:
1.) Naming convention is reverse of the paper
2.) It *mutates* the permutation being applied, which means we have to clone permutations pretty often (wasteful)
3.) Strange Permutation.online() stuff all over the place.

### DIFF
--- a/src/protocol/sort/apply.rs
+++ b/src/protocol/sort/apply.rs
@@ -21,7 +21,6 @@ pub fn apply<T>(permutation: &[u32], values: &mut [T]) {
                 pos_i = pos_j;
                 pos_j = permutation[pos_i] as usize;
             }
-            permuted.set(i, true);
         }
     }
 }
@@ -41,7 +40,6 @@ pub fn apply_inv<T>(permutation: &[u32], values: &mut [T]) {
                 permuted.set(destination, true);
                 destination = permutation[destination] as usize;
             }
-            permuted.set(i, true);
         }
     }
 }
--- a/src/protocol/sort/apply.rs
+++ b/src/protocol/sort/apply.rs
@@ -14,7 +14,7 @@ pub fn apply<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
     let mut tmp: T = T::default();
 
     for i in 0..permutation.len() {
-        if permuted[i] == false {
+        if !permuted[i] {
             mem::swap(&mut tmp, &mut values[i]);
             let mut pos_i = i;
             let mut pos_j: usize = permutation[pos_i] as usize;
@@ -39,7 +39,7 @@ pub fn apply_inv<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
     let mut tmp: T;
 
     for i in 0..permutation.len() {
-        if permuted[i] == false {
+        if !permuted[i] {
             let mut destination: usize = permutation[i] as usize;
             tmp = values[i];
             while destination != i {
--- a/src/protocol/sort/apply.rs
+++ b/src/protocol/sort/apply.rs
@@ -104,11 +104,11 @@ mod tests {
     }
 
     #[test]
-    fn permutations_a_million_long() {
-        const SUPER_LONG: usize = 16 * 16 * 16 * 16 * 16; // 1,048,576, a bit over a million
+    fn permutations_super_long() {
+        const SUPER_LONG: usize = 16 * 16 * 16 * 16; // 65,536
         let mut original_values = Vec::with_capacity(SUPER_LONG);
         for i in 0..SUPER_LONG {
-            original_values.push(format!("{:#07x}", i));
+            original_values.push(format!("{:#06x}", i));
         }
         let mut permutation: Vec<u32> = (0..SUPER_LONG)
             .map(|i| usize::try_into(i).unwrap())
--- a/src/protocol/sort/apply.rs
+++ b/src/protocol/sort/apply.rs
@@ -1,6 +1,5 @@
 use bitvec::bitvec;
 use embed_doc_image::embed_doc_image;
-use std::mem;
 
 #[embed_doc_image("apply", "images/sort/apply.png")]
 #[embed_doc_image("apply_inv", "images/sort/apply_inv.png")]
@@ -8,23 +7,20 @@ use std::mem;
 /// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
 /// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
 /// ![Apply steps][apply]
-pub fn apply<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
+pub fn apply<T>(permutation: &[u32], values: &mut [T]) {
     debug_assert!(permutation.len() == values.len());
     let mut permuted = bitvec![0; permutation.len()];
-    let mut tmp: T = T::default();
 
     for i in 0..permutation.len() {
         if !permuted[i] {
-            mem::swap(&mut tmp, &mut values[i]);
             let mut pos_i = i;
-            let mut pos_j: usize = permutation[pos_i] as usize;
+            let mut pos_j = permutation[pos_i] as usize;
             while pos_j != i {
-                values[pos_i] = values[pos_j];
+                values.swap(pos_i, pos_j);
+                permuted.set(pos_j, true);
                 pos_i = pos_j;
                 pos_j = permutation[pos_i] as usize;
-                permuted.set(pos_i, true);
             }
-            mem::swap(&mut values[pos_i], &mut tmp);
             permuted.set(i, true);
         }
     }
@@ -34,20 +30,17 @@ pub fn apply<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
 /// is moved by `apply_inv` to be the σ(i)-th item. Therefore, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is
 /// reordered into (D, C, A, B).
 /// ![Apply inv steps][apply_inv]
-pub fn apply_inv<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
+pub fn apply_inv<T>(permutation: &[u32], values: &mut [T]) {
     let mut permuted = bitvec![0; permutation.len()];
-    let mut tmp: T;
 
     for i in 0..permutation.len() {
         if !permuted[i] {
-            let mut destination: usize = permutation[i] as usize;
-            tmp = values[i];
+            let mut destination = permutation[i] as usize;
             while destination != i {
-                mem::swap(&mut tmp, &mut values[destination]);
+                values.swap(i, destination);
                 permuted.set(destination, true);
                 destination = permutation[destination] as usize;
             }
-            mem::swap(&mut values[i], &mut tmp);
             permuted.set(i, true);
         }
     }
@@ -56,6 +49,7 @@ pub fn apply_inv<T: Copy + Default>(permutation: &[u32], values: &mut [T]) {
 #[cfg(test)]
 mod tests {
     use super::{apply, apply_inv};
+    use rand::seq::SliceRandom;
 
     #[test]
     fn apply_just_one_cycle() {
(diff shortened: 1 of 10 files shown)


src/protocol/sort/apply.rs
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
+
 /// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
 /// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
 /// ![Apply steps][apply]
-pub fn apply<T: Copy + Default, S>(mut permutation: Permutation, values: &mut S)
-where
-    S: AsMut<[T]>,
-{
-    permutation.apply_slice_in_place(values);
+pub fn apply<T: Copy + Default>(permutation: &[usize], values: &mut [T]) {
```
akoshelev, line 21: maybe it is worth asserting `permutation.len() == values.len()` (debug_assert)

src/protocol/sort/apply.rs
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
+            let mut pos_i = i;
+            let mut pos_j = permutation[pos_i];
+            while pos_j != i {
+                values[pos_i] = values[pos_j];
+                pos_i = pos_j;
+                pos_j = permutation[pos_i];
+                permuted.set(pos_i, true);
+            }
+            mem::swap(&mut values[pos_i], &mut tmp);
+            permuted.set(i, true);
```
akoshelev, line 37: I am almost certain you don't need it as i th element is in place already and you should never visit it again.

src/protocol/sort/apply.rs
akoshelev, line 21: I think there is a way to require `T: Clone` only. That is probably worth doing because of #120

src/protocol/sort/apply.rs
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
 /// is moved by `apply_inv` to be the σ(i)-th item. Therefore, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is
 /// reordered into (D, C, A, B).
 /// ![Apply inv steps][apply_inv]
-pub fn apply_inv<T: Copy + Default, S>(mut permutation: Permutation, values: &mut S)
-where
-    S: AsMut<[T]>,
-{
-    permutation.apply_inv_slice_in_place(values);
+pub fn apply_inv<T: Copy + Default>(permutation: &[usize], values: &mut [T]) {
+    let mut permuted = bitvec![0; permutation.len()];
```
akoshelev, line 47: After we push this, would be interesting to experiment with either passing this bit vector as an optional argument. We could re-use the same allocation for sequentially applied permutations and pass `None` if permutation is applied in parallel. I can create an issue if you think it is worth doing (I don't really remember how often we need  to apply permutations in parallel)

src/protocol/sort/apply.rs
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
+    let mut tmp: T = T::default();
+
+    for i in 0..permutation.len() {
+        if permuted[i] == false {
+            mem::swap(&mut tmp, &mut values[i]);
+            let mut pos_i = i;
+            let mut pos_j = permutation[pos_i];
+            while pos_j != i {
+                values[pos_i] = values[pos_j];
```
akoshelev, line 31: I think we could use `Vec::swap` here. That along with removing `mem::swap` and using local variable `tmp` will lead to `T: Clone` only

something like this should work
```rust
   for i in 0..permutation.len() {
        if !permuted[i] {
            let tmp = values[i].clone(); // still cheaper than mem::swap
            ...
            while pos_j != i {
                values.swap(pos_i, pos_j); // that does not require T:Copy
                ...
            }
            values[pos_i] = tmp; // destructive move, does not require T:Copy
        }
    }
```

src/protocol/sort/compose.rs
```diff
@@ -36,39 +36,46 @@ impl Compose {
     #[allow(dead_code)]
     pub async fn execute<F: Field>(
         ctx: ProtocolContext<'_, Replicated<F>, F>,
-        sigma: Vec<Replicated<F>>,
-        rho: Vec<Replicated<F>>,
+        sigma: &mut [Replicated<F>],
```
akoshelev, line 39: ❤️

src/protocol/sort/apply.rs
akoshelev: to be more precise:

```rust
pub fn apply_inv<T: Copy + Default>(permutation: &[usize], values: &mut [T], buf: Option<&mut BitSlice>) {
   // if buf is set, re-use it for to track permutations
   // if it is not set, allocate and proceed
}

File: src/protocol/sort/apply.rs [hunk 6]
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
+            tmp = values[i];
+            while destination != i {
+                mem::swap(&mut tmp, &mut values[destination]);
+                permuted.set(destination, true);
+                destination = permutation[destination];
+            }
+            mem::swap(&mut values[i], &mut tmp);
+            permuted.set(i, true);
+        }
+    }
```
akoshelev, line 62: I stopped here

akoshelev (commented):
only partial review, looks pretty solid. I wonder if you want me to help you write some property based tests, I think testing permutations is a good use case for it

src/protocol/sort/secureapplyinv.rs
```diff
@@ -42,35 +42,45 @@ impl SecureApplyInv {
...
+            ),
         )
         .await?;
         let revealed_permutation =
             reveal_permutation(ctx.narrow(&RevealPermutation), &shuffled_sort_permutation).await?;
         // The paper expects us to apply an inverse on the inverted Permutation (i.e. apply_inv(permutation.inverse(), input))
         // Since this is same as apply(permutation, input), we are doing that instead to save on compute.
-        apply(revealed_permutation, &mut shuffled_input);
+        apply_inv(&revealed_permutation, &mut shuffled_input);
         Ok(shuffled_input)
```
richajaindce, line 71: Since we are taking a mutable ref of input, we will have more memory taken at the given time (input, sort_permutation, shuffled_input and shuffled_sort_permutation are all in memory when this function is executed). Even after this function has returned, input and shuffled_input are taking memory (since input is not consumed by the caller). Dont think this is ideal

src/protocol/sort/shuffle.rs
```diff
@@ -74,127 +67,147 @@ pub fn get_two_of_three_random_permutations(
...
+        &mut once_shuffled,
+        permutation_left,
+        permutation_right,
+        ShuffleOrUnshuffle::Shuffle,
+        &ctx,
+        Step2,
+    )
+    .await?;
+    shuffle_or_unshuffle_once(
+        &mut twice_shuffled,
```
richajaindce, line 164: At this point, we will have input, once_shuffled and twice_shuffled in the memory since none of them is dropped.

richajaindce (changes requested):

src/test_fixture/sharing.rs
```diff
@@ -42,10 +42,14 @@ pub fn validate_and_reconstruct<F: Field>(
...
-            validate_and_reconstruct((result.0[i], result.1[i], result.2[i])),
-            F::from(expected_result[i])
-        );
-    });
+    let revealed_values: Vec<F> = (0..result.0.len())
+        .map(|i| validate_and_reconstruct((result.0[i], result.1[i], result.2[i])))
+        .collect();
+
+    println!("expected results: {:#?}", expected_result);
+    println!("revealed values: {:#?}", revealed_values);
```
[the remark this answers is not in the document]
benjaminsavage: Oh whoops.. I forgot to remove this debugging code!

src/protocol/sort/apply.rs
```diff
@@ -11,77 +12,121 @@ use permutation::Permutation;
...
-where
-    S: AsMut<[T]>,
-{
-    permutation.apply_slice_in_place(values);
+pub fn apply<T: Copy + Default>(permutation: &[usize], values: &mut [T]) {
+    let mut permuted = bitvec![0; permutation.len()];
+    let mut tmp: T = T::default();
+
+    for i in 0..permutation.len() {
+        if permuted[i] == false {
```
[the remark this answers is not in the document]
benjaminsavage: Yeah... I suppressed clippy to get something out for review. I'll turn him back on =)

src/protocol/sort/apply.rs
benjaminsavage: I don't know... that's a lot of extra work. Checking that kind of stuff will make our debug builds run super slow (they already are...)

benjaminsavage: I'll give it a try.

src/protocol/sort/apply.rs
benjaminsavage: yup. This is one of the reasons I went with this refactor. I suspect we can re-use one buffer for this. As you say - next diff.

src/protocol/sort/apply.rs
benjaminsavage: I'll give it a try!

src/protocol/sort/shuffle.rs
```diff
@@ -74,127 +67,147 @@ pub fn get_two_of_three_random_permutations(
...
+/// Unshuffle calls `shuffle_or_unshuffle_once` three times with 2 helpers shuffling the shares each time in the opposite order to shuffle.
+/// Order of calling `shuffle_or_unshuffle_once` is shuffle with (H1, H2), (H3, H1) and (H2, H3)
+/// ![Unshuffle steps][unshuffle]
+pub async fn unshuffle_shares<F: Field>(
+    input: &mut [Replicated<F>],
+    permutation_left: &[usize],
+    permutation_right: &[usize],
+    ctx: ProtocolContext<'_, Replicated<F>, F>,
+) -> Result<Vec<Replicated<F>>, BoxError> {
+    let mut once_shuffled = shuffle_or_unshuffle_once(
```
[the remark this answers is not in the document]
benjaminsavage: good catch

src/protocol/sort/secureapplyinv.rs
benjaminsavage: I see what you're saying. Previously, the memory for these things was de-allocated sooner.

`input` and `shuffled_sort_permutation` were previously consumed on line 51, and need not occupy memory from that point onwards. In this version that's no longer the case. They remain in memory until the end of the calling function, when `input` is no longer referenced. 

That's a good point. I guess I didn't really need to mess with these did I. I really just wanted to eliminate the `clone` on line 51. Let me see if I can bring them back.

src/protocol/sort/shuffle.rs
benjaminsavage: That's true - and it's not ideal. Let me see if I can improve it.

src/protocol/sort/shuffle.rs
```diff
@@ -238,9 +253,9 @@ mod tests {
         assert_ne!(perm2.0, perm2.1);
         assert_ne!(perm3.0, perm3.1);
 
-        assert!(Permutation::valid(&perm1.0));
-        assert!(Permutation::valid(&perm2.0));
-        assert!(Permutation::valid(&perm3.0));
+        assert!(permutation_valid(&perm1.0));
```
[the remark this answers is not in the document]
benjaminsavage: I'm loathe to add that to actual production code - happy to keep it in test code.

src/protocol/sort/apply.rs
benjaminsavage: This is true. As an optimization I'll remove this. It just felt good to leave the `permuted` vector all ones at the end. This will leave it with a few zeroes in there.. which sort of feels yucky.. but there's no benefit to doing this. extra work so why keep it?

src/protocol/sort/apply.rs
```diff
@@ -1,87 +1,141 @@
...
+
 /// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
 /// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
 /// ![Apply steps][apply]
-pub fn apply<T: Copy + Default, S>(mut permutation: Permutation, values: &mut S)
-where
-    S: AsMut<[T]>,
-{
-    permutation.apply_slice_in_place(values);
+pub fn apply<T>(permutation: &[u32], values: &mut [T]) {
```
benjaminsavage, line 10: Check it out @akoshelev - I was able to remove ALL the bounds on this trait! No need for even Clone or Default!

src/protocol/sort/apply.rs
```diff
@@ -1,87 +1,141 @@
...
+        let expected_output_apply = ["B", "A", "C", "J", "K", "D", "E", "F", "G", "H", "I"];
+        apply_inv(&permutation, &mut values);
+        assert_eq!(values, expected_output_apply);
     }
 
     #[test]
-    pub fn apply_apply_inv_relation() {
-        // This test shows that apply(permutation, values) is same as apply_inv(permutation.inverse(), values)
-        let batchsize: usize = 100;
+    fn permutations_a_million_long() {
```
akoshelev, line 107: I don't know...do you really need to sort a billion here? this test validates for correctness, so maybe a million would do just fine (16 ms instead of 2 secs) :)

akoshelev (approved):

PR #220, opened 2022-11-11
Bitwise Sum protocol

taikiy:
This is a stacked diff. The commit needs to be reviewed is eb2777f4e8f236548643897d059403bd6963496b.

## Overview
This diff implements `BitwiseSum`, which takes two bitwise shared inputs in $\mathbb{Z}\_p$ and outputs a bitwise shared number of the sum. This is an implementation of "6.2 Bitwise Sum" described in this [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf).

## Description
This protocol is useful for adding bitwise shared numbers, but is also a part of the bit-decomposition protocol.
screenshot_2022-11-11_14-32-14

PRE∨ (PrefixOr) #195 
BIT-LT (BitwiseLessThan) #206 
SOLVED-BITS (random bit-shares) #209 
CARRIES (Carries) #219 
BIT-ADD (BitwiseSum) **this PR**

### DIFF
--- a/src/protocol/boolean/bitwise_sum.rs
+++ b/src/protocol/boolean/bitwise_sum.rs
@@ -46,7 +46,7 @@ impl BitwiseSum {
         // The paper refers `[c]_b` as `([c_1],...[c_l])`; the starting index is 1;
         // therefore, `[c_1]` is the first element, `c[0]`, in the code.
         let mut d = Vec::with_capacity(l);
-        d.push(a[0] + b[0] - (c[0] * F::from(2)));
+        d.push(a[0].clone() + &b[0] - &(c[0].clone() * F::from(2)));
 
         // Step 3. `[d_l] = [c_l]`
         // Step 4. for `i=1..l-1`, `[d_i] = [a_i] + [b_i] + [c_i] - 2[c_(i+1)]`
@@ -56,9 +56,9 @@ impl BitwiseSum {
         // 32-bit long field, the output of this protocol will be 33-bit long
         // bitwise shares.
         for i in 1..l {
-            d.push(a[i] + b[i] + c[i - 1] - (c[i] * F::from(2)));
+            d.push(a[i].clone() + &b[i] + &c[i - 1] - &(c[i].clone() * F::from(2)));
         }
-        d.push(c[l - 1]);
+        d.push(c[l - 1].clone());
 
         Ok(d)
     }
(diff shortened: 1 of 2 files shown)


src/protocol/boolean/bitwise_sum.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    #[allow(clippy::many_single_char_names)]
+    async fn bitwise_sum<F: Field>(a: F, b: F) -> Result<Vec<F>, BoxError>
+    where
+        Standard: Distribution<F>,
+    {
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<F>(&world);
+        let mut rand = StepRng::new(1, 1);
```
benjaminsavage, line 132: Creating the world and contexts within this function means you're doing so once for every test. That's a lot of times. Maybe move this out and pass them in?

src/protocol/boolean/bitwise_sum.rs
```diff
@@ -0,0 +1,212 @@
...
+            bits_to_field(&bitwise_sum(c(2_147_483_647), c(1)).await?)
+        );
+        assert_eq!(
+            c(4_294_967_290),
+            bits_to_field(&bitwise_sum(c(2_147_483_645), c(2_147_483_645)).await?)
+        );
+        assert_eq!(
+            c(0),
+            bits_to_field(&bitwise_sum(c(2_147_483_645), c(2_147_483_646)).await?)
+        );
```
benjaminsavage, line 208: This one seems like a place where rather than testing specific, hard-coded examples, we could easily just generate two random numbers, compute what the sum should be, and validate that it all worked correctly. That would get us test coverage of many more corner-cases.

benjaminsavage (changes requested):
Just a few suggestions for improving the tests.

src/protocol/boolean/bitwise_sum.rs
taikiy: If I move out `ctx`, which is an array of 3 context instances, out, I will need to `narrow()` them each time I call `bitwise_sum` helper function. I think the cost is almost the same, and easier to read this way.

benjaminsavage: I don’t think those compare in cost at all. Narrowing the context makes a single struct. Creating the world takes a long time because of the key exchange to set up the PRSS

benjaminsavage (approved):
Lots of follow up work to do, but let’s split that out into other PRs

PR #222, opened 2022-11-11
Shared map between MPC server and `HttpNetwork`

thurstonsand:
when a new query is created, there must be a way for the MPC server to send data to the protocols running that query. It does this by inserting a "query_id -> channel send" into a map shared between the server and the query creator.

When a request arrives on the server, it first grabs the query_id from the request, then gets the paired channel from the shared map. That way, the server sends the data to the correct place.

This PR implements the server side/read side of the shared map. a follow up PR will implement the query creator/write side of the shared map.


src/net/server/handlers/query.rs
```diff
@@ -64,13 +64,22 @@ impl<T> Clone for ReservedPermit<T> {
...
-pub async fn obtain_permit_mw<T: Send + 'static, B>(
-    sender: mpsc::Sender<T>,
-    mut req: Request<B>,
+pub async fn obtain_permit_mw<B: Send>(
+    message_send_map: MessageSendMap,
+    req: Request<B>,
     next: Next<B>,
 ) -> Result<Response, MpcHelperServerError> {
+    let mut req_parts = RequestParts::new(req);
+    let Path(query_id, _) = Path::from_request(&mut req_parts).await?;
```
akoshelev, line 73: it doesn't seem right to parse the entire path, drop one argument on the floor in middleware and then parse it again in handler only to pick up the dropped argument. 

I guess we need to live with it as axum does not offer anything more convenient that this? Would it work if we inject path as extension here?

src/net/server/mod.rs
```diff
@@ -95,6 +109,68 @@ impl IntoResponse for MpcHelperServerError {
...
 }
 
+/// Provides a mapping of [`QueryId`]s to senders that forward data to a [`Network`]. Every time a
+/// new query is started, a [`Network`] is created to handle communication for that query. When the
+/// server receives a request, it must know which [`Network`] to forward that request to, so it
+/// holds this mapping to accomplish that.
+///
+/// Is shareable by `clone()`ing.
+#[derive(Clone)]
+pub struct MessageSendMap {
```
akoshelev, line 119: I thought it is not stacked on 225. Is it a different `MessageSendMap` implementation?

src/net/server/handlers/query.rs
thurstonsand: yeah, i could inject path as an extension. then at least it's only cloned a couple times...

src/net/server/mod.rs
thurstonsand: honestly, this might have been a mistake on my part. i didn't intend for this to be stacked. Let me open a PR with just the stacked part of the diff

thurstonsand:
closed since this work was merged into https://github.com/private-attribution/ipa/pull/225

PR #223, opened 2022-11-12
Reveal trait and its implementations for `ProtocolContext`

akoshelev:
Another step to make sort protocol support both semi-honest and malicious security.

In semi-honest setting, `CheckZero` and `ConvertShares` will use semi-honest reveal with a cost of 1 multiplication. Malicious reveal currently has a cost of 2 multiplications and will be used only for malicious setting.

This change leverages recently stabilized [GAT](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html). ~If we like this implementation, `SecureMul` will be changed accordingly.~ I tried it and I don't think it works for multiplication

### DIFF
--- a/src/test_fixture/sharing.rs
+++ b/src/test_fixture/sharing.rs
@@ -1,5 +1,5 @@
 use crate::ff::Field;
-use crate::secret_sharing::Replicated;
+use crate::secret_sharing::{MaliciousReplicated, Replicated};
 use rand::Rng;
 use rand::RngCore;
 
@@ -16,6 +16,24 @@ pub fn share<F: Field, R: RngCore>(input: F, rng: &mut R) -> [Replicated<F>; 3] 
         Replicated::new(x2, x3),
         Replicated::new(x3, x1),
     ]
+}
+
+/// Shares `input` into 3 maliciously secure replicated secret shares using the provided `rng` implementation
+///
+#[allow(clippy::missing_panics_doc)]
+pub fn share_malicious<F: Field, R: RngCore>(
+    input: F,
+    r: [Replicated<F>; 3],
+    rng: &mut R,
+) -> [MaliciousReplicated<F>; 3] {
+    share(input, rng)
+        .iter()
+        .enumerate()
+        .map(|(i, share)| MaliciousReplicated::new(*share, r[i]))
+        // TODO: each_ref when stable
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap()
 }
 
 /// Validates correctness of the secret sharing scheme.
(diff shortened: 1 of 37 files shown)


src/test_fixture/sharing.rs
```diff
@@ -18,6 +18,24 @@ pub fn share<F: Field, R: RngCore>(input: F, rng: &mut R) -> [Replicated<F>; 3]
...
+#[allow(clippy::missing_panics_doc)]
+pub fn share_malicious<F: Field, R: RngCore>(
+    input: F,
+    r: [Replicated<F>; 3],
+    rng: &mut R,
+) -> [MaliciousReplicated<F>; 3] {
+    share(input, rng)
+        .iter()
+        .enumerate()
+        .map(|(i, share)| MaliciousReplicated::new(*share, r[i]))
```
benjaminsavage, line 32: I don’t think this is right. The second argument should be a sharing of input times r. This appears be be just r.

benjaminsavage (approved):
Looks great. Just one conflict and the issue with the test fixture (which works ATM because you’re not really using any properties of the malicious replicated secret sharing in these particular tests, but which should still be resolved)

src/test_fixture/sharing.rs
martinthomson: I think that you want:

```rust
let r = validate_and_reconstruct((r[0], r[1], r[2]));
let rx = r * input;
zip(share(input, rng), share(rx, rng))
  .map(|(x, rx)| MaliciousReplicated::new(*x, *rx))
  .collect::<Vec<_>>()
  .try_into()
  .unwrap()
```

akoshelev: yep that was a bug. I decided to generate `r` inside `share_malicious` given that I have access to PRNG there. Lmk if that is not ergonomic (we could always reconstruct $r$ from $[rx]$)

src/test_fixture/sharing.rs
```diff
@@ -18,6 +18,21 @@ pub fn share<F: Field, R: RngCore>(input: F, rng: &mut R) -> [Replicated<F>; 3]
...
+/// Shares `input` into 3 maliciously secure replicated secret shares using the provided `rng` implementation
+///
+#[allow(clippy::missing_panics_doc)]
+pub fn share_malicious<F: Field, R: RngCore>(x: F, rng: &mut R) -> [MaliciousReplicated<F>; 3] {
+    let rx = F::from(rng.gen::<u128>()) * x;
+    share(x, rng)
+        .iter()
+        .zip(share(rx, rng))
+        .map(|(x, rx)| MaliciousReplicated::new(*x, rx))
+        // TODO: each_ref when stable
```
martinthomson, line 30: Does each_ref() work with zip?

akoshelev: it should work if `array::zip` is stable https://github.com/rust-lang/rust/issues/80094. I'll mention both in the todo

akoshelev:
sorry had to force push as I screwed up during merge and it is quite painful to do it again :(

martinthomson (approved):

PR #224, opened 2022-11-14
Fix PrefixOr to work with F_p

taikiy:
## Overview
Damgård's [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf) describes the protocol in F_p, but my initial implementation used `BinaryField` to simplify boolean arithmetics. This, however, becomes a problem when we migrate to malicious because the chance of MACs not matching in Fp2 is 50%. This diff updates `PrefixOr` protocol to be generic over `F: Field`; a larger field than `BinaryField`.

## Description
This change affects one arithmetic logic where we calculate OR of two bits. (lines 49-55)

$OR([a], [b]) = ! ( ![a] * ![b] )$

### DIFF
--- a/src/protocol/boolean/prefix_or.rs
+++ b/src/protocol/boolean/prefix_or.rs
@@ -1,5 +1,5 @@
 use crate::error::BoxError;
-use crate::ff::BinaryField;
+use crate::ff::Field;
 use crate::protocol::{context::ProtocolContext, mul::SecureMul, RecordId};
 use crate::secret_sharing::Replicated;
 use futures::future::try_join_all;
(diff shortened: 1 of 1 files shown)


src/protocol/boolean/prefix_or.rs
```diff
@@ -21,35 +21,47 @@ use super::BitOpStep;
...
-        let a_and_b = ctx.multiply(record_id, a, b).await?;
-        Ok(a + b + a_and_b)
+    ) -> Result<Replicated<F>, BoxError> {
+        let ab = ctx.narrow(&Step::AMultB).multiply(record_id, a, b).await?;
+        let c = a + b - (ab * F::from(2));
+        let cab = ctx
+            .narrow(&Step::ABMultC)
+            .multiply(record_id, c, ab)
+            .await?;
+        Ok(c + ab - (cab * F::from(2)))
```
benjaminsavage, line 55: This seems like an inefficient way of computing $OR([a], [b])$. A more efficient way immediately comes to mind:

$1- MULT([1-a], [1-b])$

But ignoring that for a moment, the paper says: "Note that we can compute an Or with unbounded fan-in:

$[x]_p$ <- $\bigvee [x_j]_p$ using section 5.1 as this is a symmetric function. 

I'm not seeing this in the code right now.

src/protocol/boolean/prefix_or.rs
```diff
@@ -21,35 +21,47 @@ use super::BitOpStep;
...
+        a: &[Replicated<F>],
         k: usize,
-        ctx: ProtocolContext<'_, Replicated<B>, B>,
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
         record_id: RecordId,
-    ) -> Result<Replicated<B>, BoxError> {
+    ) -> Result<Replicated<F>, BoxError> {
         #[allow(clippy::cast_possible_truncation)]
         let mut v = a[0];
         for (i, &bit) in a[1..].iter().enumerate() {
```
benjaminsavage, line 67: This is the part of the code that ought to use the "unbounded fan-in". This is a "symmetric function" because the output only depends on the number of 1s. In this case:

0 ones => 0
1 or more ones => 1

So we can compute the number of ones by just adding up all the secret shares of the bits. Great. Next, we need a function which has the following values:
1 => 0
2 => 1
3 => 1
4 => 1
5 => 1
6 => 1

(this is assuming the u32 case... it'll be slightly different otherwise).

Now we need to use Lagrange interpolation to find a polynomial that goes through all of these points. Once we have the coefficients of that function, we can just evaluate it.

benjaminsavage (changes requested):

src/protocol/boolean/prefix_or.rs
taikiy: Per discussion offline, let me come back to this "symmetric function" part. I'll check and see which one (bit-by-bit or or an optimized symmetric function) performs better, but this optimization can be done later.

src/protocol/boolean/prefix_or.rs
```diff
@@ -326,31 +321,65 @@ impl AsRef<str> for Step {
...
+            })
+            .unzip();
+
+        // Execute
+        let pre0 = PrefixOr::new(&s0);
+        let pre1 = PrefixOr::new(&s1);
+        let pre2 = PrefixOr::new(&s2);
+        let step = "PrefixOr_Test";
+        let result = try_join_all(vec![
+            pre0.execute(ctx[0].narrow(step), RecordId::from(0_u32)),
```
benjaminsavage, line 360: I’d just in-line the PrefixOr::new() calls on drop 3 lines of code.

src/protocol/boolean/prefix_or.rs
```diff
@@ -361,40 +390,46 @@ mod tests {
...
+            let len = BITS;
+            // Generate a vector of Fp31::ZERO or Fp31::ONE from randomly picked bool values
+            let input: Vec<Fp31> = (0..len)
+                .map(|_| Fp31::from(u128::from(rng.gen::<bool>())))
+                .collect();
+            let mut expected: Vec<Fp31> = Vec::with_capacity(len);
+
+            // Calculate Prefix-Or of the secret number
+            input.iter().fold(0, |acc, &x| {
+                let sum = acc + x.as_u128();
```
benjaminsavage, line 418: This will wrap around if you attempt this on a sufficiently long number. For example, a 64 bit number will (with high probability) have over 31 ones.

src/protocol/boolean/prefix_or.rs
```diff
@@ -326,31 +321,62 @@ impl AsRef<str> for Step {
...
+    #[tokio::test]
+    /// Test PrefixOr with the input ⊆ F_2
+    pub async fn fp2() -> Result<(), BoxError> {
         let mut rng = rand::thread_rng();
 
         // Test 32-bit bitwise shares with randomly distributed bits, for 16 times.
         // The probability of i'th bit being 0 is 1/2^i, so this test covers inputs
         // that have all 0's in 5 first bits.
-        for i in 0..TEST_TRIES {
+        for _ in 0..TEST_TRIES {
```
benjaminsavage, line 379: Please ensure we test all supported bit lengths

benjaminsavage (approved):
Awesome

PR #225, opened 2022-11-14
Enforce ordered messages on server

thurstonsand:
There is a problem with implementing chunked model where there's nowhere that enforces those chunks arrive in-order. Given the situation:

* Helper 1 starts sending a stream of data to Helper 2
* Helper 1 sends chunk 0..10 to Helper 2
* Helper 1 sends chunk 11..20 to Helper 2
* Due to uncontrollable scheduling and/or network behavior, chunk 11..20 arrives to Helper 2 first
* Helper 2 interprets chunk 11..20 as the first chunk it receives, incorrectly marking it as chunk 0..10

This will all run to completion without error, but the results will be meaningless. Thus, we need a way to enforce ordering of the chunks.

This PR has the server maintain last seen offset in the network layer, and if an incoming chunk does not match the last seen offset, it will panic. Technically, a fully correct solution would have the server cache the out-of-order chunk until the in-order one arrives, but as is explained below, there is an overall better solution to the problem. So just to ensure correctness (and panicking otherwise), this PR will act as a stopgap until we have the actual solution implemented.

Importantly, this is a problem only because we are using a chunked model. We plan to shortly move to a streaming model, where the entire stream from Helper 1 will be passed to Helper 2 within 1 http body. When this implementation is done, the ordering will be handled completely by http, and the logic in this PR can be removed from the server.


src/net/server/mod.rs
```diff
@@ -71,22 +77,146 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
 
+/// Provides a mapping of [`QueryId`]s to senders that forward data to a [`Network`]. Every time a
+/// new query is started, a [`Network`] is created to handle communication for that query. When the
+/// server receives a request, it must know which [`Network`] to forward that request to, so it
+/// holds this mapping to accomplish that.
+///
+/// Is shareable by `clone()`ing.
+#[derive(Clone)]
+pub struct MessageSendMap {
+    m: Arc<Mutex<HashMap<QueryId, mpsc::Sender<MessageChunks>>>>,
```
martinthomson, line 129:
```suggestion
    map: Arc<Mutex<HashMap<QueryId, mpsc::Sender<MessageChunks>>>>,
```

src/net/server/mod.rs
```diff
@@ -71,22 +77,146 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
+    }
+}
+
+/// Keeps track of the last seen message for every [`ChannelId`]. This enables the server to ensure
+/// that messages arrive in-order. This solution is a temporary one intended to be removed once
+/// messages arrive in one stream, where ordering will be handled by http.
+/// TODO (ts): remove this when streaming solution is complete
+#[derive(Clone)]
+pub(crate) struct LastSeenMessages {
+    m: Arc<Mutex<HashMap<ChannelId, u32>>>,
```
martinthomson, line 189:
```suggestion
    messages: Arc<Mutex<HashMap<ChannelId, u32>>>,
```

martinthomson (commented):
I'm not clear on why doing streaming is harder than something like this.

thurstonsand:
I realize this is less than ideal, but we were thinking about about the time it would take to try to coordinate the change required for streaming model between the infra and network layers.

since the infra is already setup to handle this request chunking model, it takes less planning to build the network layer to match. We already have planning in place to immediately move to streaming model after this is done, so this code shouldn't need to live for very long.

src/net/server/mod.rs
```diff
@@ -77,24 +81,110 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
 }
 
+/// Provides a mapping of [`QueryId`]s to senders that forward data to a [`Network`]. Every time a
+/// new query is started, a [`Network`] is created to handle communication for that query. When the
+/// server receives a request, it must know which [`Network`] to forward that request to, so it
+/// holds this mapping to accomplish that.
+///
+/// Is shareable by `clone()`ing.
+#[derive(Clone)]
+pub struct MessageSendMap {
```
akoshelev, line 133: this implementation needs some tests as it is here for the long term

src/net/server/mod.rs
```diff
@@ -77,24 +81,110 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
+    senders: Arc<Mutex<HashMap<QueryId, mpsc::Sender<MessageChunks>>>>,
+}
+
+impl MessageSendMap {
+    /// returns the sender for a given query
+    /// # Errors
+    /// if sender does not exist
+    /// # Panics
+    /// if lock is already held by current thread
```
akoshelev, line 142: It won't panic because of that. `unwrap` on mutex panics if it is poisoned: https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning

src/net/server/mod.rs
```diff
@@ -77,24 +81,110 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
+        &self,
+        query_id: QueryId,
+    ) -> Result<mpsc::Sender<MessageChunks>, MpcHelperServerError> {
+        self.senders.lock().unwrap().get(&query_id).map_or_else(
+            || Err(MpcHelperServerError::query_id_not_found(query_id)),
+            |sender| Ok(sender.clone()),
+        )
+    }
+
+    /// adds a sender for a given query
```
akoshelev, line 153: I think it should be an error if I attempt to add sender for the same query twice?

src/net/server/mod.rs
```diff
@@ -77,24 +81,110 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
 
+/// Provides a mapping of [`QueryId`]s to senders that forward data to a [`Network`]. Every time a
+/// new query is started, a [`Network`] is created to handle communication for that query. When the
+/// server receives a request, it must know which [`Network`] to forward that request to, so it
+/// holds this mapping to accomplish that.
+///
+/// Is shareable by `clone()`ing.
+#[derive(Clone)]
+pub struct MessageSendMap {
+    senders: Arc<Mutex<HashMap<QueryId, mpsc::Sender<MessageChunks>>>>,
```
akoshelev, line 134: is there a way we can store `T: Network` instead? Do you need to share network between this struct and anything else? `Network` is sync, so any reference to it will be `Send` and therefore thread safe

src/net/server/mod.rs
```diff
@@ -77,24 +81,110 @@ impl From<axum::http::header::ToStrError> for MpcHelperServerError {
...
+    /// initialize with a [`QueryId`] already inserted.
+    /// Intended to be used only in tests
+    #[cfg(test)]
+    #[must_use]
+    pub fn filled(tx: mpsc::Sender<MessageChunks>) -> MessageSendMap {
+        let mut map = HashMap::new();
+        map.insert(QueryId, tx);
+        MessageSendMap {
+            senders: Arc::new(Mutex::new(map)),
+        }
```
akoshelev, line 176:
```suggestion
       MessageSendMap {
            senders: Arc::new(Mutex::new(HashMap::from([(QueryId, tx)]))),
        }
```

src/net/server/handlers/query.rs
```diff
@@ -64,13 +65,36 @@ impl<T> Clone for ReservedPermit<T> {
...
-    sender: mpsc::Sender<T>,
-    mut req: Request<B>,
+/// # Panics
+/// if messages arrive out of order
+pub async fn obtain_permit_mw<B: Send>(
+    req: Request<B>,
     next: Next<B>,
 ) -> Result<Response, MpcHelperServerError> {
+    let mut req_parts = RequestParts::new(req);
+    let Path(query_id, step) = Path::from_request(&mut req_parts).await?;
```
akoshelev, line 75: This seems to do exactly what Axum extractors do. In other words

```rust
pub async fn obtain_permit_mw<B: Send>(
    Path(query_id, step): Path(QueryId, Step),
)
```

is there a reason you want to parse everything inside the handler?

src/net/server/handlers/query.rs
```diff
@@ -64,13 +65,36 @@ impl<T> Clone for ReservedPermit<T> {
...
+    let Query(RoleQueryParam { role }) =
+        Query::<RoleQueryParam>::from_request(&mut req_parts).await?;
+    let record_headers = RecordHeaders::from_request(&mut req_parts).await?;
+    let Extension(last_seen_messages) =
+        Extension::<LastSeenMessages>::from_request(&mut req_parts).await?;
+    // PANIC if messages arrive out of order
+    // TODO (ts): remove this when streaming solution is complete
+    last_seen_messages
+        .update_in_place(ChannelId::new(role, step), record_headers.offset)
+        .unwrap();
```
akoshelev, line 85: why `unwrap` instead of `?`

src/net/server/handlers/query.rs
```diff
@@ -64,13 +65,36 @@ impl<T> Clone for ReservedPermit<T> {
...
     next: Next<B>,
 ) -> Result<Response, MpcHelperServerError> {
+    let mut req_parts = RequestParts::new(req);
+    let Path(query_id, step) = Path::from_request(&mut req_parts).await?;
+    let Query(RoleQueryParam { role }) =
+        Query::<RoleQueryParam>::from_request(&mut req_parts).await?;
+    let record_headers = RecordHeaders::from_request(&mut req_parts).await?;
+    let Extension(last_seen_messages) =
+        Extension::<LastSeenMessages>::from_request(&mut req_parts).await?;
+    // PANIC if messages arrive out of order
```
akoshelev, line 81: I wonder if you validated this behavior somehow. Not insisting on writing a unit test on a piece of functionality that will be gone soon, but it would be good to have some level of confidence that it will work as expected. I'll be happy with manual test if that is easier to do

src/net/server/mod.rs
```diff
@@ -35,31 +38,32 @@ pub enum MpcHelperServerError {
...
+        Self::BadPathString(format!("encountered unknown query id: {}", query_id.as_ref()).into())
     }
-}
 
-/// [`From`] implementation for [`MpcServerError::SendError`].
-/// first call `to_string` to as to drop `T` from the [`MpcServerError`]
-impl<T> From<tokio_util::sync::PollSendError<T>> for MpcHelperServerError {
-    fn from(err: tokio_util::sync::PollSendError<T>) -> Self {
-        Self::SendError(err.to_string().into())
+    pub fn out_of_order(last_seen: u32, next_seen: u32) -> Self {
```
akoshelev, line 62: if that is not too much work, would be great to add channel id here.

src/net/server/mod.rs
thurstonsand: sure. this is tested as a part of the `query.rs` tests, but i will add specific tests just for this struct

src/net/server/mod.rs
thurstonsand: gotcha, yeah that seems like the most likely reason why a lock would already be held.

if you look at the definition of `lock()`, it does have a section:

> This function might panic when called if the lock is already held by the current thread.

but like you said, basically the only reason that might happen is if it's been poisoned

src/net/server/mod.rs
thurstonsand: yeah, that's probably true. implemented that

src/net/server/mod.rs
thurstonsand: we probably could, but we'd have to change the trait to expose the opposite end of the `recv_stream`, no?

src/net/server/handlers/query.rs
thurstonsand: extractors only work on handlers, [not middleware](https://docs.rs/axum/latest/axum/middleware/fn.from_fn.html)...

src/net/server/handlers/query.rs
thurstonsand: I thought I should panic when I discover out-of-order data. would we rather just return a failure?

src/net/server/handlers/query.rs
thurstonsand: yeah, i did verify that this worked. I had to modify the tests to get them to pass.

for instance, in the `backpressure_applied` test, if you comment out the `offset += 1`, the test panics with
```
called `Result::unwrap()` on an `Err` value: OutOfOrder { last_seen: 1, next_seen: 0 }
thread 'net::server::handlers::query::tests::backpressure_applied' panicked at 'called `Result::unwrap()` on an `Err` value: OutOfOrder { last_seen: 1, next_seen: 0 }', src/net/server/handlers/query.rs:85:10
```

src/net/server/mod.rs
thurstonsand: done

src/net/server/mod.rs
akoshelev: definition of `lock()` is correct - it will panic if you attempt to grab this lock again while holding it (lets say if you call another function that also tries to lock this mutex) because mutex is not re-entrant. You can still grab a poisoned mutex and get a guard from it if you want, there is no other thread holding that lock (it panicked and released it). 

I forgot that mutex is not re-entrant, so this panic reason is legit. We could also mention poisoning (because of `unwrap`) but I don't really want to bikeshed on this. feel free to leave it as is

src/net/server/handlers/query.rs
akoshelev: that is sad :(

src/net/server/handlers/query.rs
akoshelev: `update_in_place` returns the same error type as this function, so I think it is a matter of changing `unwrap` to `?`. if you want to panic, you would probably want to change `update_in_place` definition

src/net/server/mod.rs
akoshelev: you're right `T: Network` won't work, but can we store the actual struct that implements `Network` here? It should expose methods to send data. I guess I just don't like the fact that we assume and expose the actual implementation for `Network` trait - it should be the internal business of `Network` implementation how to manage connection and communication. Structs outside it ideally just use public API of that struct to send data to it

src/net/server/handlers/query.rs
akoshelev: great to hear! if it is a matter of single line change, can we add a test for it?

src/net/server/handlers/query.rs
thurstonsand: done

src/net/server/handlers/query.rs
thurstonsand: done

src/net/server/mod.rs
thurstonsand: I agree that makes sense. The only thing I'm thinking about is the `&'a peers` field in the struct. How would I handle that ref within `MessageSendMap`?

akoshelev: have you tried something like this?

```rust
struct QueryHandler<'a> {
    inner: Arc<Mutex<HashMap<QueryId, &'a MyConfig<'a>>>>,
}

You can also clone - we need to be careful with it and that's why I think making `PeerDiscovery` return a reference was the right thing to do. but if it is a blocker, I think we can live with it

File: src/net/server/handlers/query.rs [hunk 12]
```diff
@@ -64,13 +65,34 @@ impl<T> Clone for ReservedPermit<T> {
...
+    let Extension(last_seen_messages) =
+        Extension::<LastSeenMessages>::from_request(&mut req_parts).await?;
+    // PANIC if messages arrive out of order; pretty print the error
+    // TODO (ts): remove this when streaming solution is complete
+    last_seen_messages.update_in_place(&ChannelId::new(role, step), record_headers.offset);
+
+    let Extension(message_send_map) =
+        Extension::<MessageSendMap>::from_request(&mut req_parts).await?;
+
+    let sender = message_send_map.get(query_id)?;
```
akoshelev, line 88: Storing `HttpNetwork` inside the map is a big step towards the proper encapsulation. I think what we would want here (probably in a later PR) is hide the sender channel from the axum mw/handler layer as well. It could be as simple as exposing a wrapper `Permit` struct and a method `reserve` on `HttpNetwork` that returns it. In this case middleware does not need to know how `HttpNetwork` is implemented

akoshelev (approved):

PR #227, opened 2022-11-15
Implement Distribution<F> for Standard on prime fields

martinthomson:
I tried to implement for any field, but I couldn't convince the compiler to let me do that.

This should make the testing pieces more ergonomic and maybe also some of the mainline code.

### DIFF
--- a/src/protocol/mul/semi_honest.rs
+++ b/src/protocol/mul/semi_honest.rs
@@ -71,6 +71,8 @@ pub mod tests {
         make_contexts, make_world, share, validate_and_reconstruct, TestWorld,
     };
     use futures_util::future::join_all;
+    use rand::distributions::Standard;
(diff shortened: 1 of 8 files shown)


akoshelev (approved):
👍

benjaminsavage:
This is pretty cool.

src/protocol/mul/semi_honest.rs
```diff
@@ -142,7 +144,10 @@ pub mod tests {
         a: u8,
         b: u8,
         rng: &mut R,
-    ) -> Result<u128, BoxError> {
+    ) -> Result<u128, BoxError>
+    where
+        Standard: Distribution<F>,
+    {
         let a = F::from(u128::from(a));
         let b = F::from(u128::from(b));
```
benjaminsavage, line 152: I think you missed these

PR #229, opened 2022-11-15
Be pedantic always

martinthomson:
Though clippy::pedantic can backfire on occasion, this is easier to manage.  We can deal with exceptions here as well.


src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(clippy::clone_on_ref_ptr)]
+#![deny(warnings, clippy::pedantic)]
```
akoshelev, line 1: do you want to disable `clone_on_ref_ptr` warning? I believe it is not actually included in pedantic https://rust-lang.github.io/rust-clippy/master/#clone_on_ref_ptr

akoshelev (approved):

PR #230, opened 2022-11-15
GenBitPerm can now work with both semi honest and malicious settings

richajaindce:


src/protocol/sort/bit_permutation.rs
```diff
@@ -43,18 +52,21 @@ impl<'a, F: Field> BitPermutation<'a, F> {
...
+        );
 
         let mult_input = self
             .input
             .iter()
-            .map(move |x: &Replicated<F>| share_of_one - *x)
+            .map(move |x: &S| share_of_one - *x)
             .chain(self.input.iter().copied())
-            .scan(Replicated::<F>::new(F::ZERO, F::ZERO), |sum, x| {
+            .scan(S::default(), |sum, x| {
```
richajaindce, line 69: Need to add a test to show it works for malicious. Let me know if it looks otherwise

richajaindce:
Needs some more work

src/protocol/sort/bit_permutation.rs
```diff
@@ -16,13 +21,17 @@ use futures::future::try_join_all;
 /// by K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, N. Kiribuchi, and B. Pinkas
 /// <https://eprint.iacr.org/2019/695.pdf>.
 #[derive(Debug)]
-pub struct BitPermutation<'a, F> {
-    input: &'a [Replicated<F>],
+pub struct BitPermutation<'a, F, S> {
+    input: &'a [S],
+    _marker: PhantomData<F>,
 }
```
benjaminsavage, line 27: First of all, if it helps at all, feel free to eliminate the `new` function and just turn this protocol into a single public async function as we already have done with many other protocols.

Secondly, what is this `PhantomData<F>`? I have not used this before. I don't understand why this is necessary.

Thirdly, I'm surprised that the input doesn't explain that it's connected to the field `F` in any way. It's a secret sharing of a field element, so I was expecting to see something like: `&'a [S<F>]`. Maybe if this was just a function argument to a public async function and not a property in a struct this would be easier?

src/protocol/sort/bit_permutation.rs
```diff
@@ -16,13 +21,17 @@ use futures::future::try_join_all;
...
 
-impl<'a, F: Field> BitPermutation<'a, F> {
-    pub fn new(input: &'a [Replicated<F>]) -> BitPermutation<'a, F> {
-        Self { input }
+impl<'a, S: SecretSharing<F> + Copy, F: Field> BitPermutation<'a, F, S> {
```
benjaminsavage, line 29: Why did you need to add `+ Copy`? I thought we were trying to avoid that one. 

I see that you're introducing a trait bound that `S: SecretSharing<F>`, and `F: Field`. That's good. We don't need these to be totally generic as they appear in the struct definition. I think if this protocol was just a single public async function these bounds would be enough to eliminate the need for that `PhantomData` stuff.

src/protocol/sort/bit_permutation.rs
```diff
@@ -43,18 +52,21 @@ impl<'a, F: Field> BitPermutation<'a, F> {
...
     /// ## Errors
     /// It will propagate errors from multiplication protocol.
-    pub async fn execute(
-        &self,
-        ctx: ProtocolContext<'_, Replicated<F>, F>,
-    ) -> Result<Vec<Replicated<F>>, BoxError> {
-        let share_of_one = Replicated::one(ctx.role());
+    pub async fn execute(&self, ctx: ProtocolContext<'a, S, F>) -> Result<Vec<S>, BoxError>
+    where
+        ProtocolContext<'a, S, F>: SecureMul<F, Share = S>,
```
benjaminsavage, line 57: I don't understand why this was necessary.

src/protocol/sort/bit_permutation.rs
```diff
@@ -43,18 +52,21 @@ impl<'a, F: Field> BitPermutation<'a, F> {
...
+    {
+        let share_of_one = S::one(
+            ctx.role(),
+            ctx.narrow(&ShareOfOne).prss().generate_replicated(RECORD_0),
```
benjaminsavage, line 61: We don't want to generate a new `r`. We need to use the value that the ProtocolContext is already aware of. I recommend migrating all instances of:
`S::one(...)` to `ctx.share_of_one()`. The Context knows if it is malicious or not, and it knows what `r` is, so it can generate something of the form: 
`MaliciousReplicated::new(Replicated::one(self.role()), self.r_share)`

src/protocol/sort/mod.rs
```diff
@@ -86,3 +86,18 @@ impl AsRef<str> for ComposeStep {
...
+
+impl Substep for BitPermutationStep {}
+
+impl AsRef<str> for BitPermutationStep {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::ShareOfOne => "share_of_one",
+        }
+    }
+}
```
benjaminsavage, line 103: All of this should be removed. This is not a step that needs to be done. There is already an `r` in place before this protocol is invoked (possibly) and if you generate a new one, it will not validate.

src/secret_sharing/malicious_replicated.rs
```diff
@@ -41,12 +40,6 @@ impl<F: Field> MaliciousReplicated<F> {
     pub fn rx(&self) -> Replicated<F> {
         self.rx
     }
-
-    /// Returns a pair of replicated secret sharings. One of "one", one of "r"
-    #[allow(dead_code)]
-    pub fn one(helper_role: Role, r_share: Replicated<F>) -> Self {
-        Self::new(Replicated::one(helper_role), r_share)
-    }
```
benjaminsavage, line 49: This is still definitely necessary. Don't remove this. We just need the ProtocolContext to invoke this, because it should know what `r_share` is.

src/secret_sharing/mod.rs
```diff
@@ -2,16 +2,32 @@ mod malicious_replicated;
...
+            Role::H3 => Self::new(F::ZERO, F::ONE),
+        }
+    }
+}
+impl<F: Field> SecretSharing<F> for MaliciousReplicated<F> {
+    /// Returns a pair of replicated secret sharings. One of "one", one of "r"
+    fn one(helper_role: Role, r_share: Replicated<F>) -> Self {
+        Self::new(Replicated::one(helper_role, r_share), r_share)
+    }
+}
```
benjaminsavage, line 33: Let's revert this change. They just don't accept the same arguments, and there isn't a good way to make them both do so. Ignoring the second param on the `Replicated::one()` function isn't great.

src/secret_sharing/replicated.rs
```diff
@@ -35,16 +34,6 @@ impl<F: Field> Replicated<F> {
...
-
-    /// Returns share of value one.
-    #[must_use]
-    pub fn one(helper_role: Role) -> Self {
-        match helper_role {
-            Role::H1 => Self::new(F::ONE, F::ZERO),
-            Role::H2 => Self::new(F::ZERO, F::ZERO),
-            Role::H3 => Self::new(F::ZERO, F::ONE),
-        }
-    }
```
benjaminsavage, line 47: Let's leave this here. Revert this change.

src/protocol/sort/bit_permutation.rs
richajaindce: Thanks for pointing this out, making the change

src/protocol/sort/bit_permutation.rs
richajaindce: Copy is needed since we duplicate the input at ".chain(input.iter().copied())". Is there a reason why Copy in this scenario would be harmful?

src/protocol/sort/bit_permutation.rs
richajaindce: Hmm its stored in SecurityValidator and not the context. However, context has prss and we can use it to get the r_share through prss.generate_replicated(RECORD_0) but this will make prss validation check fail (since we would end up requesting prss for RECORD_0 multiple times.

src/protocol/sort/bit_permutation.rs
martinthomson: You can use .cloned() instead and rely on the Clone implementation.

PR #232, opened 2022-11-15
Simplifying share conversion

benjaminsavage:
Daniel Masny pointed out that the share conversion protocol we had been using was optimized for pushing work into the "offline" phase. In this case, we don't gain any benefit from an "offline" phase, since we are trying to just minimize the total computation cost and latency is not as critical. If we just push it all to the "online" phase, there is actually less communication this way.

### DIFF
--- a/src/protocol/modulus_conversion/convert_shares.rs
+++ b/src/protocol/modulus_conversion/convert_shares.rs
@@ -7,8 +7,7 @@ use crate::{
     secret_sharing::Replicated,
 };
 
-use crate::protocol::reveal::Reveal;
-use futures::future::{try_join, try_join_all};
+use futures::future::try_join_all;
 use std::iter::{repeat, zip};
 
 pub struct XorShares {
@@ -24,7 +23,6 @@ pub struct ConvertShares {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 enum Step {
     DoubleRandom,
-    BinaryReveal,
 }
 
 impl crate::protocol::Substep for Step {}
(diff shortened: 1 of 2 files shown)


src/protocol/modulus_conversion/double_random.rs
```diff
@@ -106,7 +99,7 @@ impl DoubleRandom {
         record_id: RecordId,
         a: Replicated<F>,
         b: Replicated<F>,
-    ) -> Result<Replicated<F>, Error> {
+    ) -> Result<Replicated<F>, BoxError> {
```
akoshelev, line 102: is something not working with `Error`?

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -24,7 +23,6 @@ pub struct ConvertShares {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 enum Step {
```
martinthomson, line 24: Do we still need this?

src/protocol/modulus_conversion/double_random.rs
benjaminsavage: Discussed offline. Seems like we could slowly migrate everything to Error. Maybe not this diff tho.

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage: I kill it in the next diff in this stack: #233

akoshelev (approved):

PR #234, opened 2022-11-16
Add From<i32> implementation for RecordId (tests only)

akoshelev:
I can't type `RecordId::from(0_u32)` anymore, sorry

### DIFF
--- a/src/protocol/modulus_conversion/specialized_mul.rs
+++ b/src/protocol/modulus_conversion/specialized_mul.rs
@@ -209,7 +209,7 @@ pub mod tests {
             let a = rng.gen::<Fp31>();
             let b = rng.gen::<Fp31>();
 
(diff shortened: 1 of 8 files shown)


src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -359,7 +359,7 @@ pub mod tests {
 
             let a_shares = share(a, &mut rng);
 
-            let record_id = RecordId::from(0_u32);
+            let record_id = RecordId::from(0);
 
             let iteration = format!("{}", i);
```
martinthomson, line 364: Can you hold off on landing this until we do #235 ?  This code here needs to stop using `RecordId::from(0)` and putting the value of `i` into a string.

martinthomson (approved):
SGTM

Issue #237, opened 2022-11-16
[Tracking issue] Sort sometime hangs when running on large fields


akoshelev:
This is the tracking issue for sporadic failures I am observing while running the following program (bench). This is almost an exact copy of `generate_sort_permutation` test, but rans faster under bench profile and therefore have a higher chance to be reproduced:

```rust
use std::time::Instant;
use futures_util::future::try_join_all;
use rand::Rng;
use raw_ipa::error::BoxError;
use raw_ipa::ff::Fp32BitPrime;
use raw_ipa::protocol::QueryId;
use raw_ipa::test_fixture::{make_contexts, make_world_with_config, TestWorldConfig, validate_and_reconstruct};
use raw_ipa::protocol::sort::generate_sort_permutation::GenerateSortPermutation;
use raw_ipa::ff::Field;

#[tokio::main(flavor = "multi_thread", worker_threads = 3)]
async fn main() -> Result<(), BoxError> {
    let mut config = TestWorldConfig::default();
    config.gateway_config.send_buffer_config.items_in_batch = 1;
    config.gateway_config.send_buffer_config.batch_count = 1000;
    let world = make_world_with_config(QueryId, config);
    let [ctx0, ctx1, ctx2] = make_contexts::<Fp32BitPrime>(&world);
    let num_bits = 64;
    let mut rng = rand::thread_rng();

    let batchsize = 100;

    let mut match_keys: Vec<u64> = Vec::new();
    for _ in 0..batchsize {
        match_keys.push(rng.gen::<u64>());
    }

    let input_len = match_keys.len();
    let mut shares = [
        Vec::with_capacity(input_len),
        Vec::with_capacity(input_len),
        Vec::with_capacity(input_len),
    ];
    for match_key in match_keys.clone() {
        let share_0 = rng.gen::<u64>();
        let share_1 = rng.gen::<u64>();
        let share_2 = match_key ^ share_0 ^ share_1;

        shares[0].push((share_0, share_1));
        shares[1].push((share_1, share_2));
        shares[2].push((share_2, share_0));
    }

    let start = Instant::now();
    let result = try_join_all(vec![
        GenerateSortPermutation::new(&shares[0], num_bits).execute(ctx0),
        GenerateSortPermutation::new(&shares[1], num_bits).execute(ctx1),
        GenerateSortPermutation::new(&shares[2], num_bits).execute(ctx2),
    ]).await?;
    let duration = start.elapsed().as_secs_f32();
[...]

PR #238, opened 2022-11-16
Add sort bench

akoshelev:
Very rudimentary, just to reproduce #237

the way you would run it (if you want traces):

```rust
RUST_LOG=trace cargo bench --bench oneshot_sort --profile dev --features test-fixture
```


martinthomson (approved):

martinthomson:
Fix the formatting and land it.

PR #235, opened 2022-11-16
Stop making copies of Replicated

martinthomson:
This doesn't help as much as I would have liked, and there are still more calls to .clone() than would be ideal, but it does help, so I'm doing it.

Lots of minor improvements to tests in here.

PR #239, opened 2022-11-16
Gen bit perm for Semi honest and malicious setting

richajaindce:
This is stacked on #236

### DIFF
--- a/src/protocol/context_traits.rs
+++ b/src/protocol/context_traits.rs
@@ -1,0 +1,27 @@
+use crate::{
+    ff::Field,
+    secret_sharing::{MaliciousReplicated, Replicated, SecretSharing},
+};
+
+use super::context::ProtocolContext;
+
+pub trait ShareOfOne<F: Field> {
+    type Share: SecretSharing<F>;
+    fn share_of_one(&self) -> Self::Share;
+}
+
+impl<F: Field> ShareOfOne<F> for ProtocolContext<'_, Replicated<F>, F> {
+    type Share = Replicated<F>;
+
+    fn share_of_one(&self) -> Self::Share {
+        Replicated::one(self.role())
+    }
+}
+
+impl<F: Field> ShareOfOne<F> for ProtocolContext<'_, MaliciousReplicated<F>, F> {
+    type Share = MaliciousReplicated<F>;
+
+    fn share_of_one(&self) -> Self::Share {
+        MaliciousReplicated::one(self.role(), self.r_share())
+    }
+}
--- a/src/protocol/reveal/mod.rs
+++ b/src/protocol/reveal/mod.rs
@@ -251,7 +251,7 @@ mod tests {
             let v = SecurityValidator::new(ctx.narrow("MaliciousValidate"));
             let acc = v.accumulator();
 
-            ctx.upgrade_to_malicious(acc)
+            ctx.upgrade_to_malicious(acc, v.r_share().clone())
         })
     }
 }
--- a/src/protocol/reveal/mod.rs
+++ b/src/protocol/reveal/mod.rs
@@ -126,7 +126,7 @@ mod tests {
     use proptest::prelude::Rng;
     use tokio::try_join;
 
-    use crate::protocol::malicious::SecurityValidator;
+    use crate::test_fixture::make_malicious_contexts;
     use crate::{
         error::BoxError,
         error::Error,
@@ -177,9 +177,9 @@ mod tests {
 
             let record_id = RecordId::from(i);
             let results = try_join_all(vec![
-                ctx[0].clone().reveal(record_id, &share[0]),
-                ctx[1].clone().reveal(record_id, &share[1]),
-                ctx[2].clone().reveal(record_id, &share[2]),
+                ctx[0].ctx.clone().reveal(record_id, &share[0]),
+                ctx[1].ctx.clone().reveal(record_id, &share[1]),
+                ctx[2].ctx.clone().reveal(record_id, &share[2]),
             ])
             .await?;
 
@@ -203,9 +203,9 @@ mod tests {
             let share = share_malicious(input, &mut rng);
             let record_id = RecordId::from(i);
             let result = try_join!(
-                ctx[0].clone().reveal(record_id, &share[0]),
-                ctx[1].clone().reveal(record_id, &share[1]),
-                reveal_with_additive_attack(ctx[2].clone(), record_id, &share[2], Fp31::ONE),
+                ctx[0].ctx.clone().reveal(record_id, &share[0]),
+                ctx[1].ctx.clone().reveal(record_id, &share[1]),
+                reveal_with_additive_attack(ctx[2].ctx.clone(), record_id, &share[2], Fp31::ONE),
             );
 
             assert!(matches!(result, Err(Error::MaliciousRevealFailed)));
@@ -240,18 +240,5 @@ mod tests {
         .await?;
 
         Ok(left + right + share_from_left)
-    }
-
-    /// Creates malicious protocol contexts for 3 helpers. We drop security validator because
-    /// it is not used
-    fn make_malicious_contexts<F: Field>(
-        test_world: &TestWorld,
-    ) -> [ProtocolContext<'_, MaliciousReplicated<F>, F>; 3] {
-        make_contexts(test_world).map(|ctx| {
-            let v = SecurityValidator::new(ctx.narrow("MaliciousValidate"));
-            let acc = v.accumulator();
-
-            ctx.upgrade_to_malicious(acc, v.r_share().clone())
-        })
     }
 }
--- a/src/protocol/sort/bit_permutation.rs
+++ b/src/protocol/sort/bit_permutation.rs
@@ -3,7 +3,8 @@ use std::iter::{repeat, zip};
 use crate::{
     error::BoxError,
     ff::Field,
-    protocol::{context::ProtocolContext, RecordId, RECORD_0}, secret_sharing::{SecretSharing, MaliciousReplicated, Replicated},
+    protocol::{context::ProtocolContext, RecordId, RECORD_0},
+    secret_sharing::{MaliciousReplicated, Replicated, SecretSharing},
 };
 
 use crate::protocol::mul::SecureMul;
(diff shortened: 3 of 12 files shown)


src/protocol/sort/bit_permutation.rs
```diff
@@ -3,83 +3,93 @@ use std::iter::{repeat, zip};
...
+/// 5. Compute the final output, a vector of length `n`. Each element `i` in this output vector is the sum of
+/// the elements at index `i` and `i+n` from the vector computed in step 4.
+///
+/// ![Bit Permutation steps][bit_permutation]
+/// ## Panics
+/// In case the function is unable to get double size of output from multiplication step, the code will panic
+///
+/// ## Errors
+/// It will propagate errors from multiplication protocol.
+pub async fn bit_permutation<'a, F: Field, S: SecretSharing<F> + Copy>(
```
richajaindce, line 58: Will try to get rid of this Copy once Martin's changed are landed from #235

src/protocol/sort/bit_permutation.rs
```diff
@@ -3,83 +3,93 @@ use std::iter::{repeat, zip};
...
+    fn share_of_one(&self) -> Self::Share {
+        Replicated::one(self.role())
+    }
+}
+
+impl<F: Field> ShareOfOne<F> for ProtocolContext<'_, MaliciousReplicated<F>, F> {
+    type Share = MaliciousReplicated<F>;
+
+    fn share_of_one(&self) -> Self::Share {
+        MaliciousReplicated::one(self.role(), self.prss().generate_replicated(RECORD_0))
```
richajaindce, line 31: @benjaminsavage @akoshelev is this the correct way to generate r_share? Context does not hold r_share and thats why I did it like this

src/protocol/sort/bit_permutation.rs
```diff
@@ -3,83 +3,88 @@ use std::iter::{repeat, zip};
...
+pub async fn bit_permutation<'a, F: Field, S: SecretSharing<F> + Clone>(
+    ctx: ProtocolContext<'a, S, F>,
+    input: &[S],
+) -> Result<Vec<S>, BoxError> 
+where ProtocolContext<'a, S, F>: SecureMul<F, Share = S> + ShareOfOne<F, Share = S>{
+    let share_of_one = ctx.share_of_one();
+
+    let mult_input = input
+        .iter()
+        .map(|x: &S| &share_of_one - x)
```
richajaindce, line 65: cannot subtract `&S` from `&S`. consider extending the `where` clause, but there might be an alternative better way to express this requirement: `, &S: Sub<&S>`

richajaindce: For now, we are cloning share_of_one to take care of this.

src/protocol/sort/bit_permutation.rs
benjaminsavage: No, this is not correct.

`SecurityValidator` is the thing that knows what the `r_share` is. It has an accessor: `validator.r_share()`.

When the `SecurityValidator` is created via a call to `SecurityValidator::new()` that's when the global value of `r_share` is initialized. 

If you look at the tests in `malicious.rs` like `simplest_circuit` you'll see that right now nobody holds a reference to this thing... that's a problem.

Right now, I just get the `SecurityValidatorAccumulator` from it via a call to `v.accumulator()`, and this object is passed to the context via a call to `upgrade_to_malicious(acc.clone())`

Then, nobody has a reference to the actual `SecurityValidator` ever again... eventually someone needs it in order to call `validate()`

Here is what I'd suggest:

Update the `upgrade_to_malicious()` function to *also* take a clone of `r_share`. Once it has this, you can expose a new `get_share_of_one()` method on the context. Then, the context (who knows both the `role` and `r_share`) will be capable of generating a share of one in the malicious context.

benjaminsavage (changes requested):

src/protocol/sort/bit_permutation.rs
```diff
@@ -3,83 +3,90 @@ use std::iter::{repeat, zip};
...
-impl<'a, F: Field> BitPermutation<'a, F> {
-    pub fn new(input: &'a [Replicated<F>]) -> BitPermutation<'a, F> {
-        Self { input }
+impl<F: Field> ShareOfOne<F> for ProtocolContext<'_, Replicated<F>, F> {
+    type Share = Replicated<F>;
+
+    fn share_of_one(&self) -> Self::Share {
+        Replicated::one(self.role())
     }
+}
```
benjaminsavage, line 25: Stuff like this should not be defined within this file `bit_permutation.rs` - this is a particular protocol, but many protocols will require a share of one.

src/protocol/sort/bit_permutation.rs
```diff
@@ -3,83 +3,90 @@ use std::iter::{repeat, zip};
...
+/// In case the function is unable to get double size of output from multiplication step, the code will panic
+///
+/// ## Errors
+/// It will propagate errors from multiplication protocol.
+pub async fn bit_permutation<'a, F: Field, S: SecretSharing<F>>(
+    ctx: ProtocolContext<'a, S, F>,
+    input: &[S],
+) -> Result<Vec<S>, BoxError>
+where
+    ProtocolContext<'a, S, F>: SecureMul<F, Share = S> + ShareOfOne<F, Share = S>,
```
benjaminsavage, line 62: What is this? Do we need to add these types of trait bounds every time we upgrade protocols to use SecretSharing traits?

src/protocol/sort/bit_permutation.rs
richajaindce: Created a new file called context_trail.rs and moving this code there. Other context related trait definitions can also be added here.

src/protocol/sort/bit_permutation.rs
richajaindce: @akoshelev is there any better way to get rid of this? This is required by the compiler to know protocol does implement these traits

src/protocol/sort/bit_permutation.rs
richajaindce: Makes complete sense! Making the change

src/protocol/context_traits.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn share_of_one(&self) -> Self::Share {
+        Replicated::one(self.role())
+    }
+}
+
+impl<F: Field> ShareOfOne<F> for ProtocolContext<'_, MaliciousReplicated<F>, F> {
+    type Share = MaliciousReplicated<F>;
+
+    fn share_of_one(&self) -> Self::Share {
+        MaliciousReplicated::one(self.role(), self.r_share())
```
benjaminsavage, line 25: this is perfect

src/protocol/reveal/mod.rs
```diff
@@ -177,9 +177,9 @@ mod tests {
 
             let record_id = RecordId::from(i);
             let results = try_join_all(vec![
-                ctx[0].clone().reveal(record_id, &share[0]),
-                ctx[1].clone().reveal(record_id, &share[1]),
-                ctx[2].clone().reveal(record_id, &share[2]),
+                ctx[0].ctx.clone().reveal(record_id, &share[0]),
+                ctx[1].ctx.clone().reveal(record_id, &share[1]),
+                ctx[2].ctx.clone().reveal(record_id, &share[2]),
```
benjaminsavage, line 182: What is going on here? Why did you need to add a `.ctx`? And what does that even do?

src/protocol/sort/bit_permutation.rs
```diff
@@ -53,13 +55,13 @@ pub async fn bit_permutation<F: Field>(
...
 
-    assert_eq!(mult_output.len(), input.len() * 2);
+    debug_assert!(mult_output.len() == input.len() * 2);
     // Generate permutation location
     let len = mult_output.len() / 2;
     for i in 0..len {
         // we are subtracting "1" from the result since this protocol returns 1-index permutation whereas all other
         // protocols expect 0-indexed permutation
-        let less_one = &mult_output[i + len] - &share_of_one;
+        let less_one = mult_output[i + len].clone() - &share_of_one;
```
benjaminsavage, line 64: Why did you change this to clone()?

src/protocol/sort/bit_permutation.rs
```diff
@@ -112,4 +117,40 @@ mod tests {
...
+        let [ctx0, ctx1, ctx2] = make_malicious_contexts::<Fp31>(&world);
+        let mut rand = StepRng::new(100, 1);
+
+        let mut shares = [
+            Vec::with_capacity(INPUT.len()),
+            Vec::with_capacity(INPUT.len()),
+            Vec::with_capacity(INPUT.len()),
+        ];
+        for i in INPUT {
+            let share = share_malicious(Fp31::from(*i), &mut rand);
```
benjaminsavage, line 138: How can you share malicious without knowing the value of `r`?

src/protocol/sort/bit_permutation.rs
```diff
@@ -112,4 +117,40 @@ mod tests {
...
+        let h1_future = bit_permutation(ctx1.ctx, shares[1].as_slice());
+        let h2_future = bit_permutation(ctx2.ctx, shares[2].as_slice());
+
+        let result: [_; 3] = try_join_all([h0_future, h1_future, h2_future])
+            .await
+            .unwrap()
+            .try_into()
+            .unwrap();
+
+        validate_list_of_shares_malicious(EXPECTED, &result);
```
benjaminsavage, line 154: We should call `validate()` on the SecurityValidator after running the protocol and ensure it is indeed valid.

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -18,79 +18,58 @@ use embed_doc_image::embed_doc_image;
...
     num_bits: u8,
-}
-
-impl<'a> GenerateSortPermutation<'a> {
-    #[must_use]
-    pub fn new(input: &'a [(u64, u64)], num_bits: u8) -> GenerateSortPermutation {
-        Self { input, num_bits }
-    }
+) -> Result<Vec<Replicated<F>>, BoxError> {
+    let ctx_0 = ctx.narrow(&Sort(0));
```
benjaminsavage, line 41: Why did you need to add this? Can we remove this?

src/protocol/sort/bit_permutation.rs
benjaminsavage: Ugh.. I missed this when `share_malicious()` was added to the codebase... I'll fix it in another diff.

src/protocol/sort/bit_permutation.rs
benjaminsavage: I'll add this in a diff to come.

benjaminsavage (approved):
This is good work @richajaindce! Thanks for adding this `share_of_one()` functionality!

src/protocol/reveal/mod.rs
richajaindce: make malicious context returns both context and validator (so that it is not dropped). Thats why we need to explicitly call ctx here

src/protocol/sort/bit_permutation.rs
richajaindce: We are now performing S + &S vs &S + &S since addition of two &S isnt supported as yet.

PR #240, opened 2022-11-16
Random Bits protocol

taikiy:
This protocol generates a sequence of uniformly random sharing of bits in `F_p`. This achieves what Damgård's "3.1 Generating random solved bits" protocol does, but instead of using their "RAN_2" protocol, we use  our PRSS and ConvertShares.

This protocol returns `None` indicating abort if the secret number, which can be reconstructed from the generated random shared bits, is not less than the field's prime number. The caller may decide to call this protocol again for another attempt to generate random bits, but must provide a new narrowed context.


src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
+        record_id: RecordId,
+    ) -> Result<(Vec<Replicated<F>>, Replicated<F>), BoxError> {
+        // We assume the bit length we operate in would not exceed 255
+        #[allow(clippy::cast_possible_truncation)]
+        let l = F::Integer::BITS as u8;
+
+        // Generate a replicated random field values. We'll use these fields as
+        // the source of `l`-bit long uniformly random sequence of bits.
+        let r_p: Replicated<F> = ctx.prss().generate_replicated(record_id);
```
benjaminsavage, line 25: Since you're just going to immediately call as_u128() I don't see any reason to coerce these into field values. Just use the raw numbers you get from generate_values()

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        // Convert each bit to secret sharings of that bit in the target field
+        let futures = (0..l).map(|i| {
+            // again, we don't expect our prime field to be > 2^64
+            #[allow(clippy::cast_possible_truncation)]
+            let xor_shares = XorShares {
+                num_bits: l,
+                packed_bits_left: r_p.left().as_u128() as u64,
+                packed_bits_right: r_p.right().as_u128() as u64,
+            };
```
benjaminsavage, line 35: Pull this out of the for-loop. This is the same for all iterations.

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Take a slice of bits in `{0,1} ⊆ F_p`, and reconstruct the integer in `F_p`
+    fn bits_to_field<F: Field>(x: &[F]) -> F {
+        #[allow(clippy::cast_possible_truncation)]
+        let v = x
+            .iter()
+            .enumerate()
+            .fold(0, |acc, (i, &b)| acc + 2_u128.pow(i as u32) * b.as_u128());
+        F::from(v)
+    }
```
benjaminsavage, line 72: Is this duplicate code? I think this appears in two places now. Can you just move this into text_fixtures and call it from both places?

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+            let xor_shares = XorShares {
+                num_bits: l,
+                packed_bits_left: r_p.left().as_u128() as u64,
+                packed_bits_right: r_p.right().as_u128() as u64,
+            };
+            let c = ctx.narrow(&BitOpStep::Step(i as usize));
+            async move {
+                ConvertShares::new(xor_shares)
+                    .execute_one_bit(c, record_id, i as u8)
+                    .await
```
benjaminsavage, line 40: I'm always a bit fuzzy on this - do you need to call await here if you're about to try_join_all?

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+        // [r]_B must be the same bit lengths
+        assert_eq!(r0_b.len(), r1_b.len());
+        assert_eq!(r1_b.len(), r2_b.len());
+
+        // Reconstruct r_B from ([r_1]_p,...,[r_l]_p) bitwise sharings in F_p
+        let r_b = (0..r0_b.len())
+            .map(|i| validate_and_reconstruct((r0_b[i], r1_b[i], r2_b[i])))
+            .collect::<Vec<_>>();
+        // [r]_p is an XOR secret sharing of `r`
+        let r = validate_and_reconstruct_xor((r0_p, r1_p, r2_p));
```
benjaminsavage, line 105: This is a problem. You're using a different method of reconstruction here, to interpret this as XOR sharings. I don't think that works for us. What we eventually want from the SOLVED-BITs protocol is both a normal sharing of [r] (additive, not XOR) and a bitwise sharing of the same [r]. Look at the SOLVED-BITs protocol; they generate random bits and then they check that it's less than `p` and abort if not. This is just that first part.

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,129 @@
...
+            let c = ctx.narrow(&BitOpStep::Step(i as usize));
+            async move {
+                ConvertShares::new(xor_shares)
+                    .execute_one_bit(c, record_id, i as u8)
+                    .await
+            }
+        });
+        let r_b = try_join_all(futures).await?;
+
+        Ok((r_b, r_p))
```
benjaminsavage, line 45: Don't bother returning r_p. This has no meaning. It's not an equivalent secret-sharing of the same value. You're having to use a special XOR reconstruction in your test because it's not. This is what I demonstrated in that Google sheet from a few days ago, that this approach does NOT implement the SOLVED-BITs protocol. This is a perfectly fine way of generating a sequence of shared bits, which is like steps 1 and 2 of the SOLVED-BITs protocol, but it's not equivalent to the rest - we will still need to do all of that work. Specifically it's step 6 of that protocol which you need to compute the sharing of the random number in the large prime field.

benjaminsavage (changes requested):

src/protocol/boolean/random_bits.rs
taikiy: AFAIK, yes. `try_join_all` expects its argument to implement a trait with `type Output = Result`. Without `.await` here, `execute_one_bit` returns a `Future`.

src/protocol/boolean/random_bits.rs
taikiy: I used `generated_replicated()` because the generated random numbers must be less than `p`. I changed it to `generate_fields` since we don't really need `Replicated`.

src/protocol/boolean/random_bits.rs
taikiy: This inside the for-loop because `XorShares` doesn't implement `Clone`. I added `#[derive(Clone, Copy)]` for XorShares and pulled the instantiation code out of the for-loop.

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        // We assume the bit length we operate in would not exceed 255
+        #[allow(clippy::cast_possible_truncation)]
+        let l = F::Integer::BITS as u8;
+
+        // Generate a pair of random numbers. We'll use these numbers as
+        // the source of `l`-bit long uniformly random sequence of bits.
+        let r_bits: (F, F) = ctx
+            .narrow(&Step::RandomFields)
+            .prss()
+            .generate_fields(record_id);
```
benjaminsavage, line 28: I mentioned before - and I'll mention again - I think we should use generate_values() not generate_fields(). Coercing each share into a range that is smaller than the prime has no use because the additive sharings of those bits will NOT be constrained by the prime.

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp31>(&world);
+        let [c0, c1, c2] = ctx;
+
+        for i in 0..TEST_TRIES {
+            let record_id = RecordId::from(i);
+            let (r_b, r_p): (Vec<Fp31>, Fp31) =
+                random_bits([c0.clone(), c1.clone(), c2.clone()], record_id).await?;
+            // Base10 of r_b must equal r_p
+            assert_eq!(r_p, bits_to_field(&r_b));
```
benjaminsavage, line 150: I am very surprised that this works. I would assume that these random bits will (frequently) sum to a value that is larger than the prime for that field. In those cases, I wouldn't have assumed it would pass this assert. I'd like to understand what kind of wrapping / truncation is going on in these cases and why it's the same. Are we sure we have sufficient test coverage for such cases?

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp32BitPrime>(&world);
+        let [c0, c1, c2] = ctx;
+
+        for i in 0..TEST_TRIES {
+            let record_id = RecordId::from(i);
+            let (r_b, r_p): (Vec<Fp32BitPrime>, Fp32BitPrime) =
+                random_bits([c0.clone(), c1.clone(), c2.clone()], record_id).await?;
+            // Base10 of r_b must equal r_p
+            assert_eq!(r_p, bits_to_field(&r_b));
```
benjaminsavage, line 166: There are just 5 numbers that are more than this prime. Let's make sure we have some tests for them.

src/test_fixture/mod.rs
```diff
@@ -136,3 +136,13 @@ pub fn permutation_valid(permutation: &[u32]) -> bool {
...
     true
 }
+
+/// Take a slice of bits in `{0,1} ⊆ F_p`, and reconstruct the integer in `F_p`
+pub fn bits_to_field<F: Field>(x: &[F]) -> F {
+    #[allow(clippy::cast_possible_truncation)]
+    let v = x
+        .iter()
+        .enumerate()
+        .fold(0, |acc, (i, &b)| acc + 2_u128.pow(i as u32) * b.as_u128());
```
benjaminsavage, line 146: I think there is something *really* subtle going on here.

Because the {0, 1} values have type of `F: Field` they are *wrapping around* when we add them up here. 

That means that if we gave it an input like:
```
[1, 1, 1, 1, 1, 1, 1, 1]
```
This function is going to do:
1+2+4+8+16+32+64+128 = 255
Fp31::from(255)
# => 7
```

...which is kind of like saying:

"[1, 1, 1, 1, 1, 1, 1, 1] is the bitwise representation of 7"

...and that's not true... If we performed a "bit decomposition" of the number 7 and got [1, 1, 1, 1, 1, 1, 1, 1] we wouldn't be happy, right?

Comment (martinthomson) [reply src/test_fixture/mod.rs:146]: I can't help myself:

```suggestion
        .fold(0, |acc, (i, &b)| acc + (b.as_u128() << (i as u32)));
```

On the substantive part: why would get a value of 255 in like that?

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
+        record_id: RecordId,
+    ) -> Result<(Vec<Replicated<F>>, Replicated<F>), BoxError> {
+        // We assume the bit length we operate in would not exceed 255
+        #[allow(clippy::cast_possible_truncation)]
+        let l = F::Integer::BITS as u8;
+
+        // Generate a pair of random numbers. We'll use these numbers as
+        // the source of `l`-bit long uniformly random sequence of bits.
+        let r_bits: (F, F) = ctx
```
martinthomson, line 25:
```suggestion
        let (rbits_left, rbits_right): (F, F) = ctx
```

Please, if you are doing `x.0` or `x.1`, try to think of that as a problem, not a feature.

src/test_fixture/mod.rs
benjaminsavage: Oh that is kind of slick =). Will clippy complain about possible truncation or something?

On the substantive part:

We are literally generating random bits. So for each bit there are 50/50 odds of getting a 1 or a 0. So one time in 256 runs, you'll get 0b11111111

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        };
+
+        // Convert each bit to secret sharings of that bit in the target field
+        let c = ctx.narrow(&Step::ConvertShares);
+        let futures = (0..l).map(|i| {
+            // again, we don't expect our prime field to be > 2^64
+            #[allow(clippy::cast_possible_truncation)]
+            let c = c.narrow(&BitOpStep::Step(i as usize));
+            async move {
+                ConvertShares::new(xor_shares)
```
martinthomson, line 45: Why are you instantiating this object `l` times?  It's the same object every time, isn't it?

src/test_fixture/sharing.rs
```diff
@@ -41,6 +41,17 @@ where
...
 }
 
+/// Take a field value `x` and turn them into replicated bitwise sharings of three
+pub fn shared_bits<F: Field, R: RngCore>(x: F, rand: &mut R) -> Vec<[Replicated<F>; 3]>
+where
+    Standard: Distribution<F>,
+{
+    let x = x.as_u128();
+    (0..F::Integer::BITS)
+        .map(|i| share(F::from(x >> i & 1), rand))
```
martinthomson, line 51:
```suggestion
        .map(|i| share(F::from((x >> i) & 1), rand))
```

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -10,10 +10,11 @@ use crate::{
...
 use std::iter::{repeat, zip};
 
+#[derive(Clone, Copy, Debug)]
 pub struct XorShares {
-    num_bits: u8,
-    packed_bits_left: u64,
-    packed_bits_right: u64,
+    pub num_bits: u8,
+    pub packed_bits_left: u64,
+    pub packed_bits_right: u64,
```
martinthomson, line 17: This seems like a problem we can solve with a constructor instead.

src/protocol/boolean/random_bits.rs
```diff
@@ -0,0 +1,170 @@
...
+        ctx: [ProtocolContext<'_, Replicated<F>, F>; 3],
+        record_id: RecordId,
+    ) -> Result<(Vec<F>, F), BoxError>
+    where
+        Standard: Distribution<F>,
+    {
+        let [c0, c1, c2] = ctx;
+
+        // Execute
+        let result = try_join_all(vec![
```
martinthomson, line 109:
```suggestion
        let result = try_join_all([
```

No need to allocate for this.

martinthomson: That said, I've been considering whether we do a try_join3() wrapper for our code.  That might be worth doing.

src/test_fixture/mod.rs
martinthomson: I thought that that was why the values of the random fields were being clamped to the range of F, rather than taking the raw u128.

src/protocol/boolean/random_bits.rs
taikiy: I pulled this out from the loop. It was because XorShare didn't implement Copy at first, but I added `derive`.

src/protocol/boolean/random_bits.rs
taikiy: Changed to use `generate_values()`

src/protocol/boolean/random_bits.rs
taikiy: You are absolutely correct. Changed it to use `generate_values()`.

src/protocol/boolean/random_bits.rs
taikiy: This works because the base10 conversion is done in F_p. That said, I need to fix the protocol to make sure that the returned bits sharings satisfies `b < p`.

src/test_fixture/mod.rs
taikiy: I created this method to test BitwiseSum, which could return a sequence of bits outside the range of `F`. I created another method `bits_to_value() -> u128` that doesn't wrap the integer into a field.

src/protocol/boolean/random_bits.rs
taikiy: Oh, I wasn't aware. Thanks!

src/protocol/boolean/solved_bits.rs
```diff
@@ -0,0 +1,265 @@
...
+        if ctx.narrow(&Step::RevealC).reveal(record_id, &c_b).await? == F::ZERO {
+            return Ok(false);
+        }
+        Ok(true)
+    }
+
+    /// Internal use only.
+    /// Converts the prime number to a sequence of `{0,1} ⊆ F`, and creates a
+    /// local replicated share.
+    fn local_secret_shared_bits_p<F: Field>(helper_role: Role) -> Vec<Replicated<F>> {
```
benjaminsavage, line 126: There's probably some way to memoize this - so that it only needs to be computed once.

src/protocol/boolean/solved_bits.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub async fn fp31() -> Result<(), Error> {
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp31>(&world);
+        let [c0, c1, c2] = ctx;
+
+        // 0.012 secs per invocation. 100 tries = 1.2 secs
+        for i in 0..100 {
+            let record_id = RecordId::from(i);
+            // The chance of this protocol aborting 100 out of 100 tries in Fp31 is < 0.0001%.
+            // Should we test and see if it succeeds at least once?
```
benjaminsavage, line 234: Yes, we probably should. Otherwise, imagine a bug that causes this to just return None 100% of the time. This test will still pass.

benjaminsavage (approved):
This looks great! Just a few comments left then let's land it!

src/protocol/boolean/solved_bits.rs
martinthomson: Are you sure that this calculation is correct?  $31^{-100}$ is much less than 1 in a million.  That is, assuming the abort happens 1/31.

Also, one in a million is not really good enough odds for the test.  You should aim for about $2^{-100}$ odds of failure, which is only 21 runs for `Fp31`.  That's faster than this as well.

src/protocol/boolean/solved_bits.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    #[tokio::test]
+    pub async fn fp_32bit_prime() -> Result<(), Error> {
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp32BitPrime>(&world);
+        let [c0, c1, c2] = ctx;
+
+        // 0.055 secs per invocation. 20 tries = 1.1 secs
+        for i in 0..20 {
```
martinthomson, line 253: Here you should only need to run the test about three or four times.

src/protocol/boolean/solved_bits.rs
taikiy: We generate 8 random bits, each bit having a 50/50 chance of being 0 or 1. So, the chance of the number being in the range of Fp31 is, the 3 higher bits need to be 0's and others could be either 0 or 1, which is 1/2^3? (ignoring "30" where the last bit needs to be 0 if other 4 bits are 1's)

So the abort happens about 1 - 1/8 = 88%, and 100 times that is around 1 in a million, I think. For 2^-100, I need 500+ runs?

src/protocol/boolean/solved_bits.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    /// Generates a sequence of `l` random bit sharings in the target field `F`.
+    async fn generate_random_bits<F: Field>(
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
+        record_id: RecordId,
+    ) -> Result<Vec<Replicated<F>>, Error> {
+        // We assume the bit length we operate in would not exceed 255
+        #[allow(clippy::cast_possible_truncation)]
+        let l = F::Integer::BITS as u8;
```
martinthomson, line 80: The abort rate in this protocol is too high, but here is where you can fix that.

You are looking at the number of fields in the integer, not the prime.  You can calculate the number of bits you need with:

```rust
let l = F::PRIME::BITS - F::PRIME.leading_zeros();
```

src/protocol/boolean/solved_bits.rs
martinthomson: See above.  You shouldn't need that many runs.  The odds of failure can easily be much lower.

src/protocol/boolean/solved_bits.rs
benjaminsavage: I like this. Only need to generate 5 bits.

this logic will also help us if we decide to use the Mersenne prime 2^31 - 1 at some point in the future. Only need to generate 31 bits for that one! If we generate 32, it’ll have a 50% abort rate.

src/protocol/boolean/solved_bits.rs
taikiy: Rust didn't support `memoize` by default, but found a crate for it. Added a TODO for potential optimization.

src/protocol/boolean/solved_bits.rs
```diff
@@ -75,20 +76,21 @@ impl SolvedBits {
         ctx: ProtocolContext<'_, Replicated<F>, F>,
         record_id: RecordId,
     ) -> Result<Vec<Replicated<F>>, Error> {
-        // We assume the bit length we operate in would not exceed 255
-        #[allow(clippy::cast_possible_truncation)]
-        let l = F::Integer::BITS as u8;
+        // Calculate the number of bits we need to form a random number that
+        // has the same number of bits as the prime.
+        let l = u128::BITS - F::PRIME.into().leading_zeros();
```
taikiy, line 82: `const PRIME` of a generic `F` is of a type `Integer`, and only supports `Into<u128>`. Subtract from 128 to get the number of bits needed to generate.

src/protocol/boolean/solved_bits.rs
```diff
@@ -97,13 +99,23 @@ impl SolvedBits {
...
                     .await
             }
         });
 
-        try_join_all(futures).await
+        // Pad 0's at the end to return `F::Integer::BITS` long bits
+        let mut b_b = try_join_all(futures).await?;
+        #[allow(clippy::cast_possible_truncation)]
+        b_b.append(
+            &mut repeat(Replicated::ZERO)
```
taikiy, line 114: These 0'a are additive shares of 0 in `F`, but not random. I think it's okay though because this protocol leaks `b < p` anyway.

src/protocol/boolean/solved_bits.rs
```diff
@@ -0,0 +1,285 @@
...
+/// I. Damgård et al.
+pub struct SolvedBits {}
+
+impl SolvedBits {
+    // Try generating random sharing of bits, `[b]_B`, and `l`-bit long.
+    // Each bit has a 50% chance of being a 0 or 1, so there are
+    // `F::Integer::MAX - p` cases where `b` may become larger than `p`.
+    // With `Fp32BitPrime` (prime is `2^32 - 5`), that chance is around
+    // 1 * 10^-9. For Fp31, the chance of this protocol succeeding is around
+    // 1/2^3 =~ 13% (3 high bits being all 0's).
```
benjaminsavage, line 41: Comment out of date. New chance of failure is 1 out of 32 (~3.125%)

benjaminsavage (approved):
Awesome. Thanks for making all the changes and getting this PR there!

src/protocol/boolean/solved_bits.rs
taikiy: I'll include this in the next PR.

PR #241, opened 2022-11-16
tests for #135

andyleiserson:
Some tests I wrote for #135. Haven't settled on a strategy yet.

### DIFF
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -193,3 +193,71 @@ impl<'a, F: Field> ProtocolContext<'a, MaliciousReplicated<F>, F> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        ff::Fp31,
+        protocol::QueryId,
+        test_fixture::{make_contexts, make_world},
+    };
+
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    struct TestStep;
+
+    impl Substep for TestStep {}
+
+    impl AsRef<str> for TestStep {
+        fn as_ref(&self) -> &str {
+            "test"
+        }
+    }
+
+    #[tokio::test]
+    pub async fn narrow() {
+        let world = make_world(QueryId);
+        let context = make_contexts::<Fp31>(&world);
+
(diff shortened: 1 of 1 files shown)


src/protocol/context.rs
```diff
@@ -177,3 +177,74 @@ impl<'a, F: Field> ProtocolContext<'a, MaliciousReplicated<F>, F> {
...
+    impl Substep for TestStep {}
+
+    impl AsRef<str> for TestStep {
+        fn as_ref(&self) -> &str {
+            "test"
+        }
+    }
+
+    #[tokio::test]
+    pub async fn narrow() -> Result<(), BoxError> {
```
akoshelev, line 203: it seems that you don't need those tests to be async?

akoshelev (approved):

src/protocol/context.rs
andyleiserson: I tried changing them to regular tests, but there is a call to `tokio::spawn` within `InMemoryNetwork::new` within `make_world`.

src/protocol/context.rs
```diff
@@ -193,3 +193,71 @@ impl<'a, F: Field> ProtocolContext<'a, MaliciousReplicated<F>, F> {
...
+    //  // do something for RecordId i
+    // }
+    // ```
+    //
+    // ```
+    // for i in 0..2 {
+    //  // do something *without* narrowing context by `i` or using it as RecordId
+    // }
+    // ```
+    #[ignore] // broken, see #135
```
martinthomson, line 253: I'm not sure what the material difference is between this test and the previous one.  What am I missing?

andyleiserson: The idea is that, in the specific case where `i` is a record ID, narrowing a context twice in this way should not be an error. It is okay to construct the same context path multiple times, as long as you're doing it for different record IDs.

In the test case it would be possible to avoid the issue by moving the call to `narrow` outside the loop, but in general the body of the `for record_id` loop can be spread across multiple functions and it may not be desirable to pull the calls to `narrow` all the way up. (https://github.com/private-attribution/ipa/compare/main...andyleiserson:ipa:issue135 shows what that might look like.)

PR #242, opened 2022-11-17
bye bye BoxError

benjaminsavage:
Now that SecureMul doesn't throw BoxErrors, all the protocols that call it can just use normal errors. This should make it easier to match on them and stuff.


martinthomson (approved):

akoshelev:
just wanted to say thank you

PR #243, opened 2022-11-17
Fix reordering bug inside Gateway

### LINKED ISSUE #237: [Tracking issue] Sort sometime hangs when running on large fields

akoshelev:
Each time elements were inserted out of order to the send buffer inside the gateway, there was a risk of messages stuck inside the send buffer forever. This change addresses that and also adds a unit test that should prevent this from happening again.

This fixes #237

### DIFF
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -15,11 +15,12 @@ use std::ops::Range;
 /// vector has 3 regions and `X` indicates that space at that element is occupied.
 ///
 ///  region1  region2  region3
-/// [X,_,_,_][_,_,X,_][_,X,_,X]
+/// [X,_,_,_][X,_,X,_][_,X,_,X]
 ///
-/// Once `region1` is completely filled up, it is possible to drain the vector
+/// Once `region1` is completely filled up, it is possible to drain the vector. Draining will cause
+/// **all** elements from the head of the queue to be removed
 ///
-/// [X,X,X,X][_,_,X,_][_,X,_,X] -> `drain` -> [_,_,X,_][_,X,_,X][_,_,_,_]
+/// [X,X,X,X][X,_,X,_][_,X,_,X] -> `take` -> [_,X,_,_][X,_,X,_][_,_,_,_]
 ///
 /// This vector is used inside the send buffer to keep track of messages added to it. Once first
 /// batch of messages is ready (region1 is full), it drains this vector and send those messages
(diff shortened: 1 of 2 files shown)


src/helpers/buffers/fsv.rs
```diff
@@ -73,15 +74,16 @@ impl<const N: usize> FixedSizeByteVec<N> {
     /// ## Panic
     /// Panics if `ready()` is false before calling `drain()`.
     pub fn take(&mut self) -> Option<Vec<u8>> {
-        if self.added[..self.region_size].all() {
-            self.added.drain(..self.region_size).for_each(drop);
-            let r = self.data.drain(..self.region_size * N).collect();
-            self.elements_drained += self.region_size;
+        let last_inserted = self.added.leading_ones();
```
thurstonsand, line 77: so the main difference for this PR is that the returned `Vec<u8>` previously would always return with length `self.region_size`, whereas now, it returns length of whatever `self.added.leading_ones()` is?

thurstonsand (approved):
this feels like a more ergonomic way to deal with out-of-order messages getting cached and flushed

src/helpers/buffers/fsv.rs
akoshelev: that is correct. Previously it always returned chunks of the same size, even if next chunk was ready, `fsv` required another call to `take` to take it out. Now it tries to push everything it can out which is a contiguous set of elements in the front of the queue.

martinthomson (approved):

akoshelev:
#252 fixes is and also makes `fsv` easier

PR #244, opened 2022-11-17
make this test a bit better

benjaminsavage:
This seems a bit better.

### DIFF
--- a/src/protocol/mul/semi_honest.rs
+++ b/src/protocol/mul/semi_honest.rs
@@ -71,12 +71,9 @@ pub mod tests {
         make_contexts, make_world, share, validate_and_reconstruct, TestWorld,
     };
     use futures::future::try_join_all;
-    use futures_util::future::join_all;
-    use rand::distributions::Standard;
-    use rand::prelude::Distribution;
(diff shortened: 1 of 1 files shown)


src/protocol/mul/semi_honest.rs
```diff
@@ -103,40 +100,46 @@ pub mod tests {
...
+        let mut expected_outputs = Vec::with_capacity(10);
+
+        let futures: Vec<_> = zip(repeat(contexts), 0..10)
+            .map(|(ctx, i)| {
+                let a = rng.gen::<Fp31>();
+                let b = rng.gen::<Fp31>();
+                expected_outputs.push(a * b);
+
+                let a_shares: [Replicated<Fp31>; 3] = share(a, &mut rng);
+                let b_shares: [Replicated<Fp31>; 3] = share(b, &mut rng);
```
akoshelev, line 116:
```suggestion
                let a_shares = share(a, &mut rng);
                let b_shares = share(b, &mut rng);
```

there may be a reason why it wouldn't work but I don't see any - `a` and `b` are `Fp31`

akoshelev (approved):

PR #246, opened 2022-11-17
Remove deny(warnings)

akoshelev:
This setting prevents compiler from building the code with warnings. I think it is useful to be able to do that because often I prototype an implementation and want to see whether it compiles and it is ok if there are unadressed warnings such asunused imports. Also I want to be able to run tests and get feedback from them without addressing all the warnings.

I think we have good protection at PR level as GH action would prevent pushing code that violates this no warnings rule.

I also enabled validation for `cargo doc` if we ever need to use it  based on conversations in https://github.com/rust-lang/cargo/issues/8424

Verified it is working and failing build: https://github.com/akoshelev/raw-ipa/actions/runs/3486104797

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -5,6 +5,9 @@ jobs:
     name: Rust Checks
     env:
       CARGO_TERM_COLOR: always
+      RUSTFLAGS: -D warnings
+      RUSTDOCFLAGS: -D warnings
+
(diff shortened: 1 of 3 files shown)


benjaminsavage:
I'll leave this to @martinthomson

martinthomson (approved):
Can you add it back to the pre-commit?  You can always use -n if you want a checkpoint.

PR #247, opened 2022-11-17
making specialized multiplications maliciously secure

benjaminsavage:
In order to make modulus conversion maliciously secure, we need the specialised multiplications to also be maliciously secure. In this PR, I'm pulling them into the trait where normal multiplication lives.


src/protocol/mul/mod.rs
```diff
@@ -38,6 +52,28 @@ impl<F: Field> SecureMul<F> for ProtocolContext<'_, Replicated<F>, F> {
     ) -> Result<Self::Share, Error> {
         SemiHonestMul::new(self, record_id).execute(a, b).await
     }
+
+    async fn multiply_two_shares_mostly_zeroes(
```
akoshelev, line 56: I've been debating with myself whether it is better to have two additional traits for specialized multiply or two methods on `SecureMul` trait. I think it is better to proceed with your approach as it will be a bit easier to read the code that uses this specialized mul (no need to refer to use section to figure out which multiplication is used).

akoshelev (approved):

benjaminsavage:
Too old. I'll probably try this again, but abandoning for now.

PR #248, opened 2022-11-17
Reveal_permutation can now work with SecretSharing

richajaindce:


akoshelev:
I'll review it after we push #253 and rebase this on top of it

benches/oneshot/sort.rs
```diff
@@ -45,9 +45,9 @@ async fn main() -> Result<(), BoxError> {
 
     let start = Instant::now();
     let result = try_join_all(vec![
-        GenerateSortPermutation::new(&shares[0], num_bits).execute(ctx0),
-        GenerateSortPermutation::new(&shares[1], num_bits).execute(ctx1),
-        GenerateSortPermutation::new(&shares[2], num_bits).execute(ctx2),
+        generate_sort_permutation(ctx0, &shares[0], num_bits),
+        generate_sort_permutation(ctx1, &shares[1], num_bits),
+        generate_sort_permutation(ctx2, &shares[2], num_bits),
```
martinthomson, line 50: By the way, I like this move to a simple function rather than an object.  It is nicer to use by far.

I do think that maybe you could make the name shorter.  Even `generate_permutation` would be less of a mouthful.

benches/oneshot/sort.rs
```diff
@@ -45,9 +45,9 @@ async fn main() -> Result<(), BoxError> {
 
     let start = Instant::now();
     let result = try_join_all(vec![
```
martinthomson, line 47:
```suggestion
    let result = join3(
```

benches/oneshot/sort.rs
```diff
@@ -45,9 +45,9 @@ async fn main() -> Result<(), BoxError> {
...
-        GenerateSortPermutation::new(&shares[0], num_bits).execute(ctx0),
-        GenerateSortPermutation::new(&shares[1], num_bits).execute(ctx1),
-        GenerateSortPermutation::new(&shares[2], num_bits).execute(ctx2),
+        generate_sort_permutation(ctx0, &shares[0], num_bits),
+        generate_sort_permutation(ctx1, &shares[1], num_bits),
+        generate_sort_permutation(ctx2, &shares[2], num_bits),
     ])
     .await?;
```
martinthomson, line 52:
```suggestion
    )
    .await;
```
With the above, that is.

martinthomson (commented):
Yeah, this is going to take a bit of rebasing.

benches/oneshot/sort.rs
richajaindce: Taken care in #265

benches/oneshot/sort.rs
richajaindce: #265

benches/oneshot/sort.rs
richajaindce: #265

richajaindce:
Will close this and open a new one with more upgrades :)

PR #249, opened 2022-11-17
very small fix

benjaminsavage:
A small fix to #239

### DIFF
--- a/src/protocol/sort/bit_permutation.rs
+++ b/src/protocol/sort/bit_permutation.rs
@@ -50,9 +50,13 @@ where
             Some((x, sum.clone()))
         });
 
-    let async_multiply = zip(repeat(ctx), mult_input)
-        .enumerate()
-        .map(|(i, (ctx, (x, sum)))| async move { ctx.multiply(RecordId::from(i), &x, &sum).await });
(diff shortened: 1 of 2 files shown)


src/protocol/sort/bit_permutation.rs
```diff
@@ -50,9 +50,13 @@ where
...
 
-    let async_multiply = zip(repeat(ctx), mult_input)
-        .enumerate()
-        .map(|(i, (ctx, (x, sum)))| async move { ctx.multiply(RecordId::from(i), &x, &sum).await });
+    let async_multiply =
+        zip(repeat(ctx), mult_input)
+            .enumerate()
+            .map(|(i, (ctx, (x, sum)))| async move {
+                let record_id = RecordId::from(i);
+                ctx.bind(record_id).multiply(record_id, &x, &sum).await
```
richajaindce, line 58: makes sense

richajaindce (approved):

src/protocol/sort/bit_permutation.rs
martinthomson: You don't need to bind the record_id here, do you?

PR #250, opened 2022-11-17
Make share_malicious accept r

benjaminsavage:
This was definitely wrong. Our tests of malicious contexts aren't yet calling validate - so this issue wasn't detected yet. But in reality, a MaliciousReplicated sharing should be a sharing of (x, r*x) and the *same* value of `r` needs be be used for the entire protocol, until it is revealed at the time `validate()` is called.


akoshelev (approved):
I missed that `r` must be the same, sorry

PR #251, opened 2022-11-17
Run lint/Clippy first before build in GH actions

akoshelev:
Most of the time, PR is broken because of formatting and/or Clippy issues. Therefore, it is desirable to run these checks first and get the feedback earlier before running builds and tests. My observation (likely biased towards my own behavior) is that we rarely submit code that does not compile or tests that are broken.


martinthomson:
Yes, this is the correct ordering - fmt is also very fast.

A reminder to those watching that installing the pre-commit hook is a good way to avoid most build failures.

PR #252, opened 2022-11-18
Simpler FSV

martinthomson:
See also #243.

### DIFF
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -1,6 +1,5 @@
 use bitvec::bitvec;
 use bitvec::prelude::BitVec;
-use std::ops::Range;
 
 /// A vector of bytes that never grows over a certain size or shrinks below that size.
 /// Vector is segmented into regions and `max_size` is a factor of number of regions.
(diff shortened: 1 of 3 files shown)


src/helpers/buffers/fsv.rs
```diff
@@ -28,81 +28,61 @@ use std::ops::Range;
...
-            self.data[offset].copy_from_slice(&elem);
-
-            Some(r)
-        } else {
-            self.added.set(index, true);
-            self.data[offset].copy_from_slice(&elem);
-
-            None
-        }
+        assert!(!self.added[index]);
```
akoshelev, line 50: would `debug_assert` be appropriate here?

akoshelev (approved):
I like it, will abandon #243

src/helpers/buffers/fsv.rs
martinthomson: I'm on the fence about this one.  Downgrading is work, so I'm going to say that we leave it for now.  It's a little annoying to find all the `#[should_panic]` tests and add `#[cfg(debug_assertions)]`.  I will add a TODO though.

Issue #139, opened 2022-10-19
Thoughts on making Step and ProtocolContext cheaper


martinthomson:
We don't really know whether these structures are going to impose significant overheads just yet, but we have identified reasons for both needing some attention.

Warning: lots of rust code that I just typed into a text editor.  No guarantees that any of this works.

## ProtocolContext

I'm advocating for ProtocolContext being cheap so that we can make copies for every step and every record.  This should improve API ergonomics.  The structure (as of writing) is:

```rust
pub struct ProtocolContext<'a, N> {
    role: Identity,
    step: UniqueStepId, // or Step
    prss: &'a PrssEndpoint,
    gateway: &'a Gateway<N>,
}
```

That's a very small enum for role (1 bit possible, 8 bits likely), an owned string (yikes), and two references (64 bits each maybe, though this might be compressed on some platforms).

The primary offender here is `step`, which we'll get back to, but it seems like there is an obvious fix here.  Moving the shared state into another object could make this nicer.  One way to do that is to do something like this:

```rust
trait HelperState {  // name TBD
  type Network: crate::protocol::Network;
  fn role(&self) -> Identity;
  fn gateway(&self) -> &Gateway<Self::Network>;
  fn prss(&self) -> &PrssEndpoint;
}

pub struct ProtocolContext<'a, H> {
  global: &'a H,
  step: UniqueStepId, // or Step
}

impl<H: HelperState> for ProtocolContext<'_, H> {
  // This can access self.global.prss() rather than self.prss
}
```

This style of indirection might be necessary to allow the in-memory networking to refer to a single global state and a real networking setup to refer to a single helper-wide object.  Either way, the overall cost is then dominated by the step identifier.  The interface is no more complex.  The only cost is following a pointer when we want something.

## Step

In debugging and testing, we want something that is descriptive of the process.  Our current approach describes steps with strings and then composes those together into something like a path.

We can use that process happily in debugging, but step identifiers need to be passed around a LOT and that might hurt performance when we need to shift a lot of data.  What we need is a process that produces an object that can be represented in memory as a simple integer.

One way to do that is to map the path-like structure we naturally produce onto an integer, assigning each node in the path (including "directories") an integer value.  

```
protocol => 0
protocol/a => 1
protocol/a/x => 2
protocol/a/y => 3
protocol/b => 4
protocol/c => 5
protocol/c/x => 6
```

[...]

PR #253, opened 2022-11-18
Protocol context is a trait

akoshelev:
This change makes `ProtocolContext` a trait and implements it for two newly created structs that represent semi-honest and malicious contexts. 

Each protocol can decide whether it supports semi-honest/malicious security model or both by injecting `SemiHonestProtocolContext`, `MaliciousProtocolContext` or  generic type parameter `C: ProtocolContext<F>` accordingly. 

There are a couple of important changes to the context API: 

* `narrow` no longer enforces "don't narrow to the same step twice" rule.  There were two reasons for having this rule.
   * Don't use the same shared randomness more than once. #208 protects against it
   * Make sure every communication is uniquely addressable by record_id + step. If someone narrows to the same step more than once and uses the same record id to communicate, Infra is in trouble. It won't be able to resolve these two communications and likely apply last-write-wins which is obviously not correct here.  It is also mitigated because of acceptable record ranges introduced in #183. Now if protocol attempts to send a record outside of valid range, it will be caught. If it attempts to send the same record twice within the acceptable range, `fsv` will panic. 
* `bind` is gone. No need to explicitly bind context to a record because of the above


src/protocol/context/malicious.rs
```diff
@@ -0,0 +1,102 @@
+use crate::ff::Field;
```
akoshelev, line 1: moved context implementations to a separate folder

src/protocol/context/malicious.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+use crate::protocol::{Step, Substep};
+use crate::secret_sharing::{MaliciousReplicated, Replicated};
+use std::sync::Arc;
+
+/// Represents protocol context in malicious setting, i.e. secure against one active adversary
+/// in 3 party MPC ring.
+#[derive(Clone, Debug)]
+pub struct MaliciousProtocolContext<'a, F: Field> {
+    inner: ContextInner<'a>,
```
akoshelev, line 17: @martinthomson I remember we were talking about using references here, but I couldn't find a good way to do it. there must be something that holds onto this inner struct for the duration of protocol execution. It didn't seem very ergonomic but I could be missing something

src/protocol/context/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// This will panic if you have previously invoked `prss()`.
+    #[must_use]
+    fn prss_rng(&self) -> (SequentialSharedRandomness, SequentialSharedRandomness);
+
+    /// Get a set of communications channels to different peers.
+    #[must_use]
+    fn mesh(&self) -> Mesh<'_, '_>;
+
+    /// Generates a new share of one
+    fn share_of_one(&self) -> <Self as ProtocolContext<F>>::Share;
```
akoshelev, line 65: `share_of_one` is now defined inside protocol context trait. That allows malicious context to generate it without exposing the internal state (r_share)

src/protocol/mod.rs
```diff
@@ -115,18 +83,10 @@ impl Step {
         {
             let s = String::from(step.as_ref());
             assert!(!s.contains('/'), "The string for a step cannot contain '/'");
-            assert!(
```
akoshelev, line 118: step validation is gone

src/protocol/modulus_conversion/specialized_mul.rs
```diff
@@ -215,19 +216,19 @@ pub mod tests {
 
             let result_shares = try_join_all([
                 multiply_two_shares_mostly_zeroes(
-                    context[0].bind(record_id),
+                    context[0].clone(),
```
akoshelev, line 219: no more `bind` calls

src/protocol/sort/bit_permutation.rs
```diff
@@ -33,13 +32,15 @@ use futures::future::try_join_all;
 ///
 /// ## Errors
 /// It will propagate errors from multiplication protocol.
-pub async fn bit_permutation<'a, F: Field, S: SecretSharing<F>>(
-    ctx: ProtocolContext<'a, S, F>,
+pub async fn bit_permutation<
```
akoshelev, line 35: this is how the protocols would look like if they can work in semi-honest and malicious setting. 

Note that there is no more `SecureMul` constraint because it is a supertrait for  `ProtocolContext`.

src/protocol/sort/reshare.rs
```diff
@@ -36,7 +37,7 @@ impl<F: Field> Reshare<F> {
     ///    `to_helper.right` = (`rand_right`, part1 + part2) = (r0, part1 + part2)
     pub async fn execute(
         self,
-        ctx: &ProtocolContext<'_, Replicated<F>, F>,
+        ctx: &SemiHonestProtocolContext<'_, F>,
```
akoshelev, line 40: this probably needs to be fixed later

src/test_fixture/mod.rs
```diff
@@ -67,10 +70,10 @@ pub fn make_malicious_contexts<F: Field>(test_world: &TestWorld) -> [MaliciousCo
 /// # Panics
 /// Never, but then Rust doesn't know that; this is only needed because we don't have `each_ref()`.
 #[must_use]
-pub fn narrow_contexts<'a, F: Field, S: SecretSharing<F>>(
-    contexts: &[ProtocolContext<'a, S, F>; 3],
+pub fn narrow_contexts<C: Debug + ProtocolContext<F, Share = S>, F: Field, S: SecretSharing<F>>(
```
akoshelev, line 73: `Debug` because there is a call to `unwrap` inside this function

src/protocol/context/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+mod malicious;
+mod semi_honest;
+
+pub use malicious::MaliciousProtocolContext;
+pub use semi_honest::SemiHonestProtocolContext;
+
+/// Context used by each helper to perform secure computation. Provides access to shared randomness
+/// generator and communication channel.
+pub trait ProtocolContext<F: Field>:
```
akoshelev, line 21: no more `bind`

akoshelev:
I think tests from #241 now can be enabled - I would expect Infra to panic if someone tries to send the same message twice. I can also add a test for that (Friday PT). 

I also didn't update the panic message inside `fsv` to make it clear what went wrong for protocol authors.

src/protocol/context/malicious.rs
andyleiserson: I think the idea was to collect the static components of the context (role, prss, gateway) into `GlobalContext` so that a single `&GlobalContext` needs to be duplicated each time the context is refined, rather than multiple items.

This is something that I was planning to look at, but I haven't done anything with it yet. Hopefully it can be done in a way that doesn't change the context API.

src/protocol/context/mod.rs
andyleiserson: At the start of the meeting yesterday, @martinthomson and I were discussing checking for bound contexts using the type system. https://github.com/private-attribution/ipa/compare/main...andyleiserson:ipa:context2 has some changes I was working on in that direction. I was thinking it might make sense to get rid of the separate `record_id` parameter and just pass it via the context.

I had also worked on some changes to enforce one-time use restrictions. Those are in https://github.com/andyleiserson/ipa/compare/context2...andyleiserson:ipa:context3.

src/protocol/context/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    Clone + SecureMul<F, Share = <Self as ProtocolContext<F>>::Share> + Reveal
```
martinthomson, line 22:
```suggestion
    Clone + SecureMul<F, Share = <Self as ProtocolContext<F>>::Share> + Reveal<Share = <Self as ProtocolContext<F>>::Share>
```

Reveal and SecureMul are different in that one takes F and the other does not.  Can we work out which one we prefer?  If we can avoid multiple type parameters, maybe that is better.

src/protocol/context/mod.rs
martinthomson, line 21: As below, it might be possible - based on what I'm seeing anyway - to remove the F generic parameter from this and attach it to the Share type.

src/protocol/context/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn share_of_one(&self) -> <Self as ProtocolContext<F>>::Share;
+}
+
+/// Contains things that are applicable to any implementation of protocol context as see it today
+/// Every context requires access to current step, PRSS and communication and that is what this
+/// struct carries.
+#[derive(Clone, Debug)]
+struct ContextInner<'a> {
+    role: Role,
+    step: Step,
```
martinthomson, line 71: My thinking (see #139) had `Step` outside of the shared part.  That is, the shared components are kept in the inner piece, but the part that needs to change is kept with each instance.

As long as the inner object is `Sync`, then we should be able to refer to it with `Arc`.  `Arc` is just a pointer, so cloning it is as cheap as copying that and doing a relaxed atomic increment.

src/protocol/context/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn new(role: Role, prss: &'a PrssEndpoint, gateway: &'a Gateway) -> Self {
+        Self {
+            role,
+            step: Step::default(),
+            prss,
+            gateway,
+        }
+    }
+
+    fn narrow<S: Substep + ?Sized>(&self, step: &S) -> Self {
```
martinthomson, line 86: If you follow the structure above, then there is no `narrow()` on this type.

src/protocol/context/semi_honest.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::secret_sharing::Replicated;
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+/// Context for protocol executions suitable for semi-honest security model, i.e. secure against
+/// honest-but-curious adversary parties.
+#[derive(Clone, Debug)]
+pub struct SemiHonestProtocolContext<'a, F: Field> {
+    inner: ContextInner<'a>,
+    _marker: PhantomData<F>,
```
martinthomson, line 19: As noted, I think we can remove `F` from the type and have it attach to `Self::Share`.

src/protocol/context/malicious.rs
```diff
@@ -0,0 +1,101 @@
...
+    accumulator: SecurityValidatorAccumulator<F>,
+    r_share: Replicated<F>,
```
martinthomson, line 19: It seems like the inner object will need to be per-implementation rather than shared.  Otherwise, we're duplicating `r_share` and `accumulator` when the context is cloned.

src/test_fixture/mod.rs
martinthomson: It's good that `ProtocolContext` is not `Debug` in the normal case.  It's OK to be `Debug` for tests, but we shouldn't need it for real code.

martinthomson (commented):
This is definitely an improvement, but I think that we need to iterate on the structure of the `Context` implementations.

src/helpers/messaging.rs
```diff
@@ -252,6 +252,7 @@ impl Debug for ReceiveRequest {
 mod tests {
     use crate::ff::Fp31;
     use crate::helpers::Role;
+    use crate::protocol::context::ProtocolContext;
```
martinthomson, line 255: I forgot to make this suggestion.  I think that we should rename `ProtocolContext` (and `...ProtocolContext` to `Context` (and `...Context`) and save some typing.

benjaminsavage: I like this suggestion.

src/protocol/context/malicious.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn step(&self) -> &Step {
+        &self.inner.step
+    }
+
+    fn narrow<S: Substep + ?Sized>(&self, step: &S) -> Self {
+        Self {
+            inner: self.inner.narrow(step),
+            accumulator: self.accumulator.clone(),
+            // TODO (alex, mt) - is cloning ok here or we need to Cow it?
+            r_share: self.r_share.clone(),
```
benjaminsavage, line 82: Per @martinthomson's suggestion above, this should be a part of a specialized, MaliciousContext specific `inner` so that it need not be copied at all.

src/protocol/mod.rs
```diff
@@ -3,7 +3,6 @@ mod batch;
 mod boolean;
 mod check_zero;
 pub mod context;
-pub mod context_traits;
```
benjaminsavage, line 6: Nice

src/protocol/mod.rs
```diff
@@ -68,43 +62,17 @@ impl Substep for str {}
...
 #[cfg_attr(
     feature = "enable-serde",
     derive(serde::Deserialize),
     serde(from = "&str")
 )]
 pub struct Step {
     id: String,
-    /// This tracks the different values that have been provided to `narrow()`.
-    #[cfg(debug_assertions)]
-    used: Arc<Mutex<HashSet<String>>>,
```
benjaminsavage, line 81: Very nice

src/protocol/mod.rs
benjaminsavage: Hooray!

src/protocol/sort/bit_permutation.rs
benjaminsavage: That's really nice. Thanks for eliminating those constraints!

src/protocol/sort/reshare.rs
benjaminsavage: You mean the fact that it's a reference and not an actual object? @richajaindce

benjaminsavage (commented):
This is a huge win. I can't wait to land this. Thanks @akoshelev!

src/protocol/sort/reshare.rs
richajaindce: I changed it in #257 . Will rebase on this PR once we are happy with it

src/protocol/context/malicious.rs
akoshelev: With #261 I think we can make it ergonomic to use because we can just stitch global contexts inside test world and create contexts on demand. For real world implementation I am not sure if we can make it better than just holding onto both global context and protocol context

src/protocol/context/mod.rs
akoshelev: Hmm, with this change we don't need `record_id` inside the context, unless we want to change the way how we multiply shares. Currently it is  `ctx.multiply(record_id, a, b)`, do we want `ctx.multiply(a, b)`?

src/protocol/context/mod.rs
akoshelev: There was a need to make `Reveal` work with any field as `Context<F>` was used to reveal `Fp2` values. It is no longer the case, so I am going to change `Reveal` and make it look the same as `SecureMul`

src/protocol/context/mod.rs
akoshelev: I tried to look into that and I ran into some issues

* It does not seem possible to constraint the `Share` type 

this does not work

```rust
pub trait ProtocolContext:
    Clone
    + SecureMul<Share<F> = <Self as ProtocolContext>::Share<F>>
    + Reveal<Share<F> = <Self as ProtocolContext>::Share<F>>
```

and removing `F` does not work either:

```rust
pub trait ProtocolContext:
    Clone
    + SecureMul<Share = <Self as ProtocolContext>::Share>
    + Reveal<Share = <Self as ProtocolContext>::Share>
```

I also don't like that `MaliciousContext` needs to be generic over `F` to store accumulator but moving `F` to `Share` will force it to support multiplication of any `G: Field` which is wrong.

src/protocol/context/mod.rs
akoshelev: I can take a stab at moving step away from inner.

src/protocol/context/mod.rs
martinthomson: I kinda like what Andy has done here.  The `ctx.multiply(a, b)` would be `ctx.bind(record).multiply(a, b)`, which seems like a solid approach, though it looks like `ctx.multiply(record, a, b)` also works fine.

into_parts() taking a reference seems unfortunate; it would be good to understand what parts of that don't work; is it that these functions also narrow()?  It seems like you might want something on which you can narrow only in addition to the usable bits if that is the case.

The ergonomics of into_parts leaves a little to be desired.  It's not awful, but it loses some of the simplicity.

martinthomson: Ah, that's right.  I had forgotten that reveal was weird.

andyleiserson: > into_parts() taking a reference seems unfortunate

Agreed, this was temporary so that I didn't have to update all references to the context all at once. I would want to fix it before merging anything.

>  It seems like you might want something on which you can narrow only in addition to the usable bits if that is the case.

Also agreed, this was on my to do list.

akoshelev: I am a little hesitant because `ctx.multiply(a, b)` syntax invites us to solve the problem "how to bind once", while `ctx.multiply(record, a, b)` does not.

src/protocol/context/mod.rs
akoshelev: all right, I've opened a can of worms here.  I tried to avoid `Arc` but I don't seem that I can do it. The most annoying thing is that we really need `Arc` for tests only. `Arc` causes double-indirection every time we need to dereference `Gateway` or `Participant` :(  

Is using `Arc` ok for now or is there a better way to handle this? Also I think if we migrate all tests to use #261, we don't have this problem so should we push these first? 

```mermaid
graph TD;
    TestWorld-->|owns|Gateway;
    ContextInner-->|borrows|Gateway;
    subgraph ephemeral
    Context-->|borrows|ContextInner;
    end
    Test-->|owns|Context;
    Test-->|owns|TestWorld;
```

akoshelev: it is done and both contexts now use `Arc`. I've wasted another couple of hours trying to make `MaliciousContext` look like `SemiHonest` or, more precisely `Context<F, Share = Replicated<F>>`.  So sad it didn't work (https://github.com/rust-lang/rust/issues/20400)

src/protocol/context/mod.rs
andyleiserson: By "how to bind once", do you mean keeping track of whether the context is already bound and issuing a warning if it is, similar to the "context narrowed more than once" warning that is removed by this PR?

My proposal is to keep track of that with the type system. So if it compiles, the context is not bound more than once. The benefit of this is there's less chance of a bug due to passing a wrong or inconsistent record_id somewhere.

martinthomson (approved):

martinthomson:
Progress!

src/protocol/context/mod.rs
akoshelev: I guess I am debating whether we need compile-time check vs the runtime check that already exists. There are some really nice things we can do with bounded contexts - for example implement multiplication and reveal for contexts that carry `record_id` , so nobody will be able to call `multiply` on `Context<.., AnyRecord>`. On the other hand, it requires some machinery and makes code a bit harder to understand? Let me add this topic to the agenda for the Thursday sync-up - there are some really nice things we can do with the model you proposed.

PR #255, opened 2022-11-19
Add shuttle crate and randomized test for sort

akoshelev:
Unfortunately I can't enable this test right now because we have a bug somewhere. Using this randomized scheduler revealed a bug in reveal (!?). Reveal reconstructs wrong secret for some of the permutation values. I am not sure where the bug is, likely in infra but it is certainly triggered by reordering task execution. Will be looking into that, submitting this PR if someone else wants to take a look  too.

Here is how to run this test

```bash
cargo test --features shuttle --lib tests::randomized::sort
```

I also refactored the sort circuit so we can run the same test in benchmarks, unit tests and randomized tests.

this is how test currently fails

```bash
thread 'main' panicked at 'index out of bounds: the len is 50 but the index is 2988901655', /Users/koshelev/workspace/raw-ipa/src/protocol/sort/apply.rs:39:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
stack backtrace:
```

I briefly checked it and it seems permutation slice contains garbage values along with actual indices. Would need to investigate further

## Why Shuttle?

* I had some good feedback about it from the folks I worked with before. They were using both Loom and Shuttle and Loom verification took much longer to execute for them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,13 +5,14 @@ rust-version = "1.64.0"
 edition = "2021"
 
 [features]
-default = ["debug", "cli"]
 cli = ["comfy-table", "clap", "enable-serde", "metrics-tracing-context", "metrics-util", "tracing-subscriber", "web-app"]
 debug = []
+default = ["debug", "cli"]
 enable-serde = ["serde", "serde_json"]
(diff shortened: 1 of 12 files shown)


akoshelev:
closing as shuttle framework is merged already

PR #257, opened 2022-11-21
Introducing Reshare trait for SecretSharing and upgrading reshare for malicious

richajaindce:


src/protocol/sort/reshare.rs
```diff
@@ -128,4 +136,43 @@ mod tests {
         }
         assert!(new_reshares_atleast_once);
     }
+
+    #[tokio::test]
+    pub async fn test_reshare_malicious() {
```
martinthomson, line 141:
```suggestion
    pub async fn reshare_malicious() {
```

src/protocol/sort/reshare.rs
```diff
@@ -1,83 +1,91 @@
...
+    to_helper: Role,
+) -> Result<MaliciousReplicated<F>, Error> {
+    let rx_ctx = ctx.narrow(&ReshareMAC);
+    let (x, rx) = try_join(
+        reshare(ctx.to_semi_honest(), input.x(), record_id, to_helper),
+        reshare(rx_ctx.to_semi_honest(), input.rx(), record_id, to_helper),
+    )
+    .await?;
+    Ok(MaliciousReplicated::new(x, rx))
+}
```
martinthomson, line 88:
```suggestion
}

```

src/protocol/sort/reshare.rs
```diff
@@ -128,4 +136,43 @@ mod tests {
...
+
+            let h0_future = reshare_malicious(ctx0.ctx.clone(), &share0, record_id, Role::H2);
+            let h1_future = reshare_malicious(ctx1.ctx.clone(), &share1, record_id, Role::H2);
+            let h2_future = reshare_malicious(ctx2.ctx.clone(), &share2, record_id, Role::H2);
+
+            let f = try_join_all([h0_future, h1_future, h2_future])
+                .await
+                .unwrap();
+            let output_share = validate_and_reconstruct(f[0].x(), f[1].x(), f[2].x());
+            let output_macs = validate_and_reconstruct(f[0].rx(), f[1].rx(), f[2].rx());
```
martinthomson, line 166: I think that we're going to need a fixture helper to manage this validation process.

martinthomson (commented):
You should probably wait for #253 here.  That is more important.

Otherwise, I think that this is great.  I am more comfortable with this pattern than I am the other versions of this.

src/protocol/sort/reshare.rs
richajaindce: Its a test for reshare_malicious so needs to be named differently.

akoshelev (approved):

PR #259, opened 2022-11-21
Do not create the world for each test loop

taikiy:
This diff improves the unit tests for many boolean protocols I implemented lately.

* Pull out `make_world()` and `make_contexts()` out of functions that are called every time a test case is executed.
* Move dup'ed test helper functions to `test_fixture`
* Replace hard-coded test cases to random test cases as much as appropriate

`Carries` (and `BitwiseSum` which calls `Carries`) is very slow (0.4 sec per invocation), so I limited the number of test tries to 5 to shorten the test case execution time. There are TODOs for some parts of the protocol that can be optimized, but still the protocol seems too expensive. I'll open an issue to track this.

### DIFF
--- a/src/protocol/boolean/prefix_or.rs
+++ b/src/protocol/boolean/prefix_or.rs
@@ -306,24 +306,29 @@ mod tests {
     use crate::{
         error::Error,
         ff::{Field, Fp2, Fp31},
-        protocol::{QueryId, RecordId},
+        protocol::{context::ProtocolContext, QueryId, RecordId},
         secret_sharing::Replicated,
-        test_fixture::{make_contexts, make_world, share, validate_and_reconstruct, TestWorld},
+        test_fixture::{
+            logging, make_contexts, make_world, share, validate_and_reconstruct, TestWorld,
+        },
     };
     use futures::future::try_join_all;
     use rand::distributions::{Distribution, Standard};
     use rand::{rngs::mock::StepRng, Rng};
     use std::iter::zip;
 
-    const BITS: [usize; 2] = [16, 32];
+    const BITS: [usize; 1] = [32];
     const TEST_TRIES: usize = 16;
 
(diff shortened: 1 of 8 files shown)


src/protocol/boolean/prefix_or.rs
```diff
@@ -306,24 +306,29 @@ mod tests {
...
+            logging, make_contexts, make_world, share, validate_and_reconstruct, TestWorld,
+        },
     };
     use futures::future::try_join_all;
     use rand::distributions::{Distribution, Standard};
     use rand::{rngs::mock::StepRng, Rng};
     use std::iter::zip;
 
-    const BITS: [usize; 2] = [16, 32];
+    const BITS: [usize; 1] = [32];
```
taikiy, line 320: When I execute a test case for `PrefixOr` with different bit-lengths, the protocol halts during the first calls to `or()` at the second `BITS` loop. For example, if I have:

`const BITS: [usize; 2] = [16, 32];`

All cases `TEST_TRIES` times for 16-bit passes. But the first calls to `or()`, which is called from `step1()`, of 32-bit test cases stops indefinitely. How can I debug this?

[protocol__boolean__prefix_or__tests__fp31.log](https://github.com/private-attribution/ipa/files/10053935/protocol__boolean__prefix_or__tests__fp31.log)

taikiy: Resolved. `narrow`ing at the beginning of each loop before testing different bit-lengths solved the issue.

benjaminsavage:
Lots and lots of conflicts now =). Can you rebase?

benjaminsavage:
On second thought - so much has changed that it might make sense to re-evaluate this one entirely.

taikiy:
It'd be faster to work on a new PR to achieve the same. Closing for now.

PR #260, opened 2022-11-21
upgrade_to_malicious() accepts an input share

benjaminsavage:
Trying to change upgrade_to_malicious() to accept a share which needs to be turned into a MaliciousReplicated.

### DIFF
--- a/src/protocol/context.rs
+++ b/src/protocol/context.rs
@@ -6,15 +6,34 @@ use super::{
     RecordId, Step, Substep,
 };
 use crate::{
+    error::Error,
     ff::Field,
     helpers::{
         messaging::{Gateway, Mesh},
         Role,
     },
-    protocol::{malicious::SecurityValidatorAccumulator, prss::Endpoint as PrssEndpoint},
+    protocol::{
+        malicious::SecurityValidatorAccumulator, mul::SemiHonestMul, prss::Endpoint as PrssEndpoint,
+    },
+    secret_sharing::{MaliciousReplicated, Replicated, SecretSharing},
 };
 
-use crate::secret_sharing::{MaliciousReplicated, Replicated, SecretSharing};
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum MaliciousInputValidationStep {
+    RandomnessForInputValidation,
+    InputMultByR,
+}
+
+impl crate::protocol::Substep for MaliciousInputValidationStep {}
+
+impl AsRef<str> for MaliciousInputValidationStep {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::RandomnessForInputValidation => "randomness_for_input_validation",
+            Self::InputMultByR => "input_mult_by_r",
+        }
+    }
+}
 
 /// Context used by each helper to perform computation. Currently they need access to shared
 /// randomness generator (see `Participant`) and communication trait to send messages to each other.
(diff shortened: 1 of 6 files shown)


src/test_fixture/mod.rs
```diff
@@ -52,17 +54,94 @@ pub struct MaliciousContext<'a, F: Field> {
...
+    )
+}
+
+/// # Errors
+/// If any of the calls to validate fail, this will throw an error
+pub async fn validate_circuit<F: Field>(
+    contexts: [ProtocolContext<'_, MaliciousReplicated<F>, F>; 3],
+    validators: [SecurityValidator<F>; 3],
+) -> Result<(), Error> {
+    let _validation_results = try_join_all(
```
martinthomson, line 139: You shouldn't need to assign this to anything, it's a type of `()`, right?

src/protocol/context.rs
```diff
@@ -129,13 +148,41 @@ impl<'a, F: Field, SS: SecretSharing<F>> ProtocolContext<'a, SS, F> {
...
-        ProtocolContext {
+        record_id: RecordId,
+        input: Replicated<F>,
+    ) -> Result<
+        (
+            ProtocolContext<'a, MaliciousReplicated<F>, F>,
+            MaliciousReplicated<F>,
+        ),
+        Error,
+    > {
```
martinthomson, line 167: I've been thinking about this and maybe what we really need is a wrapper type being returned here.

What we really need is to pass the accumulator and `r` to the context to get a malicious context.  But we also want to ensure a few other things:

1. The inputs are passed through something to turn them from `Replicated` to `MaliciousReplicated`.  This needs to happen for multiple inputs, but we need just one context from that.  Whatever thing that is needs to have access to a semi-honest context (for multiplication) and the share of `r`.

2. The validation needs to occur at the end of the processing.  This either requires that we pass a malicious context (which is weird, because any would do there) but it would really be better as an object that is consumed.  The return value here could be that thing.

The result would look something like:

```rust
async fn full_protocol(ctx: SemiHonestContext<...>, a: &Replicated<F>, b: &Replicated<F>) -> Replicated<F> {
    // The upgrade function can generate the accumulator and shares of `r`.
    // The return value holds everything needed.
    let wrapper = ctx.upgrade();

    let a = wrapper.upgrade_share(RecordId::from(0), a); // You need to do this for all inputs.
    let b = wrapper.upgrade_share(RecordId::from(1), b);

    // Execute the protocol
    let result = wrapper.context().narrow(&Step::Multiply).multiply(a, b).await?;

    // Validate the values and release the answer.
    wrapper.validate(result).await?
}
```

Something to note here is that the `MaliciousInputValidationStep` implementation of `SubStep` can be extended here.  You currently have: 

* multiply by `r`
* randomness for each input

But you probably also need:

* a narrowing for the malicious protocol execution
* one or more narrowing for the validation steps

Then you have a clean hierarchy of steps, rather than having different types of `SubStep` coexist at the same level of the tree (which will be a problem later).

src/test_fixture/mod.rs
benjaminsavage: I think try_join_all returns a vector, so it's a Vec<Result<(), Error>>

martinthomson: Then maybe you should unwrap() it rather than throwing it away.

benjaminsavage:
Abandoning in favor of #269

PR #261, opened 2022-11-21
Simplify test

martinthomson:
This builds on #253.  View the last commit only.

### DIFF
--- a/src/test_fixture/sharing.rs
+++ b/src/test_fixture/sharing.rs
@@ -10,8 +10,11 @@ use rand::{
 
 use super::{MaliciousShares, ReplicatedShares};
 
-pub trait IntoShares<S> {
-    fn share(self) -> [S; 3];
+pub trait IntoShares<S>: Sized {
+    fn share(self) -> [S; 3] {
+        self.share_with(&mut thread_rng())
+    }
+    fn share_with<R: Rng>(self, rng: &mut R) -> [S; 3];
 }
 
 impl<F> IntoShares<Replicated<F>> for F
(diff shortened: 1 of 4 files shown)


src/test_fixture/world.rs
```diff
@@ -78,3 +87,37 @@ pub fn make(query_id: QueryId) -> TestWorld {
...
+        Standard: Distribution<F>;
+}
+
+#[async_trait]
+impl<I, A> Runner<I, A> for TestWorld
+where
+    I: 'static + IntoShares<A> + Send,
+    A: Send,
+{
+    async fn semi_honest<'a, F, O, H, X>(&'a self, input: I, mut helper_fn: H) -> [O; 3]
```
akoshelev, line 108: this is just awesome ❤️

src/test_fixture/sharing.rs
```diff
@@ -2,13 +2,55 @@ use std::iter::zip;
...
+use rand::thread_rng;
 use rand::{
     distributions::{Distribution, Standard},
     Rng, RngCore,
 };
 
 use super::{MaliciousShares, ReplicatedShares};
 
+pub trait IntoShares<S> {
+    fn share(self) -> [S; 3];
```
akoshelev, line 14: I think it would be better to inject `rng` here. I am still hopeful that we will setup property based testing for IPA and it would be very useful to generate a seed and set up PRNG with with. If that test fails, we could replay it because seed value will be saved: https://altsysrq.github.io/proptest-book/proptest/failure-persistence.html

akoshelev (approved):

PR #263, opened 2022-11-21
Use an array for PRSS in tests

martinthomson:


richajaindce (approved):
Looks good! Lets merge it :)

PR #264, opened 2022-11-22
[Sort Optimization] ApplyInv and Compose to use the same random permutation

richajaindce:
[Sort Optimization] ApplyInv and Compose use the same random permutation in shuffle. The parties receive random permutation and invokes shuffle with reveal. Thus, we can commonalize them: we can shuffle and reveal sort permutation once and share with applyinv and compose.

Regarding security, this reuse doesn't affect it since we are still uniformly randomly permuting the permutation vector.

This is OPTAPPLYINV and OPTCOMPOSE from the paper

This will also help in malicious upgrade since we need to validate before each reveal permutation. And, this PR simplifies the logic


src/protocol/sort/shuffle.rs
```diff
@@ -110,7 +110,7 @@ async fn reshare_all_shares<F: Field>(
 #[allow(clippy::cast_possible_truncation)]
 async fn shuffle_or_unshuffle_once<F: Field>(
     mut input: Vec<Replicated<F>>,
-    random_permutations: (&[u32], &[u32]),
+    random_permutations: &(Vec<u32>, Vec<u32>),
```
benjaminsavage, line 113: Why make this change?

benjaminsavage (approved):
I think I understand. That's a really nice optimization!

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -147,4 +201,51 @@ mod tests {
...
         Ok(())
     }
+
+    #[tokio::test]
+    pub async fn test_shuffle_and_reveal_permutation() {
+        const BATCHSIZE: u32 = 25;
+
+        let mut rng = rand::thread_rng();
+
+        for _ in 0..10 {
```
martinthomson, line 211: Do you really need to do this 10 times?  With 24 items, the odds of the order being the same is $\frac{1}{25!}$, which is small enough.

src/protocol/sort/generate_sort_permutation.rs
```diff
@@ -147,4 +201,51 @@ mod tests {
...
+                BATCHSIZE.try_into().unwrap(),
+                perm1,
+            );
+            let h2_future = shuffle_and_reveal_permutation(
+                ctx2.narrow("shuffle_reveal"),
+                BATCHSIZE.try_into().unwrap(),
+                perm2,
+            );
+
+            let perms_and_randoms: [_; 3] = try_join_all([h0_future, h1_future, h2_future])
```
martinthomson, line 237: join3() was written for this purpose (the one in the fixture, not the one in the future crate)

martinthomson (commented):
A couple of things for a follow-up.

src/protocol/sort/shuffle.rs
richajaindce: aesthetically this is easier to read than passing two pointers each time

martinthomson: It is very different from a functional perspective though.  In order to create the latter argument, you have to move the vector into a tuple, then reference that tuple.  In the former, you only need to create a tuple from two references.

As always, you should prefer to pass slices of things when that is what is needed.  If you need flexibility, there is also `impl IntoIterator<Item = F>`.

richajaindce: #265 Although I havent used the impl since the conversion needed is (Vec<..>, Vec<..>) -> (&[..], &[..]

PR #266, opened 2022-11-22
Bit-Decomposition protocol

taikiy:
This PR implements the bit-decomposition protocol described in Damgård's [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf) "3 Bit-Decomposition".

It takes a secret shared field value [a]_p and outputs its bitwise sharings [a]_B. This process is required for some boolean circuits such as comparing two numbers, which is used by the credit attribution and user-level capping protocols.

The bit-decomposition protocol uses `SolvedBits` to generate random sharings of bits, but `SolvedBits` may abort under some conditions. Currently, the infra freezes if the record_id doesn't start at 0, or panics if the same record_id is used twice. As a workaround, I implemented `RandomBitsGenerator`. It has an internal counter that just counts upwards and used as record IDs, generates a bunch of random-bits sharings from `SolvedBits`, and lets everyone take the sharings from the buffer as needed.

I haven't decided where would be the best place to place the buffer instance though. `ctx` seems like a good place, but `SolvedBits` calls `BitwiseLessThan` protocol which requires helper communications, hence needs `ctx` itself. One way of doing this is we could initialize the generator inside a context when it's first accessed. Another think I'm not sure about is whether we should keep cloning the instance on each `narrow`, or just throw them away.

 Suggestions are much appreciated.


src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,209 @@
...
+        ctx: ProtocolContext<'_, Replicated<F>, F>,
+        record_id: RecordId,
+        a_p: &Replicated<F>,
+    ) -> Result<Vec<Replicated<F>>, Error> {
+        // step 1 in the paper is just describing the input, `[a]_p` where `a ∈ F_p`
+
+        // Step 2. Generate random bitwise shares
+        let r = SolvedBits::execute(ctx.narrow(&Step::GenerateRandomBits), record_id).await?;
+        // Would like to do something like below. Where should I add the `RandomBitsGenerator`?
+        // let r = ctx.random_bits_generator.take_one().await?;
```
taikiy, line 33: Need to place the buffer instance somewhere. Suggestions are much appreciated :)

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,209 @@
...
+        assert_eq!(F::Integer::BITS as usize, result[2].len());
+
+        let bits = (0..result[0].len())
+            .map(|i| validate_and_reconstruct(&result[0][i], &result[1][i], &result[2][i]))
+            .collect::<Vec<_>>();
+
+        Ok(bits)
+    }
+
+    #[ignore]
```
taikiy, line 147: Remove `ignore` once `RandomBitsGenerator` is available.

src/protocol/boolean/bitwise_lt.rs
```diff
@@ -135,11 +135,12 @@ impl BitwiseLessThan {
         a: &[Replicated<F>],
         b: &[Replicated<F>],
     ) -> Result<Replicated<F>, Error> {
-        debug_assert_eq!(a.len(), b.len(), "Length of the input bits must be equal");
-        let mut e = Self::step1(a, b, ctx.narrow(&Step::AXorB), record_id).await?;
+        let (a, b) = align_bit_lengths(a, b);
```
taikiy, line 138: Edit `BitwiseLessThan` to work with inputs with different bit-lengths. `BitDecomposition` needs this.

src/protocol/boolean/bitwise_sum.rs
```diff
@@ -35,12 +36,12 @@ impl BitwiseSum {
         record_id: RecordId,
         a: &[Replicated<F>],
         b: &[Replicated<F>],
-    ) -> Result<Vec<Replicated<F>>, BoxError> {
-        debug_assert_eq!(a.len(), b.len(), "Length of the input bits must be equal");
+    ) -> Result<Vec<Replicated<F>>, Error> {
+        let (a, b) = align_bit_lengths(a, b);
```
taikiy, line 40: Same. This needs to work with inputs with different bit-lengths.

src/protocol/boolean/bitwise_sum.rs
```diff
@@ -149,50 +150,112 @@ mod tests {
...
     #[tokio::test]
-    pub async fn fp_32bit_prime_basic() -> Result<(), BoxError> {
+    pub async fn fp_32bit_prime_basic() -> Result<(), Error> {
         let c = Fp32BitPrime::from;
 
-        assert_eq!(c(0_u32), bits_to_field(&bitwise_sum(c(0), c(0)).await?));
-        assert_eq!(c(1), bits_to_field(&bitwise_sum(c(0), c(1)).await?));
-        assert_eq!(c(1), bits_to_field(&bitwise_sum(c(1), c(0)).await?));
-        assert_eq!(c(2), bits_to_field(&bitwise_sum(c(1), c(1)).await?));
+        assert_eq!(0, bits_to_value(&bitwise_sum(c(0_u32), c(0)).await?));
```
taikiy, line 170: I realized that `BitwiseSum` outputs a sequence of shared bits that is not necessarily in the inputs' field. An overflow (in the inputs' field) will not be rounded; rather the output will have +1 bit to preserve the potential carry. Remove `F::from` from asserts, and use `bits_to_value` to reconstruct values in integers, not in fields.

src/protocol/boolean/prefix_or.rs
```diff
@@ -267,10 +267,10 @@ impl PrefixOr {
         // We can modify this function if we need to support other lengths.
         let l = a.len();
         let lambda: usize = match l {
-            8 => 3,
+            8..=9 => 3,
```
taikiy, line 270: Adjust to accept ranges of bit lengths here as well.

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -0,0 +1,140 @@
...
+    use super::RandomBitsGenerator;
+    use crate::{
+        error::Error,
+        ff::Fp31,
+        protocol::QueryId,
+        test_fixture::{join3, logging, make_contexts, make_world, TestWorld},
+    };
+
+    #[tokio::test]
+    pub async fn basic() -> Result<(), Error> {
```
taikiy, line 96: This test takes about 6 seconds on my machine to complete because it generates about 500 bit shares. Would be great if I can mock `SolvedBits` to return bogus outputs. Mocking is something we'll need to support soon.

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -0,0 +1,139 @@
...
+/// item to return, and `write_pointer` points to the least recent empty slot.
+/// The pointers are `i8`. Incrementing them cause wrap around at the boundary,
+/// which makes the `buffer` hash map a ring buffer.
+#[allow(dead_code)]
+pub struct RandomBitsGenerator<'a, F: Field> {
+    context: SemiHonestContext<'a, F>,
+    buffer: HashMap<i8, RandomBitsShare<F>>,
+    counter: u32,
+    read_pointer: i8,
+    write_pointer: i8,
```
taikiy, line 23: Depends on where I place the generator instance, but most likely need to guard these with the mutex.

src/protocol/boolean/bit_decomposition.rs
benjaminsavage: The commented out lines look good - definitely better than `r.expect()` which will just fail if it aborted.

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,210 @@
...
+        let d_b = BitwiseSum::execute(ctx.narrow(&Step::AddBtoC), record_id, &c_b, &r.b_b).await?;
+
+        // Step 6. p <? d
+        let p_b = local_secret_shared_bits(F::PRIME.into(), ctx.role());
+        let q_p = BitwiseLessThan::execute(ctx.narrow(&Step::IsPLessThanD), record_id, &p_b, &d_b)
+            .await?;
+
+        // Step 7. a bitwise scalar value `f_B = bits(2^l - p)`
+        let l = F::Integer::BITS;
+        let x = 2_u128.pow(l) - F::PRIME.into();
```
benjaminsavage, line 56: I'm trying to figure out if this is another time to use just the number of bits necessary to represent the field values (i.e. 5 bits for Fp31, not 8)

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,210 @@
...
+        let g_b = f_b
+            .into_iter()
+            .map(|f_bit| q_p.clone() * f_bit)
+            .collect::<Vec<_>>();
+
+        // Step 10. [h]_B = [d + g]_B, where [h]_B = ([h]_0,...[h]_(l+1))
+        //
+        // Again, `BitwiseSum` outputs `l + 1` bits. Since [d]_B is already
+        // `l + 1` bit long, [h]_B will be `l + 2`-bit long.
+        let h_b = BitwiseSum::execute(ctx.narrow(&Step::AddDtoG), record_id, &d_b, &g_b).await?;
```
benjaminsavage, line 69: Yeah, I think we should be capable of saving ourselves some work here if we are only adding 5-bit values instead of 8-bit values (for the Fp31 case).

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,210 @@
...
+                [c0.clone(), c1.clone(), c2.clone()],
+                RecordId::from(i),
+                input,
+            )
+            .await?;
+
+            if input.as_u128() == 0 {
+                // if the protocol's input is 0, the output could either be
+                // a bitwise sharing of 0 or the prime (in the integers).
+                let x = bits_to_value(&result);
```
benjaminsavage, line 168: Actually, I think this is evidence of a bug. I don't think it should return a bit-decomposition of the prime. That's not a valid element of `F_p`. The paper doesn't mention this (as far as I can tell).

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    #[ignore]
+    #[tokio::test]
+    pub async fn fp31() -> Result<(), Error> {
+        let world: TestWorld = make_world(QueryId);
+        let ctx = make_contexts::<Fp31>(&world);
+        let [c0, c1, c2] = ctx;
+        let mut rng = rand::thread_rng();
+
+        for i in 0..10 {
```
benjaminsavage, line 156: Instead of randomly generating 10 values, perhaps it would be better to just test a few specific edge cases like: 0, 30, 15, 16, 1. That could be faster than looping 10 times and would yield good unit test coverage.

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,210 @@
...
+                RecordId::from(i),
+                input,
+            )
+            .await?;
+
+            if input.as_u128() == 0 {
+                // if the protocol's input is 0, the output could either be
+                // a bitwise sharing of 0 or the prime (in the integers).
+                let x = bits_to_value(&result);
+                assert!(x == 0 || x == u128::from(Fp32BitPrime::PRIME));
```
benjaminsavage, line 200: Should we just test the zero case specifically? Also, I remain unconvinced that it should yield a bitwise decomposition of the prime in the case the input is zero.

src/protocol/boolean/random_bits_generator.rs
benjaminsavage, line 23: Are you sure you want to use i8 and not u8?

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Takes one `RandomBitsShare` instance out of the buffer. If the number
+    /// of buffered items fall below a threshold, it'll replenish.
+    pub async fn take_one(&mut self) -> Result<RandomBitsShare<F>, Error> {
+        self.refill_if_needed().await?;
+        let r = self
+            .buffer
+            .remove(&self.read_pointer)
+            .expect("unexpected buffer depletion");
+        self.read_pointer = self.read_pointer.wrapping_add(1);
```
benjaminsavage, line 50: What does `wrapping_add(1)` do on a `i8`? When does it wrap around? Does 127 wrap around to -128 or something?

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(r)
+    }
+
+    /// Checks if the stock is short, and replenish as needed. Specifically, it
+    /// will call `SolvedBits` once the stock falls below `REFILL_THRESHOLD`,
+    /// until the `write_pointer` is one place behind the `read_pointer`.
+    async fn refill_if_needed(&mut self) -> Result<(), Error> {
+        // intentional type coercing `i8` -> `u8`
+        #[allow(clippy::cast_sign_loss)]
```
benjaminsavage, line 59: This really makes me wonder about the use of `i8`

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -0,0 +1,139 @@
...
+        if self.write_pointer.wrapping_sub(self.read_pointer) as u8 > Self::REFILL_THRESHOLD {
+            return Ok(());
+        }
+
+        while self.read_pointer.wrapping_sub(self.write_pointer) != 1 {
+            let record_id = RecordId::from(self.counter);
+            let r = SolvedBits::execute(self.context.clone(), record_id).await?;
```
benjaminsavage, line 66: This will generate them sequentially, which will be needlessly slow. I'd suggest generating them in large batches. Basically, aim to fully replenish the buffer. Compute the number of empty slots, and invoke that many in parallel. If some fail, so be it. Just insert the successful ones into the ring buffer.

benjaminsavage (changes requested):
A few necessary changes:
1. We need to invoke the RandomBitsGenerator instead of just failing if the SolvedBits protocol aborts
2. The RandomBitsGenerator should not call await in a loop. Rather, it should attempt to generate a batch of SolvedBits in parallel.
3. I'd rather use u8 instead of i8 for the read and write buffer. This seems highly preferable to me for readability.

src/protocol/boolean/bit_decomposition.rs
taikiy: OK, by shortening the bit lengths, we can reduce multiplications in BitwiseSum * 2, BitwiseLessThan. Is that correct?

If we start both [a] and [r] with 5-bit, that's 3 + 2 + 2 = 7 multiplications reduced per record for Fp31. As for Fp32BitPrime, we only save 1 multiplication?

src/protocol/boolean/bit_decomposition.rs
taikiy: I also thought it was a bug first, but the paper doesn't mention otherwise either... E.g.,

```
let input a = 0, and random bits in F_p r = 25

c = a - b = -25 mod p = 6
d = c + b = 6 + 25 = 31 <- prime
q = p <? d = 0
f = 225 (2^l - p)
g = f * q = 0
h = d + g = 31
```

This happens because BitwiseSum doesn't take the field prime into account. There are other `r` that outputs the same. i.e., 30, 32, 39. If we don't like this behavior, we could implement a "`F` wrapping" BitwiseSum? I'm not sure how that will affect other protocols though.

src/protocol/boolean/random_bits_generator.rs
taikiy: Yes, that is correct. I've changed everything to `u8`.

src/protocol/boolean/bit_decomposition.rs
benjaminsavage: I have a suspicion that this is the cause of the bitwise shares sometimes being larger than the prime.

src/protocol/boolean/bit_decomposition.rs
benjaminsavage: Wow... I think there's actually an error in the paper. The proof says:

> As for the correctness, notice that `c = a - b mod p` and `d = c + b` (in the integers). Therefore `d = a + qp` for some `q ∈ {0,1}`. Since `a ∈ {0, 1, ..., p-1}` it follows that `q = 1 iff p < d`.

That's incorrect. It should say `q = 1 iff p <= d`.  Consider the case where `q == d`. In that case we have `d = 0 + 1*p`. 

So I think to fix this, we need to compute `p <= d` which is the same thing as `!(d < p)` which we can do with just

```
q = 1 - BIT-LT(d, p)
```

src/protocol/boolean/random_bits_generator.rs
taikiy: I was thinking I have to filter the ones that failed, run it again, until all slots are filled. But ignoring those failed would be much simpler. Will update it.

taikiy:
> 1. We need to invoke the RandomBitsGenerator instead of just failing if the SolvedBits protocol aborts

Added the RandomBitsGenerator (RBG for short) inside the Context. Some notes on the design:

* The minimum requirement is to keep the RBG instance alive as long as the context lives, even across clone() and narrow(). Otherwise, every call to RBG will trigger a refetch. Context seems to be the best place to put RBG for now.
* I thought about making RBG a trait like `SecureMul`. That way, I don't need to pass a context when I call `take_one()`. But traits can't have members, so it will cause RBG instance to get generated each time ctx is cloned/narrowed.
* I also thought about making `fn random_bits_generator()` in the `Trait Context` async. Again, this is to call `take_one()` without needing to pass a context. It will make this the only async method in the `Context` and seemed out of place. Maybe that's not too bad?

As a part of this change, I needed to make `TestWorld` generic over `F`. Currently, TestWorld creates new contexts each time a `Runner` is run (i.e., `world.semi_honest(input, |ctx, a| async move {...})`) rather than clone/narrow. This forces RBG to reset and lose the buffer contents. I created a function `supply_rbg_for_tests()` to inject an existing RBG instance. This is not the best design, but supporting singletons in the context would probably take another big change. I'd love to come back to this and improve.

> 2. The RandomBitsGenerator should not call await in a loop. Rather, it should attempt to generate a batch of SolvedBits in parallel.

Fixed. We can further improve this by scheduling a task and continuing to return shares as they are generated.

> 3. I'd rather use u8 instead of i8 for the read and write buffer. This seems highly preferable to me for readability.

Fixed.

benjaminsavage:
I am not sure about putting the random bits generator into the context. Could we just make it a static that can be accessed anywhere? E.g. `RandomBitsGenerator::take_one()`?

Even if we *do* put it into the context, shouldn't it be on the `inner` context? The part that's Arc cloned? 

I'll wait for @martinthomson to review this.

taikiy:
We could create a `const` in the context like this:
```
    pub const RANDOM_BITS_GENERATOR: Lazy<RandomBitsGenerator<F>> =
        Lazy::new(|| RandomBitsGenerator::new());
```
but we'll need to specify a concrete context type to call it (i.e., `SemiHonestContext::RANDOM_BITS_GENERATOR.take_one(...)`). If we make it static for the entire app, we need to specify a concrete type for `F` at compile time. We could have multiple RBGs, but we should probably avoid global variables as much as possible.

`RandomBitsGenerator` wraps the internal buffer inside `Arc`, so we are okay on that end. The comment says `inner`'s `Arc` is for tests, so I expect it will move out of `Arc` at some point. Also, since `ContextInner` is not generic over `F`, I placed it directly under `Context`.

martinthomson:
Don't put this in the context directly, the inner is what this sort of thing is for.  The context (alongside the Step) is cloned lots.  It's a bit annoying to have to have this in what is essentially a global, but I don't see how to avoid that, or at least I have no better ideas.

taikiy:
Moved the RBG to `ContextInner`. I was able to remove that awkward `supply_rbg_for_tests()` method, too.

benjaminsavage (approved):
It's great to have this functionality landed! I've got a few PRs in flight to make it run faster. See #274 and #275

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -0,0 +1,168 @@
+use super::bitwise_sum::BitwiseSum;
+use super::dumb_bitwise_lt::BitwiseLessThan;
```
benjaminsavage, line 2: <3

src/protocol/boolean/solved_bits.rs
```diff
@@ -1,7 +1,7 @@
-use super::bitwise_lt::BitwiseLessThan;
+use super::dumb_bitwise_lt::BitwiseLessThan;
```
benjaminsavage, line 1: <3

PR #265, opened 2022-11-22
Minor fixes from #264 suggested by Martin

richajaindce:

PR #267, opened 2022-11-23
Upgrade all subprotocols of sort

richajaindce:
All subprotocols now take generic Context! Whats left
1. Add malicious tests for secureapplyinv, compose, sort (will add in next PR once #265 and #257 are approved and landed since they can conflict)
2. Validate call in sort


src/protocol/sort/reshare.rs
```diff
@@ -79,26 +88,53 @@ impl<F: Field> Reshare<F> {
     }
 }
 
+/// For malicious reshare, we run semi honest reshare protocol twice, once for x and another for rx and return the results
```
benjaminsavage, line 91: I think that this code has already landed. Can you rebase?

richajaindce: Yes let me do that

src/protocol/sort/reshare.rs
```diff
@@ -100,9 +103,6 @@ impl<F: Field> Reshare<F> for MaliciousContext<'_, F> {
         record_id: RecordId,
         to_helper: Role,
     ) -> Result<Self::Share, Error> {
-        use crate::protocol::context::SpecialAccessToMaliciousContext;
-        use crate::secret_sharing::ThisCodeIsAuthorizedToDowngradeFromMalicious;
-
```
benjaminsavage, line 105: I'm not sure you want to do that. @martinthomson - is there a specific reason why these use statements were formerly located here?

benjaminsavage (approved):
I love this change!! I can't wait to land this. Just that one question.

src/protocol/sort/reshare.rs
martinthomson: Ah, I just wanted to ensure that these were narrowly scoped.  They really shouldn't be moved to where they went here though.

richajaindce: oops i thought they were misplaced.. Let me bring it back here

richajaindce: By mistake i git commit --amend and force pushed the changes :( but it should be in correct place now

Issue #256, opened 2022-11-20
[Bug] wrong permutation indices when task scheduling reordering occurs


akoshelev:
Running concurrency test on sort revealed that we have a reordering issue somewhere.

```rust
    shuttle::check_random(
        || {
            shuttle::future::block_on(execute_sort()).unwrap();
        },
        5,
    );
```

fails with

```bash
test panicked in task 'main-thread'
index out of bounds: the len is 50 but the index is 1942298774
thread 'tests::randomized::sort' panicked at 'index out of bounds: the len is 50 but the index is 1942298774', src/protocol/sort/apply.rs:23:24
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_bounds_check
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:84:5
   3: core::slice::<impl [T]>::swap
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/slice/mod.rs:619:36
   4: raw_ipa::protocol::sort::apply::apply
             at ./src/protocol/sort/apply.rs:23:17
   5: raw_ipa::protocol::sort::compose::compose::{{closure}}
             at ./src/protocol/sort/compose.rs:50:5
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/future/mod.rs:91:19
   7: raw_ipa::protocol::sort::generate_sort_permutation::generate_sort_permutation::{{closure}}
             at ./src/protocol/sort/generate_sort_permutation.rs:69:9
   8: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/future/mod.rs:91:19
   9: <F as futures_core::future::TryFuture>::try_poll
             at /Users/koshelev/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.25/src/future.rs:82:9
  10: <futures_util::future::try_future::into_future::IntoFuture<Fut> as core::future::future::Future>::poll
[...]

PR #268, opened 2022-11-23
Refactor reshare tests

akoshelev:
While working on #256, I noticed that reshare tests consistently fail concurrency test. It was caused by always using RecordId(0) in a loop that terminates if reshare successfully generates unique shares. When loop needs more than 1 iteration it fails on PRSS uniqueness check.

While it was possible to fix that test, I did not really like the probabilistic nature of it and decided to create 2 tests instead. One validates that reshare protocol actually communicates and another one that it is correct.

It was also a good opportunity to use new test API implemented by @martinthomson in #261. It should be used everywhere.

One small change I made to the `TestWorld` to improve ergonomics changes `make_contexts` function to narrow generated contexts to some unique (within this world) step, so the following syntax is made possible

```rust
  for _ in 0..10 {
     world.semi_honest(..., |ctx, v| { ... })
  }

### CODE REVIEW

File: src/protocol/sort/reshare.rs [hunk 1]
```diff
@@ -81,52 +81,77 @@ impl<F: Field> Reshare<F> {
...
+        for &target in Role::all() {
+            let secret = thread_rng().gen::<Fp32BitPrime>();
+            let shares = world
+                .semi_honest(secret, |ctx, share| async move {
+                    let record_id = RecordId::from(0);
+
+                    // run reshare protocol for all helpers except the one that does not know the input
+                    if ctx.role() == target {
+                        // test follows the reshare protocol
+                        ctx.prss().generate_fields(record_id).into()
```
martinthomson, line 113: Won't the effect be the same if you run reshare for this target?  Or, are you trying to test that it doesn't need to participate?

src/protocol/sort/reshare.rs
```diff
@@ -81,52 +81,77 @@ impl<F: Field> Reshare<F> {
...
+                    } else {
+                        Reshare::new(share.clone())
+                            .execute(&ctx, record_id, target)
```
martinthomson, line 116: This still takes a reference to the context, when it shouldn't.

akoshelev: let me merge #257 and rebase this on top of that

src/protocol/sort/reshare.rs
akoshelev: I tried to ensure a naive `Reshare` implementation (that is - take the input and return it as output) won't pass this test. I think that was the reason for `for` loop inside the test before.

martinthomson (approved):

PR #269, opened 2022-11-23
A simpler malicious upgrade (WIP)

martinthomson:
I'm trying to make tests for malicious a little easier to stomach.  This means building a bunch of test infrastructure, but I think that it also means changing the way that inputs are upgraded.

I think that this approach is pretty neat.  You take a semi-honest context and upgrade it by wrapping it in a `MaliciousValidator`:

```rust
let v = MaliciousValidator(ctx);
```

To get a malicious shares, you just upgrade them, passing a record ID for the upgrade process:

```rust
let m_share = v.context().upgrade(record_id, share);
```

Then multiply and what-not with `v.context().multiply()` and so forth.

Note that the upgrade is on the context because you might introduce more inputs later; the record ID space is global though, so maybe that is not ideal.

Then you can validate, passing any `MaliciousReplicated` shares that contain your results:

```rust
let results = v.validate(m_results).await;
```

I've put in some safeguards that should prevent someone from accidentally looking at the values without being validated first.

I still need to migrate a few tests over to this code, so hold tight just a little longer.


src/test_fixture/world.rs
```diff
@@ -122,4 +154,24 @@ where
...
+        O: Send + Debug,
+        M: Send,
+        H: FnMut(MaliciousContext<'b, F>, M) -> R + Send,
+        R: Future<Output = O> + Send,
+        Standard: Distribution<F>,
+    {
+        // self.semi_honest(input, |ctx, args| async {
+        //     let v = MaliciousValidator::new(ctx);
+        //     let m_share = args.upgrade(v.context()).await;
+        //     helper_fn(v.context(), m_share).await
```
akoshelev, line 172: I think we should also call `v.validate()` after?

akoshelev (approved):
this looks good to me and I can't wait to see how you tackle the borrow checker in `malicious`

src/protocol/context/malicious.rs
```diff
@@ -101,23 +117,39 @@ struct ContextInner<'a, F: Field> {
...
+    upgrade_ctx: SemiHonestContext<'a, F>,
+    accumulator: MaliciousValidatorAccumulator<F>,
     r_share: Replicated<F>,
 }
 
 impl<'a, F: Field> ContextInner<'a, F> {
     fn new<B: Borrow<semi_honest::ContextInner<'a>>>(
         source: &B,
-        accumulator: SecurityValidatorAccumulator<F>,
+        upgrade_ctx: SemiHonestContext<'a, F>,
```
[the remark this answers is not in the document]
martinthomson: I'm not following your reasoning here, sorry.

src/protocol/malicious.rs
```diff
@@ -115,40 +122,59 @@ impl<F: Field> SecurityValidatorAccumulator<F> {
 }
 
 #[allow(dead_code)]
-pub struct SecurityValidator<F: Field> {
+pub struct MaliciousValidator<'a, F: Field> {
     r_share: Replicated<F>,
     u_and_w: Arc<Mutex<AccumulatorState<F>>>,
+    protocol_ctx: MaliciousContext<'a, F>,
+    propagate_uw_ctx: SemiHonestContext<'a, F>,
```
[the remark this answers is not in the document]
martinthomson: Yeah, I'm in two minds about that.  I really didn't want to keep the base context around, but it seems like maybe that's not something we can avoid in this case.

src/protocol/malicious.rs
```diff
@@ -115,40 +122,59 @@ impl<F: Field> SecurityValidatorAccumulator<F> {
...
+            check_zero_ctx,
         }
     }
 
     pub fn r_share(&self) -> &Replicated<F> {
         &self.r_share
     }
 
+    /// Get a copy of the context that can be used for malicious protocol execution.
+    pub fn context(&'a self) -> MaliciousContext<'a, F> {
```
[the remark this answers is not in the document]
martinthomson: I considered that, but I think that I'd prefer to force the caller to get the (malicious) context for doing further work.

src/protocol/malicious.rs
```diff
@@ -158,11 +184,11 @@ impl<F: Field> SecurityValidator<F> {
     /// ## Panics
     /// Will panic if the mutex is poisoned
     #[allow(clippy::await_holding_lock)]
-    pub async fn validate(self, ctx: SemiHonestContext<'_, F>) -> Result<(), Error> {
+    pub async fn validate(self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
martinthomson: And just panic if it hasn't been called?  Yeah, probably a good idea.

src/test_fixture/sharing.rs
```diff
@@ -68,19 +94,35 @@ where
...
+/// For upgrading various shapes of replicated share to malicious.
+#[async_trait]
+pub trait IntoMalicious<F: Field, M> {
+    async fn upgrade<'a>(self, ctx: MaliciousContext<'a, F>) -> M;
+}
+
+#[async_trait]
+impl<F: Field> IntoMalicious<F, MaliciousReplicated<F>> for Replicated<F> {
+    async fn upgrade<'a>(self, ctx: MaliciousContext<'a, F>) -> MaliciousReplicated<F> {
+        ctx.upgrade(RecordId::from(0_u32), self).await.unwrap()
```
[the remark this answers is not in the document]
martinthomson: Unfortunately, this code is used in benches, so I can't use your i32 hack there.

src/test_fixture/world.rs
martinthomson: Definitely.

src/protocol/context/malicious.rs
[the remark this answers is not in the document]
akoshelev: I am trying to remove transitive references to the context internals here. 

```mermaid
graph TD;
    ContextInner-->role;
    ContextInner-->upgrade_ctx;
    upgrade_ctx-->role;
```

 You could initialize `ContextInner` entirely from `upgrade_ctx`, `accumulator` and `r_share`, dropping `B`.

martinthomson:
OK, I've managed to work around the borrow checker issue.  It might be that it is a legitimate complaint here, even.  At least, my theory says that it might be stopping the code from doing something bad.  The result is pretty ugly, but I think that it holds together well enough.

Next step is to try this out on a few protocols.

src/protocol/malicious.rs
```diff
@@ -158,16 +189,19 @@ impl<F: Field> SecurityValidator<F> {
...
-        let helper_left = ctx.role().peer(Direction::Left);
+        let channel = self.propagate_uw_ctx.mesh();
+        let helper_right = self.propagate_uw_ctx.role().peer(Direction::Right);
+        let helper_left = self.propagate_uw_ctx.role().peer(Direction::Left);
 
-        let state = self.u_and_w.lock().unwrap();
+        let (u_local, w_local) = {
+            let state = self.u_and_w.lock().unwrap();
+            (state.u, state.w)
+        };
```
benjaminsavage, line 201: What is the benefit of doing this?

src/protocol/malicious.rs
```diff
@@ -115,40 +127,59 @@ impl<F: Field> SecurityValidatorAccumulator<F> {
...
-    }
-
-    pub fn accumulator(&self) -> SecurityValidatorAccumulator<F> {
-        SecurityValidatorAccumulator {
-            inner: Arc::downgrade(&self.u_and_w),
+            u_and_w,
+            protocol_ctx,
+            propagate_uw_ctx,
+            reveal_r_ctx,
+            check_zero_ctx,
```
benjaminsavage, line 170: Why do it this way?

It seems like you're taking just one context, and narrowing it 3 times, only to save all 3 of those values until validate is called.

At the time validate is called however, you still call `narrow`, specifically you call
`check_zero_ctx.narrow(&Step::CheckZero)`

Which makes me wonder: why not just pass in one context, save just one context, and in the call to validate just call narrow a few times?

src/protocol/malicious.rs
martinthomson: This limits the scope of the lock. That turned out to be a big problem building the malicious runner for tests.

martinthomson: Note also that when you overrode the clippy warning, you were burying a problem for later.

src/protocol/malicious.rs
martinthomson: I shouldn't have been narrowing in validate, but I've taken another go at this to drop the overheads a little.

The problem here is that I made `ctx.upgrade()` consume `ctx`, so this needs to reserve some narrowed contexts for later.  I think that is the right model for this, even if `upgrade()` is only an internal thing.

src/protocol/context/malicious.rs
```diff
@@ -96,28 +89,71 @@ impl<'a, F: Field> Context<F> for MaliciousContext<'a, F> {
     }
 }
 
+/// Sometimes it is required to reinterpret malicious context as semi-honest. Ideally
+/// protocols should be generic over `SecretShare` trait and not requiring this cast and taking
+/// `ProtocolContext<'a, S: SecretShare<F>, F: Field>` as the context. If that is not possible,
+/// this implementation makes it easier to reinterpret the context as semi-honest.
+impl<'a, F: Field> AccessSemiHonestContextFromMalicious<'a, F> for MaliciousContext<'a, F> {
+    fn accumulate_macs(self, record_id: RecordId, x: &MaliciousReplicated<F>) {
```
benjaminsavage, line 97: What does this function have to do with "access semi honest context from malicious"? This accumulator isn't present for Semi-Honest contexts?

src/protocol/context/semi_honest.rs
```diff
@@ -32,12 +32,15 @@ impl<'a, F: Field> SemiHonestContext<'a, F> {
     }
 
     #[must_use]
-    pub fn upgrade_to_malicious(
+    pub fn upgrade<S: Substep + ?Sized>(
         self,
-        accumulator: SecurityValidatorAccumulator<F>,
+        protocol_step: &S,
```
benjaminsavage, line 37: I don't love this name. It's very generic and it doesn't tell me how it'll be used.

benjaminsavage: Perhaps something like `mirror_step` or `duplicated_circuit_step`

src/protocol/malicious.rs
```diff
@@ -157,64 +197,64 @@ impl<F: Field> SecurityValidator<F> {
...
-        let r = ctx
-            .narrow(&Step::RevealR)
-            .reveal(RECORD_0, &self.r_share)
-            .await?;
+        let narrow_ctx = self.validate_ctx.narrow(&ValidateStep::RevealR);
+        let r = narrow_ctx.reveal(RECORD_0, &self.r_share).await?;
         let t = u_share - &(w_share * r);
 
-        let is_valid = check_zero(ctx.narrow(&Step::CheckZero), RECORD_0, &t).await?;
+        let check_zero_ctx = self.validate_ctx.narrow(&ValidateStep::CheckZero);
```
benjaminsavage, line 209: It seems awfully wasteful to set up a new communication channel to only ever evaluate a single record. Instead of narrowing 3 times for these 3 steps (propagate u&w, reveal r, check zero) could we just use a single communication channel and make these different records?

src/protocol/malicious.rs
```diff
@@ -245,52 +285,30 @@ pub mod tests {
...
-                let mult_result = a_ctx
+                let m_result = m_ctx
                     .multiply(RecordId::from(0), &a_malicious, &b_malicious)
                     .await?;
 
+                // Save some cloned values so that we can check them.
                 let r_share = v.r_share().clone();
-                v.validate(ctx.narrow("SecurityValidatorValidate")).await?;
-                Ok::<_, Error>((mult_result, r_share))
+                let result = v.validate(m_result.clone()).await?;
```
benjaminsavage, line 300: I don't understand why we need to pass in the multiplication result to the validate function. Validation is over the entire protocol thus far, and we do not necessarily want to reveal anything when we validate a protocol.

src/secret_sharing/malicious_replicated.rs
```diff
@@ -120,13 +135,81 @@ impl<F: Field> Mul<F> for MaliciousReplicated<F> {
...
+//     I: Iterator<Item = MaliciousReplicated<F>>,
+//     F: Field,
+// {
+//     type Target = DowngradeIter<I>;
+//     fn downgrade(self) -> UnauthorizedDowngradeWrapper<Self::Target> {
+//         UnauthorizedDowngradeWrapper(DowngradeIter {
+//             iter: self.into_iter(),
+//         })
+//     }
+// }
```
benjaminsavage, line 196: Do you really want to leave in this commented code?

src/test_fixture/sharing.rs
```diff
@@ -68,19 +83,52 @@ where
...
+/// For upgrading various shapes of replicated share to malicious.
+#[async_trait]
+pub trait IntoMalicious<F: Field, M> {
+    async fn upgrade(self, ctx: MaliciousContext<'_, F>) -> M;
+}
+
+#[async_trait]
+impl<F: Field> IntoMalicious<F, MaliciousReplicated<F>> for Replicated<F> {
+    async fn upgrade<'a>(self, ctx: MaliciousContext<'a, F>) -> MaliciousReplicated<F> {
+        ctx.upgrade(RecordId::from(0_u32), self).await.unwrap()
```
benjaminsavage, line 95: Why do we need this? And why does this work with a hard-coded record_id of zero? If this is just calling the `upgrade` function on the context, can't the caller just do that instead?

src/test_fixture/world.rs
```diff
@@ -125,4 +161,70 @@ where
...
+        // Current theory is that this might allow `helper_fn` to be run
+        // on multiple different threads concurrently, which would be bad.
+        // The long form below ensures that it is only run on one thread,
+        // even if it might move (with `Send`) a few times before it runs.
+
+        #[cfg(exemplary_code)]
+        {
+            self.semi_honest(input, |ctx, share| async {
+                let v = MaliciousValidator::new(ctx);
+                let m_share = share.upgrade(v.context()).await;
```
benjaminsavage, line 189: I find it very strange to invoke `upgrade` on the share - rather than the context. Why not just call:

```
let m_share = v.context().upgrade(RecordId::from(i), share).await;
```

benjaminsavage (approved):
I left a few comments about a couple of places that confuse me a bit, but this looks like a huge improvement. Let's land it!

src/test_fixture/world.rs
martinthomson: Yeah, this is a consequence of the `IntoMalicious` trait having to be implemented on the share.  Maybe we can work something else out, but I can't see how to do this easily.  One thing that makes this harder is that the trait is a test-only thing right now, because it needs to manage record IDs.

src/test_fixture/sharing.rs
martinthomson: The idea is that different tests will use different argument shapes. A multiply will take a two-tuple; the carry protocol takes an array of "bits", some of the protocols use a two-tuple of arrays of bits.

There are two translations that we use in the tests:

1. Convert from raw values into shares.  
2. Convert from a replicated share (or collection thereof) into malicious shares.

These both need an implementation for each shape of argument.  I've defined a trait for both.  Hopefully, we won't need too many different shapes in our tests, because these are a little fiddly to implement.  But once you implement one shape, they tend to be reusable for other tests.

src/protocol/malicious.rs
martinthomson: This is something that Alex and I discussed as a code hygiene thing.  This forces people who have results in `MaliciousReplicated` form to call `validate()` before they can access the results.  Otherwise, there is nothing stopping them from releasing the value of `m_share.x()` without waiting for the validation to complete.

src/protocol/malicious.rs
martinthomson: Maybe.  Later :)

src/protocol/context/malicious.rs
martinthomson: I'll see if I can find a better name.  This is really just all the stuff that we need for internal purposes.

src/protocol/malicious.rs
martinthomson: I should explain more, but the mammoth warning I got when trying to build out something that used this was amazing.  The only clue I got was that it was complaining about `MutexGuard<AccumulatorState<_>>` not implementing `Send`.

PR #271, opened 2022-11-25
Use PRSS sequential randomness to generate permutation instead of ChaCha

akoshelev:
I am on a crusade to eliminate different sources of randomness in IPA to keep PRSS and thread_rng() only. Having those prevents concurrency tests to provide "replay" functionality.

I believe we shouldn't be using any PRNG as our PRSS implementation already provides random generator that is suitable to shuffle vectors.

This change replaces ChaCha RNG with PRSS seq RNG

### DIFF
--- a/src/protocol/sort/compose.rs
+++ b/src/protocol/sort/compose.rs
@@ -86,9 +86,9 @@ mod tests {
             let [ctx0, ctx1, ctx2] = make_contexts(&world);
 
             let sigma_and_randoms: [_; 3] = try_join_all([
-                shuffle_and_reveal_permutation(ctx0.narrow("shuffle_reveal"), sigma.len(), sigma0),
-                shuffle_and_reveal_permutation(ctx1.narrow("shuffle_reveal"), sigma.len(), sigma1),
-                shuffle_and_reveal_permutation(ctx2.narrow("shuffle_reveal"), sigma.len(), sigma2),
+                shuffle_and_reveal_permutation(ctx0.narrow("shuffle_reveal"), BATCHSIZE, sigma0),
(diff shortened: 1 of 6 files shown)


akoshelev:
@richajaindce would be great if you can take a look

src/protocol/sort/compose.rs
```diff
@@ -86,9 +86,9 @@ mod tests {
             let [ctx0, ctx1, ctx2] = make_contexts(&world);
 
             let sigma_and_randoms: [_; 3] = try_join_all([
-                shuffle_and_reveal_permutation(ctx0.narrow("shuffle_reveal"), sigma.len(), sigma0),
-                shuffle_and_reveal_permutation(ctx1.narrow("shuffle_reveal"), sigma.len(), sigma1),
-                shuffle_and_reveal_permutation(ctx2.narrow("shuffle_reveal"), sigma.len(), sigma2),
+                shuffle_and_reveal_permutation(ctx0.narrow("shuffle_reveal"), BATCHSIZE, sigma0),
+                shuffle_and_reveal_permutation(ctx1.narrow("shuffle_reveal"), BATCHSIZE, sigma1),
+                shuffle_and_reveal_permutation(ctx2.narrow("shuffle_reveal"), BATCHSIZE, sigma2),
             ])
```
martinthomson, line 92: It might be good to avoid unrelated changes like this in future, but I agree that this is iffy.

martinthomson (approved):

martinthomson:
Rebasing here might be fun, but it should be strictly mechanical.  Land it once you are OK.

src/protocol/sort/compose.rs
akoshelev: yea apologies for that - should've at least mentioned this in the description.

PR #278, opened 2022-11-28
Introduce concurrency testing framework

akoshelev:
This is mostly a mechanical change to set up IPA code to support [Shuttle](https://docs.rs/shuttle/latest/shuttle/) framework. The main change is to replace std imports with shuttle if `shuttle` feature is enabled. Unfortunately that means that unit tests need to use additional flag that disables them if `shuttle` is enabled. 

Shuttle already helped to reveal a concurrency bug in infra (#256) and I plan to onboard IPA to it once the whole protocol is ready. 

There are two new concurrency tests added - they do not pass and require a fix in Infra. I just want to make sure that change gets properly reviewed and it is not hidden behind a pile of refactoring. I will be sending the fix after.


src/tests/infra.rs
```diff
@@ -0,0 +1,134 @@
...
+                                    let record_id = RecordId::from(i);
+                                    left_channel
+                                        .send(right_peer, record_id, share.left())
+                                        .await
+                                        .unwrap();
+                                    right_channel
+                                        .send(right_peer, record_id, share.right())
+                                        .await
+                                        .unwrap();
+                                }
```
benjaminsavage, line 42: This seems exactly like `Reveal`. I'm not sure what you're doing here. Why not just call reveal?

PR #280, opened 2022-11-29
User-level DP Capping

taikiy:
## UPDATE
I fixed the edge case I'm using for the test.

* "User Level Simple Capping - Optimized Parallel Approach. Step 5"
`d = cap - next.helper_bit * (cap + next.compare_bit * (cap - next.current_contribution))`
The last subtraction should be swapped.
`d = cap - next.helper_bit * (cap + next.compare_bit * (next.current_contribution - cap))`

* "User Level Simple Capping - Optimized Parallel Approach. Step 5"
`final_credit = d + compare_bit * (final_credit - d)`
In MP-SPDZ code, everything has "base = 1". I've changed to to start from index 0.

Now, the result is correctly capped, and the cap is enforced across all events with the same match key.

**Cap = 18**
image

## Overview
This PR implements the "User Level Simple Capping - Optimized Parallel Approach" described in this document: https://docs.google.com/document/d/1mswffCtW4fnpCeueq4w-nQ06w11zYwKY4b9iflukLYc/

This protocol takes output values from `AccumulateCredit` protocol, looks at credit values of each event and caps each user's total contribution at a specified amount given by a query.

There are cases where this logic outputs total contribution values that exceeds the cap (one example given below). I'm in discussion with the researchers to see if this is expected, and if not how we can fix it. For now, we should be okay to land this as-is, as a part of our effort to run E2E IPA integration test, as long as we know the input and the results produced by the current implementation.

## Example
Notations:
* `is_trigger_bit` = 0 if the event is a source event
* `helper_bit` = 1 if the previous event has the same match key
* `accumulated_credit` values are computed in the previous step; "AccumulateCredit" protocol
* `final_credit` is the result from the this `CreditCapping` step

The cap for this example is set at **20**. The light/dark blue rows are source/trigger events from match key "5678". Notice that there are two sequences, or blocks, of source event(s) followed by trigger events. The first block = row 12..14, and the second block = row 15..18.

The first block of source event (row 12) has the final_credit 20, which is larger than its original value 15. I personally think it should not exceed the original value, but this depends on the definition of attribution models.

The second dark blue source events (rows 15 and 16) have final_credit 0 and 10, which is not correct. It should be either both 0, if row 12 is 20, or should be 0 and 5 if row 12 is 15.

Typically, a last click followed by one or more conversions takes the credit, and if new click(s) happens, previous attributions are finalized and a new attribution window will begin. So, without DP Capping, the result for this example should be 15 at row 12 and 0, 10 at rows 15 and 16.

If the DP Capping follows the same rule, the result should be 15 at row 12, and 0 and 5 at rows 15 and 16.

image


src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+            , base = 1)
+        */
+        // In the above logic, all vectors have "base = 1". Why are we skipping final_credits[0]??
+
+        for i in 0..(num_rows - 1) {
+            let a = cap.clone() - &current_contribution[i + 1].clone();
+            let b = cap.clone()
+                + &ctx
+                    .narrow(&Step::FinalCreditsSourceContribution)
+                    .multiply(RecordId::from(i), &compare_bits[i + 1].clone(), &a)
```
benjaminsavage, line 241: I think that if you switch the order of arguments you can avoid a clone:
```
    .multiply(RecordId::from(i), a, &compare_bits[i+1])
```

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+    ) -> Result<(Replicated<F>, Replicated<F>), Error>
+    where
+        F: Field,
+        H: FnMut(
+            SemiHonestContext<'a, F>,
+            RecordId,
+            Replicated<F>,
+            Replicated<F>,
+            Replicated<F>,
+        ) -> R,
```
benjaminsavage, line 279: This is really hard to understand. What do all these function arguments mean?

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+        protocol::batch::Batch,
+        protocol::{attribution::AttributionInputRow, QueryId},
+        test_fixture::{share, Reconstruct, TestWorld},
+    };
+    use rand::{distributions::Standard, prelude::Distribution, rngs::mock::StepRng};
+    use std::iter::zip;
+    use tokio::try_join;
+
+    fn generate_shared_input<F: Field>(
+        input: &[[u128; 4]],
```
benjaminsavage, line 364: What is going on here?

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+        for (i, expected) in EXPECTED.iter().enumerate() {
+            let v = (
+                &result.0[i].credit,
+                &result.1[i].credit,
+                &result.2[i].credit,
+            )
+                .reconstruct();
+            assert_eq!(v.as_u128(), *expected);
+        }
+    }
```
benjaminsavage, line 476: Can we please use the new test framework?

benjaminsavage (changes requested):

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -176,30 +166,32 @@ impl<'a, F: Field> AccumulateCredit<'a, F> {
...
         }
 
-        let credit_future =
-            ctx.narrow(&Step::BTimesSuccessorCredit)
-                .multiply(record_id, &b, &successor.credit);
+        let credit_future = ctx.narrow(&Step::BTimesSuccessorCredit).multiply(
+            record_id,
+            &b,
+            &successor.interaction_bit,
+        );
```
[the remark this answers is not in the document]
taikiy: I extracted out this interaction pattern logic and generalized it. I changed the field name to `interaction_value` since it could be field values. Otherwise, there's no change to the underlying logic.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    async fn compute_current_contribution<'a, F: Field>(
+        ctx: SemiHonestContext<'a, F>,
+        input: &Batch<CreditCappingInputRow<F>>,
+        stop_bits: Batch<Replicated<F>>,
+        current_contribution: Batch<Replicated<F>>,
+    ) -> Result<Batch<Replicated<F>>, Error> {
+        let mut stop_bits = stop_bits;
+        let mut current_contribution = current_contribution;
+        #[allow(clippy::cast_possible_truncation)]
```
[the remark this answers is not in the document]
taikiy: Fixed

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+        let num_rows = input.len() as RecordIndex;
+
+        let mut iteration_step = IterStep::new("current_contribution_iteration", 0);
```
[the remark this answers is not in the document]
taikiy: Eliminated `IterStep` from the repo

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,477 @@
...
+
+        Ok(current_contribution.clone())
+    }
+
+    async fn compute_compare_bits<F: Field>(
+        ctx: SemiHonestContext<'_, F>,
+        current_contribution: &Batch<Replicated<F>>,
+        cap: u32,
+    ) -> Result<Batch<Replicated<F>>, Error> {
+        let cap = local_secret_shared_bits(cap.into(), ctx.role());
```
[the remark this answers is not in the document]
taikiy: Added as TODO

src/protocol/attribution/credit_capping.rs
taikiy: Fixed. Thanks!

src/protocol/attribution/credit_capping.rs
taikiy: Created a struct that holds these values and gave them better names.

src/protocol/attribution/credit_capping.rs
taikiy: This method creates input shares for attribution steps. Added a comment to this method.

src/protocol/attribution/credit_capping.rs
taikiy: I'll work on this in the following diff. Added TODO

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,454 @@
...
+    {
+        let end = num_rows - step_size;
+        let mut interaction_futures = Vec::with_capacity(end as usize);
+
+        let c = ctx.narrow(&InteractionPatternStep::Depth(depth));
+
+        // for each input row, create a future to execute secure multiplications
+        for i in 0..end {
+            let current = InteractionPatternInputRow {
+                is_trigger_bit: input[i].is_trigger_bit.clone(),
```
benjaminsavage, line 135: It's unfortunate that we need to clone these if they are not used.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,454 @@
...
+        }
+
+        let results = try_join_all(interaction_futures).await?;
+
+        // accumulate the contribution from this iteration
+        results
+            .into_iter()
+            .enumerate()
+            .for_each(|(i, (credit, stop_bit))| {
+                current_contribution[i] = &current_contribution[i] + &credit;
```
benjaminsavage, line 168: Can we use += operator?

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,454 @@
...
+    )
+    .await?;
+    Ok(final_credits.try_into().unwrap())
+}
+
+async fn compute_current_contribution<'a, F: Field>(
+    ctx: SemiHonestContext<'a, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    stop_bits: Batch<Replicated<F>>,
+    current_contribution: Batch<Replicated<F>>,
```
benjaminsavage, line 110: add mut in front of this.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,454 @@
...
+                RecordId::from(i),
+                current,
+                successor,
+                |ctx, record_id, input| async move {
+                    ctx.narrow(&Step::CurrentContributionBTimesSuccessorCredit)
+                        .multiply(record_id, &input.b, &input.sibling)
+                        .await
+                },
+                depth == 0,
+            ));
```
benjaminsavage, line 158:
```
interaction_futures.push(async move {
  b = computeB(
        &stop_bits[i], 
        &input[i + step_size].helper_bit,
  );
  try_join(
      ctx.narrow(&Step::CurrentContributionBTimesSuccessorCredit)
           .multiply(
               record_id, 
               &b, 
               &current_contribution[i + step_size],
          ),
      compute_stop_bit(b, &stop_bits[i + step_size]),
  }).await
);
```

benjaminsavage (changes requested):
I'd like to eliminate the `InteractionPatternInputRow`, `InteractionClosureInput`, and friends

taikiy:
Addressed @benjaminsavage's comments.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+    // Step 2. Compute the current_contribution for each event.
+    //
+    // We follow the approach used in the `AccumulateCredit` protocol. It's a
+    // reversed Prefix Sum of `final_credits`.
+    //
+    let current_contribution = compute_current_contribution(
+        ctx.clone(),
+        input,
+        &mut stop_bits.clone(),
+        &mut final_credits.clone(),
```
benjaminsavage: stop_bits.clone(),
final_credits.clone(),

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+    )
+    .await?;
+    Ok(final_credits.try_into().unwrap())
+}
+
+async fn compute_current_contribution<'a, F: Field>(
+    ctx: SemiHonestContext<'a, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    stop_bits: &mut Batch<Replicated<F>>,
+    current_contribution: &mut Batch<Replicated<F>>,
```
benjaminsavage: stop_bits: mut Batch<Replicated<F>>,
current_contribution: mut Batch<Replicated<F>>,

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+    //
+    // * `stop_bits` is used to determine when to stop the computation
+    // * `final_credits` will have credit values of only source events
+    //
+    let one = Replicated::one(ctx.role());
+    let stop_bits: Batch<Replicated<F>> = repeat(one.clone())
+        .take(input.len())
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap();
```
benjaminsavage: Let's just move this into `compute_current_contribution` and eliminate that function argument.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+            }),
+    )
+    .await?;
+    Ok(final_credits.try_into().unwrap())
+}
+
+async fn compute_current_contribution<'a, F: Field>(
+    ctx: SemiHonestContext<'a, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    stop_bits: &mut Batch<Replicated<F>>,
```
benjaminsavage: eliminate this function argument.

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+                //
+                // b = stop_bit * successor.helper_bit
+                // current_contribution += b * successor.current_contribution
+                // stop_bit = b * successor.stop_bit
+
+                let b = compute_b_bit(
+                    c.narrow(&Step::BTimesStopBit),
+                    record_id,
+                    current_stop_bit,
+                    &input[i + step_size].helper_bit,
```
benjaminsavage, line 134: Would look more consistent to pull this out above and call it `sibling_helper_bit`

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+                    depth == 0,
+                )
+                .await
+                .unwrap();
```
benjaminsavage: .await?

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+    ctx: SemiHonestContext<'_, F>,
+    record_id: RecordId,
+    b_bit: &Replicated<F>,
+    sibling_stop_bit: &Replicated<F>,
+    first_iteration: bool,
+) -> Result<Replicated<F>, Error> {
+    // Since `compute_b_bit()` will always return 1 in the first found, we can
+    // skip the multiplication in the first round.
+    let stop_bit_future = if first_iteration {
+        futures::future::Either::Left(futures::future::ok(b_bit.clone()))
```
benjaminsavage, line 198: return Ok(b_bit.clone());

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,420 @@
...
+    } else {
+        futures::future::Either::Right(ctx.multiply(record_id, b_bit, sibling_stop_bit))
```
benjaminsavage, line 200: return ctx.multiply(record_id, b_bit, sibling_stop_bit).await;

benjaminsavage (changes requested):

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,413 @@
...
+    .await?
+    .try_into()
+    .unwrap();
+    Ok(compare_bits)
+}
+
+async fn compute_final_credits<F: Field>(
+    ctx: SemiHonestContext<'_, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    current_contribution: &Batch<Replicated<F>>,
```
benjaminsavage, line 253: prefix_summed_contributions

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+        &mut original_credits,
+        cap,
+    )
+    .await?;
+
+    let final_credits: Batch<CreditCappingOutputRow<F>> = input
+        .iter()
+        .enumerate()
+        .map(|(i, x)| CreditCappingOutputRow {
+            is_trigger_bit: input[i].is_trigger_bit.clone(),
```
benjaminsavage, line 60: I'd rather remove this (if not needed)

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+async fn mask_source_credits<F: Field>(
+    input: &Batch<CreditCappingInputRow<F>>,
+    ctx: SemiHonestContext<'_, F>,
+) -> Result<Batch<Replicated<F>>, Error> {
+    let final_credits = try_join_all(
+        input
+            .iter()
+            .zip(zip(
+                repeat(ctx.narrow(&Step::MaskSourceCredits)),
+                repeat(Replicated::one(ctx.role())),
```
benjaminsavage, line 81: ctx.share_of_one()

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+    .await?;
+    Ok(final_credits.try_into().unwrap())
+}
+
+async fn credit_prefix_sum<'a, F: Field>(
+    ctx: SemiHonestContext<'a, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    mut original_credits: Batch<Replicated<F>>,
+) -> Result<Batch<Replicated<F>>, Error> {
+    let one = Replicated::one(ctx.role());
```
benjaminsavage, line 98: share_of_one()

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+        let capped_credit = if_else(
+            ctx.narrow(&Step::IfCurrentExceedsCapOrElse),
+            record_id,
+            current_prefix_summed_credit_exceeds_cap,
+            &remaining_budget,
+            original_credit,
+        )
+        .await?;
+
+        original_credits[i] = capped_credit;
```
benjaminsavage, line 308: push into a new thing

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+    .unwrap();
+    Ok(result)
+}
+
+async fn compute_final_credits<F: Field>(
+    ctx: SemiHonestContext<'_, F>,
+    input: &Batch<CreditCappingInputRow<F>>,
+    prefix_summed_credits: &Batch<Replicated<F>>,
+    exceeds_cap_bits: &Batch<Replicated<F>>,
+    original_credits: &mut Batch<Replicated<F>>,
```
benjaminsavage, line 252: let's not mutate this

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+    //
+    let exceeds_cap_bits =
+        is_credit_larger_than_cap(ctx.clone(), &prefix_summed_credits, cap).await?;
+
+    //
+    // 4. Compute the `final_credit`
+    //
+    // We compute capped credits in the method, and writes to `original_credits`.
+    //
+    compute_final_credits(
```
benjaminsavage, line 46: let final_credits = compute_final_credits(

src/protocol/attribution/credit_capping.rs
```diff
@@ -0,0 +1,422 @@
...
+        .unwrap();
+
+    Ok(final_credits)
+}
+
+async fn mask_source_credits<F: Field>(
+    input: &Batch<CreditCappingInputRow<F>>,
+    ctx: SemiHonestContext<'_, F>,
+) -> Result<Batch<Replicated<F>>, Error> {
+    let final_credits = try_join_all(
```
benjaminsavage, line 76: Maybe call this `masked_credits`

benjaminsavage (approved):
This looks pretty good. Looking forward to landing this!

PR #281, opened 2022-11-29
flatten the modulus conversion code

martinthomson:
This moves to using the new test infrastructure for some of the modulus conversion and sorting functions that couldn't do that before.

More conversions need to happen, but this should unstick others.


benjaminsavage (approved):
Perfect. Let's land it.

PR #284, opened 2022-11-29
Enable test world to print metrics per step

akoshelev:
Brought `metrics-runtime` crate because it supports scoped sinks, so step strings can be shared and stored once. Moreover it provides a nice hierarchy where `Gateway` for example does not need to know the helper role. All metrics emitted from the `Gateway` will be prefixed with the role.

There was quite a bit of plumbing to get the sink inside `IndexedEndpoint`, it looks a bit suboptimal to do it this way.

To force tests to print these metrics `RUST_LOG=raw_ipa::test_fixture::metrics=TRACE` must be set

### Sample output

```bash
2022-11-29T08:00:19.357701Z TRACE raw_ipa::test_fixture::metrics: Helper.Step,Records sent,Indexed PRSS,Sequential PRSS
2022-11-29T08:00:19.357760Z TRACE raw_ipa::test_fixture::metrics: H1.protocol/run-0/sort0/bit_permute,100,100,0
2022-11-29T08:00:19.357773Z TRACE raw_ipa::test_fixture::metrics: H1.protocol/run-0/sort0/mod_conv/xor1,50,50,0
2022-11-29T08:00:19.357781Z TRACE raw_ipa::test_fixture::metrics: H1.protocol/run-0/sort0/mod_conv/xor2,0,50,0
```


src/protocol/prss.rs
```diff
@@ -85,7 +88,9 @@ impl IndexedSharedRandomness {
         {
             self.used.insert(index);
         }
-
+        if let Some(ref sink) = self.metrics_sink {
+            sink.clone().increment_counter(INDEXED_PRSS_GENERATED, 1);
```
martinthomson, line 92: why clone?

src/test_fixture/mod.rs
```diff
@@ -39,23 +42,55 @@ pub fn narrow_contexts<C: Debug + Context<F, Share = S>, F: Field, S: SecretShar
...
-
-    [p1, p2, p3]
+#[derive(Default)]
+pub struct ParticipantSetup {
+    metrics_sink: Option<Sink>,
+}
+
+impl ParticipantSetup {
+    #[must_use]
+    pub fn new_with(metrics_sink: Sink) -> Self {
```
martinthomson, line 52: just "new" is fine

```suggestion
    pub fn new(metrics_sink: Sink) -> Self {
```

src/protocol/prss.rs
```diff
@@ -449,7 +479,7 @@ pub mod test {
     #[test]
     fn three_party_values() {
         const IDX: u128 = 7;
-        let [p1, p2, p3] = make_participants();
+        let [p1, p2, p3] = ParticipantSetup::default().into_participants();
```
martinthomson, line 482: Maybe you can make a wrapper function for this file, so that you don't need to change all those lines.

src/test_fixture/metrics.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::collections::{BTreeMap, HashMap};
+
+/// TODO: this function is horrible and I want to fix it
+pub fn print_metrics(snapshot: Snapshot) {
+    let snapshot = snapshot.into_measurements();
+    let mut metrics_table = BTreeMap::<_, HashMap<&'static str, u64>>::new();
+    for (key, measurement) in snapshot {
+        let Measurement::Counter(metric_value) = measurement else {
+            continue
+        };
```
martinthomson, line 13: I have never seen this allocation form before.  I don't think that I like it very much, but maybe it will just take some getting used to.

src/test_fixture/metrics.rs
```diff
@@ -0,0 +1,49 @@
...
+
+        if let Some((k, v)) = if key_name.contains(RECORDS_SENT) {
+            Some((RECORDS_SENT, metric_value))
+        } else if key_name.contains(INDEXED_PRSS_GENERATED) {
+            Some((INDEXED_PRSS_GENERATED, metric_value))
+        } else if key_name.contains(SEQUENTIAL_PRSS_GENERATED) {
+            Some((SEQUENTIAL_PRSS_GENERATED, metric_value))
+        } else {
+            None
+        } {
```
martinthomson, line 24: ??!?

martinthomson (commented):
I assume that the security advisory is something unrelated, but can you look into that too?

src/protocol/prss.rs
akoshelev: the design of the `Sink` is a bit weird, `increment_counter` requires an exclusive reference, despite the fact that it claims to be cheaply cloneable

src/test_fixture/metrics.rs
akoshelev: I said it is going be horrible :(

akoshelev:
I would also need to get rid of `metrics_tracing_context` because of this change (we no longer use labels from spans)

akoshelev:
I decided not to introduce a feature flag for metrics - the code would be full of duplicate methods that take additional `sink` argument and are hidden behind the feature flag. I am thinking that metrics sink is a good idea - queries running it parallel will be able to emit metrics independently and writing unit tests will be much much easier with dedicated sinks per test world.

akoshelev:
> I assume that the security advisory is something unrelated, but can you look into that too?

unfortunately it is legit. I just looked at it and this new crate is no longer maintained. I guess I need to start from scratch or implement something else

akoshelev:
closing as metrics have been merged as part of #307

PR #286, opened 2022-11-29
Small improvement to the attribution test case readability

taikiy:
I hope this makes the test case a bit easier to read.

This unit test still doesn't use `world.semi_honest()` since that requires a bit more complex `IntoShares` logic. I'll try to work on it when I finish other high-pri tasks.


martinthomson (approved):
Time to move to the new framework for this stuff.

PR #287, opened 2022-11-29
[Sort + Accumulate] Reshare objects which have Replicated<F> members

richajaindce:
The objects would need to implement
1. Resharable trait
For the tests to work, they would need to implement 
1. Reconstruct trait
2. IntoShares trait


src/protocol/attribution/accumulate_credit.rs
```diff
@@ -326,4 +373,24 @@ mod tests {
             assert_eq!(v.as_u128(), *expected);
         }
     }
+
+    #[tokio::test]
+    pub async fn test_resharable() {
+        let mut rng = rand::thread_rng();
```
martinthomson, line 379:
```suggestion
        let mut rng = crate::rand::thread_rng();
```

martinthomson (approved):

martinthomson (approved):

PR #288, opened 2022-11-29
End-To-End Network layer

thurstonsand:
DRAFT: there is an outstanding bug with the instantiation of PRSS that is fixed here: https://github.com/thurstonsand/raw-ipa/pull/8

Creates the HttpHelper that holds necessary global state for the long-running MPC instance that can communicate over HTTP. Exposes functionality for creating and executing a query


src/net/discovery/mod.rs
```diff
@@ -40,5 +41,5 @@ pub mod peer {
 /// Any potential failures should be captured in the initialization of the implementer.
 #[allow(clippy::module_name_repetitions)] // following standard naming convention
 pub trait PeerDiscovery {
-    fn peers(&self) -> &[peer::Config; 3];
+    fn peers(&self) -> [peer::Config; 3];
```
akoshelev, line 44: I think we should keep the previous interface. You can clone at the client side if you need to but after our last conversation I thought we don't need t hat

src/net/mod.rs
```diff
@@ -5,19 +5,17 @@ mod client;
...
 pub use server::tls_config_from_self_signed_cert;
-pub use server::{BindTarget, MessageSendMap, MpcHelperServer};
+pub use server::{BindTarget, MessageSendMap, MpcHelperServer, MpcHelperServerError};
 use std::collections::HashMap;
 use std::str::FromStr;
 
 use crate::{
     helpers::{network::ChannelId, MESSAGE_PAYLOAD_SIZE_BYTES},
-    net::server::MpcHelperServerError,
+    sync::{Arc, Mutex},
```
akoshelev, line 18: you would want to rebase this on top of main branch

thurstonsand: i believe it is already. is there something in particular that should be different?

src/net/discovery/mod.rs
thurstonsand: I thought it would make more sense to switch to owned so that way, the `HttpHelper` can own the peer config, and pass it out as reference to queries when they need them. If I switch this back to a reference, then the `HttpHelper` must hold the `PeerDiscovery` itself, which means it needs to carry a generic parameter with it. I thought this was a cleaner approach while still having no copies.

What is the advantage to holding the `PeerDiscovery` vs the `peer:Config`?

src/net/mod.rs
akoshelev: ah, you're right. for some reason I saw it as std::sync

src/helpers/http/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        let port = this_conf.http.origin.port().unwrap();
+        let target = BindTarget::Http(format!("127.0.0.1:{}", port.as_str()).parse().unwrap());
+        tracing::info!("starting server; binding to port {}", port.as_str());
+        self.server.bind(target).await
+    }
+
+    /// adds a query to the running server so that it knows where to send arriving data
+    /// # Errors
+    /// if a query has been previously added
+    pub fn query(&self, query_id: QueryId) -> Result<Gateway, Error> {
```
akoshelev, line 66: I looked at this interface again and I am not sure I understand why it needs to be like this. When you have a new query, the thing I would want to start the protocol is context, not gateway. why we need to keep two separate methods?

src/net/discovery/mod.rs
thurstonsand: > aren't you making a fresh copy when peers is called?

yes, but `peers()` is called only once, by `HttpHelper`. it is passed around by reference from there. I don't think there'd be anything that would need access to `peer_discovery` that wouldn't have access to `HttpHelper`.

> Components lazily grab the whole config only to pick a couple of configuration options later.

would a better model be to divide up the config into areas of interest? so you'd have a config divided up into "http" vs "prss" vs anything else?

as of now, it's divided up into "h1", "h2", "h3", but that can be changed

src/net/server/mod.rs
```diff
@@ -234,6 +236,19 @@ impl MpcHelperServer {
             .route("/echo", get(handlers::echo_handler))
     }
 
+    /// Adds a mapping between [`QueryId`] and [`HttpNetwork`] so that the server knows where to
+    /// forward arriving data.
+    /// # Errors
+    /// if a query has been previously added
```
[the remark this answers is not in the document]
thurstonsand: yes: https://github.com/thurstonsand/raw-ipa/blob/combine-all/src/net/server/mod.rs#L372

src/helpers/http/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    /// binds server as defined in the peers config list
+    /// # Panics
+    /// if peers config does not specify port
+    pub async fn bind(&self) -> (SocketAddr, JoinHandle<()>) {
+        let this_conf = &self.peers[self.role];
+        let port = this_conf.http.origin.port().unwrap();
+        let target = BindTarget::Http(format!("127.0.0.1:{}", port.as_str()).parse().unwrap());
```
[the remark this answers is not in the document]
thurstonsand: my thinking was that you could share config between all helpers. so config would contain external ips of all helpers, but when a helper is starting, it wouldn't bind to its external ip, but localhost.

src/helpers/http/mod.rs
thurstonsand: main thing is the context takes a reference to a gateway, not its value. where would that value be kept if I returned context directly from this `query` fn?

src/net/discovery/mod.rs
akoshelev: > yes, but peers() is called only once, by HttpHelper. 

My worry is that this design is based on assumptions that may no longer be true in the future. I would much rather prefer a cleaner API that allows inefficient actors (cloning at the callsite) if we really need to. Maybe I misunderstand something but I don't think we even need that here. 

> would a better model be to divide up the config into areas of interest? so you'd have a config divided up into "http" vs "prss" vs anything else?

To some extent it would help and I believe you already have it in the config (I see `http` and `prss` sections there). There might be some hiccups later when you may want to use config value across multiple layers and components, but otherwise it should work. The key part imo that components should only ask for things they need and don't get anything extra. Note that there is a way to achieve that with no extra structs being created by using traits,

thurstonsand: ok, I get your point about this now. I will modify.

I can also separate the config into different concerns if that would be more clear. don't think we'll need any traits for now, but possibly in the future.

That being said, I will separate the config within the other branch https://github.com/thurstonsand/raw-ipa/pull/8 because that branch actually deletes all of the prss config stuff since it is no longer necessary

thurstonsand: addressed.

you can see the separation of concerns change here: https://github.com/thurstonsand/raw-ipa/pull/8/commits/4e800f7ec5134bb703ee77da25343ee2410f39ec

src/helpers/http/mod.rs
akoshelev: today we discussed the need for clients to initiate new query by running prss protocol first. to do that they would need to borrow the gateway, so we should keep it somewhere. 

Can you briefly summarize here the responsibilities of each component that you're building here? I see MpcHelperServer and HttpHelper but not sure how is everything is going to work together as an application. Who is responsible of keeping track of running queries and executing prss negotiation to create context?

thurstonsand: Here's what my idea is:

we are creating a new API that a "report collector" (or test runner) will call to initiate a new query. That request handling will create the query, initiate prss, and run the protocol. everything would be kept as local variables in this context.

since we don't want to keep the request open after we empty the body, we could just spawn a new task that continues on after the body is emptied. Or maybe just send the body on a channel back to the `HttpHelper` that has an event loop that spawns the task. That might be easier since the event loop could have easy access to the functions on `HttpHelper`.

One piece that i'm not sure about is how do we handle multiple steps of the protocol? If we want to run those steps simultaneously, you'd need a way to run them independently, so maybe spawn a task for each step?

If any failure happens in the middle, then you would need a way to stop all spawned tasks. it would be possible with an event loop per task that watches for a close signal, and the failing task sends the signal to all others.

An alternative is to add a map to the `HttpHelper`. something like `query_id -> runtime_components`. so all of the spawned tasks' `JoinHandle`s are stored in the `runtime_components`. Now, if one of the tasks fails, there could be a mechanism to simply `abort` all of the handles.

In either case, the piece responsible for initiating the query would effectively be the request handler for the "create query" request.

thurstonsand:
I've discussed with Alex about next steps from here. We have a small redesign that removes the concept of `HttpHelper` and moves all of those fields in the `MpcHelperServer`. That work has already started in a branch off of this branch, so for now, let's merge this into main, and I will follow up with another PR to implement that redesign.

So, while the functions will remain largely unchanged (but moved), the `HttpHelper` struct itself will be removed

src/bin/helper.rs
```diff
@@ -14,6 +19,10 @@ struct Args {
     #[clap(flatten)]
     logging: Verbosity,
 
+    /// Indicates which role this helper is
+    #[arg(short, long, value_enum)]
+    role: Role,
```
akoshelev, line 24: this needs to be removed too, because role will be assigned per query

akoshelev (approved):

PR #289, opened 2022-11-30
Upgrade specialized mul to malicious

benjaminsavage:
Like it says on the tin


src/protocol/mul/semi_honest.rs
```diff
@@ -143,3 +298,112 @@ mod tests {
...
+            .await;
+        assert_eq!(a * b, result.reconstruct());
+    }
+
+    #[tokio::test]
+    async fn specialized_1_parallel() {
+        const COUNT: usize = 10;
+        let world = TestWorld::new(QueryId);
+
+        let mut rng = rand::thread_rng();
```
martinthomson, line 338: Please use the right RNG.  `crate::rand::{thread_rng, Rng}`

martinthomson (approved):

PR #292, opened 2022-11-30
Move RandomBitsGenerator out of context

benjaminsavage:
This was awfully messy being in context. Given that we currently only expect the user contribution capping step to use this, let's just have that protocol generate this object and pass it in.

### DIFF
--- a/src/protocol/boolean/bit_decomposition.rs
+++ b/src/protocol/boolean/bit_decomposition.rs
@@ -1,5 +1,6 @@
 use super::bitwise_less_than_prime::BitwiseLessThanPrime;
 use super::dumb_bitwise_sum::BitwiseSum;
+use super::random_bits_generator::RandomBitsGenerator;
 use crate::error::Error;
 use crate::ff::{Field, Int};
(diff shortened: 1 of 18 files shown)


src/protocol/boolean/bit_decomposition.rs
```diff
@@ -21,17 +22,15 @@ pub struct BitDecomposition {}
 impl BitDecomposition {
     #[allow(dead_code)]
     pub async fn execute<F: Field>(
+        rbg: RandomBitsGenerator<F>,
         ctx: SemiHonestContext<'_, F>,
         record_id: RecordId,
         a_p: &Replicated<F>,
```
taikiy, line 28: nit: Other protocols take `ctx` and `record_id` as first two arguments, followed by inputs. Maybe change the order to:
```
record_id: RecordId,
rbg: RandomBitsGenerator<F>,
a_p: &Replicated<F>,
```

taikiy (approved):
Just one nit comment. LGTM

martinthomson (approved):

akoshelev (approved):

PR #299, opened 2022-12-01
Add upgraded values to the MAC

martinthomson:


benjaminsavage (approved):
This was a really key catch. I'm sad I missed this reviewing the last one =(

PR #302, opened 2022-12-01
Accumulating MACs after reshare + test

richajaindce:
1. Accumulating MACs after reshare
2. Reshare malicious test which fails validation check on additive attack


src/test_fixture/world.rs
```diff
@@ -230,7 +230,8 @@ where
 
         // Perform validation and convert the results we just got: P to O
         let output = join_all(
-            zip(v, m_results).map(|(v, m_result)| async { v.validate(m_result).await.unwrap() }),
+            zip(v, m_results)
+                .map(|(v, m_result)| async move { v.validate(m_result).await.unwrap() }),
```
benjaminsavage, line 234: Why did you need to modify this line? Why add this "move"?

src/protocol/malicious.rs
```diff
@@ -197,7 +197,7 @@ impl<'a, F: Field> MaliciousValidator<'a, F> {
     ///
     /// ## Panics
     /// Will panic if the mutex is poisoned
-    pub async fn validate<D: DowngradeMalicious>(self, values: D) -> Result<D::Target, Error> {
+    pub async fn validate<D: DowngradeMalicious>(&self, values: D) -> Result<D::Target, Error> {
```
benjaminsavage, line 200: @martinthomson says we should not do this.

src/protocol/sort/reshare.rs
```diff
@@ -104,7 +108,7 @@ impl<F: Field> Reshare<F> for MaliciousContext<'_, F> {
         use crate::secret_sharing::ThisCodeIsAuthorizedToDowngradeFromMalicious;
         let rx_ctx = self.narrow(&ReshareMAC);
         let (x, rx) = try_join(
-            self.semi_honest_context().reshare(
+            self.narrow(&ReshareInput).semi_honest_context().reshare(
```
benjaminsavage, line 111: Why is this needed? I don't think this is necessary. The caller should have narrowed (if necessary) before invoking "reshare"

src/protocol/sort/reshare.rs
```diff
@@ -213,5 +225,69 @@ mod tests {
...
+            input: &MaliciousReplicated<F>,
+            record_id: RecordId,
+            to_helper: Role,
+            additive_error: F,
+        ) -> Result<MaliciousReplicated<F>, Error> {
+            use crate::protocol::context::SpecialAccessToMaliciousContext;
+            use crate::secret_sharing::ThisCodeIsAuthorizedToDowngradeFromMalicious;
+            let rx_ctx = ctx.narrow("reshare_mac");
+            let (x, rx) = try_join(
+                ctx.narrow("reshare_input").semi_honest_context().reshare(
```
benjaminsavage, line 240: I don't think you need to narrow here.

src/protocol/sort/reshare.rs
```diff
@@ -213,5 +225,69 @@ mod tests {
...
+        async fn reshare_with_additive_attack<F: Field>(
+            ctx: MaliciousContext<'_, F>,
+            input: &MaliciousReplicated<F>,
+            record_id: RecordId,
+            to_helper: Role,
+            additive_error: F,
+        ) -> Result<MaliciousReplicated<F>, Error> {
+            use crate::protocol::context::SpecialAccessToMaliciousContext;
+            use crate::secret_sharing::ThisCodeIsAuthorizedToDowngradeFromMalicious;
+            let rx_ctx = ctx.narrow("reshare_mac");
```
benjaminsavage, line 238: I think it would be better to use `&Step::ReshareMAC`

src/protocol/sort/reshare.rs
```diff
@@ -213,5 +225,69 @@ mod tests {
...
+                    record_id,
+                    to_helper,
+                ),
+                rx_ctx
+                    .semi_honest_context()
+                    .reshare(input.rx(), record_id, to_helper),
+            )
+            .await?;
+            let malicious_input =
+                MaliciousReplicated::new(x + &Replicated::new(additive_error, additive_error), rx);
```
benjaminsavage, line 251: So this is not actually the place where the additive attack happens - that's on the communication with the other helpers - where it sends a value.

benjaminsavage (changes requested):

src/protocol/sort/reshare.rs
richajaindce: We need to reshare both x and rx at this step. We also need context to calculate accumulate_macs. self.semi_honest will consume self. Another option is to clone self before resharing x.

src/protocol/sort/reshare.rs
richajaindce: U are right. Let me change this

src/test_fixture/world.rs
richajaindce: If got left from changes. Apologies

src/protocol/sort/reshare.rs
```diff
@@ -213,5 +224,120 @@ mod tests {
...
+                    .await?;
+
+                // Sleep until `to_helper.left` sends us their part1 value
+                let part1: F = channel
+                    .receive(to_helper.peer(Direction::Left), record_id)
+                    .await?;
+
+                Ok(Replicated::new(r0, part1 + part2 + additive_error))
+            } else {
+                Ok(Replicated::new(r0 + additive_error, r1 - additive_error))
```
benjaminsavage: This helper cannot wage an additive attack

benjaminsavage (approved):
I love it!

PR #303, opened 2022-12-01
Eliminate `Batch` struct

taikiy:
I originally created `Batch` to wrap `Vec` and put all "is the length of this vector less than u32::MAX?" logic, but it's making the code harder to read and sometimes create unnecessary clones. I think we are okay to assume that the elements in the vector will be less than 4B for now.


src/protocol/attribution/accumulate_credit.rs
```diff
@@ -125,7 +112,7 @@ pub async fn accumulate_credit<F: Field>(
     // of other elements, allowing the algorithm to be executed in parallel.
 
     // generate powers of 2 that fit into input len. If num_rows is 15, this will produce [1, 2, 4, 8]
-    for (depth, step_size) in std::iter::successors(Some(1u32), |prev| prev.checked_mul(2))
+    for (depth, step_size) in std::iter::successors(Some(1_usize), |prev| prev.checked_mul(2))
```
taikiy, line 115: Since we know that the size of `Batch` will be within u32, it implemented `Index<u32>` and we needed to use `u32` to access its elements instead of `usize`. This change is reverting that to use usize to access elements of a vector.

martinthomson (approved):

PR #314, opened 2022-12-03
Improve error message for duplicate records sent to Infra

akoshelev:
Before this change, infrastructure did not allow duplicates to be inserted (duplicate is a message with the same record id sent to the same channel more than once). That is expected but the error did not get propagated back to the caller (protocol in this case). `send` method did not wait for the status of the operation - regardless whether gateway can actually accept this message or no, it allowed protocols to move forward immediately after send.

New behavior provides the essential mechanism to built the backpressure later, but currently it is used to propagate errors from the send buffers.

Protocols that attempt to insert the same message twice to the infrastructure will get an error at the point where this insertion occurs and will get the proper message

```
An error occurred while sending data to channel[peer=H2,step=step=protocol/run-0]: Record RecordId(1) has been received twice
```

### DIFF
--- a/src/helpers/messaging.rs
+++ b/src/helpers/messaging.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 
 use crate::ff::{Field, Int};
-use crate::helpers::buffers::{SendBuffer, SendBufferConfig};
+use crate::helpers::buffers::{PushError, SendBuffer, SendBufferConfig};
 use crate::helpers::{MessagePayload, MESSAGE_PAYLOAD_SIZE_BYTES};
 use crate::task::JoinHandle;
 use ::tokio::sync::{mpsc, oneshot};
@@ -80,9 +80,15 @@ impl<F: Field> Message for F {
 pub struct Gateway {
     /// Sender end of the channel to send requests to receive messages from peers.
     tx: mpsc::Sender<ReceiveRequest>,
-    envelope_tx: mpsc::Sender<(ChannelId, MessageEnvelope)>,
+    envelope_tx: mpsc::Sender<SendRequest>,
     control_handle: JoinHandle<()>,
 }
+
+pub(super) type SendRequest = (
+    ChannelId,
+    MessageEnvelope,
+    oneshot::Sender<Option<PushError>>,
+);
 
 /// Channel end
 #[derive(Debug)]
@@ -153,7 +159,7 @@ pub struct GatewayConfig {
 impl Gateway {
     pub fn new<N: Network>(role: Role, network: &N, config: GatewayConfig) -> Self {
         let (tx, mut receive_rx) = mpsc::channel::<ReceiveRequest>(1);
-        let (envelope_tx, mut envelope_rx) = mpsc::channel::<(ChannelId, MessageEnvelope)>(1);
+        let (envelope_tx, mut envelope_rx) = mpsc::channel::<SendRequest>(1);
         let mut message_stream = network.recv_stream();
         let mut network_sink = network.sink();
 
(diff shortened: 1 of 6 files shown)


src/helpers/messaging.rs
```diff
@@ -80,10 +80,16 @@ impl<F: Field> Message for F {
...
     tx: mpsc::Sender<ReceiveRequest>,
-    envelope_tx: mpsc::Sender<(ChannelId, MessageEnvelope)>,
+    envelope_tx: mpsc::Sender<SendRequest>,
     control_handle: JoinHandle<()>,
 }
 
+pub(super) type SendRequest = (
+    ChannelId,
+    MessageEnvelope,
+    oneshot::Sender<Option<PushError>>,
```
martinthomson, line 90: So this is the real cost of this change.  I'm not sure about this part.

Given that this exists primarily to deal with coding issues, can we limit this to debug builds?

For non-debug builds, a panic on the task that receives messages would be fine.  This is not an error we should be tolerating.

src/helpers/messaging.rs
```diff
@@ -153,7 +159,7 @@ pub struct GatewayConfig {
 impl Gateway {
     pub fn new<N: Network>(role: Role, network: &N, config: GatewayConfig) -> Self {
         let (tx, mut receive_rx) = mpsc::channel::<ReceiveRequest>(1);
-        let (envelope_tx, mut envelope_rx) = mpsc::channel::<(ChannelId, MessageEnvelope)>(1);
+        let (envelope_tx, mut envelope_rx) = mpsc::channel::<SendRequest>(1);
```
martinthomson, line 162: Should we be configuring the size of the buffers on channels?  For tests, 1 is fine, but it seems like maybe we should let them hold a few more otherwise.

Not for this change, but it seems like we might benefit from it some time.

src/helpers/messaging.rs
```diff
@@ -175,13 +181,22 @@ impl Gateway {
...
+                                    .expect("Failed to send data to the network");
+                                None
+                            },
+                            Ok(None) => None,
+                            Err(err) => Some(err),
+                        };
+
+                        status_tx.send(status).unwrap_or_else(|_| {
+                            tracing::warn!("No listener for send request {channel_id:?}/{:?}", msg.record_id);
+                        });
```
martinthomson, line 199: This is now big enough to get its own function.

martinthomson (approved):

src/helpers/messaging.rs
akoshelev: let me create a task for it

src/helpers/messaging.rs
akoshelev: true, moved it to a separate function. One day I would create a struct for this control loop

martinthomson (approved):
This looks like a good construction.

src/helpers/messaging.rs
```diff
@@ -158,8 +154,8 @@ pub struct GatewayConfig {
 
 impl Gateway {
     pub fn new<N: Network>(role: Role, network: &N, config: GatewayConfig) -> Self {
-        let (tx, mut receive_rx) = mpsc::channel::<ReceiveRequest>(1);
-        let (envelope_tx, mut envelope_rx) = mpsc::channel::<SendRequest>(1);
+        let (recv_tx, mut recv_rx) = mpsc::channel::<ReceiveRequest>(1);
+        let (send_tx, mut send_rx) = mpsc::channel::<SendRequest>(1);
```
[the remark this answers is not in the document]
akoshelev: We're in the crypto world at the end of the day!

PR #315, opened 2022-12-05
Impl Researable for Vec<Replicated<F>>

benjaminsavage:
I am confused

### DIFF
--- a/src/protocol/sort/apply_sort/shuffle.rs
+++ b/src/protocol/sort/apply_sort/shuffle.rs
@@ -1,6 +1,6 @@
 use std::iter::{repeat, zip};
 
-use crate::secret_sharing::SecretSharing;
+use crate::secret_sharing::{SecretSharing, Replicated};
 use crate::{
     error::Error,
     ff::Field,
@@ -24,6 +24,25 @@ pub trait Resharable<F: Field>: Sized {
     async fn reshare<C>(&self, ctx: C, record_id: RecordId, to_helper: Role) -> Result<Self, Error>
     where
         C: Context<F, Share = <Self as Resharable<F>>::Share> + Send;
+}
+
+#[async_trait]
+impl<F: Field> Resharable<F> for Vec<Replicated<F>> {
+    type Share = Replicated<F>;
+
+    async fn reshare<C>(&self, ctx: C, record_id: RecordId, to_helper: Role) -> Result<Self, Error>
+    where
+        C: Context<F, Share = <Self as Resharable<F>>::Share> + Send,
+    {
+        try_join_all(
+            self.iter().map(|x| {
+                let c = ctx.narrow("foo");
+                async move {
+                    c.reshare(x, record_id, to_helper).await
+                }
+            })
+        ).await
+    }
 }
 
 async fn reshare<F, C, S, T>(input: &[T], ctx: C, to_helper: Role) -> Result<Vec<T>, Error>
--- a/src/protocol/sort/apply_sort/shuffle.rs
+++ b/src/protocol/sort/apply_sort/shuffle.rs
@@ -26,6 +26,32 @@ pub trait Resharable<F: Field>: Sized {
         C: Context<F, Share = <Self as Resharable<F>>::Share> + Send;
 }
 
+pub struct InnerVectorElementStep(usize);
+
+impl crate::protocol::Substep for InnerVectorElementStep {}
+
+impl AsRef<str> for InnerVectorElementStep {
+    fn as_ref(&self) -> &str {
+        const VEC_ELEM: [&str; 64] = [
+            "elem0", "elem1", "elem2", "elem3", "elem4", "elem5", "elem6", "elem7", "elem8",
+            "elem9", "elem10", "elem11", "elem12", "elem13", "elem14", "elem15", "elem16",
+            "elem17", "elem18", "elem19", "elem20", "elem21", "elem22", "elem23", "elem24",
+            "elem25", "elem26", "elem27", "elem28", "elem29", "elem30", "elem31", "elem32",
+            "elem33", "elem34", "elem35", "elem36", "elem37", "elem38", "elem39", "elem40",
+            "elem41", "elem42", "elem43", "elem44", "elem45", "elem46", "elem47", "elem48",
+            "elem49", "elem50", "elem51", "elem52", "elem53", "elem54", "elem55", "elem56",
+            "elem57", "elem58", "elem59", "elem60", "elem61", "elem62", "elem63",
+        ];
+        VEC_ELEM[self.0]
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -26,6 +26,48 @@ pub trait Resharable<F: Field>: Sized {
...
+#[async_trait]
+impl<F: Field> Resharable<F> for Vec<Replicated<F>> {
+    type Share = Replicated<F>;
+
+    async fn reshare<C>(&self, ctx: C, record_id: RecordId, to_helper: Role) -> Result<Self, Error>
+    where
+        C: Context<F, Share = <Self as Resharable<F>>::Share> + Send,
+    {
+        try_join_all(self.iter().enumerate().map(|(i, x)| {
+            let c = ctx.narrow(&InnerVectorElementStep::from(i));
```
martinthomson, line 64: So we can only reshare a `Vec<>` with a length of 64 or less?  That is probably nto going to scale.

This seems like a good case for ignoring the record ID.  Or at least asserting that it is zero before ignoring it.  Alternatively, use the value provided as a starting value.

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -175,5 +213,42 @@ mod tests {
...
+                    shuffle_shares(
+                        vec_of_vec_of_shares,
+                        (perms.0.as_slice(), perms.1.as_slice()),
+                        ctx,
+                    )
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct();
```
martinthomson, line 240: By reconstructing, you aren't really confirming that resharing happened.  Maybe each helper can search for `vec_of_vec_of_shares[i]` (or just `vec_of_vec_of_shares[0]`) in the output of `shuffle_shares` and confirm that it can't find that value.

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -175,5 +213,42 @@ mod tests {
...
+
+            let all_match_keys: HashSet<u32> = HashSet::from_iter(original_match_keys);
+
+            let mut match_keys_found = Vec::with_capacity(5);
```
martinthomson, line 244: 5?  How about `original_match_keys.len()`

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -175,5 +213,42 @@ mod tests {
             // Shuffled output should have same inputs
             assert_eq!(hashed_output_secret, hashed_input);
         }
+
+        #[tokio::test]
+        async fn shuffle_vec_of_replicated() {
+            let original_match_keys = [
```
martinthomson, line 219: Use `const` for this sort of thing.

As for the name, isn't this just a bunch of random values in the field?  I don't see any match keys.

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -175,5 +213,42 @@ mod tests {
...
+                .reconstruct();
+
+            let all_match_keys: HashSet<u32> = HashSet::from_iter(original_match_keys);
+
+            let mut match_keys_found = Vec::with_capacity(5);
+            for row in result {
+                let mk = u32::try_from(bits_to_value(&row)).unwrap();
+                assert!(all_match_keys.contains(&mk));
+                match_keys_found.push(mk);
+            }
```
martinthomson, line 249: You should just be able to sort the output and check for equality with the input (as you do below).  Checking that each is found isn't necessary.

martinthomson (changes requested):

src/protocol/sort/apply_sort/shuffle.rs
benjaminsavage: The use-case for this is to sort the match-keys. Each match key is a [Replicated<F>; 40]. And we need to use RecordId to represent the row. So each row will have 40 bits within in. So I don't think either of the approaches you suggested will work.

martinthomson: Oh, right.  That's a bit annoying, but I guess this has to be the way.

src/protocol/sort/apply_sort/shuffle.rs
benjaminsavage: yeah, it's just a bunch of random values.

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -26,6 +26,48 @@ pub trait Resharable<F: Field>: Sized {
...
+        const VEC_ELEM: [&str; 64] = [
+            "elem0", "elem1", "elem2", "elem3", "elem4", "elem5", "elem6", "elem7", "elem8",
+            "elem9", "elem10", "elem11", "elem12", "elem13", "elem14", "elem15", "elem16",
+            "elem17", "elem18", "elem19", "elem20", "elem21", "elem22", "elem23", "elem24",
+            "elem25", "elem26", "elem27", "elem28", "elem29", "elem30", "elem31", "elem32",
+            "elem33", "elem34", "elem35", "elem36", "elem37", "elem38", "elem39", "elem40",
+            "elem41", "elem42", "elem43", "elem44", "elem45", "elem46", "elem47", "elem48",
+            "elem49", "elem50", "elem51", "elem52", "elem53", "elem54", "elem55", "elem56",
+            "elem57", "elem58", "elem59", "elem60", "elem61", "elem62", "elem63",
+        ];
```
martinthomson, line 44: I have a macro for you here.  Upcoming.

src/protocol/sort/apply_sort/shuffle.rs
martinthomson: Can you add a comment to the effect that this is for a vector of bit-decomposed values?

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -175,5 +214,64 @@ mod tests {
...
+                        let perms = get_two_of_three_random_permutations(5, ctx.prss_rng());
+                        let shuffled_shares = shuffle_shares(
+                            vec_of_vec_of_shares,
+                            (perms.0.as_slice(), perms.1.as_slice()),
+                            ctx,
+                        )
+                        .await
+                        .unwrap();
+
+                        for row in &shuffled_shares {
```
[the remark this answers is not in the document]
benjaminsavage: Not quite... `share_appears_anywhere` expects just a share, not a Vec<> of shares.  I can do:
```
assert!(!shuffled_shares.iter().any(|row| row.iter().any(|x| share_appears_anywhere(x, &copy_of_input))));
```

martinthomson (approved):

PR #318, opened 2022-12-05
Making apply_sort generic

richajaindce:
apply_sort is now generic over malicious and semihonest context. Have declared a trait on context and implemented for both contexts. We would expect sort_permutation to be received as Replicated<F> and it is the responsibility of caller to upgrade to malicious before applying the sort permutation


richajaindce:
Ah.. this one reveals the shuffled permutation. Will need to validate that in some way :/

richajaindce:
Closing this since we will need a different approach to handle this. Good news is that apply will now become agnostic to context type!

PR #320, opened 2022-12-06
Making apply sort generic

richajaindce:


richajaindce:
apply now accepts a shuffle-revealed permutation which can be applied to the inputs. This will help us isolate the validation needed for reveal inside sort.
We will have 2 generate_permutation - one for each type of context. Malicious will do extra work needed for validation

src/protocol/sort/apply_sort/mod.rs
```diff
@@ -3,66 +3,51 @@ pub mod shuffle;
...
-    {
-        let revealed_and_random_permutation = shuffle_and_reveal_permutation(
-            ctx.narrow(&ShuffleRevealPermutation),
-            input.len().try_into().unwrap(),
-            self.0,
-        )
-        .await?;
+/// # Errors
+/// Propagates errors from shuffle/reshare
+pub async fn apply<C, F, S, I>(
```
benjaminsavage, line 18: `apply` is a fairly generic name, and it is the same as the one in `apply.rs`. Can we give this a unique name, like `apply_permutation` or something?

benjaminsavage (approved):
Looks good. Let's give that protocol a more descriptive name and land it.

src/protocol/sort/apply_sort/mod.rs
richajaindce: Just making it apply_sort_permutation for now. Earlier we were calling it through the object SortPermutation but now that wrapper is gone, this should make more sense (I think)

benjaminsavage (approved):

PR #322, opened 2022-12-06
Basics

richajaindce:
Moving basic building blocks of protocol in its own folder
This includes
1. Reshare
2. Reveal
3. Multiplication
4. Check Zero


src/protocol/basics/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod check_zero;
+pub mod mul;
+pub mod reshare;
+pub mod reveal;
```
martinthomson, line 4: The functions that each of these expose can probably be re-exported here.  That way the callers don't have to "use some::long::path;".

benjaminsavage (approved):
LGTM

PR #324, opened 2022-12-06
Malicious SolvedBits

taikiy:
As it's implemented now, `generate_random_bits()` calls `convert_bit_local()` which creates `Replicated<F>` instances. I'm moving `generate_random_bits()` as a trait and implement it for SemiHonestContext and MaliciousContext, so that protocols calling this method can be upgraded to malicious as well.

* Created `RandomBits` trait which has only one method 
* Malicious version of `generate_random_bits()` calls `upgrade_bit_triple()`
* Removed SolvedBits struct to align with other protocols. Now the protocol is just `solved_bits()` method
* Other mechanical changes for generics <F, S, C>

### DIFF
--- a/src/protocol/boolean/generate_random_bits/malicious.rs
+++ b/src/protocol/boolean/generate_random_bits/malicious.rs
@@ -1,0 +1,57 @@
+use super::{RandomBits, Step};
+use crate::error::Error;
+use crate::ff::{Field, Int};
+use crate::protocol::context::MaliciousContext;
+use crate::protocol::modulus_conversion::{convert_bit, convert_bit_local};
+use crate::protocol::{context::Context, BitOpStep, RecordId};
+use crate::secret_sharing::{MaliciousReplicated, XorReplicated};
+use async_trait::async_trait;
+use futures::future::try_join_all;
+use std::iter::repeat;
+
+#[async_trait]
+impl<F: Field> RandomBits<F> for MaliciousContext<'_, F> {
+    type Share = MaliciousReplicated<F>;
+
+    /// Generates a sequence of `l` random bit sharings in the target field `F`.
+    async fn generate_random_bits(self, record_id: RecordId) -> Result<Vec<Self::Share>, Error> {
+        // Calculate the number of bits we need to form a random number that
+        // has the same number of bits as the prime.
+        let l = u128::BITS - F::PRIME.into().leading_zeros();
+        let leading_zero_bits = F::Integer::BITS - l;
+
+        // Generate a pair of random numbers. We'll use these numbers as
+        // the source of `l`-bit long uniformly random sequence of bits.
+        let (b_bits_left, b_bits_right) = self
+            .narrow(&Step::RandomValues)
+            .prss()
+            .generate_values(record_id);
+
+        // Same here. For now, 256-bit is enough for our F_p
+        let xor_share = XorReplicated::new(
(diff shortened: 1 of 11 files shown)


src/protocol/boolean/generate_random_bits/semi_honest.rs
```diff
@@ -0,0 +1,54 @@
+use super::{RandomBits, Step};
```
taikiy, line 1: This is exactly the same as the old `generate_random_bits()`. No need to review.

src/protocol/boolean/generate_random_bits/malicious.rs
```diff
@@ -0,0 +1,57 @@
...
+            u64::try_from(b_bits_right & u128::from(u64::MAX)).unwrap(),
+        );
+
+        // Convert each bit to secret sharings of that bit in the target field
+        let c = self.narrow(&Step::ConvertShares);
+        let futures = (0..l).map(|i| {
+            let c = c.narrow(&BitOpStep::from(i));
+            let triple = convert_bit_local::<F>(c.role(), i, &xor_share);
+            async move {
+                let malicious_triple = c.upgrade_bit_triple(record_id, i, triple).await?;
```
taikiy, line 42: This is the only line different from the semi-honest version.

src/protocol/context/malicious.rs
```diff
@@ -77,9 +77,12 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
     pub async fn upgrade_bit_triple(
         &self,
         record_id: RecordId,
+        bit_index: u32,
```
taikiy, line 80: Added `bit_index`. Calling `upgrade_bit_triple()` multiple times per bit panics, because it uses the same `upgrade_ctx`. This needs to be narrowed per bit like `upgrade_bit()`.

src/test_fixture/sharing.rs
```diff
@@ -328,3 +328,13 @@ impl<F: Field> ValidateMalicious<F> for [Vec<MaliciousReplicated<F>>; 3] {
         }
     }
 }
+
+impl<F: Field> ValidateMalicious<F> for [(MaliciousReplicated<F>, Vec<MaliciousReplicated<F>>); 3] {
```
taikiy, line 332: This is for returning a tuple of `(MaliciousReplicated<F>, Vec<MaliciousReplicated<F>>)` from a closure passed to `world.malicious()`. We need this to support this tuple format for returning values inside `RandomBitsShare`.

src/protocol/boolean/generate_random_bits/malicious.rs
benjaminsavage: I do not love this level of code duplication. How about you split the semi-honest version into two separate function calls:
1. beginning of this function up until it calls `convert_bit_local`
2. a function that accepts "triples", calls "convert_bit" on them, then appends the leading zeros.

Then the malicious version can just call function #1, call `upgrade_bit_triple`, then call function #2. 
The semi-honest version will just call function #1 followed by function #2.

benjaminsavage (approved):
This looks great. I tried and failed to make this code generic - so I know how tricky this is! Well done!

src/protocol/boolean/generate_random_bits/malicious.rs
taikiy: Fixed as suggested.

PR #325, opened 2022-12-06
Ipa protocol

benjaminsavage:
Finally putting everything together to actually run the end-to-end IPA protocol (Semi-Honest)


src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    let num_bits: usize = usize::try_from(num_bits).unwrap();
+
+    let combined_match_keys_and_sidecar_data = other_inputs
+        .into_iter()
+        .enumerate()
+        .map(|(row_index, mut sidecar_data)| {
+            for i in 0..num_bits {
+                sidecar_data.push(converted_shares[i][row_index].clone());
+            }
```
benjaminsavage, line 86: Unfortunately, `converted_shares` is _transposed_. So in this case it's a Vec of length 20, each containing a Vec of length 5. 

If anyone knows a better way of doing this transpose, please let me know.

src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    let accumulated_credits =
+        accumulate_credit(ctx.narrow(&Step::AccumulateCredit), &attribution_input_rows).await?;
+
+    let _user_capped_credits = credit_capping(
+        ctx.narrow(&Step::PerformUserCapping),
+        &accumulated_credits,
+        per_user_credit_cap,
+    )
+    .await?;
+
```
benjaminsavage, line 138: We still need to call the next protocols:
1. Aggregate down to sums per breakdown key
2. Add random noise for DP

src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    let attribution_input_rows = sorted_rows
+        .iter()
+        .enumerate()
+        .map(|(i, row)| {
+            let hb = if i == 0 {
+                Replicated::ZERO
+            } else {
+                helper_bits[i - 1].clone()
+            };
```
martinthomson, line 122: You probably want to do `Some(Replicated::ZERO).into_iter().chain(helper_bits)` then zip that up with the rows.

src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    ctx: SemiHonestContext<'_, F>,
+    mk_shares: &[XorReplicated],
+    num_bits: u32,
+    other_inputs: Vec<Vec<Replicated<F>>>,
+    per_user_credit_cap: u32,
+) -> Result<Vec<AggregateCreditOutputRow<F>>, Error>
+where
+    F: Field,
+{
+    debug_assert_eq!(mk_shares.len(), other_inputs.len());
```
martinthomson, line 66: Next step: make a struct for this.

src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    let local_lists = convert_all_bits_local(ctx.role(), mk_shares, num_bits);
```
martinthomson, line 68: TODO: decrypt.

martinthomson (commented):
Progress!

src/protocol/ipa/mod.rs
benjaminsavage: I'm not sure what you mean.

martinthomson: If we want to do the whole protocol, we'll need to decrypt shares.

src/protocol/ipa/mod.rs
benjaminsavage: done!

src/protocol/ipa/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    .unwrap();
+    let sort_permutation = generate_permutation_and_reveal_shuffled(
+        ctx.narrow(&Step::GenSortPermutationFromMatchKeys),
+        &converted_shares,
+        num_bits,
+    )
+    .await
+    .unwrap();
+
+    let converted_shares = transpose(&converted_shares);
```
martinthomson, line 163: Can we flag this as a potential problem?  The use of transposition means a lot of allocations and a break from streaming that might be a performance problem later.

I think that ultimately what we'll want to do is provide the sort functions with entire input rows, with a means of it keying into that input row with a bit index (or radix N index, I guess).

martinthomson (approved):
Let's go!

src/protocol/ipa/mod.rs
benjaminsavage: You're absolutely right. This is totally sub-optimal. I've filed #330 to track this issue.

PR #326, opened 2022-12-06
AggregateCredit Protocol #1 - Sort by aggregation_bit, breakdown_key

taikiy:
`AggregateCredit` protocol takes the result from `CreditCapping` protocol, does some pre-processing (coming in the next diff), and makes two sort calls to perform the aggregation. This diff adds the sorting part of the AggregateCredit protocol.

Two sorts called during the aggregation step are:
1. Sort by aggregation_bit and then breakdown_key
2. Sort by aggregation_bit (2nd time)

`aggregation_bit` is a bit sharing of 0 or 1, and used to determine which row needs to be aggregated. This bit will be added at the beginning of the AggregateCredit protocol.

### DIFF
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -17,7 +17,10 @@ use futures::future::try_join_all;
 use std::iter::repeat;
 
 #[async_trait]
-impl<F: Field> Resharable<F> for CappedCreditsWithAggregationBit<F> {
+impl<F: Field> Resharable<F> for CappedCreditsWithAggregationBit<F>
+where
+    F: Sized,
+{
     type Share = Replicated<F>;
 
     async fn reshare<C>(&self, ctx: C, record_id: RecordId, to_helper: Role) -> Result<Self, Error>
@@ -175,7 +178,7 @@ impl AsRef<str> for Step {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(feature = "shuttle")))]
 pub(crate) mod tests {
     use super::super::tests::{BD, H};
     use super::sort_by_aggregation_bit_and_breakdown_key;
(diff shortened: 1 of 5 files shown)


src/protocol/sort/apply_sort/mod.rs
```diff
@@ -91,7 +92,7 @@ mod tests {
         });
         let expected = permutation.apply_slice(&sidecar);
 
-        let result = world
+        let result: [Vec<AttributionInputRow<Fp32BitPrime>>; 3] = world
```
richajaindce, line 95: Do we need to explicitly give it a type?

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -186,7 +187,7 @@ mod tests {
             });
             let hashed_input: HashSet<[u8; 4]> = input.iter().map(Into::into).collect();
 
-            let result = world
+            let result: [Vec<AttributionInputRow<Fp31>>; 3] = world
```
richajaindce, line 190: Same as above, is this needed?

src/protocol/sort/apply_sort/mod.rs
taikiy: Yes.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -0,0 +1,358 @@
...
+            ctx.narrow(&Step::BitDecomposeBreakdownKey),
+            &sorted_by_aggregation_bit,
+        )
+        .await?,
+    );
+
+    generate_permutation(
+        ctx.narrow(&Step::GeneratePermutationByBreakdownKey),
+        &breakdown_keys,
+        F::Integer::BITS,
```
benjaminsavage, line 124: We can probably optimize this right? This is a function of the maximum value of "breakdown key", which should be a known constant. I'd suggest making this an argument to the function.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/// Sort the input by `aggregation_bit` first, then by `breakdown_key`
+#[allow(dead_code)]
+async fn sort_by_aggregation_bit_and_breakdown_key<F: Field>(
+    ctx: SemiHonestContext<'_, F>,
+    input: &[CappedCreditsWithAggregationBit<F>],
+) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
+    // Sort by aggregation_bit
+    let sorted_by_aggregation_bit = sort_by_aggregation_bit(ctx.clone(), input).await?;
```
benjaminsavage, line 109: For the future, I wonder if it is better to just pre-pend the bit-decomposed breakdown keys with this bit and generate a single permutation, which is applied just once.

PR #327, opened 2022-12-07
Aggregation Protocol [part 2 / 2] - Add the prefix sum aggregation

taikiy:
This diff adds the prefix-sum computation of the credit values per breakdown key. The computation uses the same binary-tree prefix sum approach as other attribution protocols do.

### DIFF
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -67,15 +67,15 @@ pub async fn aggregate_credit<F: Field>(
 ) -> Result<Vec<AggregateCreditOutputRow<F>>, Error> {
     let one = ctx.share_of_one();
 
-    // Add aggregation bits and initialize with 1's. We create a new vector here
-    // because we need to push new rows needed for the aggregate computation.
+    //
+    // 1. Add aggregation bits and new rows per unique breakdown_key
+    //
     let capped_credits_with_aggregation_bits =
         add_aggregation_bits_and_breakdown_keys(&ctx, capped_credits);
 
     //
     // 2. Sort by `aggregation_bit` first, then by `breakdown_key`.
     //
-
     let sorted_input = sort_by_aggregation_bit_and_breakdown_key(
         ctx.narrow(&Step::SortByBreakdownKeyAndAttributionBit),
         &capped_credits_with_aggregation_bits,
@@ -163,7 +163,6 @@ pub async fn aggregate_credit<F: Field>(
     //
     // 4. Sort by `aggregation_bit`
     //
-
     let sorted_output =
         sort_by_aggregation_bit(ctx.narrow(&Step::SortByAttributionBit), &aggregated_credits)
             .await?;
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -1,10 +1,14 @@
-use super::CappedCreditsWithAggregationBit;
+use super::{
+    compute_b_bit, compute_stop_bit, AggregateCreditOutputRow, CappedCreditsWithAggregationBit,
+    CreditCappingOutputRow, InteractionPatternStep,
+};
 use crate::error::Error;
 use crate::ff::{Field, Int};
 use crate::helpers::Role;
 use crate::protocol::attribution::AttributionResharableStep::{
     AggregationBit, BreakdownKey, Credit, HelperBit,
 };
+use crate::protocol::basics::SecureMul;
 use crate::protocol::boolean::{random_bits_generator::RandomBitsGenerator, BitDecomposition};
 use crate::protocol::context::{Context, SemiHonestContext};
 use crate::protocol::sort::apply_sort::apply_sort_permutation;
@@ -13,7 +17,7 @@ use crate::protocol::sort::generate_permutation::generate_permutation_and_reveal
 use crate::protocol::{RecordId, Substep};
 use crate::secret_sharing::Replicated;
 use async_trait::async_trait;
-use futures::future::try_join_all;
+use futures::future::{try_join, try_join_all};
 use std::iter::repeat;
 
 #[async_trait]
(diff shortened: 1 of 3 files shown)


src/protocol/attribution/aggregate_credit.rs
```diff
@@ -100,33 +261,35 @@ async fn bit_decompose_breakdown_key<F: Field>(
...
-    // Next, sort by breakdown_key
     // TODO: Change breakdown_keys to use XorReplicated to avoid bit-decomposition calls
-    let breakdown_keys = transpose(
-        &bit_decompose_breakdown_key(
-            ctx.narrow(&Step::BitDecomposeBreakdownKey),
-            &sorted_by_aggregation_bit,
-        )
-        .await?,
+    let mut breakdown_keys = transpose(
+        &bit_decompose_breakdown_key(ctx.narrow(&Step::BitDecomposeBreakdownKey), input).await?,
```
taikiy, line 278: [Comment from #326 
](https://github.com/private-attribution/ipa/pull/326#discussion_r1041632591)

> We can probably optimize this right? This is a function of the maximum value of "breakdown key", which should be a known constant. I'd suggest making this an argument to the function.

I didn't understand what you meant by "the maximum value of "breakdown key"? Bit-decomposition is applied to all breakdown_key shares in the input rows.

Since this function is called only once per aggregation protocol call, I left it as it is for now. I'll optimize in a later diff if needed.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -52,6 +56,166 @@ where
...
+    //
+
+    let sorted_output =
+        sort_by_aggregation_bit(ctx.narrow(&Step::SortByAttributionBit), &aggregated_credits)
+            .await?;
+
+    // Take the first k elements, where k is the amount of breakdown keys.
+    let result = sorted_output
+        .iter()
+        .take(MAX_BREAKDOWN_KEY.try_into().unwrap())
```
benjaminsavage, line 174: This feels bad. Here you're assuming that the breakdown keys are sequential integers starting from zero and ending on something called `MAX_BREAKDOWN_KEY`. That name doesn't imply all of that. If I have 3 breakdown keys: [1, 10, 100] then my "Max breakdown key" is 100.

This should instead by called "NUM_BREAKDOWN_KEYS" I think.

taikiy: As a part of the protocol, we need to add dummy rows per unique breakdown_key that exists in the reports. Do we know the breakdown_key secret values ahead of time? If so, then we can just add those known values. If not, then we'll need to blindly add all possible breakdown_keys. And this part does that. `MAX_BREAKDOWN_KEY` means the maximum breakdown key value from 0.

I think it'll be okay if we limit the breakdown_key range to somewhat small like 2^16. If it goes up to 2^32, then we'll need a different approach.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -52,6 +56,166 @@ where
...
+    // need to append all possible values. For now, we assume breakdown_key
+    // is in the range of (0..MAX_BREAKDOWN_KEY).
+    (0..MAX_BREAKDOWN_KEY).for_each(|i| {
+        capped_credits_with_aggregation_bits.push(CappedCreditsWithAggregationBit {
+            helper_bit: zero.clone(),
+            aggregation_bit: zero.clone(),
+            breakdown_key: Replicated::from_scalar(ctx.role(), F::from(i)),
+            credit: zero.clone(),
+        });
+    });
```
benjaminsavage, line 214: Wait a second...

...we are just about to *sort* by these aggregation bits... but we actually know their values already. That's a big waste of energy.

Instead of pushing them to the end of this vector, why not just insert them at the beginning. Then you *know* the list is already sorted by "aggregation_bit" and you can skip the sort-by-aggregation-bit step entirely (because it is already done, by construction).

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -100,33 +261,35 @@ async fn bit_decompose_breakdown_key<F: Field>(
...
 ) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
-    // Sort by aggregation_bit
-    let sorted_by_aggregation_bit = sort_by_aggregation_bit(ctx.clone(), input).await?;
+    // Prepend the aggregation_bit bits with breakdown_key bits. This results
+    // in sorting the sidecar by aggregation_bit first, and then by breakdown_key.
+
+    let mut permutation_keys = vec![input
+        .iter()
+        .map(|x| x.aggregation_bit.clone())
+        .collect::<Vec<_>>()];
```
benjaminsavage, line 274: Yeah, none of this should be necessary. We can construct the list in such a way that it is already sorted.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -100,33 +261,35 @@ async fn bit_decompose_breakdown_key<F: Field>(
...
     );
 
+    permutation_keys.append(&mut breakdown_keys);
+
     let sort_permutation = generate_permutation_and_reveal_shuffled(
         ctx.narrow(&Step::GeneratePermutationByBreakdownKey),
-        &breakdown_keys,
-        F::Integer::BITS,
+        &permutation_keys,
+        F::Integer::BITS + 1, // breakdown_keys bits + aggregation_bit bit
```
benjaminsavage, line 286: This is wasted work.

Assuming my maximum breakdown key is 8, then you only need to sort by the least significant 3 bits of the bit-decomposed breakdown keys. Assuming you're using Fp32BitPrime, then F::Integer::BITS is 32... so you'll be doing a radix sort on 32 bits... even though we know in advance that the list is sorted after just 3 bits and everything else is a zero.

src/protocol/attribution/aggregate_credit.rs
benjaminsavage: OK. I see. You're correct.

We will definitely need to clearly document this behavior. If callers of this API are unaware of this, they could lead to massive waste.

src/protocol/attribution/aggregate_credit.rs
benjaminsavage: I left a comment explaining it below. See my comment below next to the line:
```
F::Integer::BITS + 1, // breakdown_keys bits + aggregation_bit bit
```

benjaminsavage (changes requested):
Please eliminate the sorting by aggregation bit. We can avoid that.

Also, please only do radix sort for the minimum number of bits necessary, which is a function of the MAX_BREAKDOWN_KEY.

benjaminsavage (approved):
Awesome! And with that, we basically have everything we need for IPA (except generating random noise)!

PR #328, opened 2022-12-07
Making Docs better

richajaindce:
Ben Case wants to look at the docs in an easy to understand way. This PR is making cargo doc work removing errors and also making the sort protocol docs show up properly
PS: to open cargo doc with private modules use following command:
cargo doc --document-private-items --open


src/helpers/network.rs
```diff
@@ -55,9 +55,9 @@ impl Debug for ChannelId {
     }
 }
 
-/// Wrapper around a [`PollSender`] to modify the error message to match what the [`Network`] trait
+/// Wrapper around a [`PollSender`] to modify the error message to match what the `Network` trait
 /// requires. The only error that [`PollSender`] will generate is "channel closed", and thus is the
-/// only error message forwarded from this [`Sink`].
+/// only error message forwarded from this `Sink`.
```
benjaminsavage, line 60: Why did you remove the square braces from "Network" and "Sink" but not "PollSender"?

src/net/server/mod.rs
```diff
@@ -72,14 +72,14 @@ impl MpcHelperServerError {
     }
 }
 
-/// [`From`] implementation for [`MpcServerError::InvalidHeader`]
+/// [`From`] implementation for [`MpcHelperServerError::InvalidHeader`]
```
benjaminsavage, line 75: Let's just remove all of these useless comments that say the same thing as the line below. They add no value and as we see here - have a tendency to fall out of sync.

src/net/server/mod.rs
```diff
@@ -127,9 +127,9 @@ impl IntoResponse for MpcHelperServerError {
     }
 }
 
-/// Provides a mapping of [`QueryId`]s to senders that forward data to a [`Network`]. Every time a
-/// new query is started, a [`Network`] is created to handle communication for that query. When the
-/// server receives a request, it must know which [`Network`] to forward that request to, so it
+/// Provides a mapping of [`QueryId`]s to senders that forward data to a `Network`. Every time a
```
benjaminsavage, line 130: If we're going to be consistent about removing square braces, let's remove these as well.

src/net/server/mod.rs
```diff
@@ -208,7 +208,7 @@ pub enum BindTarget {
 
 /// Contains all of the state needed to start the MPC server.
 /// For now, stub out gateway with simple send/receive
-/// TODO (ts): replace stub with real thing when [`Network`] is implemented
+/// TODO (ts): replace stub with real thing when `Network` is implemented
```
benjaminsavage, line 211: @akoshelev , @thurstonsand - can we remove this TODO now?

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -55,20 +55,20 @@ where
 /// Transpose rows of bits into bits of rows
 ///
 /// input:
-/// [
-///   [ row[0].bit0, row[0].bit1, ..., row[0].bit31 ],
-///   [ row[1].bit0, row[1].bit1, ..., row[1].bit31 ],
+/// `\[`
+///  `\[`row`\[`0`\]`.bit0,row`\[`0`\]`.bit1, ...,row`\[`0`\]`.bit31`\]`
```
benjaminsavage, line 59: I *think* you can just put one backtick at the start and one at the end, and treat the entire line as a math expression.

src/protocol/sort/compose.rs
```diff
@@ -4,24 +4,27 @@ use embed_doc_image::embed_doc_image;
...
 /// Steps
-/// ![Compose steps][compose]
+///
 /// 1. Generate random permutations using prss
 /// 2. First permutation (sigma) is shuffled with random permutations
 /// 3. Reveal the permutation
 /// 4. Revealed permutation is applied locally on another permutation shares (rho)
 /// 5. Unshuffle the permutation with the same random permutations used in step 2, to undo the effect of the shuffling
+///
+/// ![Compose steps][compose]
```
benjaminsavage, line 27: What does this do?

src/helpers/network.rs
richajaindce: The code has changed and Network and Sink no longer exist. Tweaking the comment for now

src/protocol/sort/compose.rs
richajaindce: This shows the image in the UI like this

src/protocol/attribution/aggregate_credit.rs
richajaindce: Thanks for this! I wasted some time to get the backticks in :(

src/net/server/mod.rs
richajaindce: I kept them around. Changed the names

akoshelev (approved):

PR #331, opened 2022-12-08
Malicious gen perm

richajaindce:


richajaindce:
Returns a sort permutation in a malicious context.
This runs sort in a malicious context. The caller is responsible to validate the accumulater contents and downgrade context to Semi-honest before calling this function
The function takes care of upgrading and validating while the sort protocol runs.
It then returns a semi honest context with output in Replicated format. The caller should then upgrade the output and context before moving forward
Steps
1. [Malicious Special] Upgrade the context from semihonest to malicious and get a validator
2. [Malicious Special] Upgrade 0th sort bit keys
3. Compute bit permutation that sorts 0th bit
For 1st to N-1th bit of input share
1. i. Shuffle the i-1th composition
   ii. [Malicious Special] Validate the accumulator contents
  iii. [Malicious Special] Malicious reveal
  iv. [Malicious Special] Downgrade context to semihonest
2. i. [Malicious Special] Upgrade ith sort bit keys
   ii. Sort ith bit based on i-1th bits by applying i-1th composition on ith bit
3. Compute bit permutation that sorts ith bit
4. Compute ith composition by composing i-1th composition on ith permutation
In the end, following is returned
  i. n-1th composition: This is the permutation which sorts the inputs
 ii. Validator which can be used to validate the leftover items in the accumulator

src/protocol/sort/generate_permutation.rs
```diff
@@ -34,6 +35,11 @@ pub struct RevealedAndRandomPermutations {
     pub randoms_for_shuffle: (Vec<u32>, Vec<u32>),
 }
 
+#[derive(Debug)]
+pub struct RevealPermutationOutput<'a, F: Field> {
+    pub semihonest_context: Option<SemiHonestContext<'a, F>>,
+    pub reveal_and_random_permutation: RevealedAndRandomPermutations,
+}
```
martinthomson, line 42:
```suggestion
}

```

src/protocol/sort/generate_permutation.rs
```diff
@@ -174,19 +203,143 @@ pub async fn generate_permutation_and_reveal_shuffled<F: Field>(
...
+) -> Result<(MaliciousValidator<'a, F>, Vec<MaliciousReplicated<F>>), Error>
+where
+    F: Field,
+{
+    let mut malicious_validator = MaliciousValidator::new(ctx);
+    let mut m_ctx = malicious_validator.context();
+    let m_ctx_0 = m_ctx.narrow(&Sort(0));
+    assert_eq!(sort_keys.len(), num_bits as usize);
+
+    let upgraded_sort_keys = m_ctx.upgrade_vec(0, sort_keys[0].clone()).await?;
```
martinthomson, line 259: I think that maybe you want to use a step name for the upgrades so that you don't have to continually update the starting number.  That means using a version based on `upgrade_with()`.

src/protocol/sort/generate_permutation.rs
```diff
@@ -110,27 +130,32 @@ where
...
+                reveal_permutation_output
+                    .reveal_and_random_permutation
                     .randoms_for_shuffle
                     .1
                     .as_slice(),
             ),
-            &revealed_and_random_permutations.revealed,
+            &reveal_permutation_output
+                .reveal_and_random_permutation
+                .revealed,
```
martinthomson, line 158: This really looks like it should be a method on `reveal_permutation_output` given how much rummaging around it does in the internals of that object.

src/protocol/sort/generate_permutation.rs
```diff
@@ -143,16 +168,20 @@ where
...
```
martinthomson, line 184: Same as above.

src/protocol/sort/generate_permutation.rs
```diff
@@ -62,13 +69,26 @@ pub(super) async fn shuffle_and_reveal_permutation<
     )
     .await?;
 
-    // TODO: THIS IS WHERE VALIDATOR WILL BE CALLED!
+    let mut semihonest_context = None;
+    if let Some(malicious_validator) = malicious_validator {
+        semihonest_context = Some(
+            malicious_validator
+                .validate::<MaliciousReplicated<F>>(None)
```
martinthomson, line 76: Rather than allow this to take an `Option`, we should provide a method that does nothing.  Though maybe that is not what is going on here.  We might be able to pass this the revealed permutation.  Nothing states that the Downgrade needs to do anything...

src/protocol/malicious.rs
```diff
@@ -196,7 +196,10 @@ impl<'a, F: Field> MaliciousValidator<'a, F> {
     ///
     /// ## Panics
     /// Will panic if the mutex is poisoned
-    pub async fn validate<D: DowngradeMalicious>(self, values: D) -> Result<D::Target, Error> {
+    pub async fn validate<D: DowngradeMalicious>(
+        self,
+        values: Option<D>,
+    ) -> Result<(SemiHonestContext<'a, F>, Option<D::Target>), Error> {
```
martinthomson, line 202: The other option we have is for the caller of the `MaliciousValidator` to keep the original context, so that they can keep using it.  Ostensibly, we can then think of the malicious upgrade as a `.narrow()` call.  You need to validate the output to get the answer, but you are able to keep using the input context, like so:

```rust
fn do_something_complex(c: C) where C: Context {
  let v = MaliciousValidator(c.narrow(&Step::FirstMaliciousStep));
  let m_input = v.context().upgrade(input);
  let m_output = do_something_malicious(v.context(), m_input);
  let output = v.validate(m_output);

  // and do similar again, because you still own `c`...
} 
```

That saves us from passing the context through here like this - it doesn't even demand a change to the API here.

src/protocol/sort/generate_permutation.rs
martinthomson: Carrying that thought forward...

If we make `Downgrade` an async trait, then you might consider a version where your secret-shared permutations are "downgraded" by revealing them.  We could pass in the malicious context to the downgrade function, even, allowing you to invoke `reveal()` using that context.

That's a lot more complicated, but I think that it hides some of the complexity.  What do you think?

src/protocol/sort/generate_permutation.rs
```diff
@@ -259,28 +412,95 @@ mod tests {
...
+        let mut expected = match_keys
+            .iter()
+            .map(|mk| u64::from(*mk))
+            .collect::<Vec<_>>();
+        expected.sort_unstable();
+
+        let result = world
+            .semi_honest(
+                match_keys.clone(),
+                |ctx: SemiHonestContext<Fp31>, mk_shares| async move {
```
martinthomson, line 483: Do you really need the type specification here?

martinthomson (commented):
Yep, this is a pretty gnarly change to manage.  I have a few ideas here, some more likely to be useful than others.  I think that the idea that you retain a context and then narrow before passing it to the malicious validator is probably the best of those...

src/protocol/malicious.rs
benjaminsavage: I agree. I don't like returning the validation context... that's a rather special context.

benjaminsavage: I don’t think it’s a problem. As with everything else - we need to ensure our code is correct :).

src/protocol/sort/generate_permutation.rs
richajaindce: This is a great suggestion! Let me try to code this and see if I can make it work

richajaindce: I tried to make the Downgrade trait async and it propagated to validate, and ThisCodeIsAuthorizedToDowngradeFromMalicious. Still trying to resolve errors, diff can be found at https://github.com/richajaindce/raw-ipa/tree/async_malicious

src/protocol/sort/generate_permutation.rs
richajaindce: This is going away thanks to the new downgrade trait implementation!

src/protocol/sort/generate_permutation.rs
richajaindce: It needs the type for field and not the context actually.

src/secret_sharing/malicious_replicated.rs
```diff
@@ -140,42 +153,59 @@ impl<F: Field> Mul<F> for MaliciousReplicated<F> {
...
+            output.1.access_without_downgrade(),
         ))
     }
 }
 
+#[async_trait]
+impl<'a, F: Field, S: SecretSharing<F>> Downgrade for ShuffledPermutationWrapper<'a, F, S> {
+    type Target = Vec<u32>;
+    async fn downgrade(self) -> UnauthorizedDowngradeWrapper<Self::Target> {
+        let output = reveal_permutation(self.ctx.context().narrow(&RevealPermutation), &self.val)
```
richajaindce, line 186: mismatched types
expected reference `&[MaliciousReplicated<F>]`
   found reference `&Vec<S>`
for self.val

src/test_fixture/sharing.rs
```diff
@@ -170,12 +170,34 @@ impl<F: Field> Reconstruct<F> for [&Replicated<F>; 3] {
     }
 }
 
+impl<F: Field> Reconstruct<F> for [&MaliciousReplicated<F>; 3] {
```
martinthomson, line 173: You shouldn't need this function; the sort returns the validator and so you can validate to downgrade the malicious shares that you have.

src/test_fixture/sharing.rs
```diff
@@ -170,12 +170,34 @@ impl<F: Field> Reconstruct<F> for [&Replicated<F>; 3] {
...
+    }
+}
+
 impl<F: Field> Reconstruct<F> for [Replicated<F>; 3] {
     fn reconstruct(&self) -> F {
         [&self[0], &self[1], &self[2]].reconstruct()
     }
 }
 
+impl<F: Field> Reconstruct<F> for [MaliciousReplicated<F>; 3] {
```
martinthomson, line 195: As above.

src/test_fixture/sharing.rs
```diff
@@ -170,12 +170,34 @@ impl<F: Field> Reconstruct<F> for [&Replicated<F>; 3] {
     }
 }
 
+impl<F: Field> Reconstruct<F> for [&MaliciousReplicated<F>; 3] {
+    fn reconstruct(&self) -> F {
+        use crate::secret_sharing::ThisCodeIsAuthorizedToDowngradeFromMalicious;
+        let s0 = self[0];
+        let s1 = self[1];
+        let s2 = self[2];
+        [s0.rx(), s1.rx(), s2.rx()].reconstruct();
```
martinthomson, line 179: Note that this isn't really needed.  It only validates that the shares are sensible, not that this is `r` times `x`.

src/test_fixture/sharing.rs
```diff
@@ -251,23 +275,26 @@ where
...
 
         for (m0, (m1, m2)) in zip(self[0].iter(), zip(self[1].iter(), self[2].iter())) {
-            [m0, m1, m2].validate(r);
+            [m0, m1, m2].validate(r).await;
         }
     }
 }
 
+#[async_trait]
+#[allow(unused_must_use)]
```
martinthomson, line 291: See above; you should suppress the warning at the call site, not here.

src/test_fixture/world.rs
```diff
@@ -237,6 +237,7 @@ where
         <[_; 3]>::try_from(output).unwrap()
     }
 
+    #[allow(unused_must_use)]
```
martinthomson, line 240: Please change the test to use the output if that is what you need to do.

martinthomson (commented):
Just a few nits, but largely good.

The transposition thing is something we've already noted elsewhere.  Don't fix it here, but maybe consider planning a fix.

src/protocol/sort/generate_permutation.rs
```diff
@@ -182,11 +228,125 @@ pub async fn generate_permutation_and_reveal_shuffled<F: Field>(
...
+    sort_keys: &[Vec<Replicated<F>>],
+    num_bits: u32,
+) -> Result<(MaliciousValidator<'a, F>, Vec<MaliciousReplicated<F>>), Error>
+where
+    F: Field,
+{
+    let mut malicious_validator = MaliciousValidator::new(sh_ctx.narrow(&MaliciousUpgradeContext));
+    let mut m_ctx = malicious_validator.context();
+    let m_ctx_0 = m_ctx.narrow(&Sort(0));
+    assert_eq!(sort_keys.len(), num_bits as usize);
```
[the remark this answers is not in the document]
richajaindce: Yes we have been doing a back and forth on what's the best way to represent this. Lets handle this later

src/test_fixture/sharing.rs
richajaindce: True.. Changed this

src/test_fixture/sharing.rs
richajaindce: Removing this since not needed anymore

src/test_fixture/sharing.rs
richajaindce: The compiler is confusing between MaliciousValidator::validate and ValidateMalicious::validate. Let me rename ValidateMalicious::validate and then it should be ok

src/test_fixture/world.rs
richajaindce: Realised this validate does not need to be async anymore.

src/protocol/context/malicious.rs
```diff
@@ -72,6 +74,23 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
...
     }
 
+    /// Upgrade an input vector using this context.
+    /// # Errors
+    /// When the multiplication fails. This does not include additive attacks
+    /// by other helpers.  These are caught later.
+    pub async fn upgrade_vector<SS: Substep>(
+        &self,
+        step: &SS,
+        input: Vec<Replicated<F>>,
```
martinthomson, line 84:
```suggestion
        input: impl IntoIter<Item = Replicated<F>>,
```
Or the equivalent with generic parameters.  I wonder if we'll regret not tackling the streaming problem sooner...

martinthomson (approved):

PR #333, opened 2022-12-08
XorSecretShare struct

taikiy:
Implement `XorSecretShare`. Internally, it uses `BitVec` as the storage, and stores an arbitrary number of bits specified by `N`.

This will soon replace the current `XorReplicated` struct, which is a mix of replicated secret sharing and xor secret sharing. Below is an overview of the new secret sharing arrangement.

Untitled-2022-12-08-1530

### DIFF
--- a/src/secret_sharing/xor.rs
+++ b/src/secret_sharing/xor.rs
@@ -29,6 +29,7 @@ impl XorReplicated {
 
 /// Bit store size of `XorSecretShare` in bytes.
 const BIT_STORE_SIZE_IN_BYTES: usize = 4;
+type BitStore = BitArr!(for BIT_STORE_SIZE_IN_BYTES * 8, in u8, Lsb0);
 
 /// XOR secret share.
 ///
@@ -47,12 +48,12 @@ const BIT_STORE_SIZE_IN_BYTES: usize = 4;
 /// assert_eq!(b1, 1);
 /// ```
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub struct XorSecretShare(BitArray<[u8; BIT_STORE_SIZE_IN_BYTES], Lsb0>);
+pub struct XorSecretShare(BitStore);
 
 impl XorSecretShare {
     const SIZE_IN_BYTES: usize = BIT_STORE_SIZE_IN_BYTES;
     #[allow(dead_code)]
-    const ZERO: Self = Self(BitArray::<[u8; Self::SIZE_IN_BYTES], Lsb0>::ZERO);
+    const ZERO: Self = Self(BitStore::ZERO);
 }
 
 impl BitXor for XorSecretShare {
@@ -72,7 +73,7 @@ impl BitXorAssign for XorSecretShare {
 
 impl<T: Into<u128>> From<T> for XorSecretShare {
     fn from(v: T) -> Self {
-        Self(BitArray::<_, Lsb0>::new(
+        Self(BitStore::new(
             v.into().to_le_bytes()[0..Self::SIZE_IN_BYTES]
                 .try_into()
                 .unwrap(),
--- a/src/secret_sharing/xor.rs
+++ b/src/secret_sharing/xor.rs
@@ -27,6 +27,9 @@ impl XorReplicated {
     }
 }
 
+/// Bit store size of `XorSecretShare` in bytes.
+const BIT_STORE_SIZE_IN_BYTES: usize = 4;
+
 /// XOR secret share.
 ///
 /// Bits are stored in the Little-Endian format. Accessing the first element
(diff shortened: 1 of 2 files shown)


src/secret_sharing/xor.rs
```diff
@@ -20,3 +26,155 @@ impl XorReplicated {
...
+///
+/// let s = XorSecretShare::<64>::from(2_u128);
+/// let b0 = s[0];
+/// let b1 = s[1];
+///
+/// assert_eq!(b0, 0);
+/// assert_eq!(b1, 1);
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct XorSecretShare<const N: usize>(BitVec<u8, Lsb0>);
```
akoshelev, line 47: I think you want `BitArr` here. I believe you can also omit `Lsb0` because that's the default.

src/secret_sharing/xor.rs
```diff
@@ -20,3 +26,155 @@ impl XorReplicated {
...
+/// instead.
+impl<const N: usize, T: Into<u128>> From<T> for XorSecretShare<N> {
+    fn from(v: T) -> Self {
+        Self(BitVec::from_slice(
+            &v.into().to_le_bytes()[0..Self::SIZE_IN_BYTES],
+        ))
+    }
+}
+
+impl<const N: usize> Index<usize> for XorSecretShare<N> {
```
akoshelev, line 95: you would probably need `Index<u8>` too, given that matchkeys are at most 40 bits long and we iterate over bits using `u8`?

src/secret_sharing/xor.rs
```diff
@@ -20,3 +26,155 @@ impl XorReplicated {
...
+// here allows us to store bits larger than 32 but less than 64 most
+// effectively (i.e., 40-bit = 5 * `u8` blocks).
+
+impl<const N: usize> XorSecretShare<N> {
+    const BITS: usize = N;
+    const SIZE_IN_BYTES: usize = Self::BITS / 8;
+
+    #[allow(dead_code)]
+    #[must_use]
+    pub fn new(v: BitVec<u8, Lsb0>) -> Self {
```
akoshelev, line 59: doesn't it need two values because it is secret share? also asking callers to prepare a bit array/vec is rough, you may want to make it private and provide convenient `From` implementation

akoshelev (changes requested):

src/secret_sharing/xor.rs
martinthomson: This is [`BitArray`](https://docs.rs/bitvec/latest/bitvec/array/struct.BitArray.html), but that has a drawback in that const generic expressions aren't available yet.  I don't know if we'll hit that, but we should try to avoid the heap allocation for these so that we can keep this lightweight.  (We're explicitly not `Copy` on these, but it would be good if we can approximate it.)

martinthomson (commented):
As Alex notes, this probably needs to be a replacement for `XorReplicated`, with `left` and `right` values.  You might need a trait for the individual values, but it's possible that the `const` generic argument is enough.

I also think that we'll want a trait for this, so that we can describe requirements for shares (Add and Mul, ... for arithmetic, And, Xor, Or, ... for XOR).

That is, we have a base trait that has no operations supported, but can do `Clone` and a `Resharable` and whatever else.  Then Arithmetic and Xor variants that include more trait bounds.

src/secret_sharing/xor.rs
akoshelev: I am confused, I thought https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html is heap-allocated?

taikiy: I'll change it to `BitArray`. I wanted to explicitly specify `Lsb0` so that readers are aware of this constraint.

src/secret_sharing/xor.rs
taikiy: Per offline discussion, this is intended to not have left/right for this struct. This represents a single secret-shared value. I already have `From` implementation, so I'll delete `new`. The field is already private.

src/secret_sharing/xor.rs
taikiy: Our code make use of `enumerate()` a lot, where we index individual elements via `i: usize`. I think we can avoid redundant `try_into().unwrap()` calls by using `usize` here, unless there's a big drawback.

src/secret_sharing/xor.rs
```diff
@@ -20,3 +26,155 @@ impl XorReplicated {
...
+impl<const N: usize, T: Into<u128>> From<T> for XorSecretShare<N> {
+    fn from(v: T) -> Self {
+        Self(BitVec::from_slice(
+            &v.into().to_le_bytes()[0..Self::SIZE_IN_BYTES],
+        ))
+    }
+}
+
+impl<const N: usize> Index<usize> for XorSecretShare<N> {
+    type Output = u8;
```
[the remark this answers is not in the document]
taikiy: Code paths where we access bits in xor secret shares are usually converting a bit to a Field value. I think this alleviates bool -> integer conversions at call sites.

taikiy:
Let me withdraw this PR and think about a better structures. I'm starting to think that we don't need structures for a single secret shared value, but just implement arithmetic/boolean traits on additive/xor replicated structures.

PR #336, opened 2022-12-08
Move IntoShares trait and friends to secret_sharing mod

akoshelev:
Test client I am making to benchmark and execute different scenarios needs to be able to secret share user input before sending it to different helpers.

Helper server implementation does not need that, so trait is hidden behind a feature flag

@martinthomson we discussed this and my initial approach was to implement a similar trait for CLI use, but it turned out to be exactly the same as `IntoShares`  and its implementations. Hope that exposing it to `cli` and `test_fixture` is good enough to prevent it from leaking to prod code

### DIFF
--- a/src/secret_sharing/into_shares.rs
+++ b/src/secret_sharing/into_shares.rs
@@ -1,0 +1,76 @@
+use crate::ff::Field;
+use crate::rand::{thread_rng, Rng, RngCore};
+use crate::secret_sharing::Replicated;
+use rand::distributions::{Distribution, Standard};
+
+#[cfg(any(feature = "test-fixture", feature = "cli"))]
+pub trait IntoShares<T>: Sized {
+    fn share(self) -> [T; 3] {
+        self.share_with(&mut thread_rng())
+    }
+    fn share_with<R: Rng>(self, rng: &mut R) -> [T; 3];
+}
+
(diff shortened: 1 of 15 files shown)


src/secret_sharing/into_shares.rs
```diff
@@ -0,0 +1,76 @@
...
+{
+    fn share_with<R: Rng>(self, rng: &mut R) -> [(T, U); 3] {
+        let [a0, a1, a2] = self.0.share_with(rng);
+        let [b0, b1, b2] = self.1.share_with(rng);
+        [(a0, b0), (a1, b1), (a2, b2)]
+    }
+}
+
+/// Shares `input` into 3 replicated secret shares using the provided `rng` implementation
+pub fn share<F: Field, R: RngCore>(input: F, rng: &mut R) -> [Replicated<F>; 3]
```
martinthomson, line 63: I wonder if we can inline this into the `Replicated<F>` impl above now.  Do you know who uses this?

martinthomson (approved):

src/secret_sharing/into_shares.rs
akoshelev: good idea. There were plenty of tests that used `share`, but nothing else.

PR #337, opened 2022-12-09
Bring back field tests accidentally removed

akoshelev:
Some time ago, field serde tests were removed, I think that was a mistake and this change brings them back.
It is a good time to move some of the tests inside the macro that generates the field implementation to ensure we run them for all field types.

### DIFF
--- a/src/ff/prime_field.rs
+++ b/src/ff/prime_field.rs
@@ -1,5 +1,9 @@
 use super::{field::BinaryField, Field};
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
+// something weird is happening and compiler does not see the use of this macro inside field_impl
+// but it happily complains if this import is removed.
+#[allow(unused_imports)]
+use paste::paste;
 
 macro_rules! field_impl {
     ( $field:ty, $int:ty ) => {
@@ -112,6 +116,60 @@ macro_rules! field_impl {
                 *self == other.as_u128()
             }
         }
+
+        #[cfg(test)]
+        paste! {
+            #[cfg(test)]
+            mod [< $field:lower _tests >] {
+                use super::*;
+                use proptest::proptest;
+
+                #[test]
+                fn zero() {
+                    let prime = u128::from($field::PRIME);
+                    assert_eq!($field::ZERO, $field::from(prime), "from takes a modulus",);
+                    assert_eq!($field::ZERO, $field::ZERO + $field::ZERO);
+                    assert_eq!($field::ZERO, $field::ZERO - $field::ZERO);
+                    assert_eq!($field::from(prime - 1), $field::ZERO - $field::ONE);
+                    assert_eq!($field::ZERO, $field::ZERO * $field::ONE);
+                }
+
+                #[test]
+                fn ser_not_enough_capacity() {
+                    let mut buf = [0; 0];
+                    assert!(matches!(
+                        $field::ONE.serialize(&mut buf),
+                        Err(e) if e.kind() == std::io::ErrorKind::WriteZero));
+                }
+
+                #[test]
+                fn de_buf_too_small() {
+                    let mut buf = [0; 0];
+                    assert!(matches!(
+                                    $field::deserialize(&mut buf),
+                                    Err(e) if e.kind() == std::io::ErrorKind::UnexpectedEof));
+                }
+
+                #[test]
+                fn can_write_into_buf_larger_than_required() {
+                    let mut buf = vec![0_u8; $field::SIZE_IN_BYTES as usize + 1];
+
+                    // panic will show the error while assert will just tell us that something went wrong
+                    $field::ONE.serialize(&mut buf).unwrap();
+                }
+
+                proptest! {
+                    #[test]
+                    fn serde(v in 0..$field::PRIME) {
+                        let field_v = $field(v);
+                        let mut buf = vec![0; $field::SIZE_IN_BYTES as usize];
+                        field_v.serialize(&mut buf).unwrap();
+
+                        assert_eq!(field_v, $field::deserialize(&mut buf).unwrap());
+                    }
+                }
+            }
+        }
     };
 }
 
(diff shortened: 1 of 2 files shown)


src/ff/prime_field.rs
```diff
@@ -1,5 +1,9 @@
 use super::{field::BinaryField, Field};
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
+// something weird is happening and compiler does not see the use of this macro inside field_impl
+// but it happily complains if this import is removed.
+#[allow(unused_imports)]
+use paste::paste;
```
martinthomson, line 6: Rust's approach to identifying macros is annoyingly inconsistent with how it identifies other things.  Otherwise, I would suggest you just change the below to "paste::paste! {".

But maybe we can avoid depending on that...

src/ff/prime_field.rs
```diff
@@ -112,6 +116,60 @@ macro_rules! field_impl {
...
+                    assert_eq!($field::from(prime - 1), $field::ZERO - $field::ONE);
+                    assert_eq!($field::ZERO, $field::ZERO * $field::ONE);
+                }
+
+                #[test]
+                fn ser_not_enough_capacity() {
+                    let mut buf = [0; 0];
+                    assert!(matches!(
+                        $field::ONE.serialize(&mut buf),
+                        Err(e) if e.kind() == std::io::ErrorKind::WriteZero));
```
martinthomson, line 142: Want to do two tests with capacity 0 and capacity `size_of!(Self)` (or `Self::SIZE_IN_BYTES`, I guess) ?

src/ff/prime_field.rs
```diff
@@ -112,6 +116,60 @@ macro_rules! field_impl {
...
+                }
+
+                #[test]
+                fn de_buf_too_small() {
+                    let mut buf = [0; 0];
```
martinthomson, line 147: As above.

src/ff/prime_field.rs
```diff
@@ -112,6 +116,60 @@ macro_rules! field_impl {
...
+
+                #[test]
+                fn can_write_into_buf_larger_than_required() {
+                    let mut buf = vec![0_u8; $field::SIZE_IN_BYTES as usize + 1];
+
+                    // panic will show the error while assert will just tell us that something went wrong
+                    $field::ONE.serialize(&mut buf).unwrap();
+                }
+
+                proptest! {
```
martinthomson, line 161: How much work does this take when it passes?  It probably doesn't take much, but what assurance do we have?  An exhaustive search on a large field would be awkward.

src/ff/prime_field.rs
```diff
@@ -211,14 +269,6 @@ mod test {
     use super::{Field, Fp2, Fp31, Fp32BitPrime};
 
     #[allow(clippy::eq_op)]
```
martinthomson, line 271: Might need to drop the `#[allow(...)]` now.

src/ff/prime_field.rs
```diff
@@ -112,6 +116,60 @@ macro_rules! field_impl {
                 *self == other.as_u128()
             }
         }
+
+        #[cfg(test)]
+        paste! {
+            #[cfg(test)]
+            mod [< $field:lower _tests >] {
```
martinthomson, line 123: Hmm.  The alternative is this:

```rust
#[allow(something_about_bad_names_for_modules_maybe)]
mod $f {
  struct $f(...);
  impl $f {} // and more impls

  #[cfg(test)]
  mod test {
    #[test] fn zero() { ... }
  }
}
pub use $f::$f;
```

martinthomson (commented):
Maybe you can avoid the need to import paste.

src/ff/prime_field.rs
akoshelev: it is pretty fast as proptest doesn't really do an exhaustive search. It picks values randomly from the provided range and has configurable number of iterations (default is 100 or 1000). 

```bash
cargo test --lib ff::prime_field  -- -Zunstable-options --report-time

running 20 tests
test ff::prime_field::fp2::common_tests::zero ... ok <0.000s>
test ff::prime_field::fp31::common_tests::can_write_into_buf_larger_than_required ... ok <0.000s>
test ff::prime_field::fp2::specialized_tests::fp2 ... ok <0.000s>
test ff::prime_field::fp2::specialized_tests::fp2_binary_op ... ok <0.000s>
test ff::prime_field::fp2::common_tests::can_write_into_buf_larger_than_required ... ok <0.000s>
test ff::prime_field::fp2::common_tests::de_buf_too_small ... ok <0.000s>
test ff::prime_field::fp31::common_tests::de_buf_too_small ... ok <0.000s>
test ff::prime_field::fp31::common_tests::ser_not_enough_capacity ... ok <0.000s>
test ff::prime_field::fp2::common_tests::ser_not_enough_capacity ... ok <0.000s>
test ff::prime_field::fp31::specialized_tests::fp31 ... ok <0.000s>
test ff::prime_field::fp31::common_tests::zero ... ok <0.000s>
test ff::prime_field::fp32bit::common_tests::can_write_into_buf_larger_than_required ... ok <0.000s>
test ff::prime_field::fp32bit::common_tests::de_buf_too_small ... ok <0.000s>
test ff::prime_field::fp32bit::common_tests::ser_not_enough_capacity ... ok <0.000s>
test ff::prime_field::fp32bit::common_tests::zero ... ok <0.000s>
test ff::prime_field::fp32bit::specialized_tests::thirty_two_bit_additive_wrapping ... ok <0.000s>
test ff::prime_field::fp32bit::specialized_tests::thirty_two_bit_prime ... ok <0.000s>
test ff::prime_field::fp2::common_tests::serde ... ok <0.008s>
test ff::prime_field::fp31::common_tests::serde ... ok <0.008s>
test ff::prime_field::fp32bit::common_tests::serde ... ok <0.008s>
```

src/ff/prime_field.rs
akoshelev: would write a prop test for that, too lazy creating two tests. btw it is pretty reliable as it fails with 100% probability if `buf_capacity` is allowed to be enough to fit field values

src/ff/prime_field.rs
akoshelev: Made it work without `paste` but it is a bit clunky.

akoshelev:
@martinthomson do you think it is in mergeable state now? I got rid of the `paste` crate

thurstonsand (approved):
macros are fun eh

PR #338, opened 2022-12-09
Making a bench using IPA semi honest protocol

richajaindce:
Results
<google-sheets-html-origin>

Batchsize | Time Taken
-- | --
1000 | 18.18sec
10000 | 190secs
100K | 2469sec

It does not
1. Validate results


benches/oneshot/ipa.rs
```diff
@@ -0,0 +1,101 @@
...
+impl<F> IntoShares<IPAInputRow<F>> for IPAInputTestRow
+where
+    F: Field + IntoShares<Replicated<F>>,
+    Standard: Distribution<F>,
+{
+    fn share_with<R: Rng>(self, rng: &mut R) -> [IPAInputRow<F>; 3] {
+        let match_key_shares = MaskedMatchKey::mask(self.match_key).share_with(rng);
+        let [itb0, itb1, itb2] = F::from(self.is_trigger_bit).share_with(rng);
+        let [bdk0, bdk1, bdk2] = F::from(self.breakdown_key).share_with(rng);
+        let [tv0, tv1, tv2] = F::from(self.trigger_value).share_with(rng);
```
benjaminsavage, line 30: It makes me sad to see this code duplication. @akoshelev or @martinthomson - any ideas on how to avoid this?

benches/oneshot/ipa.rs
```diff
@@ -0,0 +1,101 @@
...
+        let test_row = IPAInputTestRow {
+            match_key: rng.gen_range(0..matchkeys_upto),
+            is_trigger_bit,
+            breakdown_key: match is_trigger_bit {
+                0 => rng.gen_range(0..MAX_BREAKDOWN_KEY), // Breakdown key is only found in source events
+                1_u128..=u128::MAX => 0,
+            },
+            trigger_value: is_trigger_bit * rng.gen_range(1..MAX_TRIGGER_VALUE), // Trigger value is only found in trigger events
+        };
+        println!("{:?}", test_row);
```
benjaminsavage, line 81: Print 100K rows? Are you sure you want to leave this in?

benches/oneshot/ipa.rs
```diff
@@ -0,0 +1,101 @@
...
+    let world = TestWorld::new_with(QueryId, config);
+    let mut rng = thread_rng();
+
+    const BATCHSIZE: u64 = 100_000;
+    let matchkeys_upto: u64 = BATCHSIZE / 4;
+    const MAX_TRIGGER_VALUE: u128 = 5;
+    const PER_USER_CAP: u32 = 3;
+    const MAX_BREAKDOWN_KEY: u128 = 4;
+
+    let mut records: Vec<IPAInputTestRow> = Vec::new();
```
benjaminsavage, line 68: maybe use the with_capacity version since we know the batch size.

src/protocol/ipa/mod.rs
```diff
@@ -214,23 +216,29 @@ where
     )
     .await?;
 
-    aggregate_credit(ctx.narrow(&Step::AggregateCredit), &user_capped_credits, 3).await
+    aggregate_credit(
+        ctx.narrow(&Step::AggregateCredit),
+        &user_capped_credits,
+        max_breakdown_key,
```
benjaminsavage, line 222: whoops... thanks for fixing my bug =)

src/protocol/ipa/mod.rs
```diff
@@ -314,7 +323,7 @@ mod tests {
 
         let result = world
             .semi_honest(records, |ctx, input_rows| async move {
-                ipa::<Fp31>(ctx, &input_rows, 20, PER_USER_CAP)
+                ipa::<Fp31>(ctx, &input_rows, 20, PER_USER_CAP, 3)
```
benjaminsavage, line 326: Can you make this a constant?
`MAX_BREAKDOWN_KEY`

src/protocol/ipa/mod.rs
```diff
@@ -332,4 +341,45 @@ mod tests {
...
+        for _ in 0..BATCHSIZE {
+            let is_trigger_bit = u128::from(rng.gen::<bool>());
+            let test_row = IPAInputTestRow {
+                match_key: rng.gen_range(0..matchkeys_upto),
+                is_trigger_bit,
+                breakdown_key: match is_trigger_bit {
+                    0 => rng.gen_range(0..MAX_BREAKDOWN_KEY), // Breakdown key is only found in source events
+                    1_u128..=u128::MAX => 0,
+                },
+                trigger_value: is_trigger_bit * rng.gen_range(1..MAX_TRIGGER_VALUE), // Trigger value is only found in trigger events
```
benjaminsavage, line 368: This is duplicate code. Can you make a `IPAInputTestRow::random(&mut rng)` method for this?

src/protocol/ipa/mod.rs
```diff
@@ -332,4 +341,45 @@ mod tests {
...
+            };
+            println!("{:?}", test_row);
```
benjaminsavage, line 370: consider removing..

benjaminsavage (changes requested):

benches/oneshot/ipa.rs
richajaindce: I left it here for debug purpose. Let me comment it rather than delete so it will be easy to compare the results UNTIL we have a way to compare expected with results

benches/oneshot/ipa.rs
richajaindce: I couldnt import from the test modules even after making them public. It would be great to have test modules exposed if it is possible

src/protocol/ipa/mod.rs
richajaindce: Can! But I would still need to expose it in two places (since this is test code). Unless @martinthomson/ @akoshelev can suggest some tips

richajaindce: Actually, I am moving it to test-fixture and that should be made accessible then

benjaminsavage (approved):
Looks good! Thanks for adding this!

benjaminsavage:
One last thing - it's mad about an unused import: `use rand::thread_rng;` can you fix that?

richajaindce:
I had fixed that in local but didnt push it. Should be good to go now

PR #340, opened 2022-12-12
Making Downgrade trait async

richajaindce:
This PR is in preparation to make sort malicious. In next PR, validate which calls downgrade will return a revealed permutation for sort.


src/test_fixture/world.rs
```diff
@@ -237,6 +237,7 @@ where
         <[_; 3]>::try_from(output).unwrap()
     }
 
+    #[allow(unused_must_use)]
```
martinthomson, line 240: Why not just mark this must_use ?

martinthomson (approved):

src/test_fixture/world.rs
richajaindce: Its funny, there is a test_fixture validate which doesn't return anything. The compiler is confusing test_fixture::validate with MaliciousValidatos::validate.

richajaindce:
Closing this since it is already covered in #331

PR #342, opened 2022-12-13
`BitArray64` data type

taikiy:
A backing data type for xor secret shares.

This is a boolean variant of `Field` backing `Replicated`. The diagram below shows the overview of the new secret sharing arrangement. This PR implements red-boxed parts.

Untitled-2022-12-08-1530

### DIFF
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -18,7 +18,7 @@ type U8_8 = BitArr!(for 64, in u8, Lsb0);
 /// ```
 /// use raw_ipa::bits::BitArray64;
 ///
-/// let s = BitArray64::from(2_u128);
+/// let s = BitArray64::try_from(2_u128).unwrap();
 /// let b0 = s[0_usize];
 /// let b1 = s[1_usize];
 ///
@@ -30,9 +30,18 @@ pub struct BitArray64(U8_8);
 
 impl BitArray for BitArray64 {
     const SIZE_IN_BYTES: usize = std::mem::size_of::<Self>();
+    const BITS: u32 = 64;
 
     #[allow(dead_code)]
     const ZERO: Self = Self(U8_8::ZERO);
+
+    fn truncate_from<T: Into<u128>>(v: T) -> Self {
+        Self(U8_8::new(
+            v.into().to_le_bytes()[0..<Self as BitArray>::SIZE_IN_BYTES]
+                .try_into()
+                .unwrap(),
+        ))
+    }
 }
 
 impl BooleanOps for BitArray64 {}
@@ -83,13 +92,22 @@ impl Not for BitArray64 {
     }
 }
 
-impl<T: Into<u128>> From<T> for BitArray64 {
-    fn from(v: T) -> Self {
-        Self(U8_8::new(
-            v.into().to_le_bytes()[0..<Self as BitArray>::SIZE_IN_BYTES]
-                .try_into()
-                .unwrap(),
-        ))
+impl TryFrom<u128> for BitArray64 {
+    type Error = String;
+
+    /// Fallible conversion from `u128` to this data type. The input value must
+    /// be `Self::BITS` long. That is, leading 0's must be longer or equal to
+    /// `Self::BITS`, or it will return an error.
+    fn try_from(v: u128) -> Result<Self, Self::Error> {
+        if 128 - v.leading_zeros() <= Self::BITS {
+            Ok(Self::truncate_from(v))
+        } else {
+            Err(format!(
+                "Bit array size {} is too small to hold the value {}.",
+                Self::BITS,
+                v
+            ))
+        }
     }
 }
 
@@ -122,17 +140,24 @@ mod tests {
 
     #[test]
     pub fn basic() {
+        let one = bitarr!(u64, Lsb0; 1);
+
         assert_eq!(BitArray64::ZERO.0, bitarr!(u64, Lsb0; 0));
-        assert_eq!(BitArray64::from(1_u128).0, bitarr!(u64, Lsb0; 1));
+        assert_eq!(BitArray64::try_from(1_u128).unwrap().0, one);
+
+        let b65 = (1_u128 << BitArray64::BITS) + 1;
+        assert!(BitArray64::try_from(b65).is_err());
         assert_eq!(
-            BitArray64::from((1_u128 << (BitArray64::SIZE_IN_BYTES * 8)) + 1).0,
-            bitarr!(u64, Lsb0; 1)
+            BitArray64::try_from(b65 & u128::from(u64::MAX)).unwrap().0,
+            one
         );
+
+        assert_eq!(BitArray64::truncate_from(b65).0, one);
     }
 
     #[test]
     pub fn index() {
-        let s = BitArray64::from((1_u128 << (BitArray64::SIZE_IN_BYTES * 8)) + 1);
+        let s = BitArray64::try_from(1_u128).unwrap();
         assert_eq!(s[0_usize], 1);
         assert_eq!(s[63_u32], 0);
     }
@@ -140,7 +165,7 @@ mod tests {
     #[test]
     #[should_panic]
     pub fn out_of_count_index() {
-        let s = BitArray64::from(1_u128);
+        let s = BitArray64::try_from(1_u128).unwrap();
         // Below assert doesn't matter. The indexing should panic
         assert_eq!(s[64_usize], 0);
     }
@@ -152,13 +177,13 @@ mod tests {
             let a = rng.gen::<u128>();
             let b = rng.gen::<u128>();
 
-            let and = BitArray64::from(a & b);
-            let or = BitArray64::from(a | b);
-            let xor = BitArray64::from(a ^ b);
-            let not = BitArray64::from(!a);
+            let and = BitArray64::truncate_from(a & b);
+            let or = BitArray64::truncate_from(a | b);
+            let xor = BitArray64::truncate_from(a ^ b);
+            let not = BitArray64::truncate_from(!a);
 
-            let a = BitArray64::from(a);
-            let b = BitArray64::from(b);
+            let a = BitArray64::try_from(a & u128::from(u64::MAX)).unwrap();
+            let b = BitArray64::try_from(b & u128::from(u64::MAX)).unwrap();
 
             assert_eq!(a & b, and);
             assert_eq!(a | b, or);
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -5,11 +5,10 @@ use std::{
     ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Index, Not},
 };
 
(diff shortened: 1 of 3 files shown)


src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
+use super::{BitArray, BooleanOps};
+use bitvec::prelude::{BitArr, Lsb0};
+use std::{
+    fmt::Debug,
+    ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Index, Not},
+};
+
+/// Bit store size of `XorSecretShare` in bytes.
+const BIT_STORE_SIZE_IN_BYTES: usize = 4;
+type U8_4 = BitArr!(for BIT_STORE_SIZE_IN_BYTES * 8, in u8, Lsb0);
```
taikiy, line 10: Using `u8` blocks as backing data type makes converting from a byte array into BitArray easier. If we ever need a backing data type that is a multiple of 8 (i.e. 40), it's also easy to just change `BIT_STORE_SIZE_IN_BYTES` rather than using `u32`. Not sure if there's a performance overhead using four blocks of `u8` instead of one block of `u32`. We can measure the difference later if we decide that 32-bit is enough for xor secret shares.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+/// assert_eq!(b1, 1);
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BitArray32(U8_4);
+
+impl BitArray for BitArray32 {
+    const SIZE_IN_BYTES: usize = BIT_STORE_SIZE_IN_BYTES;
+
+    #[allow(dead_code)]
+    const ZERO: Self = Self(U8_4::ZERO);
```
taikiy, line 36: I tried creating a `const ONE`, but couldn't do it without using a macro or ::new(). I think we are okay without it though.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn from(v: T) -> Self {
+        Self(U8_4::new(
+            v.into().to_le_bytes()[0..<Self as BitArray>::SIZE_IN_BYTES]
+                .try_into()
+                .unwrap(),
+        ))
+    }
+}
+
+impl Index<usize> for BitArray32 {
```
taikiy, line 97: In the previous PR, there were comments whether this should be `Index<u8>` or `Index<u32>`. Our code make use of enumerate() a lot, where we index individual elements via i: usize. I thought we can avoid redundant `try_into().unwrap()` calls by using usize here, unless there's a big drawback.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+    type Output = u8;
```
taikiy, line 98: I made this `u8` since the code paths where we access bits in xor secret shares are usually converting a bit into a Field value. I think this alleviates bool -> integer conversions at call sites.

src/bits/bit_array.rs
akoshelev: it is a bit ugly but works and with compile-time checks that value fits into 4 byte slice

```rust
    const ONE: Self = BitArray32(bitarr![const u8, Lsb0; 1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]);
```

src/bits/bit_array.rs
akoshelev: you could provide both if you want maximum convenience.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl Not for BitArray32 {
+    type Output = Self;
+    fn not(self) -> Self::Output {
+        Self(!self.0)
+    }
+}
+
+impl<T: Into<u128>> From<T> for BitArray32 {
```
akoshelev, line 87: I think this is fallible conversion, is there a reason for having it? we could implement `From<u32>` that is guaranteed to be infallible. If we really need to, `TryFrom<u128>` can be implemented

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn from(v: T) -> Self {
+        Self(U8_4::new(
+            v.into().to_le_bytes()[0..<Self as BitArray>::SIZE_IN_BYTES]
```
akoshelev, line 90: silently truncating higher-order bits may be surprising for the callers. as I said above, probably better to provide infallible implementation for `u32` and fallible for anything else. It causes clutter at the callsites but it is a good thing in this case imo

akoshelev: for fields I believe we don't saturate at Field::PRIME bounds which is what this implementation does. It may also be an option to provide overflowing `From` implementation (similar to `Field`)

akoshelev (changes requested):

src/bits/bit_array.rs
martinthomson: I think that you will find that we'll need a different backing at some point.  A lot of the operations we're talking about here work more neatly if this is backed by u32 or u64 (or even usize, as much as Alex hates that).

src/bits/bit_array.rs
martinthomson: I don't see why we would need a definition for ONE here.  There might be value in `~ZERO` though because from that you can get (`~ZERO >> n`).  Of course, it's easy to produce that as needed without the constant.

src/bits/bit_array.rs
martinthomson: I think that I agree.  Though that might mean changing how we do fields as well.  That is, your common definition has `TryFrom<u128>` rather than `From<u128>`.  That should still work, though we'll be using `unwrap()` more often.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn index(&self, index: usize) -> &Self::Output {
+        if self.0.as_bitslice()[index] {
+            &1
+        } else {
+            &0
+        }
+    }
+}
+
+#[cfg(test)]
```
martinthomson, line 109: You can probably just derive() this instead.  The code won't be compiled if it isn't used.  Don't forget Eq too.

martinthomson (commented):
An important question for me is whether this is enough.  A 32-bit match key is probably not enough to manage the risk of collision.

src/bits/bit_array.rs
taikiy: Should we go ahead and implement `BitArray64` backed by u64?

src/bits/bit_array.rs
taikiy: I'd prefer `TryFrom<u128>` if we were creating a public library, but for the internal use I chose fewer typings. The code below could be simplified a lot by having this conversion.

```
    let xor_share = XorReplicated::new(
        u64::try_from(b_bits_left & u128::from(u64::MAX)).unwrap(),
        u64::try_from(b_bits_right & u128::from(u64::MAX)).unwrap(),
    );
```

src/bits/bit_array.rs
taikiy: I think the name `BitArray32` makes it clear that it'll only store the lower 32 bits, similar to `Fp31` or `Fp32BitPrime` accepting `u128`? As in the example above, it helps reducing `try_from().unwrap()` and `&` masks.

With this implementation, `BitArray32::from(1 << 32)` will overflow and yields 0. Correct me if I'm wrong, but I thought saturating will yield the closest valid value, which in this case is 32 1's?

src/bits/bit_array.rs
taikiy: Changed it to `BitArray64`. I kept the backing type as `u8` because that plays nicely with `serialize`/`deserialize` for `Message` trait, converting to/from a byte slice. With `u32` or `u64`, we need to call extra `to_le_bytes` and `from_le_bytes` on each communication.

src/bits/bit_array.rs
taikiy: I added `Index<u32>`. Not we need `[0_usize]` explicit typing, but I don't think we'll be using const values to access bits anyway.

src/bits/bit_array.rs
akoshelev: iirc, fields have wrapping semantic when converted from the input large than their prime. Apologies for confusion, I meant truncating add here, not saturating. But behavior is still different from what fields do. Besides I think wrapping behavior is pretty much standard for fields, while for bit arrays and arrays in general not so much. 

I worry about truncating semantic because it is lossy and allowing such conversions may lead to subtle issues that will be hard to discover. For protocol code, I think `try_from` is the right choice - if we never convert anything more than 64 bits into bit arrays, then unwrap is (almost) free. but if we have a bug in our code that produces larger than 64 bit items, we will get a panic 

for unit tests I support providing whatever semantic you find convenient when writing them as long as it panics when you do something bad and not visible for production code. `#[cfg(test)]` attribute helps with it.

taikiy: I do understand your concern of not being able to catch subtle issues , but not sure if I would call this conversion lossy. To me, this is masking (= truncating) then converting, which is the same as this code I showed above.

```
    let xor_share = XorReplicated::new(
        u64::try_from(b_bits_left & u128::from(u64::MAX)).unwrap(),
        u64::try_from(b_bits_right & u128::from(u64::MAX)).unwrap(),
    );
```

Since PRSS and fields output u128 values, masking at call sites is what we'll end up with even if we implement `try_from` for bit arrays. We should definitely create a safety net to catch unintended issues, but I don't think fallible conversion here works.

Also, I think wrapping behavior is standard for fields when executing arithmetics, but converting from larger integer types is programmatic, not mathematical. As for bit array conversions, I think masking is pretty standard too. Do you think it'll be better if we create a different function like `mask_from`?

taikiy: Per offline discussion with @akoshelev, I implemented a fallible `TryFrom<u128>` and added `truncate_from`. `truncate_from` should make it more obvious to callers that it will truncate higher-order bits and will lose information.

src/bits/bit_array.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn not(self) -> Self::Output {
+        Self(!self.0)
+    }
+}
+
+impl TryFrom<u128> for BitArray64 {
+    type Error = String;
+
+    /// Fallible conversion from `u128` to this data type. The input value must
+    /// be `Self::BITS` long. That is, leading 0's must be longer or equal to
```
akoshelev, line 99:
```suggestion
    /// be at most `Self::BITS` long. That is, the integer value must be less than or equal to $2^{Self::BITS}$
```

akoshelev (approved):
lgtm, @martinthomson do you think we can merge it now?

martinthomson (approved):

taikiy:
My apologies for the force push. I pulled Alex's "improve docs" and removed the test iterations as Martin suggested.

martinthomson:
Ah, we were working in parallel.  I made the same changes and pushed them.

PR #344, opened 2022-12-13
Align Incoming HTTP Bytes with `Field` Bytes

thurstonsand:
re-opened from PR against branch: https://github.com/thurstonsand/raw-ipa/pull/9

When accepting a BodyStream from hyper, the values on the stream are chunks of arbitrary length (tho likely aligned with packet size) bytes. We will need to convert those bytes into fields which expect a specific byte length.

This PR contains the ByteArrStream that consumes a BodyStream, and produces a new stream that outputs arrays of exactly the right length.

We can build on top of this to then convert those arrays into Fields for running protocols downstream.


akoshelev (approved):
minor feedback

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,451 @@
...
+impl<B: Send> FromRequest<B> for FieldSize {
+    type Rejection = MpcHelperServerError;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        #[cfg_attr(feature = "enable-serde", derive(serde::Deserialize))]
+        struct FieldStr {
+            field_type: String,
+        }
+
+        let Query(FieldStr { field_type }) = req.extract().await?;
```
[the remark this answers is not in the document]
thurstonsand: yep. putting `FromRequest` behind `enable-serde` flag

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,451 @@
...
+                Some(Ok(bytes)) => {
+                    self.push_back(bytes);
+                }
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl<B: HttpBody<Data = Bytes, Error = hyper::Error> + Send + 'static> FromRequest<B>
```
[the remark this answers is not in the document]
thurstonsand: > Bytes is a 32 byte pointer and it would be a waste of space to use it to pass around 4 or 8 byte chunks

this makes sense, but the reason it's done this way right now is to try to avoid copies. Are you saying that we _should_ be copying out of upstream `Bytes` array? and so this would actually be a `Stream<Item = Vec<u8>>`? we can't use arrays because len is not known at compile time.

> Right now ByteArrStream does not support chuhks smaller that minimum

not sure I quite understand. What do you mean by "chunks smaller than minimum"?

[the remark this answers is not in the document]
akoshelev: Would I be able to ask `ByteArrStream` to give me chunks of size 400 bytes if the whole input is 100 bytes?

> Are you saying that we should be copying out of upstream Bytes array? and so this would actually be a Stream<Item = Vec<u8>>?

no, all I am saying is slicing 1Gb input into chunks of 4 bytes is too wasteful. When chunk size on the heap is smaller than the pointer size, it looks very inefficient to me. We would need to be able to carve out chunks of much much larger size than 4 bytes as long as they are aligned with the field value size.

akoshelev (approved):

src/ff/prime_field.rs
```diff
@@ -2,7 +2,7 @@ use super::{field::BinaryField, Field};
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
 
 macro_rules! field_impl {
-    ( $field:ident, $int:ty, $prime:expr ) => {
+    ( $field:ident, $int:ty, $prime:expr, $type_str:expr ) => {
```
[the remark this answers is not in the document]
martinthomson: I think that `expr` is OK; literal is the literal subset of expressions, which would probably suffice here, but you might use a const expression here that would need `expr`.

src/ff/prime_field.rs
```diff
@@ -13,6 +13,7 @@ macro_rules! field_impl {
             const PRIME: Self::Integer = $prime;
             const ZERO: Self = $field(0);
             const ONE: Self = $field(1);
+            const TYPE_STR: &'static str = $type_str;
```
martinthomson, line 16: Could you instead use `stringify!($field)`?  Then you don't need to worry so much about repeating the name in the invocation.

src/ff/field.rs
```diff
@@ -119,6 +127,20 @@ pub trait Field:
     }
 }
 
+/// Mapping between a [`Field`]'s `TYPE_STR` and its `SIZE_IN_BYTES`
+/// # Errors
+/// If unknown `type_str` is passed in
+pub fn size_in_bytes_from_type_str(type_str: &str) -> Result<u32, Error> {
```
martinthomson, line 133: If you define a new trait, then you can implement it for `&str` and then you avoid the oddly-named function.  Instead you get:

```rust
use ...::FieldTypeString;
F::TYPE_STR.size_in_bytes()
```

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,452 @@
...
+}
+
+#[cfg(feature = "enable-serde")]
+#[async_trait]
+impl<B: Send> FromRequest<B> for FieldSize {
+    type Rejection = MpcHelperServerError;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        #[derive(serde::Deserialize)]
+        struct FieldStr {
```
martinthomson, line 27: Is it possible to deserialize directly into a string?

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,452 @@
...
+            let out_bytes = self.buffered[0].split_to(self.size_in_bytes as usize);
+            self.buffered_size -= self.size_in_bytes;
+            Some(out_bytes)
+        } else {
+            // First buffer is too small, so we need to take across buffers.
+            // Will require a memcopy of u8's across buffers in order to create 1 contiguous buffer
+            // to return.
+            // Removes expended buffers after taking
+
+            let mut out_bytes = Vec::with_capacity(self.size_in_bytes as usize);
```
martinthomson, line 83: Please avoid allocating on the heap for this.  It is OK to allocate a fixed 8 bytes into an array for this.

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,452 @@
...
+                if remaining_bytes == 0 {
+                    break;
+                }
+                // current `buffer` has more bytes than needed
+                if self.buffered[0].len() > remaining_bytes {
+                    let remaining = self.buffered[0].split_to(remaining_bytes);
+                    out_bytes.extend_from_slice(&remaining);
+                } else {
+                    // current `buffer` has <= bytes needed, remove and append to out_bytes
+                    out_bytes.extend_from_slice(&self.buffered.pop_front().unwrap());
```
martinthomson, line 98: You should be able to use append for this if this is a vec.  For an array, you will need to assign:

```rust
let buf = self.buffered.pop_front();
out_bytes[next..next+buf.len()]  = buf[..];
```

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,452 @@
...
+                        std::io::ErrorKind::WriteZero,
+                        format!(
+                            "expected body to align on size {}, but has insufficient bytes {}",
+                            *this.size_in_bytes, *this.buffered_size
+                        )
+                        .into(),
+                    ))));
+                }
+                // if body is finished, this stream is finished
+                None => return Poll::Ready(None),
```
martinthomson, line 137: This should produce an error if the stream ends when the buffered_size is not zero.  In that case there will be a partial value present.

martinthomson (commented):
Some drive-by feedback.

src/net/server/bytearrstream.rs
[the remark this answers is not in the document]
thurstonsand: gotcha, i see what you're saying. and your other point makes sense now as well. I can add a TODO to figure that out

src/ff/prime_field.rs
thurstonsand: is there a way to also lowercase and/or snake_case it as well?

src/net/server/bytearrstream.rs
thurstonsand: I believe axum needs the struct in order to know what the key of the query param is

src/net/server/bytearrstream.rs
thurstonsand: the problem is that I don't know what `size_in_bytes` is at compile time. are you proposing we have a max of `8`?

src/net/server/bytearrstream.rs
thurstonsand: `self.buffered.pop_front()` in this context is a  `hyper::body::Bytes`, which can be borrowed as a slice, but isn't a `Vec`, so I'm not sure how to use `append` for that.

is there a reason to avoid `extend_from_slice`?

src/net/server/bytearrstream.rs
thurstonsand: the case above this one is

```rust
None if *this.buffered_size > 0 => {
```
is that sufficient?

src/net/server/bytearrstream.rs
martinthomson: Only that it assumes you are doing heap allocations.

src/ff/prime_field.rs
akoshelev: `paste` crate can do that but I am not aware of any std macro that can do this. It may not be that big of a deal though - this string just needs to be unique and we can do case-insensitive comparison using regex if we want to

src/net/server/bytearrstream.rs
akoshelev: `BitArr` can do it at runtime I think

src/net/server/bytearrstream.rs
akoshelev: we could add `None if this.buffered_size == 0` to make it more clear but yea I got tripped here too. this code I believe handles the case when stream ends producing insufficient bytes to create another chunk.

We would likely want to change that though and use two parameters for this stream `alignment` and `buffer_size`: https://github.com/private-attribution/ipa/pull/344/files#r1048124439

src/ff/prime_field.rs
thurstonsand: i ended up using `eq_ignore_ascii_case` which seems fine

src/net/server/bytearrstream.rs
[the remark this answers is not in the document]
thurstonsand: added: https://github.com/private-attribution/ipa/pull/344/files#diff-de21be8a069c815dcad06256ba2473fcb77d6081a62e09f29e367913e38029f3R38

src/net/server/bytearrstream.rs
thurstonsand: i took a look, and it seems like it's saying lengths must be const: https://docs.rs/bitvec/0.22.3/bitvec/macro.BitArr.html

src/ff/field.rs
```diff
@@ -102,6 +110,35 @@ pub trait Field:
...
 }
 
+pub trait FieldTypeStr {
+    /// Mapping between a [`Field`]'s `TYPE_STR` and its `SIZE_IN_BYTES`
+    /// # Errors
+    /// if self is not an existing [`'Field`]'s `TYPE_STR`
+    fn size_in_bytes(&self) -> Result<u32, Error>;
+}
+
+impl FieldTypeStr for str {
```
akoshelev, line 120:
```suggestion
impl <T: AsRef<str> FieldTypeStr for T {
```

src/net/server/bytearrstream.rs
akoshelev: but field size is const, isn't it?

src/ff/field.rs
```diff
@@ -102,6 +110,35 @@ pub trait Field:
     }
 }
 
+pub trait FieldTypeStr {
```
akoshelev, line 113: I am not sure how it would play with `FieldType` enum that we would need to support different field types used in queries.

src/net/server/bytearrstream.rs
```diff
@@ -0,0 +1,456 @@
...
+use std::collections::VecDeque;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// the size of the field type specified in the query param of the request
+/// Possible sizes are are currently from:
+/// * `fp2`
+/// * `fp31`
+/// * `fp32_bit_prime`
+struct FieldSize {
```
akoshelev, line 16: this transformation is lossy because it forgets about the actual field requested by the user. if we have two fields that happen to be the same size, we won't be able to resolve this ambiguity. Besides, we need actual field type to construct the protocol

akoshelev: there is actually more evidence that using field size as alignment for input streams is not correct. For protocols that take replicated shares as input, the correct alignment is 2 * field_size

PR #345, opened 2022-12-14
Remove roles from helper config

akoshelev:
As we discussed, each helper will be assigned a role when query processing starts. That means helpers may have different roles in different queries it participates in. Helper configuration therefore must be configured around its endpoint that stays the same regardless of number of queries it processes.

This change also removes interim structs that mirrored the end config struct, but used raw types (str, [u8]). With custom deserialize implementation it is easy to support only one hierarchy of configs

### DIFF
--- a/src/net/discovery/conf.rs
+++ b/src/net/discovery/conf.rs
@@ -32,16 +32,15 @@ impl PeerDiscovery for Conf {
 
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
-    use crate::helpers::Role;
     use crate::test_fixture::net::localhost_config;
     use hyper::Uri;
 
-    const H1_PUBLIC_KEY: &str = "13ccf4263cecbc30f50e6a8b9c8743943ddde62079580bc0b9019b05ba8fe924";
-    const H2_PUBLIC_KEY: &str = "925bf98243cf70b729de1d75bf4fe6be98a986608331db63902b82a1691dc13b";
-    const H3_PUBLIC_KEY: &str = "12c09881a1c7a92d1c70d9ea619d7ae0684b9cb45ecc207b98ef30ec2160a074";
-    const H1_URI: &str = "http://localhost:3000/";
-    const H2_URI: &str = "http://localhost:3001/";
-    const H3_URI: &str = "http://localhost:3002/";
+    const PUBLIC_KEY_1: &str = "13ccf4263cecbc30f50e6a8b9c8743943ddde62079580bc0b9019b05ba8fe924";
+    const PUBLIC_KEY_2: &str = "925bf98243cf70b729de1d75bf4fe6be98a986608331db63902b82a1691dc13b";
+    const PUBLIC_KEY_3: &str = "12c09881a1c7a92d1c70d9ea619d7ae0684b9cb45ecc207b98ef30ec2160a074";
+    const URI_1: &str = "http://localhost:3000/";
+    const URI_2: &str = "http://localhost:3001/";
+    const URI_3: &str = "http://localhost:3002/";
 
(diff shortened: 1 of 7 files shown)


akoshelev:
@thurstonsand I don't know why I can't pick you as a reviewer, so tagging you in the comment instead

src/net/discovery/mod.rs
```diff
@@ -12,28 +12,52 @@ pub enum Error {
...
         pub public_key: PublicKey,
     }
 
-    #[derive(Clone)]
+    #[derive(Clone, Debug)]
+    #[cfg_attr(feature = "enable-serde", derive(serde::Deserialize))]
     pub struct Config {
-        pub http: HttpConfig,
+        #[cfg_attr(feature = "enable-serde", serde(deserialize_with = "uri_from_str"))]
+        pub origin: Uri,
```
thurstonsand, line 33: so how does a helper know which endpoint itself is?

also, should this be represented as a map of `origin -> conf`? otherwise, when we receive mapping `secret_shares -> origin` from report collector, i guess we would have to iterate through array?

src/net/discovery/conf.rs
```diff
@@ -112,43 +53,26 @@ mod tests {
...
-            .add_source(File::from_str(EXAMPLE_CONFIG, FileFormat::Toml))
-            .build()
-            .unwrap()
-            .try_deserialize()
-            .expect("config should be valid");
-        let conf = Conf::from_file_conf(&to_conf).expect("file should contain valid values");
+        let conf = localhost_config([3000, 3001, 3002]);
 
         // H1
+        assert_eq!(conf.peers[Role::H1].origin, H1_URI.parse::<Uri>().unwrap());
```
thurstonsand, line 59: should we replace `Role::*` with just indexes? seems like there shouldn't be a concept of `Role` here at all, right?

thurstonsand (changes requested):

src/net/discovery/conf.rs
thurstonsand: to summarize, we'll need 3 different kinds of mappings right?
* this mapping here, which is `endpoint -> http conf`
* mapping that report collector tells us, which is `secret_shares -> endpoint`
* mapping to assign roles, which is `endpoint -> role`

and only by combining all 3 can we accurately initialize a "ring"

thurstonsand:
> we had the exact same problem. @thurstonsand would need to be added to https://github.com/orgs/private-attribution/teams/ipa-contributors
> @martinthomson can you please help add Thurston to this :)

thanks for the tip

src/net/discovery/mod.rs
akoshelev: I was thinking that in prod each helper would have a single configuration file where the specific helper configuration is written. So there will be no ambiguity which endpoint to pick and listen to.

for test case, it does not really matter as long as each helper picks a unique entry from the config file. so incrementing an index and returning config[index] would probably work just fine?

src/net/discovery/conf.rs
akoshelev: > should we replace Role::* with just indexes? seems like there shouldn't be a concept of Role here at all, right?

yes! Nothing in our code should remind us that helpers have hardcoded roles. 

> to summarize, we'll need 3 different kinds of mappings right?

that's right. Only one of those mappings is static which is the first one, two others are decided at runtime.

thurstonsand: 👍

src/net/discovery/mod.rs
thurstonsand: gotcha. so we have an as-of-yet undefined config that would include specific helper config.

thurstonsand (approved):

PR #346, opened 2022-12-14
Sum of products implementation

richajaindce:
This will be needed for GenMultiBitPerm: https://docs.google.com/document/d/1DbMVkBf2n4LkdHKg4Q5pHFxbvCSA8oR63sx16qRPfp0/edit?usp=sharing


martinthomson (approved):
This looks great.  Just one thing:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

-- https://rust-lang.github.io/api-guidelines/naming.html

PR #347, opened 2022-12-14
Ignore end-to-end IPA test

akoshelev:
I don't have enough patience to wait for 50 seconds. We should move it to end-to-end test folder if it is going to take that long

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -302,6 +302,7 @@ pub mod tests {
 
     #[tokio::test]
     #[allow(clippy::missing_panics_doc)]
+    #[ignore]
     pub async fn random_ipa_no_result_check() {
         const BATCHSIZE: u64 = 100;
         const PER_USER_CAP: u32 = 10;


src/protocol/ipa/mod.rs
```diff
@@ -302,6 +302,7 @@ pub mod tests {
 
     #[tokio::test]
     #[allow(clippy::missing_panics_doc)]
+    #[ignore]
```
richajaindce, line 305: How about changing batchsize to 10 or even 5 if needed to keep it running?

taikiy:
+1. The test is taking too long, and it is unlikely that this will catch bugs that individual unit tests cannot. I think this should be ran as an integration test, separately from unit tests.

src/protocol/ipa/mod.rs
taikiy: I like this test because it shows us the end-to-end performance we can use to see our progress in optimizations. It won't serve the purpose if we decrease the batch size to 10 or 5, and certainly we don't need to run it as an unit test.

PR #348, opened 2022-12-15
Fields implement `ArithmeticOps`

taikiy:
Move arithmetic-ops bounds into its own trait, and have all fields implement it.

### DIFF
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -2,23 +2,12 @@ use std::any::type_name;
 use std::fmt::Debug;
 use std::io;
 use std::io::ErrorKind;
-use std::ops::{
-    Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Mul, MulAssign,
-    Neg, Not, Shr, Sub, SubAssign,
-};
+use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
+
+use super::ArithmeticOps;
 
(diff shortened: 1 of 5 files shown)


src/ff/field.rs
```diff
@@ -2,23 +2,12 @@ use std::any::type_name;
...
-    + Copy
-    + Debug
-    + Ord
-    + Sub<Output = Self>
-    + Into<u128>
-    + Shr<u32, Output = Self>
-    + BitAnd<Self, Output = Self>
-    + PartialEq
-{
+pub trait Int: Sized + Copy + Debug + Into<u128> {
```
taikiy, line 10: I removed bounds that are not used from `Int`. Let me know if I should restore them.

src/ff/mod.rs
```diff
@@ -4,6 +4,19 @@
 
 mod field;
 mod prime_field;
+use std::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 
 pub use field::{BinaryField, Field, Int};
 pub use prime_field::{Fp2, Fp31, Fp32BitPrime};
+
+pub trait ArithmeticOps:
```
akoshelev, line 12: just fyi, I think you can implement `ArithmeticOps` for any `T` (not just field) that implements `Add`, `Sub`, etc by doing something along these lines

```rust
   impl <T> ArithmeticOps for T where T: Add<Output = Self>, ...
```

in this case you don't need to implement it for each field implementation, it will  be derived automatically for it

akoshelev (approved):

PR #351, opened 2022-12-19
Delete some code

martinthomson:
I forget why we added this, but we don't need it any more.


thurstonsand (approved):

akoshelev:
I don't mind removing it, but there is a chance I will bring it back soon - we may need it for testing query processing and there is a slim chance test world being useful for it too. But it should be easy, so lgtm

akoshelev (approved):

martinthomson:
When it comes time to test this, you can add a new function to TestWorld rather than overloading this one :)

PR #353, opened 2022-12-20
Simplify the random bits generator

martinthomson:
This won't get to the real problem we're dealing with here, but it should avoid most of the overhead in tests.

### DIFF
--- a/src/protocol/boolean/random_bits_generator.rs
+++ b/src/protocol/boolean/random_bits_generator.rs
@@ -4,142 +4,21 @@ use crate::ff::Field;
 use crate::protocol::context::Context;
 use crate::protocol::RecordId;
 use crate::secret_sharing::SecretSharing;
-use futures::{future::try_join_all, TryFutureExt};
-use std::collections::HashMap;
-use std::marker::PhantomData;
-use std::sync::Arc;
-use tokio::sync::Mutex;
-
-/// An implementation of simple ring buffer that stores `u8::MAX` items.
-///
-/// Internally, it uses two pointers. `read_pointer` keeps track of the next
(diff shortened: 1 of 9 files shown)


martinthomson:
Before:
```
rows 100 benchmark complete after 5.0614614s
```

After:
```
rows 100 benchmark complete after 3.7972748s
```

martinthomson:
For `cargo test`

Before:
```
real    0m26.459s
user    1m40.707s
sys     0m8.048s
```

After:
```
real    0m9.582s
user    0m51.094s
sys     0m4.170s
```

ლ(╹◡╹ლ)

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -4,142 +4,21 @@ use crate::ff::Field;
...
+    marker::PhantomData,
+    sync::atomic::{AtomicU32, AtomicUsize, Ordering},
+};
 
 /// A struct that pre-generates and buffers random sharings of bits from the
 /// `SolvedBits` protocol. Any protocol who wish to use a random-bits can draw
-/// one by calling `take_one()`. It will call `SolvedBits` once the stock falls
-/// below `REFILL_THRESHOLD` until it fills up the empty slots.
+/// one by calling `take_one()`. Making a generator spawns a task that generates
+/// new values asynchronously up to the given capacity.
```
taikiy, line 15: The code seems to just generate one `RandomBitsShare` for each `generate()` call?

taikiy (approved):
Thanks for this!

src/protocol/boolean/random_bits_generator.rs
martinthomson: Yeah, I fixed that already.  Failed to push.

Issue #352, opened 2022-12-19
Expand definition of `Network` layer into `Transport` layer


thurstonsand:
We currently have the concept of a `Network` layer that handles the communication between helpers. It interacts directly with the Infra layer to move data for running protocols.

the `Network` trait currently looks like this:

```rust
/// Network interface for components that require communication.
pub trait Network: Sync {
    /// Type of the channel that is used to send/receive messages to/from other helpers
    type Sink: futures::Sink<MessageChunks, Error = Error> + Send + Unpin + 'static;
    type MessageStream: Stream<Item = MessageChunks> + Send + Unpin + 'static;

    /// Returns a sink that accepts data to be sent to other helper parties.
    fn sink(&self) -> Self::Sink;

    /// Returns a stream to receive messages that have arrived from other helpers. Note that
    /// some implementations may panic if this method is called more than once.
    fn recv_stream(&self) -> Self::MessageStream;
}
```

there's a way to "send" data to another helper via the `sink()`, and "receive" data from the other helpers via the `recv_stream`. This works for an actively running query, but we don't have a way for helpers to communicate to each other about _starting_ queries.

Here's how that workflow would look, in short:

* Report Collector wants to start a new query on data that it has
* Report Collector calls an externally available "Create Query" API to one of the helpers
* That helper becomes the de facto "leader" of the request, generates a query id, and sends a "Prepare Query" message to the "follower" helpers; then, it tells the Report Collector what the new query id is
* Report Collector sends all its data to the leader using a "Send Data" API
* Leader disseminates that data to the 2 followers, then they all coordinate starting the protocol at the same time

So there is a mix of externally available APIs to initiate queries, as well as helper-to-helper communication that needs to happen before a query is even started. All of this communication should be abstracted such that it could be implemented for HTTP, InMemory, maybe protobuf, or anything else. For that, I want to propose a new trait, `Transport`:

```rust
pub trait Transport: Sync {
    type CommandStream: Stream<Item = TransportCommand> + Send + Unpin + 'static;
    type Sink: futures::Sink<TransportCommand, Error = TransportError> + Send + Unpin + 'static;

    /// To be called by an entity which will handle the events as indicated by the
    /// [`SubscriptionType`]. There should be only 1 subscriber per type.
    /// # Panics
    /// May panic if attempt to subscribe to the same [`SubscriptionType`] twice
    fn subscribe(&self, subscription_type: SubscriptionType) -> Self::CommandStream;

[...]

PR #355, opened 2022-12-21
Implement Transport Layer

thurstonsand:
expand on the concept of the `network` layer into a `transport` layer. see https://github.com/private-attribution/ipa/issues/352 for discussion


src/helpers/network.rs
```diff
@@ -1,101 +1,80 @@
...
-}
-
-impl<T: Send + 'static> NetworkSink<T> {
-    #[must_use]
-    pub fn new(sender: mpsc::Sender<T>) -> Self {
+impl<T: Transport> Network<T> {
+    pub fn new(
+        transport: T,
+        query_id: QueryId,
+        roles_to_helpers: HashMap<Role, HelperIdentity>,
```
akoshelev, line 36: arrays are indexable by `Role` and hashmap lookup has a cost, I think `[HelperIdentity; 3]` is better

src/helpers/network.rs
```diff
@@ -1,101 +1,80 @@
...
+    /// if called more than once during the execution of a query.
+    pub fn recv_stream(&self) -> impl Stream<Item = MessageChunks> {
+        let query_id = self.query_id;
+        let query_command_stream = self.transport.subscribe(SubscriptionType::Query(query_id));
 
-    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        ready!(self.project().inner.poll_close(cx))?;
-        Poll::Ready(Ok(()))
+        #[allow(unreachable_patterns)] // there will be more commands in the future
+        query_command_stream.map(move |command| match command {
```
akoshelev, line 70: nice, there is no penalty of having `Network` as an extra layer here (I believe `map` is zero-cost, could be wrong though)

akoshelev:
lgtm. will be merging this to [ipa-e2e-christmas-gift](https://github.com/private-attribution/ipa/tree/ipa-e2e-christmas-gift) branch to fast-track the implementation. We can review the whole query processing + network with the team later. #352 should indicate the direction we're going with it.

PR #357, opened 2022-12-21
Define HttpTransport

thurstonsand:
This PR covers all the work to define a full `Transport` implementation for Http.


akoshelev:
merging w/o review - New Year madness. we're very close to run IPA e2e

PR #359, opened 2022-12-22
Buffer the send/recv channels

martinthomson:
This turns out to have a measurable effect on test running time, at least on my machine.  It's worth about a second overall.

I considered tuning this with more runs, but we've got enough churn ahead of us that I am going to suggest we hold off.


akoshelev (approved):
good stuff

PR #360, opened 2022-12-22
track total record count in context

andyleiserson:
Track total_records in the context. This doesn't include the changes to the infra to track progress towards the total record count and close the channel when it gets there. (I do have a version of that, which can successfully free send buffer resources when the channel is no longer needed, but with one gross sticking point -- it sends the total record count over the `mpsc` along with ChannelId and payload with every message, for lack of a good way to communicate the total record count only once.)

### DIFF
--- a/src/helpers/messaging.rs
+++ b/src/helpers/messaging.rs
@@ -90,12 +90,49 @@ pub struct Gateway {
 
 pub(super) type SendRequest = (ChannelId, MessageEnvelope);
 
+#[derive(Clone, Copy, Debug)]
+pub enum TotalRecords {
+    Unspecified,
+    Specified(NonZeroUsize),
+
+    /// Total number of records is not well-determined. When the record ID is
+    /// counting solved_bits attempts. The total record count for solved_bits
+    /// depends on the number of failures.
+    ///
+    /// The purpose of this is to waive the warning that there is a known
+    /// number of records when creating a channel. If the warning is firing
+    /// and the total number of records is knowable, prefer to specify it
+    /// rather than use this to waive the warning.
+    Indeterminate,
+}
+
+impl TotalRecords {
+    #[must_use]
+    pub fn is_unspecified(&self) -> bool {
+        matches!(self, &TotalRecords::Unspecified)
+    }
+
+    #[must_use]
+    pub fn is_indeterminate(&self) -> bool {
+        matches!(self, &TotalRecords::Indeterminate)
+    }
+}
+
+impl From<usize> for TotalRecords {
+    fn from(value: usize) -> Self {
+        match NonZeroUsize::new(value) {
+            Some(v) => TotalRecords::Specified(v),
+            None => TotalRecords::Unspecified,
+        }
+    }
+}
+
 /// Channel end
 #[derive(Debug)]
 pub struct Mesh<'a, 'b> {
     gateway: &'a Gateway,
     step: &'b Step,
-    total_records: Option<NonZeroUsize>,
+    total_records: TotalRecords,
 }
 
 pub(super) struct ReceiveRequest {
@@ -126,7 +163,7 @@ impl Mesh<'_, '_> {
             )?;
         }
 
-        if let Some(count) = self.total_records {
+        if let TotalRecords::Specified(count) = self.total_records {
             assert!(
                 usize::from(record_id) < usize::from(count),
                 "record ID {:?} is out of range for {:?} (expected {:?} records)",
@@ -155,7 +192,7 @@ impl Mesh<'_, '_> {
     /// # Panics
     /// If the context has a total record count set and the `record_id` to be received is out of range.
     pub async fn receive<T: Message>(&self, source: Role, record_id: RecordId) -> Result<T, Error> {
-        if let Some(count) = self.total_records {
+        if let TotalRecords::Specified(count) = self.total_records {
             assert!(
                 usize::from(record_id) < usize::from(count),
                 "record ID {:?} is out of range for {:?} (expected {:?} records)",
@@ -263,25 +300,8 @@ impl Gateway {
     /// ## Panics
     /// In `debug_assertions` config, if the `ALREADY_WARNED` mutex is poisoned.
     #[must_use]
-    pub fn mesh<'a, 'b>(
-        &'a self,
-        step: &'b Step,
-        total_records: Option<NonZeroUsize>,
-    ) -> Mesh<'a, 'b> {
-        // TODO: to be changed to panic or assert once all instances are eliminated.
-        // Note: update panic docs above.
-        #[cfg(debug_assertions)]
-        {
-            use once_cell::sync::Lazy;
-            use std::collections::HashSet;
-            use std::sync::Mutex;
-
-            static ALREADY_WARNED: Lazy<Mutex<HashSet<Step>>> =
-                Lazy::new(|| Mutex::new(HashSet::new()));
-            if total_records.is_none() && ALREADY_WARNED.lock().unwrap().insert(step.clone()) {
-                tracing::warn!("creating mesh for {:?} with unknown record count", step);
-            }
-        }
+    pub fn mesh<'a, 'b>(&'a self, step: &'b Step, total_records: TotalRecords) -> Mesh<'a, 'b> {
+        debug_assert!(!total_records.is_unspecified());
         Mesh {
             gateway: self,
             step,
@@ -363,6 +383,7 @@ fn print_state(role: Role, send_buf: &SendBuffer, receive_buf: &ReceiveBuffer) {
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
     use crate::ff::Fp31;
+    use crate::helpers::messaging::TotalRecords;
     use crate::helpers::Role;
     use crate::protocol::context::Context;
     use crate::protocol::{RecordId, Step};
@@ -412,7 +433,7 @@ mod tests {
         let peer = Role::H2;
         let record_id = 1.into();
         let step = Step::default();
-        let total_records = NonZeroUsize::new(2);
+        let total_records = TotalRecords::from(2);
         let channel = &world.gateway(Role::H1).mesh(&step, total_records);
 
         channel.send(peer, record_id, v1).await.unwrap();
--- a/src/helpers/messaging.rs
+++ b/src/helpers/messaging.rs
@@ -20,6 +20,7 @@ use crate::{
 use futures::StreamExt;
 use std::fmt::{Debug, Formatter};
 use std::io;
+use std::num::NonZeroUsize;
 use std::time::Duration;
 use tinyvec::array_vec;
 use tracing::Instrument;
@@ -94,6 +95,7 @@ pub(super) type SendRequest = (ChannelId, MessageEnvelope);
 pub struct Mesh<'a, 'b> {
     gateway: &'a Gateway,
     step: &'b Step,
+    total_records: Option<NonZeroUsize>,
 }
 
 pub(super) struct ReceiveRequest {
@@ -108,6 +110,9 @@ impl Mesh<'_, '_> {
     ///
     /// # Errors
     /// Returns an error if it fails to send the message or if there is a serialization error.
+    ///
+    /// # Panics
+    /// If the context has a total record count set and the `record_id` to be sent is out of range.
     pub async fn send<T: Message>(
         &self,
         dest: Role,
@@ -119,6 +124,16 @@ impl Mesh<'_, '_> {
                                                      self.step,
                                                      format!("Message {msg:?} exceeds the maximum size allowed: {MESSAGE_PAYLOAD_SIZE_BYTES}"))
             )?;
+        }
+
+        if let Some(count) = self.total_records {
+            assert!(
+                usize::from(record_id) < usize::from(count),
+                "record ID {:?} is out of range for {:?} (expected {:?} records)",
+                record_id,
+                self.step,
+                self.total_records,
+            );
         }
 
         let mut payload = array_vec![0; MESSAGE_PAYLOAD_SIZE_BYTES];
@@ -136,7 +151,20 @@ impl Mesh<'_, '_> {
     ///
     /// # Errors
     /// Returns an error if it fails to receive the message or if a deserialization error occurred
+    ///
+    /// # Panics
+    /// If the context has a total record count set and the `record_id` to be received is out of range.
     pub async fn receive<T: Message>(&self, source: Role, record_id: RecordId) -> Result<T, Error> {
+        if let Some(count) = self.total_records {
+            assert!(
+                usize::from(record_id) < usize::from(count),
+                "record ID {:?} is out of range for {:?} (expected {:?} records)",
+                record_id,
+                self.step,
+                self.total_records
+            );
+        }
+
         let payload = self
             .gateway
             .receive(ChannelId::new(source, self.step.clone()), record_id)
@@ -231,11 +259,33 @@ impl Gateway {
     /// This method makes no guarantee that the communication channel will actually be established
     /// between this helper and every other one. The actual connection may be created only when
     /// `Mesh::send` or `Mesh::receive` methods are called.
+    ///
+    /// ## Panics
+    /// In `debug_assertions` config, if the `ALREADY_WARNED` mutex is poisoned.
     #[must_use]
-    pub fn mesh<'a, 'b>(&'a self, step: &'b Step) -> Mesh<'a, 'b> {
+    pub fn mesh<'a, 'b>(
+        &'a self,
+        step: &'b Step,
+        total_records: Option<NonZeroUsize>,
+    ) -> Mesh<'a, 'b> {
+        // TODO: to be changed to panic or assert once all instances are eliminated.
+        // Note: update panic docs above.
+        #[cfg(debug_assertions)]
+        {
+            use once_cell::sync::Lazy;
+            use std::collections::HashSet;
+            use std::sync::Mutex;
+
+            static ALREADY_WARNED: Lazy<Mutex<HashSet<Step>>> =
+                Lazy::new(|| Mutex::new(HashSet::new()));
+            if total_records.is_none() && ALREADY_WARNED.lock().unwrap().insert(step.clone()) {
+                tracing::warn!("creating mesh for {:?} with unknown record count", step);
+            }
+        }
         Mesh {
             gateway: self,
             step,
+            total_records,
         }
     }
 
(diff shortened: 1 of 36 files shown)


src/helpers/messaging.rs
```diff
@@ -121,6 +126,13 @@ impl Mesh<'_, '_> {
             )?;
         }
 
+        match self.total_records {
+            Some(count) if usize::from(record_id) >= usize::from(count) => {
```
akoshelev, line 130: can this be replaced with assert?

src/protocol/context/mod.rs
```diff
@@ -44,6 +44,11 @@ pub trait Context<F: Field>:
     #[must_use]
     fn narrow<S: Substep + ?Sized>(&self, step: &S) -> Self;
 
+    /// Sets the context's total number of records field. Communication channels are
+    /// closed based on sending the expected total number of records.
+    #[must_use]
+    fn set_total_records(&self, total_records: usize) -> Self;
```
akoshelev, line 50: I wonder if we should require number of records to be an argument for 'narrow'? for prss only steps we could pass 'None'. My motivation is that number of records seems to be tightly coupled to steps and that would allow folks to keep that in mind when calling 'narrow'.

andyleiserson: We could do that, but I think there's actually a fair amount of calls to narrow where the total record count isn't known yet or is already known.

If we distinguish contexts with a known total record count from ones without using the type system, then we can possibly specialize the implementations so that a context without a known count has a two-argument narrow and a context with a known count has a one-argument narrow.

andyleiserson:
Updated to handle upgrades of malicious protocol inputs, and updated all tests.

The record count for random_bits_generator varies depending on how many failures there are. This is still a problem. One option is to address that together with the as-yet-undetermined changes mentioned in #362 related to lock-free concurrent access to channels. (My thinking is that once we have some structure in place for maintaining ownership of channels, we implement a special mechanism to close the random_bits_generator channels.)

If we want to upgrade the record_id sanity checks to hard errors before merging, then we'll need to either solve random_bits_generator, or add a special exception to the record_id checks for random_bits_generator.

src/protocol/sort/generate_permutation.rs
```diff
@@ -265,21 +268,18 @@ pub async fn malicious_generate_permutation<'a, F>(
 where
     F: Field,
 {
-    let mut malicious_validator = MaliciousValidator::new(sh_ctx.narrow(&MaliciousUpgradeContext));
-    let mut m_ctx = malicious_validator.context();
-    let m_ctx_0 = m_ctx.narrow(&Sort(0));
+    // Review note: changed context usage for bit 0 to be analogous to context usage for the rest of the bits
```
martinthomson, line 271:
```suggestion
```
Ack.  :+1:

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -286,7 +289,7 @@ mod tests {
                     let tweaked = tweak.flip_bit(ctx.role(), triple);
 
                     let v = MaliciousValidator::new(ctx);
-                    let m_ctx = v.context();
+                    let m_ctx = v.context().set_total_records(1).set_total_upgrades(1);
```
martinthomson, line 292: Seeing this reminds me that the root malicious context held by the validator can't have a `set_total_records(self, ...)` method because we need something that modifies the context in place in that case.  I don't quite know what to do about that right now.

Is it possible to pass this information in to the validator so that it constructs that root context correctly instead?  WDYT?

@akoshelev and I continue to go back and forth about the value of the split between validator and malicious context; perhaps we need to have another look at that.

src/protocol/context/malicious.rs
```diff
@@ -256,6 +294,18 @@ impl<'a, F: Field> ContextInner<'a, F> {
         })
     }
 
+    #[must_use]
+    fn set_total_upgrades(&self, total_upgrades: usize) -> Arc<Self> {
```
martinthomson, line 298: I have a comment later on that suggests that maybe this is a constructor argument on the validator itself (which is passed to the constructor of the malicious context, saving the need for a setter).

src/protocol/boolean/solved_bits.rs
```diff
@@ -144,109 +144,82 @@ impl AsRef<str> for Step {
...
-                // Base10 of `b_B ⊆ Z` must equal `b_P`
-                assert_eq!(b_p.as_u128(), bits_to_value(&b_b));
-                success += 1;
-            }
-        }
-        // The chance of this protocol aborting 4 out of 4 tries in Fp32BitPrime
-        // is about 2^-100. Assert that at least one run has succeeded.
-        assert!(success > 0);
+    pub async fn fp31() {
+        random_bits::<Fp31, 21>().await;
```
martinthomson, line 217: Maybe comment that:

```rust
// Probability of failure for one iteration is 2^-5.
// Need 21 runs to reduce failure to < 2^-100.

File: src/protocol/boolean/solved_bits.rs [hunk 7]
```diff
@@ -144,109 +144,82 @@ impl AsRef<str> for Step {
...
+    }
 
-        Ok(())
+    #[tokio::test]
+    pub async fn fp_32bit_prime() {
+        random_bits::<Fp32BitPrime, 4>().await;
```
martinthomson, line 222: Maybe comment that:

```rust
// Probability of failure for one iteration is 5/2^32 =~ 2^-30.
// Need 4 runs to reduce failure to < 2^-100.

File: src/protocol/boolean/solved_bits.rs [hunk 8]
```diff
@@ -144,109 +144,82 @@ impl AsRef<str> for Step {
...
     use std::iter::zip;
 
-    async fn random_bits<F: Field>(
-        ctx: [SemiHonestContext<'_, F>; 3],
-        record_id: RecordId,
-    ) -> Result<Option<(Vec<F>, F)>, Error>
+    /// Execute solved_bits `COUNT` times for `F`. The count should be chosen
+    /// such that the probability of that many consecutive failures in `F` is
+    /// negligible (less than 2^-100).
+    async fn random_bits<F: Field, const COUNT: usize>()
```
martinthomson, line 159: technically, you can calculate `COUNT`:

```rust
let d = F::ONE.into() << F::BITS;
let p = f64::try_from(d - F::PRIME.into()).unwrap() / f64::try_from(d).unwrap();
let count = usize::try_from((-100_f64 / p.log2()).ceil()).unwrap();
```

src/protocol/basics/reveal.rs
```diff
@@ -166,22 +169,28 @@ mod tests {
...
             )
             .await;
 
-            let results =
-                join3v(zip(v.iter(), m_shares).map(|(v, m_share)| async move {
-                    v.context().reveal(record_id, &m_share).await
-                }))
-                .await;
+            let results = join3v(
+                zip(m_ctx.clone().into_iter(), m_shares)
```
martinthomson, line 190: Do you need the into_iter() here?  `zip` takes `impl IntoIterator`.

```suggestion
                zip(m_ctx.clone(), m_shares)
```

martinthomson (commented):
I really like this.  I've gone through in some detail, but I can't really see a problem.

I note that this doesn't *do* anything with the record count other than police it.  I take it that that is the next step.

I'd still like to take this without doing that work, so that we have the API parts (which are the most disruptive piece) in place.

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -210,19 +211,22 @@ pub(crate) mod tests {
...
         ff::{Field, Fp31},
         helpers::Role,
-        protocol::attribution::{
-            accumulate_credit::accumulate_credit,
-            tests::{BD, H, S, T},
-            AttributionInputRow,
+        protocol::{
+            attribution::{
+                accumulate_credit::accumulate_credit,
+                tests::{BD, H, S, T},
```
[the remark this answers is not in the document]
taikiy: I have a local branch that removes these. Will be sending it out soon.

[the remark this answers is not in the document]
taikiy: I have a local branch that removes these. Will be sending it out soon.

src/helpers/messaging.rs
```diff
@@ -226,11 +257,26 @@ impl Gateway {
...
+        step: &'b Step,
+        total_records: Option<NonZeroUsize>,
+    ) -> Mesh<'a, 'b> {
+        // TODO: to be changed to panic or assert once all instances are eliminated.
+        // Note: update panic docs above.
+        static ALREADY_WARNED: Lazy<Mutex<HashSet<Step>>> =
+            Lazy::new(|| Mutex::new(HashSet::new()));
+        if total_records.is_none() && ALREADY_WARNED.lock().unwrap().insert(step.clone()) {
+            tracing::warn!("creating mesh for {:?} with unknown record count", step);
+        }
```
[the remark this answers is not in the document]
andyleiserson: Yes, I can do that. I was thinking that if we eliminate all cases where there isn't a known record count, we can use a straightforward assertion and get rid of the mutex. But no reason not to restrict it to debug builds in the mean time.

src/protocol/boolean/solved_bits.rs
andyleiserson: If there's an effort to avoid unnecessary loops (to keep the test suite fast?), then maybe it makes sense to change these to run up to the specified number of iterations, stopping on the first success?

(This potentially increases test non-determinism, but I don't think in a way that matters. There is a similar case where a RandomBitsGenerator is created with a total record count -- that will fail intermittently with record_id out of range when there is a solved_bits failure. I added an assertion in `RandomBitsGenerator::new` to catch this. But the truly equivalent thing here would be doing random_bits exactly once and expecting it to succeed -- then the test would intermittently fail. I can't think of a problem with stopping on the first success.)

src/protocol/basics/check_zero.rs
```diff
@@ -107,9 +107,21 @@ mod tests {
                 counter += 1;
 
                 let protocol_output = tokio::try_join!(
-                    check_zero(context[0].narrow(&iteration), record_id, &v_shares[0],),
-                    check_zero(context[1].narrow(&iteration), record_id, &v_shares[1],),
-                    check_zero(context[2].narrow(&iteration), record_id, &v_shares[2],),
+                    check_zero(
+                        context[0].narrow(&iteration).set_total_records(1),
```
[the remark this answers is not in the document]
andyleiserson: > Separately, I wonder if it would be easier all around if TestWorld::contexts() provided total records of 1 by default.

I had a similar thought of adding a record count argument to the `world.malicious` and `world.semi_honest` helpers (which could then pass it to `contexts()`).

Any thoughts on `async fn semi_honest(&self, total_records: Option<NonZeroUsize>, input: I, helper_fn: H) -> [O; 3]` and `fn contexts(&self, total_records: Option<NonZeroUsize>)`?

src/protocol/basics/reveal.rs
```diff
@@ -139,7 +140,9 @@ mod tests {
     pub async fn simple() -> Result<(), Error> {
         let mut rng = thread_rng();
         let world = TestWorld::new().await;
-        let ctx = world.contexts::<Fp31>();
+        let ctx = world
+            .contexts::<Fp31>()
+            .map(|ctx| ctx.set_total_records(10));
 
         for i in 0..10_u32 {
```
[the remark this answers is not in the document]
andyleiserson: If I understand right, this applies to the `simple`, `malicious`, and `malicious_validation_fail` tests in this file? Rationale being they're not doing anything that running the test multiple times with `cargo test` wouldn't do?

This seems like a good idea to me.

src/protocol/modulus_conversion/convert_shares.rs
andyleiserson: I will look into this.

andyleiserson: It looks like this mostly works, but with one notable difficulty. The test `protocol::basics::mul::sparse::test::check_output_malicious` relies on narrowing the upgrade context at runtime (via `upgrade_with`). Processing the inputs as records 0/1 of a common step doesn't work, I'm assuming because the zero positions are different for the two inputs so the channel usage is not consistent.

It seems a little weird to me to fix an upgrade count at the time the validator is created, but allow dynamic narrowing of the upgrade context at runtime.

The impl of `IntoMalicious` for tuples may also be problematic. If there are cases where the two sides of the tuple differ in length and differ in upgrade mechanism, that would be a problem (but I think that case may not currently exist). And the way I see to support "either same length or same mechanism" requires behaving differently depending on which of those two cases applies, which is also a bit gross.

src/protocol/basics/check_zero.rs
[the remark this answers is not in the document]
martinthomson: Given how often we do something != 1, it might be easier to do the `contexts(&self, tr: Option<NonZeroSize>)` and `TestWorld::set_total_records()` so that we don't need a second signature for `semi_honest()`.  It's state, but only on `TestWorld`.  The others are also fine.

src/protocol/context/mod.rs
akoshelev: right, I feel that it may be actually much safer if we could prevent someone from calling `set_total_records` on contexts that already have this value set. It seems a mistake to try to set it twice unless I am missing something. 

I would be perfectly happy with debug_assert there if you folks think along the same lines. 

I also wonder what would be our strategy in mid-term for it? One may say that total records is purely an optimization that can lead to better resource utilization but otherwise program will still complete correctly. On the other hand, #373 already revealed that we have batching issues and therefore we must mandate total records/batch size to be set for every channel/step. They could be unique for every step.  Curious what are you thoughts on that

src/protocol/modulus_conversion/convert_shares.rs
andyleiserson: https://github.com/andyleiserson/ipa/commit/3aba756ecc4b50de208f62c40191cf611bcdd5c7 has my experiments

andyleiserson:
In working on the check for calling set_total_records multiple times, I've found that the number of cases where tests don't have total_records of 1 seems comparable to the number that had `set_total_records(1)` before. (These are failing after adding the check, because the test harness defaults to 1, and then the implementation sets the non-singular record count.) If that changes thoughts on what the default behavior should be, let me know.

Other than that, I need to finish cleaning up total_upgrades, and it looks like there's another merge conflict to resolve. Then I think this should be pretty close, assuming we don't do anything about random_bits_generator right now.

akoshelev:
> In working on the check for calling set_total_records multiple times, I've found that the number of cases where tests don't have total_records of 1 seems comparable to the number that had `set_total_records(1)` before. (These are failing after adding the check, because the test harness defaults to 1, and then the implementation sets the non-singular record count.) If that changes thoughts on what the default behavior should be, let me know.

it feels that it should not be `TestWorld`'s responsibility to control number of records sent through each channel? I would expect tests to set up the input size before performing MPC unless I am missing something?

andyleiserson:
> it feels that it should not be `TestWorld`'s responsibility to control number of records sent through each channel? I would expect tests to set up the input size before performing MPC unless I am missing something?

The idea as I understood it was that since many tests only operate on one record, to reduce boilerplate in the tests by having TestWorld (or something in the test fixture) set total_records = 1 by default. I agree this isn't functionality that naturally belongs to TestWorld -- the MPC algorithm is what ultimately determines the record count either as a constant of the algorithm or as an input parameter. Since it appears there's a similar number of tests that would need to specify one record and tests that specify a record count within the algorithm they're running, maybe the natural functional structure is a good "tie breaker" reason to have the default be unspecified record count rather than one record count?

akoshelev:
> Since it appears there's a similar number of tests that would need to specify one record and tests that specify a record count within the algorithm they're running, maybe the natural functional structure is a good "tie breaker" reason to have the default be unspecified record count rather than one record count?

I have the same opinion given that we have the supporting evidence for it.

src/protocol/context/malicious.rs
```diff
@@ -279,52 +257,151 @@ impl<'a, F: Field> ContextInner<'a, F> {
...
+            upgrade_ctx: self.upgrade_ctx.narrow(step),
+            inner: self,
+            record_binding: record_id,
+        }
+        .upgrade(input)
+        .await
+    }
+}
+
+/// Helper to prevent using the record ID multiple times to implement an upgrade.
```
martinthomson, line 307: I think that this code from here on down could be moved to a new file (upgrade.rs).

src/protocol/context/malicious.rs
```diff
@@ -279,52 +257,151 @@ impl<'a, F: Field> ContextInner<'a, F> {
...
+pub struct NoRecord;
+impl RecordBinding for NoRecord {}
+
+impl RecordBinding for RecordId {}
+
+pub struct UpgradeContext<'a, F: Field, B: RecordBinding = NoRecord> {
+    upgrade_ctx: SemiHonestContext<'a, F>,
+    inner: &'a ContextInner<'a, F>,
+    record_binding: B,
+}
```
martinthomson, line 340: I like this additional structure in general.

I'm not 100% on the idea of binding this way, because the idea that step N has about records might not translate directly to steps that follow on from that.  For instance, our sort effectively takes $N$ input rows and performs $N.2^B$ operations, so each input row effectively maps to more than one operation in a subsequent step.  (The only thing that might save us there is that sorting requires reveals, so we're constantly upgrading and downgrading).

martinthomson (approved):
I like this.  I need to poke at the build and merge issues, but I want to land this.  It's been too long and we need to build on this.

src/protocol/context/malicious.rs
andyleiserson: Sorry, I missed these comments before merging the upgrades PR. I agree it makes sense to move to a separate file, but I think I'll leave it alone for right now since it's no longer part of this PR.

src/protocol/context/malicious.rs
andyleiserson: For now I only applied the structure to upgrades, which is a somewhat more limited set of use cases than arbitrary protocols. I'm not sure if your comment is referring to upgrades, or to the situation where we try to apply this structure to all contexts? Even for the latter, it's not necessarily a problem. It's fine if the record count varies for different parts of the protocol as long as each of those parts does it's own record_id iteration. The thing that gets messy is if there's an implementation of `foo` that operates on a single record_id (thus takes `record_id` as an argument), that wants to call some other algorithm internally that assigns a different meaning to record IDs. Then we start seeing `a * record_id + b` calculations. I don't think those show up now -- and wouldn't that also be a problem for batching/buffering unless done very carefully?

It does seem like at some point we may want to write some sort of "guide for protocol authors" document that explains some of the non-obvious issues like this one and the need to always use a complete range of [0, total) record IDs.

andyleiserson:
Latest version:
* Rebases on main after merge of #387 and removes those commits from this PR.
* Adds a change to waive the total_records check when the total cannot be known (solved_bits), and changes the check to be an assertion failure in all other cases. This is in the last commit.
* Resolves merge conflicts. The ones that are potentially interesting to review are in sum_of_products and generate_permutation.

akoshelev:
I take the merge error as my bad. I merged changes out of order, so I am going to fix this merge conflict

PR #361, opened 2022-12-22
Start with some refactoring

martinthomson:
This makes the send buffer take an absolute index, so that it can be the place we track that value.

I was going to take a look at the backpressure thing we talked about earlier, but I realized that it will need some more significant restructuring of the gateway code.  This was the groundwork I was laying, with a goal of making the `end` field on the FSV into an `AtomicUsize` that could be shared with other threads.

I did manage to get a test failure out of this where we overran the buffer, so this general line of investigation is definitely needed.

I'll follow up with a write-up on a channel-per-step design, then call it a day.

### DIFF
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -131,7 +131,6 @@ mod tests {
     #[test]
     fn gap() {
         let mut v = FixedSizeByteVec::<ELEMENT_SIZE>::new(3);
-        println!("{:?}", v.missing());
         assert!(v.missing().is_empty());
         v.insert_test_data(1);
         assert_eq!(0..1_usize, v.missing());
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -65,6 +65,7 @@ impl<const N: usize> FixedSizeByteVec<N> {
     }
 
     /// Return any gap ahead of the first missing value.
+    #[cfg(any(test, debug_assertions))]
     pub fn missing(&self) -> Range<usize> {
         let start = self.end - self.capacity;
         let absent = self.added.leading_zeros();
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -3,29 +3,16 @@ use bitvec::prelude::BitVec;
 use std::fmt::Debug;
 use std::ops::Range;
 
-/// A vector of bytes that never grows over a certain size or shrinks below that size.
-/// Vector is segmented into regions and `max_size` is a factor of number of regions.
-/// For example, if number of regions is `2`, then `max_size` must be a multiple of two.
+/// A store of bytes that allows for random access inserts, but contiguous removal.
 ///
-/// Elements can be added to this vector in random order as long as the total number of elements
-/// does not exceed the overall capacity of the vector. An attempt to do so, will lead to a panic.
-/// Every element is stored by a specific offset and it is assumed that all elements are of the same
-/// size.
+/// Fixed-sized elements can be added to this vector with [`insert`] in random order as long as
+/// the total number of elements does not exceed the overall capacity.
 ///
-/// The layout of the vector with some elements added to it is presented below. In this example,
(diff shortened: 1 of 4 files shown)


src/helpers/buffers/fsv.rs
```diff
@@ -38,27 +41,39 @@ impl<const N: usize> FixedSizeByteVec<N> {
...
     /// Panics if `index` is out of bounds or if something was previously inserted at `index`.
-    pub fn insert(&mut self, index: usize, elem: &[u8; N]) {
-        // if index is out of bounds, this line will panic, there is no need for additional check
-        // TODO: save runtime cost with `debug_assert!()`, though that would affects panic tests
-        assert!(!self.added[index]);
+    pub fn insert<D: Debug>(&mut self, channel: D, index: usize, elem: &[u8; N]) {
+        // Translate from an absolute index into a relative one.
+        let index = index
+            .checked_sub(self.end - self.capacity)
+            .unwrap_or_else(|| panic!("Duplicate send for index {index} on channel {channel:?}"));
```
akoshelev, line 56: doesn't it indicate that index is smaller than lower bound of accepted range?

src/helpers/buffers/fsv.rs
```diff
@@ -113,28 +117,54 @@ mod tests {
...
-        v.insert_test_data(2);
-        assert!(v.added(2));
-        assert!(!v.added(1));
+        assert_eq!(v.take(1), Some(test_data_at(0).to_vec()));
+    }
 
+    #[test]
+    fn gap() {
+        let mut v = FixedSizeByteVec::<ELEMENT_SIZE>::new(3);
+        println!("{:?}", v.missing());
```
akoshelev, line 134:
```suggestion
```

src/helpers/buffers/fsv.rs
```diff
@@ -113,28 +117,54 @@ mod tests {
...
+    fn insert_taken() {
+        let mut v = FixedSizeByteVec::<ELEMENT_SIZE>::new(3);
+        v.insert_test_data(0);
         assert_eq!(v.take(1), Some(test_data_at(0).to_vec()));
+        v.insert("taken", 0, &[10; ELEMENT_SIZE]);
+    }
 
-        v.insert(1, &[10; ELEMENT_SIZE]);
+    #[test]
+    #[should_panic(expected = "index 10 out of range")]
```
akoshelev, line 158: does it pass? I can't see where this message is originated from

src/helpers/buffers/fsv.rs
martinthomson: Yes.  But index is unsigned and so it cannot be below zero.  Any value below the start of the range but non-zero therefore must have already been taken and sent out, so it's a duplicate.

src/helpers/buffers/fsv.rs
martinthomson: This message comes from the bounds checks on indexing.  It's a tiny bit fragile: we're relying on rust not changing but that seems manageable.

akoshelev (approved):

src/helpers/buffers/fsv.rs
```diff
@@ -38,27 +41,39 @@ impl<const N: usize> FixedSizeByteVec<N> {
...
+            .unwrap_or_else(|| panic!("Duplicate send for index {index} on channel {channel:?}"));
         let start = index * N;
         let offset = start..start + N;
 
-        self.added.set(index, true);
+        assert!(
+            !self.added.replace(index, true),
+            "Duplicate send for index {index} on channel {channel:?}"
+        );
         self.data[offset].copy_from_slice(elem);
```
andyleiserson, line 64: It might be worth adding a comment here either explaining what it means when an OOB access occurs here, or referring to something that does (i.e. to make clear to the reader that the panic indicates erroneous usage of the buffer, and not a bug in the buffer implementation itself).

andyleiserson (approved):
At some point we may want to implement this as a circular buffer so we don't have to repeatedly copy the data, which will involve touching some of this same code again, but that still probably doesn't make it worth doing right now.

PR #365, opened 2022-12-28
Query processor end-to-end implementation

akoshelev:
Query processor now can support pretty much the whole query workflow. There is an end to end test using in-memory transport that runs test multiply protocol, including prss negotiation and storing the results.


src/query/executor.rs
```diff
@@ -0,0 +1,162 @@
...
+            let left = F::deserialize(&chunk[..=F::SIZE_IN_BYTES as usize]).unwrap();
+            let right = F::deserialize(&chunk[F::SIZE_IN_BYTES as usize..]).unwrap();
+
+            Replicated::new(left, right)
+        });
+
+        // multiply pairs
+        let mut a = None;
+        let mut record_id = 0_u32;
+        for share in shares {
```
thurstonsand, line 56: might be more obvious flow if something like
```rust
for (record_id, shares) in shares.chunks(2).enumerate() {
    assert_eq!(shares.len(), 2);
    let share_a = shares[0];
    let share_b = shares[1];

File: src/query/executor.rs [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+        let step = Step::default().narrow(&config.query_type);
+        let prss = negotiate_prss(&gateway, &step, &mut rng).await.unwrap();
+
+        match config.field_type {
+            FieldType::Fp31 => {
+                let ctx = SemiHonestContext::<Fp31>::new(&prss, &gateway);
+                match config.query_type {
+                    #[cfg(any(test, feature = "test-fixture"))]
+                    QueryType::TestMultiply => {
+                        Box::new(test_multiply(ctx, input).await) as Box<dyn Result>
```
thurstonsand, line 104: so can the compiler not figure out how to degrade type to `dyn Result` here?

src/query/executor.rs
```diff
@@ -0,0 +1,162 @@
...
+                .flat_map(|(a, b)| {
+                    let mut slice = [0_u8; 2 * SIZE];
+                    a.serialize(&mut slice).unwrap();
+                    b.serialize(&mut slice[SIZE..]).unwrap();
+
+                    slice
+                })
+                .collect::<Vec<_>>();
+
+            Box::new(stream::iter(std::iter::once(r)))
```
thurstonsand, line 145: does `r.into_iter()` not work here?

src/query/processor.rs
```diff
@@ -108,7 +146,7 @@ impl<T: Transport + Clone> Processor<T> {
     /// * query is not registered yet
     /// * creates gateway and network
     /// * registers query
-    pub async fn prepare(&self, req: &PrepareQuery) -> Result<(), PrepareQueryError> {
+    pub async fn prepare(&self, req: PrepareQuery) -> Result<(), PrepareQueryError> {
```
thurstonsand, line 149: what's the reasoning behind having a different error type for each command?

src/query/processor.rs
```diff
@@ -122,14 +160,142 @@ impl<T: Transport + Clone> Processor<T> {
...
         Ok(())
     }
 
+    /// Receive inputs for the specified query. That triggers query processing
+    pub fn receive_inputs(&self, input: QueryInput) -> Result<(), QueryInputError> {
+        let mut queries = self.queries.inner.lock().unwrap();
+        match queries.entry(input.query_id) {
+            Entry::Occupied(entry) => {
+                let state = entry.remove();
+                match state {
```
thurstonsand, line 174:
```suggestion
                match entry.remove() {
```

src/query/processor.rs
```diff
@@ -122,14 +160,142 @@ impl<T: Transport + Clone> Processor<T> {
...
+                    QueryCommand::Prepare(req) => {
+                        self.prepare(req).await.unwrap();
+                    }
+                    QueryCommand::Input(query_input) => {
+                        self.receive_inputs(query_input).unwrap();
+                    }
+                },
+                TransportCommand::StepData(_, _, _) => panic!("unexpected command: {command:?}"),
+            }
+        }
```
thurstonsand, line 220: should this have some `else tracing::warn` or something to indicate `handle_next` was called with closed `command_stream`?

src/query/processor.rs
```diff
@@ -122,14 +160,142 @@ impl<T: Transport + Clone> Processor<T> {
...
+                        let result = self.new_query(req).await.unwrap();
+                        resp.send(result).unwrap();
+                    }
+                    QueryCommand::Prepare(req) => {
+                        self.prepare(req).await.unwrap();
+                    }
+                    QueryCommand::Input(query_input) => {
+                        self.receive_inputs(query_input).unwrap();
+                    }
+                },
```
thurstonsand, line 217: i might prefer a flatter match statement here:
```suggestion
                TransportCommand::Query(QueryCommand::Create(req, resp)) => {
                    let result = self.new_query(req).await.unwrap();
                    resp.send(result).unwrap();
                },
                TransportCommand::Query(QueryCommand::Prepare(req)) => {
                    self.prepare(req).await.unwrap();
                },
                TransportCommand::Query(QueryCommand::Input(query_input)) => {
                    self.receive_inputs(query_input).unwrap();
                },
```

src/query/processor.rs
```diff
@@ -122,14 +160,142 @@ impl<T: Transport + Clone> Processor<T> {
...
+                    };
+                    queries.insert(query_id, state);
+                    Err(QueryCompletionError::StateError {
+                        source: state_error,
+                    })
+                }
+                None => Err(QueryCompletionError::NoSuchQuery(query_id)),
+            }
+            //
+            // if let Some(QueryState::Running(handle)) = queries.remove(&query_id) {
```
thurstonsand, line 249: yeah presumably there will be more `QueryState`s, so probably can remove this

src/query/processor.rs
```diff
@@ -122,14 +160,142 @@ impl<T: Transport + Clone> Processor<T> {
...
+            //     })
+            // }
+        }?;
+
+        Ok(handle.await.unwrap())
+
+        // match queries.entry(query_id) {
+        //     Entry::Occupied(entry) => {
+        //         let state = entry.remove();
+        //         drop(queries);
```
thurstonsand, line 268: fwiw i've tried doing this kind of thing before, and rust cannot understand that `drop` means you aren't carrying the (not-`Send`) mutex guard across an `await`. i've found the only thing that works is putting braces around it. something to the effect of

```rust
let state: Option<QueryState> = {
    let mut queries = self.queries.inner.lock().unwrap();
    queries.remove(query_id)
};
```
maybe moot point if you aren't going to use this code, but yeah

src/query/processor.rs
```diff
@@ -267,12 +436,96 @@ mod tests {
...
+        use crate::sync::Weak;
+        use crate::test_fixture::transport::InMemoryTransport;
+        use crate::test_fixture::Reconstruct;
+        use futures_util::future::join_all;
+        use futures_util::stream;
+        use tokio::sync::oneshot;
+
+        async fn make_three(network: &InMemoryNetwork) -> [Processor<Weak<InMemoryTransport>>; 3] {
+            let identities = HelperIdentity::make_three();
+            join_all(network.transports.iter().map(|transport| async {
```
thurstonsand, line 462: just to help my rust knowledge, why didn't you need `async move` here?

src/test_fixture/transport/routing.rs
```diff
@@ -102,30 +96,23 @@ impl Switch {
         }
 
         let handle = tokio::spawn(async move {
-            let mut query_router = QueryCommandRouter::default();
+            let mut routes = HashMap::default();
             loop {
                 ::tokio::select! {
                     Some(command) = rx.recv() => {
                         match command {
                             SwitchCommand::Subscribe(subscribe_command) => {
```
thurstonsand, line 104: could you not do
```suggestion
                            SwitchCommand::Subscribe(SubscribeRequest { subscription, link, ack_tx }) => {
```

src/test_fixture/transport/routing.rs
```diff
@@ -165,28 +159,51 @@ impl Drop for Switch {
...
+            command: value.0,
+            inner: "channel closed".into(),
+        }
     }
+}
 
-    fn subscribe(&mut self, query_id: QueryId, sender: mpsc::Sender<CommandEnvelope>) {
-        assert!(self.routes.insert(query_id, sender).is_none());
+#[async_trait]
+trait Dispatcher {
```
thurstonsand, line 183: just curious -- what alternative implementers do you see for `Dispatcher`?

src/test_fixture/world.rs
```diff
@@ -147,8 +147,8 @@ impl TestWorld {
     pub fn contexts<F: Field>(&self) -> [SemiHonestContext<'_, F>; 3] {
         let execution = self.executions.fetch_add(1, Ordering::Release);
         zip(Role::all(), zip(&self.participants, &*self.gateways))
-            .map(|(role, (participant, gateway))| {
-                SemiHonestContext::new(*role, participant, gateway)
+            .map(|(_role, (participant, gateway))| {
```
thurstonsand, line 150: can you just remove `Role::all()`?

thurstonsand (changes requested):

src/query/executor.rs
```diff
@@ -0,0 +1,162 @@
...
+use shuttle::future as tokio;
+
+pub trait Result: Send {
+    fn into_bytes(self: Box<Self>) -> Vec<u8>;
+}
+
+impl<F: Field> Result for Vec<Replicated<F>> {
+    fn into_bytes(self: Box<Self>) -> Vec<u8> {
+        // todo Result
+        let mut r = Vec::with_capacity(2 * self.len() * F::SIZE_IN_BYTES as usize);
```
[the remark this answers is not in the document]
akoshelev: yea I actually added that const already

src/query/executor.rs
```diff
@@ -0,0 +1,162 @@
...
+#[cfg(all(feature = "shuttle", test))]
+use shuttle::future as tokio;
+
+pub trait Result: Send {
+    fn into_bytes(self: Box<Self>) -> Vec<u8>;
+}
+
+impl<F: Field> Result for Vec<Replicated<F>> {
+    fn into_bytes(self: Box<Self>) -> Vec<u8> {
+        // todo Result
```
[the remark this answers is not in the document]
akoshelev: I think I need to change the return type to Result

src/query/executor.rs
```diff
@@ -0,0 +1,162 @@
...
+        let mut r = Vec::with_capacity(2 * self.len() * F::SIZE_IN_BYTES as usize);
+        for share in self.into_iter() {
+            let mut slice = [0u8; 32]; // we don't support fields > 16 bytes
+            let written = share.serialize(&mut slice).unwrap();
```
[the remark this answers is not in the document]
akoshelev: I cant try that - this code is not supposed to be running in prod, so it is not perf critical

src/query/executor.rs
akoshelev: yep it does not auto coerce types into dyn afaik

src/query/executor.rs
akoshelev: maybe it works, but I think it is the same

src/query/processor.rs
akoshelev: no actual reason, I added commands one by one. for now it works, once I have the full implementation of query processor, some errors types can be merged

src/query/processor.rs
akoshelev: true, I had to do the same trick

src/test_fixture/transport/routing.rs
akoshelev: likely only one implementation. I'll see if I can remove it

src/query/executor.rs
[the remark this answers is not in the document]
thurstonsand: so should you use that here?

src/query/executor.rs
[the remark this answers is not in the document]
thurstonsand: sure, not a big deal

src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub mod test_fixture;
 
 mod tests;
 
+extern crate core;
```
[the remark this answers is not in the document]
akoshelev: my IDE keeps adding it for some reason - will remove

src/query/executor.rs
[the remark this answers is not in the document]
akoshelev: yep, changing it

src/query/executor.rs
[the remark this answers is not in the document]
akoshelev: changed my mind - there can be only software bugs (or alloc errors) that can prevent protocol results from getting serialized into byte arr. so panic is the right choice here imo

src/query/executor.rs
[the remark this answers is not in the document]
akoshelev: that works just fine so I would change it the way you suggested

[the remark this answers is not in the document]
akoshelev: ~~this code is not supposed to be running in prod~~ not true

src/query/executor.rs
akoshelev: I don't think I can make it work because `chunks` need a container to store data and it is not implemented for iterators, only for slices.

src/query/processor.rs
akoshelev: there is actually a different issue - state needs to be inserted back if there is an error. so I would need `state` variable, but need to change `_` branch

src/query/processor.rs
akoshelev: I don't mind it either

src/query/processor.rs
akoshelev: I think polling a terminated stream would cause a panic https://docs.rs/futures/latest/futures/stream/trait.Stream.html#panics

src/query/processor.rs
akoshelev: yikes, bunch of garbage in the PR. will clean that up

src/test_fixture/transport/routing.rs
akoshelev: lgtm

src/test_fixture/transport/routing.rs
akoshelev: I don't think I can because it is an extension trait, adding behavior to foreign type

src/test_fixture/world.rs
akoshelev: yep I should

src/query/processor.rs
akoshelev: there is no requirement for things to be `'static` compared to tokio::spawn, so I can just pass a reference to identities and clone it inside the closure. My understanding is that after compiler expands the state machine for this future, it sees that future goes out of scope before `identities` and it allows this code

src/query/executor.rs
thurstonsand: ah...yeah i understand. i was expecting it to be able to coerce into `Vec`

src/query/processor.rs
thurstonsand: right. at the least, can you add a `# Panics` comment so that's more obvious?

src/query/processor.rs
thurstonsand: ok, so it's the order that things go out of scope, got it. lifetime management can be really subtle sometimes...

src/test_fixture/transport/routing.rs
thurstonsand: oh, yeah that makes sense

thurstonsand (approved):
looking good

src/query/processor.rs
akoshelev: done

PR #366, opened 2022-12-29
Query processor IPA protocol end-to-end implementation

akoshelev:
A bunch of boilerplate code had to be written to serialize and deserialize structs. that's quite annoying, need to think about a better way how to deal with it.


src/protocol/attribution/mod.rs
```diff
@@ -40,6 +41,52 @@ pub struct AggregateCreditOutputRow<F: Field> {
     credit: Replicated<F>,
 }
 
+impl<F: Field> AggregateCreditOutputRow<F> {
+    // TODO: this gets repetitive. For the most of our structures we need to be able to serialize
```
thurstonsand, line 45: yeah this kinda feels like you want to recreate the `serde::Serialize/Deserialize` macros...

src/protocol/ipa/mod.rs
```diff
@@ -75,13 +76,65 @@ impl AsRef<str> for IPAInputRowResharableStep {
...
 
+impl<F: Field> IPAInputRow<F> {
+    pub const SIZE_IN_BYTES: usize = std::mem::size_of::<Self>();
+
+    /// Splits the given slice into chunks aligned with the size of this struct and returns an
+    /// iterator that produces deserialized instances.
+    ///
+    /// ## Panics
+    /// Panics if the slice buffer is not aligned with the size of this struct.
+    pub fn from_byte_slice(input: &[u8]) -> impl Iterator<Item = Self> + '_ {
```
thurstonsand, line 95: no `deserialize` method here?

src/query/executor.rs
```diff
@@ -66,11 +91,25 @@ async fn test_multiply<F: Field, S: Stream<Item = Vec<u8>> + Send + Unpin>(
 }
 
 #[allow(clippy::unused_async)]
```
thurstonsand, line 93: remove

thurstonsand (approved):
yep, mostly just serde and hooking up the tests

src/protocol/ipa/mod.rs
akoshelev: for now no, because we only deserialize many of them

PR #367, opened 2022-12-29
implement external client api

thurstonsand:


src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub mod test_fixture;
 
 mod tests;
 
+extern crate core;
```
thurstonsand: nice

akoshelev: it is just killing me

PR #368, opened 2022-12-29
tweak framework for malicious tests

andyleiserson:


src/test_fixture/world.rs
```diff
@@ -245,72 +252,35 @@ where
...
     }
 
-    async fn malicious<'a, O, M, H, R, P>(&'a self, input: I, mut helper_fn: H) -> [O; 3]
+    async fn malicious<'a, O, M, H, R, P>(&'a self, input: I, helper_fn: H) -> [O; 3]
     where
         A: IntoMalicious<F, M>,
         O: Send + Debug,
         M: Send,
-        H: FnMut(MaliciousContext<'a, F>, M) -> R + Send,
+        H: Fn(MaliciousContext<'a, F>, M) -> R + Send + Sync,
```
akoshelev, line 260: I may be wrong, but I think @martinthomson tried hard to keep it `Send` only

akoshelev (approved):
I don't mind those new constraints, but would suggest to let @martinthomson have a look as well

src/test_fixture/world.rs
andyleiserson: I couldn't come up with a use case where it would be a problem for this to be Sync. In most cases the closure doesn't have any state at all, i.e., `helper_fn` could have type `fn(MaliciousContext<'a, F>, M) -> R`. There are a few tests in reshare.rs that do `for &role in Role::all()` and capture `role` in the closure. That case, or anything where the closure captures some static data, is still Sync.

martinthomson: Yeah, I'm happy to see this code cleaned up like this.  The extra constraints don't seem to affect the code we have, so that's pretty strong evidence that we don't need the extra flexibility that you get from `FnMut` or non-`Sync` functions.  Same goes for `P` being `!Clone`.

src/test_fixture/world.rs
```diff
@@ -245,72 +252,35 @@ where
...
+        let (m_results, r_shares, output) = split_array_of_tuples(
             self.semi_honest(input, |ctx, share| async {
                 let v = MaliciousValidator::new(ctx);
                 let m_share = share.upgrade(v.context()).await;
-                let res = helper_fn(v.context(), m_share).await;
-                v.validate(res).await.unwrap()
+                let m_result = helper_fn(v.context(), m_share).await;
+                let m_result_clone = m_result.clone();
+                let r_share = v.r_share().clone();
+                let output = v.validate(m_result_clone).await.unwrap();
```
martinthomson, line 273:
```suggestion
                let r_share = v.r_share().clone();
                let output = v.validate(m_result.clone()).await.unwrap();
```

martinthomson (approved):

src/test_fixture/world.rs
martinthomson: ... and now I see why you needed to do this explicitly.  That `await` really messes with things.

queenbeastmode89-hub (commented):
7f482a68a3c5bade1d5674d70d79e53b784e7822

PR #369, opened 2023-01-03
Make the fixed sized buffer fixed

martinthomson:
... and circular

This avoids some of the churn in this object, though it doesn't avoid the part where it returns a vector rather than a slice.  That might be something we need to do later.  As we're sending the data, having it be owned isn't strictly bad, but we might see some benefits from using slices once we get deeper into optimization.

### DIFF
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -131,7 +131,6 @@ mod tests {
     #[test]
     fn gap() {
         let mut v = FixedSizeByteVec::<ELEMENT_SIZE>::new(3);
-        println!("{:?}", v.missing());
         assert!(v.missing().is_empty());
         v.insert_test_data(1);
         assert_eq!(0..1_usize, v.missing());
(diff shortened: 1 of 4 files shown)


martinthomson:
Note: this builds on #361.  One commit different only.

src/helpers/buffers/fsv.rs
```diff
@@ -38,62 +28,93 @@ impl<const N: usize> FixedSizeByteVec<N> {
...
+            buf.extend_from_slice(&self.data[..(wrap * N)]);
+            Some(buf)
+        } else {
+            Some(self.data[(i * N)..((i + tail) * N)].to_owned())
+        }
     }
 }
 
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
```
akoshelev, line 117: I would feel a bit more confident if we added a randomized test that uses capacity values picked arbitrarily and tests that take/insert work correctly. bugs in circular buffers are often subtle. Having said that, I don't see anything wrong with this implementation

src/helpers/buffers/fsv.rs
```diff
@@ -38,62 +28,93 @@ impl<const N: usize> FixedSizeByteVec<N> {
         Self {
             data: vec![0_u8; N * capacity],
             added: bitvec![0; capacity],
-            taken: 0,
+            capacity,
```
akoshelev, line 31: maybe it is a good time to reject `capacity = 0`? does not have to be in this change though

akoshelev (approved):

PR #370, opened 2023-01-04
`SharedValue` trait for secret shared value types

taikiy:
## Overview
Another refactoring towards supporting XOR replicated shares (see the diagram below for the overall design). This diff adds a new intermediate trait for secret shared values i.e., fields, bit arrays, that helps abstract away the implementations of the secrets from contexts and context dependent protocols such as `SecureMul` (red box in the diagram).

For now, I bounded the context to `Context<V: SharedValue + ArithmeticOps>`, basically equivalent to `F: Field` leaving the current behavior intact.

## Design
Untitled-2022-12-08-1530

### DIFF
--- a/src/bits/mod.rs
+++ b/src/bits/mod.rs
@@ -1,4 +1,4 @@
-use crate::secret_sharing::SharedValue;
+use crate::secret_sharing::BooleanShare;
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Index, Not};
 
 mod bit_array;
@@ -7,14 +7,7 @@ pub use bit_array::BitArray64;
 
 /// Trait for data types storing arbitrary number of bits.
 // TODO: Implement `Message`
-pub trait BitArray: SharedValue + BooleanOps + TryFrom<u128> + Index<usize> {
-    /// Size of this data type in bytes. This is the size in memory allocated
-    /// for this data type to store the number of bits specified by `BITS`.
-    /// `SIZE_IN_BYTES * 8` could be larger than `BITS`, but this type will
-    /// store exactly `BITS` number of bits.
-    const SIZE_IN_BYTES: usize;
-    /// Number of bits stored in this data type.
-    const BITS: u32;
+pub trait BitArray: BooleanShare + TryFrom<u128> + Index<usize> {
     /// A bit array with all its elements initialized to 0.
     const ZERO: Self;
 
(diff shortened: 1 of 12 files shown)


src/bits/mod.rs
```diff
@@ -7,19 +7,7 @@ pub use bit_array::BitArray64;
...
-    + Debug
-    + Send
-    + Sync
-    + Sized
-    + 'static
-{
+pub trait BitArray: SharedValue + BooleanOps + TryFrom<u128> + Index<usize> {
     /// Size of this data type in bytes. This is the size in memory allocated
     /// for this data type to store the number of bits specified by `BITS`.
     /// `SIZE_IN_BYTES * 8` could be larger than `BITS`, but this type will
```
martinthomson, line 13: I wonder whether `BITS` and `SIZE_IN_BYTES` belong on `SharedValue`.  WDYT?

`Field` also has the same constants (more or less).

src/protocol/basics/reveal.rs
```diff
@@ -1,24 +1,24 @@
...
+use crate::secret_sharing::{MaliciousReplicated, Replicated, SecretSharing, SharedValue};
 use crate::{error::Error, helpers::Direction, protocol::RecordId};
 use async_trait::async_trait;
 use embed_doc_image::embed_doc_image;
 use futures::future::{try_join, try_join_all};
 
 /// Trait for reveal protocol to open a shared secret to all helpers inside the MPC ring.
 #[async_trait]
-pub trait Reveal<F: Field> {
+pub trait Reveal<V: SharedValue + ArithmeticOps> {
```
martinthomson, line 13: I'm seeing a few of these, which is fine, but I wonder if we might define a simple alias trait like `trait ArithmeticShare: SharedValue + ArithmeticOps`.

src/ff/field.rs
```diff
@@ -102,14 +92,4 @@ pub trait Field:
...
-    + BitAnd<Output = Self>
-    + BitAndAssign
-    + BitOr<Output = Self>
-    + BitOrAssign
-    + BitXor<Output = Self>
-    + BitXorAssign
-    + Not<Output = Self>
-{
-}
+pub trait BinaryField: Field + BooleanOps {}
```
martinthomson, line 95: I just searched and we don't use `BinaryField` or `Fp2` any more.  Can these be removed?

I started by trying to think why we would want something like this and came up short.  The `BitArray`-based version works better for every case I could think of.

src/bits/mod.rs
taikiy: Sounds great. I moved these two constants to `SharedValue`.

src/ff/field.rs
taikiy: Removed Fp2.

martinthomson (approved):

PR #371, opened 2023-01-04
[Sort] Multi Gen Bit Perm : Computing permutation for multiple bits at a time

richajaindce:


src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,174 @@
...
+    ctx: C,
+    input: &[Vec<S>],
+) -> Result<Vec<S>, Error> {
+    let share_of_one = ctx.share_of_one();
+    let num_multi_bits = input.len();
+    let num_recs = input[0].len();
+    let twonminusone = 2 << (num_multi_bits - 1);
+
+    let mut mult_outputs = Vec::with_capacity(twonminusone);
+    for j in 0..twonminusone {
```
richajaindce, line 48: All this can definitely be written in a more compact way. Need some help to make it more concise.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,175 @@
...
+/// by K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, N. Kiribuchi, and B. Pinkas
+/// <https://eprint.iacr.org/2019/695.pdf>.
+///
+/// Protocol to compute a secret sharing of a permutation, after sorting on multiple bits `num_multi_bits`.
+/// At a high level, the protocol works as follows:
+/// 1. Start with a vector of list of `L*n` secret shares `[[x1_1 ... x1_n], .. , [xL_1 ... xL_n]]` where each is a secret sharing of either zero or one.
+/// 2. For j in 0 to 2 pow `num_multi_bits`
+///    i. Get binary representation of j (Bk)
+///    ii. For i in `num_multi_bits`
+///      a. Locally compute `mult_inputs` as (Bk * k + (1-Bk)(1- k))
```
benjaminsavage, line 26: Reading this, I lose you here. Not sure what "k" is, or how it relates to Bk.

richajaindce: Tried to make it better.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,175 @@
...
+            j_mult_output.push(
+                multiply_all_shares(
+                    ctx.narrow(&MultiplyAcrossBits),
+                    RecordId::from(j * num_recs + rec),
+                    mult_inputs_all.as_slice(),
+                )
+                .await?,
+            );
+        }
+        mult_outputs.push(j_mult_output);
```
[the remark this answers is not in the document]
richajaindce: Tried to incorporate this

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+                mult_input.as_slice(),
+            )
+            .await
+        });
+    }
+    try_join_all(equality_check_futures).await
+}
+
+/// Get binary representation of an integer as a vector of bool
+fn get_binary_from_int(input_num: usize, num_bits: usize) -> Vec<bool> {
```
akoshelev, line 142: I don't think you need a vector here. or an array. you just need to loop over all bits of an integer, returning `impl Iterator<Item = bool>` is enough. The good news is probably you don't even need to manually implement it, because `bitvec` crate has it. So it should be something similar to 

```rust
   let bit_iter = idx.view_bits::<Lsb0>().iter().take(..) // not sure if usize is easy to deal with here but we have a limit of 1B records so downcasting idx to u32 should be fine
```


in general, try to avoid using vectors unless it is absolutely necessary. Vectors are relatively expensive because they need to heap allocate in order to keep the elements. If you know that the number of items you need to keep is small, prefer tinyvec arrays.

akoshelev: to make it a bit more clear - you can get rid of this function and replace it with bitvec `BitView` `view_bits` method.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+            // multiply all mult_input for this j for each record => f(j)
+            multiply_all_shares(
+                ctx_across_bits,
+                RecordId::from(idx * num_records + rec),
+                mult_input.as_slice(),
+            )
+            .await
+        });
+    }
+    try_join_all(equality_check_futures).await
```
akoshelev, line 138: `try_join_all` takes an iterator, you probably don't need to keep futures inside a vector. `try_join_all(num_records.map(..)).await?` will likely be more efficient

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+                .await
+        });
+    }
+    try_join_all(permutation_futures).await
+}
+
+/// For a given `idx` check if each of the record has same value as idx.
+/// Steps
+/// 1. Get bit representation of `idx`
+/// 2. Calculate equality check
```
akoshelev, line 91: I am not sure I follow what the term`equality` means in this context.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+    for rec in 0..num_records {
+        let ctx_across_bits = ctx.narrow(&MultiplyAcrossBits);
+        let share_of_one = ctx_across_bits.share_of_one();
+
+        let mut bits_for_record = Vec::with_capacity(num_multi_bits);
+        for jth_bits in input.iter().take(num_multi_bits) {
+            bits_for_record.push(jth_bits[rec].clone());
+        }
+        let mut mult_input = Vec::with_capacity(num_multi_bits);
+        for jth_bit in j_bit_representation.iter().take(num_multi_bits) {
```
akoshelev, line 121: I feel like you should be able to zip `input` and `j_bits` like this

```rust
   input.iter().zip(j_bits).map(|(record, jth_bit)| { if jth_bit { record[rec] } else { -record[rec] + share_of_one } })
```

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+            .push(async move { get_bit_equality_checkers(j, input, ctx_equality_checker).await });
+    }
+
+    let equality_checks = try_join_all(equality_check_futures).await?;
+
+    // Compute accumulated sum
+    let mut sop_inputs_transposed = Vec::with_capacity(num_possible_bit_values * num_records);
+    let mut initial_state = S::ZERO;
+    for equality_check in equality_checks {
+        let out = equality_check
```
akoshelev, line 56: what are you trying to do here? there might be a cleaner way to do this

richajaindce: I am trying to compute a cumulative sum across records and bits as in this explainer: https://docs.google.com/document/d/1DbMVkBf2n4LkdHKg4Q5pHFxbvCSA8oR63sx16qRPfp0/edit#bookmark=id.3z1s7zmkwpwg

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Thanks for the suggestion! This is much better.

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -60,36 +60,36 @@ impl AsRef<str> for Step {
 pub async fn sum_of_products<F>(
     ctx: MaliciousContext<'_, F>,
     record_id: RecordId,
-    a: &[&MaliciousReplicated<F>],
-    b: &[&MaliciousReplicated<F>],
+    pairs: &[(&MaliciousReplicated<F>, &MaliciousReplicated<F>)],
```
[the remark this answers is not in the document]
richajaindce: I thought this format might be more relatable since sum of products would be done on each pair. How does this affect the efficiency? (I am changing it in the code but trying to understand the efficiency consideration here)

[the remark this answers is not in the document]
akoshelev: my assumption is that `a` and `b` are separate contiguous regions in memory (as previous implementation assumed). You somehow accumulated `a` values and `b` values before doing sum of products (maybe from two different sub-protocols). Once you hold onto `a` and `b` regions, you can just pass references to `a` and `b` to `sum_of_products`. Your implementation takes a slice of pairs, so clients that hold `a` and `b` will have to construct this new region (vector) of pairs by copying references to values inside `a` and `b`. That would require another allocation and therefore less efficient.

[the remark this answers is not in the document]
martinthomson: You could achieve similar semantics without copying if you took an argument that was `impl IntoIterator<Item = (&M, &M)>`, but I don't know if that is any better.  If these really need to be held in a random access store in order to apply permutations, then there is no value in that change.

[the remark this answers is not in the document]
benjaminsavage: Looking down below at how this is actually used - I don't think there's any difference because we *aren't* dealing with consecutive regions of memory. We are creating new vectors either way. So in practice I don't think there's a difference between creating two new vectors vs creating one vector of pairs.

What I like about this approach is that you have a compile-time guarantee that the vectors have the same length.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+        let num_multi_bits = INPUT.len();
+        let num_records = INPUT[0].len();
+        let num_possible_bit_values = 2 << (num_multi_bits - 1);
+
+        let result = world
+            .semi_honest(input, |ctx, m_shares| async move {
+                let mut equality_check_futures =
+                    Vec::with_capacity(num_possible_bit_values * num_records);
+                for j in 0..num_possible_bit_values {
+                    let ctx_equality_checker = ctx.narrow(&EqualityBitChecker);
```
benjaminsavage, line 215: I don't think this narrow call is necessary, because within `get_bit_equality_checkers` you're immediately calling narrow again:
`let ctx_across_bits = ctx.narrow(&MultiplyAcrossBits);`
So as far as I can tell, you never actually use this `ctx_equality_checker` context on its own.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+    input: &[Vec<S>],
+) -> Result<Vec<S>, Error> {
+    let num_multi_bits = input.len();
+    let num_records = input[0].len();
+    let num_possible_bit_values = 2 << (num_multi_bits - 1);
+
+    // Equality bit checker: this checks if each secret shared record is equal to any of numbers between 0 and num_possible_bit_values
+    let mut equality_check_futures = Vec::with_capacity(num_possible_bit_values * num_records);
+    for j in 0..num_possible_bit_values {
+        let ctx_equality_checker = ctx.narrow(&EqualityBitChecker);
```
benjaminsavage, line 45: I don't think this narrow call is necessary, because within get_bit_equality_checkers you're immediately calling narrow again:
let ctx_across_bits = ctx.narrow(&MultiplyAcrossBits);
So as far as I can tell, you never actually use this ctx_equality_checker context on its own.

src/protocol/sort/multi_bit_permutation.rs
benjaminsavage: Yeah, I think there must be a simpler way of writing this. My gut instinct would be to just write a nested for loop:
```
let mut sop_inputs_transposed = Vec::with_capacity(num_possible_bit_values * num_records);
let mut cumulative_sum = S::ZERO;
for equality_check in equality_checks {
    for check in equality_check {
        cumulative_sum += check;
        sop_inputs_transposed.push(cumulative_sum);
    }
}
```

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+            .unwrap_or_else(|| panic!("We should never reach here"))
+            .1
+            .clone();
+        sop_inputs_transposed.push(out);
+    }
+
+    // Take sum of products of output of equality check and accumulated sum
+    let mut permutation_futures = Vec::with_capacity(num_records);
+    for rec in 0..num_records {
+        let ctx_sop = ctx.narrow(&Sop);
```
benjaminsavage, line 74: Is this narrow call necessary at all?
If it is, why does it need to be called from within the loop body? Can it be moved outside?

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+        let mut sop_inputs = Vec::with_capacity(num_possible_bit_values);
+        for sop_input_transposed in sop_inputs_transposed.iter().take(num_possible_bit_values) {
+            sop_inputs.push((&sop_input_transposed[rec].0, &sop_input_transposed[rec].1));
+        }
```
benjaminsavage, line 78: It seems that in the current implementation `sop_inputs_transposed` contains a copy of the original "equality_checks", which are subsequently copied again here. That seems wasteful. I'd recommend not generating that intermediate copy. You can still construct this vector of pairs here from the original `equality_checks` and the cumulative sums.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context<F, Share = S>>(
+    ctx: C,
+    input: &[Vec<S>],
+) -> Result<Vec<S>, Error> {
+    let num_multi_bits = input.len();
+    let num_records = input[0].len();
+    let num_possible_bit_values = 2 << (num_multi_bits - 1);
+
+    // Equality bit checker: this checks if each secret shared record is equal to any of numbers between 0 and num_possible_bit_values
+    let mut equality_check_futures = Vec::with_capacity(num_possible_bit_values * num_records);
```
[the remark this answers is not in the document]
richajaindce: This is correct. It got missed in the changes done from last review. Thanks for spotting this

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Since we have an async move just after this, it expects to receive a context which can be consumed by the async block. Changing it to clone the context instead (which is what others are also doing in the code)

src/protocol/sort/multi_bit_permutation.rs
richajaindce: I started with this approach and then wanted to do something similar to what we did in bit_permutation. For now, lets keep it simple and use what u proposed Ben.

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Sounds good.. made this change

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,206 @@
...
+    F: Field,
+    C: Context<F, Share = S>,
+    S: SecretSharing<F>,
+{
+    let num_multi_bits = multi_bit_input.len();
+    let num_records = multi_bit_input[0].len();
+
+    let ctx_across_bits = ctx.narrow(&MultiplyAcrossBits);
+
+    let idx_in_bits = idx.view_bits::<Lsb0>()[..num_multi_bits].iter().rev();
```
akoshelev, line 103: you can use MSB order and that should eliminate the need for `rev`

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,206 @@
...
+    let equality_check_futures = (0..num_records).map(|rec| {
+        let ctx_across_bits = ctx_across_bits.clone();
+        let share_of_one = ctx_across_bits.share_of_one();
+
+        let mult_input = multi_bit_input
+            .iter()
+            .zip(idx_in_bits.clone())
```
akoshelev, line 110: that clones more than we probably need. `idx_in_bits` should be created once and put on the stack. On each iteration we can just borrow it

src/protocol/sort/multi_bit_permutation.rs
akoshelev: hmm, it is actually pretty much the same thing, so I don't think it is worth doing: `view_bits::<Msb0>.iter().skip(idx.size() - num_multi_bits)`

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,206 @@
...
+            .map(|(single_bit_input, bit)| {
+                if *bit {
+                    single_bit_input[rec].clone()
+                } else {
+                    -single_bit_input[rec].clone() + &share_of_one
+                }
+            })
+            .collect::<Vec<_>>();
+
+        async move {
```
akoshelev, line 120: you can lift it up to the closure definition to save on indents 

```rust
    let equality_check_futures = (0..num_records).map(|rec| async move { ...
```

src/protocol/basics/sum_of_product/malicious.rs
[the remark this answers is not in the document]
akoshelev: I looked at the usages (I was only able to find usages in tests, so could be missing something) and it seems that `a` and `b` are coming as separate allocations most of the time?

for example

```rust
let res = world
            .semi_honest((av, bv), |ctx, (a, b)| async move {
                let mut pairs = Vec::with_capacity(MULTI_BIT_LEN);
                for i in 0..a.len() {
                    pairs.push((&a[i], &b[i]));
                }
                ctx.sum_of_products(RecordId::from(0), pairs.as_slice())
                    .await
                    .unwrap()
            })
```

with previous implementation can probably be written without another vector

```rust
let res = world
            .semi_honest((av, bv), |ctx, (a, b)| async move {
                ctx.sum_of_products(RecordId::from(0), &a, &b)
                    .await
                    .unwrap()
            })
```

I agree that ergonomic of using `&[(a, b)]` is better, but efficiency loss may be significant if we need to allocate another large vector just to combine them. If we are truly assembling the `a` and `b` vectors at the same time (within the same protocol), then I am in favor of this change

src/protocol/sort/multi_bit_permutation.rs
richajaindce: agreed.. they both looked similar and this looked a bit cleaner (since we dont do calculation of how much to skip)

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Replaced this with repeat which looks cleaner

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -96,36 +98,37 @@ where
...
-        async move {
+    let equality_check_futures = (0..num_records)
+        .zip(repeat(ctx_across_bits))
+        .zip(repeat(idx_in_bits))
+        .map(|((rec, ctx), idx_in_bits)| async move {
+            let share_of_one = ctx.share_of_one();
+
+            let mult_input = multi_bit_input
+                .iter()
+                .zip(idx_in_bits)
```
martinthomson, line 115: This might read better as `let mult_input = zip(&multi_bit_input, idx_in_bits).map(...)`

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -96,36 +98,37 @@ where
...
+                        -single_bit_input[rec].clone() + &share_of_one
+                    }
+                })
+                .collect::<Vec<_>>();
+
             multiply_all_shares(
-                ctx_across_bits,
+                ctx,
                 RecordId::from(idx * num_records + rec),
                 mult_input.as_slice(),
```
martinthomson, line 128: This is a case where you might be better off having `multiply_all_shares()` take `impl IntoIterator<Item = ...>` rather than a slice.  Then you might not need to `.collect()` into `mult_input`.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,256 @@
...
+
+/// This is an implementation of `GenMultiBitSort` (Algorithm 11) described in:
+/// "An Efficient Secure Three-Party Sorting Protocol with an Honest Majority"
+/// by K. Chida, K. Hamada, D. Ikarashi, R. Kikuchi, N. Kiribuchi, and B. Pinkas
+/// <https://eprint.iacr.org/2019/695.pdf>.
+///
+/// Protocol to compute a secret sharing of a permutation, after sorting on multiple bits `num_multi_bits`.
+/// At a high level, the protocol works as follows:
+/// 1. Start with a vector of list of `L*n` secret shares `[[x1_1 ... x1_n], .. , [xL_1 ... xL_n]]` where each is a secret sharing of either zero or one.
+///    Here, L is the number of multi bits which are procrssed together (`num_multi_bits`) and n is the number of records
```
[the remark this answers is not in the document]
benjaminsavage: While this is marked as resolved - it's not showing an update.

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Inside the multiply_all_shares code, we access random locations (2*i and 2*i+1) for each i. Then, we would anyways need to collect inside that function. It is just a matter of doing outside or inside. I think doing outside makes it clear to the caller that a vector is expected. Let me know your thoughts

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,209 @@
...
+    let mut equality_check_futures = Vec::with_capacity(num_possible_bit_values);
+    for j in 0..num_possible_bit_values {
+        let ctx = ctx.clone();
+        equality_check_futures.push(async move { get_bit_equality_checkers(j, input, ctx).await });
+    }
+
+    let equality_checks = try_join_all(equality_check_futures).await?;
+
+    // Compute accumulated sum
+    let mut sop_inputs_transposed = Vec::with_capacity(num_possible_bit_values * num_records);
```
benjaminsavage, line 55: I don't love this variable name. It's not really a transposition... it's a flattenning? I'd rather just call it "prefix_sum".

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,209 @@
...
+        let ctx = ctx.clone();
+        let mut sop_inputs = Vec::with_capacity(num_possible_bit_values);
+        for idx in 0..num_possible_bit_values {
+            sop_inputs.push((
+                &equality_checks[idx][rec],
+                &sop_inputs_transposed[idx * num_records + rec],
+            ));
+        }
+        permutation_futures.push(async move {
+            ctx.sum_of_products(RecordId::from(rec), sop_inputs.as_slice())
```
benjaminsavage, line 76: This seems like it will require a call to `narrow` once we try to use this protocol within something else. Right now you're counting on the context that is passed it to be "clean" or "unused". I think we should probably just narrow before doing this.

richajaindce: I thought this is the expectation from the caller that they will handover a clean context for subprotocols to use. All the sort subprotocols get a narrowed context from their callers.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -0,0 +1,207 @@
...
+        let ctx = ctx.clone();
+        let mut sop_inputs = Vec::with_capacity(num_possible_bit_values);
+        for idx in 0..num_possible_bit_values {
+            sop_inputs.push((
+                &equality_checks[idx][rec],
+                &prefix_sum[idx * num_records + rec],
+            ));
+        }
+        permutation_futures.push(async move {
+            ctx.sum_of_products(RecordId::from(rec), sop_inputs.as_slice())
```
benjaminsavage, line 76: Right now we are:

1. Making a new copy for `N*M` cumulative sums.
2. Copying all of these AND also `N*M` elements from equality_checks into short vectors of pairs.
I think we can skip step 1 by just writing the cumulative sums directly into the vectors of pairs.

Here is a sketch (that prolly won't compile)

```
let mut permutation_inputs = Vec::with_capacity(num_records);
for (j, equality_check) in equality_checks.iter().enumerate() {
    for (i, check) in equality_check.iter().enumerate() {
        cumulative_sum += check;
        permutation_inputs.push((
            &equality_checks[j][i],
            cumulative_sum.clone(),
        ));
    }
}
let mut permutation_futures = Vec::with_capacity(num_records);
for (i, row) in permutation_inputs.iter().enumerate() {
    permutation_futures.push(async move {
        ctx.sum_of_products(RecordId::from(i), row.as_slice());
     });
}
try_join_all(permutation_futures).await
```

richajaindce: We are not copying the elements but their locations only for cumulative_sum

benjaminsavage: I see, I was wrong - you're not copying just taking a reference. Ignore this comment.

benjaminsavage (approved):
LGTM

PR #372, opened 2023-01-04
Kill

benjaminsavage:
We do not seem to be using these files. We can always recover them from version control if we need them later.


richajaindce:
I am using sum of product in #371. Please dont delete the code

richajaindce (changes requested):
Please dont delete sum of product code. It is used in #371

PR #373, opened 2023-01-05
How to break IPA in one easy step

martinthomson:
Something for @akoshelev and I to spend some time on.  This locks up for me.


akoshelev:
I am looking into it now

akoshelev:
@martinthomson looks like you got to the bottom of it - there is no one-buffer-size-fits-all solution. Are you working towards exposing buffer capacity to protocols (meaning they will be set per step) or just rejecting such inputs (after timeout occurs perhaps?)

PR #363, opened 2022-12-23
`Transport` trait definition and in-memory implementation

akoshelev:
As we discussed in #352, we need another abstraction on the network layer that is capable of sending messages and delivering events. This change brings the trait defition to main branch along with in-memory implementation for it (both @thurstonsand and I were working on it). `TestWorld` have been migrated to use transport layer too - hence the size of this change. The reason why I want to merge it to main is because @martinthomson, @andyleiserson are working on the same codebase at the moment and merging it later will be super painful. 

We can focus this review around `Transport` trait definition, `InMemoryTransport` and `Network` structs. Other changes are pretty mechanical.

PR #374, opened 2023-01-07
Enable concurrency testing

akoshelev:
I finally found the bug that was blocking shuttle tests. The issue outlined in #256 was fixed in
#363 but there was another bug inside the concurrency test itself. This change fixes it and enables concurrency testing as part of PR workflow approval.

It also adds a test for semi-honest IPA but given that it is incredibly slow, it only runs it for 2 iterations. Once we improve the IPA latency, number of iterations can be increased.

### DIFF
--- a/src/tests/protocol.rs
+++ b/src/tests/protocol.rs
@@ -40,7 +40,7 @@ fn semi_honest_ipa() {
                     .reconstruct();
 
                 assert_eq!(MAX_BREAKDOWN_KEY, result.len() as u128);
-            })
+            });
         },
         10,
     );
--- a/src/tests/protocol.rs
+++ b/src/tests/protocol.rs
@@ -10,11 +10,11 @@ fn semi_honest_ipa() {
     shuttle::check_random(
         || {
             shuttle::future::block_on(async {
-                const BATCHSIZE: usize = 20;
+                const BATCHSIZE: usize = 5;
                 const PER_USER_CAP: u32 = 10;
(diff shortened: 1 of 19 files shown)


src/tests/protocol.rs
```diff
@@ -0,0 +1,54 @@
+#[cfg(all(feature = "shuttle", test))]
+mod randomized {
```
martinthomson, line 2: This doesn't need a module here, does it?  You could make this:

```suggestion
#![cfg(all(feature = "shuttle", test))]
```

src/tests/protocol.rs
```diff
@@ -0,0 +1,54 @@
...
+    use crate::ff::Fp32BitPrime;
+    use crate::protocol::ipa::ipa;
+    use crate::rand::thread_rng;
+    use crate::test_fixture::{IPAInputTestRow, Reconstruct, Runner, TestWorld};
+
+    #[test]
+    fn semi_honest_ipa() {
+        shuttle::check_random(
+            || {
+                shuttle::future::block_on(async {
```
martinthomson, line 12: That's a lot of indent.

src/tests/protocol.rs
```diff
@@ -0,0 +1,54 @@
...
+                    const BATCHSIZE: usize = 20;
```
martinthomson, line 13: Would a smaller batch allow you to increase the number of iterations?

martinthomson (approved):

PR #377, opened 2023-01-10
Rename `Replicated`

taikiy:
Simple renamings of `Replicated` and `MaliciousReplicated` to `ReplicatedAdditiveShares` and `MaliciousReplicatedAdditiveShares`. Touches many files but no logic change is introduced by this diff.

Also moved the `const ZERO` to `SharedValue` as it's common to both arithmetic and boolean shares.

### DIFF
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -23,7 +23,6 @@ pub trait Field: ArithmeticShare + From<u128> + Into<Self::Integer> {
     type Integer: Int;
 
     const PRIME: Self::Integer;
-
     /// Multiplicative identity element
     const ONE: Self;
 
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -23,8 +23,7 @@ pub trait Field: ArithmeticShare + From<u128> + Into<Self::Integer> {
     type Integer: Int;
 
     const PRIME: Self::Integer;
-    /// Additive identity element
-    const ZERO: Self;
+
     /// Multiplicative identity element
     const ONE: Self;
 
--- a/src/secret_sharing/replicated/mod.rs
+++ b/src/secret_sharing/replicated/mod.rs
@@ -1,0 +1,2 @@
+pub mod malicious;
+pub mod semi_honest;
--- a/src/secret_sharing/replicated/semi_honest/mod.rs
+++ b/src/secret_sharing/replicated/semi_honest/mod.rs
@@ -1,0 +1,3 @@
+mod additive_share;
+
+pub use additive_share::AdditiveShare;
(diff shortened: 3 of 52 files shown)


src/ff/field.rs
```diff
@@ -23,8 +23,7 @@ pub trait Field: ArithmeticShare + From<u128> + Into<Self::Integer> {
     type Integer: Int;
 
     const PRIME: Self::Integer;
-    /// Additive identity element
-    const ZERO: Self;
+
```
martinthomson, line 26:
```suggestion
```

martinthomson (commented):
Skimming through this, I was struck by just how long and unwieldy the new name is.

Is there some other name we can use that is shorter?  Is it time for us to consider an acronym for this (like RAS, MRAS, RXS, MRXS) or is that going too far?

taikiy:
@akoshelev @martinthomson, I moved `Replicated` and `MaliciousReplicated` to separate module paths. Also, I decided to use aliasing (`use crate::secret_sharing::replicated::semi_honest::AdditiveShare as Replicated`) rather than renaming all the references in the code to keep the change minimal. It'll probably easier for you to review it by going to "[Files changed](https://github.com/private-attribution/ipa/pull/377/files)" rather than the new commit.

src/secret_sharing/replicated/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod malicious;
+pub mod semi_honest;
```
martinthomson, line 2: This is probably my one reservation about this, though I'm on the fence about it.

You have a layer here where there are no peers, which suggests that you could maybe flatten this.  That said, this is nicely descriptive.

src/secret_sharing/replicated/semi_honest/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod additive_share;
+
+pub use additive_share::AdditiveShare;
```
martinthomson, line 3: A no-op mod.rs probably isn't helping much here.

You could drop the module and even the subdirectory and have malicious.rs and semi_honest.rs in the replicated/ direcory.

martinthomson (approved):
LGTM.

I'd like a second opinion though on the general shape of this, but I think we should take this.  And soon; this stuff rots fast.

src/secret_sharing/replicated/semi_honest/mod.rs
taikiy: I'm going to add xor_share.rs under replicated/semi_honest/. Not sure whether we'll have the malicious version of it, but it'll be nice to have a consistent directory structure?

benjaminsavage (approved):
LGTM

PR #379, opened 2023-01-11
Refactoring sum of product to take a pair instead of vector of a's and b's

richajaindce:
- Also includes minor changes to handle cargo doc errors

### DIFF
--- a/src/protocol/basics/sum_of_product/malicious.rs
+++ b/src/protocol/basics/sum_of_product/malicious.rs
@@ -110,17 +110,14 @@ mod test {
 
     #[tokio::test]
     pub async fn simple() {
-        const MULTI_BIT_LEN: usize = 10;
+        const BATCHSIZE: usize = 10;
         let world = TestWorld::new().await;
 
         let mut rng = thread_rng();
 
-        let (mut av, mut bv) = (
-            Vec::with_capacity(MULTI_BIT_LEN),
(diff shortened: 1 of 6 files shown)


src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -130,9 +130,11 @@ mod test {
 
         let res = world
             .malicious((av, bv), |ctx, (a_share, b_share)| async move {
-                let a_refs = a_share.iter().collect::<Vec<_>>();
-                let b_refs = b_share.iter().collect::<Vec<_>>();
-                ctx.sum_of_products(RecordId::from(0), a_refs.as_slice(), b_refs.as_slice())
+                let mut pairs = Vec::with_capacity(MULTI_BIT_LEN);
```
benjaminsavage, line 133: MULTI_BIT_LEN does not belong here. This file is just for the abstract dot-product computation. Can you rename it something generic like VECTOR_LENGTH?

benjaminsavage (approved):
Just change that one constant name.

If as @akoshelev worries - this generates a bunch of copying things around - we can always just change it later.

src/protocol/basics/sum_of_product/malicious.rs
richajaindce: Changed to BATCHSIZE.

benjaminsavage (approved):

PR #383, opened 2023-01-12
Multi bit permutation should be 0 indexed

richajaindce:


benjaminsavage (approved):

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -39,9 +41,14 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
 ) -> Result<Vec<S>, Error> {
     let num_multi_bits = input.len();
     assert!(num_multi_bits > 0);
+    if num_multi_bits == 1 {
+        return bit_permutation(ctx, &input[0]).await;
+    }
```
benjaminsavage, line 46: I'm not sure if we will even keep `bit_permutations`, can't we just use this for all, even one bit?

PR #384, opened 2023-01-12
Optimized sort which takes in num_multi_bits as an argument and run an optimized sort

richajaindce:


src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -0,0 +1,197 @@
...
+            bit_num + num_multi_bits
+        };
+
+        let futures =
+            (bit_num..last_bit_num)
+                .zip(repeat(ctx_bit.clone()))
+                .map(|(idx, ctx_bit)| async move {
+                    secureapplyinv(
+                        ctx_bit.narrow(&MultiApplyInv(idx)),
+                        sort_keys[idx as usize].clone(),
```
benjaminsavage, line 104: Hmmm... This isn't what I had expected to see.

You're calling `secureapplyinv` in a loop. Can't we just apply this one time, to all of the affected rows? The same way that we eventually apply the sort permutation to ALL of the rows of the inputs to the IPA query.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -0,0 +1,197 @@
...
+                });
+        let bits_i_sorted_by_less_significant_bits = try_join_all(futures).await?;
+
+        let bits_i_permutation = multi_bit_permutation(
+            ctx_bit.narrow(&BitPermutationStep),
+            &bits_i_sorted_by_less_significant_bits,
+        )
+        .await?;
+
+        let composed_i_permutation = compose(
```
benjaminsavage, line 118: These names which include an `_i_` don't make sense now that we are doing multi-bit. I'd suggest:

`next_few_bits_sorted_by_less_significant_bits`

and

`next_few_bits_permutation`

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -0,0 +1,197 @@
...
+                revealed_and_random_permutations
+                    .randoms_for_shuffle
+                    .1
+                    .as_slice(),
+            ),
+            &revealed_and_random_permutations.revealed,
+            bits_i_permutation,
+        )
+        .await?;
+        composed_less_significant_bits_permutation = composed_i_permutation;
```
benjaminsavage, line 134: Why even have this local variable `composed_i_permutation` in the first place then? Why not just assign directly to `composed_less_significant_bits_permutation`?

benjaminsavage (changes requested):

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Each loop iteration each async move so should happen in parallel.

src/protocol/sort/generate_permutation_opt.rs
richajaindce: I just copy pasted the code original sort. Then it was somehow needed to assign it.

src/protocol/sort/generate_permutation_opt.rs
benjaminsavage: Sure, it's parallelized, but is there any more communication going on this way?

benjaminsavage (approved):
Looks good!

We can continue the discussion on my one comment, and if necessary follow up with another PR.

Issue #380, opened 2023-01-12
Explore whether cloning context can be made more efficient


akoshelev:
Our software design for IPA require protocol contexts to be cheaply cloneable because we clone them a lot. 

```rust
pub struct SemiHonestContext<'a, F: Field> {
    inner: Arc<ContextInner<'a>>,
    step: Step,
}
```

Making steps cloning inexpensive is tracked in #139, this issue focuses on improvements that can be made on protocol contexts, specifically on `ContextInner` clone. It looks innocent (atomic load/store on each clone) but when doing it often it may lead to false sharing issues on CPU caches. There may be a better way to do this if we look at the structure of the `ContextInner` struct:

```rust
pub(super) struct ContextInner<'a> {
    pub role: Role,
    pub prss: &'a PrssEndpoint,
    pub gateway: &'a Gateway,
}
```

There is a role bit and two references that never change no matter how many times we clone the context. There might be an opportunity to cache those values (not references) in one place. Root-level future that is created to run protocol may be a good place to do that and Tokio crate provides some support for it: https://docs.rs/tokio/latest/tokio/task/struct.LocalKey.html.

## 10000 foot view

This is how query processing may look like

* Helpers receive a request to process new query. They dynamically allocate roles for each of them and set up PRSS.
* Each helper stores role, PRSS endpoint and Gateway inside task local data cell (tokio localkey)
* Query processing is launched by spawning a tokio task (or any other runtime task). In case of tokio it would look like

```rust
   tokio::spawn(|| async {
        CONTEXT_INNER.scope((prss, gateway, role), async move {
              let context = SemiHonestContext::new("step");  // no need to pass PRSS or Gateway anymore      
         }).await;
   });

   impl SemiHonestContext {
          pub fn prss(&self) -> &PrssEndpoint {
              CONTEXT_INNER.with(|inner| inner.0) // not sure if that works fine
          }
   }
```

This design does not require changing protocols code, everything is contained within Infra layer 

## Complications
* Our plan is to run IPA software on Cloudflare Workers that do not support Tokio: https://github.com/cloudflare/workers-rs. `LocalKey` implementation however does not depend on tokio runtime so there  may be no issues

andyleiserson:
For the malicious case, there's also some data in ContextInner that isn't static -- `upgrade_ctx` and `accumulator`. (And I discovered today that upgrades are not just for inputs, they're also used in `solved_bits`). We may need to revisit the handling of these.

akoshelev:
[...]

PR #387, opened 2023-01-12
rework upgrades for malicious protocol

andyleiserson:
Split this off from #360.

The purpose of these changes is to allow managing total_records on the upgrade_ctx without exposing a
set_total_upgrades function on MaliciousContext.

This ended up being a bit more involved than I thought it would be originally. I could pretty easily change this to get rid of set_total_upgrades without the UpgradeContext stuff, by adding an upgrade_vector function on ContextInner and narrowing the IntoMalicious impl for IntoIter to be specific to vectors.

I do think it's possible we'll want UpgradeContext as an entity, either as part of the changes to have contexts own open channels, or as part of the changes to make context cloning more efficient (#380).

### DIFF
--- a/src/protocol/boolean/generate_random_bits/malicious.rs
+++ b/src/protocol/boolean/generate_random_bits/malicious.rs
@@ -20,7 +20,7 @@ impl<F: Field> RandomBits<F> for MaliciousContext<'_, F> {
         let ctx = &c;
         let malicious_triples =
             try_join_all(triples.into_iter().enumerate().map(|(i, t)| async move {
-                ctx.upgrade_bit_triple(&BitOpStep::from(i), record_id, t)
+                ctx.upgrade_for_record_with(&BitOpStep::from(i), record_id, t)
                     .await
             }))
             .await?;
--- a/src/protocol/context/malicious.rs
+++ b/src/protocol/context/malicious.rs
@@ -286,7 +286,9 @@ impl<'a, F: Field> ContextInner<'a, F> {
 /// ```no_run
 /// use raw_ipa::protocol::{context::{NoRecord, UpgradeContext, UpgradeToMalicious}, RecordId};
 /// use raw_ipa::ff::Fp31;
-/// use raw_ipa::secret_sharing::{MaliciousReplicated, Replicated};
+/// use raw_ipa::secret_sharing::replicated::{
+///     malicious::AdditiveShare as MaliciousReplicated, semi_honest::AdditiveShare as Replicated,
+/// };
 /// let _ = <UpgradeContext<Fp31, NoRecord> as UpgradeToMalicious<Replicated<Fp31>, _>>::upgrade;
 /// let _ = <UpgradeContext<Fp31, RecordId> as UpgradeToMalicious<Replicated<Fp31>, _>>::upgrade;
 /// let _ = <UpgradeContext<Fp31, NoRecord> as UpgradeToMalicious<(Replicated<Fp31>, Replicated<Fp31>), _>>::upgrade;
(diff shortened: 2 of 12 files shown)


src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -225,10 +224,7 @@ mod tests {
 
                 let v = MaliciousValidator::new(ctx);
                 let m_ctx = v.context();
-                let m_triple = m_ctx
-                    .upgrade_bit_triple(&BitOpStep::from(0), RecordId::from(0), triple)
-                    .await
-                    .unwrap();
+                let m_triple = m_ctx.upgrade(triple).await.unwrap();
```
benjaminsavage, line 227: This new version is much shorter and nicer looking!

src/protocol/sort/generate_permutation.rs
```diff
@@ -270,9 +270,7 @@ where
     let m_ctx_0 = m_ctx.narrow(&Sort(0));
     assert_eq!(sort_keys.len(), num_bits as usize);
 
-    let upgraded_sort_keys = m_ctx
-        .upgrade_vector(&MaliciousUpgradeInput, sort_keys[0].clone())
-        .await?;
+    let upgraded_sort_keys = m_ctx.upgrade(sort_keys[0].clone()).await?;
```
benjaminsavage, line 273: It's also nice you were able to take care of vectors!

src/protocol/boolean/generate_random_bits/malicious.rs
```diff
@@ -20,7 +20,7 @@ impl<F: Field> RandomBits<F> for MaliciousContext<'_, F> {
         let ctx = &c;
         let malicious_triples =
             try_join_all(triples.into_iter().enumerate().map(|(i, t)| async move {
-                ctx.upgrade_bit_triple(&BitOpStep::from(i), record_id, t)
+                ctx.upgrade_for_record_with(&BitOpStep::from(i), record_id, t)
```
benjaminsavage, line 23: This one seems different from all the others. Is this the only place where the client needs to provide a record_id when upgrading?

andyleiserson: Yes... this is the only place where the entire input isn't presented for upgrade all at once (it's not really an input at all -- it's randomness I think used for bit decomposition), so it's not possible to refactor it such that the record IDs are generated internal to the upgrade implementation for the thing being upgraded.

It might be possible to refactor this to `upgrade_for_record<T>(record_id, input: T)` with  `T = Vector<BitConversionTriple>` (getting rid of the `_with`) if that's preferable.

src/protocol/context/malicious.rs
```diff
@@ -335,3 +387,87 @@ impl<'a, F: Field> ContextInner<'a, F> {
...
+#[async_trait]
+impl<'a, F, T, TM, U, UM> UpgradeToMalicious<(T, U), (TM, UM)> for UpgradeContext<'a, F, NoRecord>
+where
+    F: Field,
+    T: Send + 'static,
+    U: Send + 'static,
+    TM: Send + Sized,
+    UM: Send + Sized,
+    for<'u> UpgradeContext<'u, F, NoRecord>: UpgradeToMalicious<T, TM>,
+    for<'u> UpgradeContext<'u, F, NoRecord>: UpgradeToMalicious<U, UM>,
```
martinthomson, line 400:
```suggestion
    for<'u> UpgradeContext<'u, F, NoRecord>: UpgradeToMalicious<T, TM> + UpgradeToMalicious<U, UM>,
```

src/protocol/boolean/generate_random_bits/malicious.rs
martinthomson: Not really worth holding this change up.  Want to open an issue so that we can capture any ideas you have and maybe revisit later?

martinthomson (approved):

akoshelev:
merging it as we have all necessary approvals

PR #388, opened 2023-01-12
Eliminate more calls to tokio::try_join!

akoshelev:
Apologies, missed more calls to `tokio::try_join!` :(

Verified that after this change there are no more calls to it

```bash
 grep -r "tokio::try_join" . --include \*.rs | wc -l
       0
```

### DIFF
--- a/src/protocol/attribution/credit_capping.rs
+++ b/src/protocol/attribution/credit_capping.rs
@@ -310,12 +310,11 @@ impl AsRef<str> for Step {
 
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
-    use super::super::tests::generate_shared_input;
     use crate::{
         ff::{Field, Fp32BitPrime},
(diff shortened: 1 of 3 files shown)


src/protocol/attribution/credit_capping.rs
```diff
@@ -310,12 +310,11 @@ impl AsRef<str> for Step {
 
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
-    use super::super::tests::generate_shared_input;
     use crate::{
         ff::{Field, Fp32BitPrime},
         protocol::attribution::{
             credit_capping::credit_capping,
-            tests::{BD, H, S, T},
+            tests::{generate_shared_input, BD, H, S, T},
```
akoshelev, line 317: this is unrelated change: I moved things around here and Clion decided to format it this way. Decided to keep it because PR is small and it is better to import it this way

PR #389, opened 2023-01-13
WIP

benjaminsavage:
This doesn't compile yet

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -12,18 +12,23 @@ use crate::{
         context::Context,
         sort::{
             apply_sort::apply_sort_permutation,
-            generate_permutation::generate_permutation_and_reveal_shuffled,
+            generate_permutation::{
+                generate_permutation_and_reveal_shuffled,
+                malicious_shuffle_and_reveal_permutation,
+            },
+            SortStep::ShuffleRevealPermutation,
         },
         RecordId,
     },
-    secret_sharing::{Replicated, XorReplicated},
+    secret_sharing::{Replicated, XorReplicated, SecretSharing, MaliciousReplicated},
 };
 use async_trait::async_trait;
 use futures::future::{try_join, try_join_all};
+use hyper::upgrade;
 
-use super::{attribution::AggregateCreditOutputRow, context::SemiHonestContext};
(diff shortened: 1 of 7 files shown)


src/protocol/ipa/mod.rs
```diff
@@ -129,6 +146,140 @@ impl<F: Field + Sized> Resharable<F> for IPAModulusConvertedInputRow<F> {
     }
 }
 
+#[async_trait]
+impl<F: Field + Sized> Resharable<F> for IPAMaliciousModulusConvertedInputRow<F> {
+    type Share = MaliciousReplicated<F>;
```
richajaindce, line 151: Is there any other way to avoid this duplication of code for just declaring a Share type as Malicious? @akoshelev @martinthomson ?

src/protocol/ipa/mod.rs
```diff
@@ -129,6 +146,140 @@ impl<F: Field + Sized> Resharable<F> for IPAModulusConvertedInputRow<F> {
...
+                    .await?;
+                Ok::<_, Error>(IPAMaliciousModulusConvertedInputRow {
+                    mk_shares: m_mk_shares,
+                    is_trigger_bit: m_is_trigger_bit,
+                    breakdown_key: m_breakdown_key,
+                    trigger_value: m_trigger_value,
+                })
+            }
+        })
+        .collect::<Vec<_>>();
```
richajaindce, line 270: Should we put this in a separate function of its own?
Also, no need to collect this future iterator. try_join_all can take in an iterator

src/protocol/ipa/mod.rs
```diff
@@ -129,6 +146,140 @@ impl<F: Field + Sized> Resharable<F> for IPAModulusConvertedInputRow<F> {
     }
 }
 
+#[async_trait]
+impl<F: Field + Sized> Resharable<F> for IPAMaliciousModulusConvertedInputRow<F> {
```
martinthomson, line 150: Isn't Field already Sized?

src/protocol/ipa/mod.rs
martinthomson: The way we have done this in the past is to have `IpaModulusConvertedInputRow` be generic over `S: SecretSharing` (and probably also `F: Field`).  Then you can implement `Resharable<F> for IpaModulusConvertedInputRow<S, F>`.

This line becomes `type Share = S`, I think.  The interaction between type constraints and associated types is still something I'm not 100% on.

benjaminsavage:
Abandoning in favor of #430

PR #390, opened 2023-01-13
Create a macro to generate BitArray variations

taikiy:
Similar to `Field`, create a macro to generate BitArray structs with a specified bit length. Added `BitArray40` for the match key.

`BitArray64` is intact which we could use for other report fields i.e. breakdown key, but I suppose we'll use a lot shorter bit-length struct for that.

### DIFF
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -1,193 +1,199 @@
 use super::BitArray;
 use crate::secret_sharing::SharedValue;
 use bitvec::prelude::{BitArr, Lsb0};
-use std::{
-    fmt::Debug,
-    ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Index, Not},
-};
 
+/// Bit store type definition. Five `u8` blocks.
+type U8_5 = BitArr!(for 40, in u8, Lsb0);
 /// Bit store type definition. Eight `u8` blocks.
 type U8_8 = BitArr!(for 64, in u8, Lsb0);
 
-/// 64-bit array of bits. Similar to `u64`, but supports boolean algebra, and
-/// provides access to individual bits via index.
-///
-/// Bits are stored in the Little-Endian format. Accessing the first element
-/// like `s[0]` will return the LSB.
-///
-/// ## Example
-/// ```
-/// use raw_ipa::bits::BitArray64;
-///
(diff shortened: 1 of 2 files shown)


src/bits/bit_array.rs
```diff
@@ -1,193 +1,199 @@
...
+                        "Bit array size {} is too small to hold the value {}.",
+                        Self::BITS,
+                        v
+                    ))
+                }
+            }
+        }
+
+        impl std::ops::Index<usize> for $array {
+            type Output = u8;
```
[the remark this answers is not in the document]
taikiy: Thanks! Removed the custom `Index` and `TryFrom`, and all other trait implementations that are provided by a bare BitArray.

src/bits/mod.rs
```diff
@@ -1,19 +1,16 @@
...
-pub trait BitArray: BooleanShare + TryFrom<u128> + Index<usize> {
+pub trait BitArray: BooleanShare + Index<usize> {
     /// A bit array with all its elements initialized to 0.
     const ZERO: Self;
 
     /// Truncates the higher-order bits larger than `Self::BITS`, and converts
-    /// into this data type. This conversion is lossy. Callers are encouraged
-    /// to use `try_from` if the input is not known in advance.
+    /// into this data type. This conversion is lossy.
     fn truncate_from<T: Into<u128>>(v: T) -> Self;
```
martinthomson, line 14: This could be marked `#[cfg(test)]` to avoid accidental misuse.

martinthomson (commented):
WFM

src/bits/mod.rs
taikiy: This will be used in production. I spoke with Alex and this is the name we agreed on for the lossy conversion. Alternatively, we could use BitSlice or bitarr! macro to create a new BiArray instance, but neither is not user-friendly. I think we'll almost always use this function. Or we might need to bring back TryFrom<u128>.

taikiy:
Per discussion offline with @martinthomson. Addressed PR comments other than removing `TryFrom<u128>` and using a bare BitArray. I think we'll need a wrapper struct to implement i.e. `Distribution` for the bit array structs for `IntoShares`.

benjaminsavage (commented):
Thanks for adding this @taikiy

benjaminsavage (approved):

PR #391, opened 2023-01-13
More efficient multi-bit permutation generation

benjaminsavage:
tl;dr

Sorting 1000 rows:

= Old version =
(3-bit multi-sort) Total records sent: 1,950,000
(4-bit multi-sort) Total records sent: 3,129,000

= New version =
(3-bit multi-sort) Total records sent: 1,194,000
(4-bit multi-sort) Total records sent: 1,353,000

= Summary =
(3-bit multi-sort) 
- 756,000 fewer records sent
- 38.8% reduction

(4-bit multi-sort)
- 1,776,000 fewer records sent
- 56.7% reduction

For a 3-bit multi-bit-sort, here is what you do for each row:

- check if it's a bitwise representation of zero
- check if it's a bitwise representation of one
- check if it's a bitwise representation of two
- check if it's a bitwise representation of three
- check if it's a bitwise representation of four
- check if it's a bitwise representation of five
- check if it's a bitwise representation of six
- check if it's a bitwise representation of seven

So let's say that for a given row, you have three secret shares representing the three bits. Let's call it:

[ $b_0, b_1, b_2$ ]

Logically speaking, the way you check all 8 of these conditions is to compute the following:

- $(1-b_0) \times (1-b_1) \times (1 - b_2)$
- $(1-b_0) \times (1-b_1) \times b_2$
- $(1-b_0) \times b_1 \times (1 - b_2)$
- $(1-b_0) \times b_1 \times b_2$
- $b_0 \times (1-b_1) \times (1 - b_2)$
- $b_0 \times (1-b_1) \times b_2$
- $b_0 \times b_1 \times (1 - b_2)$
- $b_0 \times b_1 \times b_2$

Prior to this diff, that's exactly what we did. Very simple. Just two multiplications per equality check, so a total of 2 * 8 = 16 multiplications per row.

But you can expand these expressions out:
- $1 - b_0 - b_1 - b_2 + b_0 b_1 + b_0 b_2 + b_1 b_2$
- $b_2 - b_0 b_2 - b_1 b_2 + b_0 b_1 b_2$
- $b_1 - b_0 b_1 - b_1 b_2 + b_0 b_1 b_2$
- $b_1 b_2 - b_0 b_1 b_2$
- $b_0 - b_0 b_1 - b_0 b_2 + b_0 b_1 b_2$
- $b_0 b_2 - b_0 b_1 b_2$
- $b_0 b_1 - b_0 b_1 b_2$
- $b_0 b_1 b_2$

And as you can see, they are all just linear combinations of the same set of common terms! So we only need to pre-compute the following 4 things:

- $b_0 \times b_1$
- $b_0 \times b_2$
- $b_1 \times b_2$
- $b_0 \times b_1 \times b_2 $

...and then we can compute all 8 equality checks by just adding / subtracting those things in various combinations.

The tricky thing is determining *which* terms to use for each equality check, and what sign each should have.

I thought of it this way:
- The first term is either $b_0$ or $(1 - b_0)$. 
- Then we multiply by the second term, which is either $b_1$ or $(1 - b_1)$
- Then we multiply by the third term, which is either $b_2$ or $(1 - b_2)$

So I pre-compute these terms in that order:

First I add two elements:
$[1, b_0]$

Then, I multiply both of these by $b_1$ and append the results to the array:

$[1, b_0, b_1, b_0 b_1]$
(multiplying $1 * b_1$ clearly doesn't need a real multiplication - we can skip that one).

Then we multiply all four elements by $b_2$ and append the results to the array:

$[1, b_0, b_1, b_0 b_1, b_2, b_0 b_2, b_1 b_2, b_0 b_1 b_2]$
(multiplying $1 * b_2$ clearly doesn't need a real multiplication - we can skip that one too..)

If we were using more than 3 bits, we would just continue in this way.

Now this array of pre-computed values can be thought of like a *binary tree*. At each node in the tree, you can either go left (multiply by one), or go right (multiply by the next bit). To go "left" you just stay at the same index. To go "right" you just add the "step" which is 2^bit. 

Now, when we are trying to compute an equality check to a specific number, we can use this "binary tree" to help us. Let's say we are checking equality to "4". First, we compute the binary version of this number, 100. Then we use each bit to guide us. The first bit is a "1", so originally we go "right" (because we just multiply by $b_0$ in this case, not by $(1 - b_0)$. ). The next bit is a "0". This indicates that we need to multiply by $(1 - b_1)$. So now we have to go left AND right, and add the recursive results. Finally, the 3rd bit is also "0", indicating that we need to multiply by $(1 - b_2)$, so we need both of the branches from the last step to both go left AND right. In this way, we have re-derived that we need to compute $b_0 - b_0 b_1 - b_0 b_2 + b_0 b_1 b_2$, and we grabbed those four terms from out of our pre-computed array.

### DIFF
--- a/src/protocol/sort/multi_bit_permutation.rs
+++ b/src/protocol/sort/multi_bit_permutation.rs
@@ -106,13 +106,40 @@ where
     // and checks if the bits are equal to this value. It does so my computing a linear combination of the
     // pre-computed coefficients.
     //
-    // Observe that the coefficients necessary to compute a given equality check form a Sierpiński triangle
-    // <https://en.wikipedia.org/wiki/Sierpi%C5%84ski_triangle#/media/File:Multigrade_operator_AND.svg>
+    // Observe that whether a given precomputed coefficient contributes to a
+    // given equality check follows a Sierpiński triangle
+    // https://en.wikipedia.org/wiki/Sierpi%C5%84ski_triangle#/media/File:Multigrade_operator_AND.svg.
     //
-    // Martin Thomson found that the sign of the coefficients follows the following pattern:
-    // 1. Take the row and column and bitwise AND them: $(!i & j)$
-    // 2. Count the number of non-zero bits in the result
-    // 3. If the result is an odd number, the coefficient is negative. If the result is even, the coefficient is positive.
+    // For example, for a three bit value, we have the following:
+    // 1 | 1 | 1 | 1 | 1 | 1 | 1 | 1
+    // 0 | 1 | 0 | 1 | 0 | 1 | 0 | 1
+    // 0 | 0 | 1 | 1 | 0 | 0 | 1 | 1
+    // 0 | 0 | 0 | 1 | 0 | 0 | 0 | 1
+    // 0 | 0 | 0 | 0 | 1 | 1 | 1 | 1
+    // 0 | 0 | 0 | 0 | 0 | 1 | 0 | 1
+    // 0 | 0 | 0 | 0 | 0 | 0 | 1 | 1
+    // 0 | 0 | 0 | 0 | 0 | 0 | 0 | 1
+    //
+    // This can be computed from row (i) and column (j) indices with i & j == i
+    //
+    // The sign of the inclusion is less obvious, but we discovered that this
+    // can be found by taking the same row (i) and column (j) indices:
+    // 1. Invert the row index and bitwise AND the values: a = !i & j
+    // 2. Count the number of bits that are set: b = a.count_ones()
+    // 3. An odd number means a positive coefficient; an odd number means a negative.
+    //
+    // For example, for a three bit value, step 1 produces (in binary):
+    // 000 | 001 | 010 | 011 | 100 | 101 | 110 | 111
+    // 000 | 000 | 010 | 010 | 100 | 100 | 110 | 110
+    // 000 | 001 | 000 | 001 | 100 | 101 | 100 | 101
+    // 000 | 000 | 000 | 000 | 100 | 100 | 100 | 100
+    // 000 | 001 | 010 | 011 | 000 | 001 | 010 | 011
+    // 000 | 000 | 010 | 010 | 000 | 000 | 010 | 010
+    // 000 | 001 | 000 | 001 | 000 | 001 | 000 | 001
+    // 000 | 000 | 000 | 000 | 000 | 000 | 000 | 000
+    //
+    // Where 000, 101, 011, and 110 mean positive contributions, and
+    // 001, 010, 100, and 111 mean negative contributions.
     let side_length = 1 << num_bits;
     let mut equality_checks = Vec::with_capacity(side_length);
     for i in 0..side_length {
@@ -170,27 +197,19 @@ where
     for (bit_idx, column) in input.iter().enumerate() {
         let bit = &column[record_idx];
         let step = 1 << bit_idx;
-        let num_children_to_add = precomputed_combinations.len();
-        precomputed_combinations.push(bit.clone());
-        if bit_idx == 0 {
-            continue;
-        }
-        let mut multiplication_results = try_join_all(
-            precomputed_combinations
-                .iter()
-                .enumerate()
-                .skip(1)
-                .take(num_children_to_add - 1)
-                .map(|(j, precomputed_combination)| {
+        let mut multiplication_results =
+            try_join_all(precomputed_combinations.iter().skip(1).enumerate().map(
+                |(j, precomputed_combination)| {
                     let child_idx = j + step;
                     ctx.narrow(&BitOpStep::from(child_idx)).multiply(
                         record_id,
                         precomputed_combination,
                         bit,
                     )
-                }),
-        )
-        .await?;
+                },
+            ))
+            .await?;
+        precomputed_combinations.push(bit.clone());
         precomputed_combinations.append(&mut multiplication_results);
     }
     Ok(precomputed_combinations)
--- a/src/protocol/sort/multi_bit_permutation.rs
+++ b/src/protocol/sort/multi_bit_permutation.rs
@@ -95,32 +95,28 @@ where
 {
     let record_id = RecordId::from(record_idx);
     let num_bits = input.len();
-    let mut step = 1 << num_bits;
-    let mut precomputed_combinations = vec![S::ZERO; step];
+    let mut precomputed_combinations = Vec::with_capacity(1 << num_bits);
+    precomputed_combinations.push(ctx.share_of_one());
     for bit_idx in 0..num_bits {
         let bit = &input[num_bits - bit_idx - 1][record_idx];
-        step >>= 1;
+        let step = 1 << bit_idx;
+        let num_children_to_add = precomputed_combinations.len();
+        precomputed_combinations.push(bit.clone());
         if bit_idx == 0 {
-            precomputed_combinations[0] = ctx.share_of_one();
-            precomputed_combinations[step] = bit.clone();
-        } else {
-            let num_children_to_add = 1 << bit_idx;
-            let mut multiplication_futures = Vec::with_capacity(num_children_to_add - 1);
-            for j in 1..num_children_to_add {
-                let parent_idx = 2 * j * step;
-                let child_idx = parent_idx + step;
-                let parent = &precomputed_combinations[parent_idx];
-                multiplication_futures.push(
-                    ctx.narrow(&BitOpStep::from(child_idx))
-                        .multiply(record_id, parent, bit),
-                );
-            }
-            let multiplication_results = try_join_all(multiplication_futures).await?;
-            precomputed_combinations[step] = (*bit).clone();
-            for (j, mult_result) in multiplication_results.into_iter().enumerate() {
-                precomputed_combinations[step * (2 * (j + 1) + 1)] = mult_result;
-            }
+            continue;
         }
+        let mut multiplication_futures = Vec::with_capacity(num_children_to_add - 1);
+        #[allow(clippy::needless_range_loop)]
+        for j in 1..num_children_to_add {
+            let child_idx = j + step;
+            multiplication_futures.push(ctx.narrow(&BitOpStep::from(child_idx)).multiply(
+                record_id,
+                &precomputed_combinations[j],
+                bit,
+            ));
+        }
+        let mut multiplication_results = try_join_all(multiplication_futures).await?;
+        precomputed_combinations.append(&mut multiplication_results);
     }
 
     let mut equality_checks = Vec::with_capacity(1 << num_bits);
@@ -131,6 +127,7 @@ where
             0,
             1_i8,
             precomputed_combinations.as_slice(),
+            1,
         ));
     }
     Ok(equality_checks)
@@ -142,6 +139,7 @@ fn check_equality_to<F: Field, S: SecretSharing<F>>(
     idx: usize,
     sign: i8,
     precomputed_combinations: &[S],
+    step: usize,
 ) -> S {
     if bit_number == 0 {
         if sign == 1_i8 {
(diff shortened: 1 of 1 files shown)


src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+            num_bits,
+            precomputed_combinations.as_slice(),
+            0,
+            0,
+        ));
+    }
+    Ok(equality_checks)
+}
+
+fn check_equality_to<F: Field, S: SecretSharing<F>>(
```
martinthomson, line 136: This is obviously correct (the tests pass), but the code is incomprehensible to me.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+            multiplication_futures.push(ctx.narrow(&BitOpStep::from(child_idx)).multiply(
+                record_id,
+                &precomputed_combinations[j],
+                bit,
+            ));
+        }
+        let mut multiplication_results = try_join_all(multiplication_futures).await?;
+        precomputed_combinations.append(&mut multiplication_results);
+    }
+
```
richajaindce, line 122: Lets put this in a separate function "precompute_all_combinations".

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+            0,
+        ));
+    }
+    Ok(equality_checks)
+}
+
+fn check_equality_to<F: Field, S: SecretSharing<F>>(
+    value: u32,
+    num_bits: usize,
+    tree: &[S],
```
richajaindce, line 139: lets call it precomputed_combinations to avoid confusion

richajaindce (changes requested):
Still reviewing and thinking how to make it simpler. Its definitely hard to follow :(

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
-                .await
-            }),
-    )
-    .await
+    let record_id = RecordId::from(record_idx);
+    let num_bits = input.len();
+    let mut precomputed_combinations = Vec::with_capacity(1 << num_bits);
+    precomputed_combinations.push(ctx.share_of_one());
+    #[allow(clippy::needless_range_loop)]
+    for bit_idx in 0..num_bits {
```
richajaindce, line 101:
```suggestion
    for (bit_idx, iter) in input.iter().enumerate() {
        let bit = &iter[record_idx];
```

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+        let bit = &input[bit_idx][record_idx];
+        let step = 1 << bit_idx;
+        let num_children_to_add = precomputed_combinations.len();
+        precomputed_combinations.push(bit.clone());
+        if bit_idx == 0 {
+            continue;
+        }
+        let mut multiplication_futures = Vec::with_capacity(num_children_to_add - 1);
+        #[allow(clippy::needless_range_loop)]
+        for j in 1..num_children_to_add {
```
richajaindce, line 111:
```suggestion
        for (j, precomputed_combination) in precomputed_combinations.iter().enumerate().skip(1) {
```

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+    }
+    Ok(equality_checks)
+}
+
+fn check_equality_to<F: Field, S: SecretSharing<F>>(
+    value: u32,
+    num_bits: usize,
+    tree: &[S],
+    bit_idx: usize,
+    idx: usize,
```
richajaindce, line 141:
```suggestion
    tree_idx: usize,
```

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +83,75 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+    }
+    let bit = (value >> bit_idx) & 1;
+    let step = 1 << bit_idx;
+    let next_bit_idx = bit_idx + 1;
+    let times_bit = check_equality_to(value, num_bits, tree, next_bit_idx, idx + step);
+    if bit == 0 {
+        let times_one = check_equality_to(value, num_bits, tree, next_bit_idx, idx);
+        return times_one - &times_bit;
+    }
+    times_bit
```
richajaindce, line 154:
```suggestion
) fn check_equality_to<F: Field, S: SecretSharing<F>>(
    value: u32,
    num_bits: usize,
    precomputed_combinations: &[S],
    bit_idx: usize,
    tree_idx: usize,
) -> S {
    if bit_idx == num_bits {
        return precomputed_combinations[tree_idx].clone();
    }
    // 1<<bit_idx + tree_idx gives the right subtree equivalent for this bit (i.e. for 1-b it finds location of b)
    let times_bit = check_equality_to(value, num_bits, precomputed_combinations, bit_idx + 1, tree_idx + 1 << bit_idx);
    let bit = (value >> bit_idx) & 1;
    if bit == 0 {
        let times_one = check_equality_to(value, num_bits, precomputed_combinations, bit_idx + 1, tree_idx);
        return times_one - &times_bit;
    }
    times_bit
}
```
This isnt the best, but atleast helps know what each branch is trying to do. Some diagram would definitely be helpful here

src/protocol/sort/multi_bit_permutation.rs
benjaminsavage: Thank you! I'm incorporating this now. I couldn't figure this out before - I was missing the `take()`. I honestly still don't understand why that is required.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -50,20 +43,22 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
-        (0..num_possible_bit_values)
-            .zip(repeat(ctx.clone()))
-            .map(|(j, ctx)| async move { get_bit_equality_checkers(j, input, ctx).await }),
-    )
-    .await?;
+    let mut equality_check_futures = Vec::with_capacity(num_records);
+    for i in 0..num_records {
+        equality_check_futures.push(check_everything(ctx.clone(), i, input));
+    }
+    let equality_checks = try_join_all(equality_check_futures).await?;
```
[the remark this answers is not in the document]
benjaminsavage: Thanks! OK, I'm incorporating this as well.

src/protocol/sort/multi_bit_permutation.rs
benjaminsavage: Thank you again. I'm incorporating this as well.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +102,200 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+fn generate_lookup_table(num_bits: usize) -> Vec<Vec<i8>> {
+    let side_length = 1 << num_bits;
+    let mut sterpinski_triangle = Vec::with_capacity(side_length);
+    for _ in 0..side_length {
+        sterpinski_triangle.push(Vec::with_capacity(side_length));
+    }
+    let mut binary_representation = 1;
+    for i in 0..side_length {
+        if i > 0 {
+            binary_representation = binary_representation ^ (binary_representation << 1);
```
martinthomson, line 245: I have a far simpler sierpinski calculation:

```rust
let a = i & j;
let bit = (a) == i;
let sign = 1 - 2 * i8::try_from(a.count_ones() & 1).unwrap();
map[row].push(bit * sign);
```

martinthomson: Given that, it might be easier to calculate it at runtime.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +81,120 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
-                .await
-            }),
-    )
-    .await
+    // This loop just iterates over all the possible values this N-bit input could potentially represent
+    // and checks if the bits are equal to this value. It does so my computing a linear combination of the
+    // pre-computed coefficients.
+    //
+    // Observe that the coefficients necessary to compute a given equality check form a Sierpiński triangle
+    // <https://en.wikipedia.org/wiki/Sierpi%C5%84ski_triangle#/media/File:Multigrade_operator_AND.svg>
```
martinthomson, line 110: // Observe that whether a given precomputed coefficient contributes to a
// given equality check follows a Sierpiński triangle
// <https://en.wikipedia.org/wiki/Sierpi%C5%84ski_triangle#/media/File:Multigrade_operator_AND.svg>.
//
// For example, for a three bit value, we have the following:
// ```
// 1 | 1 | 1 | 1 | 1 | 1 | 1 | 1
// 0 | 1 | 0 | 1 | 0 | 1 | 0 | 1
// 0 | 0 | 1 | 1 | 0 | 0 | 1 | 1
// 0 | 0 | 0 | 1 | 0 | 0 | 0 | 1
// 0 | 0 | 0 | 0 | 1 | 1 | 1 | 1
// 0 | 0 | 0 | 0 | 0 | 1 | 0 | 1
// 0 | 0 | 0 | 0 | 0 | 0 | 1 | 1
// 0 | 0 | 0 | 0 | 0 | 0 | 0 | 1
// ```
// This can be computed from row (i) and column (j) indices with `i & j == i`
//
// The sign of the inclusion is less obvious, but we discovered that this
// can be found by taking the same row (i) and column (j) indices:
// 1. Invert the row index and bitwise AND the values: `a = !i & j`
// 2. Count the number of bits that are set: `b = a.count_ones()`
// 3. An odd number means a positive coefficient; an odd number means a negative.
//
// For example, for a three bit value, step 1 produces (in binary):
// ```
// 000 | 001 | 010 | 011 | 100 | 101 | 110 | 111
// 000 | 000 | 010 | 010 | 100 | 100 | 110 | 110
// 000 | 001 | 000 | 001 | 100 | 101 | 100 | 101
// 000 | 000 | 000 | 000 | 100 | 100 | 100 | 100
// 000 | 001 | 010 | 011 | 000 | 001 | 010 | 011
// 000 | 000 | 010 | 010 | 000 | 000 | 010 | 010
// 000 | 001 | 000 | 001 | 000 | 001 | 000 | 001
// 000 | 000 | 000 | 000 | 000 | 000 | 000 | 000
// ```
// Where 000, 101, 011, and 110 mean positive contributions, and
// 001, 010, 100, and 111 mean negative contributions.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -91,55 +81,120 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+        precomputed_combinations.push(bit.clone());
+        if bit_idx == 0 {
+            continue;
+        }
+        let mut multiplication_results = try_join_all(
+            precomputed_combinations
+                .iter()
+                .enumerate()
+                .skip(1)
+                .take(num_children_to_add - 1)
```
martinthomson, line 183: There might be a cleaner version of this.

You want to skip index 0, so that your step counter makes sense.  That seems right; you skip the bit where you multiply by one.  But you have to use `take()` to avoid multiplying bit by bit, which we don't need.  You could move the addition of the extra value until after the loop.

That means running the loop always, but I believe that `try_join_all(v)` for `v.is_empty()` is fast enough that the simplification might be worth it.

You can also reorder the enumerate and skip steps.  The only effect that has is on child_idx, which will be slightly smaller as a consequence.  But since `step = 1<<bit_idx` and `num_children_to_add < 1<<bit_idx`, you are never at risk of overflowing.

martinthomson (approved):

PR #392, opened 2023-01-13
Move Http Impl

thurstonsand:
Moves the HTTP impl of `Transport` to its own top level module. Also, moves `ByteArrStream` out of `net` and into `helpers`


src/helpers/mod.rs
```diff
@@ -1,9 +1,8 @@
 pub mod messaging;
 pub mod network;
-pub mod old_http;
-#[deprecated(note = "Use `Transport` instead")]
-pub mod old_network;
-/// TODO: the whole http module need to be moved to `net`
+// pub mod old_http;
+// #[deprecated(note = "Use `Transport` instead")]
+// pub mod old_network;
```
akoshelev, line 5:
```suggestion
```

src/http/server/handlers/create_query.rs
```diff
@@ -1,22 +1,19 @@
...
 };
 use axum::{routing::post, Extension, Json, Router};
 use hyper::{Body, Request};
 use tokio::sync::{mpsc, oneshot};
 
-#[cfg_attr(feature = "enable-serde", derive(serde::Deserialize))]
-struct CreateQueryBody {
-    helper_positions: [HelperIdentity; 3],
-}
+// TODO: do we need this?
```
akoshelev, line 12: if it is not used, probably not?

akoshelev:
can you move `http` module to `net`?

src/http/server/handlers/query_input.rs
```diff
@@ -34,9 +33,9 @@ async fn handler(
...
     let input_stream = RequestParts::new(req)
-        .extract::<ByteArrStream>()
-        .await
-        .map_err(|_| Error::BadQueryString("TODO: move ByteArrStream to the right package".into()))?
+        .extract::<ByteArrStreamFromReq>()
+        .await?
+        .0
         .and_then(|bytes| futures::future::ok(bytes.to_vec()))
         .map_err(TransportError::from);
     let query_input = QueryInput {
```
akoshelev, line 41: I think we need another todo here. `FieldType` should not be sent twice per query. RC initiates the query telling us what field type to use and then submits the input assuming we know the field type already

akoshelev (changes requested):

src/http/server/handlers/query_input.rs
thurstonsand: right. this is one of the points i wanted to discuss over VC. i have a proposed solution and wanted to pass it by you

src/http/server/handlers/create_query.rs
thurstonsand: ok. it will be easy enough to add back if we ever do need it

thurstonsand:
i can move `http` to `net`, but let me do that in the last PR in the stack: https://github.com/thurstonsand/raw-ipa/pull/12 since I have made changes to the `http` folder along the way. easier to resolve diff

src/http/server/handlers/query_input.rs
thurstonsand: this field has been removed in the last PR: https://github.com/thurstonsand/raw-ipa/pull/12 (see commit: https://github.com/thurstonsand/raw-ipa/pull/12/commits/dd1fac7d5fe15ca8d00373b45193fb656226c779)

PR #393, opened 2023-01-13
Sanitize features in cargo.toml and update benches readme

akoshelev:
Initially I just wanted to update the readme file to show folks how to print per-step level metrics in benchmarks. After I started working on it, I realized it is not possible in the current cargo workspace configuration because debug spans are removed at compile time.

I thought I can just fix it with a one liner, but that turned into an exercise to clean up our existing code that does not protect imports behind feature gates. 

What I am proposing here is less than ideal, I would really want to move http related stuff into its own crate after looking at build timings

Screenshot 2023-01-13 at 3 08 27 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,14 +5,18 @@ rust-version = "1.64.0"
 edition = "2021"
 
 [features]
-default = ["debug", "cli"]
-cli = ["comfy-table", "clap", "enable-serde", "metrics-tracing-context", "metrics-util", "tracing-subscriber", "web-app"]
-debug = []
+# by default remove all TRACE, DEBUG spans from release builds
+default = ["web-app", "tracing/max_level_trace", "tracing/release_max_level_info"]
+cli = ["comfy-table", "clap"]
 enable-serde = ["serde", "serde_json"]
-web-app = ["tokio", "tokio-stream", "tokio-util", "axum", "axum-server", "hyper", "hyper-tls", "tower", "tower-http"]
+# TODO move web-app to a separate crate. It adds a lot of build time to people who mostly write protocols
+# TODO Consider moving out benches as well
+web-app = ["axum", "axum-server", "enable-serde", "hex", "hyper", "hyper-tls", "tower", "tower-http"]
 self-signed-certs = ["hyper-tls"]
 test-fixture = []
 shuttle = ["shuttle-crate", "test-fixture"]
+debug-trace = ["tracing/max_level_trace", "tracing/release_max_level_debug"]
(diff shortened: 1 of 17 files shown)


martinthomson:
How often do people working on the protocols need to worry about the build time of the other stuff?

richajaindce:
> How often do people working on the protocols need to worry about the build time of the other stuff?

I think now that we will be doing performance benchmarking, we would need to know if the bottleneck is in protocol implementation or in infrastructure. Can this information provide clarity on that?

akoshelev:
> How often do people working on the protocols need to worry about the build time of the other stuff?

I don't think they ever want to build http server, most of the tests we run as part of dev cycle are unit tests and touch Infra only

martinthomson:
I guess the question is really how much the added dependencies add to build times when you are running an incremental compilation.

akoshelev:
> I guess the question is really how much the added dependencies add to build times when you are running an incremental compilation.

that's a good point. I am seeing more full rebuilds than I expected (after every git pull for instance) but I can live with that. 

this change does not need that though, it is just feature cleanup + readme

benches/README.md
```diff
@@ -7,17 +7,57 @@ By convention, Criterion benchmarks have prefix `ct`, iai benchmark names start
 To execute a benchmark, run
 
 ```bash
-cargo bench -F test-fixture --bench <benchmark_name>
+cargo bench -F enable-benches --bench <benchmark_name>
```
martinthomson, line 10: I find it annoying that `cargo bench --bench foo` doesn't change the build configuration so that the necessary features are turned on.

Cargo.toml
```diff
@@ -23,16 +27,18 @@ bitvec = "1.0.1"
 clap = { version = "4.0.10", optional = true, features = ["derive"] }
 comfy-table = { version = "6.0.0", optional = true }
 config = "0.13.2"
+criterion = { version = "0.4.0", optional = true, default-features = false, features = ["async_tokio", "plotters", "html_reports"] }
```
martinthomson, line 30: New? or just added explicitly so that you can limit the features?

martinthomson (approved):
Ship it.

Cargo.toml
akoshelev: yea I moved it to dependencies so I can make it optional. dev-dependencies cannot be optional for some reason and we don't really need them for regular builds (just for benchmarks)

benches/README.md
akoshelev: yea, I also wish I could specify default features for profiles.

PR #394, opened 2023-01-14
Ordering MPSC

martinthomson:
This creates a channel that accepts inputs in any order, then reorders them, using a buffer.  It only accepts each item once (panicking if a duplicate is sent) and it will block (asynchronously) if the value doesn't currently fit in the buffer.

Hopefully, this will be useful for our sending side.  We'll need a store of these and a way to hook them up to the connection.

### DIFF
--- a/src/helpers/buffers/fsv.rs
+++ b/src/helpers/buffers/fsv.rs
@@ -2,6 +2,8 @@ use bitvec::bitvec;
 use bitvec::prelude::BitVec;
 use std::fmt::Debug;
 use std::num::NonZeroUsize;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering::{AcqRel, Acquire};
 
 /// A store of bytes that allows for random access inserts, but contiguous removal.
 ///
@@ -18,7 +20,7 @@ pub struct FixedSizeByteVec<const N: usize> {
     data: Vec<u8>,
     added: BitVec,
     capacity: NonZeroUsize,
-    end: usize,
+    end: AtomicUsize,
 }
 
 impl<const N: usize> FixedSizeByteVec<N> {
@@ -29,7 +31,7 @@ impl<const N: usize> FixedSizeByteVec<N> {
             data: vec![0_u8; N * capacity.get()],
             added: bitvec![0; capacity.get()],
             capacity,
-            end: capacity.get(),
+            end: AtomicUsize::new(capacity.get()),
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/helpers/buffers/ordering_mpsc.rs
```diff
@@ -0,0 +1,418 @@
...
+        self.end.incr(tail + wrap);
+        self.added[i..(i + tail)].fill(false);
+        if wrap > 0 {
+            self.added[..wrap].fill(false);
+            let mut buf = Vec::with_capacity((tail + wrap) * M::SIZE_IN_BYTES);
+            buf.extend_from_slice(&self.buf[(i * M::SIZE_IN_BYTES)..]);
+            buf.extend_from_slice(&self.buf[..(wrap * M::SIZE_IN_BYTES)]);
+            Some(buf)
+        } else {
+            Some(self.buf[(i * M::SIZE_IN_BYTES)..((i + tail) * M::SIZE_IN_BYTES)].to_owned())
```
akoshelev, line 135: not for this change, but this looks like a good use case for `Bytes` (shared pointer to a slice of [u8])`?

src/helpers/messaging.rs
```diff
@@ -35,13 +35,13 @@ use shuttle::future as tokio;
 /// Trait for messages sent between helpers
 pub trait Message: Debug + Send + Sized + 'static {
     /// Required number of bytes to store this message on disk/network
-    const SIZE_IN_BYTES: u32;
+    const SIZE_IN_BYTES: usize;
```
akoshelev, line 38: usize: 10
alex: 0

src/helpers/buffers/fsv.rs
```diff
@@ -29,7 +31,7 @@ impl<const N: usize> FixedSizeByteVec<N> {
             data: vec![0_u8; N * capacity.get()],
             added: bitvec![0; capacity.get()],
             capacity,
-            end: capacity.get(),
+            end: AtomicUsize::new(capacity.get()),
```
akoshelev, line 34: It looks like `ordering_mpsc` is a replacement for `fsv`, so I am wondering why we need this change?

src/helpers/buffers/ordering_mpsc.rs
```diff
@@ -0,0 +1,418 @@
...
+
+struct OrderingMpscEnd {
+    end: AtomicUsize,
+    notify: Notify,
+}
+
+/// A multi-producer, single-consumer channel that performs buffered reordering
+/// of inputs, with back pressure if the insertion is beyond the end of its buffer.
+/// This requires that each item be serializable and fixed size.
+pub fn ordering_mpsc<M: Message, S: AsRef<str>>(
```
akoshelev, line 44: do you want a shuttle test for it?

akoshelev (approved):

akoshelev:
I love the way it looks and how easy it is to understand what is going on.

src/helpers/buffers/fsv.rs
martinthomson: Ah, because I started by modifying fsv, and forgot to revert this, I missed than in my self-review.  Will revert.

src/helpers/buffers/ordering_mpsc.rs
martinthomson: Yeah, probably.  I'll get on to that.

src/helpers/buffers/ordering_mpsc.rs
martinthomson: `Bytes` is probably a little heavyweight for this application.  We don't need the refcounting and all that stuff in `Bytes`.

I almost had this return a slice, except for the wrapping part.  It might be possible to change the calling side to make that work, but that can come later, as you say.

akoshelev: oh nice, if we can eventually get to returning a slice, that's even better

PR #395, opened 2023-01-16
Feature turn-off-prss and protocol-perf-testing

richajaindce:
Introducing feature to switch off infrastructure pieces so that we can perf test and improve the IPA protocol.

This is under the feature protocol-perf-testing which switches off prss
To run bench using this feature do
 RUST_LOG=raw_ipa=debug cargo bench --bench <bench> --features="test-fixture protocol-perf-testing"

For the original sort, turning off PRSS brought the runtime down from 106secs to 96secs


src/protocol/prss.rs
```diff
@@ -4,11 +4,16 @@ use crate::rand::{CryptoRng, RngCore};
 use crate::secret_sharing::Replicated;
 use crate::sync::{Arc, Mutex};
 
+#[cfg(not(feature = "turn-off-prss"))]
```
martinthomson, line 7: Rather than puncturing this code, it might be easier to build a parallel file that you include like this:

```rust
#[cfg(feature = "no-prss")]
mod prss_zero;
#[cfg(feature = "no-prss")]
pub use prss_zero::{...};

#[cfg(not(feature = "no-prss"))]
mod prss;
#[cfg(not(feature = "no-prss"))]
pub use prss::{...};
```

There will be a bit of duplication, but I think that this would be easier to read and manage.

martinthomson (commented):
I'm always uncertain about conditional compilation and its effect on code rotting, but I think we're OK here.  Anyone who finds this code to be rotten can fix it trivially.

src/protocol/prss.rs
richajaindce: makes sense. let me do this

src/protocol/no_prss.rs
```diff
@@ -0,0 +1,102 @@
+use crate::ff::Field;
```
akoshelev, line 1: lets update the comments to show that this is a no-op PRSS

src/protocol/mod.rs
```diff
@@ -5,8 +5,12 @@ pub mod context;
 pub mod ipa;
 pub mod malicious;
 pub mod modulus_conversion;
+#[cfg(feature = "no-prss")]
```
akoshelev, line 8: I am a bit surprised that you need to re-export those. I thought having `prss` module exported would be enough.

src/protocol/no_prss.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// The key exchange component of a participant.
+pub struct KeyExchange {
+    sk: EphemeralSecret,
+}
+
+impl KeyExchange {
+    pub fn new<R: RngCore + CryptoRng>(r: &mut R) -> Self {
+        Self {
+            sk: EphemeralSecret::new(r),
```
akoshelev, line 68: I don't think you need a secret for no-op PRSS. `KeyExchange` can just return a fixed public key (you can create it from [0_u8; 32] and `GeneratorFactory` is a no-op already.

src/protocol/no_prss.rs
```diff
@@ -0,0 +1,102 @@
...
+        Generator {}
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Generator {}
+/// The basic generator.  This generates values based on an arbitrary index.
+impl Generator {
+    /// Generate the value at the given index.
+    /// This uses the MMO^{\pi} function described in <https://eprint.iacr.org/2019/074>.
```
akoshelev, line 97:
```suggestion

```

akoshelev (changes requested):

src/protocol/no_prss.rs
richajaindce: Makes sense. I was just scared of touching that code. Done now

src/protocol/mod.rs
richajaindce: Removing "pub" here but we would still need to re-export the traits/functions defined in no_prss or use_prss since they are not exposed in prss.rs anymore right? i.e. line 2-4 in prss.rs

akoshelev: yep, removing `pub` is what I meant. It is still looks a bit upside down: `prss` module conditionally re-exports functions from parent submodules.

would it be cleaner if we move `prss` to its own folder inside the protocol? 

```
-> protocol
    -> prss
          -> mod.rs // conditionally re-exports either no-op PRSS or actual PRSS
          -> no_op.rs
          -> crypto.rs 
```

richajaindce: Done

akoshelev (approved):
looks great! thanks for working on making it more structured!

src/protocol/prss/crypto.rs
```diff
@@ -0,0 +1,147 @@
...
+pub trait SharedRandomness {
+    /// Generate two random values, one that is known to the left helper
+    /// and one that is known to the right helper.
+    #[must_use]
+    fn generate_values<I: Into<u128>>(&self, index: I) -> (u128, u128);
+
+    /// Generate two random field values, one that is known to the left helper
+    /// and one that is known to the right helper.
+    #[must_use]
+    fn generate_fields<F: Field, I: Into<u128>>(&self, index: I) -> (F, F) {
```
martinthomson, line 22: I had a thought...  Will execute on it shortly.

martinthomson (approved):

PR #399, opened 2023-01-18
Implement `Ord` for BitArray types

taikiy:
The original implementation of `Ord` for `bitvec::BitArray` compares two arrays from LSB, and at the first index where the arrays differ, the array with the high bit is greater.

For our use case, however, we want to compare two arrays by their integer values represented by the bits. In other words, if `a < b` is true, then `BitArrayN::try_from(a).unwrap() < BitArrayN::try_from(b).unwrap()` must also be true.

### DIFF
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -101,26 +101,20 @@ macro_rules! bit_array_impl {
                 }
             }
 
-            impl TryFrom<$name> for u128 {
-                type Error = String;
-
-                /// Fallible conversion from this data type to `u128`. The inner value must be
-                /// at most `Self::BITS` long. That is, the integer value must be less than
-                /// or equal to `2^Self::BITS`, or it will return an error.
-                fn try_from(value: $name) -> Result<Self, Self::Error> {
-                    if <$name>::BITS <= 128 {
-                        Ok(value
-                            .0
-                            .iter()
-                            .by_refs()
-                            .enumerate()
-                            .fold(0_u128, |acc, (i, b)| acc + ((*b as u128) << i)))
-                    } else {
-                        Err(format!(
-                            "Bit array size {} is too large to convert to u128.",
-                            Self::BITS,
-                        ))
-                    }
+            #[allow(clippy::from_over_into)]
+            impl Into<u128> for $name {
+                /// Infallible conversion from this data type to `u128`. We assume that the
+                /// inner value is at most 128-bit long. That is, the integer value must be
+                /// less than or equal to `2^Self::BITS`. Should be long enough for our use
+                /// case.
+                fn into(self) -> u128 {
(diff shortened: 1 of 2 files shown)


src/bits/bit_array.rs
```diff
@@ -117,6 +140,27 @@ macro_rules! bit_array_impl {
...
             }
 
+            /// Compares two `BitArray`s by their representational ordering
+            ///
+            /// The original implementation of `Ord` for `bitvec::BitArray` compares two arrays
+            /// from LSB, and at the first index where the arrays differ, the array with the high
+            /// bit is greater. For our use case, however, we want to compare two arrays by their
+            /// integer values represented by the bits. In other words, if `a < b` is true, then
+            /// `BitArray::try_from(a).unwrap() < BitArray::try_from(b).unwrap()` must also be true.
+            impl Ord for $name {
```
akoshelev, line 150: do we ever use bit arrays that store more than 128 bits? if we do, panicking in `Ord` is not the best practice imo.
It does not seem that we allow constructing bitarrays from anything expect a `u128`, so maybe we can define `TryFrom/Into` for u128 instead of `TryFrom/TryInto` and getting rid of `unwrap` here?

taikiy: Sounds good. I think we are okay with constraining to <= 128 bits for now.

taikiy:
@akoshelev, let me know if it looks good. Would be great if you could merge it once accepted. Thanks!

src/bits/bit_array.rs
```diff
@@ -117,6 +134,25 @@ macro_rules! bit_array_impl {
                 }
             }
 
+            /// Compares two `BitArray`s by their representational ordering
```
akoshelev, line 137: I would also suggest to follow up and move this doc to the struct definition. Basically what we're saying here is that bit array uses LSB order to store elements and MSB order to compare two arrays

akoshelev (approved):

src/bits/bit_array.rs
```diff
@@ -179,6 +215,28 @@ macro_rules! bit_array_impl {
...
+                }
+
+                #[test]
+                pub fn ordering() {
+                    let mut rng = thread_rng();
+                    let a = rng.gen::<u128>() & MASK;
+                    let b = rng.gen::<u128>() & MASK;
+
+                    println!("a: {a}");
+                    println!("b: {b}");
```
akoshelev, line 236:
```suggestion
```

akoshelev:
> @akoshelev, let me know if it looks good. Would be great if you could merge it once accepted. Thanks!

yea apologies for missing the update yesterday. Let me know if you want to merge it as is and follow up on minor feedback or you prefer to address comments in this PR.

benjaminsavage:
lets merge and fix additional issues in more PRs

PR #402, opened 2023-01-20
Reverting #379

benjaminsavage:
As it turns out, the only time we ever compute the dot-product of two vectors, we independently have them in memory as two vectors. So it's better to just use them as-is, rather than creating a new vector of pairs of them.

### DIFF
--- a/src/protocol/sort/multi_bit_permutation.rs
+++ b/src/protocol/sort/multi_bit_permutation.rs
@@ -61,18 +61,22 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
     }
 
     // Take sum of products of output of equality check and accumulated sum
-    let mut one_off_permutation =
-        try_join_all((0..num_records).zip(repeat(ctx)).map(|(rec, ctx)| {
-            let mut sop_inputs = Vec::with_capacity(num_possible_bit_values);
-            for idx in 0..num_possible_bit_values {
-                sop_inputs.push((&equality_checks[rec][idx], &prefix_sum[rec][idx]));
-            }
-            async move {
-                ctx.sum_of_products(RecordId::from(rec), sop_inputs.as_slice())
-                    .await
-            }
-        }))
-        .await?;
+    let mut one_off_permutation = try_join_all(
+        equality_checks
+            .into_iter()
+            .zip(prefix_sum.into_iter())
+            .zip(repeat(ctx))
+            .enumerate()
+            .map(|(i, ((eq_checks, prefix_sums), ctx))| async move {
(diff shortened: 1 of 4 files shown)


src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -61,18 +61,22 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
     }
 
     // Take sum of products of output of equality check and accumulated sum
-    let mut one_off_permutation =
-        try_join_all((0..num_records).zip(repeat(ctx)).map(|(rec, ctx)| {
-            let mut sop_inputs = Vec::with_capacity(num_possible_bit_values);
-            for idx in 0..num_possible_bit_values {
-                sop_inputs.push((&equality_checks[rec][idx], &prefix_sum[rec][idx]));
-            }
```
benjaminsavage, line 69: All of this was eliminated. No need to create a new vector for every row, and push a bunch of pairs into it.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -61,18 +61,22 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
...
+        equality_checks
+            .into_iter()
+            .zip(prefix_sum.into_iter())
+            .zip(repeat(ctx))
+            .enumerate()
+            .map(|(i, ((eq_checks, prefix_sums), ctx))| async move {
+                ctx.sum_of_products(
+                    RecordId::from(i),
+                    eq_checks.iter().collect::<Vec<_>>().as_slice(),
+                    prefix_sums.iter().collect::<Vec<_>>().as_slice(),
```
benjaminsavage, line 74: This is stupid.

This is here because `sum_of_products` really wants to accept a slice of *references* to shares... what we have is vectors of *shares*. So this converts them into vectors of *references*...

I attempted to eliminate this, and make `sum_of_products` just accept a slice of shares... but unfortunately, while this is easy for the semi-honest one, this isn't easy for the malicious version. That's because it attempts to call the semi-honest version, but it doesn't *have* vectors of replicated shares - it has vectors of *malicious replicated*...

I couldn't fix this without copy-pasting a lot of code. We can return to this later. There is another optimisation waiting here.

richajaindce: Lets open an issue for this so we address it soon

richajaindce: #403

richajaindce (approved):
Looks good!

Issue #403, opened 2023-01-20
sum_of_products should accept slice of shares and not slice of reference of shares


richajaindce:
This is stupid.

This is here because `sum_of_products` really wants to accept a slice of *references* to shares... what we have is vectors of *shares*. So this converts them into vectors of *references*...

I attempted to eliminate this, and make `sum_of_products` just accept a slice of shares... but unfortunately, while this is easy for the semi-honest one, this isn't easy for the malicious version. That's because it attempts to call the semi-honest version, but it doesn't *have* vectors of replicated shares - it has vectors of *malicious replicated*...

I couldn't fix this without copy-pasting a lot of code. We can return to this later. There is another optimisation waiting here.

_Originally posted by @benjaminsavage in https://github.com/private-attribution/ipa/pull/402#discussion_r1082174365_

PR #404, opened 2023-01-20
Attempt to remove needless iterators

benjaminsavage:
This is an attempt to address issue #403. I don't like it. There's a lot of duplicate code here.

### DIFF
--- a/src/protocol/basics/sum_of_product/malicious.rs
+++ b/src/protocol/basics/sum_of_product/malicious.rs
@@ -1,11 +1,13 @@
 use crate::error::Error;
 use crate::ff::Field;
-use crate::protocol::basics::sum_of_product::SecureSop;
+use crate::helpers::Direction;
+use crate::protocol::prss::SharedRandomness;
 use crate::protocol::{
     context::{Context, MaliciousContext},
     RecordId,
 };
 use crate::secret_sharing::replicated::malicious::AdditiveShare as MaliciousReplicated;
+use crate::secret_sharing::replicated::semi_honest::AdditiveShare as Replicated;
 use futures::future::try_join;
 use std::fmt::Debug;
 
@@ -60,8 +62,8 @@ impl AsRef<str> for Step {
 pub async fn sum_of_products<F>(
     ctx: MaliciousContext<'_, F>,
     record_id: RecordId,
-    a: &[&MaliciousReplicated<F>],
-    b: &[&MaliciousReplicated<F>],
+    a: &[MaliciousReplicated<F>],
+    b: &[MaliciousReplicated<F>],
 ) -> Result<MaliciousReplicated<F>, Error>
 where
     F: Field,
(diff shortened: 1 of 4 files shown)


src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -127,10 +151,8 @@ mod test {
         }
 
         let res = world
-            .malicious((av, bv), |ctx, (a_share, b_share)| async move {
-                let a_refs = a_share.iter().collect::<Vec<_>>();
-                let b_refs = b_share.iter().collect::<Vec<_>>();
```
benjaminsavage, line 132: This absurdity can be eliminated.

src/protocol/basics/sum_of_product/semi_honest.rs
```diff
@@ -117,9 +117,7 @@ mod test {
 
         let res = world
             .semi_honest((av, bv), |ctx, (a, b)| async move {
-                let a_refs = a.iter().collect::<Vec<_>>();
-                let b_refs = b.iter().collect::<Vec<_>>();
```
benjaminsavage, line 121: This absurdity can be eliminated.

src/protocol/basics/sum_of_product/semi_honest.rs
```diff
@@ -137,11 +135,8 @@ mod test {
         let b: Vec<_> = b.iter().map(|x| Fp31::from(*x)).collect();
 
         let result = world
-            .semi_honest((a, b), |ctx, (a_share, b_share)| async move {
-                let a_refs = a_share.iter().collect::<Vec<_>>();
-                let b_refs = b_share.iter().collect::<Vec<_>>();
```
benjaminsavage, line 142: This absurdity can be eliminated.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -70,8 +70,8 @@ pub async fn multi_bit_permutation<'a, F: Field, S: SecretSharing<F>, C: Context
             .map(|(i, ((eq_checks, prefix_sums), ctx))| async move {
                 ctx.sum_of_products(
                     RecordId::from(i),
-                    eq_checks.iter().collect::<Vec<_>>().as_slice(),
-                    prefix_sums.iter().collect::<Vec<_>>().as_slice(),
```
benjaminsavage, line 74: And here is the main reason I'm doing this: to eliminate this wastefulness. There is really no need to map these into new vectors of references...

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -70,30 +72,52 @@ where
...
+        right_sops_m +=
+            arx.right() * bx.right() + arx.left() * bx.right() + arx.right() * bx.left();
+    }
+
+    // notify helper on the right that we've computed our value
+    let channel = ctx.mesh();
+    let channel_m = duplicate_multiply_ctx.mesh();
+    try_join(
+        channel.send(role.peer(Direction::Right), record_id, right_sops),
+        channel_m.send(role.peer(Direction::Right), record_id, right_sops_m),
```
benjaminsavage, line 104: Note that this is pretty just all copy-pasted from src/protocol/basics/sum_of_product/semi_honest.rs

...I don't like that...

akoshelev: yea we need to do something about it. I know that Richa is having the same issue, trying to implement a protocol that is *almost* the same for semi-honest and malicious. I added this to our Thursday agenda sync

akoshelev (approved):
nice work! not just the code looks better, it is also more efficient

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -60,8 +62,8 @@ impl AsRef<str> for Step {
 pub async fn sum_of_products<F>(
     ctx: MaliciousContext<'_, F>,
     record_id: RecordId,
-    a: &[&MaliciousReplicated<F>],
-    b: &[&MaliciousReplicated<F>],
+    a: &[MaliciousReplicated<F>],
+    b: &[MaliciousReplicated<F>],
```
martinthomson, line 66: Looking at this, why can't `sum_of_products()` take `impl IntoIterator` so that we don't even need to `collect()` at the call site?

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -127,10 +151,8 @@ mod test {
         }
 
         let res = world
-            .malicious((av, bv), |ctx, (a_share, b_share)| async move {
-                let a_refs = a_share.iter().collect::<Vec<_>>();
-                let b_refs = b_share.iter().collect::<Vec<_>>();
-                ctx.sum_of_products(RecordId::from(0), a_refs.as_slice(), b_refs.as_slice())
+            .malicious((av, bv), |ctx, (a, b)| async move {
+                ctx.sum_of_products(RecordId::from(0), a.as_slice(), b.as_slice())
```
martinthomson, line 155: Note that you can do `a.as_slice()` with `&a` if `a` is a `Vec`.

martinthomson (commented):
We can maybe do better yet.

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -70,30 +72,52 @@ where
...
+    let mut right_sops: F = s1 - s0;
+    let mut right_sops_m: F = s1_m - s0_m;
+
+    for i in 0..vec_len {
+        let ax = a[i].x().access_without_downgrade();
+        let bx = b[i].x().access_without_downgrade();
+        let arx = a[i].rx();
+        right_sops += ax.right() * bx.right() + ax.left() * bx.right() + ax.right() * bx.left();
+        right_sops_m +=
+            arx.right() * bx.right() + arx.left() * bx.right() + arx.right() * bx.left();
```
martinthomson, line 96: Do we know whether this would be faster?

```rust
right_sops += ax.right() * (bx.left() + bx.right()) + ax.left() * bx.right();
```

It seems like it could, depending on how good the compiler is at modulo math.

PR #405, opened 2023-01-20
Secure apply inv now works on multi bits

richajaindce:
This is a building block to support multi bit permutation to work on multi bit slices at a time


benjaminsavage (approved):
Works for me!

PR #406, opened 2023-01-20
ByteArrStream now returns aligned chunks

thurstonsand:
reopened from https://github.com/thurstonsand/raw-ipa/pull/11.

Previously, ByteArrStream would strictly return size_in_bytes len chunks. Now, it will return the largest possible chunk that is aligned with size_in_bytes.

This is currently causing the e2e test to fail due to some expectations there. that will be fixed in an upcoming PR.

Also correctly passes QueryType::IPA


src/http/server/handlers/mod.rs
```diff
@@ -58,6 +73,31 @@ impl<B: Send> FromRequest<B> for QueryConfigFromReq {
...
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        #[derive(serde::Deserialize)]
+        struct FieldTypeParam {
+            field_type: FieldType,
+        }
+
+        let Query(FieldTypeParam { field_type }) = req.extract().await?;
+        let body: BodyStream = req.extract().await?;
+        // TODO: multiply the field_type by 22; this is a hacky way of making sure IPA can run,
```
akoshelev, line 93: can you follow up and make a small change here?

this todo is very important to address, so I am thinking that we should also print a warning each time this code is executed

akoshelev (approved):

src/http/server/handlers/mod.rs
thurstonsand: following up to say this was fixed

PR #408, opened 2023-01-20
Merge main

thurstonsand:


src/net/server/mod.rs
```diff
@@ -163,3 +163,256 @@ ShF2TD9MWOlghJSEC6+W3nModkc=
 
     RustlsConfig::from_pem(cert.as_bytes().to_vec(), key.as_bytes().to_vec()).await
 }
+
```
thurstonsand, line 166: will address in follow-up PR

src/secret_sharing/mod.rs
```diff
@@ -32,9 +51,12 @@ pub trait SecretSharing<F>:
     const ZERO: Self;
 }
 
-impl<F: Field> SecretSharing<F> for Replicated<F> {
-    const ZERO: Self = Replicated::ZERO;
+/// TODO: why did I need this change?
```
thurstonsand, line 54: I had to add `crate::ff::Field` here, and I'm not sure why

src/secret_sharing/mod.rs
```diff
@@ -32,9 +51,12 @@ pub trait SecretSharing<F>:
...
+impl<V: ArithmeticShare + crate::ff::Field> SecretSharing<V> for SemiHonestAdditiveShare<V> {
+    const ZERO: Self = SemiHonestAdditiveShare::ZERO;
 }
-impl<F: Field> SecretSharing<F> for MaliciousReplicated<F> {
-    const ZERO: Self = MaliciousReplicated::ZERO;
+
+/// TODO: why did I need this change?
```
thurstonsand, line 59: I had to add `crate::ff::Field` here, and I'm not sure why

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -46,55 +45,62 @@ pub trait ThisCodeIsAuthorizedToDowngradeFromMalicious<T> {
...
     }
 }
 
-impl<F: Field> MaliciousReplicated<F> {
-    pub const ZERO: Self = Self {
-        x: Replicated::ZERO,
-        rx: Replicated::ZERO,
-    };
-
+/// TODO: why did I need this change?
```
thurstonsand, line 64: I had to add `+ Field` here, and I'm not sure why

src/secret_sharing/mod.rs
thurstonsand: is it possible to implement `AdditiveShare` without `F: Field`?

thurstonsand: spoke with @taikiy, and we actually do need `Field`

src/query/executor.rs
```diff
@@ -141,8 +141,10 @@ pub fn start_query(
                         Box::new(execute_test_multiply(ctx, input).await) as Box<dyn Result>
                     }
                     QueryType::IPA(config) => {
-                        let input = input.align(IPAInputRow::<Fp31>::SIZE_IN_BYTES);
-                        Box::new(execute_ipa(ctx, config, input).await) as Box<dyn Result>
+                        let input = input.align(IPAInputRow::<Fp31, BitArray40>::SIZE_IN_BYTES);
+                        // TODO: how to know `BitArray` type?
```
thurstonsand, line 145: do we need to add another value to `IPAQueryConfig` here?

thurstonsand: if so, get ready for exponential match statements...

thurstonsand: looks like we will try to make a `MatchKey` alias to use in place of `BitArray40`

akoshelev:
I skimmed over it and I don't see anything bad but given the size of it it is likely that I could've missed something. 
Lets merge it as is and see what the diff with main reveals

thurstonsand:
it's going to be one _major_ diff with main...

PR #410, opened 2023-01-21
HPKE for match key encryption

akoshelev:
This commit provides basic HPKE functionality to decrypt match keys encrypted on UA side according with https://github.com/patcg-individual-drafts/ipa/pull/31.

### DIFF
--- a/src/crypto/hpke/aad.rs
+++ b/src/crypto/hpke/aad.rs
@@ -1,0 +1,99 @@
+use std::fmt::{Display, Formatter};
+
+use super::{Epoch, KeyIdentifier};
+
+const SALT: &str = "private-attribution";
+
+#[derive(Debug)]
+pub struct HelperOrigin<'a>(pub &'a str);
+#[derive(Debug)]
+pub struct SiteOrigin<'a>(pub &'a str);
+
+/// Represents the [`info`] part of the receiver context, that is: application specific data
+/// for each encryption.
+///
+/// IPA uses key identifier, event epoch and both helper and site origins to authenticate match key
+/// encryptions. It is not guaranteed that the same receiver can be used for anything else.
+///
+/// [`info`]: https://www.rfc-editor.org/rfc/rfc9180.html#name-creating-the-encryption-con
+pub struct AssociatedData<'a> {
+    key_id: KeyIdentifier,
+    epoch: Epoch,
+    helper_origin: &'a HelperOrigin<'a>,
+    site_origin: &'a SiteOrigin<'a>,
+}
+
+#[derive(Debug)]
+pub struct NonAsciiStringError<'a> {
+    input: &'a str,
+}
+
+impl Display for NonAsciiStringError<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "string contains non-ascii symbols: {}", self.input)
+    }
+}
+
+impl<'a> From<&'a str> for NonAsciiStringError<'a> {
+    fn from(input: &'a str) -> Self {
+        Self { input }
+    }
+}
+
+impl<'a> AssociatedData<'a> {
+    /// Creates new instance
+    ///
+    /// ## Errors
+    /// if helper or site origin is not a valid ASCII string.
+    pub fn new(
+        key_id: KeyIdentifier,
+        epoch: Epoch,
+        helper_origin: &'a HelperOrigin<'a>,
+        site_origin: &'a SiteOrigin<'a>,
+    ) -> Result<Self, NonAsciiStringError<'a>> {
+        if !helper_origin.0.is_ascii() {
+            return Err(helper_origin.0.into());
+        }
+
+        if !site_origin.0.is_ascii() {
+            return Err(site_origin.0.into());
+        }
+
+        Ok(Self {
+            key_id,
+            epoch,
+            helper_origin,
+            site_origin,
+        })
+    }
+
+    pub(super) fn key_id(&self) -> KeyIdentifier {
+        self.key_id
+    }
+
+    pub(super) fn epoch(&self) -> Epoch {
+        self.epoch
+    }
+
+    pub(super) fn to_bytes(&self) -> Box<[u8]> {
+        let info_len = SALT.len()
+            + 2 // account for 2 delimiters
+            + self.helper_origin.0.len()
+            + self.site_origin.0.len()
+            + std::mem::size_of_val(&self.key_id)
+            + std::mem::size_of_val(&self.epoch);
+        let mut r = Vec::with_capacity(info_len);
+
+        r.extend_from_slice(SALT.as_bytes());
+        r.extend_from_slice(self.helper_origin.0.as_bytes());
+        r.push(b'\0');
+        r.extend_from_slice(self.site_origin.0.as_bytes());
+        r.push(b'\0');
+
+        r.push(self.key_id);
+        // Spec dictates epoch to be encoded in BE
+        r.extend_from_slice(&self.epoch.to_be_bytes());
+
+        r.into_boxed_slice()
+    }
+}
--- a/src/crypto/hpke/aad.rs
+++ b/src/crypto/hpke/aad.rs
@@ -63,21 +63,13 @@ impl<'a> Info<'a> {
         })
     }
 
-    pub(super) fn key_id(&self) -> KeyIdentifier {
-        self.key_id
-    }
-
-    pub(super) fn epoch(&self) -> Epoch {
-        self.epoch
-    }
-
     /// Converts this instance into an owned byte slice that can further be used to create HPKE
     /// sender or receiver context.
     pub(super) fn into_bytes(self) -> Box<[u8]> {
         let info_len = DOMAIN.len()
-            + 3 // account for 3 delimiters
             + self.helper_origin.len()
             + self.site_origin.len()
+            + 3 // account for 3 delimiters
             + std::mem::size_of_val(&self.key_id)
             + std::mem::size_of_val(&self.epoch);
         let mut r = Vec::with_capacity(info_len);
--- a/src/crypto/hpke/aad.rs
+++ b/src/crypto/hpke/aad.rs
@@ -2,12 +2,7 @@ use std::fmt::{Display, Formatter};
 
 use super::{Epoch, KeyIdentifier};
 
-const SALT: &str = "private-attribution";
-
-#[derive(Debug)]
-pub struct HelperOrigin<'a>(pub &'a str);
-#[derive(Debug)]
-pub struct SiteOrigin<'a>(pub &'a str);
+const DOMAIN: &str = "private-attribution";
 
 /// Represents the [`info`] part of the receiver context, that is: application specific data
 /// for each encryption.
@@ -16,11 +11,11 @@ pub struct SiteOrigin<'a>(pub &'a str);
 /// encryptions. It is not guaranteed that the same receiver can be used for anything else.
 ///
 /// [`info`]: https://www.rfc-editor.org/rfc/rfc9180.html#name-creating-the-encryption-con
-pub struct AssociatedData<'a> {
+pub struct Info<'a> {
     key_id: KeyIdentifier,
     epoch: Epoch,
-    helper_origin: &'a HelperOrigin<'a>,
-    site_origin: &'a SiteOrigin<'a>,
+    helper_origin: &'a str,
+    site_origin: &'a str,
 }
 
 #[derive(Debug)]
@@ -40,7 +35,7 @@ impl<'a> From<&'a str> for NonAsciiStringError<'a> {
     }
 }
 
-impl<'a> AssociatedData<'a> {
+impl<'a> Info<'a> {
     /// Creates new instance
     ///
     /// ## Errors
@@ -48,15 +43,15 @@ impl<'a> AssociatedData<'a> {
     pub fn new(
         key_id: KeyIdentifier,
         epoch: Epoch,
-        helper_origin: &'a HelperOrigin<'a>,
-        site_origin: &'a SiteOrigin<'a>,
+        helper_origin: &'a str,
+        site_origin: &'a str,
     ) -> Result<Self, NonAsciiStringError<'a>> {
-        if !helper_origin.0.is_ascii() {
-            return Err(helper_origin.0.into());
+        if !helper_origin.is_ascii() {
+            return Err(helper_origin.into());
         }
 
-        if !site_origin.0.is_ascii() {
-            return Err(site_origin.0.into());
+        if !site_origin.is_ascii() {
+            return Err(site_origin.into());
         }
 
         Ok(Self {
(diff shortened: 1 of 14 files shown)


src/crypto/hpke/aad.rs
```diff
@@ -0,0 +1,99 @@
...
+            + 2 // account for 2 delimiters
+            + self.helper_origin.0.len()
+            + self.site_origin.0.len()
+            + std::mem::size_of_val(&self.key_id)
+            + std::mem::size_of_val(&self.epoch);
+        let mut r = Vec::with_capacity(info_len);
+
+        r.extend_from_slice(SALT.as_bytes());
+        r.extend_from_slice(self.helper_origin.0.as_bytes());
+        r.push(b'\0');
```
martinthomson, line 89:
```suggestion
        r.push(0);
```

src/crypto/hpke/aad.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    pub(super) fn key_id(&self) -> KeyIdentifier {
+        self.key_id
+    }
+
+    pub(super) fn epoch(&self) -> Epoch {
+        self.epoch
+    }
+
+    pub(super) fn to_bytes(&self) -> Box<[u8]> {
```
martinthomson, line 78:
```suggestion
    pub(super) fn to_bytes(self) -> Box<[u8]> {
```
Any reason not to consume `self`?

There's a reason I suggest this.  It's possible that we could batch up decryption, but we'd want to have this operation done as few times as possible.  Taking a reference means that you are encouraging people to keep calling this function, when that is just busy work in the end.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// Opens ciphertext in place by first obtaining the secret key from [`key_registry`]
+/// using epoch and key from [`aad`] and then applying [`HPKE decryption`] to the provided ciphertext.
+///
+/// ## Errors
+/// If ciphertext cannot be opened for any reason.
+///
+/// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
+pub fn open_in_place(
+    key_registry: &KeyRegistry,
```
martinthomson, line 54: Is this a function of `KeyRegistry`?

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    encap_key: &[u8],
```
martinthomson, line 55: Spec uses 'enc' for this.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    ciphertext: &mut [u8],
+    tag: &[u8],
+    aad: &AssociatedData,
+) -> Result<(), BottomError> {
+    //TODO: log errors, but don't return them
+    let info = aad.to_bytes();
+    let encap_key = <IpaKem as hpke::Kem>::EncappedKey::from_bytes(encap_key)?;
+    let tag = AeadTag::<IpaAead>::from_bytes(tag)?;
+    let sk = key_registry.private_key(aad.epoch(), aad.key_id());
```
martinthomson, line 64: I think that we'd want to lookup by key_id only.  My expectation is that the one key is good for a couple of months.  And we don't want to make it too easy to use different keys (which is what a large key identifier space can do).

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+type IpaAead = hpke::aead::AesGcm128;
+type IpaKdf = hpke::kdf::HkdfSha256;
+
+pub type KeyIdentifier = u8;
+pub type Epoch = u16;
+type IpaPublicKey = <IpaKem as hpke::kem::Kem>::PublicKey;
+type IpaPrivateKey = <IpaKem as hpke::kem::Kem>::PrivateKey;
+
+#[derive(Debug)]
+pub struct BottomError;
```
martinthomson, line 26: Bottom?  DecryptionError or DecapsulationError perhaps.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// ## Errors
+/// If ciphertext cannot be opened for any reason.
+///
+/// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
+pub fn open_in_place(
+    key_registry: &KeyRegistry,
+    encap_key: &[u8],
+    ciphertext: &mut [u8],
+    tag: &[u8],
```
martinthomson, line 57: I know that the in-place function needs to have two slices for the ciphertext and tag, but my preference is that we do the split for them.  That is, we take `ciphertext: &mut [u8]` here and do:

```rust
let (ct, tag) = ciphertext.split_at_mut(ciphertext.len() - IpaAead::N_T);
```
(Though sadly the constants defined in the spec aren't exposed on the AEAD trait in the crate.  Maybe we can upstream something...; until then, a constant in this code (TAG_LEN) should do.)

Then you can return an immutable reference to `ct`.  Note that you need to use lifetime annotations so that the compiler gets it right.

src/crypto/hpke/aad.rs
```diff
@@ -0,0 +1,99 @@
...
+        let info_len = SALT.len()
+            + 2 // account for 2 delimiters
```
martinthomson, line 80: I updated the PR in response to feedback from @dominiccooney and we're putting a 0 at the end of the fixed string too.  (It might pay to make functions for strlen() and strencode() here anyway).

src/crypto/hpke/aad.rs
```diff
@@ -0,0 +1,99 @@
+use std::fmt::{Display, Formatter};
+
+use super::{Epoch, KeyIdentifier};
+
+const SALT: &str = "private-attribution";
```
martinthomson, line 5: This isn't really salt (salt is generally randomized), but a domain separation string.  That is, something that ensures that - if you use the same randomness for multiple application - you don't end up accidentally producing the same keys (which would be disastrous).

src/crypto/hpke/registry.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.key_pair(epoch, key_id).public_key()
+    }
+
+    fn key_pair(&self, epoch: Epoch, key_id: KeyIdentifier) -> &KeyPair {
+        self.keys
+            .get(&epoch)
+            .and_then(|keys| keys.get(key_id as usize))
+            .unwrap_or_else(|| {
+                panic!("no key registered for epoch '{epoch}' and key id '{key_id}'")
+            })
```
martinthomson, line 74: I don't think that it is a good idea to panic when a lookup fails.  This code is driven by the record collector, so we probably need this to return `Result<&KeyPair, _>` instead.

src/secret_sharing/xor.rs
```diff
@@ -1,10 +1,14 @@
...
+use std::io;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct XorReplicated {
     left: u64,
     right: u64,
 }
 
 impl XorReplicated {
+    pub const SIZE_IN_BYTES: usize = 2 * std::mem::size_of::<u64>();
```
martinthomson, line 10: I think that you want to look at Taiki's latest in #409 here.  And isn't this supposed to attach to an implementation of `Message`?

src/secret_sharing/xor.rs
```diff
@@ -19,4 +23,44 @@ impl XorReplicated {
...
+    /// if buffer capacity is not sufficient to hold all the bytes.
+    pub fn serialize(&self, buf: &mut [u8]) -> io::Result<usize> {
+        if buf.len() < Self::SIZE_IN_BYTES {
+            Err(io::Error::new(
+                io::ErrorKind::WriteZero,
+                "not enough buffer capacity",
+            ))
+        } else {
+            buf[..8].copy_from_slice(&self.left.to_le_bytes());
+            buf[8..16].copy_from_slice(&self.right.to_le_bytes());
```
martinthomson, line 61: Is there any reason you can't:

```suggestion
            buf[8..].copy_from_slice(&self.right.to_le_bytes());
```
... on the expectation that you will only be given SIZE_IN_BYTES bytes to write to?

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+        let match_key = XorReplicated::new(u64::MAX, u64::MAX / 2);
+
+        let enc = suite.seal(0, match_key);
+        let r = suite.open(0, enc).unwrap();
+
+        assert_eq!(match_key, r);
+    }
+
+    #[test]
+    fn decrypt_wrong_aad() {
```
martinthomson, line 193: Is this a case where you could do a proptest?

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    mod proptests {
+        use super::*;
+        use proptest::prelude::ProptestConfig;
+
+        proptest::proptest! {
+            #![proptest_config(ProptestConfig::with_cases(50))]
+            #[test]
+            fn arbitrary_corruption(bad_byte in 0..15_usize, bad_bit in 0..7_usize, seed: [u8; 32]) {
```
martinthomson, line 219: Corrupting the ciphertext and tag separately work here, but not if we change the size of the ciphertext.  I recommend building two tests.  See also above with respect to the info string.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+        let mut suite = EncryptionSuite::new(1, rng);
+        let match_key = XorReplicated::new(u64::MAX, u64::MAX / 2);
+        let enc = suite.seal(0, match_key);
+        suite.advance_epoch();
+
+        let _ = suite.open(0, enc).unwrap_err();
+    }
+
+    #[test]
+    fn decrypt_wrong_key() {
```
martinthomson, line 204: Also, when we don't have the identified key.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+        assert_eq!(match_key, r);
+    }
+
+    #[test]
+    fn decrypt_wrong_aad() {
+        let rng = StdRng::from_seed([1_u8; 32]);
+        let mut suite = EncryptionSuite::new(1, rng);
+        let match_key = XorReplicated::new(u64::MAX, u64::MAX / 2);
+        let enc = suite.seal(0, match_key);
+        suite.advance_epoch();
```
martinthomson, line 198: What are you planning to get from advancing the epoch like this?  We should be able decrypt epoch 0 when we reach epoch 5.  The only reason we wouldn't is a) policy reasons, or b) we rolled the keys and threw out the old ones (likely justified by the policy reasons).

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    use crate::secret_sharing::XorReplicated;
+
+    use hpke::{single_shot_seal_in_place_detached, OpModeS, Serializable};
+    use rand::rngs::StdRng;
+    use rand_core::{CryptoRng, RngCore, SeedableRng};
+
+    use crate::crypto::hpke::aad::{HelperOrigin, SiteOrigin};
+
+    #[derive(Clone)]
+    struct MatchKeyEncryption {
```
martinthomson, line 89: It strikes me that we might want a way to represent the thing that we parse off the wire.  Maybe this isn't it (we might want to reference a common `info`), but it might be worth thinking that through a little, to make decrypting lots of stuff easier.

martinthomson (changes requested):
A few things to look at here.  Mostly, Taiki's changes to XOR sharing.

I also wonder whether we need to build both ends of this, even if we only use the encryption routines for tests.

src/crypto/hpke/mod.rs
akoshelev: I added some proptests already - what is the specific scenario you want to test here?

src/crypto/hpke/mod.rs
akoshelev: yep, this suite is designed to help with unit tests, by no means I want to use it outside of `cfg(test)`. It kinda says "current epoch is X, can I decrypt this ct using key_id = 0`. If key_id was the same, but epoch changed, we expect decryption to fail and that is what this test validates

src/crypto/hpke/mod.rs
akoshelev: that seems to be validated at compile time unless I misunderstood the question here

src/crypto/hpke/mod.rs
akoshelev: do you want to test that keeping the tag intact and incrementing the size of the ciphertext leads to decryption error?

src/crypto/hpke/registry.rs
akoshelev: fair point, will fix

src/secret_sharing/xor.rs
akoshelev: I was a little cautious with that trait because it kinda carries two meanings - something that is serializable and something that can be sent between helpers. I am thinking that I should probably create a supertrait to `Message` and make xor share implement it?

src/secret_sharing/xor.rs
akoshelev: yea the way it is written, it supports both cases (exactly 2*SIZE_IN_BYTES and more)

src/crypto/hpke/mod.rs
martinthomson: Yes, but my expectation would be that the same key identifier would be good for multiple epochs.  There should be no need to synchronize key and epoch rotation.

src/crypto/hpke/mod.rs
martinthomson: You damage the ciphertext (and tag, which I generally don't treat as a distinct thing).  But you don't damage the info.  That's all.

src/crypto/hpke/mod.rs
martinthomson: See the above note about key identifiers and epochs.

src/crypto/hpke/mod.rs
martinthomson: How do you mean?  If I pass in a key identifier that you don't know about (say 100 in this test where you only have 0..10), I think the code will crash.

src/crypto/hpke/mod.rs
martinthomson: No, I'm suggesting that a 64-bit match key produces ciphertext and tag that are both 16 bytes, which allows you to cheat and use the same test to damage either from the same entropy.  A 40-bit match key (10 bytes of ciphertext) won't work the same way.

src/secret_sharing/xor.rs
martinthomson: Either way works for me.  I had the same thought, but I couldn't bring myself to care.

src/crypto/hpke/aad.rs
```diff
@@ -0,0 +1,99 @@
...
+///
+/// IPA uses key identifier, event epoch and both helper and site origins to authenticate match key
+/// encryptions. It is not guaranteed that the same receiver can be used for anything else.
+///
+/// [`info`]: https://www.rfc-editor.org/rfc/rfc9180.html#name-creating-the-encryption-con
+pub struct AssociatedData<'a> {
+    key_id: KeyIdentifier,
+    epoch: Epoch,
+    helper_origin: &'a HelperOrigin<'a>,
+    site_origin: &'a SiteOrigin<'a>,
```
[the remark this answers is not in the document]
akoshelev: fair enough, I like that I can catch the issue at compile time if I swap the arguments, but this is not a hill I am willing to die on :)

src/crypto/hpke/mod.rs
akoshelev: yep that is what I meant here

src/crypto/hpke/mod.rs
akoshelev: ah gotcha, so we don't rotate keys per epoch

src/crypto/hpke/mod.rs
akoshelev: ah sure I can do that - randomly flip bits in info and validate ct cannot be opened

src/crypto/hpke/mod.rs
akoshelev: ah that's right. That goes to the comment that we shouldn't crash if we don't have the key

src/crypto/hpke/mod.rs
akoshelev: got you. You're right, I cheated because initially I had two tests and they were very slow because of encryption. So I danced around it and used the same ct and tag to validate two separate test cases.

I will build two tests instead

src/crypto/hpke/mod.rs
akoshelev: synced up offline, we are keeping the match key encryption as [enc, ct] where ct = ct + tag. Once we start consuming it from the body, we will be able to see whether it is more convenient to keep it as is or to include the encapped key into the same slice

src/crypto/hpke/mod.rs
akoshelev: yep, keys and epochs are not connected. The general sentiment for this test still stands though imo - if epoch does not match, we should not be able to decrypt even when the same key is used and that is what this test validates by advancing the epoch and trying to decrypt using the same key/site origin/helper origin

src/crypto/hpke/mod.rs
akoshelev: I am going to randomly flip bits in random fields in `Info` struct. The API for `open` does not expose info byte string and I'll probably want to keep it that way

src/crypto/hpke/mod.rs
akoshelev: after combining ct + tag I no longer need to have two tests to validate corruption, but I built another one that checks corruption in `enc`

akoshelev:
> I also wonder whether we need to build both ends of this, even if we only use the encryption routines for tests.

somehow missed that comment. I think it would be easy to do later when we have CLIs that encrypt the matchkeys before sending them to the helpers.

akoshelev:
hmmm, it is not merged yet, only the serializable part of it

PR #411, opened 2023-01-24
free send and receive buffer resources when no longer needed

andyleiserson:
This is still pretty rough but it works for tests that don't use RandomBitsGenerator.

There is a check for channels not cleaned up, but the check is only effective if the test calls `world.join`, which most don't.

This is on top of the total-records PR. For a diff with only the latest changes, see https://github.com/andyleiserson/ipa/compare/total-records-3...andyleiserson:ipa:total-records-2a?diff=split

### DIFF
--- a/src/helpers/buffers/receive.rs
+++ b/src/helpers/buffers/receive.rs
@@ -131,6 +131,11 @@ impl ReceiveBuffer {
         }
     }
 
+    pub fn open_channels(&self) -> usize {
+        assert_eq!(self.inner.len(), self.record_ids.len());
+        self.inner.len()
+    }
+
     #[cfg(debug_assertions)]
     pub(in crate::helpers) fn waiting(&self) -> super::waiting::WaitingTasks {
         use super::waiting::WaitingTasks;
--- a/src/helpers/buffers/receive.rs
+++ b/src/helpers/buffers/receive.rs
@@ -3,9 +3,9 @@ use crate::{
     helpers::messaging::TotalRecords,
     protocol::RecordId,
 };
-use std::{collections::hash_map::Entry, num::NonZeroUsize};
-use std::collections::HashMap;
+use std::collections::{HashMap, hash_map::Entry};
 use std::fmt::Debug;
+use std::num::NonZeroUsize;
 use tokio::sync::oneshot;
 
 #[derive(Debug)]
--- a/src/helpers/buffers/receive.rs
+++ b/src/helpers/buffers/receive.rs
@@ -52,8 +52,9 @@ impl ReceiveBuffer {
         channel_id: ChannelId,
         record_id: RecordId,
         sender: oneshot::Sender<MessagePayload>,
+        total_records: NonZeroUsize,
     ) {
-        let total_records = channel_id.total_records.expect("can't receive without a known total record count");
+        //let total_records = channel_id.total_records.expect("can't receive without a known total record count");
         let channel = self.inner.entry(channel_id.clone()).or_insert_with(|| {
             tracing::trace!("create rx channel for rx request {:?}", &channel_id);
             ReceiveChannel {
@@ -84,10 +85,10 @@ impl ReceiveBuffer {
     /// chunk will belong to range of records [0..chunk.len()), second chunk [chunk.len()..2*chunk.len())
     /// etc. It does not require all chunks to be of the same size, this assumption is baked in
     /// send buffers.
-    pub fn receive_messages(&mut self, channel_id: &ChannelId, messages: &[u8]) {
+    pub fn receive_messages(&mut self, channel_id: &ChannelId, messages: &[u8], total_records: NonZeroUsize) {
         // TODO: this is probably wrong. Need to maintain total count as
         // unknown until we see a receive request.
-        let total_records = channel_id.total_records.expect("can't receive without a known total record count");
+        //let total_records = channel_id.total_records.expect("can't receive without a known total record count");
 
         let offset = self
             .record_ids
--- a/src/helpers/buffers/receive.rs
+++ b/src/helpers/buffers/receive.rs
@@ -1,5 +1,6 @@
 use crate::{
     helpers::{network::ChannelId, MessagePayload, MESSAGE_PAYLOAD_SIZE_BYTES},
+    helpers::messaging::TotalRecords,
     protocol::RecordId,
 };
 use std::{collections::hash_map::Entry, num::NonZeroUsize};
@@ -9,7 +10,7 @@ use tokio::sync::oneshot;
 
 #[derive(Debug)]
 struct ReceiveChannel {
-    total_records: NonZeroUsize,
+    total_records: TotalRecords,
     received_records: usize,
     items: HashMap<RecordId, ReceiveBufItem>,
 }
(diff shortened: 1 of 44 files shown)


src/helpers/buffers/receive.rs
```diff
@@ -57,21 +92,37 @@ impl ReceiveBuffer {
...
             .or_insert_with(|| RecordId::from(0_u32));
+        *offset += 1;
+        // TODO: need to pre-increment and drop this reference so we can
+        // call update_record_count below.  Can we do better?
+        let offset = *offset;
 
         for msg in messages.chunks(MESSAGE_PAYLOAD_SIZE_BYTES) {
             let payload = msg.try_into().unwrap();
-            match self
+            let channel = self
```
akoshelev, line 110: It does seem possible to move this construction out of the `for` loop and it could also solve the offset issue for you. I didn't think clearly when I wrote this, but now I think we can rely on `record_ids` and `inner` to be in sync and if there is nothing in `record_ids` for this channel, we can insert in both (`or_insert_with_key`?) and then use mutable references to receive channel and offset inside this for loop. 

maybe we can do even better and get rid of `record_ids` entirely, storing `(RecordId, ReceiveChannel)` pair inside `inner`

src/helpers/buffers/send.rs
```diff
@@ -49,35 +57,59 @@ impl SendBuffer {
...
+                }
             })
         };
 
         // TODO: avoid the copy here and size the element size to the message type.
         let mut payload = [0; ByteBuf::ELEMENT_SIZE_BYTES];
         payload[..msg.payload.len()].copy_from_slice(&msg.payload);
-        buf.insert(channel_id, usize::from(msg.record_id), &payload);
-        buf.take(self.items_in_batch.get())
+        channel.buf.insert(channel_id, usize::from(msg.record_id), &payload);
```
akoshelev, line 89: it would be great if `SendChannel` could insert items into its own buffer and update `send_records` counter by exposing an `insert` method. Then this code wouldn't need to worry about proper bookkeeping inside the send channel

src/helpers/buffers/send.rs
```diff
@@ -49,35 +57,59 @@ impl SendBuffer {
...
+        let res = channel.buf.take(self.items_in_batch.get());
+        if let Some(ref vec) = res {
+            assert!(vec.len() % ByteBuf::ELEMENT_SIZE_BYTES == 0);
+            channel.sent_records += vec.len() / ByteBuf::ELEMENT_SIZE_BYTES;
+        }
+        if NonZeroUsize::new(channel.sent_records) == Some(channel.total_records.try_into().unwrap()) {
+            assert!(NonZeroUsize::new(channel.buf.taken()) == Some(channel.total_records.try_into().unwrap()));
+            assert!(channel.buf.is_empty());
+            self.inner.remove(&channel_id);
```
akoshelev, line 98: I am wondering if @martinthomson and you had a chance to discuss whether we need to keep a tombstone after closing the channel. It will cost us something but also may catch some gnarly bugs in the future. Sending through a tombstone channel is an immediate panic. It is more work so even if we decide to do it, it does not need to be in this change, but I am just curious what you think about it

src/helpers/messaging.rs
```diff
@@ -85,15 +88,68 @@ pub struct Gateway {
...
+
+#[derive(Debug)]
+pub struct TotalRecordsUnknownError;
+
+impl TryInto<NonZeroUsize> for TotalRecords {
+    type Error = TotalRecordsUnknownError;
+    fn try_into(self) -> Result<NonZeroUsize, Self::Error> {
+        match self {
+            TotalRecords::Specified(value) => Ok(value),
+            _ => Err(TotalRecordsUnknownError),
```
akoshelev, line 142: a nit: it does not really unknown to us at this point. We could say exactly why we can't convert

src/helpers/buffers/send.rs
```diff
@@ -49,35 +57,59 @@ impl SendBuffer {
...
         payload[..msg.payload.len()].copy_from_slice(&msg.payload);
-        buf.insert(channel_id, usize::from(msg.record_id), &payload);
-        buf.take(self.items_in_batch.get())
+        channel.buf.insert(channel_id, usize::from(msg.record_id), &payload);
+        let res = channel.buf.take(self.items_in_batch.get());
+        if let Some(ref vec) = res {
+            assert!(vec.len() % ByteBuf::ELEMENT_SIZE_BYTES == 0);
+            channel.sent_records += vec.len() / ByteBuf::ELEMENT_SIZE_BYTES;
+        }
+        if NonZeroUsize::new(channel.sent_records) == Some(channel.total_records.try_into().unwrap()) {
```
akoshelev, line 95: would this explode if number of records is unspecified/indeterminate?

src/helpers/messaging.rs
```diff
@@ -121,6 +180,16 @@ impl Mesh<'_, '_> {
             )?;
         }
 
+        if let TotalRecords::Specified(count) = self.total_records {
```
akoshelev, line 183: maybe add a method to `TotalRecords` to validate this as it is used in more than one place?

src/helpers/messaging.rs
```diff
@@ -182,16 +267,18 @@ impl Gateway {
...
                         tracing::trace!("new {:?}", receive_request);
-                        receive_buf.receive_request(receive_request.channel_id, receive_request.record_id, receive_request.sender);
+                        let total_records = *channels.lock().unwrap().get(&receive_request.channel_id.step).expect("unknown channel");
+                        receive_buf.receive_request(receive_request.channel_id, receive_request.record_id, receive_request.sender, total_records);
                     }
                     Some((channel_id, messages)) = message_stream.next() => {
                         tracing::trace!("received {} bytes from {:?}", messages.len(), channel_id);
                         receive_buf.receive_messages(&channel_id, &messages);
                     }
                     Some((channel_id, envelope)) = send_rx.recv(), if pending_sends.is_empty() => {
```
martinthomson, line 277: This arrangement means that we need to do a hashtable lookup for every send.  We put the channel id, record number, message, and max record number in every message we send.

I've been trying to work out how I would do this for the ordering_mpsc I put together, and I don't think that this is the way.  The nice thing there is that the channel only carries a record ID and message, with the other stuff being setup at creation time.  As much as possible, I'd like to get to the point where we don't replicate information.  (It's necessary here because we have a single channel for all steps, but I think that's something we want to get away from.)

I think that we can - in most cases - know how many records there will be before we create any such channel.  That simplifies things because the channel can be constructed with knowledge of when it ends.

The indeterminate length version is a harder problem.  For that, it might be necessary to have a different implementation that uses an atomic to communicate where to end.  That is slightly more expensive to read, but it should be possible to set that to a non-zero value once the last record is sent.  It should be possible to deal with that by presenting different functions for each.

src/helpers/messaging.rs
```diff
@@ -207,6 +294,18 @@ impl Gateway {
                         #[cfg(debug_assertions)]
                         print_state(role, &send_buf, &receive_buf);
                     }
+                    Ok(()) = &mut shutdown_rx => {
```
martinthomson, line 297: Can you explain why you see a need for an explicit shutdown?  Do we have a problem with circular references, such that this sort of cleanup is necessary?

src/helpers/buffers/send.rs
martinthomson: I think that if we move to a channel per step, we'll be in a better position to handle that.

src/helpers/messaging.rs
andyleiserson: It's not functionally needed, but it makes it possible to check that all channels have been cleaned up at the end of tests.

src/helpers/messaging.rs
andyleiserson: Agree, I don't want this to be the final implementation.

And I agree it's knowable (in most cases) how many records there will be before the channel is first used, but since our current channel construction is "create on first use", will require some restructuring of the code to actually make use of that knowledge.

andyleiserson:
Closing-- this is addressed by the new infrastructure.

PR #412, opened 2023-01-25
IPA now runs with multi bit sort

richajaindce:
This just plugs multi bit sort inside IPA. Still working on PR to restructure code to accept a Vec<2D vector of records per multi bit slices>

RUST_LOG=raw_ipa=DEBUG cargo bench --bench oneshot_ipa --features="enable-benches" --no-default-features 
Communication down by 11.5% from 410490 records sent to 362970 for 100 rows


benches/oneshot/ipa.rs
```diff
@@ -34,9 +35,16 @@ async fn main() -> Result<(), Error> {
     let start = Instant::now();
     let result = world
         .semi_honest(records, |ctx, input_rows| async move {
-            ipa::<Fp32BitPrime>(ctx, &input_rows, 20, PER_USER_CAP, MAX_BREAKDOWN_KEY)
-                .await
-                .unwrap()
+            ipa::<Fp32BitPrime>(
+                ctx,
+                &input_rows,
+                20,
```
martinthomson, line 41: How does this map on to #409?

richajaindce: #409  is still calling generate_permutation and not generate_permutation_opt inside generate_permutation_and_reveal_shuffled. Once it is landed will stack this on top of that

akoshelev:
Merging as there is a bit of urgency here and I don't see any issues. We can follow up if someone has any comments

akoshelev (approved):

PR #413, opened 2023-01-25
[WIP] Refactoring code to work smoothly with multi bit sort

richajaindce:
Now, convert_all_bits returns vector of 2D vectors where each 2D vector is of size [ record_count ; num_multi_bits].
This helps in code clarity and is also required for malicious sort to do upgrade


richajaindce:
This is a rebase on top of #409 and #412 since #409 is already approved and hopefully will be merged soon and 412 is quite straightforward. 
Once this PR is landed, this will unblock malicious sort to use multi bit permutation

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -274,15 +278,14 @@ async fn sort_by_aggregation_bit<F: Field>(
 ) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
     // Since aggregation_bit is a 1-bit share of 1 or 0, we'll just extract the
     // field and wrap it in another vector.
-    let aggregation_bits = &[input
+    let aggregation_bits = vec![input
         .iter()
-        .map(|x| x.aggregation_bit.clone())
+        .map(|x| vec![x.aggregation_bit.clone()])
```
akoshelev, line 283: is it possible to avoid allocating a vector here? would `&[x.aggregation_bit]` serve the same purpose?

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,80 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
+/// Panics if input doesn't have same number of bits as `num_bits`
 #[must_use]
-pub fn transpose<F: Field>(input: &[Vec<Replicated<F>>]) -> Vec<Vec<Replicated<F>>> {
-    (0..input[0].len())
-        .map(|i| input.iter().map(|b| b[i].clone()).collect::<Vec<_>>())
+pub fn split_into_multi_bit_slices<F: Field>(
+    input: &[Vec<Replicated<F>>],
+    num_bits: u32,
+    num_multi_bits: u32,
+) -> Vec<Vec<Vec<Replicated<F>>>> {
```
akoshelev, line 37: same here, it feels like you should be able to borrow from `input` to get rid of the most inner vector. Is it possible to return an iterator here and cut the memory footprint to a constant value? this nested structure looks frightening when I think about 1M items input

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,80 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
+/// output:
+/// `[`
+///     `[ row[0].bit0, row[0].bit1, ..., row[0].bit31 ]`,
+///     `[ row[1].bit0, row[1].bit1, ..., row[1].bit31 ]`,
+///     ...
+///     `[ row[n].bit0, row[n].bit1, ..., row[n].bit31 ]`,
+///  `]`
+
+/// # Errors
+/// If any underlying protocol fails
```
akoshelev, line 74: that does not seem like a correct description

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,80 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
+/// # Panics
+/// Panics if input doesn't have same number of bits as `num_bits`
+#[must_use]
+pub fn combine_slices<F: Field>(
+    input: &[Vec<Vec<Replicated<F>>>],
+    num_bits: u32,
+) -> Vec<Vec<Replicated<F>>> {
```
akoshelev, line 81: same idea - see if it is possible to return an iterator here

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,80 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
+    let record_count = input[0].len();
+    let mut output = Vec::with_capacity(record_count);
+    output.resize_with(record_count, || {
+        Vec::with_capacity(num_bits.try_into().unwrap())
+    });
+    for slice in input {
+        output.push(Vec::new());
```
akoshelev, line 88: this makes `output` size  `record_count * input.len()` while you only fill the first `record_count` rows. Is it correct?

akoshelev (changes requested):

src/protocol/modulus_conversion/mod.rs
richajaindce: This indeed is the right description. we are basically collapsing same record bits together in the same row

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -134,27 +151,33 @@ where
...
-        })
-        .collect::<Vec<_>>();
-    let converted_shares = try_join_all(futures).await?;
-    Ok(converted_shares)
+    try_join_all(
+        (0..num_bits)
+            .step_by(num_multi_bits.try_into().unwrap())
+            .map(|bit_num| {
+                let last_bit = std::cmp::min(num_multi_bits + bit_num, num_bits) as usize;
+                try_join_all(locally_converted_bits.enumerate().map(|(idx, record)| {
```
richajaindce, line 171: Doesnt compile saying "cannot move out of `locally_converted_bits`, a captured variable in an `FnMut` closure
move occurs because `locally_converted_bits` has type `I`, which does not implement the `Copy` trait"

src/protocol/modulus_conversion/mod.rs
akoshelev: I meant that you don't return `Result` so there are no errors that can be reported from `combine_slice`

src/protocol/modulus_conversion/convert_shares.rs
akoshelev: I pushed a small change to your branch that makes this code compile. basically it did not work because you wanted to consume this iterator multiple times within `map` function and Rust wouldn't allow that. [`by_ref`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref) lets you exclusively borrow this iterator multiple times and it should work in this case, given that you own `locally_converted_bits`

richajaindce: The code compiles with this change but the iterator is empty after first iteration.

src/protocol/modulus_conversion/mod.rs
richajaindce: Oh my bad.. fixing this

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Made it work

src/protocol/modulus_conversion/mod.rs
richajaindce: Fixed it

src/protocol/modulus_conversion/mod.rs
richajaindce: This can return an iterator but the consumer of output (sort) still needs a slice. In one of the places where sort is called, we need to use the input of the function in next step: https://github.com/private-attribution/ipa/blob/main/src/protocol/ipa/mod.rs#L156

Can we keep working on the iterators in the next PR?

src/protocol/attribution/aggregate_credit.rs
richajaindce: This needs another bunch of refactoring since this is an input to sort which is called from multiple places. Lets revisit this in the next PR?

richajaindce:
I am still trying to see how much of the code can be iter-i-fied. But in the meantime, can we work towards landing this code and handle iterators in follow up PRs?

richajaindce:
This looks a bit complex so making this a draft to keep discussing making the iterator change

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -253,19 +250,21 @@ async fn sort_by_breakdown_key<F: Field>(
...
     // breakdown key values.
     let valid_bits_count = u128::BITS - (max_breakdown_key - 1).leading_zeros();
 
-    let sort_permutation = generate_permutation_and_reveal_shuffled(
-        ctx.narrow(&Step::GeneratePermutationByBreakdownKey),
-        &breakdown_keys[..valid_bits_count as usize],
+    let bit_decompose_breakdown_key =
+        bit_decompose_breakdown_key(ctx.narrow(&Step::BitDecomposeBreakdownKey), input).await?;
+    let breakdown_keys = split_into_multi_bit_slices(
+        &bit_decompose_breakdown_key,
```
benjaminsavage, line 260: I think @taikiy is/has removed this.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -86,22 +87,37 @@ pub fn convert_bit_local<F: Field, B: BitArray>(
...
-    }
-    total_list
+    num_bits: u32,
+) -> impl Iterator<Item = Vec<BitConversionTriple<Replicated<F>>>> + '_ {
+    input.iter().map(move |record| {
+        (0..num_bits)
+            .map(|bit_index: u32| {
+                let left = u128::from(record.left()[bit_index]);
+                let right = u128::from(record.right()[bit_index]);
+                BitConversionTriple(match helper_role {
```
benjaminsavage, line 100: This seems highly duplicative of `convert_bit_local`. Why not invoke that?

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -134,28 +150,47 @@ where
...
-                one_column,
-            )
-        })
-        .collect::<Vec<_>>();
-    let converted_shares = try_join_all(futures).await?;
-    Ok(converted_shares)
+        .for_each(|(idx, record)| {
+            (0..count).for_each(|slice_num| {
+                let first_bit =
+                    num_multi_bits * <usize as TryInto<u32>>::try_into(slice_num).unwrap();
```
benjaminsavage, line 177: That cast tho... why is it so complicated looking?

benjaminsavage, line 177: I think you can do this in a much more idiomatic way with `.chunks(num_multi_bits)`.

https://docs.rs/itertools/0.10.5/itertools/trait.Itertools.html#method.chunks

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -134,28 +150,47 @@ where
...
     C: Context<F, Share = S>,
-    S: ArithmeticSecretSharing<F>,
+    S: ArithmeticSecretSharing<F> + 'a,
+    I: Iterator<Item = Vec<BitConversionTriple<S>>>,
 {
-    let futures = locally_converted_bits
-        .iter()
+    let num_records = locally_converted_bits.size_hint().0;
+
+    let count = ((num_bits + num_multi_bits - 1) / num_multi_bits) as usize;
```
benjaminsavage, line 167: I'll admit - I have no idea where this equation came from or what it does.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -134,28 +150,47 @@ where
...
+            (0..count).for_each(|slice_num| {
+                let first_bit =
+                    num_multi_bits * <usize as TryInto<u32>>::try_into(slice_num).unwrap();
+                let last_bit = std::cmp::min(first_bit + num_multi_bits, num_bits) as usize;
+                futures[slice_num].push(convert_bit_list(
+                    ctx.narrow(&ModulusConversion(first_bit))
+                        .set_total_records(num_records),
+                    record[first_bit as usize..last_bit].to_vec(),
+                    RecordId::from(idx),
+                ));
```
benjaminsavage, line 184: I don't think we want to call "convert_bit_list" on ALL of the bits at once. When it comes time to upgrade to malicious security, we will want to just convert the bits that are inputs to the next stage of the sort algorithm, use them, call validate, and then we can call "convert_bit_list" on the next batch, as they are inputs to the next stage, and need to be validated along with the circuit where they are used.

src/protocol/attribution/aggregate_credit.rs
taikiy: #418. Not merged yet but it will be removed.

src/protocol/modulus_conversion/convert_shares.rs
martinthomson: This rounds `num_bits` up to the next multiple of `num_multi_bits`.  I've not looked at why...

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: because rust doesnt let me do slice_num.try_into().unwrap() when multiplying usize numbers :(

martinthomson: `u32::try_from(slice_num).unwrap()` would be easier to read.

richajaindce:
Will work on iterators in future PRs

PR #414, opened 2023-01-25
Serializable trait

akoshelev:
This is taken out from #410 hence the crazy commit history. All the changes related to HPKE encryption have been excluded.

The motivation for this change is to polish a bit our serialization story. Before it, some things were serializable (Field) while others (BooleanShare) were not. For IPA pretty much everything needs to be put on the wire at some point, so it makes sense to create a trait and force our basic primitives to implement it.

I intentionally split the `Message` trait and made it a marker trait instead. I want code to opt-in in situations when they want to send data between helpers and be explicit about it. 

There are two main reasons why I want to push it separately:

* Changes to core primitives. Forcing `Serializable` implementation on `SharedValue` turned out to be quite involved. HPKE change looks bigger that it should because of it. 
* Unblock @thurstonsand work on merging two branches


src/bits/bit_array.rs
```diff
@@ -252,3 +252,85 @@ macro_rules! bit_array_impl {
 
 bit_array_impl!(bit_array_64, BitArray64, U8_8, 64);
 bit_array_impl!(bit_array_40, BitArray40, U8_5, 40);
+
+impl Serializable for BitArray40 {
```
thurstonsand, line 256: is there a reason we couldn't implement this inside `bit_array_impl`?

src/bits/bit_array.rs
```diff
@@ -252,3 +252,85 @@ macro_rules! bit_array_impl {
...
+
+    fn serialize(self, buf: &mut [u8]) -> std::io::Result<()> {
+        let req = <Self as Serializable>::SIZE_IN_BYTES;
+
+        if buf.len() >= req {
+            buf[..req].copy_from_slice(self.0.as_raw_slice());
+            Ok(())
+        } else {
+            let error_text = format!(
+                "Buffer with total capacity {} cannot hold field value {:?} because \
```
thurstonsand, line 267: > field value

change to

> bit array

?

and elsewhere

thurstonsand (changes requested):
what's going on with `src/secret_sharing/xor.rs`? empty file?

src/bits/bit_array.rs
akoshelev: no real reason, I thought it won't be possible because the size of the bit array *store* needs to be known at compile time, but Rust seems to be handling it just fine

andyleiserson (approved):

thurstonsand (approved):
nice!

src/bits/bit_array.rs
```diff
@@ -159,6 +159,47 @@ macro_rules! bit_array_impl {
                 }
             }
 
+            impl Serializable for $name {
+                const SIZE_IN_BYTES: usize = Self::BITS as usize / 8;
```
martinthomson, line 163: This isn't correct, or at least, not strictly.

You want:

```rust
const SIZE_IN_BYTES: usize = (Self::BITS as usize + 7) / 8;
```

src/ff/field.rs
```diff
@@ -31,6 +32,10 @@ pub trait Field: ArithmeticShare + From<u128> + Into<Self::Integer> {
         let int: Self::Integer = (*self).into();
         int.into()
     }
+}
+
+impl<F: Field> Serializable for F {
+    const SIZE_IN_BYTES: usize = F::BITS as usize / 8;
```
martinthomson, line 38: See above.

src/bits/mod.rs
```diff
@@ -53,3 +54,23 @@ impl<T> BooleanRefOps for T where
     T: for<'a> BitXor<&'a Self, Output = Self> + for<'a> BitXorAssign<&'a Self>
 {
 }
+
+/// Trait for items that have fixed-byte length representation.
+pub trait Serializable: Sized {
```
martinthomson, line 59: Is this the right location for this trait?

src/bits/mod.rs
```diff
@@ -53,3 +54,23 @@ impl<T> BooleanRefOps for T where
...
+    /// Required number of bytes to store this message on disk/network
+    const SIZE_IN_BYTES: usize;
+
+    /// Serialize this message to a mutable slice. Implementations need to ensure `buf` has enough
+    /// capacity to store this message.
+    ///
+    /// ## Errors
+    /// Returns an error if `buf` does not have enough capacity to store at least `SIZE_IN_BYTES` more
+    /// data.
+    fn serialize(self, buf: &mut [u8]) -> io::Result<()>;
```
martinthomson, line 69: Do we really want to return `Result` from these?  It seems that we might instead insist that the caller provide a precisely-sized slice to both functions.  After all, the caller knows.

I see a lot of checking in the implementations of these and that is really unnecessary.  Given that these are likely on hot paths, that extra checking comes with costs.

(Ideally, this would be a runtime constraint and the argument would be `&[u8; Self::SIZE_IN_BYTES]`, but I don't know if that works, even with generic const expressions.)

martinthomson (commented):
A great opportunity to find some ways to improve code :)

PR #415, opened 2023-01-26
HPKE for match key encryption (2nd attempt)

akoshelev:
As #410 got closed by #414, submitting hpke changes only. All comments from #410 have been addressed


src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+
+                encryption.enc.as_mut()[bad_byte] ^= 1 << bad_bit;
+                let _ = suite.open(0, encryption).unwrap_err();
+            }
+        }
+
+        proptest::proptest! {
+            #![proptest_config(ProptestConfig::with_cases(50))]
+            #[test]
+            fn arbitrary_info_corruption(corrupted_info_field in 1..4, seed: [u8; 32]) {
```
martinthomson, line 303: Would it be easier to damage the encoded info block?

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+                        let idx = rng.gen_range(0..site_origin.len());
+                        site_origin.remove(idx);
+
+                        Info {
+                            site_origin: site_origin.as_ref(),
+                            ..encryption.info
+                        }
+                    },
+                    4 => {
+                        let idx = rng.gen_range(0..helper_origin.len());
```
martinthomson, line 330: You are removing a byte, which is probably OK, but there is also adding and modifying.  Could this use `origin in "[a-z]{1,10}"` or whatever proptest provides?

src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub mod bits;
 pub mod chunkscan;
 #[cfg(feature = "cli")]
 pub mod cli;
+pub mod crypto;
```
martinthomson, line 11: Do we have another "crypto" module that we might add here?  ~~"nft" perhaps?~~ I can't think of anything that fits.

martinthomson (approved):
A few residual things, but nothing blocking.  Land once you have fixed the naming issues.

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+/// needs to know which epoch to use for each individual event.
+///
+/// [`ipa-spec`]: https://github.com/patcg-individual-drafts/ipa/blob/main/IPA-End-to-End.md#other-key-terms
+pub type Epoch = u16;
+type IpaPublicKey = <IpaKem as hpke::kem::Kem>::PublicKey;
+type IpaPrivateKey = <IpaKem as hpke::kem::Kem>::PrivateKey;
+
+/// Total len in bytes for an encrypted matchkey including the authentication tag.
+pub const MATCHKEY_CT_LEN: usize = <XorReplicated as crate::bits::Serializable>::SIZE_IN_BYTES
+    + <AeadTag<IpaAead> as hpke::Serializable>::OutputSize::USIZE;
```
thurstonsand, line 42: would this work?
```suggestion
pub const MATCHKEY_CT_LEN: usize = XorReplicated::SIZE_IN_BYTES
    + AeadTag::<IpaAead>::OutputSize::USIZE;
```

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum DecryptionError {
+    #[error("Unknown key {0}")]
+    NoSuchKey(KeyIdentifier),
+    #[error("Failed to open ciphertext")]
+    Other,
+}
+
+impl From<hpke::HpkeError> for DecryptionError {
+    fn from(_value: hpke::HpkeError) -> Self {
```
thurstonsand, line 53: any particular reason you are not exposing the underlying error?

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        Self::Other
+    }
+}
+
+/// Opens the given ciphertext in place by first obtaining the secret key from [`key_registry`]
+/// using epoch and key from [`info`] and then applying [`HPKE decryption`] to the provided ciphertext.
+///
+/// This function mutates the provided ciphertext slice and replaces it with the plaintext obtained
+/// after opening the ciphertext. The result will contain a pointer to the plaintext slice.
+/// of the plaintext. Note that if the ciphertext slice does not include authentication tag, decryption
```
thurstonsand, line 69:
```suggestion
/// Note that if the ciphertext slice does not include authentication tag, decryption
```

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        sk,
+        &encap_key,
+        &info,
+        ct,
+        &[],
+        &tag,
+    )?;
+
+    // at this point ct is no longer a pointer to the ciphertext.
+    let pt = ct;
```
thurstonsand, line 104: what is this doing that `Ok(ct)` does not?

src/crypto/hpke/mod.rs
akoshelev: fixed that as well

src/lib.rs
akoshelev: eh maybe not, lemme lift hpke module

src/crypto/hpke/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+    /// [`aead`]: IpaAead
+    enc: [u8; 32],
+
+    /// Ciphertext + tag
+    ct: [u8; MATCHKEY_CT_LEN],
+
+    /// Info part of the receiver context as defined in [`url`] specification.
+    ///
+    /// [`url`]: https://datatracker.ietf.org/doc/html/rfc9180#section-5.1
+    info: Info<'a>,
```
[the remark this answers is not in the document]
martinthomson: Yeah, I expect that there will be a very limited set of `info` values in a full query.  You have the cross product of: site, epoch, and key identifier.  All of these being low cardinality.

src/hpke/registry.rs
```diff
@@ -0,0 +1,53 @@
+use super::{IpaPrivateKey, IpaPublicKey, KeyIdentifier};
+
+/// A pair of secret key and public key. Public keys used by UA to encrypt the data towards helpers
+/// secret keys used by helpers to open the ciphertexts. Each helper needs access to both
+/// TODO: we may decide to use different HPKE settings for each key identifier.
```
martinthomson, line 5: That's a big lift.  It's a classic failing in software to not prepare for the obsolescence of the crypto you are using.  And rust only makes it works.

On the other hand, I don't see a way to fix that without making the code much worse, so let's not concern ourselves with it.

```suggestion
```

martinthomson (approved):

PR #417, opened 2023-01-26
Add multiplicative identity const to SharedValue

akoshelev:
Both boolean and arithmetic fields used in IPA make use of additive and multiplicative identity elements. This change removes the discrepancy between arithmetic and boolean field defitions by requiring `SharedValue` to define that constant (ONE) and implements it for bit arrays that are used to represent boolean field values

### DIFF
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -3,10 +3,13 @@ use crate::bits::Serializable;
 use crate::secret_sharing::SharedValue;
 use bitvec::prelude::{BitArr, Lsb0};
 
+type Container = u8;
+type Order = Lsb0;
+
 /// Bit store type definition. Five `u8` blocks.
-type U8_5 = BitArr!(for 40, in u8, Lsb0);
+type U8_5 = BitArr!(for 40, in Container, Order);
 /// Bit store type definition. Eight `u8` blocks.
-type U8_8 = BitArr!(for 64, in u8, Lsb0);
+type U8_8 = BitArr!(for 64, in Container, Order);
 
 macro_rules! bit_array_impl {
     ( $modname:ident, $name:ident, $store:ty, $bits:expr ) => {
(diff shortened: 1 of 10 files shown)


thurstonsand (approved):

thurstonsand:
that's actually pretty straightforward

src/secret_sharing/mod.rs
```diff
@@ -16,7 +16,11 @@ pub trait SharedValue:
     /// Number of bits stored in this data type.
     const BITS: u32;
 
+    /// Additive identity element.
     const ZERO: Self;
+
+    /// Multiplicative identity element
```
martinthomson, line 22:
```suggestion
    /// Multiplicative identity element.
```

src/bits/bit_array.rs
```diff
@@ -24,6 +27,7 @@ macro_rules! bit_array_impl {
             impl SharedValue for $name {
                 const BITS: u32 = $bits;
                 const ZERO: Self = Self(<$store>::ZERO);
+                const ONE: Self = Self(bitvec::bitarr!(const Container, Order; 1; $bits));
```
martinthomson, line 30: I think that we avoided doing this because it requires that we consider the bit array to be something other than an ordered collection of bits.  This explicitly says that the bit array is a representation of a number in a particular order.

I think that's OK, but it's worth documenting that assumption somewhere..

src/bits/bit_array.rs
```diff
@@ -296,6 +303,24 @@ macro_rules! bit_array_impl {
 
                     assert_eq!(a, $name::deserialize(&buf).unwrap());
                 }
+
+                proptest! {
+                   #[test]
+                   fn identities(a in 2_u128..1 << $name::BITS - 1) {
```
martinthomson, line 309: I'm not clear on the operator precedence rules here.  Which did you mean?

```suggestion
                   fn identities(a in 2_u128..((1 << $name::BITS) - 1)) {
```

or 

```suggestion
                   fn identities(a in 2_u128..(1 << ($name::BITS - 1))) {
```

Either way, this is a bit of an odd test.  You are testing all values that aren't 1 or 0 and then checking operations that only affect the last bit.  That's not going to be of much use.

martinthomson (commented):
I'm not 100% of this.  It's a reframing of our conception of the bit array type.

akoshelev:
don't need it anymore

PR #418, opened 2023-01-27
GenericReport input struct

taikiy:
# Overview

Most changes are moving definitions and renamings. New logic `src/test_fixture/input/mod.rs` is for tests.

Currently we have 2 structs for each protocol's input i.e., `AttributionInputRow`, `CreditCappingInputRow`, `IPAInputRow` containing replicated shares, and `__TestInputRow` structs that implement `IntoShares`. For inputs that contain the match key, we also have modulus-converted structs (from `XorReplicated<B>` to `Vec<Replicated<F>>`). The list will keep growing as we add new attribution/aggregation protocols. This PR introduces generic structs that implements `IntoShares`. They contain all fields for source and trigger events defined in the IPA spec, and generate shares used in tests. 

With this change, I changed `breakdown_key` to xor share and remove the bit-decomposition step from the aggregation protocol. The benchmark result is shown below.

# Performance
Run oneshot/ipa.rs 10 times on main and this PR.

Command:
cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches"

Parameters:
BATCHSIZE = 8192
NUM_MULTI_BITS = 3
MAX_BREAKDOWN_KEY = 4
MatchKey bits = 40
BreakdownKey bits = 8

Time:
main = 142.912
new = 121.631
-14.9%
(MP-SPDZ: 8.564s)

Steps:
main = 30,969,822
latest = 25,068,534
-19.0%


taikiy:
@thurstonsand this PR impacts your #408.

src/protocol/input.rs
```diff
@@ -0,0 +1,340 @@
...
+        let f_match_key = match self.match_key.clone() {
+            None => async { Ok::<_, Error>(None) }.left_future(),
+            Some(v) => async move {
+                let a = v
+                    .reshare(c.narrow(&ReshareStep::MatchKey), record_id, to_helper)
+                    .await
+                    .unwrap();
+                Ok(Some(a))
+            }
+            .right_future(),
```
taikiy, line 169: Is there a better way to handle Option + Future?

src/protocol/input.rs
```diff
@@ -0,0 +1,340 @@
...
+use futures::future::{try_join4, try_join_all};
+use futures::FutureExt;
+use std::iter::{repeat, zip};
+
+pub type MatchKey = BitArray40;
+pub type BreakdownKey = BitArray8;
+
+// Struct that holds all possible fields of the input to IPA.
+#[derive(Debug)]
+pub struct GenericReportShare<F: Field, MK: BitArray, BK: BitArray> {
```
martinthomson, line 21: I think that this sort of catch-all type isn't the best way to approach this.

Taking a high-level look at this, we have a protocol that glues a bunch of discrete stages together into a coherent whole.  That overall protocol will have an input of a particular form (this with some fields removed and other fields made non-optional).

Internally, we need something as well.  Internal steps in that process carry stuff along to future steps, but don't care what that payload looks like (we can use a very generic `T: Resharable` or something for the stuff that we sort).  And some protocol parts directly chain together with concrete types.

Having a single type with lots of optional bits makes it hard to understand because the type doesn't force values to be valid in any given context.  Can I provide something with only breakdown keys and trigger values?  This type permits that.

It would be better overall to have each internal step define its own input format, then to try to rationalize them throughout by making them consistent with each other (with naming, types, etc...).

martinthomson (changes requested):
I haven't looked closely, but the code to generate somewhat plausible inputs seems quite useful.  The structures though seem like they might make this harder to work with.

src/protocol/input.rs
taikiy: Reducing the number of `*InputRow`/`*OutputRow` was something @benjaminsavage and I discussed, but I do agree that there's a benefit of having distinct structs for each protocol.

My motivation was to reduce duplicate trait implementations like `IntoShares`, `Reconstruct` and `Resharable`. Since two of those three are for tests, I've changed my approach a bit. We still have the generic structs but only for tests, while keeping the distinct input/output structs for production code. I think the protocols' data flow are easier to understand with this new change, while reducing some of the redundancy.

src/protocol/attribution/credit_capping.rs
```diff
@@ -13,15 +14,16 @@ use crate::secret_sharing::replicated::semi_honest::AdditiveShare as Replicated;
 use futures::future::{try_join, try_join_all};
 use std::iter::{repeat, zip};
 
+/// User-level credit capping protocol.
 ///
-/// # Errors
-/// Propagates errors from multiplications
+/// ## Errors
```
thurstonsand, line 19:
```suggestion
/// # Errors
```

thurstonsand (approved):
thanks for letting me know. will have to resolve a bit of duplicate work with my branch, but no big deal

src/protocol/attribution/input.rs
```diff
@@ -0,0 +1,172 @@
...
+#[derive(Debug)]
+pub struct AccumulateCreditInputRow<F: Field, BK: BitArray> {
+    pub is_trigger_report: AdditiveShare<F>,
+    pub helper_bit: AdditiveShare<F>,
+    pub breakdown_key: XorShare<BK>,
+    pub trigger_value: AdditiveShare<F>,
+}
+
+#[derive(Debug)]
+pub struct MCAccumulateCreditInputRow<F: Field> {
```
benjaminsavage, line 24: What does "MCA" stand for? I don't quite get this name.

taikiy: MC = Modulus Converted. Names getting too long, it might be a time we need to use abbreviations.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -186,23 +160,40 @@ pub async fn aggregate_credit<F: Field>(
...
+        .collect::<Vec<_>>();
+
+    let converted_breakdown_keys = convert_all_bits(
+        &ctx.narrow(&Step::ModulusConversionForBreakdownKey),
+        &convert_all_bits_local(ctx.role(), &unique_breakdown_keys),
+        BK::BITS,
+        num_multi_bits,
+    )
+    .await
+    .unwrap();
```
benjaminsavage, line 184: This seems wasteful. I don't think it's necessary to make this an async function, and actually perform real multiplications here given that we are just generating secret-sharings of known values.

I'm OK with removing this in a follow-up PR though.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -211,55 +202,35 @@ fn add_aggregation_bits_and_breakdown_keys<F: Field>(
...
+                helper_bit: one.clone(),
+                aggregation_bit: one.clone(),
             })
             .collect::<Vec<_>>(),
     );
 
-    unique_breakdown_keys
-}
-
-async fn bit_decompose_breakdown_key<F: Field>(
```
benjaminsavage, line 226: Woohoo!

src/protocol/attribution/input.rs
```diff
@@ -0,0 +1,172 @@
...
+            f_breakdown_key,
+            try_join_all([f_aggregation_bit, f_helper_bit, f_value]),
+        )
+        .await?;
+
+        Ok(MCCappedCreditsWithAggregationBit {
+            breakdown_key,
+            helper_bit: fields.remove(0),
+            aggregation_bit: fields.remove(0),
+            credit: fields.remove(0),
```
benjaminsavage, line 146: Some day, I hope to find a better way to run these in parallel. This whole `.remove(0)` thing feels so awkward.

PR #420, opened 2023-01-28
Experimental `Serialize` trait backed by GenericArray

akoshelev:
`serialize`/`deserialize` now work with fixed-size arrays. Unfortunately it does not eliminate all the runtime checks because in some places I had to use `GenericArray::from_slice`.

The benefit of this approach is that `Serialize` implementors do not need to check the size of the buffer, it is guaranteed to be set correctly.

this is still rough (I want to fix import and add some doc tests) but should demonstrate the idea.

### DIFF
--- a/src/bits/bit_array.rs
+++ b/src/bits/bit_array.rs
@@ -2,6 +2,8 @@ use super::BitArray;
 use crate::bits::Serializable;
 use crate::secret_sharing::SharedValue;
 use bitvec::prelude::{BitArr, Lsb0};
+use sha2::digest::generic_array::GenericArray;
+use sha2::digest::typenum::Unsigned;
 
 /// Bit store type definition. Five `u8` blocks.
 type U8_5 = BitArr!(for 40, in u8, Lsb0);
@@ -9,7 +11,7 @@ type U8_5 = BitArr!(for 40, in u8, Lsb0);
 type U8_8 = BitArr!(for 64, in u8, Lsb0);
 
 macro_rules! bit_array_impl {
-    ( $modname:ident, $name:ident, $store:ty, $bits:expr ) => {
+    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $arraylen:ty ) => {
         mod $modname {
             use super::*;
 
@@ -28,11 +30,8 @@ macro_rules! bit_array_impl {
 
             impl BitArray for $name {
                 fn truncate_from<T: Into<u128>>(v: T) -> Self {
-                    Self(<$store>::new(
-                        v.into().to_le_bytes()[0..<Self as Serializable>::SIZE_IN_BYTES]
-                            .try_into()
-                            .unwrap(),
-                    ))
+                    let v = &v.into().to_le_bytes()[..<Self as Serializable>::Size::to_usize()];
+                    Self(<$store>::new(v.try_into().unwrap()))
                 }
             }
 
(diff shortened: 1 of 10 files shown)


src/bits/bit_array.rs
```diff
@@ -303,5 +273,6 @@ macro_rules! bit_array_impl {
     };
 }
 
-bit_array_impl!(bit_array_64, BitArray64, U8_8, 64);
-bit_array_impl!(bit_array_40, BitArray40, U8_5, 40);
+use sha2::digest::typenum::{U5, U8};
```
martinthomson, line 276: Do you really have to use the sha2 crate for this?  Can we get this from the typenum crate directly?

src/bits/mod.rs
```diff
@@ -1,5 +1,7 @@
 use crate::secret_sharing::BooleanShare;
-use std::io;
+
+use sha2::digest::generic_array::{ArrayLength, GenericArray};
```
martinthomson, line 3: Again, generic_array has its own crate.

src/secret_sharing/replicated/semi_honest/xor_share.rs
```diff
@@ -73,47 +78,33 @@ impl<V: BooleanShare> BitXorAssign<&Self> for XorShare<V> {
...
-            let right = V::deserialize(&buf[V::SIZE_IN_BYTES..2 * V::SIZE_IN_BYTES])?;
-
-            Ok(Self::new(left, right))
-        }
+impl<V: BooleanShare> Serializable for XorShare<V>
+where
+    V::Size: Add<V::Size>,
+    <V::Size as Add<V::Size>>::Output: ArrayLength<u8>,
+{
+    type Size = <V::Size as Add<V::Size>>::Output;
```
martinthomson, line 86: Clever.  typenum is completely bonkers.  The use of Add to generate derived types seems like a misuse of the trait; but this is fine.

src/bits/mod.rs
```diff
@@ -58,19 +60,19 @@ impl<T> BooleanRefOps for T where
...
     /// Required number of bytes to store this message on disk/network
-    const SIZE_IN_BYTES: usize;
+    type Size: ArrayLength<u8>;
 
     /// Serialize this message to a mutable slice. Implementations need to ensure `buf` has enough
     /// capacity to store this message.
     ///
     /// ## Errors
     /// Returns an error if `buf` does not have enough capacity to store at least `SIZE_IN_BYTES` more
     /// data.
```
martinthomson, line 70:
```suggestion
```

I guess that using `GenericArray` means that you can't mess this up.

src/bits/mod.rs
```diff
@@ -58,19 +60,19 @@ impl<T> BooleanRefOps for T where
...
-    fn serialize(self, buf: &mut [u8]) -> io::Result<()>;
+    fn serialize(self, buf: &mut GenericArray<u8, Self::Size>);
 
     /// Deserialize message from a sequence of bytes.
     ///
     /// ## Errors
     /// Returns an error if the provided buffer does not have enough bytes to read (EOF).
```
martinthomson, line 76:
```suggestion
```

martinthomson (changes requested):
A good direction.  Please update Cargo.toml to include the typenum and generic_array crates, since we are using them directly.

src/bits/bit_array.rs
akoshelev: yes that was the plan if this is the right direction

src/bits/bit_array.rs
```diff
@@ -273,6 +273,6 @@ macro_rules! bit_array_impl {
     };
 }
 
-use sha2::digest::typenum::{U5, U8};
+use typenum::{U5, U8};
```
martinthomson, line 276: This might go up top with the other `use` statements.

martinthomson (approved):

andyleiserson:
I like the safety this provides for things like the size of a serialized XorShare.

I have been looking at some performance issues and in particular the potential performance benefit of vectorization. I hope to have some concrete results to share soon, but one issue that has raised is whether we will need to send larger messages across the wire. And if we need larger messages, we may need variable-sized messages, which may mean changing our serialization framework to e.g. serialize a payload of unknown length to a `Vec<u8>`. Even if we do, I still think this change provides valuable safety in some places in the serialization routines, so I'm mentioning this as general context and something to think about, not a suggestion to do anything in particular with this change.

akoshelev:
Good point @andyleiserson. My understanding is that we certainly need to send larger chunks across the wire. Much much larger than a single serialized xorshare or a single field value. I hoped that every implementation of `Field` or `Share` would have the same size on the wire, the only difference is how many of those things we want to pack together before sending them out. 

The existing hack of using 8 bytes for serialized messages is bad but we could probably fix it w/o affecting serialization protocol for individual values. I really wanted to make message size a const on the step, so we could tune buffer size for each step individually, but const generics is not a thing yet. Maybe you have an idea how we could send messages with no overhead today, it would be great to fix this mess.  

Using `GenericArray` provides compile-time safety for implementers of `Serialize` trait, it is still very easy to cause a panic at the callsite by wrapping a slice of the wrong length.

```rust
   let xor_share = XorShare::new(..);
   let mut slice = [0_u8; 1];
   xor_share.serialize(GenericArray::from_mut_slice(&mut slice));
```

but at the same time it still allows serializing fields/shares directly to heap, so hopefully it shouldn't be a blocker if we decide to do so.

PR #421, opened 2023-01-30
Make upgrade work for 2D vectors of the form [num_records * multiple columns]

richajaindce:


richajaindce:
@andyleiserson Can you please help review this

andyleiserson:
Is there a problem with the following? When I wrote this I used `BitOpStep` since it was convenient, but if that doesn't seem like an appropriate use of `BitOpStep`, it could also use the `Upgrade2DVectors` step that you defined.

I pulled the commit you sent me (7350120) and changed the upgrade impl to the below, and it at least fixed the `malicious_sort_in_semi_honest` test.

Functionally, I think what you have in this PR is fine, but there is some redundancy because it is manually inlining the 1D implementation within the 2D implementation.

```rust
#[async_trait]
impl<'a, F> UpgradeToMalicious<Vec<Vec<Replicated<F>>>, Vec<Vec<MaliciousReplicated<F>>>>
    for UpgradeContext<'a, F, NoRecord>
where
    F: Field,
    for<'u> UpgradeContext<'u, F, NoRecord>: UpgradeToMalicious<Vec<Replicated<F>>, Vec<MaliciousReplicated<F>>>,
{
    async fn upgrade(
        self,
        input: Vec<Vec<Replicated<F>>>,
    ) -> Result<Vec<Vec<MaliciousReplicated<F>>>, Error> {
        try_join_all(input.into_iter().enumerate().map(|(i, bit_shares)| {
            self.narrow(&BitOpStep::from(i)).upgrade(bit_shares)
        }))
        .await
    }
}
```

src/protocol/context/malicious.rs
```diff
@@ -470,6 +473,58 @@ where
...
+    V(usize),
+}
+impl crate::protocol::Substep for Upgrade2DVectors {}
+
+impl AsRef<str> for Upgrade2DVectors {
+    fn as_ref(&self) -> &str {
+        const COLUMN: [&str; 64] = repeat64str!["upgrade_2d"];
+
+        match self {
+            Self::V(i) => COLUMN[*i],
```
andyleiserson, line 486: This will panic at runtime if the input dimension exceeds 64. It looks like many if not all of our existing steps that are implemented this way have the same problem.

For now I guess that's okay since we know our input size is limited? But I worry that the more places this ripples to in the code, the more likely somebody overlooks this restriction and gets a confusing panic.

Maybe it would be better to require that all steps of this kind be constructed with `TryFrom` so that it's clear the construction may fail? Or at least panic on construction of the step with an out of range value, rather than later.

richajaindce:
> Is there a problem with the following? When I wrote this I used `BitOpStep` since it was convenient, but if that doesn't seem like an appropriate use of `BitOpStep`, it could also use the `Upgrade2DVectors` step that you defined.
> 
> I pulled the commit you sent me ([7350120](https://github.com/private-attribution/ipa/commit/7350120701002b9a34e4d629e7780b811148ee55)) and changed the upgrade impl to the below, and it at least fixed the `malicious_sort_in_semi_honest` test.
> 
> Functionally, I think what you have in this PR is fine, but there is some redundancy because it is manually inlining the 1D implementation within the 2D implementation.
> 
> ```rust
> #[async_trait]
> impl<'a, F> UpgradeToMalicious<Vec<Vec<Replicated<F>>>, Vec<Vec<MaliciousReplicated<F>>>>
>     for UpgradeContext<'a, F, NoRecord>
> where
>     F: Field,
>     for<'u> UpgradeContext<'u, F, NoRecord>: UpgradeToMalicious<Vec<Replicated<F>>, Vec<MaliciousReplicated<F>>>,
> {
>     async fn upgrade(
>         self,
>         input: Vec<Vec<Replicated<F>>>,
>     ) -> Result<Vec<Vec<MaliciousReplicated<F>>>, Error> {
>         try_join_all(input.into_iter().enumerate().map(|(i, bit_shares)| {
>             self.narrow(&BitOpStep::from(i)).upgrade(bit_shares)
>         }))
>         .await
>     }
> }
> ```

This means that we are creating a new step per record which I thought we don't want to do. Is creating a new step per record acceptable?

andyleiserson (approved):
After some offline discussion with @richajaindce I understand why we need this implementation.

src/protocol/context/malicious.rs
richajaindce: Agree on this. This is an assumption we are making currently in the code that steps wont be more than 64

benjaminsavage: I am OK with a panic. I think we could add a "panics" documentation to these things - but as the error would present itself immediately to the developer (and should never crop up in unexpected ways at runtime) I'd rather avoid the annoyance of the codebase having a bunch of casting littered about.

benjaminsavage (approved):
Merging, as this is functionally correct and helpful to us reaching the goal of shipping a malicious variant of IPA quite soon - but I'd like us to follow up with PRs to address the extra copying.

src/protocol/context/malicious.rs
```diff
@@ -470,6 +473,64 @@ where
...
+
+        try_join_all(zip(repeat(all_ctx), input.iter()).enumerate().map(
+            |(record_idx, (all_ctx, one_input))| async move {
+                try_join_all(zip(all_ctx, one_input).map(|(ctx, share)| async move {
+                    let ctx_ref = &ctx;
+                    self.inner
+                        .upgrade_one(
+                            ctx_ref.clone(),
+                            RecordId::from(record_idx),
+                            share.clone(),
```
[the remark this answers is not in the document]
richajaindce: Yes I have it fixed in the next PR. We should not need any clones

PR #422, opened 2023-01-30
Refactor of sort to take multi bits per record

richajaindce:
Now, convert_all_bits returns vector of 2D vectors where each 2D vector is of size [ record_count ; num_multi_bits].
This helps in code clarity and is also required for malicious sort to do upgrade

### DIFF
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -11,7 +11,7 @@ use crate::protocol::attribution::AttributionResharableStep::{
 use crate::protocol::basics::SecureMul;
 use crate::protocol::boolean::{random_bits_generator::RandomBitsGenerator, BitDecomposition};
 use crate::protocol::context::{Context, SemiHonestContext};
-use crate::protocol::modulus_conversion::transpose;
+use crate::protocol::modulus_conversion::split_into_multi_bit_slices;
 use crate::protocol::sort::apply_sort::apply_sort_permutation;
 use crate::protocol::sort::apply_sort::shuffle::Resharable;
 use crate::protocol::sort::generate_permutation::generate_permutation_and_reveal_shuffled;
@@ -169,12 +169,9 @@ pub async fn aggregate_credit<F: Field>(
     //
     // 4. Sort by `aggregation_bit`
     //
-    let sorted_output = sort_by_aggregation_bit(
-        ctx.narrow(&Step::SortByAttributionBit),
-        &aggregated_credits,
-        num_multi_bits,
-    )
-    .await?;
+    let sorted_output =
+        sort_by_aggregation_bit(ctx.narrow(&Step::SortByAttributionBit), &aggregated_credits)
+            .await?;
 
     // Take the first k elements, where k is the amount of breakdown keys.
     let result = sorted_output
@@ -253,19 +250,20 @@ async fn sort_by_breakdown_key<F: Field>(
     num_multi_bits: u32,
 ) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
     // TODO: Change breakdown_keys to use XorReplicated to avoid bit-decomposition calls
-    let breakdown_keys = transpose(
-        &bit_decompose_breakdown_key(ctx.narrow(&Step::BitDecomposeBreakdownKey), input).await?,
-    );
 
     // We only need to run a radix sort on the bits used by all possible
     // breakdown key values.
     let valid_bits_count = u128::BITS - (max_breakdown_key - 1).leading_zeros();
 
-    let sort_permutation = generate_permutation_and_reveal_shuffled(
-        ctx.narrow(&Step::GeneratePermutationByBreakdownKey),
-        &breakdown_keys[..valid_bits_count as usize],
+    let breakdown_keys = split_into_multi_bit_slices(
+        &bit_decompose_breakdown_key(ctx.narrow(&Step::BitDecomposeBreakdownKey), input).await?,
         valid_bits_count,
         num_multi_bits,
+    );
+
+    let sort_permutation = generate_permutation_and_reveal_shuffled(
+        ctx.narrow(&Step::GeneratePermutationByBreakdownKey),
+        &breakdown_keys,
     )
     .await?;
 
@@ -280,20 +278,17 @@ async fn sort_by_breakdown_key<F: Field>(
 async fn sort_by_aggregation_bit<F: Field>(
     ctx: SemiHonestContext<'_, F>,
     input: &[CappedCreditsWithAggregationBit<F>],
-    num_multi_bits: u32,
 ) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
     // Since aggregation_bit is a 1-bit share of 1 or 0, we'll just extract the
     // field and wrap it in another vector.
-    let aggregation_bits = &[input
+    let aggregation_bits = vec![input
         .iter()
-        .map(|x| x.aggregation_bit.clone())
+        .map(|x| vec![x.aggregation_bit.clone()])
         .collect::<Vec<_>>()];
 
     let sort_permutation = generate_permutation_and_reveal_shuffled(
         ctx.narrow(&Step::GeneratePermutationByAttributionBit),
-        aggregation_bits,
-        1,
-        num_multi_bits,
+        aggregation_bits.as_slice(),
     )
     .await?;
 
--- a/src/protocol/modulus_conversion/convert_shares.rs
+++ b/src/protocol/modulus_conversion/convert_shares.rs
@@ -90,16 +90,36 @@ pub fn convert_bit_local<F: Field, B: BitArray>(
 pub fn convert_all_bits_local<F: Field, B: BitArray>(
     helper_role: Role,
     input: &[XorReplicated<B>],
+    num_bits: u32,
 ) -> Vec<Vec<BitConversionTriple<Replicated<F>>>> {
-    let mut total_list = Vec::new();
-    for bit_index in 0..B::BITS {
-        let one_list = input
-            .iter()
-            .map(|v| convert_bit_local::<F, B>(helper_role, bit_index, v))
-            .collect::<Vec<_>>();
-        total_list.push(one_list);
-    }
-    total_list
+    input
+        .iter()
+        .map(move |record| {
+            (0..num_bits)
+                .map(|bit_index: u32| {
+                    let left = u128::from(record.left()[bit_index]);
+                    let right = u128::from(record.right()[bit_index]);
+                    BitConversionTriple(match helper_role {
+                        Role::H1 => [
+                            Replicated::new(F::from(left), F::ZERO),
+                            Replicated::new(F::ZERO, F::from(right)),
+                            Replicated::new(F::ZERO, F::ZERO),
+                        ],
+                        Role::H2 => [
+                            Replicated::new(F::ZERO, F::ZERO),
+                            Replicated::new(F::from(left), F::ZERO),
+                            Replicated::new(F::ZERO, F::from(right)),
+                        ],
+                        Role::H3 => [
+                            Replicated::new(F::ZERO, F::from(right)),
+                            Replicated::new(F::ZERO, F::ZERO),
+                            Replicated::new(F::from(left), F::ZERO),
+                        ],
+                    })
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>()
 }
 
 /// Convert a locally-decomposed single bit into field elements.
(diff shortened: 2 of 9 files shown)


src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -90,16 +90,36 @@ pub fn convert_bit_local<F: Field, B: BitArray>(
...
-    }
-    total_list
+    input
+        .iter()
+        .map(move |record| {
+            (0..num_bits)
+                .map(|bit_index: u32| {
+                    let left = u128::from(record.left()[bit_index]);
+                    let right = u128::from(record.right()[bit_index]);
+                    BitConversionTriple(match helper_role {
```
benjaminsavage, line 102: Why not just invoke convert_bit_local?

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -137,25 +157,36 @@ where
...
-                one_column,
-            )
-        })
-        .collect::<Vec<_>>();
-    let converted_shares = try_join_all(futures).await?;
-    Ok(converted_shares)
+    let num_records = locally_converted_bits.len();
+    try_join_all(
+        (0..num_bits)
+            .step_by(num_multi_bits.try_into().unwrap())
```
benjaminsavage, line 171: `step_by` seems like not exactly what you want here. I think you want something more like `.chunks(num_multi_bits)` 

https://docs.rs/itertools/0.7.8/itertools/trait.Itertools.html#method.chunks

Looks like we use this a few places in our project already.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -137,25 +157,36 @@ where
...
+            .map(|bit_num| {
+                let last_bit = std::cmp::min(num_multi_bits + bit_num, num_bits) as usize;
+                try_join_all(
+                    locally_converted_bits
+                        .iter()
+                        .enumerate()
+                        .map(|(idx, record)| {
+                            convert_bit_list(
+                                ctx.narrow(&ModulusConversion(bit_num))
+                                    .set_total_records(num_records),
```
benjaminsavage, line 181: I feel like we should pull these lines out of the inner loop since they do not depend on the record at all. Right now you're calling narrow num_records times, when we only need to call it once and from then on just clone the context.

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,69 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
 /// `]`
 #[must_use]
-pub fn transpose<F: Field>(input: &[Vec<Replicated<F>>]) -> Vec<Vec<Replicated<F>>> {
-    (0..input[0].len())
-        .map(|i| input.iter().map(|b| b[i].clone()).collect::<Vec<_>>())
+pub fn split_into_multi_bit_slices<F: Field>(
+    input: &[Vec<Replicated<F>>],
+    num_bits: u32,
+    num_multi_bits: u32,
+) -> Vec<Vec<Vec<Replicated<F>>>> {
```
benjaminsavage, line 37: You might not want to do it all at once. 

When we upgrade to malicious, each batch of 3 columns will need to get upgraded to malicious as a part of a different validation phase (they'll use a different value of the random sharing of `r`) as they're inputs to different circuits, each of which is processed after a reveal.

Consider just emitting one batch at a time. It'll have the added benefit of simplifying the type signature.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -137,25 +157,36 @@ where
 pub async fn convert_all_bits<F, C, S>(
     ctx: &C,
     locally_converted_bits: &[Vec<BitConversionTriple<S>>],
-) -> Result<Vec<Vec<S>>, Error>
+    num_bits: u32,
+    num_multi_bits: u32,
+) -> Result<Vec<Vec<Vec<S>>>, Error>
```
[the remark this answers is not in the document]
benjaminsavage: You might not want to do it all at once. 

When we upgrade to malicious, each batch of 3 columns will need to get upgraded to malicious as a part of a different validation phase (they'll use a different value of the random sharing of `r`) as they're inputs to different circuits, each of which is processed after a reveal.

Consider just emitting one batch at a time. It'll have the added benefit of simplifying the type signature.

src/protocol/sort/generate_permutation.rs
```diff
@@ -446,46 +438,4 @@ mod tests {
             perms_and_randoms[1].randoms_for_shuffle.1
         );
     }
-
-    #[tokio::test]
-    pub async fn malicious_sort_in_semi_honest() {
```
benjaminsavage, line 451: Why is this test just gone?

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -43,29 +44,22 @@ use futures::future::try_join_all;
...
-    .await?;
+    let lsb_permutation =
+        multi_bit_permutation(ctx_0.narrow(&BitPermutationStep), &sort_keys[0]).await?;
 
     let input_len = u32::try_from(sort_keys[0].len()).unwrap(); // safe, we don't sort more that 1B rows
 
     let mut composed_less_significant_bits_permutation = lsb_permutation;
-    for bit_num in (num_multi_bits..num_bits).step_by(num_multi_bits.try_into().unwrap()) {
-        let ctx_bit = ctx.narrow(&Sort(bit_num));
+    for (bit_num, one_slice) in sort_keys.iter().enumerate().skip(1) {
```
benjaminsavage, line 61: `bit_num` isn't a good variable name, because you're enumerating the "slices", so it's like a "chunk_number" now.

Also, `one_slice` is an odd name. Why one? Perhaps `next_few_bits`?

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -46,7 +49,7 @@ pub async fn multi_bit_permutation<
     let equality_checks = try_join_all(
         (0..num_records)
             .zip(repeat(ctx.set_total_records(num_records)))
-            .map(|(i, ctx)| check_everything(ctx, i, input)),
+            .map(|(i, ctx)| check_everything(ctx, i, input[i].as_slice())),
     )
```
benjaminsavage, line 53: I'm surprised clippy didn't yell at you about this loop. When I write code like this, it usually prefers me to just loop over the thing I'm indexing into, rather than loop over an int, and just use that int to index into stuff.

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -94,20 +97,15 @@ pub async fn multi_bit_permutation<
 /// `2^N` possible values this could have.
 /// This function checks all of these possible values, and returns a vector of secret-shared results.
 /// Only one result will be a secret-sharing of one, all of the others will be secret-sharings of zero.
-async fn check_everything<F, C, S>(
-    ctx: C,
-    record_idx: usize,
-    input: &[Vec<S>],
```
benjaminsavage, line 100: This is amazing. So much cleaner than passing the whole damn thing in!

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -190,7 +188,7 @@ where
 async fn pregenerate_all_combinations<F, C, S>(
     ctx: C,
     record_idx: usize,
-    input: &[Vec<S>],
```
benjaminsavage, line 193: A huge improvement.

benjaminsavage (changes requested):
I like this diff. It's pretty close. I think with a few minor changes it's ready to go.

akoshelev:
I am in favor of pushing it as is because we need to get some numbers for malicious sort and trying to optimize the vectors out later (maybe next week, when we are all going to be in the same place). I still worry about the cost of 3d vectors, but don't want to hold this change because of it

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Missed it in multiple iterations. Changing it

src/protocol/modulus_conversion/convert_shares.rs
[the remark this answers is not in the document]
richajaindce: Makes sense. Lets do this in future PRs. It might be a bit of refactoring needed.

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Agreed, changing this

src/protocol/sort/generate_permutation.rs
richajaindce: Since we changed the function output of convert_all_bits, we would need to take extra effort to maintain this test. This test is going to come back in PR with malicious sort.

src/protocol/sort/generate_permutation_opt.rs
richajaindce: I was thinking this is one slice which will be sent to bit permutation at a time. Changing it

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -85,21 +79,13 @@ where
...
-                        sort_keys[idx as usize].clone(),
-                        (randoms_for_shuffle0, randoms_for_shuffle1),
-                        revealed,
-                    )
-                    .await
-                });
-        let next_few_bits_sorted_by_less_significant_bits = try_join_all(futures).await?;
+        let next_few_bits_sorted_by_less_significant_bits = secureapplyinv_multi(
+            ctx_bit.narrow(&MultiApplyInv(bit_num.try_into().unwrap())),
+            one_slice.clone(),
```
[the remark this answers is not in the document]
richajaindce: Cannot since secureapplyinv does in-place processing hence needs a copy of data.

src/protocol/sort/multi_bit_permutation.rs
richajaindce: Changed it

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: chunks needs a vector to work on (and not a range) so needed to collect the bit nums before calling it. Let me know if it looks ok

src/protocol/modulus_conversion/mod.rs
```diff
@@ -16,17 +16,69 @@ use crate::{ff::Field, secret_sharing::replicated::semi_honest::AdditiveShare as
...
+///     `[ row[0].bitL+1, ..., row[0].bit2L ], [ row[1].bitL+1, ..., row[1].bit2L ], .. [ row[n].bitL+1, ..., row[n].bit2L ], `,
 ///     ...
-///     `[ row[0].bit31, row[1].bit31, ..., row[n].bit31 ]`,
+///     `[ row[0].bitmL,  ..., row[0].bit31 ], [ row[1].bitmL, ..., row[n].bit31 ], .. [ row[n].bitmL, ..., row[n].bit31 ]`,
 /// `]`
 #[must_use]
-pub fn transpose<F: Field>(input: &[Vec<Replicated<F>>]) -> Vec<Vec<Replicated<F>>> {
-    (0..input[0].len())
-        .map(|i| input.iter().map(|b| b[i].clone()).collect::<Vec<_>>())
+pub fn split_into_multi_bit_slices<F: Field>(
```
richajaindce, line 33: I know this is going away in @taikiy's PR. Based on who lands the code first, will remove this.

src/protocol/modulus_conversion/convert_shares.rs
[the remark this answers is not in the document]
benjaminsavage: That works for me.

src/protocol/modulus_conversion/convert_shares.rs
benjaminsavage: Oh, I see. Does calling "collect" result in instantiating a new vector?

src/protocol/sort/generate_permutation.rs
benjaminsavage: OK

src/protocol/sort/generate_permutation_opt.rs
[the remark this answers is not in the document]
benjaminsavage: OK

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -253,19 +250,20 @@ async fn sort_by_breakdown_key<F: Field>(
     num_multi_bits: u32,
 ) -> Result<Vec<CappedCreditsWithAggregationBit<F>>, Error> {
     // TODO: Change breakdown_keys to use XorReplicated to avoid bit-decomposition calls
-    let breakdown_keys = transpose(
-        &bit_decompose_breakdown_key(ctx.narrow(&Step::BitDecomposeBreakdownKey), input).await?,
-    );
```
benjaminsavage, line 258: Huge win.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -90,16 +90,16 @@ pub fn convert_bit_local<F: Field, B: BitArray>(
...
-        let one_list = input
-            .iter()
-            .map(|v| convert_bit_local::<F, B>(helper_role, bit_index, v))
-            .collect::<Vec<_>>();
-        total_list.push(one_list);
-    }
-    total_list
+    input
+        .iter()
+        .map(move |record| {
```
benjaminsavage, line 97: I think you want to use an "into_iter()" in a case like this - but I think it's the same?

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Yes it does

src/protocol/modulus_conversion/convert_shares.rs
richajaindce: Yes, clippy wants it to be iter()

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -174,7 +182,14 @@ where
     try_join_all(
         zip(repeat(ctx), locally_converted_bits.iter())
             .enumerate()
-            .map(|(i, (ctx, row))| async move { convert_bit(ctx, RecordId::from(i), row).await }),
+            .map(|(i, (ctx, bit))| async move {
+                convert_bit(
+                    ctx.narrow(&ModulusConversion(i.try_into().unwrap())),
```
benjaminsavage, line 187: Can you make the `ModulusConversion(` accept whatever this type is in order to avoid all of this casting?

src/protocol/sort/generate_permutation_opt.rs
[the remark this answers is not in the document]
benjaminsavage: But I don't see `one_slice` being used for anything else. Do we need it for something else?

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -236,11 +235,15 @@ mod tests {
...
-        &[0, 0, 1, 0, 1, 0],
-        &[0, 1, 1, 0, 0, 0],
-        &[1, 0, 1, 0, 1, 0],
+    const INPUT: [[u128; 3]; 6] = [
+        [0, 0, 1],
+        [0, 1, 0],
+        [1, 1, 1],
+        [0, 0, 0],
+        [1, 0, 1],
+        [0, 0, 0],
```
benjaminsavage, line 244: Can you add a comment above this saying "This is in Little-Endian format"? I gets me every time.

benjaminsavage (approved):
This is pretty good. I think it's best to land it right now, and follow up with smaller PRs to improve on the various little nits that I've commented on.

PR #424, opened 2023-01-30
improve safety for indexed protocol steps

andyleiserson:
This creates some guardrails around construction of indexed protocol steps.

I didn't convert `BitOpStep` from `from` to `try_from` since there are a number of places it needs to change, but I can do that if desired (after there's agreement on the general approach).

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -15,6 +15,7 @@ use crate::{
 use std::fmt::{Debug, Formatter};
 use std::hash::Hash;
 use std::ops::AddAssign;
+use std::ops::Index;
 
 pub type MatchKey = BitArray40;
 pub type BreakdownKey = BitArray8;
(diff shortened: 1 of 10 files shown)


src/protocol/mod.rs
```diff
@@ -123,6 +124,45 @@ macro_rules! repeat64str {
     }
 }
 
+/// Bounds-checked index type for protocol steps indexed 0..64.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
akoshelev, line 128: how do you keep them sorted? do you have to do it manually or there is a magic cargo command I could use to have the same result?

src/protocol/mod.rs
```diff
@@ -123,6 +124,45 @@ macro_rules! repeat64str {
...
+                    _ => Err(StepIndexFromIntError),
+                }
+            }
+        }
+    };
+}
+
+step_index_try_from!(usize);
+step_index_try_from!(u32);
+step_index_try_from!(i32);
```
akoshelev, line 164: it is probably unlikely that we need this except in tests?

akoshelev (approved):
I like it - I am always in favour of using fallible conversions if they can fail

src/protocol/mod.rs
martinthomson: When I last looked at this, I discovered that ordering matters.  That is, `cargo fmt` wasn't able to reorder these.  (I'm surprised that Eq and PartialEq can appear in this order, but maybe my info is old.)

andyleiserson: I just happened to write it in sorted order without thinking about Eq / PartialEq, I guess that must have changed recently. There are a few places in the code that list them in Eq, PartialEq order now.

Maybe cargo fmt will start sorting them at some point.

src/protocol/mod.rs
andyleiserson: It was used for numeric literals (not just in tests), but that can be fixed by specifying `u32` suffix on the literals, which seems like the right tradeoff given that a negative step index is nonsensical.

src/protocol/mod.rs
martinthomson: Yeah, they must have worked out how to address the ordering problem in the compiler, which is pretty good if it means we can eliminate variance at some point in the future.

akoshelev: @martinthomson can you point me to this issue if you have it handy? I am curious what could be the problem.

PR #425, opened 2023-01-31
Malicious sort now uses multi bit permutations optimization

richajaindce:
This is stacked on top of #421


src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -81,6 +82,25 @@ where
     try_join_all(reshares).await
 }
 
+#[async_trait]
+impl<F: Field> Resharable<F> for Vec<MaliciousReplicated<F>> {
```
benjaminsavage, line 86: It feels like we shouldn't have to special case this. It feels like we should just have one implementation for ANY vector of Reshareables. Maybe we can add that later.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
     Ok(composed_less_significant_bits_permutation)
 }
 
+#[allow(dead_code)]
+#[embed_doc_image("malicious_sort", "images/sort/malicious-sort.png")]
+/// Returns a sort permutation in a malicious context.
+/// This runs sort in a malicious context. The caller is responsible to validate the accumulater contents and downgrade context to Semi-honest before calling this function
```
benjaminsavage, line 123: spelling: "accumulator"

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+/// If sort keys dont have num of bits same as `num_bits`
+/// # Errors
+pub async fn malicious_generate_permutation_opt<'a, F>(
+    sh_ctx: SemiHonestContext<'a, F>,
+    sort_keys: &[Vec<Vec<Replicated<F>>>],
+) -> Result<(MaliciousValidator<'a, F>, Vec<MaliciousReplicated<F>>), Error>
+where
+    F: Field,
+{
+    let mut malicious_validator = MaliciousValidator::new(sh_ctx.narrow(&Sort(0)));
```
benjaminsavage, line 156: Why do we need to narrow here? It seems like this context is not really the one used for `Sort(0)`, as we narrow again before calling `multi_bit_permutation` for the first time. Is one of those un-necessary, or should this be renamed to indicate something else?

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+    let mut m_ctx_bit = malicious_validator.context();
+    let input_len = u32::try_from(sort_keys[0].len()).unwrap(); // safe, we don't sort more than 1B rows
+
+    let upgraded_sort_keys = m_ctx_bit.upgrade(sort_keys[0].clone()).await?;
```
benjaminsavage, line 160: We never use sort_keys[0] ever again - so can we avoid this `clone()` somehow?

src/protocol/sort/apply_sort/shuffle.rs
richajaindce: Agree on this. Even before that, is it possible to share Resharable for MaliciousReplicated and Replicated? I tried but couldnt get it to working @akoshelev

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+///    i. n-1th composition: This is the permutation which sorts the inputs
+///   ii. Validator which can be used to validate the leftover items in the accumulator
+///
+/// ![Malicious sort permutation steps][malicious_sort]
+/// # Panics
+/// If sort keys dont have num of bits same as `num_bits`
+/// # Errors
+pub async fn malicious_generate_permutation_opt<'a, F>(
+    sh_ctx: SemiHonestContext<'a, F>,
+    sort_keys: &[Vec<Vec<Replicated<F>>>],
```
benjaminsavage, line 151: This type signature is unhelpful. It doesn't inform the reader that we are expecting a highly specific format.

I'd really like to get some kind of compile-time assistance on this one. Could we somehow make a few types that at compile-time are known to be of specific lengths? For example, a "chunk" could internally be implemented as a vector of arrays of length 3, if the compiler just knew that the number of multi-bits was 3. That would also make for a much more informative type. I'm not sure exactly how to achieve this, but imagine if this was something like our BitArray40 type, where the size is known at compile time. Then we could have something like:

```
pub async fn malicious_generate_permutation_opt<'a, F, Ch>(
    sh_ctx: SemiHonestContext<'a, F>,
    sort_keys: &[Vec<Ch>],
) -> blah blah
where
    F: Field,
    Ch: MultiBitChunk<F>,
{
```

Where we could have various types that all implement `MultiBitChunk`, one of which would be like `ThreeBitChunk` or something.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+        let revealed_and_random_permutations = malicious_shuffle_and_reveal_permutation(
+            m_ctx_bit.narrow(&ShuffleRevealPermutation),
+            input_len,
+            composed_less_significant_bits_permutation,
+            malicious_validator,
+        )
+        .await?;
+
+        malicious_validator =
+            MaliciousValidator::new(sh_ctx.narrow(&Sort(chunk_num.try_into().unwrap())));
```
benjaminsavage, line 175: I'm annoyed that we need this casting. What's `chunk_num`, a usize? Whatever it is, can we just make the `Sort()` enum accept this type?

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+        m_ctx_bit = malicious_validator.context();
+        let upgraded_sort_keys = m_ctx_bit.upgrade(chunk.clone()).await?;
```
benjaminsavage, line 177: For a follow-up PR:

I *think* it would be more efficient to wait to convert from XorReplicated sharings until *this point in time*. But let's revisit this later on.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+            revealed_and_random_permutations
+                .randoms_for_shuffle
+                .0
+                .as_slice(),
+            revealed_and_random_permutations
+                .randoms_for_shuffle
+                .1
+                .as_slice(),
+            revealed_and_random_permutations.revealed.as_slice(),
+        );
```
benjaminsavage, line 189: There has got to be a more compact way to do this de-structuring...

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -113,6 +117,104 @@ where
...
+
+        let next_few_bits_sorted_by_less_significant_bits = secureapplyinv_multi(
+            m_ctx_bit.narrow(&MultiApplyInv(chunk_num.try_into().unwrap())),
+            upgraded_sort_keys,
```
benjaminsavage, line 193: Actually, it might even be more efficient to apply this sort permutation to XorReplicated sharings, and convert them to a Vec<MaliciousReplicated<F>> **after** this step, as the re-shares will be cheaper for XorReplicated sharings.

martinthomson: Would resharing be vulnerable to attack in that case?

benjaminsavage: I *think* our malicious re-share doesn't depend on anything except the ability to add and subtract. The same goes for the malicious reshare. We should be able to detect additive errors. But we need to researchers to help us look into that.

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Cannot since we just have a reference to the Vec<Vec>> and upgrade needs a copy of input to work on

src/protocol/sort/generate_permutation_opt.rs
richajaindce: agreed, this should be doable

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Since it is stored in a struct, it looks like this and we cant destructure like a simple tuple. Should we change it to a tuple instead? That would make the readability a bit hard though

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Aren't we resharing to avoid identifying shares from before?

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Added a todo for this

src/protocol/sort/generate_permutation_opt.rs
richajaindce: Actually the chunks are all same size apart from the last one. e.g. if num_bits = 40 and multi bits = 3, then we have 13 * [;3] and 1 [;1]. So we would need to handle that scenario as well. Lets work on this in follow up PRs?

src/protocol/sort/apply_sort/shuffle.rs
richajaindce: Thanks to @taikiy this works now!!

src/protocol/context/malicious.rs
```diff
@@ -492,37 +492,32 @@ impl AsRef<str> for Upgrade2DVectors {
...
-        try_join_all(zip(repeat(all_ctx), input.iter()).enumerate().map(
+        try_join_all(zip(repeat(all_ctx), input.into_iter()).enumerate().map(
             |(record_idx, (all_ctx, one_input))| async move {
                 try_join_all(zip(all_ctx, one_input).map(|(ctx, share)| async move {
-                    let ctx_ref = &ctx;
-                    self.inner
-                        .upgrade_one(
-                            ctx_ref.clone(),
-                            RecordId::from(record_idx),
-                            share.clone(),
```
benjaminsavage, line 522: I *LOVE* that you were able to remove these `clone()` calls!!!!

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -47,8 +46,8 @@ impl From<usize> for InnerVectorElementStep {
 }
 
 #[async_trait]
-impl<F: Field> Resharable<F> for Vec<Replicated<F>> {
-    type Share = Replicated<F>;
+impl<T: Arithmetic<F>, F: Field> Resharable<F> for Vec<T> {
```
benjaminsavage, line 49: Love this.

src/protocol/sort/generate_permutation.rs
```diff
@@ -158,7 +158,7 @@ where
 
     let mut composed_less_significant_bits_permutation = bit_0_permutation;
     for bit_num in 1..num_bits {
-        let ctx_bit = ctx.narrow(&Sort(bit_num));
+        let ctx_bit = ctx.narrow(&Sort(bit_num.try_into().unwrap()));
```
benjaminsavage, line 161: We shouldn't need this casting now, right? `Sort()` accepts a usize now.  Isn't that what the type of `bit_num` is?

src/protocol/sort/generate_permutation.rs
```diff
@@ -288,7 +288,8 @@ where
         )
         .await?;
 
-        malicious_validator = MaliciousValidator::new(sh_ctx.narrow(&Sort(bit_num)));
+        malicious_validator =
+            MaliciousValidator::new(sh_ctx.narrow(&Sort(bit_num.try_into().unwrap())));
```
benjaminsavage, line 292: I still don't understand why we need to cast here.

src/protocol/sort/generate_permutation_opt.rs
benjaminsavage: Let's just leave it for now.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -169,4 +277,48 @@ mod tests {
...
+
+        let [(v0, result0), (v1, result1), (v2, result2)] = world
+            .semi_honest(
+                match_keys.clone(),
+                |ctx: SemiHonestContext<Fp31>, mk_shares| async move {
+                    let local_lists = convert_all_bits_local(ctx.role(), &mk_shares);
+                    let converted_shares =
+                        convert_all_bits(&ctx, &local_lists, BitArray40::BITS, NUM_MULTI_BITS)
+                            .await
+                            .unwrap();
```
benjaminsavage, line 303: This conversion wasn't validated... but it's a test for `malicious_generate_permutation_opt`, so I guess that doesn't really matter.

But in the actual IPA protocol we will need to validate this upgrade.

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -169,4 +277,48 @@ mod tests {
...
+                        .unwrap()
+                },
+            )
+            .await;
+        let result = join3(
+            v0.validate(result0),
+            v1.validate(result1),
+            v2.validate(result2),
+        )
+        .await;
```
benjaminsavage, line 316: Can't we call "Reconstruct" or something?

benjaminsavage (approved):
LGTM! Thanks for making all of these suggested changes!

src/protocol/sort/generate_permutation.rs
richajaindce: This is the original generate_permutation without any optimization. It somehow works best with u32's in the code. Since we will be eliminating this code anyways, so didnt bother changing it here

src/protocol/sort/generate_permutation.rs
richajaindce: This code will be purged in favour of multi bit so, please ignore it for now

src/protocol/sort/generate_permutation_opt.rs
richajaindce: The validate will be run before each reveal or downgrade of context right?

PR #427, opened 2023-01-31
[Aggregate Protocol] Remove unnecessary modulus conversion

taikiy:
In the aggregation protocol (last step in the attribution), we append new rows with unique breakdown keys to the input where we store aggregated credits for each breakdown key. Because these values are publicly known, we can skip the modulus conversion in MPC and directly convert them to bitwise shares of F.

### DIFF
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -5,20 +5,16 @@ use super::{
     },
     InteractionPatternStep,
 };
+use crate::bits::BitArray;
 use crate::error::Error;
 use crate::ff::Field;
 use crate::protocol::basics::SecureMul;
 use crate::protocol::context::{Context, SemiHonestContext};
-use crate::protocol::modulus_conversion::{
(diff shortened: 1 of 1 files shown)


src/protocol/attribution/aggregate_credit.rs
```diff
@@ -160,41 +150,38 @@ pub async fn aggregate_credit<F: Field, BK: BitArray>(
...
-            helper_bit: zero.clone(),
-            aggregation_bit: zero.clone(),
-            credit: zero.clone(),
+    let mut unique_breakdown_keys = (0..max_breakdown_key)
+        .map(|i| {
+            // Since these breakdown keys are publicly known, we can directly convert them to Vec<Replicated<F>>
+            let bk_bits = BK::truncate_from(i);
+            let converted_bk = (0..BK::BITS)
+                .map(|i| {
+                    if bk_bits[i] {
```
taikiy, line 171: Indexing a `BitArray` value returns `bool`.

PR #430, opened 2023-02-01
[WIP] Malicious IPA - upgraded until first sort is called

richajaindce:


src/protocol/ipa/mod.rs
```diff
@@ -321,6 +453,56 @@ pub mod tests {
         }
     }
 
+    #[tokio::test]
+    #[allow(clippy::missing_panics_doc)]
+    pub async fn malicious_wip() {
```
martinthomson, line 458:
```suggestion
    async fn malicious_wip() {
```

benjaminsavage (approved):
Cool! Let's convert this to malicious, one step at a time!

PR #431, opened 2023-02-01
Unordering receiver

martinthomson:
OK, the next small increment in my journey to smooth the flow of data between protocol and network.

This one takes a stream of data off the network and lets callers receive at different indices in any order.

You might have 10 messages, with receivers 0 through 9 running in different tasks.  Each can call `recv()` with their index (i.e., `RecordId`) in any order, from any thread.  The data will be read off as it arrives and delivered to each in order.

The form of the `Stream` that this consumes isn't *exactly* the same as the one that hyper produces from its requests, but a simple adapter should work for that.

This isn't the whole of the work involved, but it's the fiddly part.  From here, there is a bunch of plumbing to go.  I might be able to split off the code for matching incoming streams to steps, but the rest might have to go all together, because the integration starts to get harder to do in parts.


src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+    /// Wake the waker from the next future, if the next receiver has been polled.
+    fn wake_next(&mut self) {
+        self.next += 1;
+        let index = self.next % self.wakers.len();
+        if let Some(w) = self.wakers[index].take() {
+            w.wake();
+        }
+        // See the documentation for `overflow_wakers`.
+        if self.next % (self.wakers.len() / 2) == 0 {
+            for w in self.overflow_wakers.drain(..) {
```
akoshelev, line 202: I think `drain` never shrinks the allocated capacity so the memory footprint may be less than ideal. Is it reasonable to assume the average capacity to be equal to `self.wakers.capacity()` and set it after each drain?

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+    marker::PhantomData,
+    num::NonZeroUsize,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{Context, Poll},
+};
+use typenum::Unsigned;
+
+/// A future for receiving item `i` from an `UnorderedReceiver`.
+#[pin_project]
```
akoshelev, line 18: It feels that you don't need to pin anything. `usize` is `Unpin` and `OperatingState` pins the stream

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+        // We don't save a waker at `self.next`, so `>` and not `>=`.
+        if i > self.next + self.wakers.len() {
+            self.overflow_wakers.push(waker);
+        } else {
+            assert!(
+                i > self.next,
+                "Awaiting a read that has already been fulfilled"
+            );
+            let index = i % self.wakers.len();
+            self.wakers[index] = Some(waker);
```
akoshelev, line 189: maybe another debug assert that `self.wakers[index] == None`?

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+    /// aren't ready at the time of the call.  If the future is invoked prior
+    /// to the value being ready, the `Waker` is saved here.
+    wakers: Vec<Option<Waker>>,
+    /// If we ever find that a waker doesn't fit in `wakers`, this is where
+    /// they are stashed.  These are more than `c` items past the current
+    /// item into the future when registered (c = capacity or `wakers.len()`).
+    /// So we don't want to wake them frequently.  Instead, these are woken on
+    /// a fixed cadence of every `c/2` items.
+    ///
+    /// When polled again, any that were only a little past the capacity will
```
akoshelev, line 130: is it worth considering a future optimization where we only poll overflow wakers when there is nothing to be polled in `wakers`?

akoshelev (commented):
very elegant solution!

src/bits/mod.rs
```diff
@@ -71,5 +71,5 @@ pub trait Serializable: Sized {
     /// buffer has enough capacity to fit instances of this trait.
     ///
     /// [`serialize`]: Self::serialize
-    fn deserialize(buf: GenericArray<u8, Self::Size>) -> Self;
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Self;
```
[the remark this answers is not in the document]
martinthomson: We had a few cases where we used `GenericArray::clone_from_slice` in order to produce this value.  That involved a copy, which wasn't necessary, whereas `GenericArray::from_slice` doesn't.  But that returns a reference.

src/helpers/buffers/unordered_receiver.rs
martinthomson: Hmm, I added this early on and didn't come back to it (what you see is not what I started with).  I'll take another look.

src/helpers/buffers/unordered_receiver.rs
martinthomson: That might be better, the more I think on it.  I was concerned that we might have to walk the main store to ensure that things were fine, but we can count the number of wakers that are in there.  The other cost is that we might flush the extras too often, but I think that only happens after you finish reading the last item in a chunk in most cases.  With that design, any time you read an item out and there are no outstanding receivers, you will flush the overflow.  That will flush more often than this, probably, but it might be simpler to reason about.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+    ///
+    /// # Panics
+    ///
+    /// If `i` is for an message that has already been read.
+    ///
+    /// [`recv`]: UnorderedReceiver::recv
+    /// [`poll`]: Future::poll
+    fn add_waker(&mut self, i: usize, waker: Waker) {
+        // We don't save a waker at `self.next`, so `>` and not `>=`.
+        if i > self.next + self.wakers.len() {
```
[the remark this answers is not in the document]
martinthomson: My understanding is that capacity is a measure of the size of the underlying store, which could be larger than what you asked for.  If the allocation is larger, then this code will be incorrect.

src/helpers/buffers/unordered_receiver.rs
martinthomson: That's not going to be the case.  I originally had that and found that there are some cases (the synchronous test is one), where we replace a waker.  Basically, if the future is polled more than once in this state, we will be getting another waker for the same item.  I think that a strict interpretation of the `Future` contract requires us to store both wakers in that case, but the note above explains why I think that's unnecessary.

(I spent a lot of time thinking about this and I'm somewhat sure that this is not OK in theory, but maybe OK in practice.  That is, what if the first to poll is kept, the second is polled and discarded? We won't wake the first.  But then, we won't be doing that, I hope.)

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+
+    /// Wake the waker from the next future, if the next receiver has been polled.
+    fn wake_next(&mut self) {
+        self.next += 1;
+        let index = self.next % self.wakers.len();
+        if let Some(w) = self.wakers[index].take() {
+            w.wake();
+        }
+        // See the documentation for `overflow_wakers`.
+        if self.next % (self.wakers.len() / 2) == 0 {
```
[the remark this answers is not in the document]
martinthomson: Well, *this* design depends on this being done unconditionally, but I think I'm going to try your suggested design instead.

src/helpers/buffers/unordered_receiver.rs
martinthomson: Yeah, I might instead use `std::mem::take()` here.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,502 @@
...
+    S: Stream<Item = C>,
+    C: AsRef<[u8]>,
+{
+    /// Wrap a stream for unordered reading.
+    ///
+    /// The capacity here determines how far ahead a read can be.  In most cases,
+    /// this should be the same as the value given to [`ordering_mpsc`].
+    ///
+    /// [`ordering_mpsc`]: crate::helpers::buffers::ordering_mpsc::ordering_mpsc
+    pub fn new(stream: Pin<Box<S>>, capacity: NonZeroUsize) -> Self {
```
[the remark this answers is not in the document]
martinthomson: I don't think that makes this easier in this case.  It would make it less flexible for one (this allows you to read different-sized messages).  This basically takes that functionality and implements it internally, which should be more efficient overall.

src/bits/bit_array.rs
```diff
@@ -10,8 +10,17 @@ type U8_1 = BitArr!(for 8, in u8, Lsb0);
...
+/// macro processor.  This noop ensures that the instance of `GenericArray` used
+/// is `Copy`.  It should be - it's the same size as the `BitArray` instance.
+fn assert_copy<C: Copy>(c: C) -> C {
+    c
+}
+
 macro_rules! bit_array_impl {
     ( $modname:ident, $name:ident, $store:ty, $bits:expr, $arraylen:ty ) => {
+        // $store
+
```
martinthomson, line 23:
```suggestion
```

src/bits/mod.rs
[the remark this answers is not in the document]
akoshelev: that's right. I don't mind changing it this way, just saying that for places where I believe currently deserialize is "free" (bitarray) it will come with a cost of cloning.

martinthomson:
OK, I just implemented a change to track wakers differently and hit a deadlock.

The logic I implemented looks like this:

```rust
if let Some(w) = self.wakers[self.next % N].take() {
  w.wake();
  self.active_wakers -= 1;
} else if self.active_wakers == 0 {
  for w in std::mem::take(&mut self.overflow_wakers) { 
    w.wake();
  }
}
```

That is, if we ever find that there are no wakers left in the active set, flush the overflow.  By reactivating those, you will get more things to work on.

The problem that leads to a deadlock is where you get a gap.

Let's say that you wake item N, which is the last item in the active store.  Item N+1 is in the overflow, but you know that item N is coming back soon, so you don't bother flushing out the store just yet.

Before item N is polled again, item N+2 is polled.  This is put into the active store.  Now, when item N is polled again, the store is not empty, but item N+1 is stuck in the overflow.

I conclude from this that you have to flush the overflow any time that it is not empty and the next slot in the active store is empty.  That's much less efficient.

I'm going to go back to the c/2 design I started out with :)  Thanks to shuttle again for finding this one (I'm also glad I used a capacity of 3 rather than a capacity of 2, because this only happens for c>2).

src/helpers/buffers/unordered_receiver.rs
akoshelev: > I was concerned that we might have to walk the main store to ensure that things were fine, but we can count the number of wakers that are in there.

yea I was also thinking about maintaining a counter of active wakers inside `wakers`. that shouldn't require walking through the vector and counting

src/bits/mod.rs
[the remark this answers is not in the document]
martinthomson: Yeah, though I realized that this is what we really want in those cases.  We want to own the copy of whatever came off the wire.  I also discovered that those cases can rely on the type being `Copy`, so it's cheap.

src/helpers/buffers/unordered_receiver.rs
martinthomson: I've got a check for this that should ensure that it is robust.  Not guaranteed (`Waker::will_wake` can return false in cases where there is no issue), but then we shouldn't be hitting that code path outside of the few tests I have.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -0,0 +1,504 @@
...
+impl<S, C, M> Future for Receiver<S, C, M>
+where
+    S: Stream<Item = C>,
+    C: AsRef<[u8]>,
+    M: Message,
+{
+    type Output = Result<M, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.as_ref();
```
akoshelev, line 38: you could use `self` because pin implements `Deref`

src/helpers/buffers/unordered_receiver.rs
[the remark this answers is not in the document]
akoshelev: right, I thought if len() != capacity() this code will be either incorrect or super inefficient, potentially storing everything but one item inside the overflow wakers.

src/helpers/buffers/unordered_receiver.rs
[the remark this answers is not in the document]
akoshelev: ah I think I got it now. You wake the overflow wakers not just because there might be a chance that their future is ready, but also because you want to move some of them to the `wakers` vector. In this case you have to wake them up on schedule

akoshelev (approved):

PR #432, opened 2023-02-01
Split out some of the longer-running tasks

martinthomson:
An additional workflow can run in parallel.  This increases overall work, but it might allow us to reduce the overall time to getting a positive answer.


martinthomson:
Concurrency tests seem to be the long pole here.  I want to get them running alongside the regular tests.

akoshelev:
I could tune them a bit to make them run faster. Every single improvement in IPA makes concurrency tests better too :)

akoshelev (approved):

PR #434, opened 2023-02-01
Run cargo audit directly

### LINKED ISSUE #433: actions-rs actions are effectively unmaintained
We're getting warnings about node 12 deprecation, so these will stop working soon-ish.

It looks like we can just run `cargo audit` for ourselves.  We don't get the fancy reporting and whatnot, but that's probably tolerable.  We can also probably consolidate the audits; the steps taken are the same.

martinthomson:
Closes #433.


.github/workflows/audit.yml
```diff
@@ -19,6 +20,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/audit-check@v1
+      - uses: dtolnay/rust-toolchain@stable
```
akoshelev, line 23: nice!

akoshelev (approved):

.github/workflows/audit.yml
akoshelev: well, not so nice, as it does not seem to work :(

PR #435, opened 2023-02-01
Client tests

thurstonsand:
expand on the client-side tests for the `HttpTransport` impl.


src/helpers/network.rs
```diff
@@ -151,3 +151,122 @@ impl<T: Transport> Network<T> {
...
+    }
+
+    #[tokio::test]
+    async fn successfully_sends() {
+        let message_chunks = (
+            ChannelId::new(Role::H1, Step::default().narrow("successfully-sends")),
+            vec![0u8; MESSAGE_PAYLOAD_SIZE_BYTES],
+        );
+        let roles = RoleAssignment::new(HelperIdentity::make_three());
+        let this_identity = roles.identity(Role::H2);
```
akoshelev, line 206: nit: instead of this and that, would be easier to read if you name it `h1_identity`, `h2_identity`

src/helpers/network.rs
```diff
@@ -151,3 +151,122 @@ impl<T: Transport> Network<T> {
...
+            )
+            .await;
+        }
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "out-of-order delivery of data for")] // just the prefix
+    async fn rejects_bad_offset() {
+        let expected_query_id = QueryId;
+        let message_chunks = (
```
akoshelev, line 234: this usage of message chunks looks very confusing to me. It took me some time to realize that you use payload on the send side and channel id on the receiving side. It would make code easier if you could do something like this

```rust
   let sample_payload = [...];
   let h1_transport = transports.transport(...);
   let h2_transport = ...

   let h1_input_stream = h1_network.recv_stream().await;
   h2_transport.send(h1, command).await.unwrap();
   assert_eq((ChannelId::new(h2, step), sample_payload), h1_input_stream.next().await.unwrap());
   ....
```

src/net/client/mod.rs
```diff
@@ -67,27 +65,31 @@ impl MpcHelperClient {
...
-            .build()?)
-    }
-
     /// Responds with whatever input is passed to it
     /// # Errors
     /// If the request has illegal arguments, or fails to deliver to helper
-    pub async fn echo(&self, s: &str) -> Result<Vec<u8>, Error> {
-        let uri = self.build_uri(http_serde::echo::uri(s))?;
+    pub async fn echo(&self, s: &str) -> Result<String, Error> {
+        const FOO: &str = "foo";
```
akoshelev, line 72: It seems that it's been reviewed already. Is it the case? can you maybe rebase it on top of e2e branch?

thurstonsand: it's rebased now

src/helpers/network.rs
thurstonsand: alright, i split it apart

src/helpers/network.rs
thurstonsand: done

src/net/client/mod.rs
```diff
@@ -188,3 +188,293 @@ impl MpcHelperClient {
...
+                step,
+                payload,
+                offset,
+            } => {
+                assert_eq!(query_id, expected_query_id);
+                assert_eq!(step, expected_step);
+                assert_eq!(payload, expected_payload);
+                assert_eq!(offset, expected_offset);
+            }
+            other @ TransportCommand::Query(_) => {
```
akoshelev, line 447: neat

akoshelev (approved):

PR #437, opened 2023-02-01
Remove bit padding

taikiy:
When I first implemented bitwise protocols, I used `F::Integer::BITS` to determine a bit length of a value into bits for the bit decomposition. Instead, we can computer the exact number of bits needed to fit a value by counting the number of leading zeros of the targeted field's prime number.

### DIFF
--- a/src/protocol/boolean/bit_decomposition.rs
+++ b/src/protocol/boolean/bit_decomposition.rs
@@ -2,7 +2,7 @@ use super::bitwise_less_than_prime::BitwiseLessThanPrime;
 use super::dumb_bitwise_sum::bitwise_sum;
 use super::random_bits_generator::RandomBitsGenerator;
 use crate::error::Error;
-use crate::ff::{Field, Int};
+use crate::ff::Field;
 use crate::protocol::boolean::local_secret_shared_bits;
 use crate::protocol::context::Context;
 use crate::protocol::RecordId;
@@ -61,15 +61,19 @@ impl BitDecomposition {
         .await?;
 
         // Step 7. a bitwise scalar value `f_B = bits(2^l - p)`
-        let l = F::Integer::BITS;
+        let l = u128::BITS - F::PRIME.into().leading_zeros();
         let x = 2_u128.pow(l) - F::PRIME.into();
         let f_b = (0..l).map(|i| F::from(x >> i & 1));
 
         // Step 8, 9. [g_i] = [q] * f_i
-        let g_b = f_b
+        let mut g_b = f_b
             .into_iter()
             .map(|f_bit| q_p.clone() * f_bit)
             .collect::<Vec<_>>();
+
+        // Since bitwise-sum will return +1 bit from the carry, d_b.len() is 1 bit longer
+        // than g_b. Append ZERO to make them the same size.
+        g_b.push(S::ZERO);
 
         // Step 10. [h]_B = [d + g]_B, where [h]_B = ([h]_0,...[h]_(l+1))
         //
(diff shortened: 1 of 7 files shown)


src/protocol/boolean/bit_decomposition.rs
```diff
@@ -61,16 +61,20 @@ impl BitDecomposition {
...
         // Step 8, 9. [g_i] = [q] * f_i
-        let g_b = f_b
+        let mut g_b = f_b
             .into_iter()
             .map(|f_bit| q_p.clone() * f_bit)
             .collect::<Vec<_>>();
 
+        // Since bitwise-sum will return +1 bit from the carry, d_b.len() is 1 bit longer
+        // than g_b. Append ZERO to make them the same size.
+        g_b.push(S::ZERO);
```
martinthomson, line 76: Rather than this, you can use `.chain(std::iter::once())` before collecting.  This will result in the allocation for the vector being correct.

src/protocol/boolean/dumb_bitwise_sum.rs
```diff
@@ -59,9 +58,10 @@ where
     C: Context<F, Share = S>,
     S: ArithmeticSecretSharing<F>,
 {
-    let (a, b) = align_bit_lengths(a, b); // TODO: remove
+    debug_assert_eq!(a.len(), b.len());
```
martinthomson, line 61: Can we make this a compile-time condition?

Like define `a: &[S; B]` with `<const B: usize>` in the signature.  Not sure if that helps any, or if it is even possible.

taikiy: I think this is similar to what's discussed [here](https://github.com/private-attribution/ipa/pull/425#discussion_r1092126039). Would be nice to have a compile-time check on this, but bitwise_sum could be a bit troublesome as it outputs +1 bit from the carry.

martinthomson: @benjaminsavage mentioned that maybe we can exploit the fact that we know certain values are zero here.  Does that apply here also?  Would be better to not feed the overflow bit in? We can add it from the potential carry more easily if the function knows that it is zero.

taikiy: It's possible we could just ignore the MSB from bitwise sum for our use cases. In fact, I tried that instead of adding a bit like in the [above comment](https://github.com/private-attribution/ipa/pull/437#discussion_r1093799569), but tests failed so I didn't go further on that for this PR. I'm guessing that tests failed because of some assumptions we made but not necessary.

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -61,14 +61,17 @@ impl BitDecomposition {
...
         let f_b = (0..l).map(|i| F::from(x >> i & 1));
 
         // Step 8, 9. [g_i] = [q] * f_i
         let g_b = f_b
             .into_iter()
             .map(|f_bit| q_p.clone() * f_bit)
+            // Since bitwise-sum will return +1 bit from the carry, d_b.len() is 1 bit longer
+            // than g_b. Append ZERO to make them the same size.
+            .chain(std::iter::once(S::ZERO))
             .collect::<Vec<_>>();
```
benjaminsavage, line 75: This code is all removed in #443.

taikiy: At least for this PR, I need this line or tests will break. Either one of us needs to do merging :)

src/protocol/boolean/dumb_bitwise_sum.rs
benjaminsavage: I wouldn't worry about it - I removed this entire file in #443 =)

src/protocol/boolean/mod.rs
```diff
@@ -34,7 +34,7 @@ where
     C: Context<F, Share = S>,
```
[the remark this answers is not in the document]
taikiy: You are right. In my next PR, I'm removing this from `credit_capping`, too. R.I.P.

PR #439, opened 2023-02-02
Consolidate auditing and CI tasks

martinthomson:
There's a lot of overlap here, so it makes sense to consolidate like this.

I also added caching for the audit task, which - if it runs every day - might save some time.  It doesn't look like it will save much based on the last run.


.github/workflows/audit.yml
```diff
@@ -19,6 +23,7 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/audit-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: cargo audit
+        run: |
+          cargo audit
```
akoshelev, line 29: this seems to be missing cargo install?

martinthomson: See https://github.com/actions/runner-images/blob/ubuntu22/20230129.2/images/linux/Ubuntu2204-Readme.md#rust-tools

PR #440, opened 2023-02-02
Upgrading from_scalar

richajaindce:


martinthomson (commented):
This all looks pretty good, but you have picked up more than just the change from the title.

Why not drop the `share_of_one()` function entirely?  We can construct that from a single value.  It's maybe a tiny bit more peformant in the case where a == 1, but we can eat that cost, I think.

Also, maybe "scalar" isn't the best name if we're going to expose this on XOR shares.  Perhaps "known value" (`share_known_value`) is what this is.

richajaindce:
Yes it is actually a stacked diff on #438 (which should be quite straightforward) :)

.github/workflows/audit-check.yml
```diff
@@ -20,8 +19,7 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions-rs/audit-check@v1
         with:
-          components: cargo-audit
-      - name: cargo audit
-        run: cargo audit
+          token: ${{ secrets.GITHUB_TOKEN }}
+
```
richajaindce, line 25: I dont know what's happening in this file, I copy pasted the contents from the main branch but it still shows as a diff :O

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -86,6 +86,21 @@ impl<V: ArithmeticShare> AdditiveShare<V> {
...
+        helper_role: Role,
+        value: V,
+        r_share: &SemiHonestAdditiveShare<V>,
+    ) -> Self {
+        let x = SemiHonestAdditiveShare::share_known_value(helper_role, value);
+        let rx = match helper_role {
+            Role::H1 => SemiHonestAdditiveShare::new(r_share.left() * value, V::ZERO),
+            Role::H2 => SemiHonestAdditiveShare::share_known_value(helper_role, value),
+            Role::H3 => SemiHonestAdditiveShare::new(V::ZERO, r_share.right() * value),
+        };
```
benjaminsavage, line 101: I'm 99% sure this is incorrect.

This is not actually a secret-sharing of r * x. Let's consider the case of x = 1. This will produce a 3-way sharing of the following values:

`H1's r_share.left(), 0, 0`

That is NOT a 3-way secret-sharing of r * 1. That's just one of the 3 random components of r. 

I think what you want is just:

```
let rx = r_share * value;
```

benjaminsavage (changes requested):
This is logically incorrect.

src/secret_sharing/replicated/malicious/additive_share.rs
richajaindce: r_share is a Replicated<F> so, it would be of form (left, right)
for x = 1, shouldnt the shares be 
H1: ((1, 0), (left * 1, 0))
H2: ((0, 0), (0, 0))
H3: ((0, 1), (0, right * 1))

Here H1's left would be same as H3's right?

richajaindce: You are right, fixing it

benjaminsavage (approved):

PR #442, opened 2023-02-02
Optimize bit-decomposition by making a specialized protocol to add a …

benjaminsavage:
…bitwise secret shared number to a constant


src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -0,0 +1,192 @@
...
+                .narrow(&BitOpStep::from(bit_index))
+                .multiply(record_id, &S::ZERO, &S::ZERO) // this is stupid
+                .await?;
+
+            S::ZERO
+        } else {
+            ctx.narrow(&BitOpStep::from(bit_index))
+                .multiply(record_id, &last_carry, bit)
+                .await?
+        };
```
taikiy, line 79:
```suggestion
        let mult_result = if last_carry_known_to_be_zero {
            async move { Ok::<_, Error>(S::ZERO) }.left_future()
        } else {
            ctx.narrow(&BitOpStep::from(bit_index))
                .multiply(record_id, &last_carry, bit)
                .right_future()
        }
        .await?;
```

martinthomson: Unfortunately, this stupid multiply is necessary until we work something else out.  We need to ensure that there is a contiguous series of records for each step and with the random revealed value, this optimization leaves gaps at the low bits any time there is a contiguous series of zero bits.

Ultimately, we'll have to build something to track how many preceding records had these gaps so that we can adjust the record ID for these multiplies down (that is, for bit B, record R, how many in records between 0 and R-1 had B-1 or more zero-valued bits in the revealed value, then subtract that from R to get the functional record ID).

taikiy: Oh I see. So this fails in other tests like oneshot/ipa? I ran the unit tests in this file and it checked out.

richajaindce: Yes IPA test cries due to missing processing of records :(

src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        let next_bit_a_one = (b >> bit_index) & 1 == 1;
+        let next_carry = if next_bit_a_one {
+            last_carry_known_to_be_zero = false;
+            -mult_result + &last_carry + bit
+        } else {
+            mult_result
+        };
+
+        let result_bit = if next_bit_a_one {
```
richajaindce, line 89: Given the math and restructure involved, it would be good to add a comment about what is happening at this specific step

richajaindce (approved):

PR #443, opened 2023-02-02
Bitwise add constant (maybe!)

benjaminsavage:
Our Bit-Decomposition protocol was the *only* place we ever invoked the Bitwise Addition protocol. Since #442 removed one reference, I started looking at the other.

It seems that the logic is this:

1. compute a secret-shared value that is either 1 or 0. Let's call that value `maybe`
2. Compute the number 5. (It's just 5 in our case, because 2^32 - PRIME = 5)
3. Bitwise add some bitwise-secret-shared random number to `maybe * 5`

...so I made a specialized protocol to _maybe_ add a constant.

The win is in cases (like this one) where you're adding a constant that's *mostly* zeroes in binary. In those cases, the carry is very simple, it's just one multiplication, the unknown-bit * the last carry. In those cases where the constant has a 1 bit, you have to do more work, but there are only 2 of those, and one is the very first bit, when there *is* no other carry! So it's a special case as well.

This saves 18,900 field values of communication when invoking the "oneshot_ipa" script (which calls IPA on 100 rows).


src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -0,0 +1,448 @@
...
+        };
+        output.push(result_bit);
+
+        last_carry = next_carry;
+    }
+    output.push(last_carry);
+    Ok(output)
+}
+
+pub async fn bitwise_add_constant_maybe<F, C, S>(
```
richajaindce, line 102: Lets add the comments about what function does here also from the PR summary

src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -102,16 +102,104 @@ where
...
+// The win is in cases (like in bit decomposition) where you're adding a constant that's mostly zeroes in binary.
+// In those cases, the carry is very simple, it's just one multiplication, the unknown-bit * the last carry.
+// In those cases where the constant has a 1 bit, you have to do more work,
+// but there are only 2 of those, and one is the very first bit, when there is no other carry! So it's a special case as well.
+pub async fn bitwise_add_constant_maybe<F, C, S>(
+    ctx: C,
+    record_id: RecordId,
+    a: &[S],
+    b: u128,
+    maybe: &S,
```
martinthomson, line 116: You need to more clearly document that `maybe` has to be either 1 or 0 or this function fails spectacularly.

src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -102,16 +102,104 @@ where
...
+    if (b & 1) == 1 {
+        let next_carry = ctx
+            .narrow(&BitOpStep::from(0))
+            .multiply(record_id, &a[0], maybe)
+            .await?;
+        output.push(-next_carry.clone() * F::from(2) + &a[0] + maybe);
+        last_carry = next_carry;
+    } else {
+        output.push(a[0].clone());
+        // In practice, I don't expect to exercize this path, because bit-decomposition
```
martinthomson, line 135:
```suggestion
        // In practice, I don't expect to exercise this path, because bit-decomposition
```

src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -102,16 +102,104 @@ where
...
+        // involves (maybe) adding `2^l - PRIME`, which in our case is:
+        // `2^32 - 4_294_967_291 = 5`
+        // ...so the least significant bit will be a `1`.
+        // In fact, for ANY prime number, the least significant bit will ALWAYS be `1`
+        // Since all primes are odd (apart from 2).
```
martinthomson, line 140: You could probably `assert_eq!(b & 1, 1)` instead.  That would be faster probably.

src/protocol/boolean/dumb_bitwise_add_constant.rs
```diff
@@ -102,16 +102,104 @@ where
...
+    a: &[S],
+    b: u128,
+    maybe: &S,
+) -> Result<Vec<S>, Error>
+where
+    F: Field,
+    C: Context<F, Share = S>,
+    S: ArithmeticSecretSharing<F>,
+{
+    let mut output = Vec::with_capacity(a.len() + 1);
```
martinthomson, line 123: You want to assert that the output length is < 128.

src/protocol/boolean/dumb_bitwise_add_constant.rs
benjaminsavage: OK. I'll have to update the tests then =)

PR #444, opened 2023-02-02
Bringing in F,C,T

richajaindce:
Most of the code has been made generic. We still need to upgrade code after sort in wip_ipa_malicious.
aggregate_credit is intentionally left semihonest since it has call to sort twice and for now, we can validate context before downgrading and sending it to the function


src/protocol/attribution/accumulate_credit.rs
```diff
@@ -124,7 +128,7 @@ pub async fn accumulate_credit<F: Field>(
             .into_iter()
             .enumerate()
             .for_each(|(i, (credit, stop_bit))| {
-                credits[i] = &credits[i] + &credit;
+                credits[i] = credits[i].clone() + &credit;
```
benjaminsavage, line 131: I don't understand why this changed.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -5,38 +5,46 @@ use super::{
...
 /// propagates errors from multiplications
-pub async fn aggregate_credit<F: Field, BK: BitArray>(
+pub async fn aggregate_credit<F, BK>(
     ctx: SemiHonestContext<'_, F>,
-    capped_credits: &[MCAggregateCreditInputRow<F>],
+    capped_credits: &[MCAggregateCreditInputRow<F, Replicated<F>>],
     max_breakdown_key: u128,
     num_multi_bits: u32,
-) -> Result<Vec<MCAggregateCreditOutputRow<F>>, Error> {
+    _is_malicious: bool,
```
benjaminsavage, line 31: If this is unused - why is it here? Can we remove this?

src/protocol/attribution/accumulate_credit.rs
richajaindce: Compiler cries because moving out of the vector issue.

src/protocol/attribution/aggregate_credit.rs
richajaindce: Yes, it is removed in next PR. Let me remove from here itself

src/protocol/attribution/accumulate_credit.rs
benjaminsavage: But why did it work before?

src/protocol/attribution/aggregate_credit.rs
benjaminsavage: It's fine - I'll assume it's used in the next PR. No need to remove.

benjaminsavage (approved):
We're very close to having *everything* generic now!

PR #445, opened 2023-02-02
Bitwise less than (greater than) constant

taikiy:
# Overview

This PR implements "2.1 Comparison with Bitwise Shared Input – LTBits Protocol", bitwise comparison with a constant value.

The logic is pretty simple. Take XOR of the two values, do prefix-or to find the first differing bit, and the value with that differing bit is larger than the other one.

Since we are computing an XOR of a constant and a bitwise shared value, we can perform it locally. The paper uses \$F_2\$ to execute the logics, but we do it over a generic `F`. To output a share of a one or zero, some arithmetic operations' order matters, and I marked them in the comments.

# Performance

* Command:
cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches"

* Parameters:
BATCHSIZE = 8192
NUM_MULTI_BITS = 3
MatchKey bits = 40
BreakdownKey bits = 8

* Time
main: 124.562s
new: 116.146s
-6.8%

* Steps
main: 25,068,438
new: 22,807,446
-9.0%

* bitwise_lt protocol steps
dumb_bitwise_lt: 189
bitwise_lt_const: 75
-60%

* credit_capping steps
main: 24,741
new: 19,497
-21.2%


src/protocol/boolean/bitwise_gt_constant.rs
```diff
@@ -0,0 +1,262 @@
...
+    ctx.narrow(&Step::DotProduct)
+        .sum_of_products(record_id, &first_diff_bit, a)
+        .await
+}
+
+async fn first_differing_bit<F, C, S>(
+    ctx: &C,
+    record_id: RecordId,
+    a: &[S],
+    b: &[F],
```
benjaminsavage, line 44: I think it'll be easier if you just pass in the value. Don't bother bit-decomposing it. Like I did in #442

src/protocol/boolean/bitwise_gt_constant.rs
```diff
@@ -0,0 +1,262 @@
...
+            // This is the same as adding a local share `[b] = b == 1 ? one : zero`.
+            // Now we have `S + S` and computed `S ^ F` locally.
+            let v = a_bit.clone() - &(a_bit.clone() * *b_bit * F::from(2));
+            let b_share = if b_bit.as_u128() == 1 {
+                one.clone()
+            } else {
+                S::ZERO
+            };
+            v + &b_share
+        })
```
benjaminsavage, line 75: I don't understand why this is necessary. Isn't it just:

```
.map(|bit_index, a_bit| {
    if (b >> bit_index) & 1 == 0 {
        a_bit.clone()
    } else {
        one.clone() - a_bit.clone()
    }
}
```

src/protocol/boolean/dumb_bitwise_lt.rs
```diff
@@ -18,8 +18,10 @@ use std::iter::zip;
 /// the little-endian; the least-significant byte at the smallest address (0'th
 /// element).
 ///
+#[allow(dead_code)]
 pub struct BitwiseLessThan {}
```
benjaminsavage, line 22: Can we just delete this file now?

benjaminsavage (changes requested):
Can you please simplify the overly complex code?

src/protocol/boolean/bitwise_gt_constant.rs
taikiy: oh my god, so simple...

src/protocol/boolean/bitwise_gt_constant.rs
```diff
@@ -0,0 +1,249 @@
...
+    ctx.narrow(&Step::DotProduct)
+        .sum_of_products(record_id, &first_diff_bit, a)
+        .await
+}
+
+async fn first_differing_bit<F, C, S>(
+    ctx: &C,
+    record_id: RecordId,
+    a: &[S],
+    b: u32,
```
benjaminsavage, line 42: If you're going to use a u32 for `b`, then you should `assert` that `a.len() <= 32`. Or use a u128, and assert `<= 128`

src/protocol/boolean/bitwise_gt_constant.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    // Subtract neighboring bits to yield all 0's and a single 1 at the index of the first
+    // differing bit. Note that at the index where the transition from 0 to 1 happens,
+    // `prefix_or[i + 1] > prefix_or[i]`. Do not change the order of the subtraction unless
+    // we use Fp2, or the result will be `[p-1]`.
+    let mut first_diff_bit = Vec::with_capacity(prefix_or.len());
+    for i in 0..(prefix_or.len() - 1) {
+        first_diff_bit.push(prefix_or[i].clone() - &prefix_or[i + 1]);
+    }
+    first_diff_bit.push(prefix_or[prefix_or.len() - 1].clone());
```
benjaminsavage, line 94: For a follow-up diff...

...I think we can compute the diff of each bit with its neighbor as a part of the previous loop. In that fashion I think we can avoid some extra allocations. Let's do that in a later PR.

benjaminsavage (approved):
I LOVE this!

PR #450, opened 2023-02-03
Reduce the memory footprint in `bitwise_gt_constant`

taikiy:
* Removed one vector allocation by combining the neighboring bits subtraction with the prefix-or step
* Removed one `reverse()` operation by using `rev()` iteration


benjaminsavage (approved):
Thanks for following up to make this optimization!

PR #451, opened 2023-02-03
Bitwise equal constant

taikiy:
Similar to other bitwise protocols, this is a specialized protocol to check an equality of a bitwise share with a constant value. 

I plan to use this in a modified aggregation protocol where we compare breakdown keys rather than sorting by breakdown keys to sum up credits.


benjaminsavage (approved):
Looks good!

PR #452, opened 2023-02-03
[Malicious IPA] IPA IS NOW MALICIOUS!

richajaindce:
The code is quite copied and there is definitely room for combining and refactoring code. But here is the first cut malicious IPA. Malicious is 2.3x more expensive in communication than semi honest
image


benches/oneshot/ipa.rs
```diff
@@ -3,16 +3,17 @@ use raw_ipa::error::Error;
 use raw_ipa::ff::Fp32BitPrime;
 use raw_ipa::ipa_test_input;
 use raw_ipa::protocol::ipa::ipa;
+// use raw_ipa::protocol::ipa::{ipa_wip_malicious};
```
richajaindce, line 6: Leaving it commented so it is easy to compare two different settings

src/protocol/ipa/mod.rs
```diff
@@ -380,27 +409,32 @@ where
...
         per_user_credit_cap,
     )
     .await?;
 
-    aggregate_credit::<F, BK>(
-        sh_ctx.narrow(&Step::AggregateCredit),
+    //Validate before calling sort with downgraded context
+    // let user_capped_credits = malicious_validator.validate(user_capped_credits).await?;
+
+    malicious_aggregate_credit::<F, BK>(
```
richajaindce, line 427: Returning Replicated<F> as output since there is compute after this and in aggregate_credit, last communication operation was sort

benches/oneshot/ipa.rs
```diff
@@ -3,16 +3,17 @@ use raw_ipa::error::Error;
...
-use raw_ipa::secret_sharing::replicated::semi_honest::AdditiveShare as Replicated;
 use raw_ipa::test_fixture::input::GenericReportTestInput;
 use raw_ipa::test_fixture::{Runner, TestWorld, TestWorldConfig};
 use std::num::NonZeroUsize;
 use std::time::Instant;
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 3)]
 async fn main() -> Result<(), Error> {
-    const BATCHSIZE: usize = 100;
+    const BATCHSIZE: usize = 1000;
```
benjaminsavage, line 16: This will change Ben Case's benchmarks =). We should warn him!

richajaindce: I will change it back to 100?

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -160,6 +163,160 @@ where
...
+
+    let aggregated_credits = malicious_validator.validate(aggregated_credits).await?;
+    //
+    // 4. Sort by `aggregation_bit`
+    //
+    let sorted_output = sort_by_aggregation_bit(
+        sh_ctx.narrow(&Step::SortByAttributionBit),
+        aggregated_credits,
+    )
+    .await?;
```
benjaminsavage, line 304: Is this a malicious sort?

benjaminsavage (approved):
I'm not 100% convinced that we have yet to fully reach malicious security - but this PR takes us a LOT closer to that point. Let's land this, and then do a thorough review of the current code to see if we are missing anything.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -160,6 +163,160 @@ where
...
+/// # Panics
+/// It probably won't
+///
+/// # Errors
+/// propagates errors from multiplications
+#[allow(clippy::too_many_lines)]
+pub async fn malicious_aggregate_credit<F, BK>(
+    m_ctx: MaliciousContext<'_, F>,
+    malicious_validator: MaliciousValidator<'_, F>,
+    sh_ctx: SemiHonestContext<'_, F>,
```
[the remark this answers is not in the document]
richajaindce: Yes, currently we dont give a way of getting semi honest context from malicious context not the validator :( We should simplify this part

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -160,6 +163,160 @@ where
...
+    // 2. Sort by `breakdown_key`. Rows with `aggregation_bit` = 0 must
+    // precede all other rows in the input. (done in the previous step).
+    //
+    let sorted_input = sort_by_breakdown_key(
+        sh_ctx.narrow(&Step::SortByBreakdownKey),
+        capped_credits_with_aggregation_bits,
+        max_breakdown_key,
+        num_multi_bits,
+    )
+    .await?;
```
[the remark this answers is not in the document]
richajaindce: No, #457 should take care of that. Thanks for spotting this

src/protocol/attribution/aggregate_credit.rs
richajaindce: No, https://github.com/private-attribution/ipa/pull/457 should take care of that. Thanks for spotting this

PR #453, opened 2023-02-03
No sort aggregation

taikiy:
This is a stack diff on top of #451. The commit relevant to this PR is 87b144df33dcd208cb85463c608377d6f2ea60eb.

# Overview

This diff eliminates two sort operations from the aggregate_credit protocol. Instead of using sorts, we use `bitwise_equal_constant` to compare bitwise shared breakdown keys (ones that IPA receive from RC) with known breakdown keys (ones we add within IPA for aggregation).

# Performance
* aggregate_credit unit test
main: 5772
new: 4104

* IPA oneshot
main: 10782
new: 9759


taikiy:
I'll send a new PR after #452 is merged.

PR #457, opened 2023-02-03
Upgrading sorts used in IPA 

richajaindce:


src/protocol/ipa/mod.rs
```diff
@@ -522,7 +522,7 @@ pub mod tests {
             (Fp31, MatchKey, BreakdownKey)
         );
 
-        let result: Vec<GenericReportTestInput<Fp31, MatchKey, BreakdownKey>> = world
```
benjaminsavage, line 525: This seems like a regression. Why not call `reconstruct`?

src/protocol/ipa/mod.rs
```diff
@@ -534,20 +534,11 @@ pub mod tests {
...
-                *expected,
-                [
-                    result[i].breakdown_key.as_u128(),
-                    result[i].trigger_value.as_u128()
-                ]
-            );
-        }
+        assert_eq!(EXPECTED.len(), result0.len());
+        assert_eq!(EXPECTED.len(), result1.len());
+        assert_eq!(EXPECTED.len(), result2.len());
```
benjaminsavage, line 541: Wait a minute... we lost the check that the output was correct?

src/protocol/ipa/mod.rs
richajaindce: Unfortunately reconstruct doesn't exist for malicious. Another option is to return a validator, validate results and then check. I was not sure if we want to return validator at the end of ipa

src/protocol/ipa/mod.rs
```diff
@@ -654,7 +645,7 @@ pub mod tests {
 
         let world = TestWorld::new_with(*TestWorldConfig::default().enable_metrics()).await;
 
-        let _: Vec<GenericReportTestInput<Fp32BitPrime, MatchKey, BreakdownKey>> = world
+        let _ = world
```
benjaminsavage, line 648: nice! Can we eliminate all the mess on the equivalent line above: 623?

src/protocol/ipa/mod.rs
richajaindce: Let me bring it back with the validator. I was not sure which way to go .

benjaminsavage: Yeah, I think calling the validator and then checking the results would be better.

PR #459, opened 2023-02-03
Parametrize IPA benchmark with match key bits

akoshelev:
The goal is to allow more flexibility comparing MP-SPDZ and Rust implementation. This change does not make match key size dynamic as it is significantly more work, but rather instructs the IPA protocol (specifically modulus conversion) to ignore higher bits. 

## Testing

Using 20 bits from each match key

```bash
RUST_LOG=debug cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches debug-trace" -- --match-key-bits 20 | grep match_key_shares | cut -d'/' -f7 | cut -d',' -f1 | sed -e "s/elem//g" | sort -nr | uniq | head -n1
19
```

Omit `--match-key-bits` argument and IPA uses all 40 bits

```bash
RUST_LOG=debug cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches debug-trace" | grep match_key_shares | cut -d'/' -f7 | cut -d',' -f1 | sed -e "s/elem//g" | sort -nr | uniq | head -n1
39
```

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -1,3 +1,4 @@
+use clap::Parser;
 use rand::Rng;
 use raw_ipa::error::Error;
 use raw_ipa::ff::Fp32BitPrime;
@@ -5,11 +6,21 @@ use raw_ipa::ipa_test_input;
 use raw_ipa::protocol::ipa::ipa;
 // use raw_ipa::protocol::ipa::{ipa_wip_malicious};
 
-use raw_ipa::protocol::{BreakdownKey, MatchKey};
+use raw_ipa::protocol::{ipa, BreakdownKey, MatchKey};
+use raw_ipa::secret_sharing::SharedValue;
 use raw_ipa::test_fixture::input::GenericReportTestInput;
 use raw_ipa::test_fixture::{Runner, TestWorld, TestWorldConfig};
 use std::num::NonZeroUsize;
 use std::time::Instant;
+
+#[derive(Debug, Parser)]
+pub struct IpaBenchArgs {
+    #[arg(long, help = "bits of match key to use in IPA circuit.")]
+    pub match_key_bits: Option<u8>,
+
+    #[arg(short, long, help = "ignored")]
+    pub bench: bool,
+}
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 3)]
 async fn main() -> Result<(), Error> {
@@ -17,7 +28,7 @@ async fn main() -> Result<(), Error> {
     const MAX_TRIGGER_VALUE: u128 = 5;
     const PER_USER_CAP: u32 = 3;
     const MAX_BREAKDOWN_KEY: u128 = 4;
-    const NUM_MULTI_BITS: u32 = 3;
+    let args = IpaBenchArgs::parse();
 
     let mut config = TestWorldConfig::default();
     config.gateway_config.send_buffer_config.items_in_batch = NonZeroUsize::new(1).unwrap();
@@ -26,9 +37,18 @@ async fn main() -> Result<(), Error> {
     let mut rng = rand::thread_rng();
 
     let max_match_key = u128::try_from(BATCHSIZE / 4).unwrap();
+    let match_key_bits = if let Some(bits) = args.match_key_bits {
+        bits
+    } else {
+        u8::try_from(MatchKey::BITS).unwrap()
+    };
 
-    let mut records: Vec<GenericReportTestInput<Fp32BitPrime, MatchKey, BreakdownKey>> =
-        Vec::with_capacity(BATCHSIZE);
+    assert!(
+        u32::from(match_key_bits) <= MatchKey::BITS,
+        "Requested to use more bits than available in match keys"
+    );
+
+    let mut records = Vec::with_capacity(BATCHSIZE);
 
     for _ in 0..BATCHSIZE {
         records.push(ipa_test_input!(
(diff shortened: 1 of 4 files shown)


src/protocol/ipa/mod.rs
```diff
@@ -155,24 +156,60 @@ impl<F: Field + Sized, T: Arithmetic<F>> Resharable<F> for IPAModulusConvertedIn
     }
 }
 
+/// Describes a request to execute IPA protocol. Consists of inputs for the computation
+/// and query parameters
+pub struct Query<'a, F: Field, MK: BitArray = MatchKey, BK: BitArray = BreakdownKey> {
+    input: &'a [IPAInputRow<F, MK, BK>],
+    /// Maximum contribution allowed per match key/user.
+    per_user_credit_cap: u32,
+    /// ?
```
benjaminsavage, line 165: IPA queries report a total per "breakdown key". These keys are specified by the report collector, and secret-shared along with "source reports". All breakdown keys should be between [0, max_breakdown_key).

src/protocol/ipa/mod.rs
```diff
@@ -155,24 +156,60 @@ impl<F: Field + Sized, T: Arithmetic<F>> Resharable<F> for IPAModulusConvertedIn
...
+    max_breakdown_key: u128,
+    /// That many bits will be sorted at the same time. Setting it to 1 would mean IPA sorts
+    /// 1 bit at a time.
```
benjaminsavage, line 168: Experimentally, we have determined that 3 is the optimal number to reduce communication overhead.

src/protocol/modulus_conversion/convert_shares.rs
```diff
@@ -136,6 +136,7 @@ where
 pub async fn convert_all_bits<F, C, S>(
     ctx: &C,
     locally_converted_bits: &[Vec<BitConversionTriple<S>>],
+    // TODO: u8 should be enough to represent bits
```
benjaminsavage, line 139: I think it was @martinthomson who said that rust just really wants you to use u32 for bitwise stuff.

benjaminsavage (approved):
I like this change. Passing in a "query" object is a lot nicer and fewer lines than like, 6 parameters or whatever we are up to.

benches/oneshot/ipa.rs
```diff
@@ -1,23 +1,34 @@
...
+use raw_ipa::secret_sharing::SharedValue;
 use raw_ipa::test_fixture::input::GenericReportTestInput;
 use raw_ipa::test_fixture::{Runner, TestWorld, TestWorldConfig};
 use std::num::NonZeroUsize;
 use std::time::Instant;
 
+#[derive(Debug, Parser)]
+pub struct IpaBenchArgs {
+    #[arg(long, help = "bits of match key to use in IPA circuit.")]
+    pub match_key_bits: Option<u8>,
```
martinthomson, line 19: There are a lot of other tuning parameters in this file that might benefit from similar treatment.  See BATCHSIZE in particular.  Also, your Query struct is full of goodies that might benefit from slapping clap annotations on it.

You can (and maybe should) use the collapsing mode (`#[flatten]`, from memory) and just include the IPA options as a sub-element of this.  More comments below.

benches/oneshot/ipa.rs
```diff
@@ -45,21 +65,17 @@ async fn main() -> Result<(), Error> {
...
-            ipa::<Fp32BitPrime, MatchKey, BreakdownKey>(
-                ctx,
-                &input_rows,
-                PER_USER_CAP,
-                MAX_BREAKDOWN_KEY,
-                NUM_MULTI_BITS,
-            )
-            .await
-            .unwrap()
+            ipa(ctx, query).await.unwrap()
```
martinthomson, line 71: I think that we should keep two inputs here.  One for the query configuration stuff (how many bits from the match key to process, etc...) and one for all of the inputs.

If you think about how you might arrange this in a real deployment, the configuration will be shared and likely provided to the second and third helpers by the first helper, after receiving it from the record collector.  The inputs will be provided separately.  On that basis, it makes sense to keep the two separate.

src/protocol/ipa/mod.rs
martinthomson: I think that num_breakdown_keys would be a better name for that then.

src/protocol/ipa/mod.rs
martinthomson: It's fine to have this as a tuning parameter, but please set a default if you annotate this.

src/protocol/ipa/mod.rs
```diff
@@ -155,24 +156,60 @@ impl<F: Field + Sized, T: Arithmetic<F>> Resharable<F> for IPAModulusConvertedIn
     }
 }
 
+/// Describes a request to execute IPA protocol. Consists of inputs for the computation
+/// and query parameters
+pub struct Query<'a, F: Field, MK: BitArray = MatchKey, BK: BitArray = BreakdownKey> {
+    input: &'a [IPAInputRow<F, MK, BK>],
```
martinthomson, line 162: I think that you should remove this (keep it separate) and then name the struct IpaConfig or something.

src/protocol/modulus_conversion/convert_shares.rs
martinthomson: [`type_of(u64::BITS) == u32`](https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.BITS)

Seems overkill to me, but I'd prefer to use compatible types than to have to continually think about type conversion.

martinthomson (changes requested):
I think that this is a good start, but I want moar!

akoshelev:
covered by #569

PR #461, opened 2023-02-04
Memoize b

benjaminsavage:
This takes us down BELOW the communication overhead of MP-SPDZ for attribution! Just 251 Mb of comms for 1 million rows (vs MP-SPDZ which uses 487). This is a huge drop compared to what we had before: 1354 Mb!!!

It was a really simple fix. We were just multiplying the same two numbers together over and over again every loop. I just memoized them so that we only do that mult once.


benjaminsavage (commented):
Merging this to improve our communication overhead as much as possible prior to the in-person conference next week.

PR #464, opened 2023-02-06
End to End IPA over network

thurstonsand:
This massive PR covers all the work needed to bootstrap the HTTP server so that we can run IPA end-to-end. As Alex and I worked on this in a sprint near the end of the year, we hadn't merged any of it to main. It's now ready to merge to main


src/ff/prime_field.rs
```diff
@@ -171,6 +160,9 @@ macro_rules! field_impl {
                 }
             }
         }
+
+        // Make sure FieldType has a member for this field implementation.
+        const _FIELD_TYPE_VALUE: FieldType = crate::ff::FieldType::$field;
```
thurstonsand, line 165:
```suggestion
        const _FIELD_TYPE_VALUE: FieldType = FieldType::$field;
```
?

I like this tho. catch the issue at compile time

src/ff/field.rs
```diff
@@ -62,6 +53,7 @@ impl<F: Field> Serializable for F {
 pub trait BinaryField: Field + BooleanOps {}
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "enable-serde", derive(serde::Serialize, serde::Deserialize))]
```
thurstonsand, line 56: my only concern (and maybe this is whatever) is that you've basically got the str repr of `FieldType` defined in 2 places. one in derived serde, and another in `AsRef`. so if you ever change one, you have to remember to change the other. and it would be hard to remember because it's not obvious that they're linked.

of course we do have tests that would check for this, but still.

src/net/server/handlers/query/step.rs
```diff
@@ -0,0 +1,262 @@
...
+    helpers::{transport::TransportCommand, CommandEnvelope, CommandOrigin},
+    net::{http_serde, server::Error},
+    protocol::QueryId,
+    sync::{Arc, Mutex},
+};
+use axum::{routing::post, Extension, Router};
+use std::collections::HashMap;
+use tokio::sync::mpsc;
+
+#[allow(clippy::type_complexity)] // it's a hashmap
```
martinthomson, line 11: Would a type alias make this warning go away in a cleaner way?

src/net/transport.rs
```diff
@@ -0,0 +1,331 @@
...
+        })
+    }
+
+    /// Binds self to port described in `peers_conf`.
+    /// # Panics
+    /// if self id not found in `peers_conf`
+    pub async fn bind(&self) -> (SocketAddr, JoinHandle<()>) {
+        let this_conf = &self.peers_conf[self.id];
+        let port = this_conf.origin.port().unwrap();
+        let target = BindTarget::Http(format!("127.0.0.1:{}", port.as_str()).parse().unwrap());
```
martinthomson, line 54:
```suggestion
        let target = BindTarget::Http(format!("0.0.0.0:{}", port.as_str()).parse().unwrap());
```

We should avoid binding to loopback.

src/protocol/attribution/input.rs
```diff
@@ -34,7 +35,24 @@ pub struct MCAccumulateCreditInputRow<F: Field, T: Arithmetic<F>> {
...
+        helper_bit: T,
+        breakdown_key: Vec<T>,
+        trigger_value: T,
+    ) -> Self {
+        Self {
+            is_trigger_report,
+            helper_bit,
+            breakdown_key,
+            trigger_value,
+            _marker: PhantomData::default(),
```
martinthomson, line 53:
```suggestion
            _marker: PhantomData,
```

I learned a trick.

src/protocol/ipa/mod.rs
```diff
@@ -96,19 +93,149 @@ impl AsRef<str> for IPAInputRowResharableStep {
...
+        <<Replicated<F> as Serializable>::Size as Add<
+            <<XorReplicated<BK> as Serializable>::Size as Add<
+                <Replicated<F> as Serializable>::Size,
+            >>::Output,
+        >>::Output,
+    >>::Output;
+
+    fn serialize(self, buf: &mut GenericArray<u8, Self::Size>) {
+        self.mk_shares.serialize(GenericArray::from_mut_slice(
+            &mut buf[..<XorReplicated<MK> as Serializable>::Size::USIZE],
```
martinthomson, line 142:
```rust
let mk_sz = <XorReplicated<MK> as Serializable>::Size::USIZE;
```

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -182,6 +183,41 @@ impl<V: ArithmeticShare> Mul<V> for AdditiveShare<V> {
     }
 }
 
+impl<V: ArithmeticShare> Serializable for AdditiveShare<V>
```
martinthomson, line 186: todo serde macro for these collections so we can hide the crazy size calculations

src/test_fixture/transport/mod.rs
```diff
@@ -1,13 +1,21 @@
-pub mod network;
+mod network;
```
martinthomson, line 1: This file is empty!

martinthomson (approved):

src/test_fixture/transport/mod.rs
akoshelev: which file? network.rs has something it seems

src/ff/field.rs
akoshelev: yea we probably don't need `AsRef` anywhere except in tests so I took the liberty to get rid of it

akoshelev (approved):

PR #466, opened 2023-02-08
Merge ArithmeticShare and BooleanShare

taikiy:
Goal: Make `XorShare` re-sharable, so that we can shuffle xor-shared match keys.

We need to support addition/subtraction for `XorShare` in order for it to be re-sharable. That implies the shared values in the XorShare, `BitArray`, needs to implement `ArithmeticOps`. By doing so, `Field` and `BitArray` are almost the same except that one is a single value, and the other is a sequence of values.

This PR implements ArithmeticOps for former BitArray (now called `SharedValueArray`), and BooleanOps for Field. `SharedValueArray` implementation is unchanged (specific to bitvec's BitArray), but in the future, we probably want to change it in the future so that it can hold an array of `Field` values.


src/bits/mod.rs
```diff
@@ -1,17 +1,14 @@
...
-
-pub use bit_array::{BitArray40, BitArray8};
+mod shared_value_array;
+pub use shared_value_array::{BitArray40, BitArray8};
 
 /// Trait for data types storing arbitrary number of bits.
 // TODO: Implement `Message`
-pub trait BitArray:
-    BooleanShare + TryFrom<u128> + Into<u128> + Index<usize, Output = bool> + Index<u32, Output = bool>
+pub trait SharedValueArray:
```
martinthomson, line 10: This is maybe a bit misleading given the implementation of these.  It might be better to think of this as `Fp2Array` instead.

src/ff/prime_field.rs
```diff
@@ -85,6 +85,52 @@ macro_rules! field_impl {
...
+            }
+        }
+
+        impl std::ops::BitAndAssign for $field {
+            fn bitand_assign(&mut self, rhs: Self) {
+                *self = *self & rhs;
+            }
+        }
+
+        impl std::ops::BitOr for $field {
```
martinthomson, line 101: Is this right?

Consider `Fp31`.  What do you get if you invoke `Fp31(30) | Fp31(1)`?  The result based on the code below is `Fp31(31)`, which isn't valid.

You might think that implementing this in terms of `Fp31::from()` instead might fix the problem.  That would ensure that the answer is at least in the field (`Fp31(0)`), but I don't know if that is logically consistent with the definition of the type.

I don't see a way to make bit-level operations work for fields.  But perhaps we don't need to.

That is, while we might define arithmetic operations that operate on arrays of `Fp2`, there is no need to define bitwise operations on generic fields.

src/bits/mod.rs
```diff
@@ -1,17 +1,14 @@
...
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Index, Not};
 
-mod bit_array;
-
-pub use bit_array::{BitArray40, BitArray8};
+mod shared_value_array;
+pub use shared_value_array::{BitArray40, BitArray8};
 
 /// Trait for data types storing arbitrary number of bits.
 // TODO: Implement `Message`
```
martinthomson, line 9: Can we not implement serializable generically?

src/secret_sharing/mod.rs
```diff
@@ -11,18 +11,20 @@ use crate::ff::ArithmeticOps;
...
-    Clone + Copy + PartialEq + Debug + Send + Sync + Sized + Serializable + 'static
+    Clone
+    + Copy
+    + PartialEq
+    + Debug
+    + Send
+    + Sync
+    + Sized
+    + ArithmeticOps
+    + BooleanOps
```
martinthomson, line 22: I think that you might need to add one extra layer to this, based on the above comments on fields.

That is, we include `ArithmeticOps` for all, but we need an additional constraint for `Fp2`.

If we ever move to a ring in $F_{2^n}$, we'll be in a position to use the binary operations, but we can't do that in a generic field.

martinthomson (changes requested):

src/bits/mod.rs
taikiy: @akoshelev can we? I can make the change together in this diff if we could.

src/bits/mod.rs
taikiy: Renamed to `Fp2Array`.

src/secret_sharing/mod.rs
taikiy: If we ever actually need Fp2 field values to form `Fp2Array`, I'll add an extra layer to for boolean operations. For now, I added `BooleanOps` bound directly to `Fp2Array`.

Here's the new diagram.

image

martinthomson: Works for me, thanks.

src/bits/mod.rs
martinthomson: I would do this in a later PR, because it is distinct from this.

src/secret_sharing/replicated/semi_honest/additive_share.rs
```diff
@@ -27,6 +31,9 @@ impl<V: SharedValue> Default for AdditiveShare<V> {
 }
 
 impl<V: SharedValue> AdditiveShare<V> {
+    /// Replicated secret share where both left and right values are `F::ZERO`
+    pub const ZERO: Self = Self(V::ZERO, V::ZERO);
```
martinthomson, line 35: Do we have anywhere that wants to use this?

martinthomson (approved):

akoshelev (approved):
lgtm, this PR clashed with #473 and #458, so I had to resolve conflicts before merging

PR #467, opened 2023-02-08
Remove input_len argument from shuffle_and_reveal_permutation

akoshelev:
Our code asserts that input_len == input_permutation.len() in apply/applyinv so there is no reason to have this argument as it is implied to be equal to input_permutation.len()


richajaindce (approved):
LGTM!

PR #469, opened 2023-02-10
eliminate `Arc`s in contexts

andyleiserson:
This eliminates the `Arc`s in contexts, which has been raised as a performance concern (#380). However the larger reason I started down this road is to make it possible to use contexts to hold additional state, in particular, channel endpoints.

Unfortunately doing that resulted in a bunch of rote changes and additional lifetime parameters. I don't know if it makes sense to take this in the current form, but sharing it for discussion. I have an idea of how this could be split into two changes if that were preferable.

The major changes here are:
* Add an additional lifetime parameter to various Contexts. The additional lifetime parameter is necessary so that the contexts can be invariant over the `'a` lifetime parameter (which may appear in the stored data), while still allowing references to contexts with lifetimes shorter than `'a`.
* Add `Buf`-suffixed variants of the contexts that own their state (think `Path` vs. `PathBuf`). There is a function `get_ref` that obtains a `Context` (reference) for a `ContextBuf`.
* In order to create contexts within the `semi_honest` and `malicious` test helpers in `TestWorld`, the test closures had to be changed to return `Pin<Box<dyn Future>>`. The pair of bounds `H: Fn(SemiHonestContext<'a, F>, A) -> R + Send + Sync` and `R: Future<Output = O> + Send` (which refer to a lifetime parameter `'a` that has to outlive the test helper) change to `H: for<'c> Fn(MaliciousContext<'c, '_, F>, M) -> Pin<Box<dyn Future<Output = P> + 'c + Send>> + Send + Sync` which allows a shorter lifetime `'c`. What would be nice here is a way to declare `for<'c>` over the scope of both the `H` and `R` bounds describing the opaque future type.

### DIFF
--- a/src/protocol/context/malicious.rs
+++ b/src/protocol/context/malicious.rs
@@ -3,6 +3,7 @@ use std::marker::PhantomData;
 
 use async_trait::async_trait;
 use futures::future::{try_join, try_join_all};
+use once_cell::sync::OnceCell;
 
 use crate::error::Error;
 use crate::ff::Field;
@@ -24,36 +25,59 @@ use crate::secret_sharing::replicated::{
     malicious::AdditiveShare as MaliciousReplicated, semi_honest::AdditiveShare as Replicated,
 };
 use crate::secret_sharing::Arithmetic;
-use crate::sync::Arc;
 
 /// Represents protocol context in malicious setting, i.e. secure against one active adversary
 /// in 3 party MPC ring.
 #[derive(Clone, Debug)]
-pub struct MaliciousContext<'a, F: Field> {
-    /// TODO (alex): Arc is required here because of the `TestWorld` structure. Real world
-    /// may operate with raw references and be more efficient
-    inner: Arc<ContextInner<'a, F>>,
+pub struct MaliciousContext<'c, 'a, F: Field> {
+    inner: &'c ContextInner<'c, 'a, F>,
     step: Step,
     total_records: TotalRecords,
+    // Ensure that the struct is invariant over 'a, because that may be required in the future.
+    _dummy: PhantomData<OnceCell<&'a ()>>,
 }
 
-pub trait SpecialAccessToMaliciousContext<'a, F: Field> {
+/// Represents protocol context in malicious setting, i.e. secure against one active adversary
+/// in 3 party MPC ring.
+#[derive(Debug)]
+pub struct MaliciousContextBuf<'c, 'a, F: Field> {
+    inner: ContextInner<'c, 'a, F>,
+    step: Step,
+    total_records: TotalRecords,
+    // Ensure that the struct is invariant over 'a, because that may be required in the future.
+    _dummy: PhantomData<OnceCell<&'a ()>>,
+}
+
+impl<'c, 'a, F: Field> MaliciousContextBuf<'c, 'a, F> {
+    pub fn get_ref(&'c self) -> MaliciousContext<'c, 'a, F>
+    {
+        MaliciousContext {
+            inner: &self.inner,
+            step: self.step.clone(),
+            total_records: self.total_records,
+            _dummy: PhantomData,
+        }
+    }
+}
+
+pub trait SpecialAccessToMaliciousContext<'c, 'a, F: Field> {
     fn accumulate_macs(self, record_id: RecordId, x: &MaliciousReplicated<F>);
-    fn semi_honest_context(self) -> SemiHonestContext<'a, F>;
+    fn semi_honest_context(self) -> SemiHonestContext<'c, 'a, F>;
 }
 
-impl<'a, F: Field> MaliciousContext<'a, F> {
+impl<'c, 'a, F: Field> MaliciousContext<'c, 'a, F> {
     pub(super) fn new<S: Substep + ?Sized>(
-        source: &SemiHonestContext<'a, F>,
+        source: &SemiHonestContext<'c, 'a, F>,
         malicious_step: &S,
-        upgrade_ctx: SemiHonestContext<'a, F>,
+        upgrade_ctx: SemiHonestContext<'c, 'a, F>,
         acc: MaliciousValidatorAccumulator<F>,
         r_share: Replicated<F>,
-    ) -> Self {
-        Self {
+    ) -> MaliciousContextBuf<'c, 'a, F> {
+        MaliciousContextBuf {
             inner: ContextInner::new(upgrade_ctx, acc, r_share),
             step: source.step().narrow(malicious_step),
             total_records: TotalRecords::Unspecified,
+            _dummy: PhantomData,
         }
     }
 
@@ -63,7 +87,7 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
     /// by other helpers.  These are caught later.
     pub async fn upgrade<T, M>(&self, input: T) -> Result<M, Error>
     where
-        for<'u> UpgradeContext<'u, F>: UpgradeToMalicious<T, M>,
+        for<'u> UpgradeContext<'c, 'u, F>: UpgradeToMalicious<T, M>,
     {
         self.inner.upgrade(input).await
     }
@@ -92,7 +116,7 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
         input: T,
     ) -> Result<M, Error>
     where
-        for<'u> UpgradeContext<'u, F, RecordId>: UpgradeToMalicious<T, M>,
+        for<'u> UpgradeContext<'c, 'u, F, RecordId>: UpgradeToMalicious<T, M>,
     {
         self.inner
             .upgrade_for_record_with(step, record_id, input)
@@ -100,7 +124,7 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
     }
 }
 
-impl<'a, F: Field> Context<F> for MaliciousContext<'a, F> {
+impl<'c, 'a, F: Field> Context<F> for MaliciousContext<'c, 'a, F> {
     type Share = MaliciousReplicated<F>;
 
     fn role(&self) -> Role {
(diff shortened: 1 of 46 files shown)


src/protocol/context/semi_honest.rs
```diff
@@ -1,44 +1,134 @@
...
-#[derive(Clone, Debug)]
-pub struct SemiHonestContext<'a, F: Field> {
-    /// TODO (alex): Arc is required here because of the `TestWorld` structure. Real world
-    /// may operate with raw references and be more efficient
-    pub(super) inner: Arc<ContextInner<'a>>,
+#[derive(Debug)]
+pub struct ContextStore<'a> {
+    pub(super) static_data: ContextStatic<'a>,
+    // Ensure that the struct is invariant over 'a, because that may be required in the future.
+    pub(super) _dummy: PhantomData<OnceCell<&'a ()>>,
```
martinthomson, line 23: Presumably this also works, if your goal is just to pull in the `'a` lifetime.

```suggestion
    pub(super) _dummy: PhantomData<&'a ()>,
```

src/protocol/context/semi_honest.rs
```diff
@@ -1,44 +1,134 @@
...
-        Self::new_with_total_records(participant, gateway, TotalRecords::Unspecified)
-    }
+#[derive(Debug)]
+pub struct SemiHonestContextBuf<'a, F: Field> {
+    pub(super) store: ContextStore<'a>,
+    pub(super) step: Step,
+    pub(super) total_records: TotalRecords,
+    // Ensure that the struct is invariant over 'a, because that may be required in the future.
+    _dummy: PhantomData<OnceCell<&'a ()>>,
+    _marker: PhantomData<F>,
```
martinthomson, line 52: It might be better to combine these:

```suggestion
    _marker: PhantomData<(&'a (), F)>,
```

src/protocol/context/malicious.rs
```diff
@@ -24,36 +25,59 @@ use crate::secret_sharing::replicated::{
...
     total_records: TotalRecords,
+    // Ensure that the struct is invariant over 'a, because that may be required in the future.
+    _dummy: PhantomData<OnceCell<&'a ()>>,
 }
 
-pub trait SpecialAccessToMaliciousContext<'a, F: Field> {
+/// Represents protocol context in malicious setting, i.e. secure against one active adversary
+/// in 3 party MPC ring.
+#[derive(Debug)]
+pub struct MaliciousContextBuf<'c, 'a, F: Field> {
```
martinthomson, line 43: Would it be better to name this `MaliciousContext` and then rename `MaliciousContext` as `MaliciousContextRef` to reflect that the latter is really just a reference to the active state?

src/protocol/context/malicious.rs
```diff
@@ -3,6 +3,7 @@ use std::marker::PhantomData;
 
 use async_trait::async_trait;
 use futures::future::{try_join, try_join_all};
+use once_cell::sync::OnceCell;
```
martinthomson, line 6: I don't see you using `OnceCell` here, which is good, I think.

src/protocol/context/malicious.rs
```diff
@@ -24,36 +25,59 @@ use crate::secret_sharing::replicated::{
...
 }
 
-pub trait SpecialAccessToMaliciousContext<'a, F: Field> {
+/// Represents protocol context in malicious setting, i.e. secure against one active adversary
+/// in 3 party MPC ring.
+#[derive(Debug)]
+pub struct MaliciousContextBuf<'c, 'a, F: Field> {
+    inner: ContextInner<'c, 'a, F>,
+    step: Step,
+    total_records: TotalRecords,
```
martinthomson, line 46: How much do we depend on these two fields (step/total_records) being available on this `...Buf` struct?  How often is this specific type used as opposed to the reference type?

It seems to me like this type could omit these fields and the `get_ref()` function could populate the values with a default/initial value.

Looking at how you use `get_ref()`, it might be the latter: you might pass a `Substep` implementation to the function.

martinthomson (commented):
The additional lifetime parameter is a little annoying, but not enough to concern me.  I have some questions and feedback on the overall structure, but this looks like it might be worth doing.

(You know, it would be really nice if Rust were to build in some way to express common or shared type constraints.  We're starting to get enough type constraints on things that compressing those is likely to help, both in terms of making code easier to type out and in making it less error-prone to define new functions and types.)

src/protocol/context/semi_honest.rs
andyleiserson: There is a subtle difference with and without `OnceCell`. Because the `OnceCell` introduces interior mutability, the struct becomes invariant in the lifetime `'a` rather than covariant in the lifetime `'a`. In the covariant case, a function that takes `ContextStore<'a>` will also accept `ContextStore<'b>` for `'b: 'a`. In the invariant case, the function will only accept exactly `ContextStore<'a>` and not any other lifetime.

Minimal example, if you want to play with it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=386bd282eb9543dd8b991d7e8c7234fb

andyleiserson: > (You know, it would be really nice if Rust were to build in some way to express common or shared type constraints. We're starting to get enough type constraints on things that compressing those is likely to help, both in terms of making code easier to type out and in making it less error-prone to define new functions and types.)

Agreed. In some situations a trait can be used for this purpose, but that results in some number of `impl MyTraitUsedOnlyAsBoundsAlias for SomeType {}`, which is kind of ugly, especially for library consumers.

Or a macro, but I'm in camp macro-dubious, so I don't particularly like that either.

src/protocol/context/semi_honest.rs
andyleiserson: My thought was that `_dummy` goes away once we add some other member that makes use of interior mutability, whereas `_marker` isn't going anywhere, but I do agree that it otherwise makes sense to be less verbose by combining the two `PhantomData`s.

src/protocol/context/malicious.rs
andyleiserson: I thought about that... I think one reason I didn't was to try and keep the diff smaller, but at that point I was still hoping not to change the lifetime parameter. If we have to touch all the instances of `SemiHonestContext<...>` to add the new lifetime parameter, then it's no more work to change the name. I don't have a strong preference between `Context` / `ContextBuf` and `ContextRef` / `Context`.

src/protocol/context/malicious.rs
andyleiserson: Only in `PhantomData`, for now. I got aggressive about putting placeholders in all the types to make sure that I rooted out anything that was going to have a problem with invariance over the lifetime parameter. In cases where we think we may want to make use of that in the foreseeable future, then keeping the placeholder avoids regressions. But I don't see a specific need for it in the malicious context structs, so I'm fine dropping it. (At the moment, the malicious contexts contain a semi-honest context, so they'll inherit the invariance that way.)

src/protocol/context/semi_honest.rs
martinthomson: I looked at your covariance/invariance example and I'm a bit baffled.  What about `OnceCell` makes it behave this way when the simple wrapper implementation doesn't?  Is it the use of `UnsafeCell`?

(Count me as in the macro-dubious camp also.  They tend to be opaque; I'm looking for something that is maximally transparent.)

src/protocol/context/malicious.rs
andyleiserson: > How much do we depend on these two fields (step/total_records) being available on this ...Buf struct? How often is this specific type used as opposed to the reference type?

The answer to the second question is that the owned type is used rarely. I think I may have tried removing the step/records from the buf type at some point during development, but I can't find the version of the code where I had done that, nor can I remember why I decided not to. Possibly because SpecialAccessToMaliciousContext appeared to need to store step/records in the owned type? But I don't think that's a good reason, SpecialAccessTopMaliciousContext is a special case constructor and I've in fact already changed it to return a borrowed instead of owned context in the latest version of my changes. There were also periods where I thought that I might be able to have ContextBuf deref to Context or to otherwise have the test closures accept ContextBuf as their argument rather than Context, in which case ContextBuf might have more need to hold the full state of a context.

Another variant would be to eliminate the ContextBuf owned type entirely, and just use ContextStore as the owned type (or keep ContextBuf, and get rid of ContextStore).

src/protocol/context/semi_honest.rs
andyleiserson: Yes, it is the `UnsafeCell` that triggers this. https://doc.rust-lang.org/nomicon/subtyping.html has a useful table.

There is a discussion to be had whether we need this ability to mutate the context. If we don't, then (modulo some cleanup) I already have a branch that achieves the goal of "eliminate Arcs in contexts" with a significantly smaller change than this. However, absent some even larger refactoring of how we handle state associated with communication channels, I felt like some mutability through the context is likely to be necessary.

There's no reason the marker has to be a OnceCell -- but it does have to be `Sync`, which some of the other likely candidates (`Cell` or `&mut`) are not.

src/protocol/context/malicious.rs
akoshelev: I may also prefer `ContextRef`/`Context`. `Buf` to me implies that something may be growing or shrinking

src/protocol/sort/generate_permutation.rs
```diff
@@ -117,7 +119,7 @@ pub(super) async fn malicious_shuffle_and_reveal_permutation<F: Field>(
     Ok(RevealedAndRandomPermutations {
         revealed: revealed_permutation,
         randoms_for_shuffle: random_permutations_for_shuffle,
-    })
+    })*/
```
andyleiserson, line 122: I meant to leave a comment here when I opened the PR. The trouble with this code is that `MaliciousValidator::validate` takes the validator by value, which is incompatible with `m_ctx` having an active borrow.

One idea that I had for dealing with this is to have `DowngradeMalicious::downgrade` take a context as an argument. However, I'm not sure I like that as a solution, since many kinds of values can be downgraded without a context.

So I'm still thinking about how best to handle this, and I'm open to suggestions.

PR #470, opened 2023-02-10
[Iterating] Sort now takes iterator as input

richajaindce:


src/protocol/modulus_conversion/mod.rs
```diff
@@ -66,14 +66,14 @@ pub fn split_into_multi_bit_slices<T: Clone>(
 ///     ...
 ///     `[ row[n].bit0, row[n].bit1, ..., row[n].bit31 ]`,
 ///  `]`
-pub fn combine_slices<T: Clone>(
-    input: &[Vec<Vec<T>>],
+pub fn combine_slices<'a, T: Clone + 'a>(
+    input: impl Iterator<Item = &'a Vec<Vec<T>>>,
```
martinthomson, line 70: You should use `impl IntoIterator` rather than `impl Iterator` where possible.

src/protocol/modulus_conversion/mod.rs
```diff
@@ -66,14 +66,14 @@ pub fn split_into_multi_bit_slices<T: Clone>(
 ///     ...
 ///     `[ row[n].bit0, row[n].bit1, ..., row[n].bit31 ]`,
 ///  `]`
-pub fn combine_slices<T: Clone>(
-    input: &[Vec<Vec<T>>],
+pub fn combine_slices<'a, T: Clone + 'a>(
```
martinthomson, line 69: This looks like the much-dreaded `transpose()` we tried very hard to eliminate.

Though this takes and produces an iterator, it is still very expensive because it allocates a vector.

The input vector appears to be a 3d collection, but it appears as though that is only a fraction of `num_bits` item deep on the third dimension.  (The documentation here is unhelpful as both input and output in the comment are 2d.  It would be easier to understand if this explained some of the context into which this function is expected to fit.)

It seems to me like this is a hard problem to solve if you are operating on an entire collection in this way, as the chunks into which we split the bits are spread throughout the input.  To avoid the allocation, you need all bits that need to be adjacent in the output to be adjacent in the input.

For that to work out, we probably need to save something, but I would prefer it if we changed the bit conversion so that we store all bits for each record in a single row, rather than produce a 3d vector.  We then need to index into the correct bits as needed when sorting, but at least we don't need this expensive conversion once the bits are sorted.

martinthomson (commented):
The use of an iterator here only really papers over deeper problems that need a lot more thought.  There is still a huge allocation involved here.

src/protocol/modulus_conversion/mod.rs
richajaindce: The documentation was incorrect, I should have added one [] in each row.

richajaindce: In the next PR i was hoping that we could return iterator from convert_all_bits which would be slices based on what sort needs in that iteration (e.g. iterator on m..n bits for each record). By doing what you suggested, we would need to pre-process modulus conversion on all the records together. We also have an upcoming optimization to interleave modulus conversion with sort which would be easier with the current format: https://github.com/private-attribution/ipa/blob/main/src/protocol/sort/generate_permutation_opt.rs#L172

Let me know your thoughts!

martinthomson: The problem with this is that you are doing a full-sized allocation here, which is maybe OK if this is the single place where that allocation occurs, but that doesn't seem to be where we are.

richajaindce: Agree this PR does not take care of this. It is just dumbly replacing Vec's by iterator. Let me remove combine_slices change from this PR. This issue should be handled properly once XorReplicated is refactored. Sort should then convert XorReplicated shares, use them to sort, combine and return it to the caller. Does that sound ok?

src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -45,20 +45,20 @@ use embed_doc_image::embed_doc_image;
 /// Panics if input doesn't have same number of bits as `num_bits`
 pub async fn generate_permutation_opt<F>(
     ctx: SemiHonestContext<'_, F>,
-    sort_keys: &[Vec<Vec<Replicated<F>>>],
+    mut sort_keys: impl Iterator<Item = &Vec<Vec<Replicated<F>>>>,
```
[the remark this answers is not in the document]
martinthomson: For functions that already have where clauses, it might be better to use those rather than mix `impl` and `where`.

src/protocol/sort/apply_sort/mod.rs
```diff
@@ -122,16 +122,15 @@ mod tests {
...
                         &convert_all_bits_local(ctx.role(), &bk_shares),
                         BreakdownKey::BITS,
-                        NUM_MULTI_BITS,
+                        BreakdownKey::BITS,
                     )
                     .await
                     .unwrap();
-                    let converted_bk_shares =
-                        combine_slices(&converted_bk_shares, BreakdownKey::BITS);
+                    let converted_bk_shares = converted_bk_shares.remove(0).into_iter();
```
martinthomson, line 133:
```suggestion
                    let converted_bk_shares = converted_bk_shares.into_iter().skip(1);
```

martinthomson (commented):
You are using remove(), which I thought was something that can be replaced by skip, but I think that you need to use `next()` instead because you want to keep the iterator for use later.

src/protocol/sort/apply_sort/shuffle.rs
```diff
@@ -229,16 +227,15 @@ mod tests {
...
                             &convert_all_bits_local(ctx.role(), &bk_shares),
                             BreakdownKey::BITS,
-                            NUM_MULTI_BITS,
+                            BreakdownKey::BITS,
                         )
                         .await
                         .unwrap();
-                        let converted_bk_shares =
-                            combine_slices(&converted_bk_shares, BreakdownKey::BITS);
+                        let converted_bk_shares = converted_bk_shares.remove(0).into_iter();
```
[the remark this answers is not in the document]
richajaindce: Used pop instead.. Actually this code was doing unnecessary work of splitting converted_bk_shares, so just ensuring it does the right thing for now. I plan to refactor it when we interleave convert_all_bits with sort. Will add a TODO

src/protocol/sort/apply_sort/mod.rs
richajaindce: Used pop instead

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -157,16 +153,16 @@ mod tests {
...
                         &convert_all_bits_local(ctx.role(), &bk_shares),
                         BreakdownKey::BITS,
-                        NUM_MULTI_BITS,
+                        BreakdownKey::BITS,
                     )
                         .await
                         .unwrap();
                     let converted_bk_shares =
-                        combine_slices(&converted_bk_shares, BreakdownKey::BITS);
+                    converted_bk_shares.pop().unwrap();
```
martinthomson, line 165: I find it fascinating that you were able to change a `.remove(0)` to `.pop()` without any tests failing.  Then I realized that we don't really depend on our sort doing a sort.

Though we rely on the stability aspect to ensure that we don't perturb any pre-sorting (based on timestamp), the order in which we apply permutations doesn't really matter.

Is there any way we can exploit this fact here?  Maybe there isn't any communications savings, but maybe we can save on some of the transient state costs.

martinthomson (approved):

PR #471, opened 2023-02-10
Make oneshot IPA compile again

akoshelev:
After merging https://github.com/private-attribution/ipa/pull/464, oneshot IPA benchmark stopped compiling w/o
default features enabled. This commit fixes it by gating web-app
features behind `web-app` feature flag.

One place where it was particularly problematic was `ByteArrStream`
with a strong dependency on Hyper. I had to replace it with `Bytes`.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -53,3 +53,5 @@ cargo clippy --tests --features shuttle -- -D warnings || error "Clippy concurre
 cargo test || error "Test failures"
 
 cargo test --features shuttle || error "Concurrency test failures"
+
+cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches" || error "Failed to run IPA benchmark"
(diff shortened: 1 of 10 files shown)


andyleiserson (approved):

pre-commit
```diff
@@ -53,3 +53,5 @@ cargo clippy --tests --features shuttle -- -D warnings || error "Clippy concurre
 cargo test || error "Test failures"
 
 cargo test --features shuttle || error "Concurrency test failures"
+
+cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches" || error "Failed to run IPA benchmark"
```
richajaindce, line 57: Later, should we add an argument for batchsize and run this only for 10 rows in pre-commit. We are slowly increasing the pre-commit time more and more

richajaindce (approved):

PR #472, opened 2023-02-13
Save MERGE_* in pre-commit execution

taikiy:
When `git commit` is executed to finalize a merge process, MERGE_* files in .git/ are silently removed by `git stash push` in `pre-commit` script. This results in merge's `git commit` to fail, and one needs to restart the merging process from the beginning.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -26,24 +26,24 @@ if [[ "$1" == "install" ]]; then
 fi
 
 save_merge_files() {
-    # MERGE_[HEAD|MODE|MSG]
+    # Move MERGE_[HEAD|MODE|MSG] files to the root directory, and let `git stash push` saves them.
     for f in "$root"/.git/MERGE_*; do
 	if [[ -e "$f" ]]; then
             t=$(basename $f)
-            cp -f "$f" "$t"
+            mv -f "$f" "$t"
 	fi
     done
 }
 
 restore_merge_files() {
+    # Moves MERGE files restored by `git stash pop` back into .git/ directory.
     for f in MERGE_*; do
 	if [[ -e "$f" ]]; then
-            t=$(basename $f)
-	    if [[ -e "${root}/.git/${t}" ]]; then
-		echo "Failed to restore ${t}. File already exists"
-		exit 1
+	    if [[ -e "${root}/.git/${f}" ]]; then
+		echo "Failed to restore ${f}. File already exists" 1>&2
+	    else
+		mv -f "$f" "${root}/.git/${f}"
 	    fi
-            mv -f "$f" "${root}/.git/${t}"
 	fi
     done
 }
--- a/pre-commit
+++ b/pre-commit
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
     exit
 fi
 
+save_merge_files() {
+    # MERGE_[HEAD|MODE|MSG]
+    for f in "$root"/.git/MERGE_*; do
+	if [[ -e "$f" ]]; then
+            t=$(basename $f)
+            cp -f "$f" "$t"
+	fi
+    done
+}
+
+restore_merge_files() {
+    for f in MERGE_*; do
+	if [[ -e "$f" ]]; then
+            t=$(basename $f)
+	    if [[ -e "${root}/.git/${t}" ]]; then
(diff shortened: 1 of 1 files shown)


pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
...
+
+restore_merge_files() {
+    for f in MERGE_*; do
+	if [[ -e "$f" ]]; then
+            t=$(basename $f)
+	    if [[ -e "${root}/.git/${t}" ]]; then
+		echo "Failed to restore ${t}. File already exists"
+		exit 1
+	    fi
+            mv -f "$f" "${root}/.git/${t}"
```
martinthomson, line 46: Given that this is in the cleanup, maybe just include the warning.

```suggestion
		echo "Failed to restore ${t}. File already exists" 1>&2
            else
		mv -f "$f" "${root}/.git/${t}"
            fi
```

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
     exit
 fi
 
+save_merge_files() {
+    # MERGE_[HEAD|MODE|MSG]
+    for f in "$root"/.git/MERGE_*; do
+	if [[ -e "$f" ]]; then
+            t=$(basename $f)
```
martinthomson, line 32: This is moving the files into the root.  Is the goal to ensure that `git stash` saves these files?  Add a comment to that effect perhaps.

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
...
+            cp -f "$f" "$t"
```
martinthomson, line 33:
```suggestion
            mv -f "$f" "$t"
```

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
...
+	fi
+    done
+}
+
+restore_merge_files() {
+    for f in MERGE_*; do
+	if [[ -e "$f" ]]; then
+            t=$(basename $f)
```
martinthomson, line 41: Shouldn't be necessary.

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
...
+            cp -f "$f" "$t"
+	fi
+    done
+}
+
+restore_merge_files() {
+    for f in MERGE_*; do
+	if [[ -e "$f" ]]; then
```
martinthomson, line 40: You should need this if statement: the for loop guarantees that a file exists here.

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
     exit
 fi
 
+save_merge_files() {
+    # MERGE_[HEAD|MODE|MSG]
+    for f in "$root"/.git/MERGE_*; do
+	if [[ -e "$f" ]]; then
```
martinthomson, line 31: See below.

martinthomson (commented):
A useful contribution, for sure.  I've started committing merges with `-n`, but this is better.

pre-commit
taikiy: I thought so too, but `mv` actually fails if there's no match. "$f" is filed with "MERGE_*" in that case.

`mv: rename /Users/taiki/Source/raw-ipa/.git/MERGE_* to MERGE_*: No such file or directory`

martinthomson: Ah, right.  I'm use to `make`, where `$(wildcard MERGE_*)` would be empty if there is no match.

You could use `$(ls MERGE_*)`, I guess.

pre-commit
```diff
@@ -25,9 +25,33 @@ if [[ "$1" == "install" ]]; then
     exit
 fi
 
+save_merge_files() {
+    # Move MERGE_[HEAD|MODE|MSG] files to the root directory, and let `git stash push` saves them.
```
martinthomson, line 29:
```suggestion
    # Move MERGE_[HEAD|MODE|MSG] files to the root directory, and let `git stash push` save them.
```

martinthomson (approved):
Just one tweak and I'll merge it.

akoshelev (approved):
fwiw, I am using the same script as pre-push hook, leaving pre-commit empty.

PR #473, opened 2023-02-13
Add constructor to CappedCreditsWIthAggregationBit struct

akoshelev:
to avoid exposing `PhantomData` marker


richajaindce (approved):
Agree on asking the client to send phantomdata. Change looks good!

PR #475, opened 2023-02-16
[Refactoring] Remove removes and make _markers private

richajaindce:


src/protocol/attribution/input.rs
```diff
@@ -238,15 +247,15 @@ impl<F: Field, T: Arithmetic<F>> Resharable<F> for MCAccumulateCreditInputRow<F,
...
         .await?;
 
         Ok(MCAccumulateCreditInputRow {
             breakdown_key,
-            is_trigger_report: fields.remove(0),
-            helper_bit: fields.remove(0),
-            trigger_value: fields.remove(0),
+            is_trigger_report: fields.pop().unwrap(),
+            helper_bit: fields.pop().unwrap(),
+            trigger_value: fields.pop().unwrap(),
```
martinthomson, line 258: Rather than collecting these into a Vec, you should probably change to using a tuple.  That will (might?) avoid a heap allocation for the Vec and you can use a destructuring assignment, like:

```rust
        let (breakdown_key, (value, helper_bit, trigger_bit)) = try_join(
            f_breakdown_key,
            try_join3(f_value, f_helper_bit, f_trigger_bit),
        )
        .await?;
```

That also avoids the unnatural reverse order you have here.

martinthomson (commented):
One suggestion.

src/protocol/attribution/input.rs
richajaindce: Thanks for the suggestion!

PR #476, opened 2023-02-16
eliminate use of downgrade auth trait in attribution::input

andyleiserson:
Partly prompted by https://github.com/private-attribution/ipa/pull/458#discussion_r1096418881 and partly because I'm looking at handling of downgrades for `ShuffledPermutationWrapper` in the context of https://github.com/private-attribution/ipa/pull/469#discussion_r1106472691.

With a more complicated version of the map trait that allows for contexts and record IDs, this could possibly be extended to support upgrades, reveals, and reshares, but I stopped myself before I went that far.

My thoughts for ShuffledPermutationWrapper are to add another trait (something like `MaliciousOutput`) that takes a context. `MaliciousOutput` can be implemented for anything that is `Downgrade` and will just not use the context in that case. `MaliciousOutput` will also be implemented for `ShuffledPermutation` to reveal the permutation. A more general version would implement some kind of `Reveal` trait and implement `MaliciousOutput` for anything that is `Reveal`, but again, that seems like unnecessary complexity.

Besides the issue of needing a context to reveal the permutation, I think it's somewhat incorrect to put a reveal operation in the same class with downgrades. Whether the output value needs to be downgraded or revealed isn't implied by the output data type -- it's dictated by the role of the value in the protocol.

### DIFF
--- a/src/helpers/map.rs
+++ b/src/helpers/map.rs
@@ -1,0 +1,37 @@
+/// Trait for transforming values in the manner of `Option::map` or `Iter::map`.
+///
+/// This is currently used to implement [malicious downgrades], and can be used to
+/// generically implement other operations in the future.
+///
+/// The type parameter `M` represents the mapping being performed. It is analogous to the `F`
+/// closure type parameter on `Option::map`.
+///
+/// [malicious downgrades]: crate::secret_sharing::replicated::malicious::UncheckedDowngrade
+pub trait Map<M: Mapping> {
+    type Output;
+
+    fn map(self) -> Self::Output;
+}
+
+pub trait Mapping {}
+
+impl<T, U, M: Mapping> Map<M> for (T, U)
+where
+    T: Map<M>,
+    U: Map<M>,
+{
+    type Output = (<T as Map<M>>::Output, <U as Map<M>>::Output);
+    fn map(self) -> Self::Output {
+        (self.0.map(), self.1.map())
+    }
+}
+
+impl<T, M: Mapping> Map<M> for Vec<T>
+where
+    T: Map<M>,
+{
+    type Output = Vec<<T as Map<M>>::Output>;
+    fn map(self) -> Self::Output {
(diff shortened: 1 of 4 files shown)


src/helpers/map.rs
```diff
@@ -0,0 +1,31 @@
+/// Trait for transforming values in the manner of `Option::map` or `Iter::map`.
+///
+/// This is used to implement malicious downgrades.
+pub trait Map<M: Mapping> {
```
akoshelev, line 4: I wish there was a way to express this without an extra generic type that has no meaning in `Map`. I guess associated type does not work because it is unstable?

akoshelev (approved):

src/helpers/map.rs
andyleiserson: I think of the type parameter `M` as representing the degree of freedom to specify what mapping is being performed -- somewhat like the parameter `F` on [`iter::Map`](https://doc.rust-lang.org/std/iter/struct.Map.html).

On the impls of `Map<M>` for various types, the parameter signifies "this code is generic over the type of mapping being performed".

Associated type defaults are unstable, but I'm not sure what I would do with an associated type for `M`. I see how the type parameter looks like unnecessary baggage in most of the `Map` impls, but it's necessary at the leaf (`impl<F: Field> Map<UncheckedDowngrade> for AdditiveShare<F>`), and I'm not sure how to accomplish the same goal without it. 

This PR doesn't really demonstrate the power of the generic `Map`, but to illustrate, a version of `Map` that understood contexts and record IDs could replace the `Reshare` impls in [#493](https://github.com/private-attribution/ipa/pull/493/files#diff-eeea3aba623b98182cbcf841810cefca926225cbe6c0d766097b9b807831d577R227). Once `Map` is implemented once for `MCCappedCreditsWithAggregationBit`, you would automatically get `Reshare`, `Reveal`, `UpgradeMalicious`, `DowngradeMalicious`, and any other operation that can be applied to complex structs by applying the operation to each member of the struct.

PR #478, opened 2023-02-18
Adding specialized functionality for when the per-user cap is one

benjaminsavage:
One common ads use-case is to just *count* conversions. In this case, the "trigger value" is irrelevant. Every conversion counts the same, just one.

Furthermore, a common use-case is to *count distinct* users. Another way of saying this, is that each user can at most contribute a value of *one* to the output. 

In this special case, we can completely avoid bit-decomposition, and indeed we can avoid large chunks of the IPA protocol entirely. As this is not a rare use-case, it seems worthwhile to specialise this, so that we can run a much faster, more communication efficient algorithm.

Testing out on 1000 rows, "Records Sent" drops from 2,286,120 to 1,652,889 - a 27.7% drop in total communication.

Looking at the breakdown by stage:

- Accumulate drops from 77,796 to 2,997 (now it's just ONE multiplication per row!)
- User Capping drops from 668,790 to 110,601 (no more bit-decomposition! Just a prefix-OR!)

### DIFF
--- a/src/protocol/attribution/accumulate_credit.rs
+++ b/src/protocol/attribution/accumulate_credit.rs
@@ -22,6 +22,18 @@ impl AsRef<str> for Step {
     }
 }
 
+///
+/// When `PER_USER_CAP` is set to one, this function is called
+/// In this case, `trigger_value` is ignored entirely. Instead, each `trigger_report` counts as one.
+/// So in the event that a `source report` is followed by multiple `trigger reports`, only one will count.
+/// As such, this function can be simplified a great deal. All that matters is when a `source report` is
+/// immediately followed by a `trigger report` from the same `match key`. As such, each row only needs to
+/// be compared to the following row.
+/// If there are multiple attributed conversions from the same `match key` they will be removed in the
+/// next stage; `user capping`.
+///
+/// This method implements "last touch" attribution, so only the last `source report` before a `trigger report`
+/// will receive any credit.
 async fn accumulate_credit_cap_one<F, C, T>(
     ctx: C,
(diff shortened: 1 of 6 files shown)


akoshelev:
do you want to add the baseline test for this feature?

src/protocol/attribution/accumulate_credit.rs
```diff
@@ -22,6 +22,46 @@ impl AsRef<str> for Step {
     }
 }
 
+async fn accumulate_credit_cap_one<F, C, T>(
```
akoshelev, line 25: would be nice to have some documentation for this function explaining the key differences

akoshelev (approved):

src/protocol/ipa/mod.rs
```diff
@@ -714,6 +806,115 @@ pub mod tests {
         assert_eq!(MAX_BREAKDOWN_KEY, result.len() as u128);
     }
 
+    struct TestRawDataRecord {
```
martinthomson, line 809: There is a lot of code duplication here.

benjaminsavage: Yes, I am in the process of eliminating the `random_ipa_no_result_check` test, and merging it into this one which actually does test outputs.

benjaminsavage: Here's the PR: https://github.com/private-attribution/ipa/pull/480

...but in the process of merging I discovered we have a logic error... and the test fails for per_user_cap > 1... so this is just step one to get this test up so that we can figure out where the issue is.

PR #480, opened 2023-02-20
Making the random data test actually check outputs

benjaminsavage:
The "random" IPA test did not actually validate that the output was correct.

This PR creates a single "random" IPA test, which uses a seeded random number generator and logs the seed. That way, if it ever fails, we can re-test the seed that led to the failure.

Since the logic differs (significantly) if the PER_USER_CAP is 1 or another number, the same input is tested twice, first with a PER_USER_CAP of 1, then with a PER_USER_CAP of 3.

The expected output is computed in a straightforward way that's easy to verify is correct.

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -760,62 +760,18 @@ pub mod tests {
         }
     }
 
-    #[tokio::test]
-    #[allow(clippy::missing_panics_doc)]
-    #[ignore]
-    pub async fn random_ipa_no_result_check() {
-        const BATCHSIZE: u128 = 20;
-        const PER_USER_CAP: u32 = 10;
-        const MAX_BREAKDOWN_KEY: u128 = 8;
-        const MAX_TRIGGER_VALUE: u128 = 5;
-        const NUM_MULTI_BITS: u32 = 3;
-
-        let max_match_key: u128 = BATCHSIZE / 10;
-
-        let world = TestWorld::new().await;
-        let mut rng = thread_rng();
-
-        let mut records = Vec::new();
-
-        for _ in 0..BATCHSIZE {
-            records.push(ipa_test_input!(
-                {
-                    match_key: rng.gen_range(0..max_match_key),
-                    is_trigger_report: rng.gen::<u32>(),
-                    breakdown_key: rng.gen_range(0..MAX_BREAKDOWN_KEY),
-                    trigger_value: rng.gen_range(0..MAX_TRIGGER_VALUE),
-                };
-                (Fp32BitPrime, MatchKey, BreakdownKey)
-            ));
-        }
-        let result: Vec<GenericReportTestInput<Fp32BitPrime, MatchKey, BreakdownKey>> = world
-            .semi_honest(records, |ctx, input_rows| async move {
-                ipa::<Fp32BitPrime, MatchKey, BreakdownKey>(
-                    ctx,
-                    &input_rows,
-                    PER_USER_CAP,
-                    MAX_BREAKDOWN_KEY,
-                    NUM_MULTI_BITS,
-                )
-                .await
-                .unwrap()
-            })
-            .await
-            .reconstruct();
-
-        assert_eq!(MAX_BREAKDOWN_KEY, result.len() as u128);
-    }
-
+    #[derive(Debug)]
     struct TestRawDataRecord {
         user_id: usize,
         timestamp: usize,
         is_trigger_report: bool,
         breakdown_key: usize,
+        trigger_value: u32,
     }
 
     #[tokio::test]
     #[allow(clippy::missing_panics_doc)]
-    #[ignore]
+    //#[ignore]
     pub async fn random_ipa_cap_one_check() {
         const MAX_BREAKDOWN_KEY: usize = 16;
         const NUM_USERS: usize = 20;
@@ -827,7 +783,8 @@ pub mod tests {
         let mut rng = thread_rng();
 
         let mut raw_data = Vec::with_capacity(NUM_USERS * MAX_RECORDS_PER_USER);
-        let mut expected_results = vec![0_u128; MAX_BREAKDOWN_KEY];
+        let mut expected_results = vec![0_u32; MAX_BREAKDOWN_KEY];
+        let per_user_cap: u32 = 1;
         for _ in 0..NUM_USERS {
             let random_user_id = rng.gen_range(0..MAX_USER_ID);
             let num_records_for_user = rng.gen_range(1..MAX_RECORDS_PER_USER);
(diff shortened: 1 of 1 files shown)


src/protocol/ipa/mod.rs
```diff
@@ -760,158 +762,141 @@ pub mod tests {
...
-                    };
-                    (Fp32BitPrime, MatchKey, BreakdownKey)
-                )
-            })
-            .collect::<Vec<_>>();
+                raw_data.append(&mut records_for_user);
+            }
+            raw_data.sort_unstable_by(|a, b| a.timestamp.cmp(&b.timestamp));
+
+            //println!("raw data (before secret-sharing): {raw_data:#?}");
```
martinthomson, line 849:
```suggestion
```

src/protocol/ipa/mod.rs
```diff
@@ -760,158 +762,141 @@ pub mod tests {
...
                 }
 
-                if record.is_trigger_report {
-                    encountered_trigger = true;
-                } else if encountered_trigger {
-                    expected_results[record.breakdown_key] += 1;
-                    contributed_once = true;
-                    encountered_trigger = false;
+                // sort in reverse time order
+                records_for_user.sort_unstable_by(|a, b| b.timestamp.cmp(&a.timestamp));
```
martinthomson, line 824: This (from the `for _ in 0..NUM_USERS` onward) looks like something that could be its own function.  Maybe `random_user_records(...)`.  I'll leave it to you as to whether you want to pass constants as arguments or move them to the function.

src/protocol/ipa/mod.rs
```diff
@@ -760,158 +762,141 @@ pub mod tests {
...
+                        pending_trigger_value += record.trigger_value;
+                    } else if pending_trigger_value > 0 {
+                        let delta_to_per_user_cap = per_user_cap - total_contribution;
+                        let capped_contribution =
+                            std::cmp::min(delta_to_per_user_cap, pending_trigger_value);
+                        expected_results[record.breakdown_key] += capped_contribution;
+                        total_contribution += capped_contribution;
+                        pending_trigger_value = 0;
+                    }
                 }
```
martinthomson, line 843: This bit where you add to the expectation looks like another useful function to break out. It can take `records_for_user: &[...], expected_results: &mut Vec<u32>`.

src/protocol/ipa/mod.rs
```diff
@@ -760,158 +762,141 @@ pub mod tests {
...
-            records_for_user.sort_unstable_by(|a, b| b.timestamp.cmp(&a.timestamp));
+        let random_seed = thread_rng().gen();
+        println!("Using random seed: {random_seed}");
 
-            let mut encountered_trigger = false;
-            let mut contributed_once = false;
-            for record in &records_for_user {
-                if contributed_once {
-                    continue;
+        for per_user_cap in [1, 3] {
```
martinthomson, line 789: The body of this loop is probably another function too.

martinthomson (changes requested):
Cleanup time.

src/protocol/ipa/mod.rs
```diff
@@ -771,132 +771,195 @@ pub mod tests {
...
-                        let delta_to_per_user_cap = per_user_cap - total_contribution;
-                        let capped_contribution =
-                            std::cmp::min(delta_to_per_user_cap, pending_trigger_value);
-                        expected_results[record.breakdown_key] += capped_contribution;
-                        total_contribution += capped_contribution;
-                        pending_trigger_value = 0;
-                    }
-                }
+            if total_contribution >= per_user_cap {
+                continue;
```
martinthomson, line 839:
```suggestion
                break;
```

src/protocol/ipa/mod.rs
```diff
@@ -771,132 +771,195 @@ pub mod tests {
...
-    #[allow(clippy::missing_panics_doc, clippy::too_many_lines)]
-    //#[ignore]
-    pub async fn random_ipa_check() {
-        const MAX_BREAKDOWN_KEY: usize = 16;
-        const MAX_TRIGGER_VALUE: u32 = 5;
-        const NUM_USERS: usize = 10;
-        const MAX_RECORDS_PER_USER: usize = 8;
-        const NUM_MULTI_BITS: u32 = 3;
+    fn generate_random_user_records_in_reverse_chronological_order(
+        rng: &mut StdRng,
```
martinthomson, line 775:
```suggestion
        rng: &mut impl Rng,
```

src/protocol/ipa/mod.rs
```diff
@@ -771,132 +771,195 @@ pub mod tests {
...
+        for _ in 0..NUM_USERS {
+            let records_for_user = generate_random_user_records_in_reverse_chronological_order(
+                &mut rng,
+                MAX_RECORDS_PER_USER,
+                MAX_BREAKDOWN_KEY,
+                MAX_TRIGGER_VALUE,
+            );
+            random_user_records.push(records_for_user);
+        }
+        let mut raw_data = random_user_records.concat();
```
martinthomson, line 936: I would use `Vec::append()` for each additional user rather than building two vectors.

src/protocol/ipa/mod.rs
```diff
@@ -771,132 +771,195 @@ pub mod tests {
...
+
+        // Sort the records in chronological order
+        // This is part of the IPA spec. Callers should do this before sending a batch of records in for processing.
+        raw_data.sort_unstable_by(|a, b| a.timestamp.cmp(&b.timestamp));
```
martinthomson, line 940: One thing that we might loosen as a constraint here is that maybe the caller only needs to sort those events that *might* be from the same person.  That means that if they know records are from different people, they can ignore the relative sort order. That might make certain types of sorting more efficient.

I was realized that this test would work fine without this step.  Of course, then our sort doesn't really do anything of use, so of course we should do the full sort in our test case...

martinthomson (approved):

src/protocol/ipa/mod.rs
benjaminsavage: Oh nice. Now that we've refactored this into a separate function I can easily do that. It was a bit gnarly when it was inside a doubly nested for loop before =)

src/protocol/ipa/mod.rs
benjaminsavage: The good reason to do this, is that I can pull the random test data *outside* of the for loop instead of re-generating it each time. But since the expected result changes when you modify per_user_cap, I need to recalculate that, and if I've already appended all of the records from all the users together, that's a lot harder to do.

src/protocol/ipa/mod.rs
benjaminsavage: That is an interesting observation. I'm not sure how we can utilize it - but I really hope we can find a way!

If you're running source-fanout query, and you are showing ads to signed-in users, you can easily segment your source-events into multiple groups of non-overlapping users. Let's say 100 groups. If you had a load of trigger-events you wanted to try to join up with them, I suppose you could just add them to ALL of the groups, and then sort all 100 groups separately (and do attribution / capping / aggregation separately as well). Any given trigger event could only affect one of those groups, so there should be no double counting (might need some kind of proof that it was segmented right to be sure tho...)

This would *definitely* be faster if trigger events are relatively rare compared to source events. It would *definitely* be a lot of extra work in the case that source and trigger events come in equal volumes.

PR #482, opened 2023-02-20
Minor optimization. Skip one row.

benjaminsavage:
Extremely minor performance improvement.

In the PER_USER_CAP = 1 variant of the code:
...we checking if we need to filter out a given row.
...we do that by checking if the _following_ rows contain at least one contribution from the same user.

So that means nobody was ever reading the very *first* row.

This PR just avoids computing this value which was never read.


richajaindce (approved):
Reducing cost one skip at a time :)

PR #485, opened 2023-02-21
Implement look-back window

taikiy:
This PR implements "attribution window" capping protocol. If a trigger event's timestamp from its nearest source event exceeds a given "window" (varies depending on the type of ads. 1, 7, 28 days are commonly seen), we set trigger_value to 0. This protocol will be the first to run before any attribution protocols. Trigger events with 0 value will not be counted in later attribution protocols. One exception is when per_user_cap is set to 1, then all trigger events, including value = 0, are attributed. We could conditionally count trigger events with values >= 1 *if* attribution window is configured.

The protocol uses the binary tree thing, but I couldn't merge the code because the logic differs. I'm actually wondering if we could eliminate the use of `b` in `do_the_binary_tree_thing()`. Anyway, that's for another diff.


src/protocol/attribution/input.rs
```diff
@@ -19,9 +19,52 @@ use std::marker::PhantomData;
...
+            is_trigger_report,
+            helper_bit,
+            breakdown_key,
+            trigger_value,
+            _marker: PhantomData,
+        }
+    }
+}
+
+pub type MCApplyAttributionWindowOutputRow<F, T> = MCAccumulateCreditInputRow<F, T>;
```
akoshelev, line 62: maybe we can work on more concise naming here. this type is in inside `attribution` module so we can eliminate the `attribution`word from it. In general it feels like the input and output for each circuit could be expressed as `protocol::circuit::input[row]`, `protocol::circuit::output[row]`. For this example, they could look like this

```rust
mod attribution {
     ... 
    mod window {
         ...
         pub struct InputRow<F, T> { ... }
         pub type OutputRow<F, T> = ...
    
         async fn apply<F, C, T>(ctx: C, input: &[InputRow<F, T>], ...
) -> Result<Vec<OutputRow<F, T>>, Error> { .. }
    }
}

```

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+/// of time deltas in `F`.
+///
+/// # Errors
+/// Fails if the bit-decomposition, bitwise comparison, or multiplication fails.
+async fn zero_out_expired_trigger_values<F, C, T>(
+    ctx: &C,
+    input: &[MCApplyAttributionWindowInputRow<F, T>],
+    time_delta: &[T],
+    cap: u32,
+) -> Result<Vec<T>, Error>
```
akoshelev, line 175: you may want to talk to @richajaindce as she recently put a lot of effort refactoring these allocations for protocols that process input rows one by one. This seems like another good use case for iterators that could simplify our life in the future. As a bonus point it should allow zipping `time_delta` and `input` together with no runtime overhead

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+        .enumerate()
+    {
+        let record_count = num_rows - step_size;
+
+        let depth_i_ctx = ctx
+            .narrow(&InteractionPatternStep::from(depth))
+            .set_total_records(record_count);
+        let mt_ctx = depth_i_ctx.narrow(&Step::PreviousStopBitTimesCurrent);
+
+        let mut futures = Vec::with_capacity(record_count);
```
akoshelev, line 122: I don't think you need a vector here, `try_join_all` takes an iterator, so you can feed it one created from `0..record_count` range

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+        .into_iter()
+        .chain(
+            try_join_all(
+                zip(input.iter(), input.iter().skip(1))
+                    .zip(stop_bits.iter().skip(1))
+                    .enumerate()
+                    .map(|(i, ((prev, curr), b))| {
+                        let c = t_delta_context.clone();
+                        let record_id = RecordId::from(i);
+                        let delta = curr.timestamp.clone() - &prev.timestamp;
```
akoshelev, line 101: this clone is annoying - one day I tried to generalize `Add` for `ArithmeticRefOps` to work for `&T` and it was a spectacular waste of 3 hours. If someone knows how to do it, please let me know

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+        .collect::<Vec<_>>();
+
+    // First, create a vector of timedeltas. This vector contains non-zero values only for
+    // rows with `stop_bit` = 1, meaning that the row is a trigger event, and has the same
+    // match key as the event one above.
+    let t_delta_context = ctx
+        .narrow(&Step::InitializeTimeDelta)
+        .set_total_records(num_rows - 1);
+    let mut t_delta = Some(T::ZERO)
+        .into_iter()
```
akoshelev, line 92: isn't it just `std::iter::once(T::ZERO).chain(...)`?

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+            .for_each(|(i, (delta, stop_bit))| {
+                t_delta[i + step_size] += &delta;
+                stop_bits[i + step_size] = stop_bit;
+            });
+    }
+
+    Ok(t_delta)
+}
+
+/// Creates a vector of trigger values where values are set to `0` if the time delta
```
akoshelev, line 162: is it possible to replace `trigger_value` instead of making a fresh new vector with updated values? that may be cheaper if we can just update them by mutably borrowing the inputs

src/test_fixture/input/mod.rs
```diff
@@ -51,6 +51,28 @@ macro_rules! ipa_test_input {
     };
 }
 
+#[macro_export]
+macro_rules! attribution_window_test_input {
```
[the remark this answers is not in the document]
taikiy: I think this is helpful that it lets us write, and more importantly, see what we are passing as test inputs to our protocols.

```
ipa_test_input!(
    [
        { match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
        { match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
        { match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
```

If we were to use a function, we'll see integers only. That was what we had before and it and made very difficult to understand the tests. If we use struct constructions, line get too long. But, I'll be more than happy to change it as long as we can ensure the readability.

src/protocol/attribution/input.rs
taikiy: I do agree with your idea of separating words into module paths we discussed in NYC. That said, this particular case, "Attribution Window", the words are inseparable as it's a technical term. It's like separating `SemiHonestContext` to `mod semi { mod honest { struct Context } }`.

src/protocol/attribution/apply_attribution_window.rs
taikiy: Ah, right. I noticed the change after I started to work on this. Will change them to iterators where possible.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+    // match key as the event one above.
+    let t_delta_context = ctx
+        .narrow(&Step::InitializeTimeDelta)
+        .set_total_records(num_rows - 1);
+    let mut t_delta = Some(T::ZERO)
+        .into_iter()
+        .chain(
+            try_join_all(
+                zip(input.iter(), input.iter().skip(1))
+                    .zip(stop_bits.iter().skip(1))
```
benjaminsavage, line 96: This makes me wonder why you added a ZERO to the front of this vector. If we are just going to skip it, why not not add it in the first place?

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+                        let delta = curr.timestamp.clone() - &prev.timestamp;
+                        async move { c.multiply(record_id, &delta, b).await }
+                    }),
+            )
+            .await?,
+        )
+        .collect::<Vec<_>>();
+
+    // Do the prefix-sum accumulation of timedeltas. This is similar to
+    // `do_the_binary_tree_thing()`, but it sums up values in top-down order.
```
benjaminsavage, line 110: I think you can achieve this same effect by simply reversing the iterator, then calling do_the_binary_tree_thing()

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+            let previous_stop_bit = &stop_bits[index - step_size];
+            let current_stop_bit = &stop_bits[index];
+            let previous_t_delta = &t_delta[index - step_size];
+
+            futures.push(async move {
+                try_join(
+                    // if the current stop_bit is 0, don't accumulate
+                    c1.multiply(record_id, previous_t_delta, current_stop_bit),
+                    // next `stop_bit`
+                    c2.multiply(record_id, previous_stop_bit, current_stop_bit),
```
benjaminsavage, line 142: We definitely need to call the same `do_the_binary_tree_thing` function. This is missing so many of the optimizations we put into that thing, like the whole "don't compute the stop bits on the final iteration" stuff.

benjaminsavage: Oh, and also the whole "don't compute stop bits on the final record" optimization.

src/protocol/attribution/apply_attribution_window.rs
benjaminsavage: I agree.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+    let bit_decomposition_ctx = c.narrow(&Step::TimeDeltaBitDecomposition);
+    let cmp_ctx = c.narrow(&Step::TimeDeltaLessThanCap);
+    let mul_ctx = c.narrow(&Step::CompareBitTimesTriggerValue);
+
+    let random_bits_generator =
+        RandomBitsGenerator::new(ctx.narrow(&Step::RandomBitsForBitDecomposition));
+    let rbg = &random_bits_generator;
+
+    try_join_all(
+        zip(input, time_delta)
```
benjaminsavage, line 193: I think you can use input.iter().skip(1) or something to skip the first row because:
- It's either a source or a trigger report
- If it's a source report, then the result will be zero due to the final multiplication
- if it's a trigger report, then there is no way there is any source event before it to which it could have been attributed.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+                let c3 = mul_ctx.clone();
+                let record_id = RecordId::from(i);
+
+                async move {
+                    let delta_bits = BitDecomposition::execute(c1, record_id, rbg, delta).await?;
+                    let compare_bit = one
+                        - &bitwise_greater_than_constant(c2, record_id, &delta_bits, cap.into())
+                            .await?;
+                    c3.multiply(record_id, &row.trigger_value, &compare_bit)
+                        .await
```
benjaminsavage, line 208: Seems about right! 

On the one hand, I'm glad we have all of these powerful and useful protocols available for use, so that we can implement new functionality like this really quickly!

On the other hand, I'm crying inside because these are such expensive protocols!

src/protocol/attribution/input.rs
```diff
@@ -19,9 +19,52 @@ use std::marker::PhantomData;
...
+    pub timestamp: AdditiveShare<F>,
+    pub is_trigger_report: AdditiveShare<F>,
+    pub helper_bit: AdditiveShare<F>,
+    pub breakdown_key: XorShare<BK>,
+    pub trigger_value: AdditiveShare<F>,
+}
+
+#[derive(Debug)]
+pub struct MCApplyAttributionWindowInputRow<F: Field, T: Arithmetic<F>> {
+    pub timestamp: T,
```
benjaminsavage, line 35: I do not like this name. This actually indicates a _time delta_, not a timestamp. That confused me when I was looking at the test and expected to see monotonically increasing timestamps.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+    #[tokio::test]
+    pub async fn attribution_window() {
+        const ATTRIBUTION_WINDOW: u32 = 600;
+        const EXPECTED: &[u128; 23] = &[
+            0, 0, 0, 10, 2, 1, 5, 1, 0, 0, 0, 10, 0, 3, 12, 0, 0, 6, 4, 0, 6, 1, 0,
+        ];
+
+        let input: Vec<GenericReportTestInput<Fp32BitPrime, MatchKey, BreakdownKey>> = attribution_window_test_input!(
+            [
+                { timestamp: 500, is_trigger_report: 0, helper_bit: 0, breakdown_key: 3, credit: 0 }, // delta: 0
```
benjaminsavage, line 270: Are these actually timestamps, or are they deltas?

If they are actually timestamps, then 500 should not come before 100. That was super confusing to me.

benjaminsavage: Oh I see... it's from a different user... so it's OK. wow this is subtle.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,351 @@
...
+                { timestamp: 130, is_trigger_report: 0, helper_bit: 1, breakdown_key: 4, credit: 0 }, // delta: 0
+                { timestamp: 150, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 10 },// delta: 20
+                { timestamp: 250, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 2 }, // delta: 120
+                { timestamp: 310, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 1 }, // delta: 180
+                { timestamp: 420, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 5 }, // delta: 290
+                { timestamp: 540, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 1 }, // delta: 410
+                { timestamp: 890, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 3 }, // delta: 760
+                { timestamp: 920, is_trigger_report: 1, helper_bit: 1, breakdown_key: 0, credit: 5 }, // delta: 790
+                { timestamp: 110, is_trigger_report: 0, helper_bit: 0, breakdown_key: 1, credit: 0 }, // delta: 0
+                { timestamp: 310, is_trigger_report: 1, helper_bit: 0, breakdown_key: 0, credit: 10 },// delta: 0
```
benjaminsavage, line 281: comment is wrong. delta is actually 200

benjaminsavage: Gah... nm... helper_bit = 0, so it's just not the same user... Maybe instead of `delta: 0`, you could just say: `different match key`

src/protocol/attribution/input.rs
```diff
@@ -19,9 +19,52 @@ use std::marker::PhantomData;
 use typenum::Unsigned;
 
 //
-// `accumulate_credit` protocol
+// 'apply_attribution_window` protocol
```
benjaminsavage, line 22: use backticks at the beginning and end

src/protocol/attribution/input.rs
benjaminsavage: nevermind... it actually is timestamps =)

benjaminsavage (changes requested):
Looks pretty good.

Just putting it back in your queue to merge with the normal `do_the_binary_tree_thing` function.

src/test_fixture/input/mod.rs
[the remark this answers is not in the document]
taikiy: Then, we'll need to implement that constructor for many structs and more duped code? Not sure which is better.

src/protocol/attribution/apply_attribution_window.rs
taikiy: Not sure how I do this because `time_delta` cannot be moved inside `async move`?

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,303 @@
...
+where
+    F: Field,
+    C: Context,
+    T: Arithmetic<F> + BasicProtocols<C, F>,
+{
+    let num_rows = input.len();
+
+    // Pre-compute `is_trigger_bit * helper_bit'.
+    // Later in the prefix-sum loop, this vector is updated in each iteration to help
+    // accumulate timedeltas and determine when to stop accumulating.
```
benjaminsavage, line 71: We already did this back in the accumulate step! I'll bet that (someday) we can find a way of just doing this one time to save on mults.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -0,0 +1,303 @@
...
+        .chain(
+            try_join_all(
+                zip(input.iter(), input.iter().skip(1))
+                    .zip(stop_bits.clone().skip(1))
+                    .enumerate()
+                    .map(|(i, ((prev, curr), b))| {
+                        let c = t_delta_context.clone();
+                        let record_id = RecordId::from(i);
+                        let delta = curr.timestamp.clone() - &prev.timestamp;
+                        async move { T::multiply(c, record_id, &delta, &b).await }
```
benjaminsavage, line 101: I kind of feel like this should be taken care of by the `do_the_binary_tree_thing` function. I'm not sure why you have to do this multiplication.

src/protocol/attribution/apply_attribution_window.rs
benjaminsavage: We can do this in a follow-up diff.

src/protocol/attribution/apply_attribution_window.rs
benjaminsavage: ...but this is a *micro* optimization. So feel free to do this in some later diff.

src/test_fixture/input/mod.rs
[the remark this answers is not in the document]
benjaminsavage: As discussed in person, if we can actually get a concrete measurement of how much extra compile-time these macros cost, then we can make an informed tradeoff. At the moment, I'm in favor of leaving them around as we don't have data to prove that removing them will make compilation faster.

src/protocol/attribution/apply_attribution_window.rs
taikiy: I'll follow up in a later diff.

src/protocol/attribution/apply_attribution_window.rs
taikiy: This part is the conversion from timestamps to time deltas. There might be a clever way of doing this without multiplications.

PR #486, opened 2023-02-21
Format imports with rustfmt

akoshelev:
This change uses [unstable rustfmt option](
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity) to auto format imports. There is a [little trick](https://github.com/rust-lang/rustfmt/issues/5511) to make it work on a stable channel. It seems to me that benefits outweigh the inconvenience of adding arguments to `cargo fmt`. 

It also updates pre-commit and CI checks and fixes imports for pretty much every file.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,7 +41,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
 
       - name: Check formatting
-        run: cargo fmt --all -- --check
+        run: cargo fmt --all -- --check --config imports_granularity=Crate
 
       - name: Clippy
         if: ${{ success() || failure() }}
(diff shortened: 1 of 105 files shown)


pre-commit
```diff
@@ -53,9 +53,9 @@ stashfile=$(mktemp .pre-commit.stashXXXXXX)
 trap 'set +e;git stash pop -q; rm -f "$stashfile"; restore_merge_files' EXIT
 save_merge_files
 git stash push -k -u -q -m "pre-commit stash"
-if ! errors=($(cargo fmt -- --check -l)); then
+if ! errors=($(cargo fmt -- --check --config imports_granularity=Crate -l)); then
```
akoshelev, line 56: key change

.github/workflows/check.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
 
       - name: Check formatting
-        run: cargo fmt --all -- --check
+        run: cargo fmt --all -- --check --config imports_granularity=Crate
```
akoshelev, line 44: key change

martinthomson (approved):
Wow, very nice.

PR #487, opened 2023-02-22
Remove the intermediate `b` bit

taikiy:
The aggregation protocols accumulate values using a binary-tree style prefix-sum. The current code computes an intermediate `b` bit and multiply it by `credit` to determine whether a value should be accumulated in each iteration.

How we currently compute the `b` bit is by `sibling_helper_bit * current_stop_bit`. Since the initial `stop_bit` is computed by `sibling_helper_bit * sibling_is_trigger_bit`, we can eliminate `b` and just compute `current_stop_bit * sibling_value`.

Since call sites do not need helper_bits after calling `do_the_binary_tree_thing()`, I changed helper_bits to move. We also don't need to handle a special case for depth 0 and not need to create new vectors for credits, I changed credits to mutable borrow.

This change reduces 1 mult from per depth * N rows * 3 protocols, and 3N heap allocations * 3 protocols.

```
Baseline for semi-honest IPA (cap = 1) has improved! Expected 7557, got 7551.
Baseline for malicious IPA (cap = 1) has improved! Expected 18849, got 18837.
Baseline for semi-honest IPA (cap = 3) has improved! Expected 10740, got 10731.
Baseline for malicious IPA (cap = 3) has improved! Expected 26410, got 26392.
```

### DIFF
--- a/src/protocol/attribution/mod.rs
+++ b/src/protocol/attribution/mod.rs
@@ -162,49 +162,30 @@ where
 /// ## Panics
 /// Nah, it doesn't.
 ///
-pub async fn do_the_binary_tree_thing<'a, F, C, S, I>(
+pub async fn do_the_binary_tree_thing<F, C, S>(
     ctx: C,
-    helper_bits: &[S],
-    credits: I,
+    stop_bits: &[S],
+    values: &mut [S],
 ) -> Result<Vec<S>, Error>
 where
     F: Field,
     C: Context<F, Share = S>,
-    S: ArithmeticSecretSharing<F> + 'a,
-    I: Iterator<Item = &'a S>,
+    S: ArithmeticSecretSharing<F>,
 {
-    let num_rows = helper_bits.len() + 1;
-    let depth_0_ctx = ctx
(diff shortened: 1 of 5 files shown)


src/protocol/attribution/mod.rs
```diff
@@ -162,77 +162,52 @@ where
 /// ## Panics
 /// Nah, it doesn't.
 ///
-pub async fn do_the_binary_tree_thing<'a, F, C, S, I>(
+pub async fn do_the_binary_tree_thing<F, C, S>(
     ctx: C,
-    helper_bits: &[S],
-    credits: I,
+    stop_bits: &[S],
+    values: &mut [S],
```
taikiy, line 168: Renamed to `values` since we can probably use this logic to accumulate time deltas for lookback window protocol.

benjaminsavage (approved):
Amazing. It's not often you get to see a diff which simultaneously makes code simpler to read, shorter, AND more efficient. Well done @taikiy! Nice work spotting this optimization!

PR #488, opened 2023-02-22
remove type parameter from `Context` and `SemiHonestContext`

andyleiserson:


src/protocol/basics/share_known_value/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+            Role::H1 => Self::new(value, V::ZERO),
+            Role::H2 => Self::new(V::ZERO, V::ZERO),
+            Role::H3 => Self::new(V::ZERO, value),
+        }
+    }
+}
+
+impl<'a, F: Field> ShareKnownValue<MaliciousContext<'a, F>, F> for MaliciousReplicated<F> {
+    fn share_known_value(ctx: &MaliciousContext<'a, F>, value: F) -> Self {
+        // TODO: I don't actually think r_share should be exposed like this
```
martinthomson, line 35: Say more?  You are probably right, in that the share should not be used, except under tightly controlled conditions, but this seems like pretty clean code and an exception to the usual rule where values should be upgraded and added to the MAC, rather than using r_share().

src/protocol/basics/share_known_value/mod.rs
```diff
@@ -0,0 +1,41 @@
+use crate::bits::Fp2Array;
```
martinthomson, line 1: I know that we've done this before, but I don't like it when we make a module that is a single mod.rs in a directory.  It's harder to find.

src/protocol/basics/reshare.rs
```diff
@@ -30,36 +30,39 @@ use futures::future::try_join;
...
-
-    async fn reshare(
-        self,
-        input: &Self::Share,
-        record: RecordId,
+pub trait Reshare<C: Context>: Sized {
+    async fn reshare<'fut>(
+        ctx: C,
+        record_id: RecordId,
+        input: &Self,
```
martinthomson, line 37: Any reason that this isn't `&self`?  I don't see why not.

src/protocol/context/mod.rs
```diff
@@ -80,7 +63,7 @@ pub trait Context<V: SharedValue>:
     #[must_use]
     fn mesh(&self) -> Mesh<'_, '_>;
 
-    fn share_known_value(&self, scalar: V) -> <Self as Context<V>>::Share;
+    //fn share_known_value(&self, scalar: V) -> <Self as Context<V>>::Share;
```
martinthomson, line 66:
```suggestion
```

src/protocol/context/semi_honest.rs
```diff
@@ -121,9 +113,11 @@ impl<'a, F: Field> Context<F> for SemiHonestContext<'a, F> {
         self.inner.gateway.mesh(self.step(), self.total_records)
     }
 
+    /*
     fn share_known_value(&self, scalar: F) -> <Self as Context<F>>::Share {
         Replicated::share_known_value(self.role(), scalar)
     }
+    */
```
martinthomson, line 120:
```suggestion
```

martinthomson (approved):
Ship it.  I like it.  I don't really care that it's a draft.

akoshelev:
I like this change too. I think originally the idea was to customize the buffer size inside `Gateway` using the compile-time information about `F` but we don't strictly need that - context creation happens in one place and we can just run a runtime check there. 

We don't seem to lose anything on upgrade to malicious as well - it will still prevent to multiply if field type does not match

akoshelev (approved):

src/protocol/basics/reshare.rs
andyleiserson: We could do it that way, and it would save needing to call the methods as `Type::method`. (I don't find `T::method` too bad; in other places it is `Replicated::method` or `MaliciousReplicated::method` which is worse.)

I kind of like having the context be the first argument (and that's what we do in a bunch of existing places), but there's no reason it has to be. But, for multiplies, we have to take a 2nd operand:
* `Replicated::multiply(ctx, record, a, b)`
* `a.multiply(ctx, record, b)`
* `a.multiply(b, ctx, record)`
* `(a, b).multiply(ctx, record)` (I think this goes in the "too clever for it's own good" department.)

I would probably pick the 3rd over the 2nd (keep the operands closer together), but that introduces an inconsistency if we otherwise put `ctx` first.

src/protocol/basics/share_known_value/mod.rs
andyleiserson: The thing that made me cringe (and where I first added this comment) was what this calls -- `pub fn r_share(&self)` on MaliciousContext. To me it's a similar argument to why `MaliciousReplicated` shouldn't allow direct access to `x`.

In my cleanups I changed this so r_share is not exposed -- I moved the `known * r` calculation back inside a method on MaliciousContext.

This specific issue aside, I do have some concerns about the malicious share-known-value protocol -- it seems dangerous to me to combine the private shares of `r` with publicly known constants. The resultant shares are effectively as sensitive as shares of `r`. If you do something like share-known-value followed by reveal (admittedly a pointless combination), then you've leaked the `r` share to another party. I made a note to think about this some more -- or track down a reference for the malicious share-known-value protocol.

martinthomson: Sounds like you have it in hand, which is good.

Yes, `known * r` is a problem if revealed, but that is true for any value that might be revealed.  Revealing is something we only do once we have confidence that the value is safe to reveal.

We require that `r` is revealed before revealing anything that includes it.  I wonder if there is any way to structure the code to enforce that constraint.  It's the opposite of the usual lifetime thing in Rust though (`'x: 'share - 'r` isn't a thing you can do, and probably for good reason.)

src/protocol/basics/reshare.rs
martinthomson: The third is better, yeah.  I am not a huge fan of having to bring in the trait (with `use`) and then use `Struct::method` to invoke what is better conceived of as a function on the operands.  If only because it is more verbose than necessary and finding the right type to use as a base is awkward.

martinthomson: Oh, I should say, I'm going to merge this now; do that part later :)

PR #489, opened 2023-02-22
Add more events to the baseline test

taikiy:
Credit capping executes the prefix-or function on each row log N times. I was working on an optimization to:

* Reduce 1 mult per iteration from N log N loop
* Add N mults per function call

The baseline test's result was an increase in the communication because the loop executed only once. As the `N` increases, however, the performance gain is obvious.

This PR adds 4 more rows, just enough to execute the said loop few more times, to be able to see the optimization results better. As the test is executed in parallel, it doesn't change the overall test execution time.

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -1002,16 +1002,16 @@ pub mod tests {
         const NUM_MULTI_BITS: u32 = 3;
 
         /// empirical value as of Feb 4, 2023.
-        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 10740;
+        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 19290;
 
         /// empirical value as of Feb 14, 2023.
-        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 26410;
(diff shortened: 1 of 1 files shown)


benjaminsavage:
Hmmm... Is it really necessary to add those extra N mults per function call? Can those potentially be eliminated? I don't feel like adding more rows to the test is the right solution. It obfuscates the performance regression on small sets. It'll also make the runtime longer.

taikiy:
Just created #491 for reference.

> Is it really necessary to add those extra N mults per function call?

I'll try and see if I can eliminate all or some of them.

That said, I think, with (extremely) small sets, obfuscation is greater. It's difficult to see the impact from reducing a fixed number of mults vs. repetitive mults. The baseline test for #491 will only fail for 6 or less events, but passes for 7+, which is probably 100% of cases in real world. Even with larger sets, reducing a fixed number of mults will still be visible, so I believe it'll only benefit us to have a larger set, as long as tests finish in a reasonable duration.

As for the runtime, I executed `cargo test` multiple times and found that the total runtime was not affected by this change.

benjaminsavage:
OK, now that we've landed #491 - I think this PR makes sense. This baseline test is just too small to really be useful as a means of evaluating optimizations. We need a few more rows to evaluate impacts to the ( n * log(n) ) parts of the code.

akoshelev (approved):

PR #490, opened 2023-02-22
Skipping the computation of 'stop bits' on the last iteration

benjaminsavage:
On the last iteration, there is no need to update the "stop bits" because they won't be read in the last iteration. We can skip this and save on communication bandwidth!


taikiy (approved):
Yet another neat optimization :)

PR #491, opened 2023-02-22
Remove `b` bit from `prefix_or_binary_tree_style`

taikiy:
Same as #487 + #490. The baseline result doesn't change because it only has 5 events, but the improvement is still 1 less mult from O(N log N) logic.

### DIFF
--- a/src/protocol/attribution/mod.rs
+++ b/src/protocol/attribution/mod.rs
@@ -58,7 +58,7 @@ where
 ///
 pub async fn prefix_or_binary_tree_style<F, C, S>(
     ctx: C,
-    helper_bits: &[S],
+    stop_bits: &[S],
     uncapped_credits: &[S],
 ) -> Result<Vec<S>, Error>
 where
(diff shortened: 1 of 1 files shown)


src/protocol/attribution/mod.rs
```diff
@@ -52,86 +52,56 @@ where
...
-    assert_eq!(helper_bits.len() + 1, uncapped_credits.len());
+    assert_eq!(stop_bits.len() + 1, uncapped_credits.len());
 
     let num_rows = uncapped_credits.len();
-    let depth_0_ctx = ctx
-        .narrow(&InteractionPatternStep::from(0))
-        .set_total_records(num_rows - 1);
 
-    let b_times_sibling_credit_ctx = depth_0_ctx.narrow(&Step::BTimesSuccessorCredit);
+    let mut uncapped_credits = uncapped_credits.to_owned();
```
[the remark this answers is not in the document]
taikiy: Now I remember that credit_capping is a bit different from accumulation and aggregation. At the beginning of the capping protocol, we flip is_trigger_bit and multiply by credits to mask source values. Then we proceed to the binary-tree logic. So, the logic is still the same in this PR, except for it removes a redundant the squared helper_bit when computing the `b` bit.

src/protocol/attribution/mod.rs
```diff
@@ -58,100 +58,88 @@ where
 ///
 pub async fn prefix_or_binary_tree_style<F, C, S>(
     ctx: C,
-    helper_bits: &[S],
+    stop_bits: &[S],
     uncapped_credits: &[S],
```
taikiy, line 62: There is only one reference to this function, and the call site uses both input vectors after calling this function. Hence I kept these as references, different from how I changed them to move/mutable borrow.

benjaminsavage (approved):
I'm surprised this didn't drop the communication baseline test. You added an optimization (removing the computation of stop bits on the last iteration). It should have dropped stuff... Is that because another regression was introduced?

taikiy:
This is because the baseline test input length is 5. If the binary-tree loops twice, the communication baseline is the same.

baseline test input len = 5
accumulate_credit_cap_one output len = 4
prefix_or_binary_tree_style input len = 3

Before:
Depth 0 => 2 events * 2 muls
Depth 1 => 1 event * 4 muls
Total 8 muls

After:
Depth 0 => 2 events * 3 muls
Depth 1 => 1 event * 2 muls
Total 8 muls

src/protocol/attribution/mod.rs
```diff
@@ -58,100 +58,88 @@ where
...
-    prefix_or.push(uncapped_credits[num_rows - 1].clone());
-
-    // Create stop_bit vector.
-    // This vector is updated in each iteration to help accumulate values
+    // This vector is updated in each iteration to help accumulate credits
     // and determine when to stop accumulating.
-    let mut stop_bits = helper_bits.to_owned();
-    stop_bits.push(S::share_known_value(&ctx, F::ONE));
+    let mut stop_bits = stop_bits.to_owned();
+    stop_bits.push(S::ZERO);
```
benjaminsavage, line 78: I just noticed that this swapped from ONE to ZERO... why is that?

taikiy: This is because of that *semantic change* of the stop bits. We now compute `current_stop_bit * sibling_credit`, and the binary-tree loop skips the bottom N - step_size rows. It doesn't matter what value we store here.

And I just realized we can probably skip the last `current_stop_bit * sibling_stop_bit multiplication` in each depth, because it'll always be 0.

Issue #492, opened 2023-02-23
Update encryption


martinthomson:
I just changed the proposal to include the origin of the match key provider.  We also need to change from site origin to site registrable domain (which it should have been all along).  There are lots of rules about how you construct these inputs, but we can just take them as strings.

PR #494, opened 2023-02-24
Update HPKE encryption of match keys

akoshelev:
It addresses the updates in the proposal https://github.com/patcg-individual-drafts/ipa/pull/31/ and IPA issue #492

### DIFF
--- a/src/hpke/info.rs
+++ b/src/hpke/info.rs
@@ -51,7 +51,7 @@ impl<'a> Info<'a> {
         site_registrable_domain: &'a str,
     ) -> Result<Self, NonAsciiStringError<'a>> {
         if !match_key_provider_origin.is_ascii() {
-            return Err(match_key_provider_origin.into())
+            return Err(match_key_provider_origin.into());
         }
 
         if !helper_origin.is_ascii() {
--- a/src/hpke/info.rs
+++ b/src/hpke/info.rs
@@ -7,14 +7,16 @@ const DOMAIN: &str = "private-attribution";
 /// Represents the [`info`] part of the receiver context, that is: application specific data
 /// for each encryption.
 ///
-/// IPA uses key identifier, event epoch and both helper and site origins to authenticate match key
-/// encryption. It is not guaranteed that the same receiver can be used for anything else.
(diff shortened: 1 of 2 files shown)


src/hpke/info.rs
```diff
@@ -7,16 +7,18 @@ const DOMAIN: &str = "private-attribution";
...
 ///
 /// [`info`]: https://www.rfc-editor.org/rfc/rfc9180.html#name-creating-the-encryption-con
 #[derive(Clone)]
 pub struct Info<'a> {
     pub(super) key_id: KeyIdentifier,
     pub(super) epoch: Epoch,
+    pub(super) match_key_provider_origin: &'a str,
     pub(super) helper_origin: &'a str,
-    pub(super) site_origin: &'a str,
+    pub(super) site_registrable_domain: &'a str,
```
martinthomson, line 21:
```suggestion
    pub(super) site_domain: &'a str,
```

This might be easier to type for us.

src/hpke/mod.rs
```diff
@@ -148,8 +148,9 @@ mod tests {
     }
 
     impl<R: RngCore + CryptoRng> EncryptionSuite<R> {
+        const MKP_ORIGIN: &'static str = "";
         const HELPER_ORIGIN: &'static str = "foo";
-        const SITE_ORIGIN: &'static str = "bar";
+        const SRD: &'static str = "xn--mozilla.com.xn--example.com";
```
martinthomson, line 153: "xn--" hah

martinthomson (approved):

src/hpke/mod.rs
akoshelev: this was my interpretation of this sentence

```
The ASCII serialization of the registrable domain for the current site as a period-separated sequence of
A-labels
```

by changing it I wanted to validate my understanding by you - it seems we have to prefix everything with `xn--` according to the [spec](https://www.rfc-editor.org/rfc/rfc5890#section-2.3.2.5) ?

martinthomson: Ah, my error: NR-LDH is also OK.  These are perfectly OK as NR-LDH labels.

PR #495, opened 2023-02-24
Remove the last stop_bit computation from each binary-tree depth

taikiy:
In #487, I changed how we compute the prefix-sum. Determining whether we should accumulate the `sibling_value` only depends on the `current_stop_bit`. That means we skip the bottom stop bits at `stop_bits[(num_rows - step_size + 1)..]`, which allows us to reduce 1 mult from each depth = log N mult reduction.

### DIFF
--- a/src/protocol/attribution/mod.rs
+++ b/src/protocol/attribution/mod.rs
@@ -94,6 +94,7 @@ where
         let mut stop_bit_futures = Vec::with_capacity(end);
 
         for i in 0..end {
+            let last_row = i == end - 1;
             let c1 = new_credit_ctx.clone();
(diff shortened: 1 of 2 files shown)


benjaminsavage (changes requested):
This is great, but can you do the same to the normal path (not just cap=1)?

Also - now you can remove the code that inserts a F::ZERO into the end of the vectors. That's very slow as it resizes the vec.

taikiy:
Removed `push()`

benjaminsavage (approved):
Love it. This is a very nice improvement! I love that we aren't resizing that vector anymore.

PR #497, opened 2023-02-24
Correct record count for one less stop_bit mult

taikiy:
Missed to fix `set_total_records` in #495.

### DIFF
--- a/src/protocol/attribution/mod.rs
+++ b/src/protocol/attribution/mod.rs
@@ -84,12 +84,16 @@ where
         let first_iteration = step_size == 1;
         let last_iteration = step_size * 2 >= num_rows;
         let end = num_rows - step_size;
-        let depth_i_ctx = ctx
-            .narrow(&InteractionPatternStep::from(depth))
+        let depth_i_ctx = ctx.narrow(&InteractionPatternStep::from(depth));
+        let new_credit_ctx = depth_i_ctx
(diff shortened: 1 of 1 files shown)


src/protocol/attribution/mod.rs
```diff
@@ -84,12 +84,16 @@ where
...
             .set_total_records(end);
-        let new_credit_ctx = depth_i_ctx.narrow(&Step::CurrentStopBitTimesSuccessorCredit);
-        let new_stop_bit_ctx = depth_i_ctx.narrow(&Step::CurrentStopBitTimesSuccessorStopBit);
-        let credit_or_ctx = depth_i_ctx.narrow(&Step::CurrentCreditOrCreditUpdate);
+        let credit_or_ctx = depth_i_ctx
+            .narrow(&Step::CurrentCreditOrCreditUpdate)
+            .set_total_records(end);
+        let new_stop_bit_ctx = depth_i_ctx
+            .narrow(&Step::CurrentStopBitTimesSuccessorStopBit)
+            .set_total_records(end - 1);
```
benjaminsavage, line 96: @andyleiserson - can you add some kind of warning to help us catch these types of cases? This time we caught it, but I can easily see us missing this in the future.
If a channel is still open and expecting like, one more record... and its been like 30 seconds and it hasn't come... maybe log a warning or something?

benjaminsavage (approved):
Thanks @taikiy!

PR #498, opened 2023-02-24
Stop computing unused stop bits

benjaminsavage:
Stop bits are used in the next iteration. No sense generating more than will be used in the next iteration.


akoshelev (approved):

akoshelev:
not so great sign - https://github.com/private-attribution/ipa/actions/runs/4263887708/jobs/7421205444.
workflow is stuck running tests. I am not seeing which test caused it, so there is a small chance that it is a glitch in CI, but will keep an eye on it and asking folks to report long-running tests if they see them

PR #502, opened 2023-02-26
Pass by value not reference

benjaminsavage:
This one function was inconsistent, and that bothered me. Pretty much everywhere else we create a new context and pass that in. This was the one exception where we pass by reference.


taikiy (approved):

akoshelev:
lets rebase and merge it @benjaminsavage ?

PR #503, opened 2023-02-27
Make oneshot-sort work again

richajaindce:


akoshelev (approved):
thanks! I am still puzzled why `cargo build --benches --features enable-benches` does not catch those

richajaindce:
> thanks! I am still puzzled why `cargo build --benches --features enable-benches` does not catch those

Its not a compile error but a runtime error. We should run oneshot-sort in precommit hook

pre-commit
```diff
@@ -79,3 +79,5 @@ cargo test || error "Test failures"
 cargo test --features shuttle || error "Concurrency test failures"
 
 cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches" || error "Failed to run IPA benchmark"
+
+cargo bench --bench oneshot_sort --no-default-features --features="enable-benches" || error "Failed to run IPA benchmark"
```
martinthomson, line 83: Missing end of line here.

Also, can we just run very, very small tests in the pre-commit?  This is a pretty large batch.

PR #504, opened 2023-02-27
Optimizing aggregation for small breakdown key counts

benjaminsavage:
When there are very few breakdown keys (i.e. less than 32) it's less communication to just check each breakdown key for equality for every possible breakdown key. Sorting is better when there are a large number of breakdowns.

I did an analysis, plotting the total communication for both approaches. Results: https://docs.google.com/spreadsheets/d/1Nix-mr6VBpGFOnKIwIvniAV3KO5vhk3ydSwxcfh2-OM/edit?usp=sharing

Screenshot 2023-02-27 at 14 52 19

### DIFF
--- a/src/protocol/attribution/aggregate_credit.rs
+++ b/src/protocol/attribution/aggregate_credit.rs
@@ -228,7 +228,7 @@ where
 
 async fn simple_aggregate_credit<F, BK>(
     ctx: SemiHonestContext<'_>,
-    capped_credits: &[MCAggregateCreditInputRow<F, Replicated<F>>],
+    capped_credits: impl Iterator<Item = MCAggregateCreditInputRow<F, Replicated<F>>>,
     max_breakdown_key: u128,
 ) -> Result<Vec<MCAggregateCreditOutputRow<F, Replicated<F>, BK>>, Error>
 where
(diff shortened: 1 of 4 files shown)


src/protocol/sort/mod.rs
```diff
@@ -137,3 +145,139 @@ impl AsRef<str> for MultiBitPermutationStep {
...
+    let mut precomputed_combinations = Vec::with_capacity(1 << num_bits);
+    precomputed_combinations.push(S::share_known_value(&ctx, F::ONE));
+    for (bit_idx, bit) in input.iter().enumerate() {
+        let step = 1 << bit_idx;
+        let mut multiplication_results =
+            try_join_all(precomputed_combinations.iter().skip(1).enumerate().map(
+                |(j, precomputed_combination)| {
+                    let child_idx = j + step;
+                    S::multiply(
+                        ctx.narrow(&BitOpStep::from(child_idx)),
```
taikiy, line 271: Doesn't this cause the BitOpStep to go out of range?

I think you need (2 ^ num_bits - 1) + 1 multiplications, and BitOpStep only has 64 strings. Also, this isn't exactly a bit-by-bit operation, so you might want to create a new step for this.

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -219,6 +226,84 @@ where
...
+        .set_total_records(capped_credits.len());
+    let check_times_credit_context = ctx
+        .narrow(&Step::CheckTimesCredit)
+        .set_total_records(capped_credits.len());
+
+    let futures = capped_credits.iter().enumerate().map(|(i, row)| {
+        let c1 = equality_check_context.clone();
+        let c2 = check_times_credit_context.clone();
+        let credit = &row.credit;
+        let valid_bits_count = (u128::BITS - (max_breakdown_key - 1).leading_zeros()) as usize;
```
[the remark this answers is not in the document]
benjaminsavage: Already done =)

src/protocol/attribution/aggregate_credit.rs
```diff
@@ -219,6 +226,84 @@ where
...
+        let copy_of_valid_bk_bits = row
+            .breakdown_key
+            .iter()
+            .take(valid_bits_count)
+            .cloned()
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
benjaminsavage: Oh wow... that's MUCH better

src/protocol/sort/mod.rs
```diff
@@ -137,3 +145,139 @@ impl AsRef<str> for MultiBitPermutationStep {
...
+    F: Field,
+    C: Context,
+    S: ArithmeticSecretSharing<F> + BasicProtocols<C, F>,
+{
+    let num_bits = record.len();
+    let precomputed_combinations =
+        pregenerate_all_combinations(ctx, record_idx, record, num_bits).await?;
+
+    // This loop just iterates over all the possible values this N-bit input could potentially represent
+    // and checks if the bits are equal to this value. It does so my computing a linear combination of the
```
taikiy, line 173: s/my/by/

src/protocol/sort/mod.rs
benjaminsavage: It uses a lot fewer multiplications.

To check 4 requires 1 mult
To check 8 requires 4 mults
To check 16 requires 11 mults
To check 32 requires 26 mults
To check 64 requires 57 mults

It *will* break if you attempt to use this method for 65 or more breakdown keys, as that requires *120* multiplications per row...

...but the code is only invoked if you are using 32 or less breakdown keys.

I don't see the point in making yet another "Step" enum, as it won't really add any information.

taikiy: Ah, OK. I was worried that if the bit length is 7+ it'll break, but 32 was the cap.

taikiy (approved):

PR #505, opened 2023-02-27
Remove trailing bits from the bit decomposition protocol

martinthomson:
We just throw these out, so remove 'em.  The function now operates modulo 2^l, which is all we use.  We can't do the same for the earlier addition as we need to know about the overflow.

Savings are modest.  So 3 multiplications per event. 27 records in semi-honest; 54 in malicious for our 9 event test.

I also renamed both functions.  The old names annoyed me, because context provided a lot of information that was again repeated in the name.  The new names are shorter.  We should remove "bitwise" from other "boolean" functions, because bitwise is implied in most cases.

### DIFF
--- a/src/protocol/boolean/bit_decomposition.rs
+++ b/src/protocol/boolean/bit_decomposition.rs
@@ -1,6 +1,6 @@
 use super::{
+    add_constant::{add_constant, maybe_add_constant_mod2l},
     bitwise_less_than_prime::BitwiseLessThanPrime,
-    dumb_bitwise_add_constant::{bitwise_add_constant, bitwise_add_constant_maybe},
     random_bits_generator::RandomBitsGenerator,
     RandomBits,
 };
@@ -53,11 +53,10 @@ impl BitDecomposition {
 
         // Step 5. Add back [b] bitwise. [d]_B = BitwiseSum(c, [b]_B) where d ∈ Z
         //
-        // `BitwiseSum` outputs `l + 1` bits, so [d]_B is (l + 1)-bit long.
-        let d_b = bitwise_add_constant(ctx.narrow(&Step::AddBtoC), record_id, &r.b_b, c.as_u128())
-            .await?;
+        // `BitwiseSum` outputs one more bit than its input, so [d]_B is (el + 1)-bit long.
+        let d_b = add_constant(ctx.narrow(&Step::AddBtoC), record_id, &r.b_b, c.as_u128()).await?;
 
-        // Step 6. p <=? d. The paper says "p <? d", but should actually be "p <=? d"
+        // Step 6. p <? d.
         let q_p = BitwiseLessThanPrime::greater_than_or_equal_to_prime(
             ctx.narrow(&Step::IsPLessThanD),
             record_id,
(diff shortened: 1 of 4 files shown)


src/protocol/boolean/bit_decomposition.rs
```diff
@@ -53,32 +53,31 @@ impl BitDecomposition {
...
         // Step 5. Add back [b] bitwise. [d]_B = BitwiseSum(c, [b]_B) where d ∈ Z
         //
-        // `BitwiseSum` outputs `l + 1` bits, so [d]_B is (l + 1)-bit long.
-        let d_b = bitwise_add_constant(ctx.narrow(&Step::AddBtoC), record_id, &r.b_b, c.as_u128())
-            .await?;
+        // `BitwiseSum` outputs one more bit than its input, so [d]_B is (el + 1)-bit long.
+        let d_b = add_constant(ctx.narrow(&Step::AddBtoC), record_id, &r.b_b, c.as_u128()).await?;
 
-        // Step 6. p <=? d. The paper says "p <? d", but should actually be "p <=? d"
+        // Step 6. p <? d.
```
benjaminsavage, line 59: I don't think you can drop the equals sign here. We are checking if the input is "greater than or equal to" the prime. If it's equal, this returns true. So I think it must remain:

```// Step 6. p <=? d```

File: src/protocol/boolean/bit_decomposition.rs [hunk 2]
```diff
@@ -53,32 +53,31 @@ impl BitDecomposition {
...
         // Step 8, 9. [g_i] = [q] * f_i
-        // Step 10. [h]_B = [d + g]_B, where [h]_B = ([h]_0,...[h]_(l+1))
+        // Step 10. [h]_B = [d + g]_B, where [h]_B = ([h]_0,...[h]_(el+1))
         //
-        // Again, `BitwiseSum` outputs `l + 1` bits. Since [d]_B is already
-        // `l + 1` bit long, [h]_B will be `l + 2`-bit long.
-        let h_b = bitwise_add_constant_maybe(ctx.narrow(&Step::AddDtoG), record_id, &d_b, x, &q_p)
-            .await?;
+        // Again, `BitwiseSum` outputs an extra bit. Since [d]_B is already
+        // `el + 1` bits long, [h]_B will be `el + 2`-bits long.
```
benjaminsavage, line 75: ...well it isn't any more. Perhaps update the comment to say something about how normal addition would result it i having two extra bits, so to avoid needless waste we are calling the "mod 2 el" version of this function.

src/protocol/ipa/mod.rs
```diff
@@ -1071,8 +1071,8 @@ pub mod tests {
                 "Baseline for semi-honest IPA (cap = {per_user_cap}) has DEGRADED! Expected {semi_honest_baseline}, got {records_sent}.");
 
             if records_sent < semi_honest_baseline {
-                tracing::warn!("Baseline for semi-honest IPA (cap = {per_user_cap}) has improved! Expected {semi_honest_baseline}, got {records_sent}.\
-                                Strongly consider adjusting the baseline, so the gains won't be accidentally offset by a regression.");
+                tracing::warn!("Baseline for semi-honest IPA (cap = {per_user_cap}) has improved! Expected {semi_honest_baseline}, got {records_sent}. \
+                                Consider adjusting the baseline, so the gains won't be accidentally offset by a regression.");
```
benjaminsavage, line 1075: lol. Not strongly? =P

benjaminsavage (approved):
Nice catch! Every optimisation counts!

src/protocol/ipa/mod.rs
martinthomson: Mostly this is just because the missing space annoyed me.

PR #513, opened 2023-03-01
Build integration between new transport and OrderingMpsc 

akoshelev:
see commit messages. Basically this PR does two things
* Make `OrderingMpscReceiver` implement a stream
* Get rid of one associated type on transport and make `send` method to take any stream provided by the caller.

### DIFF
--- a/src/helpers/buffers/ordering_mpsc.rs
+++ b/src/helpers/buffers/ordering_mpsc.rs
@@ -4,10 +4,12 @@ use crate::{
     helpers::{messaging::Message, Error},
 };
 use bitvec::{bitvec, vec::BitVec};
-use futures::FutureExt;
+use futures::{FutureExt, Stream};
 use generic_array::GenericArray;
 use std::{
+    fmt::{Debug, Formatter},
     num::NonZeroUsize,
+    pin::Pin,
     sync::{
         atomic::{
             AtomicUsize,
@@ -15,6 +17,7 @@ use std::{
         },
         Arc,
     },
+    task::{Context, Poll},
 };
 use tokio::sync::{
     mpsc::{self, Receiver, Sender},
@@ -154,7 +157,7 @@ impl<M: Message> OrderingMpscReceiver<M> {
     }
 
     /// Take the next set of values.  This will not resolve until `min_count` values are available.
-    pub async fn next(&mut self, min_count: usize) -> Option<Vec<u8>> {
+    pub async fn take_next(&mut self, min_count: usize) -> Option<Vec<u8>> {
         // Read all values that are available before trying to return anything.
         while let Some(read) = self.rx.recv().now_or_never() {
             if let Some((index, msg)) = read {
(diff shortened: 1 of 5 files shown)


src/helpers/buffers/ordering_mpsc.rs
```diff
@@ -412,11 +447,20 @@ mod unit {
...
         assert_eq!(buf.len(), 3 * FP32BIT_SIZE);
         let send3 = tx.send_test(3);
         assert!(send3.now_or_never().is_some());
     }
+
+    #[tokio::test]
+    async fn recv_stream() {
+        let (tx, mut rx) = ordering_mpsc("test", NonZeroUsize::new(2).unwrap());
+        tx.send_test(1).await;
+        assert_eq!(None, StreamExt::next(&mut rx).now_or_never());
```
martinthomson, line 460:
```suggestion
        assert!(StreamExt::next(&mut rx).now_or_never().is_none());
```

src/helpers/buffers/ordering_mpsc.rs
```diff
@@ -194,6 +197,37 @@ impl<M: Message> OrderingMpscReceiver<M> {
...
+}
+
+/// [`OrderingMpscReceiver`] is a [`Stream`] that yields chunks of maximum capacity
+/// this instance can hold.
+impl<M: Message> Stream for OrderingMpscReceiver<M> {
+    type Item = Vec<u8>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = Pin::get_mut(self);
+        loop {
```
martinthomson, line 217: I think that this is correct, though the (renamed) async take_next() is quite a bit more complex.  I think that it has to be, so that's OK, but it is maybe worth highlighting.

The subtle difference is that that code can't block if there are available unread items in the channel, whereas this never blocks, so it doesn't need to worry about that.  It makes be wonder whether we can't just rely on the provided implementation of next() that the stream provides.

src/helpers/buffers/ordering_mpsc.rs
```diff
@@ -324,6 +358,7 @@ mod unit {
         },
     };
     use futures::{future::join, FutureExt};
+    use futures_util::StreamExt;
```
martinthomson, line 361: futures_util::StreamExt or futures::stream::StreamExt ?  Or are these aliases?

src/helpers/transport/channelled_transport.rs
```diff
@@ -100,16 +100,15 @@ impl RouteParams<RouteId, QueryId, Step> for (RouteId, QueryId, Step) {
 /// Transport that supports per-query,per-step channels
 #[async_trait]
 pub trait ChannelledTransport: Send + Sync + 'static {
-    type DataStream: Stream<Item = Vec<u8>>;
     type RecordsStream: Stream<Item = Vec<u8>>;
 
     fn identity(&self) -> HelperIdentity;
 
-    async fn send<Q, S, R>(
+    async fn send<D: Stream<Item = Vec<u8>> + Send + 'static, Q, S, R>(
```
martinthomson, line 107: Move to the where clause, given the size, I'd say.

src/test_fixture/transport/channeled_transport.rs
```diff
@@ -127,9 +131,12 @@ impl ChannelledTransport for InMemoryChannelledTransport {
         let channel = self.get_channel(dest);
         let packet = Addr::from_route(self.identity, &route);
```
martinthomson, line 132:
```suggestion
        let addr = Addr::from_route(self.identity, &route);
```

martinthomson (approved):
Progress!

src/helpers/buffers/ordering_mpsc.rs
akoshelev: The reason I left `take_next` was because it gives extra functionality (take N items) whereas stream must always return chunks of the same size. Given that `take_next` is quite well tested I couldn't force myself to remove it because we may need it one day maybe.

martinthomson: > stream must always return chunks of the same size

I don't think that is true.  We're returning a vec, which could be any size.

src/helpers/buffers/ordering_mpsc.rs
akoshelev: looks like an alias to me https://docs.rs/futures-util/latest/futures_util/#reexports, but it is worth importing it from `futures` crate for consistency. My Intellij makes it too easy to import without paying attention :(

src/helpers/buffers/ordering_mpsc.rs
akoshelev: That's true, the type system does not have this limitation, the implementation however does exactly that

PR #514, opened 2023-03-01
Impl Reshare for XorShare

taikiy:
Implement `Reshare` trait for semi-honest and malicious `XorShare`. For now, the malicious XOR reshare implementation is the same as semi-honest.

### DIFF
--- a/src/helpers/messaging.rs
+++ b/src/helpers/messaging.rs
@@ -8,7 +8,6 @@
 //!
 use crate::{
     bits::Serializable,
-    ff::Field,
     helpers::{
         buffers::{ReceiveBuffer, SendBuffer, SendBufferConfig},
         network::{ChannelId, MessageEnvelope, Network},
@@ -17,6 +16,7 @@ use crate::{
         Error, MessagePayload, Role, MESSAGE_PAYLOAD_SIZE_BYTES,
     },
     protocol::{RecordId, Step},
+    secret_sharing::SharedValue,
     task::JoinHandle,
     telemetry::{labels::STEP, metrics::RECORDS_SENT},
 };
@@ -40,7 +40,7 @@ use shuttle::future as tokio;
 pub trait Message: Debug + Send + Serializable + 'static {}
 
 /// Any field value can be send as a message
-impl<F: Field> Message for F {}
+impl<S: SharedValue> Message for S {}
 
 /// Entry point to the messaging layer managing communication channels for protocols and provides
 /// the ability to send and receive messages from helper peers. Protocols request communication
--- a/src/helpers/messaging.rs
+++ b/src/helpers/messaging.rs
@@ -39,7 +39,7 @@ use shuttle::future as tokio;
 /// Trait for messages sent between helpers. Everything needs to be serializable and safe to send.
 pub trait Message: Debug + Send + Serializable + 'static {}
 
-/// Any field value can be send as a message
+/// Any secret shared value can be send as a message
 impl<S: SharedValue> Message for S {}
 
 /// Entry point to the messaging layer managing communication channels for protocols and provides
--- a/src/protocol/basics/reshare.rs
+++ b/src/protocol/basics/reshare.rs
@@ -76,16 +76,13 @@ pub trait ReshareBorrowed<C: Context, B: RecordBinding> {
 /// We compute the new shares by adding/subtracting shared values and random values. Since `Fp2Array`
 /// implements `ArithmeticOps` using binary operations i.e., `Add`/`Sub` -> `BitXor`, we can use the
 /// same code for both sharing schemes.
-async fn semi_honest_reshare<'a, 'fut, V: SharedValue, S: ReplicatedSecretSharing<V>>(
+async fn semi_honest_reshare<'a, V: SharedValue, S: ReplicatedSecretSharing<V>>(
     ctx: SemiHonestContext<'a>,
     record_id: RecordId,
     to_helper: Role,
     share: &S,
     (r0, r1): (V, V),
-) -> Result<(V, V), Error>
-where
-    SemiHonestContext<'a>: 'fut,
-{
+) -> Result<S, Error> {
     let channel = ctx.mesh();
 
     // `to_helper.left` calculates part1 = (self.0 + self.1) - r1 and sends part1 to `to_helper.right`
@@ -101,7 +98,7 @@ where
             .receive(to_helper.peer(Direction::Right), record_id)
             .await?;
 
-        Ok((part1 + part2, r1))
+        Ok(S::new(part1 + part2, r1))
     } else if ctx.role() == to_helper.peer(Direction::Right) {
         // `to_helper.right` calculates part2 = (self.left() - r0) and sends it to `to_helper.left`
         // This is same as (a3 - r3) in the diagram
@@ -115,9 +112,9 @@ where
             .receive(to_helper.peer(Direction::Left), record_id)
             .await?;
 
-        Ok((r0, part1 + part2))
+        Ok(S::new(r0, part1 + part2))
     } else {
-        Ok((r0, r1))
+        Ok(S::new(r0, r1))
     }
 }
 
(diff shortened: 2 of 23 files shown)


src/secret_sharing/replicated/mod.rs
```diff
@@ -1,2 +1,9 @@
+use super::{SecretSharing, SharedValue};
+
 pub mod malicious;
 pub mod semi_honest;
+
+pub trait ReplicatedSecretSharing<V: SharedValue>: SecretSharing<V> {
+    fn left(&self) -> V;
+    fn right(&self) -> V;
+}
```
taikiy, line 9: A new trait to surface left() and right(). Used in the common reshare code.

src/helpers/messaging.rs
```diff
@@ -40,7 +40,7 @@ use shuttle::future as tokio;
 pub trait Message: Debug + Send + Serializable + 'static {}
 
 /// Any field value can be send as a message
```
benjaminsavage, line 42: need to update this comment now =)

src/protocol/basics/reshare.rs
```diff
@@ -156,6 +181,45 @@ impl<'a, F: Field> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousRepli
...
+        Ok(XorReplicated::new(s0, s1))
+    }
+}
+
+#[async_trait]
+/// Malicious version of xor resharing is executed in semi-honest context.
+///
+/// As of March 2023, we aren't sure what is a proper way of doing reshare in malicious context.
+/// We believe that, even though an additive attack is possible, a malicious helper can ONLY
+/// corrupt the protocol output, but cannot learn any private info.
```
benjaminsavage, line 207: I suggest you file an issue about this and link to it from here

src/protocol/basics/reshare.rs
```diff
@@ -280,6 +346,37 @@ mod tests {
...
+                        let record_id = RecordId::from(0);
+                        let ctx = ctx.set_total_records(1);
+
+                        // run reshare protocol for all helpers except the one that does not know the input
+                        if ctx.role() == target {
+                            // test follows the reshare protocol
+                            ctx.prss().generate_bit_arrays(record_id).into()
+                        } else {
+                            share.reshare(ctx, record_id, target).await.unwrap()
+                        }
```
benjaminsavage, line 367: Why can't you just run the reshare protocol for all? Doesn't the reshare protocol just do this same behaviour when this helper is the target?

src/protocol/context/mod.rs
```diff
@@ -89,26 +89,11 @@ mod tests {
...
-
     /// This looks weird because it uses `MaliciousReplicated::rx()` value instead of `x`.
     /// Malicious context intentionally disallows access to `x` without validating first and
     /// here it does not matter at all. It needs just some value to send (any value would do just
     /// fine)
-    impl<F: Field> AsReplicated<F> for MaliciousReplicated<F> {
+    impl<F: Field> ReplicatedSecretSharing<F> for MaliciousReplicated<F> {
         fn left(&self) -> F {
             (self as &MaliciousReplicated<F>).rx().left()
         }
```
benjaminsavage, line 99: Wait a minute...

This looks like it is ONLY accessing `rx` and never `x`...

...I suspect that this "reshare" isn't even secure in a semi-honest sense, because if the `x` value isn't getting shuffled too, then you can still trace where records went...  or am I missing something?

src/secret_sharing/replicated/semi_honest/xor_share.rs
```diff
@@ -89,6 +83,12 @@ impl<V: Fp2Array> BitXorAssign<&Self> for XorShare<V> {
     }
 }
 
+impl<V: Fp2Array> From<(V, V)> for XorShare<V> {
+    fn from(s: (V, V)) -> Self {
+        XorShare::new(s.0, s.1)
+    }
+}
+
```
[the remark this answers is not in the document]
benjaminsavage: I don't think we need this, because I don't think the test should have to call that.

src/protocol/basics/reshare.rs
```diff
@@ -280,6 +346,37 @@ mod tests {
...
 
+        /// Validates that reshare protocol actually generates new xor shares using PRSS.
+        #[tokio::test]
+        async fn generates_unique_xor_shares() {
+            let world = TestWorld::new().await;
+
+            for &target in Role::all() {
+                let secret = thread_rng().gen::<BitArray40>();
+                let shares = world
+                    .semi_honest(secret, |ctx, share: XorShare<BitArray40>| async move {
```
benjaminsavage, line 357: Please add an additional test for `.malicious' and validate that *both* parts (`x` and `rx`) have changed.

benjaminsavage (changes requested):
I think I found a serious pricavy bug with the malicious implementation.

tl;rd I think that the `left` and `right` for `MaliciousReplicated` _just_ return the `rx` values... meaning that the `x` values are not getting shuffled.

Both need to be shuffled, resulting in sending twice as many bits over the wire.

src/protocol/basics/reshare.rs
```diff
@@ -156,6 +181,45 @@ impl<'a, F: Field> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousRepli
...
+    }
+}
+
+#[async_trait]
+/// Malicious version of xor resharing is executed in semi-honest context.
+///
+/// As of March 2023, we aren't sure what is a proper way of doing reshare in malicious context.
+/// We believe that, even though an additive attack is possible, a malicious helper can ONLY
+/// corrupt the protocol output, but cannot learn any private info.
+impl<'a, F: Field, B: Fp2Array> Reshare<MaliciousContext<'a, F>, RecordId> for XorReplicated<B> {
```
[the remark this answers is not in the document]
taikiy: I was also wondering how we could fool-proof this.

There are two code locations where we call `reshare`. Let say we are in `shuffle_once` (src/protocol/sort/apply_sort/shuffle.rs@39), the function accepts a generic context `C` and an input `Vec<I>` where `I` is `Reshare`, is there a way to specialize the code when `C` is malicious and `I` is an xor share?

src/protocol/basics/reshare.rs
taikiy: I tried to factor out the common part but I realized it'd add more code than reduce them. Namely, `generate_bit_arrays` and `generate_fields` need a new function that returns `SecretSharing`. The calls to these functions are at the deepest part of the loop, and given this is a test code, I figured I would just copy them.

src/protocol/basics/reshare.rs
taikiy: We don't have `x` and `rx` for XorShare. `MaliciousContext` is generic over `F`, so there's no way to upgrade XorShare.

src/protocol/basics/reshare.rs
```diff
@@ -156,6 +181,45 @@ impl<'a, F: Field> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousRepli
...
+        &self,
+        ctx: MaliciousContext<'a, F>,
+        record_id: RecordId,
+        to_helper: Role,
+    ) -> Result<Self, Error>
+    where
+        MaliciousContext<'a, F>: 'fut,
+    {
+        self.reshare(ctx.semi_honest_context(), record_id, to_helper)
+            .await
```
martinthomson, line 214:
```suggestion
        panic!("we don't know how to do this");
```

taikiy: I assume that our position might have changed regarding this due to a possible attack mentioned by Daniel and Ben Case, but the first intention was to run reshare using semi-honest context even in the malicious settings. If we insert panic here, we won't be able to run e2e malicious IPA, and we might just as well not land this PR for now?

martinthomson: I only want to ensure that we track this issue properly.  Either we put the insecure code in and open an issue, or we ensure that we only deploy secure code.  I don't mind which.

src/protocol/basics/reshare.rs
taikiy: Opened #518

src/protocol/basics/reshare.rs
taikiy: Opened #518.

benjaminsavage: OK, I think I now know of a way how to do this. Let's land this PR with the `panic!()` call as @martinthomson has suggested, and then follow up with a diff to fix it.

benjaminsavage (approved):
I think we now know how to proceed. Let's land this and then land the fix for malicious reshare.

PR #515, opened 2023-03-02
Attribution wrapper protocol

taikiy:
Wraps the three attribution sub-protocols (accumulate, cap, aggregate) into one protocol, making it more friendly for callers that they do not need to know the details and the order of the sub-protocols.

It also lays the foundation of calling `apply_attribution_window` protocol. This wrapper protocol can determine whether to apply the attribution window logic depending on a parameter, and sub-protocols' behaviors will change accordingly. All these details will be contained in this super protocol.

### DIFF
--- a/src/helpers/transport/query.rs
+++ b/src/helpers/transport/query.rs
@@ -120,7 +120,7 @@ impl Substep for QueryType {}
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct IpaQueryConfig {
     pub per_user_credit_cap: u32,
-    pub max_breakdown_key: u128,
+    pub max_breakdown_key: u32,
     pub num_multi_bits: u32,
 }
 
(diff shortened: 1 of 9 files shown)


src/helpers/transport/query.rs
```diff
@@ -120,7 +120,7 @@ impl Substep for QueryType {}
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct IpaQueryConfig {
     pub per_user_credit_cap: u32,
-    pub max_breakdown_key: u128,
+    pub max_breakdown_key: u32,
```
benjaminsavage, line 123: I'm not sure why you changed all of these from u128 to u32. Doesn't seem related to the title and summary of this PR. What was the motivation for this?

taikiy: This is to just align with other parameter types. I initially used `u128` to make conversions between `usize` easier, but it's clearly unnecessarily large for breakdown keys.

benjaminsavage (approved):
This looks great. I like that the IPA protocol is shorter with all of this detail nicely encapsulated into this wrapper that'll be able to handle all the various input parameters (like per user cap, time windows, etc.)

I'd merge it now, but there is a conflict. Can you resolve that?

PR #519, opened 2023-03-03
make `multiply` a method

andyleiserson:
I think this is the last cleanup from https://github.com/private-attribution/ipa/pull/488#discussion_r1115059175. Reshare and reveal were done previously. This PR does multiply. The remaining things in `protocol::basics` are:

* `sum_of_products` is still an associated function, but it's less clear the right way to make that a method. If it's implemented as a method on slices, then various things that currently require `S: BasicProtocols` instead need to require `[S]: SumOfProducts`, which somewhat offsets any ergonomic benefit of being able to call as a method rather than an associated function. Sum-of-products is only used a couple places anyways.
* `share_known_value` is an associated function, but that makes sense because it's a constructor of sorts, not an operation on a value.
* `check_zero` is just a plain function, and it doesn't seem that important to make a trait out of it.

### DIFF
--- a/src/protocol/attribution/accumulate_credit.rs
+++ b/src/protocol/attribution/accumulate_credit.rs
@@ -54,7 +54,11 @@ where
     let credits = try_join_all(input.iter().skip(1).enumerate().map(|(i, x)| {
         let c = memoize_context.clone();
         let record_id = RecordId::from(i);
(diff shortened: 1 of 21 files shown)


akoshelev (approved):
lgtm

Issue #520, opened 2023-03-06
Wrapping-add attack


taikiy:
Imagine an attacker sends the following reports for the same user.

(Assume Fp31, per-user-cap = 2)

```        
                     trigger_value
         row 1   SE:      0
         row 2   TE:      2
         row 3   SE:      0
         row 4   TE:      29
         row 5   SE:      0
         row 6   TE:      2
         ...
         row N-1 SE:      0
         row N   SE:      29
```

`credit_capping` protocol computes the reversed (bottom to top) prefix-sum as follows:

```        
                     trigger_value  prefix_sum
         row 1   SE:      0              0
         row 2   TE:      2              29
         row 3   SE:      0              29
         row 4   TE:      29             0
         row 5   SE:      0              0
         row 6   TE:      2              29
         ...
         row N-1 SE:      0              29
         row N   TE:      29             0
```

Then, `is_credit_larger_than_cap` will compute credit >? cap, and yields {0, 1} bits.

```
                     trigger_value  prefix_sum  exceeds_cap_bits
         row 1   SE:      0              0             0
         row 2   TE:      2              29            1
         row 3   SE:      0              29            1
         row 4   TE:      29             0             0
         row 5   SE:      0              0             0
         row 6   TE:      2              29            1
         ...
         row N-1 SE:      0              29            1
         row N   TE:      29             0             0
```
        
Finally, `compute_final_credits` will apply the cap to the original trigger values using `exceeds_cap_bits`.
        
```
                     trigger_value  prefix_sum  exceeds_cap_bits  final_credits
         row 1   SE:      0              0             0               2
         row 2   TE:      2              29            1               0
         row 3   SE:      0              29            1               2
         row 4   TE:      29             0             0               0
         row 5   SE:      0              0             0               2
         row 6   TE:      2              29            1               0
         ...
         row N-1 SE:      0              29            1               2
         row N   TE:      29             0             0               0
```
        
In the next aggregation stage, we accumulate `final_credits`, and we'll end up with this user's contribution exceeding the cap. The reason we are not detecting the wrap-around is because we assume that the prefix-sum has never wrapped around and only check the *current* report's prefix-sum value is less than the cap.

Another case is where `exceeds_cap_bits` become all zeros. Here's an example of such a case.

```
                     trigger_value  prefix_sum  exceeds_cap_bits  final_credits
         row 1   SE:      0              0             0               29
         row 2   TE:      29             2             0               0
         row 3   SE:      0              2             0               2
[...]

PR #521, opened 2023-03-06
Prevent the wrapping-add attack

taikiy:
Preventing the wrapping-add attack - Issue #520


taikiy:
I found a case where this fix would break. We could have 4 rows:

```
SE trigger_value=0, breakdown_key=1
TE trigger_value=1, breakdown_key=0
SE trigger_value=0, breakdown_key=2
TE trigger_value=-1, breakdown_key=0
```

Then -1 will be added to breakdown_key=2 contribution. We need a different approach to solving this. Closing this PR.

PR #522, opened 2023-03-07
OrderingSender

martinthomson:
This is an alternative to the `OrderingMpsc[Sender|Receiver]`.  It's a fair bit simpler overall, as it doesn't use a channel to collect sends.

The other major feature of this is that it can send arbitrary-sized messages, just like the `UnorderedReceiver`.  That might allow us to drop some generic parameters from some of our key types.  It will also allow us to build more complex protocols (though we don't currently seem to need to).

The primary consequence of the design is that sends will block, so that back pressure is more immediate (at least in terms of the sending part; more on the task part below).  Ultimately, the promises for each send will resolve in order because each send accesses the buffer more or less directly.

The other thing to consider here is that stream data will only be available once a certain threshold is hit.  I didn't want to make that configurable, so it is capacity/2 for now.  You need to close the stream if there is less than that amount of data in it and you need to have it sent.  There might be a few places where this is not going to work out for us, so we'll need to be careful.  I think that this will affect malicious validation and it might affect places where we can't close when we are done.

One nice thing is that you can close as soon as you know how many things you are sending.  That is, if you are willing to continue polling the resulting task.  To avoid having to poll like that, it is better to send everything and then close, though you could create the close task and only poll it once the sending is complete.

---

The theory that this operates on is that sends for the same step will occur on the same thread more or less, so that contention for the mutex that guards the buffer should be lightly contested.  The atomic counter used to track the index of the next sender will further protect that mutex.  The only contention will be between sends and polling the stream.

If sends are registered with the scheduler in order, they seem to be executed in order, so sends that are executed in order won't ever have to be saved for later re-awakening until the buffer fills up.  That should keep the overheads low.  I've built a sharded collection for those wakers that we do need to save.  I don't know exactly how to shard there to be most effective, so maybe that could be simplified a little.

I'm not sure that the backpressure here is going to work without changes to the scheduler or how we use tasks.  If we schedule a ton of tasks that are blocked on the send here and separately schedule reading from the stream, we'll be saving a bunch of wakers from those sends if those are polled ahead of the read.  If we poll the stream ahead of the sending, then the same applies.  What we really want is for a finite number of sends to be scheduled.

This matters most for those places we use `[try_]join_all`, which pulls in a whole bunch of tasks and polls them all together.  Ordinary tasks that operate in series are no trouble (though they are slow and should be avoided for that reason).  The fix there is to probably to operate in a sliding window with a finite number of outstanding tasks active.  That might be the next task to look at for scaling this thing out better.

I've left the debugging in for now, which might help with understanding how this works in action.  I'm also less than 100% happy with test coverage.  Hence the draft status.


akoshelev:
> I'm not sure that the backpressure here is going to work without changes to the scheduler or how we use tasks. If we schedule a ton of tasks that are blocked on the send here and separately schedule reading from the stream, we'll be saving a bunch of wakers from those sends if those are polled ahead of the read. If we poll the stream ahead of the sending, then the same applies. What we really want is for a finite number of sends to be scheduled.

I agree with your intuition here. I've been profiling `OrderingMpscSender` this week and I am seeing this pattern. Most protocols use multiplication and it first sends then receives:

1  tlc flamegraph

If I understand correctly what is going on here, then we are getting a lot of thrashing on the send side. 

* For 1000 rows we generate 1000 sends. 16 of them (buffer capacity) get through, the remaining 984 put to sleep on notify
* 16 tasks now can execute receive. We get our first 16 multiplications done (for records 0..15)
* now we wake 984 tasks and only 16 of them can really send data again. 
* rinse and repeat

src/helpers/buffers/ordering_sender.rs
```diff
@@ -0,0 +1,566 @@
...
+            Ordering::Greater => {
+                panic!("attempt to write/close at index {i} twice");
+            }
+            Ordering::Equal => {
+                // OK, now it is our turn, so we need to hold a lock.
+                // No one else should be incrementing this atomic, so
+                // there should be no contention on this lock except for
+                // any calls to `take()`, which is fine.
+                let mut b = self.state.lock().unwrap();
+                let res = f(&mut b);
```
akoshelev, line 246: do you think it is worth dropping `b` right after this line?

src/helpers/buffers/ordering_sender.rs
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    fn write<M: Message>(&mut self, m: &M, cx: &Context<'_>) -> bool {
+        let b = &mut self.buf[self.written..];
+        if M::Size::USIZE <= b.len() {
+            self.written += M::Size::USIZE;
+            m.serialize(GenericArray::from_mut_slice(&mut b[..M::Size::USIZE]));
+
+            if self.written * 2 >= self.buf.len() {
+                Self::wake(&mut self.stream_ready);
```
akoshelev, line 70: is there an opportunity for deadlock here? if message size is not aligned with buffer capacity, for example writing 2 10 byte messages into a 16 byte buffer, send may never complete? If that's true, do we need another warning here?

src/helpers/buffers/ordering_sender.rs
```diff
@@ -0,0 +1,566 @@
...
+            }
+            true
+        } else {
+            Self::save_waker(&mut self.write_ready, cx);
```
akoshelev, line 74: maybe worth doing `debug_assert(M::Size::USIZE <= b.capacity())` to better handle obvious cases where progress cannot be made at all?

akoshelev (commented):
can't wait to integrate it with new infra, at most one extra wake per N messages sent!

PR #523, opened 2023-03-08
[IPA oneshot test] Previous test did not validate the results were same as expected.

richajaindce:
This test runs for query size = NUM_USERS * MAX_RECORDS_PER_USER as compared to earlier one which ran for a BATCHSIZE.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -3,7 +3,7 @@ use raw_ipa::{
     error::Error,
     test_fixture::{
         generate_random_user_records_in_reverse_chronological_order, test_ipa,
-        update_expected_output_for_user, TestWorld,
+        update_expected_output_for_user, IpaSecurityModel, TestWorld,
     },
 };
 
@@ -49,7 +49,7 @@ async fn main() -> Result<(), Error> {
             &expected_results,
             per_user_cap,
             MAX_BREAKDOWN_KEY,
-            true,
+            IpaSecurityModel::Malicious,
         )
         .await;
     }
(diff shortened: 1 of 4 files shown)


src/protocol/ipa/mod.rs
```diff
@@ -962,12 +833,13 @@ pub mod tests {
...
             let world = TestWorld::new().await;
 
-            test_ipa_semi_honest(
+            test_ipa(
                 world,
                 &raw_data,
                 &expected_results,
                 per_user_cap,
                 MAX_BREAKDOWN_KEY,
+                false,
```
akoshelev, line 842: it's a nit, but I really dislike boolean arguments that dictate which logic a subroutine must execute. it is very hard to read on the caller's site and understand what this `true` or `false` mean. imo it would be better if you could define an enum `IpaSecurityModel::{SemiHonest,Malicious}` (name is debatable ofc) and use it here.

akoshelev (approved):

benches/oneshot/ipa.rs
```diff
@@ -1,62 +1,57 @@
...
-                is_trigger_report: rng.gen::<u32>(),
-                breakdown_key: rng.gen_range(0..MAX_BREAKDOWN_KEY),
-                trigger_value: rng.gen_range(0..MAX_TRIGGER_VALUE),
-            };
-            (Fp32BitPrime, MatchKey, BreakdownKey)
-        ));
+    const MAX_BREAKDOWN_KEY: usize = 16;
+    const MAX_TRIGGER_VALUE: u32 = 5;
+    const NUM_USERS: usize = 8;
+    const MAX_RECORDS_PER_USER: usize = 10;
```
benjaminsavage, line 15: The randomness here will make it more difficult to benchmark IPA on N records. The total number of records generated will now be a probabilistic outcome. That's OK, but we might want to consider the usecase of "How fast is N rows?" and make some way to do that.

PR #524, opened 2023-03-08
A test coverage script

martinthomson:
This is a little bit of an indulgence for me, but I find that generating coverage reports was a little bit tedious to setup.  This script can be pointed at a file to generate a coverage report.  It's a little elaborate, but it automates the process completely, so it might be useful.

I haven't enabled cargo options (like `-F`) but those can be added.

### DIFF
--- a/scripts/test-coverage.sh
+++ b/scripts/test-coverage.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+#
+# Generate test coverage reports.
+#
+# To use, point this at the files you want a coverage report on.  This opens a
+# browser with a report or point you at the report if it can't find a browser.
+#
+# You will need a few tools installed, but this script should be able to install
+# those for you using `rustup` or `cargo`.
+
+set -e
+
+tests=()
+index="."
+for a in "$@"; do
+    if [[ -e "$a" ]]; then
+        for f in $(find "$a" -name '*.rs' -type f -exec realpath --relative-to="$(dirname "$0")/../src" {} \+); do
+            if [[ "$index" == "." ]]; then
+                index="coverage/$(realpath "$a")"
+            else
+                index="index"
+            fi
+            f="${f//\//::}"
+            tests+=("${f%.rs}")
+        done
+    fi
+done
+[[ "$index" == "." ]] && index="index"
+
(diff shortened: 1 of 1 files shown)


scripts/test-coverage.sh
```diff
@@ -0,0 +1,74 @@
...
+#
+# To use, point this at the files you want a coverage report on.  This opens a
+# browser with a report or point you at the report if it can't find a browser.
+#
+# You will need a few tools installed, but this script should be able to install
+# those for you using `rustup` or `cargo`.
+
+set -e
+
+tests=()
```
richajaindce, line 13: Can you provide an example of how to use this script and how will the output look like? Maybe can attach a screenshot with the output?

akoshelev:
would something like https://github.com/taiki-e/cargo-llvm-cov give us a better path forward? I haven't looked closely into it, but it seems to be a wrapper around llvm-cov and supports CI/CD which may be a plus for us in the future

scripts/test-coverage.sh
martinthomson: I ran:

```
$ ./scripts/test-coverage.sh src/helpers/buffers/unordered_receiver.rs
```

After building and running the tests in that file, it opened a browser to a coverage report for that file.  That shows a few assertions and even a few lines of real code that are not covered by those unit tests.

To Alex's point, `cargo llvm-cov` does much the same thing.  I've just automated it more.  I'm happy to keep this for my own use if others don't find it useful.

akoshelev (approved):

scripts/test-coverage.sh
akoshelev: If we can automate it to a point where we can see coverage on PRs, I would be happy

martinthomson:
It should be possible to do something, but it will require a bit of additional work.  An easy approach here is to run coverage for tests (using a script like this) and then append the summary to the `$GITHUB_STEP_SUMMARY` file.  That won't show up in PR discussion (that requires extra privileges), but it will show up in the run summary.

I do this for another tool, see https://github.com/martinthomson/i-d-template/issues/330#issuecomment-1132594585 for an example of that.

PR #525, opened 2023-03-09
add non-bitwise comparison vs. constant

andyleiserson:


src/protocol/boolean/comparison.rs
```diff
@@ -2,10 +2,126 @@ use super::or::or;
...
+
+impl crate::protocol::Substep for GreaterThanConstantStep {}
+
+impl AsRef<str> for GreaterThanConstantStep {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::Reveal => "reveal",
+            Self::CompareLo => "compare_lo",
+            Self::CompareHi => "compare_hi",
+            Self::Multiply => "multiply",
```
martinthomson, line 120:
```suggestion
            Self::And => "and",
```

It's a logical AND, so maybe this is clearer.

martinthomson (commented):
The less_than_constant function will need testing.  Other than that, I don't see anything to complain about here.  Nice work.

src/protocol/boolean/comparison.rs
```diff
@@ -2,10 +2,126 @@ use super::or::or;
...
+//
+/// # Errors
+/// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+/// back via the error response
+pub async fn greater_than_constant<F, C, S>(
+    ctx: C,
+    record_id: RecordId,
+    rbg: &RandomBitsGenerator<F, S, C>,
+    a: &S,
+    c: u128,
```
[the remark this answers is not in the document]
benjaminsavage: I don't think so. Below this is compared to b.as_u128(). Since "less than" and "more than" aren't really *defined* in a Ring like this - I think it's best to use a u128.

src/protocol/ipa/mod.rs
```diff
@@ -1024,11 +1024,11 @@ pub mod tests {
         const MAX_BREAKDOWN_KEY: u128 = 3;
         const NUM_MULTI_BITS: u32 = 3;
 
-        /// empirical value as of Feb 27, 2023.
-        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 17154;
+        /// empirical value as of Mar 8, 2023.
+        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 15453;
```
benjaminsavage, line 1028: 10% reduction!!! Incredible!

src/protocol/boolean/comparison.rs
```diff
@@ -2,10 +2,126 @@ use super::or::or;
...
+//  ⟺  b - c ≤ r ≤ b
+//  ⟺  b - b ≤ b - r ≤ b - (b - c)
+//  ⟺  0 ≤ a ≤ c
+//  ⟺  ~(a > c)
+//
+// Case 2: b ≤ c
+//  b < r < b + p - c
+//  ⟺  b - (b + p - c) < b - r < b - b
+//  ⟺  -(p - c) < a < 0
+//  ⟺  a > c
```
benjaminsavage, line 49: This is brilliant. Great work finding this excellent algorithm @andyleiserson! But wow... there are a lot of opportunities for an "off-by-one" error here. I tried to work the math out on paper myself, and all of the < vs <= signs, and the extra +1 terms here and there are *tricky* to get right.

I'm sure your implementation is correct, but I agree with @martinthomson it would be great to see some tests for this `greater_than_constant` function.

Man, I can't wait to merge this though!

benjaminsavage (changes requested):
Just sending back to your queue @andyleiserson for the extra tests and a few naming suggestions @martinthomson had.

src/protocol/boolean/comparison.rs
akoshelev: it would be great to add some property tests here, to make sure we cover larger inputs. Other than that, lgtm. Great work!

src/protocol/boolean/comparison.rs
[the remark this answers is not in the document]
akoshelev: in this case we would probably need to enforce `c` < `F::PRIME`?

src/protocol/boolean/comparison.rs
```diff
@@ -2,10 +2,126 @@ use super::or::or;
...
+// Goal: Compute `a > c` as `~(0 ≤ a ≤ c)`
+//
+// Strategy:
+//  1. Generate random r
+//  2. Reveal b = a + r
+//  3. Derive bounds r_low and r_high from public values, such that the desired result is a simple
+//     function of `r_low < r` and `r < r_high`.
+//
+// Case 1: b > c
+// b - c - 1 < r < b + 1
```
akoshelev, line 39: I feel dumb, but I can't figure out why we need to add 1 to $r_{high}$ and subtract 1 from $r_{low}$. I am looking at Fig 5 and the expression and it seems that it is enough for $r < p - c_2 + c$ to be the upper bound for $r$ for $a$ to be outside of ($c_1$, $c_2$). Clearly I am missing something, but can't figure out what. Can someone help me here?

andyleiserson: It is easy to get confused because of the differences between the paper and the implementation. I think the most succinct argument I have that $\pm 1$ are needed in the Fig. 5 case in the paper is by evaluation at the boundary point $r = p - c_2 + c$. At that point, $a = c - r = c - (p - c_2 + c) = -(p - c_2) \equiv c_2 \pmod p$. Since we are computing the result as $\neg(r_{low} < r < r_{high})$, and we want a negative result when $a = c_2$ (we're going for $c_1 < a < c_2$), we need to adjust $r_{high}$ by $+1$. Similar argument can be made at the low side.

Does that make sense? If not I can offer a few more ways of analyzing it. Also, while going through this, I found a mistake in the correspondence I specified between the paper variables and implementation variables, so thank you for prompting me to spend a few more minutes looking at it so I could catch that.

src/protocol/boolean/comparison.rs
andyleiserson: I didn't see how to write a proptest for async code, but with some tweaks and a small amount of duplication, I was able to proptest a non-async version.

src/protocol/boolean/comparison.rs
akoshelev: Thanks Andy for your explanation, I think I get it now. Elements in $\mathbb{Z}_p$ are discrete and paper indeed tests for inequality only, hence the shift by 1. In our case we care about equality on the lower bound only, so we keep `-1` and omit `+1`

benjaminsavage (approved):

PR #528, opened 2023-03-10
Making it work for upto 1000 records

richajaindce:


martinthomson (approved):

benches/oneshot/ipa.rs
```diff
@@ -3,32 +3,45 @@ use raw_ipa::{
...
     let random_seed = thread_rng().gen();
     println!("Using random seed: {random_seed}");
     let mut rng = StdRng::seed_from_u64(random_seed);
 
+    let mut total_count = 0;
+
     let mut random_user_records = Vec::with_capacity(NUM_USERS);
-    for _ in 0..NUM_USERS {
-        let records_for_user = generate_random_user_records_in_reverse_chronological_order(
+    while random_user_records.len() < NUM_USERS && total_count < MAX_QUERY_SIZE {
```
benjaminsavage, line 30: Maybe you don’t need both now. Perhaps just a total number of records is sufficient. The “num users” won’t necessarily even matter if you pick the wrong combination of numbers.

martinthomson:
Race condition, sorry Ben.

benches/oneshot/ipa.rs
```diff
@@ -3,32 +3,45 @@ use raw_ipa::{
...
             MAX_TRIGGER_VALUE,
         );
+        if records_for_user.len() > MAX_QUERY_SIZE - total_count {
+            records_for_user.truncate(MAX_QUERY_SIZE - total_count);
+        }
+        total_count += records_for_user.len();
         random_user_records.push(records_for_user);
     }
     let mut raw_data = random_user_records.concat();
+    println!("Running test for {:?} records", raw_data.len());
```
benjaminsavage, line 44: Can probably remove this now

PR #530, opened 2023-03-10
Run concurrency tests in release mode

akoshelev:
It leads to quite significant improvement in runtime. In debug mode it currently takes ~40 seconds on my machine to run them all. In release mode, it only takes 3 seconds

before

```bash
test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 41.59s
```

after

```bash
test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.49s
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -95,7 +95,7 @@ jobs:
         run: cargo build --features shuttle
 
       - name: Run concurrency tests
-        run: cargo test --features shuttle
+        run: cargo test --release --features shuttle
 
       - name: Run IPA bench
         run: cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches"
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         run: cargo build --features shuttle
```
martinthomson, line 95: You should build this with `--release` too, right?  I mean, we wouldn't be building or testing with debug and shuttle, but that's fine.

martinthomson (approved):

.github/workflows/check.yml
akoshelev: I tried to cheat here and use this command to cover debug builds (if shuttle test fails, it is very hard to debug shuttle tests in release build) as well. I can create another CI step to keep it clean - after taking another look, I no longer like this approach

PR #533, opened 2023-03-13
Adding an implementation of multiply for Bit Array

benjaminsavage:
See the super long comment =).

Eventually, this will be used as an "Extension Field" to be able to operate on arithmetic circuits in F2, with any security parameter we choose, as outlined in https://eprint.iacr.org/2019/386.pdf

### DIFF
--- a/src/bits/fp2_array.rs
+++ b/src/bits/fp2_array.rs
@@ -4,12 +4,12 @@ use crate::{
 };
 use bitvec::prelude::{BitArr, Lsb0};
 use generic_array::GenericArray;
-use typenum::{Unsigned, U1, U5, U8};
+use typenum::{Unsigned, U1, U4, U5};
 
 // Bit store type definitions
 type U8_1 = BitArr!(for 8, in u8, Lsb0);
+type U8_4 = BitArr!(for 32, in u8, Lsb0);
 type U8_5 = BitArr!(for 40, in u8, Lsb0);
-type U8_8 = BitArr!(for 64, in u8, Lsb0);
 
 /// The implementation below cannot be constrained without breaking Rust's
 /// macro processor.  This noop ensures that the instance of `GenericArray` used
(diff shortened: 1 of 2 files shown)


src/bits/fp2_array.rs
```diff
@@ -123,10 +125,111 @@ macro_rules! bit_array_impl {
...
+            // 5. Inverse elements: For any non-zero element a in the field, there exists a unique element -a such that a+(-a) = 0, and there exists a unique element a^(-1) such that a×a^(-1) = 1.
+            // 6. Distributivity: Multiplication distributes over addition, meaning that a(b+c) = ab+ac for any elements a, b, and c in the field.
+            //
+            // Tests of Associativity, Commutativity and Distributivity are below.
             impl std::ops::Mul for $name {
                 type Output = Self;
                 fn mul(self, rhs: Self) -> Self::Output {
-                    self & rhs
+                    debug_assert!(2 * $bits < u128::BITS);
+                    let a = <Self as Into<u128>>::into(self);
```
taikiy, line 219: Let's change this to
```suggestion
                    let a = <Self as Fp2Array>::as_u128(self);
```

src/bits/fp2_array.rs
```diff
@@ -123,10 +125,111 @@ macro_rules! bit_array_impl {
...
+                    let mut product = 0;
+                    for i in 0..Self::BITS {
+                        let bit = u128::from(rhs[i]);
+                        product ^= bit * (a << i);
+                    }
+
+                    while (u128::BITS - product.leading_zeros()) > $bits {
+                        let poly = <Self as Fp2Array>::POLYNOMIAL;
```
taikiy, line 227: Move this to the top of this function?

src/bits/fp2_array.rs
```diff
@@ -335,6 +470,32 @@ macro_rules! bit_array_impl {
...
-bit_array_impl!(bit_array_40, BitArray40, U8_5, 40, U5);
-bit_array_impl!(bit_array_8, BitArray8, U8_1, 8, U1);
+bit_array_impl!(
+    bit_array_40,
+    BitArray40,
+    U8_5,
+    40,
+    U5,
+    // x^40 + x^5 + x^3 + x^2 + 1
+    0b1_0000_0000_0000_0000_0000_0000_0000_0000_0010_1101_u128
```
taikiy, line 480: like the underscores 👍

taikiy (approved):
Just one nit suggestion. LGTM

src/bits/fp2_array.rs
benjaminsavage: Why would you suggest this? Why not the more generic Into?

src/bits/fp2_array.rs
benjaminsavage: Fair enough!

src/bits/fp2_array.rs
benjaminsavage: clippy made me do it =P

PR #535, opened 2023-03-13
Galois field 3

benjaminsavage:
Renmaing the module fp2_array to galois_field


src/ff/galois_field.rs
```diff
@@ -335,6 +470,32 @@ macro_rules! bit_array_impl {
     };
 }
 
-bit_array_impl!(bit_array_64, BitArray64, U8_8, 64, U8);
-bit_array_impl!(bit_array_40, BitArray40, U8_5, 40, U5);
-bit_array_impl!(bit_array_8, BitArray8, U8_1, 8, U1);
+bit_array_impl!(
+    bit_array_40,
+    GF_2_pow_40,
```
taikiy, line 475: Should we stick with the camel casing e.g., Gf2Pow40, or even shorter Gf40 or Gf40Bit.

taikiy:
I think we should remove Bitwise ops traits from GF, but I can do that together with merging GF with the secret sharing code.

Issue #508, opened 2023-02-27
Address in use errors


martinthomson:
```
---- net::transport::e2e_tests::fails_if_not_subscribed stdout ----
thread 'net::transport::e2e_tests::fails_if_not_subscribed' panicked at 'Failed to serve: Os { code: 98, kind: AddrInUse, message: "Address already in use" }', src/net/server/mod.rs:77:26
thread 'net::transport::e2e_tests::fails_if_not_subscribed' panicked at 'Failed to bind server to a port', src/net/server/mod.rs:95:14
```

andyleiserson:
I believe we haven't seen this in a long time, so I'm going to close this issue.

Issue #376, opened 2023-01-10
The complications of dynamic role allocation


martinthomson:
We assume that shares are allocated in a ring as follows:

* H1 gets s1, s2
* H2 gets s2, s3
* H3 gets s3, s1

However, clients and record collectors will identify helpers by their identity (for instance: A, B, C) and roles will be allocated by the leader.

If the leader allocates (A, B, C) -> (H1, H2, H3) or (B, C, A) or (C, A, B), then the order of allocations will ensure that the shares will be properly assigned.  These allocations preserve the order of helps so that A is to the left of B is to the left of C is to the left of A.

However, the other three possible allocations: (A, C, B), (B, A, C), and (C, B, A) reverse the ring order.  This works, but only if each helper knows that this is the case, so that it swaps the left and right values.

I think that we need something in the helper party network configuration that sets a canonical ring order, otherwise we'll potentially end up with a swapped order.  A failure here would either be disastrous for all of our security guarantees or would cause an immediate failure.  I'm not sure which: all of our algorithms make assumptions about how shares are replicated.

richajaindce:
There are several subprotocols in sort which make this assumption. Can we tie the helper identity from report collection and let it remain the same throughout the computation?

martinthomson:
Role allocation should be stable for an entire query - anything else is too hard to reason about.  The main point here is that we should either maintain ring order, or - if the order is reversed - also flip the two replicated shares.  I believe that is all that is necessary.

akoshelev:
good catch, I can't think of any security implications, but the whole computation will likely result in garbage output. I feel like it is probably easier to start with canonical order and avoid "illegal" allocations. I suspect that we may want to reconsider that in the future because likely it will result in poor load balancing capabilities of the helper rings, but it should be easy to tweak later.

PR #537, opened 2023-03-14
create separate server / network config structs

andyleiserson:
`NetworkConfig` contains information describing the helper network as a whole. The intent is that this can be the same on all three servers. `ServerConfig` is for a single server and has information unique to the local server (e.g. where to find private keys).

Fix use of kernel-selected free ports for tests (#508).

Removed specification of identities when creating a `Processor`. The association of helper identites 1/2/3 with discovery information is established in `NetworkConfig`.


src/net/server/mod.rs
```diff
@@ -55,6 +68,37 @@ impl MpcHelperServer {
...
+                        AddrStream,
+                        hyper::Request<hyper::Body>,
+                    >>::Service,
+                > + Clone
+                + Send
+                + Sync
+                + 'static,
+            A::Stream: AsyncRead + AsyncWrite + Unpin + Send,
+            A::Service: SendService<Request<hyper::Body>> + Send,
+            A::Future: Send,
```
andyleiserson, line 89: This arose out of an effort to reduce code duplication in the `match` statement below. Given how ugly it turned out, I'm inclined to put it back the way it was and live with the duplication of `server.handle(handle).serve(svc).await.expect("Failed to serve")`.

src/net/transport.rs
```diff
@@ -236,22 +256,24 @@ mod e2e_tests {
...
+        sockets: [TcpListener; 3],
+        server_conf: [ServerConfig; 3],
+        network_conf: Arc<NetworkConfig>,
     ) -> [Processor<Arc<HttpTransport>>; 3] {
-        join_all(ids.map(|id| {
-            let conf = Arc::clone(&conf);
-            async move {
-                let transport = HttpTransport::new(id, conf);
-                transport.bind().await;
-                Processor::new(transport, ids).await
```
andyleiserson, line 246: Note that this is not ensuring that the local helper appears first in `ids`, which was supposed to be the case, prior to the changes in this PR.

src/query/processor.rs
```diff
@@ -114,14 +116,11 @@ impl<T: Transport + Clone> Processor<T> {
         let handle = self.queries.handle(query_id);
         handle.set_state(QueryState::Preparing(req))?;
 
-        // invariant: this helper's identity must be the first element in the array.
-        let this = self.identities[0];
-        let right = self.identities[1];
-        let left = self.identities[2];
+        let id = self.identity();
+        let [right, left] = id.others();
```
andyleiserson, line 120: This relates to #376.

I added a test `helpers::tests::role_assignment_tests::multiply_with_various_roles` that is covering basic functionality in both ring orders, and it works.

I did try randomly assigning to `right` and `left` here, and it somewhat worked, but I didn't resolve all of the test failures. It's possible that the failures are just due to not being careful enough about which arrays are helper-indexed vs. role-indexed, and could be fixed easily (along the lines of the fix to `TestWorld::new_with` in this diff), but I wasn't sure if we really wanted to add random role assignments as a variable in our work right now. If there's interest in that, I'm happy to look at the rest of the test failures.

akoshelev: yea it does not seem we can handle it. It would be great to have canonical order and stick with it for some time - for 2023 it would probably be more than enough

src/test_fixture/config.rs
```diff
@@ -0,0 +1,98 @@
...
+        let mut sockets = None;
+        let ports = self.ports.unwrap_or_else(|| {
+            let socks = [(); 3].map(|_| TcpListener::bind("127.0.0.1:0").unwrap());
+            let ports = socks
+                .iter()
+                .map(|sock| sock.local_addr().unwrap().port())
+                .collect::<Vec<_>>()
+                .try_into()
+                .unwrap();
+            sockets = Some(socks);
```
akoshelev, line 43: there is still a chance for port collision here, we drop the socket at the end and kernel is free to release it between the call to `build` and `listen`. it certainly does not happen often, but it may be nice to have a generic solution, similar to https://docs.rs/axum-server/latest/axum_server/struct.Handle.html

akoshelev (approved):

akoshelev:
merging as I want to get #532 and #522 in. No critical feedback here, we can push separate PRs if we need

PR #538, opened 2023-03-14
Remove bitwise ops from Galois Field

taikiy:
This leaves GF with just arithmetic operations which will let us treat GF exactly the same as prime fields.

Next up is removing bitwise operations from `XorShare`, and merge `AdditiveShare` and `XorShare`.

### DIFF
--- a/src/ff/mod.rs
+++ b/src/ff/mod.rs
@@ -3,13 +3,16 @@
 // This is where we store arithmetic shared secret data models.
 
 mod field;
+mod galois_field;
 mod prime_field;
 
-pub use field::{BinaryField, Field, FieldType, Int};
+pub use field::{Field, FieldType, Int};
+pub use galois_field::{Gf40Bit, Gf8Bit};
 pub use prime_field::{Fp31, Fp32BitPrime};
 
 use crate::secret_sharing::SharedValue;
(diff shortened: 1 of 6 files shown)


src/ff/mod.rs
```diff
@@ -3,13 +3,16 @@
...
 
-pub use field::{BinaryField, Field, FieldType, Int};
+pub use field::{Field, FieldType, Int};
+pub use galois_field::{Gf40Bit, Gf8Bit};
 pub use prime_field::{Fp31, Fp32BitPrime};
 
 use crate::secret_sharing::SharedValue;
-use std::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+use generic_array::{ArrayLength, GenericArray};
+use std::ops::{Add, AddAssign, BitXor, BitXorAssign, Index, Mul, MulAssign, Neg, Sub, SubAssign};
```
benjaminsavage, line 15: Why do you need to import these now?

benjaminsavage (approved):

src/ff/mod.rs
taikiy: Just moving what was on line 66-67 to the top.

PR #542, opened 2023-03-15
Rename Arithmetic and Boolean secret sharing traits

taikiy:
`trait Arithmetic` -> `trait Linear`
`trait Boolean` -> `trait Bitwise`

This PR only renames all references. `Bitwise` should be bound to `Linear`, but it'll need `Mul` implementation, so I'll do that in the next diff.

### DIFF
--- a/src/secret_sharing/scheme.rs
+++ b/src/secret_sharing/scheme.rs
@@ -8,7 +8,7 @@ pub trait SecretSharing<V: SharedValue>: Clone + Debug + Sized + Send + Sync {
 }
 
 /// Secret share of a secret that has additive and multiplicative properties.
-pub trait Arithmetic<V: SharedValue>: SecretSharing<V> + ArithmeticRefOps<V> {}
+pub trait Linear<V: SharedValue>: SecretSharing<V> + ArithmeticRefOps<V> {}
 
-/// Secret share of a secret with bit operations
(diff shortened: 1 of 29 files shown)


martinthomson (approved):
Renames rot fast, but I want @benjaminsavage to take a look too.

src/secret_sharing/scheme.rs
```diff
@@ -8,7 +8,7 @@ pub trait SecretSharing<V: SharedValue>: Clone + Debug + Sized + Send + Sync {
 }
 
 /// Secret share of a secret that has additive and multiplicative properties.
-pub trait Arithmetic<V: SharedValue>: SecretSharing<V> + ArithmeticRefOps<V> {}
+pub trait Linear<V: SharedValue>: SecretSharing<V> + ArithmeticRefOps<V> {}
 
-/// Secret share of a secret with bit operations
-pub trait Boolean<V: GaloisField>: SecretSharing<V> + BooleanRefOps {}
+/// Secret share of a secret in bits. It has additive and multiplicative properties.
+pub trait Bitwise<V: GaloisField>: SecretSharing<V> + BooleanRefOps {}
```
benjaminsavage, line 14: Unclear why we need this at all now.

benjaminsavage (approved):

PR #545, opened 2023-03-16
make PrimeField and GaloisField subtypes of Field

andyleiserson:
@benjaminsavage @taikiy Prompted by the concerns in the description of #543, I did this as something of a lark to see how bad it would be and I think the conclusion is not trivial, but not too bad. There are some things that would need to be cleaned up, especially around the conversions between field types and u128.

I think with this in place, we could generalize malicious `AdditiveShare` to something like:

```
pub struct AdditiveShare<V: SharedValue, U: SharedValue + FieldExtension<V> = V> {
    x: SemiHonestAdditiveShare<V>,
    rx: SemiHonestAdditiveShare<U>,
}
```

Note the second type parameter has a default, so that `AdditiveShare<V>` as it is used today would still work and mean exactly the same thing.

That's not 100% of what we want, since the representation of `V` (`Gf1Bit`?) presumably ends up using at least one byte on the wire, so we only get 75% of the theoretical savings we're going for. For maximum savings we'd have to do something like:
```
struct PackedAdditiveShare<V: SharedValue, U: SharedValue + FieldExtension<V>> {
    x: SemiHonestAdditiveShare<V>,
    rx: Vec<SemiHonestAdditiveShare<U>>,
}
```
Here `V` would be `Gf40Bit`, `U` is `Gf32Bit` or whatever the security parameter dictates, and the `Vec` has length 40.

On the other hand, if the idea that I had tossed out the other day works, then I don't think we'd need to change malicious `AdditiveShare` at all. In fact, most of the necessary infrastructure may already be here? With these changes, reshare of malicious `AdditiveShare<Gf40Bit>` is supported.

### DIFF
--- a/src/ff/galois_field.rs
+++ b/src/ff/galois_field.rs
@@ -1,8 +1,8 @@
 use crate::{
-    ff::{GaloisField, Serializable},
+    ff::{Field, GaloisField, Serializable, Int},
     secret_sharing::SharedValue,
 };
-use bitvec::prelude::{BitArr, Lsb0};
+use bitvec::prelude::{BitArr, bitarr, Lsb0};
 use generic_array::GenericArray;
 use typenum::{Unsigned, U1, U4, U5};
 
@@ -10,6 +10,18 @@ use typenum::{Unsigned, U1, U4, U5};
 type U8_1 = BitArr!(for 8, in u8, Lsb0);
 type U8_4 = BitArr!(for 32, in u8, Lsb0);
 type U8_5 = BitArr!(for 40, in u8, Lsb0);
+
+impl Int for U8_1 {
+    const BITS: u32 = 8;
+}
+
+impl Int for U8_4 {
+    const BITS: u32 = 32;
+}
+
+impl Int for U8_5 {
+    const BITS: u32 = 40;
+}
 
 /// The implementation below cannot be constrained without breaking Rust's
 /// macro processor.  This noop ensures that the instance of `GenericArray` used
@@ -19,7 +31,7 @@ fn assert_copy<C: Copy>(c: C) -> C {
 }
 
 macro_rules! bit_array_impl {
-    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $arraylen:ty, $polynomial:expr ) => {
+    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $arraylen:ty, $one:expr, $polynomial:expr ) => {
         #[allow(clippy::suspicious_arithmetic_impl)]
         #[allow(clippy::suspicious_op_assign_impl)]
         mod $modname {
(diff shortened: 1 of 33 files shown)


benjaminsavage:
I like the direction you're going with this.

I think one possible approach we could pursue is to break this work up into a couple of smaller PRs by just:

1. Simply rename "Field" to "PrimeField"
2. Simple rename "SharedValue" to "Field"

That would make the remaining work somewhat more tractable to review. Thoughts?

src/ff/galois_field.rs
```diff
@@ -201,6 +235,7 @@ macro_rules! bit_array_impl {
                 }
             }
 
+            /*
```
benjaminsavage, line 238: Note: We probably want this back.

benjaminsavage: And maybe make an equivalent thing for the prime field.

src/secret_sharing/replicated/semi_honest/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod additive_share;
-mod xor_share;
+//mod xor_share;
 
 pub use additive_share::AdditiveShare;
-pub use xor_share::XorShare;
+//pub use xor_share::XorShare;
```
benjaminsavage, line 5: Worth cleaning up

src/ff/galois_field.rs
martinthomson: With From implemented, the TryFrom implementation isn't necessary.  We should choose (I favour the fallible TryFrom given that we need to handle overflow in most cases).

benjaminsavage: We need it for the generation from PRSS. There, we want to just go modulo the prime for PrimeField, and just take the lower 32 bits for Galois field. We were thinking that *most* cases can use the try_from approach, but at least these should truncate.

src/protocol/basics/reshare.rs
```diff
@@ -173,23 +158,7 @@ impl<'a, F: Field> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousRepli
...
-    ) -> Result<Self, Error>
-    where
-        SemiHonestContext<'a>: 'fut,
-    {
-        let r = ctx.prss().generate_bit_arrays(record_id);
-        semi_honest_reshare(ctx, record_id, to_helper, self, r).await
-    }
-}
-
+/*
```
benjaminsavage, line 161: I think it's OK to just delete this now.

src/ff/galois_field.rs
```diff
@@ -424,6 +459,7 @@ bit_array_impl!(
     U8_5,
     40,
     U5,
+    bitarr!(const u8, Lsb0; 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1),
```
taikiy, line 462: Since `Lsb0`, this should be [1, 0, ..., 0].

andyleiserson:
@taikiy took this over and merged it in several pieces; closing this one.

PR #547, opened 2023-03-16
make GaloisField a subtype of Field

taikiy:
Commandeering @andyleiserson 's #545 plus a bit more refactoring/renaming/cleanup. Thanks Andy!

### DIFF
--- a/src/ff/field.rs
+++ b/src/ff/field.rs
@@ -1,24 +1,18 @@
-use crate::secret_sharing::SharedValue;
-use generic_array::ArrayLength;
+use crate::secret_sharing::{Block, SharedValue};
 use std::fmt::Debug;
+use typenum::{U1, U4};
 
-// Trait for primitive integer types used to represent the underlying type for field values
-pub trait Int: Sized + Copy + Debug {
-    const BITS: u32;
+impl Block for u8 {
+    type Size = U1;
+    const VALID_BIT_LENGTH: u32 = u8::BITS;
 }
 
-impl Int for u8 {
-    const BITS: u32 = u8::BITS;
+impl Block for u32 {
+    type Size = U4;
+    const VALID_BIT_LENGTH: u32 = u32::BITS;
 }
 
-impl Int for u32 {
-    const BITS: u32 = u32::BITS;
-}
-
-pub trait Field: SharedValue + From<u128> + Into<Self::Integer> {
-    type Integer: Int;
-    type Size: ArrayLength<u8>;
-
+pub trait Field: SharedValue + From<u128> + Into<Self::Storage> {
     /// Multiplicative identity element
     const ONE: Self;
 
--- a/src/ff/galois_field.rs
+++ b/src/ff/galois_field.rs
@@ -1,15 +1,47 @@
 use crate::{
-    ff::{GaloisField, Serializable},
-    secret_sharing::SharedValue,
+    ff::{Field, Serializable},
+    secret_sharing::{Block, SharedValue},
 };
-use bitvec::prelude::{BitArr, Lsb0};
+use bitvec::prelude::{bitarr, BitArr, Lsb0};
 use generic_array::GenericArray;
+use std::ops::Index;
 use typenum::{Unsigned, U1, U4, U5};
+
+/// Trait for data types storing arbitrary number of bits.
+pub trait GaloisField:
+    Field + Into<u128> + Index<usize, Output = bool> + Index<u32, Output = bool>
+{
+    const POLYNOMIAL: u128;
+
+    fn as_u128(self) -> u128 {
+        <Self as Into<u128>>::into(self)
+    }
+
+    /// Truncates the higher-order bits larger than `Self::BITS`, and converts
+    /// into this data type. This conversion is lossy. Callers are encouraged
+    /// to use `try_from` if the input is not known in advance.
+    fn truncate_from<T: Into<u128>>(v: T) -> Self;
+}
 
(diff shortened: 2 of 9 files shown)


src/ff/field.rs
```diff
@@ -1,24 +1,18 @@
-use crate::secret_sharing::SharedValue;
-use generic_array::ArrayLength;
+use crate::secret_sharing::{Block, SharedValue};
 use std::fmt::Debug;
+use typenum::{U1, U4};
 
-// Trait for primitive integer types used to represent the underlying type for field values
-pub trait Int: Sized + Copy + Debug {
```
taikiy, line 6: This is replaced by `Block` in secret_sharing/mod.rs

src/ff/field.rs
```diff
@@ -1,24 +1,18 @@
...
+    type Size = U4;
+    const VALID_BIT_LENGTH: u32 = u32::BITS;
 }
 
-impl Int for u32 {
-    const BITS: u32 = u32::BITS;
-}
-
-pub trait Field: SharedValue + From<u128> + Into<Self::Integer> {
-    type Integer: Int;
```
taikiy, line 19: This is replaced by `type Storage: Block` in `SharedValue`.

src/ff/field.rs
```diff
@@ -1,24 +1,18 @@
...
-    type Size: ArrayLength<u8>;
```
taikiy, line 20: This is moved to `Block` trait.

src/ff/galois_field.rs
```diff
@@ -1,16 +1,48 @@
...
+impl Block for U8_4 {
+    type Size = U4;
+    const VALID_BIT_LENGTH: u32 = 32;
+}
+
+impl Block for U8_5 {
+    type Size = U5;
+    const VALID_BIT_LENGTH: u32 = 40;
+}
+
```
taikiy, line 45: These are equivalent to former `Int` trait implementations, i.e., storage types.

src/ff/galois_field.rs
```diff
@@ -19,7 +51,7 @@ fn assert_copy<C: Copy>(c: C) -> C {
 }
 
 macro_rules! bit_array_impl {
-    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $arraylen:ty, $polynomial:expr ) => {
```
taikiy, line 22: Now `$bits` and `$arraylen` are defined in `Block` trait and passed to this macro as `$store`.

src/ff/galois_field.rs
```diff
@@ -47,6 +88,13 @@ macro_rules! bit_array_impl {
                 }
             }
 
+            // TODO (taikiy): Remove this infallible conversion and bring back to TryFrom
+            impl From<u128> for $name {
+                fn from(v: u128) -> Self {
+                    Self::truncate_from(v)
+                }
+            }
```
taikiy, line 96: GaloisField (or BitArray) used to implement a fallible conversion of TryFrom. By making the `GaloisField` a subtype of `Field`, we need to exclusively choose either `From` or `TryFrom`. I agree with Martin that we should implement TryFrom to catch any unexpected lossy conversions, but there are literary hundreds of references to `F::from()`, so I'm going to make a separate change to remove `From` from `Field` and replace it by `TryFrom`. For now, `GaloisField` needs to follow the trait bound and implement `From` but this is temporal.

src/ff/galois_field.rs
```diff
@@ -201,22 +249,9 @@ macro_rules! bit_array_impl {
...
-                fn try_from(v: u128) -> Result<Self, Self::Error> {
-                    if u128::BITS - v.leading_zeros() <= Self::BITS {
-                        Ok(Self::truncate_from(v))
-                    } else {
-                        Err(format!(
-                            "Bit array size {} is too small to hold the value {}.",
-                            Self::BITS,
-                            v
-                        ))
-                    }
```
taikiy, line 219: As commented above, this is removed to resolve a conflict, but I will bring this back in the next PR.

src/ff/galois_field.rs
```diff
@@ -422,8 +458,7 @@ bit_array_impl!(
     bit_array_40,
     Gf40Bit,
     U8_5,
-    40,
-    U5,
+    bitarr!(const u8, Lsb0; 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
```
taikiy, line 461: Field::ONE implementation

src/ff/mod.rs
```diff
@@ -66,23 +66,6 @@ where
 {
 }
 
-/// Trait for data types storing arbitrary number of bits.
-// TODO: Implement `Message`
-pub trait GaloisField:
```
taikiy, line 71: This is moved inside galois_field.rs file.

src/ff/prime_field.rs
```diff
@@ -25,41 +28,40 @@ impl<F: PrimeField> Serializable for F {
 }
 
 macro_rules! field_impl {
-    ( $field:ident, $int:ty, $prime:expr, $arraylen:ty ) => {
```
taikiy, line 28: Same as the `GaloisField` macro. `$arraylen` is now defined in `Block` so no need to pass it here.

benjaminsavage (approved):
Lgtm!

PR #548, opened 2023-03-16
Parallel processing inside credit capping `compute_final_credits`

akoshelev:
Noticed that this method needlessly processes input rows one-by-one while working on new Infra. It is not possible to set up buffering correctly for it, so I decided to make it process inputs in parallel.

### DIFF
--- a/src/protocol/attribution/credit_capping.rs
+++ b/src/protocol/attribution/credit_capping.rs
@@ -268,7 +268,6 @@ where
 {
     let num_rows = input.len();
     let cap = T::share_known_value(&ctx, F::from(cap.into()));
-    let mut final_credits = original_credits.to_vec();
 
     // This method implements the logic below:
     //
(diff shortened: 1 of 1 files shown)


src/protocol/attribution/credit_capping.rs
```diff
@@ -286,44 +285,50 @@ where
...
-            &remaining_budget,
-            original_credit,
-        )
-        .await?;
+    try_join_all(
+        (0..num_rows)
+            .zip(repeat((ctx, cap)))
+            .map(|(i, (ctx, cap))| async move {
+                if i == num_rows - 1 {
+                    return Ok(original_credits[i].clone());
```
benjaminsavage, line 293: Could this value exceed the cap?

benjaminsavage: I guess:
1. The logic matches what was there before
2. The final row must have “original_credits” equal to zero because there is no following trigger event to attribute to this even if it is a source event.

benjaminsavage (approved):
Parallelization FTW!

src/protocol/attribution/credit_capping.rs
akoshelev: That's right, previously it was iterating over first N - 1 rows, leaving the last value inside `final_credits` intact. It may be better to explicitly set to 0 and explain why - lmk if you want me to make this change

PR #549, opened 2023-03-17
Use two channels inside random bits generator 

akoshelev:
This is just a prelude to a larger refactoring after which `RandomBitsGenerator` will be able to efficiently use the infrastructure buffering capabilities.

Currently, it does not change anything for the callers, but the idea is that `fallback_generator` will opt out from using buffering and will be continuously sending data as it goes in. With the right field size, the probability of `RandomBitsGenerator` calling it more than once is negligible, calling it once for `Fp32BitPrime` is $10^-9$

### DIFF
--- a/src/protocol/attribution/apply_attribution_window.rs
+++ b/src/protocol/attribution/apply_attribution_window.rs
@@ -139,8 +139,10 @@ where
     let cmp_ctx = c.narrow(&Step::TimeDeltaLessThanCap);
     let mul_ctx = c.narrow(&Step::CompareBitTimesTriggerValue);
 
-    let random_bits_generator =
-        RandomBitsGenerator::new(ctx.narrow(&Step::RandomBitsForBitDecomposition));
+    let random_bits_generator = RandomBitsGenerator::new(
+        ctx.narrow(&Step::RandomBitsForBitDecomposition),
+        input.len(),
+    );
     let rbg = &random_bits_generator;
 
     try_join_all(
--- a/src/protocol/boolean/random_bits_generator.rs
+++ b/src/protocol/boolean/random_bits_generator.rs
@@ -13,6 +13,8 @@ use std::{
     marker::PhantomData,
     sync::atomic::{AtomicU32, AtomicUsize, Ordering},
 };
+use crate::protocol::Substep;
+
 
 /// A struct that pre-generates and buffers random sharings of bits from the
 /// `SolvedBits` protocol. Any protocol who wish to use a random-bits can draw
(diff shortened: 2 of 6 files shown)


akoshelev:
tests are flaky because it seems we have a bug here https://github.com/private-attribution/ipa/blob/58698e35f28e8bdd41caa9ca32a53b4a6843cbb1/src/protocol/boolean/generate_random_bits/malicious.rs#L27

malicious upgrade takes a substep and narrows **the internal context**, so the step hierarchy used inside random bits is not used. Hence when fallback generator is used, malicious context uses the exact same step

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -54,12 +90,15 @@ where
     /// inner members multiple times, I/O errors while executing MPC protocols,
     /// read from an empty buffer, etc.
     pub async fn generate(&self) -> Result<RandomBitsShare<F, S>, Error> {
```
martinthomson, line 94: I think that this should retain the atomic counter, but only use it for the backup.

Then this can take an index for use with the primary one.

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -98,6 +160,18 @@ mod tests {
         let _: Fp31 = result.reconstruct(); // reconstruct() will validate the value.
     }
 
+    #[tokio::test]
+    pub async fn uses_fallback_channel() {
```
martinthomson, line 164: Not 100% on this test.  This won't be closing any of the channels.

src/protocol/boolean/random_bits_generator.rs
martinthomson, line 94: At the call site for this function, I think that we should close out the fallback generator.

martinthomson (commented):
Seems like a good direction.  I'd really prefer if we had some way to close out the fallback, but maybe we can do that once we close out the higher-layer step.

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -6,28 +6,60 @@ use crate::{
...
+#[derive(Debug)]
+struct CountingGenerator<F, S, C> {
+    counter: AtomicU32,
+    ctx: C,
     _marker: PhantomData<(F, S)>,
 }
 
+enum RBGStep {
+    /// Special communication channel used when values generated using the standard communications
+    /// failed to pass the prime test. It is widely inefficient to use, so field must be large enough
```
benjaminsavage, line 48: Wildly?

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -36,13 +68,19 @@ where
 {
     #[must_use]
     #[allow(clippy::needless_pass_by_value)] // TODO: pending resolution of TotalRecords::Indeterminate
```
benjaminsavage, line 70: Is this Clippy error suppression still necessary?

akoshelev: unfortunately yes, we still call only methods that take `&self`

src/protocol/boolean/random_bits_generator.rs
akoshelev: I don't have a better word for sending 32 bytes through and waiting for response :(

src/protocol/boolean/random_bits_generator.rs
akoshelev: given that we know how many records this generator is expected to process, we could close it inside `generate` once the last record is processed

would it be easier to work with? I've pushed this version, please take a look and let me know what you think

src/protocol/boolean/random_bits_generator.rs
akoshelev: yea I only wanted to test that uses that channel - without extra plumbing it is not easy to see which channel is being used. I also think we can test main channel closure easily by setting up a limit on number of records and then exhausting it. Not sure how we would test fallback channel closure though if it happens inside - I could expose `is_closed(&self)` method on the gateway to make it work

martinthomson (commented):
I've got some proposed changes inbound for this.  I also want to take a closer look at #550, because I think that our upgrade context design has outstayed its welcome here and we're seeing the first signs of that being a real problem here.

What I'm going to look into is making the multiplication and RNG steps for upgrade relative to the current step, so that callers just upgrade with whatever context they have.  This might take a little bit of care, but I think we'll find that it will work out well enough.

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -139,8 +139,10 @@ where
     let cmp_ctx = c.narrow(&Step::TimeDeltaLessThanCap);
     let mul_ctx = c.narrow(&Step::CompareBitTimesTriggerValue);
 
-    let random_bits_generator =
-        RandomBitsGenerator::new(ctx.narrow(&Step::RandomBitsForBitDecomposition));
+    let random_bits_generator = RandomBitsGenerator::new(
+        ctx.narrow(&Step::RandomBitsForBitDecomposition),
+        input.len(),
```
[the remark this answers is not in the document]
akoshelev: I guess I could do that too - we seem to be avoiding that assumption in many places (malicious upgrade for instance) but I don't see anything wrong with enforcing that at RBG layer

PR #550, opened 2023-03-17
MaliciousContext::narrow must narrow the inner context as well

akoshelev:
Found this issue while working on #549 - narrowing malicious context did not change the step for upgrade.

root -> substep1 -> substep2 -> upgrade
  -> another_substep1 -> another_substep2 -> upgrade

Before this change, both upgrades will use the same step `upgrade_inputs/...` and the second will fail because of record_id re-use.

I stared at this for some time and initially I did not like the idea of narrowing the internal context - it can lead to step explosion. However, `RandomBitsGenerator` after #549 needs to do exactly that - call malicious upgrade with the same record id twice (one through regular channel and sometimes again via the fallback). So I don't really have a better option now, but open for suggestions if there is a better way

### DIFF
--- a/src/protocol/context/malicious.rs
+++ b/src/protocol/context/malicious.rs
@@ -130,7 +130,13 @@ impl<'a, F: Field> Context for MaliciousContext<'a, F> {
 
     fn narrow<S: Substep + ?Sized>(&self, step: &S) -> Self {
         Self {
-            inner: Arc::clone(&self.inner),
+            // todo: it is inefficient, we only need to change the context, but end up
+            // cloning everything from inner
+            inner: ContextInner::new(
+                self.inner.upgrade_ctx.narrow(step),
+                self.inner.accumulator.clone(),
+                self.inner.r_share.clone(),
+            ),
             step: self.step.narrow(step),
             total_records: self.total_records,
         }
--- a/src/protocol/context/mod.rs
+++ b/src/protocol/context/mod.rs
@@ -69,7 +69,11 @@ mod tests {
     use crate::{
         ff::{Field, Fp31},
         helpers::Direction,
-        protocol::{malicious::Step::MaliciousProtocol, prss::SharedRandomness, RecordId},
+        protocol::{
+            malicious::{MaliciousValidator, Step::MaliciousProtocol},
+            prss::SharedRandomness,
+            RecordId,
+        },
         secret_sharing::replicated::{
             malicious::AdditiveShare as MaliciousReplicated,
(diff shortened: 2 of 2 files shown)


martinthomson (commented):
So this is unique to the changes we're making to the random bit generator, so I want to make this change along with that.

The problem is that we have a malicious context created where we upgrade the random bits from the mainline random bits generator, but if we have a fallback because the random value is too large ($\ge p$), the fallback will also request upgrade.  The method we use for upgrading bits starts from the same place, which isn't ideal.

Given the shape of the API as it stands, it might be easier to remove the upgrade context from the malicious context, so that we can just use the current context (by accessing its internals) for upgrades alongside everything else.

benjaminsavage:
Here's the PR in question: https://github.com/private-attribution/ipa/pull/557/files

Whatever solution we come up with here, I hope that it also fixes the issue I encountered there.

andyleiserson:
I'm in favor of something like #553. I don't think it makes sense to pay the cost of maintaining two parallel contexts that both need to be narrowed.

Another possibility would be to use the something in the `upgrade_with` family of functions to provide a distinct context for the fallback RBG without maintaining two parallel contexts. But that only makes sense if an entirely separate context hierarchy for malicious upgrades is important, and I don't see why it is.

akoshelev:
> Given the shape of the API as it stands, it might be easier to remove the upgrade context from the malicious context, so that we can just use the current context (by accessing its internals) for upgrades alongside everything else.

this makes sense to me and would be ideal. Over the weekend I also had a need to propagate `set_total_records` call, in order to close/flush upgraded contexts. It does not seem right that we have upgrade context isolated from the current one where upgrade is requested. 

I would also want us to decide on the strategy here - if making changes to malicious upgrade is straightforward and can be done quickly, we should abandon this PR, otherwise I would suggest to push it as it is blocking more than one working stream. 


> I'm in favor of something like #553. I don't think it makes sense to pay the cost of maintaining two parallel contexts that both need to be narrowed.

I don't see how #553 addresses that problem tbh. The issue is inside the upgrade call, I wouldn't blame RBG for it

akoshelev:
> I don't see how https://github.com/private-attribution/ipa/pull/553 addresses that problem tbh. The issue is inside the upgrade call, I wouldn't blame RBG for it

` Large diffs are not rendered by default. ` :(

benjaminsavage:
Is this issue resolved now? Do we still need this PR?

martinthomson:
Nope, it was resolved.

PR #552, opened 2023-03-17
WIP: trying to get to extension field validation

benjaminsavage:
This is a work in progress

### DIFF
--- a/src/ff/galois_field.rs
+++ b/src/ff/galois_field.rs
@@ -19,6 +19,7 @@ pub trait GaloisField:
 }
 
 // Bit store type definitions
+type U8_0 = BitArr!(for 1, in u8, Lsb0);
 type U8_1 = BitArr!(for 8, in u8, Lsb0);
 type U8_4 = BitArr!(for 32, in u8, Lsb0);
 type U8_5 = BitArr!(for 40, in u8, Lsb0);
@@ -480,4 +481,13 @@ bit_array_impl!(
     bitarr!(const u8, Lsb0; 1, 0, 0, 0, 0, 0, 0, 0),
     // x^8 + x^4 + x^3 + x + 1
     0b1_0001_1011_u128
+);
+
+bit_array_impl!(
+    bit_array_1,
+    Gf2,
+    U8_0,
+    bitarr!(const u8, Lsb0; 1, 0),
+    // x
+    0b10_u128
 );
--- a/src/protocol/basics/mul/malicious.rs
+++ b/src/protocol/basics/mul/malicious.rs
@@ -6,7 +6,7 @@ use crate::{
         context::{Context, MaliciousContext},
         RecordId,
     },
-    secret_sharing::replicated::malicious::AdditiveShare as MaliciousReplicated,
+    secret_sharing::replicated::malicious::{AdditiveShare as MaliciousReplicated, ExtendableField},
 };
 use futures::future::try_join;
 use std::fmt::Debug;
@@ -67,7 +67,7 @@ pub async fn multiply<F>(
     zeros_at: MultiplyZeroPositions,
 ) -> Result<MaliciousReplicated<F>, Error>
 where
-    F: Field,
+    F: Field + ExtendableField,
 {
     use crate::{
         protocol::context::SpecialAccessToMaliciousContext,
@@ -84,7 +84,7 @@ where
             zeros_at,
         ),
         a.rx().multiply_sparse(
-            b.x().access_without_downgrade(),
+            b.get_induced_share().access_without_downgrade(),
             duplicate_multiply_ctx.semi_honest_context(),
             record_id,
             (ZeroPositions::Pvvv, zeros_at.1),
--- a/src/protocol/basics/reshare.rs
+++ b/src/protocol/basics/reshare.rs
@@ -13,7 +13,7 @@ use crate::{
     },
     secret_sharing::{
         replicated::{
-            malicious::AdditiveShare as MaliciousReplicated,
+            malicious::{AdditiveShare as MaliciousReplicated, ExtendableField},
             semi_honest::{AdditiveShare as Replicated, XorShare as XorReplicated},
             ReplicatedSecretSharing,
         },
@@ -138,7 +138,7 @@ impl<'a, F: Field> Reshare<SemiHonestContext<'a>, RecordId> for Replicated<F> {
 /// For malicious reshare, we run semi honest reshare protocol twice, once for x and another for rx and return the results
 /// # Errors
 /// If either of reshares fails
-impl<'a, F: Field> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousReplicated<F> {
+impl<'a, F: Field + ExtendableField> Reshare<MaliciousContext<'a, F>, RecordId> for MaliciousReplicated<F> {
     async fn reshare<'fut>(
         &self,
         ctx: MaliciousContext<'a, F>,
@@ -187,27 +187,6 @@ impl<'a, B: GaloisField> Reshare<SemiHonestContext<'a>, RecordId> for XorReplica
     {
         let r = ctx.prss().generate_bit_arrays(record_id);
         semi_honest_reshare(ctx, record_id, to_helper, self, r).await
-    }
-}
-
-#[async_trait]
-/// Malicious version of xor resharing is executed in semi-honest context.
-///
-/// As of March 2023, we aren't sure what is a proper way of doing reshare in malicious context.
-/// We believe that, even though an additive attack is possible, a malicious helper can ONLY
-/// corrupt the protocol output, but cannot learn any private info.
-impl<'a, F: Field, B: GaloisField> Reshare<MaliciousContext<'a, F>, RecordId> for XorReplicated<B> {
-    #[allow(clippy::missing_panics_doc)]
-    async fn reshare<'fut>(
-        &self,
-        _ctx: MaliciousContext<'a, F>,
-        _record_id: RecordId,
-        _to_helper: Role,
-    ) -> Result<Self, Error>
-    where
-        MaliciousContext<'a, F>: 'fut,
-    {
-        panic!("we don't know how to do this");
     }
 }
 
--- a/src/protocol/basics/sum_of_product/malicious.rs
+++ b/src/protocol/basics/sum_of_product/malicious.rs
@@ -8,7 +8,7 @@ use crate::{
         RecordId,
     },
     secret_sharing::replicated::{
-        malicious::AdditiveShare as MaliciousReplicated, semi_honest::AdditiveShare as Replicated,
+        malicious::{AdditiveShare as MaliciousReplicated, ExtendableField}, semi_honest::AdditiveShare as Replicated,
         ReplicatedSecretSharing,
     },
 };
@@ -70,7 +70,7 @@ pub async fn sum_of_products<F>(
     b: &[MaliciousReplicated<F>],
 ) -> Result<MaliciousReplicated<F>, Error>
 where
-    F: Field,
+    F: Field + ExtendableField,
 {
     use crate::{
         protocol::context::SpecialAccessToMaliciousContext,
@@ -97,9 +97,10 @@ where
         let ax = a[i].x().access_without_downgrade();
         let bx = b[i].x().access_without_downgrade();
         let arx = a[i].rx();
+        let bx_induced = b[i].get_induced_share().access_without_downgrade();
         right_sops += ax.right() * bx.right() + ax.left() * bx.right() + ax.right() * bx.left();
         right_sops_m +=
-            arx.right() * bx.right() + arx.left() * bx.right() + arx.right() * bx.left();
+            arx.right() * bx_induced.right() + arx.left() * bx_induced.right() + arx.right() * bx_induced.left();
     }
 
     // notify helper on the right that we've computed our value
--- a/src/protocol/context/malicious.rs
+++ b/src/protocol/context/malicious.rs
@@ -30,8 +30,8 @@ use crate::{
     repeat64str,
     secret_sharing::{
         replicated::{
-            malicious::AdditiveShare as MaliciousReplicated,
-            semi_honest::AdditiveShare as Replicated,
+            malicious::{AdditiveShare as MaliciousReplicated, ExtendableField},
+            semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing,
         },
         Linear as LinearSecretSharing,
     },
@@ -41,7 +41,7 @@ use crate::{
 /// Represents protocol context in malicious setting, i.e. secure against one active adversary
 /// in 3 party MPC ring.
 #[derive(Clone)]
-pub struct MaliciousContext<'a, F: Field> {
+pub struct MaliciousContext<'a, F: Field + ExtendableField> {
     /// TODO (alex): Arc is required here because of the `TestWorld` structure. Real world
     /// may operate with raw references and be more efficient
     inner: Arc<ContextInner<'a, F>>,
@@ -49,12 +49,12 @@ pub struct MaliciousContext<'a, F: Field> {
     total_records: TotalRecords,
 }
 
-pub trait SpecialAccessToMaliciousContext<'a, F: Field> {
+pub trait SpecialAccessToMaliciousContext<'a, F: Field + ExtendableField> {
     fn accumulate_macs(self, record_id: RecordId, x: &MaliciousReplicated<F>);
     fn semi_honest_context(self) -> SemiHonestContext<'a>;
 }
 
-impl<'a, F: Field> MaliciousContext<'a, F> {
+impl<'a, F: Field + ExtendableField> MaliciousContext<'a, F> {
     pub(super) fn new<S: Substep + ?Sized>(
         source: &SemiHonestContext<'a>,
         malicious_step: &S,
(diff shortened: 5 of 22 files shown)


src/protocol/context/malicious.rs
```diff
@@ -111,15 +111,19 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
             .await
     }
 
+    fn get_induced_field_value(x: F) -> F {
+        x.clone()
```
benjaminsavage, line 115: wrong - haven't replaced with new thing yet.

src/protocol/basics/mul/malicious.rs
```diff
@@ -76,15 +79,20 @@ where
...
         a.x().access_without_downgrade().multiply_sparse(
-            b.x().access_without_downgrade(),
+            b_x,
             ctx.semi_honest_context(),
             record_id,
             zeros_at,
         ),
         a.rx().multiply_sparse(
-            b.x().access_without_downgrade(),
+            &b_induced_share,
```
benjaminsavage, line 96: See protocol 4.15 in https://eprint.iacr.org/2019/386.pdf.

In Step 4: "Circuit emulation:", it says:

"If Gj is a multiplication gate: Given pairs ([x], [[rb· x]]) and ([y], [[rb· y]]) on the left and right input wires
respectively, the parties compute ([x · y], [[rb· x · y]]) as follows:
(a) The parties call Fmult on [x] and [y] to receive [x · y].
(b) The parties locally compute the induced share [[y]] = f([y], 0, . . . , 0).
(c) The parties call Fbmult on [[rb· x]] and [[y]] to receive [[rb· x · y]].
"

src/protocol/context/malicious.rs
```diff
@@ -451,7 +454,10 @@ impl<'a, F: Field> ContextInner<'a, F> {
         x: Replicated<F>,
         zeros_at: ZeroPositions,
     ) -> Result<MaliciousReplicated<F>, Error> {
-        let rx = x
+        let induced_share =
+            Replicated::new(x.left().get_induced_value(), x.right().get_induced_value());
+
+        let rx = induced_share
```
benjaminsavage, line 460: In protocol 4.15 of https://eprint.iacr.org/2019/386.pdf

Step 3: Randomization of inputs: 

"For each input wire sharing [vj ] (where j ∈ {1, . . . , M}), the parties locally
compute the induced share [[vj ]] = f([vj ], 0, . . . , 0). Then, the parties call Fbmult on [[rb]] and [[vj ]] to receive
[[rb· vj ]]."

src/ff/galois_field.rs
```diff
@@ -481,3 +482,12 @@ bit_array_impl!(
     // x^8 + x^4 + x^3 + x + 1
     0b1_0001_1011_u128
 );
+
+bit_array_impl!(
+    bit_array_1,
+    Gf2,
```
benjaminsavage, line 488: A binary field. Just two elements, 0 and 1.

src/protocol/basics/sum_of_product/malicious.rs
```diff
@@ -86,20 +87,26 @@ where
...
+        let bx_induced = Replicated::new(
+            bx.left().get_induced_value(),
+            bx.right().get_induced_value(),
+        );
         right_sops += ax.right() * bx.right() + ax.left() * bx.right() + ax.right() * bx.left();
-        right_sops_m +=
-            arx.right() * bx.right() + arx.left() * bx.right() + arx.right() * bx.left();
+        right_sops_m += arx.right() * bx_induced.right()
+            + arx.left() * bx_induced.right()
+            + arx.right() * bx_induced.left();
```
benjaminsavage, line 109: Same deal as multiplication. See comment above.

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -23,16 +23,37 @@ use std::{
...
+    fn get_induced_value(&self) -> Self::LargeFieldType {
+        *self
+    }
+}
+
+impl ExtendableField for Gf2 {
+    type LargeFieldType = Gf32Bit;
+
+    fn get_induced_value(&self) -> Self::LargeFieldType {
+        Gf32Bit::try_from(self.as_u128()).unwrap()
```
benjaminsavage, line 48: @taikiy - is there a better way to do this?

taikiy: I think the current implementation is good. Maybe just move `ExtendableField` to `field.rs`? Also, line 48 can be

`Self::LargeFieldType::try_from(self.as_u128()).unwrap()`

martinthomson: Don't we have a truncate version as well?

src/ff/galois_field.rs
martinthomson: Is this Fp2 or GF(2)?

martinthomson:
> I did unfortunately have to change a lot of our tests. Any time we want a test with malicious security, we can no longer use Fp31. We need to use a field which actually implements the ExtendableField trait to extend to a real degree of statistical security. Since Fp31 does not provide this, I don't think it makes sense for it to implement that trait.

I think that we could implement this for Fp31 for use in tests only.  That is, `#[cfg(test)] impl ExtendableField for Fp31`.  It's just too useful a construct not to be able to use it in tests.  For one, having a prime with a non-trivial probability of error is very, very useful in a lot of tests.  It would be a shame not to be able to exploit that for testing purposes.

src/ff/galois_field.rs
```diff
@@ -19,6 +19,7 @@ pub trait GaloisField:
 }
 
 // Bit store type definitions
+type U8_0 = BitArr!(for 1, in u8, Lsb0);
```
taikiy, line 22: This doesn't work as intended. I created a new PR to fix the issue.

tl;dr, You can specify `U8_1` for `Gf2`.

Longer version:
`U8_0` and `U8_1` generates the same struct which is `[u8; 1]`. We need to implement the `Block` trait to let the macro know of bit-lengths and sizes of the internal storage types, but the compiler will complain if we implement one for `U8_0` because we already have the conflicting implementation for `U8_1`.

benjaminsavage: Is this resolved now?

src/secret_sharing/replicated/malicious/additive_share.rs
benjaminsavage: Since this is specific to the GF2 => Gf32Bit transition, I was thinking we could literally just clone the least significant byte of storage. How can I do that @taikiy ?

src/ff/galois_field.rs
taikiy: > Is this resolved now?
Yes, it's in the main. Please remove `U8_0` and just use the existing `U8_1`.

src/secret_sharing/replicated/malicious/additive_share.rs
taikiy: This code does exactly what you described - in a safe way via `try_from`. `truncate_from`, which is called by `try_from`, copies the least significant bits and creates a new field instance.

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -23,16 +23,37 @@ use std::{
...
-pub struct AdditiveShare<V: SharedValue> {
+pub struct AdditiveShare<V: SharedValue + ExtendableField> {
     x: SemiHonestAdditiveShare<V>,
-    rx: SemiHonestAdditiveShare<V>,
+    rx: SemiHonestAdditiveShare<V::LargeFieldType>,
 }
 
-impl<V: SharedValue> SecretSharing<V> for AdditiveShare<V> {
+pub trait ExtendableField {
+    type LargeFieldType: Field;
```
martinthomson, line 32: I think we should call this...

```suggestion
    type InducedField: Field;
```

or

```suggestion
    type ExtendedField: Field;
```

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -23,16 +23,37 @@ use std::{
...
+    fn get_induced_value(&self) -> Self::LargeFieldType;
```
martinthomson, line 33: With suggestion 1 above, you should maybe call this `to_induced()`:

```suggestion
    fn to_induced(&self) -> Self::InducedField;
```

or with suggestion 2:

```suggestion
    fn to_extended(&self) -> Self::ExtendedField;
```

Brevity, consistency and all that.  The paper tends to use "induced", but it seems like you might prefer "extended".  I don't quite know which is bette.r

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -23,16 +23,37 @@ use std::{
...
+}
+
+impl ExtendableField for Fp32BitPrime {
+    type LargeFieldType = Fp32BitPrime;
+
+    fn get_induced_value(&self) -> Self::LargeFieldType {
+        *self
+    }
+}
```
martinthomson, line 42:
```rust
impl<T> ExtendableField for T where T: PrimeField {
    type ExtendedField = Self;
    fn to_extended(&self) -> Self::ExtendedField { *self }
}
```

I know that Fp31 is too small here, but maybe we can address that in other ways, like sticking a `#[cfg(test)]` on it so that it never gets used in anger.

Otherwise, this is just another thing we need to fix if we go to a different field size.

What we might want is some way to make compilation conditional on the size of the prime, but we don't get to have that (yet).

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -23,16 +23,37 @@ use std::{
...
+
+impl ExtendableField for Gf2 {
+    type LargeFieldType = Gf32Bit;
+
+    fn get_induced_value(&self) -> Self::LargeFieldType {
+        Gf32Bit::try_from(self.as_u128()).unwrap()
+    }
+}
```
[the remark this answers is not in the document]
martinthomson: Another great comment that would be good to see in code, rather than GitHub.

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -275,6 +303,19 @@ mod tests {
...
+    // That said, it's still very useful for this to work in *tests*. One reason is that it's useful to have
+    // a small field where collissions can often happen.
+    #[cfg(test)]
+    impl ExtendableField for Fp31 {
+        type LargeFieldType = Fp31;
+
+        fn get_induced_value(&self) -> Self::LargeFieldType {
+            *self
+        }
+    }
```
martinthomson, line 317: As noted, I think that we can move this annotation to the declaration of `Fp31` itself.

martinthomson (commented):
This is much more straightforward than I expected.  A few nits for you.

src/ff/galois_field.rs
```diff
@@ -71,7 +71,8 @@ macro_rules! bit_array_impl {
                 }
 
                 fn truncate_from<T: Into<u128>>(v: T) -> Self {
-                    let v = &v.into().to_le_bytes()[..<Self as Serializable>::Size::to_usize()];
+                    const MASK: u128 = u128::MAX >> (u128::BITS - <$name>::BITS);
+                    let v = &(v.into() & MASK).to_le_bytes()[..<Self as Serializable>::Size::to_usize()];
                     Self(<$store>::new(v.try_into().unwrap()))
                 }
```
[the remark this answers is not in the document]
benjaminsavage: I'm not sure what you're suggesting... I'm sorry - I'm missing something.

src/ff/galois_field.rs
```diff
@@ -288,6 +290,7 @@ macro_rules! bit_array_impl {
                 type Output = bool;
 
                 fn index(&self, index: u32) -> &Self::Output {
+                    debug_assert!(index < <$name>::BITS);
```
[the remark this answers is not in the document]
benjaminsavage: We had to add it for Gf2. That's using an 8-bit storage primitive, so indexing didn't panic like the test asserts.

src/ff/galois_field.rs
benjaminsavage: I think those two are isomorphic, so I guess it's both. But given the way that multiplication is defined here, it sort of looks and smells more like a Galois Field.

src/protocol/basics/reshare.rs
```diff
@@ -450,6 +455,7 @@ mod tests {
                                 record_id,
                                 to_helper,
                                 Fp32BitPrime::ONE,
+                                Fp32BitPrime::ONE,
```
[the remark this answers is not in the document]
benjaminsavage: Will do!

src/protocol/boolean/comparison.rs
```diff
@@ -350,7 +351,7 @@ mod tests {
             .reconstruct();
 
         let m_result = world
-            .malicious(input.clone(), |ctx, a_share| async move {
+            .malicious(input, |ctx, a_share| async move {
```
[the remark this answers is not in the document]
benjaminsavage: I realized that close wasn't needed. Only line 345 needed the clone. This one is the final usage of `input`. Must have originally been a copy-paste error.

You're right that it's also un-necessary to clone on line 389 - I'll change that one too.

src/protocol/boolean/comparison.rs
```diff
@@ -504,6 +532,42 @@ mod tests {
         assert_eq!(zero, bitwise_lt(&world, c(Fp31::PRIME - 1), 0).await);
     }
 
+    #[tokio::test]
+    pub async fn bw_lt_fp32bit_prime() {
```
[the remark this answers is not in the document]
benjaminsavage: Good point. I just noticed we lacked test coverage for this bitwise_lt method. I'll find a way to make them share code.

src/protocol/context/malicious.rs
```diff
@@ -97,7 +98,10 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
         x: Replicated<F>,
         zeros_at: ZeroPositions,
     ) -> Result<MaliciousReplicated<F>, Error> {
-        let rx = x
+        let induced_share =
```
benjaminsavage, line 101: I'll add a comment here as well, explaining the relevant part of the paper.

src/protocol/malicious.rs
```diff
@@ -131,12 +140,15 @@ impl<F: Field> MaliciousValidatorAccumulator<F> {
...
         let random_constant = prss.generate_replicated(record_id);
-        let u_contribution = Self::compute_dot_product_contribution(&random_constant, input.rx());
-        let w_contribution = Self::compute_dot_product_contribution(
-            &random_constant,
-            input.x().access_without_downgrade(),
-        );
+        let u_contribution: F::LargeFieldType =
+            Self::compute_dot_product_contribution(&random_constant, input.rx());
+        let w_contribution: F::LargeFieldType =
+            Self::compute_dot_product_contribution(&random_constant, &induced_share);
```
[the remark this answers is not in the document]
benjaminsavage: I'll add a comment to the code saying this.

src/secret_sharing/replicated/malicious/additive_share.rs
benjaminsavage: ok, I'll go with "ExtendedField"

src/secret_sharing/replicated/malicious/additive_share.rs
benjaminsavage: OK, I'll go with "to_extended()"

src/secret_sharing/replicated/malicious/additive_share.rs
benjaminsavage: So I currently do have this configured to work for Fp31 (only in tests). I used to have that up here, but clippy was all mad, claiming that the import of Fp31 was "unused" due to it being wrapped in the #[cfg(test)]. I moved it down into the tests module to make that error go away.

What are you suggesting I change this to? Where would the #[cfg(test)] part go in the code you've shown above (defining it on PrimeField itself)?

src/secret_sharing/replicated/malicious/additive_share.rs
benjaminsavage: I'm still a bit confused about where you'd recommend I move this to. To PrimeField? In this file, or another file? And where should I put the #[cfg(test)]?

src/ff/galois_field.rs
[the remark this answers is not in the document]
taikiy: I believe what Martin is saying is that `v.into()` allocates the intermediate 16 bytes when we are converting values from `v: T` where T is Into<u128> to e.g., GF(2^8). I'm not sure how we can bind T to numeric types though. Maybe use `[num_traits](https://docs.rs/num-traits/latest/num_traits/trait.Num.html)`?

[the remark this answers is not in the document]
martinthomson: I'll elaborate...

You don't need specific implementations: this one is entirely generic and could be part of the definition of `Field`.  If the assumption is that truncation is able to drop high order bits because the caller has ensured that the value fits (I think that is the assumption) then this only requires that the implementation supports `try_from()`.

Of course, `TryFrom` for GF uses this method, a dependency order that would need to be inverted for that to work.  But the upshot would be that Field contains an implementation that is generic and specific implementations could write optimized versions.  For `Gf32Bit` that implementation could take the `u128,` truncate to `u32` and put those values in.

That leads me to ask: why are we still using `BitArray` here?  Won't this work for integers?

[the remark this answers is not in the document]
taikiy: Initially, `truncate_from()` was equivalent to `Field::from()`; an infallible conversion used to draw a random value from PRSS. Now that we have prime fields and Galois fields both sub types of `Field`, I changed `PrimeField::from()` to `PrimeField::truncate_from()` for consistency. Don't they need specific implementations? PrimeField takes modulo prime, and GaloisField just ignores high order bits.

I think `BitArray` is useful here if we ever need to create GF(2^n) if `n` is not a multiple of 8. It lets us store bits in u8's, and accessing bits is (they claim) faster.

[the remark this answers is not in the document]
martinthomson: I'm inclined to believe the claim if they are inclined to make it.  No need to change anything here for now.

martinthomson:
Regarding `Fp31,` Ben and I talked and I am suggesting that we go ahead and implement `Fp31` -> `Fp31` (even though security is laughable there).  What we should later do is `#[cfg(test)]` ALL of `Fp31` so that it never gets used in anger.  It is still extremely useful as a field in which we are able to generate bad or malicious results with reasonable probability, but only for testing.

martinthomson (approved):

PR #553, opened 2023-03-20
Cleanup RBG stuff

martinthomson:
NOTE: this is a pull request on a pull request (#549 specifically in this case).

This does a bunch of cleanup, but does not fix the underlying problem (yet).  I was able to make a commit without disabling the pre-commit hook, but only after a couple of goes, which means that this really isn't good enough yet.

### DIFF
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -418,11 +418,23 @@ impl TotalRecords {
     /// Returns true iff the total number of records is specified and the given record is the final
     /// one to process.
     #[must_use]
-    pub fn last<I: Into<RecordId>>(&self, record_id: I) -> bool {
+    pub fn is_last<I: Into<RecordId>>(&self, record_id: I) -> bool {
         match self {
             Self::Unspecified | Self::Indeterminate => false,
             Self::Specified(v) => usize::from(record_id.into()) == v.get() - 1,
         }
+    }
+
+    /// Overwrite this value.  This is only valid if either this or the new value is unspecified.
+    /// In debug builds it will panic if it isn't.
+    #[must_use]
+    pub fn overwrite<T: Into<TotalRecords>>(&self, value: T) -> TotalRecords {
+        let value = value.into();
+        debug_assert!(
+            self.is_unspecified() || value.is_indeterminate(),
+            "impermissible transition for TotalRecords {self:?} -> {value:?}",
+        );
+        value
     }
 }
 
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -398,9 +398,9 @@ pub enum TotalRecords {
     /// depends on the number of failures.
     ///
     /// The purpose of this is to waive the warning that there is a known
-    /// number of records when creating a channel. If the warning is firing
-    /// and the total number of records is knowable, prefer to specify it
-    /// rather than use this to waive the warning.
+    /// number of records when creating a channel.
+    ///
+    /// Using this is very inefficient, so avoid it.
     Indeterminate,
 }
 
@@ -425,16 +425,19 @@ impl TotalRecords {
         }
     }
 
-    /// Overwrite this value.  This is only valid if either this or the new value is unspecified.
-    /// In debug builds it will panic if it isn't.
+    /// Overwrite this value.
+    /// # Panics
+    /// This panics if the transition is invalid.
+    /// Any new value is OK if the current value is unspecified.
+    /// Otherwise the new value can be indeterminate if the old value is specified.
     #[must_use]
     pub fn overwrite<T: Into<TotalRecords>>(&self, value: T) -> TotalRecords {
-        let value = value.into();
-        debug_assert!(
-            self.is_unspecified() || value.is_indeterminate(),
-            "impermissible transition for TotalRecords {self:?} -> {value:?}",
-        );
-        value
+        match (self, value.into()) {
+            (Self::Unspecified, v) => v,
+            (_, Self::Unspecified) => panic!("TotalRecords needs a specific value for overwriting"),
+            (Self::Specified(_), Self::Indeterminate) => Self::Indeterminate,
+            (old, new) => panic!("TotalRecords bad transition: {old:?} -> {new:?}"),
+        }
     }
 }
 
--- a/src/protocol/boolean/random_bits_generator.rs
+++ b/src/protocol/boolean/random_bits_generator.rs
@@ -28,22 +28,18 @@ pub struct RandomBitsGenerator<F, S, C> {
     _marker: PhantomData<(F, S)>,
 }
 
-enum RBGStep {
-    /// Special context that is used when values generated using the standard method are larger
-    /// than the prime for the field. It is grossly inefficient to use, because communications
-    /// are unbuffered, but a prime that is close to a power of 2 helps reduce how often we need it.
-    Fallback,
-}
+/// Special context that is used when values generated using the standard method are larger
+/// than the prime for the field. It is grossly inefficient to use, because communications
+/// are unbuffered, but a prime that is close to a power of 2 helps reduce how often we need it.
+struct FallbackStep;
 
-impl AsRef<str> for RBGStep {
+impl AsRef<str> for FallbackStep {
     fn as_ref(&self) -> &str {
-        match self {
-            RBGStep::Fallback => "fallback",
-        }
+        "fallback"
     }
 }
 
-impl Substep for RBGStep {}
+impl Substep for FallbackStep {}
 
 impl<F, S, C> RandomBitsGenerator<F, S, C>
 where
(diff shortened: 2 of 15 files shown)


src/protocol/context/malicious.rs
```diff
@@ -58,17 +71,47 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
...
         Self {
-            inner: ContextInner::new(upgrade_ctx, acc, r_share),
+            inner: ContextInner::new(source, acc, r_share),
             step: source.step().narrow(malicious_step),
             total_records: TotalRecords::Unspecified,
         }
     }
 
+    /// TODO: This is not fast, but we can't just `reinterpret_cast` here.
+    fn as_semi_honest(&self) -> SemiHonestContext<'a> {
```
martinthomson, line 85: This really isn't easy to fix.  It might be that what we really need to write a sparse multiply op that is passed the gateway and PRSS directly so that we can use that for upgrading.

src/protocol/context/malicious.rs
```diff
@@ -58,17 +71,47 @@ impl<'a, F: Field> MaliciousContext<'a, F> {
...
+        let rx = x
+            .multiply_sparse(
+                &self.inner.r_share,
+                self.as_semi_honest(),
+                record_id,
+                (zeros_at, ZeroPositions::Pvvv),
+            )
+            .await?;
+        let m = MaliciousReplicated::new(x, rx);
+        let narrowed = self.narrow(&RandomnessForValidation);
```
martinthomson, line 109: I also don't want to narrow the entire context in order to get this value back, but the whole instrumentation layer makes that more awkward than I'd like.

src/protocol/context/malicious.rs
```diff
@@ -136,22 +187,18 @@ impl<'a, F: Field> Context for MaliciousContext<'a, F> {
...
     fn set_total_records<T: Into<TotalRecords>>(&self, total_records: T) -> Self {
-        debug_assert!(
-            self.is_total_records_unspecified(),
-            "attempt to set total_records more than once"
-        );
         Self {
             inner: Arc::clone(&self.inner),
             step: self.step.clone(),
-            total_records: total_records.into(),
+            total_records: self.total_records.overwrite(total_records),
```
martinthomson, line 194: This code was factored out.

src/protocol/context/malicious.rs
```diff
@@ -741,19 +669,15 @@ where
...
+        self.ctx
+            .upgrade_one(self.record_binding, input, ZeroPositions::Pvvv)
             .await
     }
 }
 
 // Impl for upgrading things that can be upgraded using a single record ID using a non-record-bound
-// context. This gets used e.g. when the protocol takes a single `Replicated<F>` input.
+// context. This is only used for tests where the protocol takes a single `Replicated<F>` input.
+#[cfg(test)]
```
martinthomson, line 680: These functions were test-only, which is good, because they are a bit fragile.  Some tests now set total records (to 1), but others don't.  Rather than try to standardize, this was easier.

src/protocol/context/malicious.rs
```diff
@@ -551,40 +505,19 @@ impl<'a, F: Field> ContextInner<'a, F> {
...
-            inner: self.inner,
+            ctx: self.ctx.narrow(step),
             record_binding: self.record_binding,
         }
     }
 }
 
-// This could also work on a record-bound context, but it's only used in one place for tests where
-// that's not currently required.
-impl<'a, F: Field> UpgradeContext<'a, F, NoRecord> {
```
martinthomson, line 571: moved below

src/protocol/context/malicious.rs
```diff
@@ -38,6 +38,19 @@ use crate::{
     sync::Arc,
 };
 
+/// This step is not used at the same place.
+/// Upgrades
```
martinthomson, line 42:
```suggestion
/// Upgrades all use this step to distinguish protocol steps from the step that is used to upgrade inputs.
```

src/protocol/boolean/random_bits_generator.rs
```diff
@@ -93,110 +68,102 @@ where
...
+                if let Some(v) = solved_bits(self.fallback_ctx.clone(), RecordId::from(i)).await? {
                     break v;
                 }
             }
         };
 
-        if self.total_records.last(record_id) {
-            self.fallback_generator.close().await;
+        if self.ctx.is_last_record(record_id) {
+            // TODO: close indeterminate channels
```
martinthomson, line 83: I'm still not sure how to address this one.

What I was working toward, which might not work out, is to have the context carry an enum:

```rust
enum RecordCloser {
  KnownRecordCount(usize),
  IndeterminateRecordCount(Step),
}
```

The `Step` type here would be a step that you use as a key into an `Arc<Mutex<HashMap<Step, Vec<GatewaySender>>>>` (or a `DashMap` along similar lines) that is owned by the context (in the `inner` part, that is).

The idea is that when you call `set_total_records(TotalRecords::Indeterminate)` a new item is added to the store and whenever a new channel is created from a context that has this `RecordCloser::IndeterminateRecordCount(s)`, a copy of the sender is saved to the identified store in the collection.

Any code that creates an indeterminate step would be responsible for closing all the senders that are gathered into this collection when they are done, using the top-level step to identify which ones need closing.

src/protocol/mod.rs
```diff
@@ -102,6 +102,23 @@ impl Step {
             id: self.id.clone() + "/" + step.as_ref(),
         }
     }
+
+    /// Determine if this step is "narrower" than the `other` step.
```
martinthomson, line 106: I don't think that this is going to be needed, assuming that the above design works.

src/protocol/boolean/random_bits_generator.rs
martinthomson: Note that I don't think we should over-index on trying to reduce contention on this store.  We expect low utilization.  The only thing we really need to do this for is `RandomBitsGenerator`, which has to be constructed just once by design.  Therefore, we can be fairly confident that adding the item to the context store is sufficiently rare and opening channels is also rare.  Starting with something crude should do, which might mean not using `DashMap` or similar.

andyleiserson (commented):
I think this strategy makes sense. The downside of this approach is that it's easier for malicious upgrades to get spread around in various places, which might increase the chance of missing a malicious upgrade. But given that upgraded values have a different type, I think that's a pretty minimal risk. And the need to do upgrades for random bits generation isn't consistent with a model of "collect all inputs and upgrade them all at once".

It might be possible to get rid of UpgradeContext entirely, but I don't think it's worth trying to do that in this PR.

src/protocol/boolean/random_bits_generator.rs
akoshelev: > The idea is that when you call set_total_records(TotalRecords::Indeterminate) a new item is added to the store and whenever a new channel is created from a context that has this RecordCloser::IndeterminateRecordCount(s), a copy of the sender is saved to the identified store in the collection.

I had the same thought, but it seemed to me that we are trying to optimize the rare case and being explicit here shouldn't be expensive. Sure, RBG is unique, let it explicitly close the fallback channel.

src/helpers/mod.rs
```diff
@@ -418,12 +418,27 @@ impl TotalRecords {
...
+    /// # Panics
+    /// This panics if the transition is invalid.
+    /// Any new value is OK if the current value is unspecified.
+    /// Otherwise the new value can be indeterminate if the old value is specified.
+    #[must_use]
+    pub fn overwrite<T: Into<TotalRecords>>(&self, value: T) -> TotalRecords {
+        match (self, value.into()) {
+            (Self::Unspecified, v) => v,
+            (_, Self::Unspecified) => panic!("TotalRecords needs a specific value for overwriting"),
+            (Self::Specified(_), Self::Indeterminate) => Self::Indeterminate,
```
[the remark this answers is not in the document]
martinthomson: So this reflects where the code currently is.  That is, we set a record count at a high level, which flows down into RBG and so we need to go from specific to indeterminate for the fallback.  We could make a special API for making the limit indeterminate, but I think that we understand from seeing `TotalRecords::Indeterminate` in code that this is not a good thing, so I'm not concerned about that.  I'm a little concerned about specific->specific transitions that appear without care, but apparently we are good so far.

src/protocol/boolean/random_bits_generator.rs
martinthomson: > we seem to require it in more than one protocol and the current design requires callers to provide record_id, which means we need an instance per step?

You currently need a single RBG for all records on a step, though you can create a different RBG for a different step.

martinthomson: > RBG is unique, let it explicitly close the fallback channel.

So that is hard.  If you look at RBG, it spawns multiple steps, even for fallback.  So what we need is to close all of those channels.  Hence the suggested design.  As noted, this should be rare, so we don't need lots of machinery, or very fast code, but it's a little more complex than just closing a single channel.

PR #554, opened 2023-03-20
Remove XorShare

taikiy:
XorShare's underlying secret share value type `Fp2Array` was changed to `GaloisField`. Now that we can treat both prime fields and galois fields the same using `AdditiveShare`, we don't need `XorShare` anymore.


benjaminsavage (approved):
I love it. Thank you @taikiy!

PR #555, opened 2023-03-20
Hanging tests

martinthomson:
This builds on both #553 and @akoshelev's `ordering-sender-new-infra` branch.  I wasn't able to figure out why things were stalling here though.


src/helpers/gateway/send.rs
```diff
@@ -55,47 +97,64 @@ impl<M: Message> SendingEnd<M> {
...
+            let sender = Arc::new(GatewaySender::new(
+                channel_id.clone(),
+                OrderingSender::new(write_size, SPARE.unwrap()),
+                total_records,
+            ));
+            if senders
+                .insert(channel_id.clone(), Arc::clone(&sender))
+                .is_some()
+            {
+                panic!("TODO - make sender creation contention less dangerous");
```
martinthomson, line 144: I want to flag this as a potentially very bad bug.  I've put a `panic!()` here, but it might need some more careful thought instead.

src/helpers/gateway/send.rs
```diff
@@ -55,47 +97,64 @@ impl<M: Message> SendingEnd<M> {
...
+            }
+            let stream = GatewaySendStream {
+                inner: Arc::clone(&sender),
+            };
+            (sender, Some(stream))
         }
     }
 }
+
+impl Stream for GatewaySendStream {
```
martinthomson, line 154: I don't think that we need this wrapper.

akoshelev:
I think we should merge it - all tests and benchmark pass. This PR brings `OrderingSender`, channel closure and lots of other good stuff. We can build `seq_join_all` on top of it

akoshelev (approved):

martinthomson:
Merging a PR entitled "hanging tests" makes me happy for some reason...

PR #558, opened 2023-03-21
Replace most try_join_all calls with seq_try_join_all

martinthomson:
I tried to go all-in with replacing `try_join_all` and discovered that things don't compile.

I backed out many of the changes until I'm down to a single compilation error, which is spectacularly unhelpful.

Probably the most informative change is the one in `xor_all_the_bits()`.  That generates a new failure if you replace `try_join_all` with `seq_try_join_all`.

All the reading I've done suggests that the compilation error comes down to various things not having `Send` when they need to.  I marked as many things `Send` as I dared, but made no real progress.  My current theory is that the difference between `try_join_all` and `seq_try_join_all` is that the former collects all of the futures before it ever awaits, whereas the new implementation carries the input iterator or stream across.  Something about that causes the compiler to complain.

<details>
<summary>Error messages</summary>

```
error: higher-ranked lifetime error
   --> src/test_fixture/ipa.rs:126:44
    |
126 |               IpaSecurityModel::Malicious => world
    |  ____________________________________________^
127 | |                 .semi_honest(records, |ctx, input_rows| async move {
128 | |                     ipa_malicious::<Fp32BitPrime, MatchKey, BreakdownKey>(
129 | |                         ctx,
...   |
136 | |                     .unwrap()
137 | |                 })
    | |__________________^
    |
    = note: could not prove `[async block@src/test_fixture/ipa.rs:127:57: 137:18]: std::marker::Send`

error: higher-ranked lifetime error
   --> src/test_fixture/ipa.rs:140:45
    |
140 |               IpaSecurityModel::SemiHonest => world
    |  _____________________________________________^
141 | |                 .semi_honest(records, |ctx, input_rows| async move {
142 | |                     ipa::<Fp32BitPrime, MatchKey, BreakdownKey>(
143 | |                         ctx,
...   |
150 | |                     .unwrap()
151 | |                 })
    | |__________________^
    |
    = note: could not prove `[async block@src/test_fixture/ipa.rs:141:57: 151:18]: std::marker::Send`

error: implementation of `std::marker::Send` is not general enough
   --> src/test_fixture/ipa.rs:140:45
    |
140 |               IpaSecurityModel::SemiHonest => world
    |  _____________________________________________^
141 | |                 .semi_honest(records, |ctx, input_rows| async move {
142 | |                     ipa::<Fp32BitPrime, MatchKey, BreakdownKey>(
143 | |                         ctx,
...   |
150 | |                     .unwrap()
151 | |                 })
    | |__________________^ implementation of `std::marker::Send` is not general enough
    |
    = note: `std::marker::Send` would have to be implemented for the type `&'0 secret_sharing::replicated::semi_honest::additive_share::AdditiveShare<ff::prime_field::fp32bit::Fp32BitPrime>`, for any lifetime `'0`...
    = note: ...but `std::marker::Send` is actually implemented for the type `&'1 secret_sharing::replicated::semi_honest::additive_share::AdditiveShare<ff::prime_field::fp32bit::Fp32BitPrime>`, for some specific lifetime `'1`
```

</details>


martinthomson:
https://users.rust-lang.org/t/implementation-is-not-general-enough/57433 might contain clues to sorting out the problems that I'm seeing.

src/seq_futures.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        if let Some(f) = this.active.front_mut() {
+            if let Poll::Ready(v) = Future::poll(Pin::as_mut(f), cx) {
+                drop(this.active.pop_front());
+                Poll::Ready(Some(v))
+            } else {
+                for f in this.active.iter_mut().skip(1) {
+                    let res = Future::poll(Pin::as_mut(f), cx);
+                    assert!(res.is_pending(), "future resolved out of order");
```
andyleiserson, line 147: I think there may be a race that will cause a false positive here, if the first future has not resolved when it is checked on line 141, and then multiple futures resolve in quick succession.

martinthomson:
Thanks for looking into this Andy, this is really encouraging to see (and I'm glad that you somehow managed to convince this to do something approximately sensible.  I really want to test out what you have built here, but time has not been my friend this week.

I have a few comments, but I do want to point out that I'm unable to do a lot of work on this in the next week because I'll be somewhat distracted by the IETF meeting.  If either you or @akoshelev want to keep running with it, then please do.

I had hoped to take a `Stream` as input here because that is the shape of the output and it might be nice if we could chain one stream into the next.  However, I'll take the wins and if we have to use `Iterator`, so be it.

The main thing I was concerned about from your comment...

> I changed SequentialFutures to eagerly collect the futures.

Turned out to be not true, thankfully.  One thing I really want to avoid is drawing down on the input (the `Iterator` in this case, the `Stream` in my original vision) any more than `active` times.  You kept that part of my code, which is great.

I think that the Pin is not quite right.  The values inside the `VecDeque` do need to be pinned, but they are already.  The `VecDeque` itself doesn't because references to that don't leak.  The same for the iterator.  Thus, those values don't need to be pinned (I think).  

That Rust is unable to statically analyze these properties (we're hiding behind `unsafe` blocks inside the `pin_project` implementation), is another black mark for async Rust in my view.  The first being this whole business with lifetimes and how that manifest in terms of what the compiler said about this code.  We're essentially writing unsafe code here, without safeguards and that is pretty distressing.

Separately, for the compiler/borrow checker bug, I'm surprised that you only needed `assert_send()` in that one file.

As for the race condition, if there are cases for inputs `[a, b]` where `a` resolves after polling such that polling `b` produces a result, then that would make those functions unsuitable for use here.  For our purposes, we guarantee sequencing via both sender and receiver now, which are where our `await` calls sit.  I'm fairly confident then that we won't have bad sequencing in our protocols as a result.  After all, `a` should not make process without being polled, so if resolution of `b` depends on `a` being resolved, we're fine.  Otherwise, we're in for a world of hurt.

A final thought.  I hadn't gotten around to this, but our send buffer will determine the size of the `active` set here.  I've hard-coded that, which is probably unwise, because if the send buffer is too large we'll deadlock.  Basically, the current active set is 256 items, but the send buffer won't flush until it hits 4 kilobytes.  For our 4 byte fields, we need 1024 items in the active set before we'll fill the send buffers and release data.

I know that this makes the system that much more fragile, but such is the cost of performance sometimes.  I don't know how to manifest that precondition in code because it crosses so many layers.  If you have any ideas, I'd really like to hear them.  We can maybe add runtime checks, passing stuff around using task-local state.  But I don't know if that is safe because the cost of checking would be high enough that we'd want to turn it off in release mode builds, where we might be inclined to use different configuration.  Anyway, more thought needed, I think.

PR #561, opened 2023-03-21
[Oneshot IPA] fix test to honor QUERY_SIZE

richajaindce:
remove NUM_USERS requirement from oneshot so that QUERY_SIZE is honored while generating input for the test


benches/oneshot/ipa.rs
```diff
@@ -28,21 +27,20 @@ async fn main() -> Result<(), Error> {
     let mut total_count = 0;
 
     let mut random_user_records = Vec::with_capacity(NUM_USERS);
```
martinthomson, line 29: This is your only use for NUM_USERS now.

I would suggest that you calculate this value.  Setting it to QUERY_SIZE / Expected(records_per_user) is probably better than keeping a constant that will rot.

Also, I think we should change the distribution from which we draw the number of records for each user.  Uniform is not that realistic.

richajaindce: we actually don't draw from a fixed number of users. Inside the function, it generates a random number upto MAX_RECORDS_PER_USER and uses that to decide the count.

benches/oneshot/ipa.rs
```diff
@@ -13,36 +13,33 @@ use std::time::Instant;
...
             &mut rng,
             MAX_RECORDS_PER_USER,
             MAX_BREAKDOWN_KEY,
             MAX_TRIGGER_VALUE,
         );
-        if records_for_user.len() > MAX_QUERY_SIZE - total_count {
-            records_for_user.truncate(MAX_QUERY_SIZE - total_count);
+        if records_for_user.len() > QUERY_SIZE - total_count {
+            records_for_user.truncate(QUERY_SIZE - total_count);
         }
```
martinthomson, line 38: You can just truncate unconditionally.

```suggestion
        records_for_user.truncate(QUERY_SIZE - total_count);
```

martinthomson (approved):
LGTM.  Check with @akoshelev about #562, but this one looks more ready to me.

benjaminsavage (approved):

PR #562, opened 2023-03-22
Fix send gateway buffer capacity

akoshelev:
As Martin predicted, send capacity was destined to be broken because of the bad config name. So we ended up with send buffer capacity that is 4 times smaller than we wanted.

I thought it might be a good idea to align buffer capacity with the field size as it should work for the majority of the channels. That required making sure IPA uses the same field, so I had to plug in a generic argument to it

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -1,6 +1,7 @@
 use rand::{rngs::StdRng, thread_rng, Rng, SeedableRng};
 use raw_ipa::{
     error::Error,
+    ff::Fp32BitPrime,
     helpers::GatewayConfig,
     test_fixture::{
         generate_random_user_records_in_reverse_chronological_order, test_ipa,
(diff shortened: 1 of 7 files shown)


benches/oneshot/ipa.rs
```diff
@@ -16,10 +17,12 @@ async fn main() -> Result<(), Error> {
     const MAX_QUERY_SIZE: usize = 1000;
     const NUM_USERS: usize = 300;
     const MAX_RECORDS_PER_USER: usize = 10;
+    type BenchField = Fp32BitPrime;
 
     let mut config = TestWorldConfig::default();
-    config.gateway_config =
-        GatewayConfig::symmetric_buffers((NUM_USERS * MAX_RECORDS_PER_USER).clamp(16, 1024));
+    config.gateway_config = GatewayConfig::symmetric_buffers::<BenchField>(
+        (NUM_USERS * MAX_RECORDS_PER_USER).clamp(16, 1024),
```
martinthomson, line 24: @richajaindce made a change elsewhere to remove the variability in this test.  Look at #561.

What happened to #459, by the way?  I kinda liked that, but I guess got too ambitious.

martinthomson (approved):
My predictions always come true.  A few interactions here that you might want to think about before landing though.

benches/oneshot/ipa.rs
akoshelev: yea I think #459 tried to bite more than I could swallow. We probably need something simpler that just adds parameters to this IPA benchmark. Naturally we could see where it is going and course-correct if necessary

PR #566, opened 2023-03-22
WIP: using field extension to drop IPA communications

benjaminsavage:
This PR is stacked on top of #552 - and makes use of that new functionality to save communication overhead in IPA.

This uses a *binary field* for the bits of the match key, and while in malicious mode, the `rx` component is still a 4-byte value, the `x` part is just a single bit. For the moment, this results in 1-byte of communication - but we can address that later.

### DIFF
--- a/src/protocol/attribution/malicious.rs
+++ b/src/protocol/attribution/malicious.rs
@@ -6,12 +6,13 @@ use super::{
 };
 use crate::{
     error::Error,
-    ff::{GaloisField, PrimeField, Serializable},
+    ff::{GaloisField, Gf2, PrimeField, Serializable},
     protocol::{
-        boolean::bitwise_equal::bitwise_equal,
+        boolean::bitwise_equal::bitwise_equal_gf2,
         context::{Context, SemiHonestContext},
         ipa::IPAModulusConvertedInputRow,
         malicious::MaliciousValidator,
+        modulus_conversion::{convert_bit, convert_bit_local},
         RecordId, Substep,
     },
     secret_sharing::replicated::{
@@ -29,6 +30,8 @@ use std::iter::{repeat, zip};
 pub async fn secure_attribution<'a, F, BK>(
     sh_ctx: SemiHonestContext<'a>,
     malicious_validator: MaliciousValidator<'a, F>,
+    binary_malicious_validator: MaliciousValidator<'a, Gf2>,
+    sorted_match_keys: Vec<Vec<AdditiveShare<Gf2>>>,
     sorted_rows: Vec<IPAModulusConvertedInputRow<F, AdditiveShare<F>>>,
     per_user_credit_cap: u32,
     max_breakdown_key: u32,
@@ -42,24 +45,42 @@ where
     SemiHonestAdditiveShare<F>: Serializable,
 {
     let m_ctx = malicious_validator.context();
+    let m_binary_ctx = binary_malicious_validator.context();
 
     let futures = zip(
         repeat(
-            m_ctx
+            m_binary_ctx
                 .narrow(&Step::ComputeHelperBits)
-                .set_total_records(sorted_rows.len() - 1),
+                .set_total_records(sorted_match_keys.len() - 1),
         ),
-        sorted_rows.iter(),
+        sorted_match_keys.iter(),
     )
-    .zip(sorted_rows.iter().skip(1))
+    .zip(sorted_match_keys.iter().skip(1))
     .enumerate()
-    .map(|(i, ((m_ctx, row), next_row))| {
+    .map(|(i, ((c, row), next_row))| {
         let record_id = RecordId::from(i);
-        async move { bitwise_equal(m_ctx, record_id, &row.mk_shares, &next_row.mk_shares).await }
+        async move { bitwise_equal_gf2(c, record_id, &row, &next_row).await }
     });
+    let helper_bits_gf2 = try_join_all(futures).await?;
+
+    let validated_helper_bits_gf2 = binary_malicious_validator.validate(helper_bits_gf2).await?;
+
+    let hb_mod_conv_ctx = sh_ctx
+        .narrow(&Step::ModConvHelperBits)
+        .set_total_records(sorted_match_keys.len() - 1);
+    let semi_honest_helper_bits = try_join_all(validated_helper_bits_gf2.iter().enumerate().map(
+        |(i, gf2_bit)| {
+            let bit_triple = convert_bit_local::<F, Gf2>(sh_ctx.role(), 0, gf2_bit);
+            let record_id = RecordId::from(i);
+            let c = hb_mod_conv_ctx.clone();
+            async move { convert_bit(c, record_id, &bit_triple).await }
+        },
+    ))
+    .await?;
+
     let helper_bits = Some(AdditiveShare::ZERO)
         .into_iter()
-        .chain(try_join_all(futures).await?);
+        .chain(m_ctx.upgrade(semi_honest_helper_bits).await?);
 
     let attribution_input_rows = zip(sorted_rows, helper_bits)
         .map(|(row, hb)| {
@@ -105,6 +126,7 @@ pub enum Step {
     AccumulateCredit,
     PerformUserCapping,
     AggregateCredit,
+    ModConvHelperBits,
 }
 
 impl Substep for Step {}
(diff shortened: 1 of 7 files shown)


src/protocol/boolean/bitwise_equal.rs
```diff
@@ -46,6 +46,26 @@ where
...
+    S: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    debug_assert!(a.len() == b.len());
+    let c = a
+        .iter()
+        .zip(b.iter())
+        .map(|(a_bit, b_bit)| a_bit.clone() - b_bit)
+        .collect::<Vec<_>>();
+
+    no_ones(ctx, record_id, &c).await
```
martinthomson, line 66: No one's which exactly?

These functions should probably be named `all_` or `any_`.  `no_` was confusing.

src/protocol/ipa/mod.rs
```diff
@@ -223,11 +221,11 @@ where
...
-        let f_mk_shares = self.mk_shares.reshare(
-            ctx.narrow(&IPAInputRowResharableStep::MatchKeyShares),
-            record_id,
-            to_helper,
-        );
+        // let f_mk_shares = self.mk_shares.reshare(
+        //     ctx.narrow(&IPAInputRowResharableStep::MatchKeyShares),
+        //     record_id,
+        //     to_helper,
+        // );
```
martinthomson, line 228: ...

src/protocol/ipa/mod.rs
```diff
@@ -313,20 +306,31 @@ where
...
+                    Replicated::new(
+                        Gf2::truncate_from(row.mk_shares.left()[i]),
+                        Gf2::truncate_from(row.mk_shares.right()[i]),
+                    )
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>();
+
+    let combined_match_keys_and_sidecar_data = converted_bk_shares
```
martinthomson, line 323: The name of this is no longer correct.

src/protocol/ipa/mod.rs
```diff
@@ -401,7 +414,29 @@ where
         MaliciousValidator::<F>::new(sh_ctx.narrow(&Step::AfterConvertAllBits));
     let m_ctx = malicious_validator.context();
 
-    let converted_mk_shares = combine_slices(&converted_mk_shares, MK::BITS);
+    //
+    //New stuff
```
martinthomson, line 418: "New"?

src/protocol/boolean/bitwise_equal.rs
```diff
@@ -46,6 +46,26 @@ where
...
+    b: &[S],
+) -> Result<S, Error>
+where
+    C: Context,
+    S: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    debug_assert!(a.len() == b.len());
+    let c = a
+        .iter()
+        .zip(b.iter())
```
martinthomson, line 62:
```suggestion
    let c = zip(a.iter(), b.iter())
```
Might be easier to read.

src/protocol/attribution/malicious.rs
```diff
@@ -41,24 +44,42 @@ where
...
+        .set_total_records(sorted_match_keys.len() - 1);
+    let semi_honest_helper_bits = try_join_all(validated_helper_bits_gf2.iter().enumerate().map(
+        |(i, gf2_bit)| {
+            let bit_triple = convert_bit_local::<F, Gf2>(sh_ctx.role(), 0, gf2_bit);
+            let record_id = RecordId::from(i);
+            let c = hb_mod_conv_ctx.clone();
+            async move { convert_bit(c, record_id, &bit_triple).await }
+        },
+    ))
+    .await?;
```
martinthomson, line 78: This function is kinda long and awkward now.  Can you think of a way to label a chunk of this early stuff and put it in a function so that someone reading this function doesn't have to think about the details so much?

Calculating the helper bits is taking up a lot of space, so maybe just take all of that together.

martinthomson (commented):
The gains look pretty solid.  I've not followed the GF conversion closely, but this looks like it might help.  I've spoken to @akoshelev about improving our metrics so that we count bytes as well as messages, but we should look into that.

Would it help us to have a 31-bit GF polynomial?  Do we ever send both pieces at once?  Or should we be looking to improve the sending infrastructure to do bit packing (somehow).

src/protocol/boolean/bitwise_equal.rs
benjaminsavage: "no ones" makes sure a list of shares is "all_zeroes". I could call it that instead.

src/protocol/attribution/mod.rs
```diff
@@ -231,12 +239,66 @@ where
...
+    S: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    let narrowed_ctx = ctx
+        .narrow(&Step::ComputeHelperBits)
+        .set_total_records(sorted_match_keys.len() - 1);
+
+    try_join_all(
+        sorted_match_keys
+            .iter()
+            .zip(sorted_match_keys.iter().skip(1))
```
taikiy, line 257: I found a cool technique for this where we want to have a "sliding window".

```
sorted_match_keys.windows(2).enumerate().map(|(i, rows)| {
    ...
    async move { bitwise_equal_gf2(c, record_id, rows[0], rows[1]).await }
})
```

src/protocol/ipa/mod.rs
```diff
@@ -810,17 +842,17 @@ pub mod tests {
...
 
-        /// empirical value as of Feb 27, 2023.
-        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 11784;
+        /// empirical value as of Mar 24, 2023.
+        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 10848;
 
-        /// empirical value as of Feb 28, 2023.
-        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 29046;
+        /// empirical value as of Mar 24, 2023.
+        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 27189;
```
taikiy, line 855: ❤

src/protocol/attribution/mod.rs
martinthomson: I think that you need to do this:

```rust
sorted_match_keys/*.iter() ?*/.windows(2).enumerate().map(|(i, [first, second])| {
  async move { bitwise_equal_gf2(c, RecordId::from(i), second, first).await }
```

The drawback here is that you need a slice.  This won't work otherwise, though with the double-iteration requirement, that's maybe inevitable.

A specific adapter might be even better.

```rust
let mut it = sorted_match_keys.into_iter(); // See note.
let mut prev = it.next().ok_or(Error::NoDataWtf)?;
it.map(|v| { let r = (prev, v); prev = v; r }).enumerate(|i, (first, second)| {
  // ... as above ...
});
```

Note that I'm using `into_iter()` here and assuming that the values are `Copy`.  That's not ideal (or even true), but it simplifies things.  Otherwise, you need to clone every item in the list.

You could maybe do this with references and avoid the clone, but it might be a little fiddly.

martinthomson: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fad373f3c12798257189c36d21eb9733

taikiy: I pulled the PR and tried the following. Seems to work okay :)

```
    try_join_all(sorted_match_keys.windows(2).enumerate().map(|(i, rows)| {
        let c = narrowed_ctx.clone();
        let record_id = RecordId::from(i);
        async move { bitwise_equal_gf2(c, record_id, &rows[0], &rows[1]).await }
    }))
    .await
```

There is `array_windows` that yields an iterator like in Martin's suggestion, but I think we are okay with just `windows()`.

martinthomson: That doesn't work generically though, sadly.  Try this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=99a63b7f87733b461183b8454bfffc83

martinthomson: Homework for someone who wants a challenge: implement this for pairs that aren't adjacent so that we can to the GraphSC thing over a small window.  That won't help for capping (which has to traverse the entire thing, such that a slice is necessary), but it might be OK for some applications.

benjaminsavage: That is so cool @taikiy! Thanks for this code snippet, I've added it in!

PR #567, opened 2023-03-23
Pass timestamps to IPA

taikiy:
Add a new `AdditiveShare<F>` field for timestamps to IPA e2e protocol input.

semi-honest cap3: 15453 -> 15507
malicious cap3: 38400 -> 38535
semi-honest cap1: 11784 -> 11838
malicious cap1: 29046 -> 29181

We can save the cost by not applying permutation if the attribution window parameter passed to IPA is 0 (disable attribution window). But that requires different input structures and many if-else in several code paths. If we expect reasonable amount of queries to disable applying attribution window, then we should consider doing so. However, my guess is that almost all queries would require attribution window computations.

### DIFF
--- a/src/query/executor.rs
+++ b/src/query/executor.rs
@@ -235,11 +235,11 @@ mod tests {
 
         let records: Vec<GenericReportTestInput<Fp31, MatchKey, BreakdownKey>> = ipa_test_input!(
             [
-                { match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
-                { match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
-                { match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
-                { match_key: 12345, is_trigger_report: 1, breakdown_key: 0, trigger_value: 5 },
-                { match_key: 68362, is_trigger_report: 1, breakdown_key: 0, trigger_value: 2 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
+                { timestamp: 0, match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 1, breakdown_key: 0, trigger_value: 5 },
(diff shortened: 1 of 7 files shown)


src/query/executor.rs
```diff
@@ -235,11 +235,11 @@ mod tests {
...
-                { match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
-                { match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
-                { match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
-                { match_key: 12345, is_trigger_report: 1, breakdown_key: 0, trigger_value: 5 },
-                { match_key: 68362, is_trigger_report: 1, breakdown_key: 0, trigger_value: 2 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
+                { timestamp: 0, match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 0, match_key: 12345, is_trigger_report: 1, breakdown_key: 0, trigger_value: 5 },
+                { timestamp: 0, match_key: 68362, is_trigger_report: 1, breakdown_key: 0, trigger_value: 2 },
```
taikiy, line 242: We are not using timestamps yet, so just fill with 0's. I will create new test cases once the attribution window protocol is applied.

martinthomson (commented):
This change looks fine (other than the cost, which seems pretty tolerable overall), but I want to better understand the approach you are taking with respect to timestamps.

What size are timestamps?  How many bits, what range?  How does that manifest in code?  Are you expecting the report collector to generate them?  How will they be expressed in submissions?

There are a few optimizations we might want to consider, but also some trade-offs.

taikiy:
I don't have all the answers, but my understanding of the timestamp values according to the e2e doc and Ben Case's paper is:

* an offset in seconds into a given epoch
* assuming an epoch is 1 week, we need 604,800 seconds = 20 bits
* timestamps are in clear

I suppose that the report collector will ensure that epoch and timestamp are in whatever the format we end up with. From protocol layer POV, as you mentioned, this code shouldn't be affected much. One thing we haven't really talked about is how the epoch works. I remember in our group discussions we talked about how match keys are tied to epochs, meaning the encrypted value will change when the epoch rolls, but someone said that's not the case anymore? How we do cross-epoch attribution is another thing, although if match keys stay the same across epochs, then we don't have a problem.

benjaminsavage (approved):

PR #568, opened 2023-03-24
Add bytes.sent metric to count bytes sent between helpers

akoshelev:
A companion to `records.sent` that measures the total communication in bytes between helpers. With recent developments in IPA, the total number of records may stay the same, but size of messages sent is reduced. `records.sent` does not capture it, so we would rely on `bytes.sent` to check those improvements.

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -851,15 +851,19 @@ pub mod tests {
 
         /// empirical value as of Mar 24, 2023.
         const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 14517;
+        const BYTES_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 48780;
 
         /// empirical value as of Mar 24, 2023.
         const RECORDS_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 36543;
+        const BYTES_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 136_884;
 
         /// empirical value as of Mar 24, 2023.
         const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 10848;
+        const BYTES_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 34104;
 
         /// empirical value as of Mar 24, 2023.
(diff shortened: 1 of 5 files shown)


src/telemetry/step_stats.rs
```diff
@@ -35,13 +35,13 @@ impl CsvExporter for Metrics {
         // then dump them to the provided Write interface
         // TODO: include role dimension. That requires rethinking `Metrics` implementation
         // because it does not allow such breakdown atm.
-        writeln!(w, "Step,Records Sent,Indexed PRSS,Sequential PRSS")?;
+        writeln!(w, "Step,Bytes sent,Indexed PRSS,Sequential PRSS")?;
```
benjaminsavage, line 38: Why not both?

src/protocol/ipa/mod.rs
```diff
@@ -860,19 +864,25 @@ pub mod tests {
                 .reconstruct();
 
             let snapshot = world.metrics_snapshot();
-            let records_sent = snapshot.get_counter(RECORDS_SENT);
-            let semi_honest_baseline = if per_user_cap == 1 {
-                RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1
+            let (records_baseline, bytes_baseline) = if per_user_cap == 1 {
+                (
+                    RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1,
+                    FIELD_SIZE * RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1,
```
benjaminsavage, line 870: I don’t think this is right. Currently IPA runs two separate validators, each over a subset of the protocol, so some records are sent in field 1 and others in field 2

src/telemetry/step_stats.rs
akoshelev: I was thinking that number of records may no longer be relevant now, but I don't mind adding it back

benjaminsavage: I think it’s really helpful for analyzing where we should focus on optimizations.

src/protocol/ipa/mod.rs
akoshelev: yep, my rebase was 4 hours old and stale already!

benjaminsavage (approved):
Thanks Alex, this is super useful. I'm already building all my new PRs on top of it, so I'm happy to merge it!

PR #570, opened 2023-03-27
Script to generate a report with benchmarking results

richajaindce:
===============================
Command to run
===============================
scripts/benchmark_ipa.sh QUERY_SIZE

This would run oneshot_ipa with QUERY_SIZE and output the benchmark results

===============================
Sample output (on the terminal)
===============================
Step,Bytes sent,Records Sent
Verify,153600,38400
Sort,716040,197010
Attribution,85419,30042
Capping,436256,109064
Aggregation,64800,16200
Others,148956,37239
Total,1605071,427955

### DIFF
--- a/scripts/benchmark_ipa.sh
+++ b/scripts/benchmark_ipa.sh
@@ -11,7 +11,7 @@ if [ "$#" -ne 1 ]; then
 fi
 
 temp_file=".benchmark_result$num_rows.csv"
-RUST_LOG=raw_ipa=DEBUG cargo bench --bench oneshot_ipa --features="enable-benches" --no-default-features -- -n $num_rows > $temp_file
+RUST_LOG=ipa=DEBUG cargo bench --bench oneshot_ipa --features="enable-benches" --no-default-features -- -n $num_rows > $temp_file
  
 until_now_records=0
 until_now_bytes=0
--- a/scripts/benchmark_ipa.sh
+++ b/scripts/benchmark_ipa.sh
@@ -23,17 +23,17 @@ until_now_bytes=0
 
 echo "Step,Bytes sent,Records Sent"
 step_mapping=( 
-        "mod_conv_breakdown_key\|mod_conv_match_key:Verify"
-        "apply_sort_permutation\|gen_sort_permutation_from_match_keys:Sort"
-	"accumulate_credit\|compute_helper_bits:Attribution"
+    "mod_conv_breakdown_key\|mod_conv_match_key:Verify"
+    "apply_sort_permutation\|gen_sort_permutation_from_match_keys:Sort"
+    "accumulate_credit\|compute_helper_bits:Attribution"
   	"user_capping:Capping"
-        "check_times_credit\|compute_equality_checks:Aggregation"
+    "check_times_credit\|compute_equality_checks:Aggregation"
 )
         
 
 for step in "${step_mapping[@]}" ; do
-    key=${step%%:*}
-    value=${step#*:}
+    key="${step%%:*}"
+    value="${step#*:}"
     records_sent=`grep "$key" $temp_file | awk -F, '{sum+=$2;}END{print sum;}'`
     bytes_sent=`grep "$key" $temp_file | awk -F, '{sum+=$3;}END{print sum;}'`
 
--- a/scripts/benchmark_ipa.sh
+++ b/scripts/benchmark_ipa.sh
@@ -10,8 +10,13 @@ if [ "$#" -ne 1 ]; then
     exit
 fi
 
-temp_file=".benchmark_result$num_rows.csv"
-RUST_LOG=ipa=DEBUG cargo bench --bench oneshot_ipa --features="enable-benches" --no-default-features -- -n $num_rows > $temp_file
+temp_file=`mktemp -q /tmp/ipa-bench.XXXXXX`
+ if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


scripts/benchmark_ipa.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# To use, supply the number of rows you want to run benchmarking on
+
+num_rows=$1
+if [ "$#" -ne 1 ]; then
+    echo "Please supply query size as an argument to the script"
+    exit
+fi
+
+temp_file=".benchmark_result$num_rows.csv"
```
akoshelev, line 13:
```suggestion
temp_file=`mktemp -q /tmp/ipa-bench.XXXXXX`
 if [ $? -ne 0 ]; then
         echo "$0: Can't create temp file, exiting..."
         exit 1
 fi
```

that way, even if someone presses ctrl-c while running ipa bench, they won't face "file exists" error message after re-running this script

scripts/benchmark_ipa.sh
```diff
@@ -0,0 +1,47 @@
...
+        "mod_conv_breakdown_key\|mod_conv_match_key:Verify"
+        "apply_sort_permutation\|gen_sort_permutation_from_match_keys:Sort"
+	"accumulate_credit\|compute_helper_bits:Attribution"
+  	"user_capping:Capping"
+        "check_times_credit\|compute_equality_checks:Aggregation"
+)
+        
+
+for step in "${step_mapping[@]}" ; do
+    key=${step%%:*}
```
martinthomson, line 35:
```suggestion
    key="${step%%:*}"
```
Variable expansions should be quoted.

scripts/benchmark_ipa.sh
richajaindce: Didnt know about mktemp.. it looks nice!

scripts/benchmark_ipa.sh
```diff
@@ -0,0 +1,47 @@
...
+# To use, supply the number of rows you want to run benchmarking on
+
+num_rows=$1
+if [ "$#" -ne 1 ]; then
+    echo "Please supply query size as an argument to the script"
+    exit
+fi
+
+temp_file=".benchmark_result$num_rows.csv"
+RUST_LOG=ipa=DEBUG cargo bench --bench oneshot_ipa --features="enable-benches" --no-default-features -- -n $num_rows > $temp_file
```
martinthomson: There is an alternative approach that doesn't involve a temporary file.  Try:

```sh
declare -A bytes
declare -A records
RUST_LOG=ipa... cargo bench ... | while IFS=, read line k a b c; do
    # Maybe permute k so that you can get the step more easily:
    case "$k" in
      */mod_conv_breakdown_key/*|*/mod_conv_match_key/*) step=Verify ;;
      # etc...
      *) step=Other ;;
    esac
    records[$step]=$(($records[$step] + $a))
    bytes[$step]=$((${bytes[$step]} + $b))
done

for s in "${!records[@]}"; do # Or you can just enumerate the steps so that the order is consistent.
    echo "$s: ${records[$s]} records, ${bytes[$s]} bytes"
done
```

martinthomson: I only used associative arrays because that is easier to manage.  You can use different variables instead; it just looks a bit ugly.

Also, blame Apple.  They refuse to ship a GPL-3 version of bash, so they are stuck in the past.

richajaindce: I am having hard time to make it work. Looks like piping data makes the variable disappear since piping creates a SubShell which runs in its own environment : http://mywiki.wooledge.org/BashFAQ/024

richajaindce: Took me a while to figure this out.. but hopefully it looks good now :)

scripts/benchmark_ipa.sh
```diff
@@ -0,0 +1,47 @@
...
+  bytes+=(0);
+done
+
+while IFS=, read -d ' ' actual_step_name a b c d; do
+    if [[ $actual_step_name != *"protocol"* ]]; then 
+      continue
+    fi
+
+    # In case you add a new stage to DISPLAY_STAGE, please add corresponding step 
+    # in below case to match expected DISPLAY_STAGE array index
```
martinthomson, line 31: You could maybe remove maintenance overhead with some meta-programming, but this is fine for now...

```bash
steps() {
if [[ "$1" == "f" ]]; then
    echo 'match_step() {'
    echo '    case "$1" in'
fi
i=0
while read name pattern; do
    if [[ "$1" == "f" ]]; then
        echo "        ${pattern}) step=$i ;;"
    else
        echo "DISPLAY_STAGE+=(\"$name\");"
    fi
    i=$((i + 1))
done <<EOSTEPS
Verify */mod_conv_breakdown_key/*|*/mod_conv_match_key/*
Sort */apply_sort_permutation/*|*/gen_sort_permutation_from_match_keys/*
EOSTEPS
if [[ "$1" == "f" ]]; then
   echo '    esac'
    echo '}'
fi
}
eval "$(steps f)"
DISPLAY_STAGE=()
eval "$(steps n)"
```

martinthomson (approved):
LGTM

scripts/benchmark_ipa.sh
richajaindce: This makes the script very hard to read :p can we leave it as is for now?

akoshelev:
merging as it's been open for a week and major feedback is addressed.

PR #571, opened 2023-03-28
Breakdown keys GF2

benjaminsavage:
This is a draft - but it shows the direction I'm exploring, using the GF(2) extension field to optimize the breakdown key bits


martinthomson:
No longer needed?

PR #573, opened 2023-03-28
Use dynamic dispatch for transport callbacks

akoshelev:
Having spent many hours trying to make static dispatch work for in memory network I am swinging towards this direction. Setting up one callback for 3 in memory helpers is extremely verbose and because closures are not equal to each other, requires **3** generic arguments. 

```rust
  /// callbacks for 3 helpers (receive query only)
  struct NetworkSetup<RQC1: ReceiveQueryCallback, RQC2: ReceiveQueryCallback, RQC3: ReceiveQueryCallback> {
      transport1: Setup<RQC1>,
      transport2: Setup<RQC2>,
      transport3: Setup<RQC3>,
  }
```

And that is only for one callback, where we would have more: receive query callback, prepare query callback, receive inputs callback, etc.

It will come with some cost of boxing, but I don't see an issue here - those callbacks are not on the hot path anyway.

P.S. I'd love to move away from callbacks, but I don't see another way. Transports could potentially be polled for signals, so `Transport` -> `QueryProcessor` delivery is easy, but transport also needs to wait for the outcome and return it back to network. So In reality the link is bidirectional: `Transport` <-> `QueryProcessor`.

### DIFF
--- a/src/test_fixture/network/transport.rs
+++ b/src/test_fixture/network/transport.rs
@@ -42,7 +42,7 @@ pub struct InMemoryTransport {
 
 impl InMemoryTransport {
     #[must_use]
-    pub fn with_stub_callbacks(identity: HelperIdentity) -> Setup<impl ReceiveQueryCallback> {
+    pub fn with_stub_callbacks(identity: HelperIdentity) -> Setup {
         Setup::new(identity, stub_callbacks())
     }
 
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Yes, this seems like the best of a bunch of poor options here.  As you say, performance isn't critical here and this might help simplify elsewhere.

PR #574, opened 2023-03-28
Adding `apply_attribution_window` step to secure_attribution

taikiy:
This is a stacked diff on top of #567. The diff to be reviewed is b8e543c4a6d9822d19882c7c5c9e73c120608057.

This PR adds `apply_attribution_window` step to the e2e IPA protocol. If a query provides the attribution window parameter, then we'll apply the protocol.

Currently, attribution window will not work as expected if `per_user_cap` is 1. This is because the current attribution window protocol works by setting 0 to expired (outside the attribution window) trigger values. I'll send out a new solution in the next diff.

cap = 3
  semi_honest: 14571 -> 18258
  malicious: 36678 -> 46068

cap = 1
  semi_honest: 10902 -> 14589
  malicious: 27324 -> 36714

The cost increase comes from:
1. Time delta computation (prefix-sum and two mults)
2. Comparison of the publicly known attribution window and the time delta values (random bits generator, comparison, and a mult)


src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -20,31 +20,49 @@ use std::iter::{repeat, zip};
...
 /// # Errors
 /// Fails if sub-protocols fails.
-#[allow(dead_code)]
-async fn apply_attribution_window<F, C, T>(
+pub async fn apply_attribution_window<F, C, T>(
     ctx: C,
     input: &[MCApplyAttributionWindowInputRow<F, T>],
     attribution_window_seconds: u32,
-) -> Result<impl Iterator<Item = MCApplyAttributionWindowOutputRow<F, T>> + '_, Error>
+) -> Result<Vec<MCApplyAttributionWindowOutputRow<F, T>>, Error>
```
benjaminsavage, line 27: Fine for now (temporary), but I think we will probably wind up converting everything back to iterators soon.

src/protocol/ipa/mod.rs
```diff
@@ -605,11 +645,121 @@ pub mod tests {
...
+
+        let world = TestWorld::default();
+
+        let records: Vec<GenericReportTestInput<Fp31, MatchKey, BreakdownKey>> = ipa_test_input!(
+            [
+                { timestamp: 0, match_key: 12345, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 2, match_key: 12345, is_trigger_report: 0, breakdown_key: 2, trigger_value: 0 },
+                { timestamp: 3, match_key: 68362, is_trigger_report: 0, breakdown_key: 1, trigger_value: 0 },
+                { timestamp: 12, match_key: 12345, is_trigger_report: 1, breakdown_key: 0, trigger_value: 5 },
+                { timestamp: 15, match_key: 68362, is_trigger_report: 1, breakdown_key: 0, trigger_value: 2 },
```
benjaminsavage, line 762: Is this the same as the rows above? Can you eliminate this duplication? Maybe just make it a constant or something?

benjaminsavage (approved):
LGTM

PR #575, opened 2023-03-29
Use type aliases for Gateway and ReceivingEnd

akoshelev:
Builds on top of #573 

It turns out the enum dispatch for the transport layer
implemented in `TransportImpl` is unusable with multiple transport implementations.

It does not allow decorators/wrappers to be used without adding another enum variant
which is horrible in my opinion. Testing `QueryProcessor` requires use of specialized
transports (`DelayedTransport`) and I couldn't just make it work even when I use the
atrocious downcasting `Transport` -> `InMemoryTransport`.

We still want `Gateway` to be generic free, but it is just not flexible for the infrastructure
layer tests. So I thought that I could maybe make it work if I lift type aliases one level above:
i.e. instead of using type aliases for transport implementations, I could use them to provide
different `Gateway` definitions depending on what underlying transport is used.

That turns out to cover my needs on the infra side and requires no change on MPC side. I can now
instantiate `Gateway` with whatever `Transport` implementation I have handy and the horrible `TransportImpl`
is gone now, which is another win imo.

### DIFF
--- a/src/helpers/gateway/mod.rs
+++ b/src/helpers/gateway/mod.rs
@@ -2,32 +2,48 @@ mod receive;
 mod send;
 mod transport;
 
-pub use receive::ReceivingEnd;
 pub use send::SendingEnd;
 
 use std::{fmt::Debug, num::NonZeroUsize};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


akoshelev:
addressed by #584

PR #576, opened 2023-03-29
Revert gateway send buffer capacity back from bytes to units

akoshelev:
Recent [change](https://github.com/private-attribution/ipa/pull/566) made it possible for IPA to send different field values, so send buffer capacity must be adjusted to that. Setting up raw bytes capacity no longer works and here is why:

assume we set send buffer capacity to 1024 bytes. That would mean that it could keep up to 256 `Fp32BitPrime` values or 1024 `Gf2` values. Given that we are moving towards defining a sliding window of records we process in parallel (see @martinthomson's efforts to implement [`seq_try_join_all`](https://github.com/private-attribution/ipa/commit/bfc7a09b46bfb4f3de18693882192424d83addff)) we no longer can rely on universal send buffer capacity.

The proposed fix includes defining send buffer capacity in terms of items to process and adjusting byte capacity by calculating the size of one message (field value) and multiplying it by number of concurrent items it needs to hold.

Using the example above: if send buffer capacity is 1024 items, then `Gf2` buffer capacity will be set to 1024 bytes and `Fp32BitPrime` buffer capacity will be 4096 bytes. Note that if we want to optimize the network utilization, we would rather set `Gf2` buffer capacity to 4096 items while keeping `Fp32BitPrime` at 1024 items, but I don't think it is possible with `seq_try_join_all` API that uses number of records.

To explain why it wouldn't work: if `seq_try_join_all` processes up to 4096 items in parallel, it may underutilize send buffer for `Gf2` if it is ever set to be more than 1024 items, while being perfectly capable of using less than 25% of send buffer capacity for `Fp32BitPrime`.

Overall, it feels easier to use the same units across the system, currently it seems the number of records prevails, so lets use it everywhere then.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -49,8 +49,7 @@ async fn main() -> Result<(), Error> {
 
     let prep_time = Instant::now();
     let mut config = TestWorldConfig::default();
-    config.gateway_config =
-        GatewayConfig::symmetric_buffers::<BenchField>(args.query_size.clamp(16, 1024));
+    config.gateway_config = GatewayConfig::symmetric_buffers(args.query_size.clamp(16, 1024));
 
     let seed = args.random_seed.unwrap_or_else(|| thread_rng().gen());
     println!(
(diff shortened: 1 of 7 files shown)


martinthomson (approved):
Thanks Alex, this is a big improvement in addition to fixing the bug.

PR #579, opened 2023-04-05
Cap 1 look-back window

taikiy:
This diff introduces a new auxiliary bit share called "active_bit". An active_bit is a result of `greater_than_constant()` that indicates whether the corresponding credit is outside the window (0) or valid (1). If the per_user_cap = 1, we have a special branch of accumulation protocol which ignores trigger values entirely, but instead uses helper_bits and is_trigger_bits to *count* the number of events. Active bits will be used to zero-out those events if they are outside the window.

As a result, this diff adds 1 multiplication per row.


src/protocol/attribution/accumulate_credit.rs
```diff
@@ -15,29 +15,54 @@ use crate::{
...
 where
     F: Field,
     C: Context,
     T: LinearSecretSharing<F> + BasicProtocols<C, F>,
 {
-    let output = input.iter().zip(stop_bits).map(|(x, credit)| {
+    // if `attribution_window_seconds` is 0, we use `stop_bits` directly. Otherwise, we need to invalidate
+    // credits that are outside the attribution window by multiplying them by `active_bit`. active_bit is
+    // 0 if the trigger report's time-delta to the nearest source report is greater than the attribution window.
+    let credits = if attribution_window_seconds == 0 {
```
benjaminsavage, line 39: I don't like the name "credits". It confused me for a while here. Since these are always 0 or 1, they just indicate "was there an attributed trigger event" right? I wonder if we could fine another name to evoke this idea. Perhaps something like:

`attributed_trigger_events_in_window`

src/protocol/attribution/input.rs
```diff
@@ -87,26 +89,59 @@ impl<F: Field, T: LinearSecretSharing<F>> MCAccumulateCreditInputRow<F, T> {
...
-pub type MCAccumulateCreditOutputRow<F, T> = MCAccumulateCreditInputRow<F, T>;
+pub type AccumulateCreditOutputRow<F, BK> = CreditCappingInputRow<F, BK>;
+pub type MCAccumulateCreditOutputRow<F, T> = MCCreditCappingInputRow<F, T>;
 
 //
 // `credit_capping` protocol
 //
-pub type CreditCappingInputRow<F, BK> = AccumulateCreditInputRow<F, BK>;
-pub type MCCreditCappingInputRow<F, T> = MCAccumulateCreditInputRow<F, T>;
+pub struct CreditCappingInputRow<F: Field, BK: GaloisField> {
```
benjaminsavage, line 113: @martinthomson - what's your opinion about all of these various struct definitions for each step of the protocol?

Personally, I am not in love with it. I think they have the effect of making the code look more complicated than it actually is. We also find ourselves serialising back and forth quite a bit from vectors of these structs, to vectors of just one element from the struct.

I also really don't like how much code these create, for all of the `reshare` and `upgrade` and test input, and reconstruct functions.

I think I'm in favor of just eliminating 100% of these structs, and simply passing around vectors of each of the things inside of them.

e.g.

```
pub async fn accumulate_credit<F, C, T>(
    ctx: C,
//    input: &[MCAccumulateCreditInputRow<F, T>], eliminate this
    is_trigger_report_bits: &[T],
    helper_bits: &[T],
//    breakdown_keys: &[T] I don't think this is even needed for accumulate credit step
    trigger_values: &[T],
    stop_bits: &[T],
    per_user_credit_cap: u32,
    attribution_window_seconds: u32,
) -> Result<Vec<MCAccumulateCreditOutputRow<F, T>>, Error>
where
    F: Field,
    C: Context,
    T: LinearSecretSharing<F> + BasicProtocols<C, F>,
{
```

src/protocol/ipa/mod.rs
```diff
@@ -1004,13 +1097,13 @@ pub mod tests {
...
+        // empirical value as of Apr 6, 2023.
+        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 14_589;
+        const BYTES_SENT_SEMI_HONEST_BASELINE_CAP_1: u64 = 49_068;
 
-        // empirical value as of Mar 30, 2023.
-        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 36_666;
-        const BYTES_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 137_376;
+        // empirical value as of Apr 6, 2023.
+        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 36_714;
+        const BYTES_SENT_MALICIOUS_BASELINE_CAP_1: u64 = 137_568;
```
benjaminsavage, line 1106: This is OK. Very modest increases.

src/protocol/ipa/mod.rs
```diff
@@ -876,6 +876,99 @@ pub mod tests {
...
+                    result[i].breakdown_key.as_u128(),
+                    result[i].trigger_value.as_u128()
+                ]
+            );
+        }
+    }
+
     #[tokio::test]
     #[allow(clippy::missing_panics_doc)]
     pub async fn random_ipa_check() {
```
benjaminsavage, line 974: This is actually my favorite test, and the easiest for me to reason about. I'd love to update this one to perform windowing of say 1-day (currently the records all have random timestamps over the course of a week).

I don't like reading tests with hard-coded inputs because it's just a lot of mental work to think about what they mean and what the right answer is supposed to be.

Could you please update the random_ipa_check test?

taikiy: I have the diff ready :)

src/protocol/attribution/accumulate_credit.rs
taikiy: Renamed!

src/protocol/attribution/input.rs
martinthomson: I'm not sure that there is an easy answer here.  I definitely agree that transforming stuff is going to be costing us more than is good for us, but massive function signatures and no encapsulation has some other consequences that aren't great either.

If there are places where these inputs can be logically grouped in ways that work across multiple stages of the computation, then we might do what you have there, with some groupings that make sense.

PR #582, opened 2023-04-05
Go faster with seq_join

martinthomson:
This isn't perfect, but it is quite a bit faster.  I was able to shave about 25% of the run time of a 10k records test.  The `-a` argument to the bench determines how much work is kept in flight and I saw modest gains going from 1k to 2k and then 4k, with each saving almost 10% each, with the 1k value being about 10% better than the code on `main`.

There are some extra bits in here that I've split out into separate pull requests.


src/protocol/attribution/credit_capping.rs
```diff
@@ -284,50 +288,67 @@ where
...
-
-                let original_credit = &original_credits[i];
-                let next_prefix_summed_credit = &prefix_summed_credits[i + 1];
-                let current_prefix_summed_credit_exceeds_cap = &exceeds_cap_bits[i];
-                let next_credit_exceeds_cap = &exceeds_cap_bits[i + 1];
-                let next_event_has_same_match_key = &input[i + 1].helper_bit;
-
-                let remaining_budget = if_else(
-                    ctx.narrow(&Step::IfNextEventHasSameMatchKeyOrElse),
+    let capped = zip(
```
martinthomson, line 291: This is the most interesting item.  We can't do what we did before, which had the processing for the last record return immediately.  That violates the seq_join contract, and I don't want to append to the vector and maybe reallocate.  So I've used seq_join here directly.

src/protocol/attribution/mod.rs
```diff
@@ -134,8 +135,8 @@ where
         }
 
         let (stop_bit_updates, credit_updates) = try_join(
-            try_join_all(stop_bit_futures),
-            try_join_all(credit_update_futures),
+            assert_send(ctx.try_join_all(stop_bit_futures)),
```
martinthomson, line 138: Credit to Andy for finding these and wrapping them.

src/seq_join.rs
```diff
@@ -0,0 +1,361 @@
...
+};
+
+/// This helper function might be necessary to convince the compiler that
+/// the return value from [`seq_try_join_all`] implements `Send`.
+/// Use this if you get higher-ranked lifetime errors that mention `std::marker::Send`.
+// pub fn assert_send<'a, F, O>(
+//     fut: F,
+// ) -> F where F: Future<Output = O> + Send + 'a {
+//     fut
+// }
```
martinthomson, line 19: I should remove this.  It is basic desugaring as far as I can tell, but it doesn't work.

src/seq_join.rs
```diff
@@ -0,0 +1,361 @@
...
+/// Unlike [`StreamExt::buffered`], Futures from the stream must resolve in the
+/// same order in which they are produced.
+///
+/// This API accepts an active stream count and a [`Stream`] instance,
+/// from which `Future`s are drawn.
+/// To replicate the effect of [`try_join_all`], an input `Iterator` can be adapted using
+/// [`futures::stream::iter`]; the output can be collected into a [`Vec`] (or any
+/// collection) using [`StreamExt::collect`], as follows:
+///
+/// ```ignore
```
martinthomson, line 41: These docs are a little out of date.  I still want to take a `Stream` as input here, but working out how to compile that was taking too much energy.

src/seq_join.rs
```diff
@@ -0,0 +1,361 @@
...
+    #[should_panic(expected = "future resolved out of order")]
+    async fn out_of_order() {
+        let capacity = NonZeroUsize::new(3).unwrap();
+        let unresolved: BoxFuture<'_, u32> = Box::pin(pending());
+        let it = once(unresolved)
+            .chain((0..3_u32).map(|i| -> BoxFuture<'_, u32> { Box::pin(async move { i }) }));
+        drop(seq_join(capacity, it).collect::<Vec<_>>().await);
+    }
+
+    #[cfg(seq_join_stream)]
```
martinthomson, line 212: This is the same as the comments.  I want streams, but apparently we can't have nice things.

src/test_fixture/mod.rs
```diff
@@ -111,8 +107,8 @@ where
     T::Ok: Debug,
     T::Error: Debug,
 {
```
martinthomson, line 109: Note to self, remove panics doc for this.

src/seq_join.rs
```diff
@@ -0,0 +1,356 @@
...
+    fut
+}
+
+/// Sequentially join futures from an iterator.
+///
+/// This function polls futures from a stream in strict sequence.
+/// If any future blocks, up to `active - 1` futures after it will be polled so
+/// that they make progress.
+///
+/// Unlike [`StreamExt::buffered`], Futures from the stream must resolve in the
```
akoshelev, line 27: this requirement is interesting, I read the code and I believe it is coming from the implementation where at the time when future is polled and resolved out of order, we don't have a container to store the value received from it. Given that we box anyway, we could probably use an enum to describe elements within the deque: either future or value received from it. 

I don't think we need to do it in this PR, just wondering if it is something we should consider in the future. I don't think we would see impact for in-memory implementations, but curious if this could help us avoid crashes when we start doing real IO

akoshelev (approved):

src/protocol/attribution/apply_attribution_window.rs
```diff
@@ -12,7 +12,6 @@ use crate::{
     },
     secret_sharing::Linear as LinearSecretSharing,
 };
-use futures::future::try_join_all;
```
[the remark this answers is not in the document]
martinthomson: The problem we have right now is that try_join_all is sometimes necessary.  Perhaps that's OK and we can just use `#[allow...]`.  Let me look into it.

src/protocol/attribution/credit_capping.rs
```diff
@@ -284,50 +288,67 @@ where
...
+        },
+    );
 
-                Ok::<_, Error>(capped_credit)
-            }),
-    )
-    .await
+    let last = original_credits.last().ok_or(Error::Internal).cloned();
+
+    seq_join(ctx.active_work(), capped)
```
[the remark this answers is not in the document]
martinthomson: Probably.  Though this is a bit of an odd case.

src/seq_join.rs
martinthomson: This might be the way we resolve some of the weird ordering problems I'm dealing with, yeah.  The random bit generator (may it burn in a very special place) has this property and maybe switching types would help there.  I'll put that into a follow-up (just as I want to follow up with the change to using a `Stream` at some point).

src/protocol/context/malicious.rs
```diff
@@ -530,28 +538,25 @@ impl<'a, F: Field + ExtendableField>
...
-                    self.record_binding,
-                    v2,
-                    ZeroPositions::Pzzv,
-                ),
-            ])
-            .await?
-            .try_into()
-            .unwrap(),
-        ))
+        let (t0, t1, t2) = try_join3(
```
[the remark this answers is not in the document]
martinthomson: I'm not seeing a bug, by the way, just a case where we didn't need to be using an array/iterator input form.

akoshelev:
Lets close this one and iterate if we need to - tons of useful improvements here!

PR #583, opened 2023-04-05
Clean up the leak in handles_reordering test

martinthomson:


akoshelev (approved):
just curious if you saw an impact of this leak. Agree that it is better not to have it at all

PR #584, opened 2023-04-05
Bring back query processing functionality

akoshelev:
This change gets rid of many `#[cfg(never)]` blocks, but not all. The main focus is to integrate `QueryProcessor` with the new transport interface, which turned out to be not so easy. 

The outcome here is that end-to-end in memory query processing works again. The HTTP part is not ready yet  (@andyleiserson is working on it). 

Some other notable changes here:
* Public key exchange no longer requires sending chunks of public key, can be done in one shot.
* Keeping context free of `Transport` generic argument continues to be a challenge. At some point the conversion from `T` to `TransportImpl` needs to happen and query processor hits that path when it is ready to start the computation. It is ugly and I am open to suggestions how to make it better. 
* `Transport::send` contract now requires the acknowledgment from the receiving end. Hopefully it shouldn't be a problem for HTTP implementation.


src/helpers/prss_protocol.rs
```diff
@@ -21,198 +18,41 @@ impl AsRef<str> for PrssExchangeStep {
...
 impl Substep for PrssExchangeStep {}
 
-pub const PUBLIC_KEY_CHUNK_COUNT: usize = 4;
-
 /// establish the prss endpoint by exchanging public keys with the other helpers
 /// # Errors
 /// if communication with other helpers fails
-pub async fn negotiate<R: RngCore + CryptoRng>(
-    gateway: &Gateway,
+pub async fn negotiate<T: Transport, R: RngCore + CryptoRng>(
```
akoshelev, line 24: key exchange is much simpler now

src/helpers/transport/callbacks.rs
```diff
@@ -0,0 +1,57 @@
...
+    helpers::{
+        query::{PrepareQuery, QueryConfig},
+        TransportError,
+    },
+    protocol::QueryId,
+};
+use std::{future::Future, pin::Pin};
+
+/// Called when helper receives a new query request from an external party.
+pub trait ReceiveQueryCallback<'a, T>:
```
akoshelev, line 11: callbacks live in their own module now

src/query/processor.rs
```diff
@@ -445,130 +435,43 @@ mod tests {
...
-                    ))
-                    .await;
-                handle_resps.push(rx);
-            }
+        async fn complete_query_test_multiply() -> Result<(), BoxError> {
+            let app = TestApp::default();
+            let a = Fp31::truncate_from(4u128);
+            let b = Fp31::truncate_from(5u128);
+            let results = app
+                .execute_query(
```
akoshelev, line 453: `TestApp` makes testing queries a bit easier

src/query/state.rs
```diff
@@ -77,17 +76,23 @@ pub enum StateError {
 
 /// Keeps track of queries running on this helper.
 pub struct RunningQueries {
-    pub inner: Arc<Mutex<HashMap<QueryId, QueryState>>>,
+    pub inner: Mutex<HashMap<QueryId, QueryState>>,
```
akoshelev, line 79: `Arc` was redundant here

src/test_fixture/network/receive.rs
```diff
@@ -0,0 +1,57 @@
+use crate::test_fixture::network::stream::{StreamCollection, StreamKey};
```
akoshelev, line 1: no functional changes here, just moving things around. In memory transport module was too big already

src/test_fixture/network/stream.rs
```diff
@@ -0,0 +1,134 @@
+use crate::{
+    helpers::HelperIdentity,
```
akoshelev, line 2: same here, just moving things out from `transport` module

src/tests/infra.rs
```diff
@@ -127,3 +132,43 @@ fn send_receive_parallel() {
         1000,
     );
 }
+
+#[test]
+fn execute_query() {
```
akoshelev, line 137: shuttle test to cover query processing logic

src/helpers/gateway/mod.rs
```diff
@@ -2,30 +2,42 @@ mod receive;
...
+
+pub type Gateway = GatewayBase<TransportImpl>;
+pub type ReceivingEnd<M> = ReceivingEndBase<TransportImpl, M>;
+
+/// Gateway into IPA Infrastructure systems. This object allows sending and receiving messages.
+/// As it is generic over network/transport layer implementation, type alias [`Gateway`] should be
+/// used to avoid carrying `T` over.
+///
+/// [`Gateway`]: crate::helpers::Gateway
+pub struct GatewayBase<T: Transport> {
```
andyleiserson, line 36: `pub struct Gateway<T: Transport = TransportImpl>` is also a possibility.

(I assume you're using the type parameter to catch accidental use of non-generic methods on whichever impl is configured in a particular build? If the concern is just making sure that `InMemoryTransport` and `HttpTransport` share a common interface, there's an idiom I can suggest to enforce that without having to code against the `Transport` trait, but it won't prevent non-trait functionality from leaking through.)

src/helpers/transport/mod.rs
```diff
@@ -97,24 +107,38 @@ impl RouteParams<RouteId, QueryId, Step> for (RouteId, QueryId, Step) {
...
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io {
+        #[from]
+        inner: io::Error,
+    },
+    #[error("Request rejected by remote {dest:?}: {inner:?}")]
```
andyleiserson, line 122: "Error from remote helper"?

andyleiserson (approved):
> Keeping context free of Transport generic argument continues to be a challenge. At some point the conversion from T to TransportImpl needs to happen and query processor hits that path when it is ready to start the computation. It is ugly and I am open to suggestions how to make it better.

I'm not sure which part you think is ugly?

> Transport::send contract now requires the acknowledgment from the receiving end. Hopefully it shouldn't be a problem for HTTP implementation.

Interesting question. Hopefully not, but it will depend how much visibility hyper gives us to the progress of the request. I'll find out soon.

src/test_fixture/app.rs
```diff
@@ -0,0 +1,115 @@
...
+{
+    fn into_buf(self) -> Vec<u8> {
+        let this = self.into_iter();
+        let (lb, ub) = this.size_hint();
+        let cnt = ub.unwrap_or(lb);
+        let item_size: usize = <S as Serializable>::Size::USIZE;
+
+        let mut buf = vec![0u8; cnt * item_size];
+        for (i, item) in this.enumerate() {
+            let sl = &mut buf[i * item_size..(i + 1) * item_size];
```
martinthomson, line 35: I wonder if we can't use something more clever here.

```rust
for (item, chunk) in zip(self.into_iter(), buf.chunks_mut(item_size)) {
  item.serialize(GenericArray::from_mut_slice(chunk));
}
```

src/test_fixture/app.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl<I, S> IntoBuf for I
+where
+    I: IntoIterator<Item = S>,
+    S: Serializable,
+{
+    fn into_buf(self) -> Vec<u8> {
+        let this = self.into_iter();
+        let (lb, ub) = this.size_hint();
+        let cnt = ub.unwrap_or(lb);
```
martinthomson, line 30: This seems a bit sketchy to me.  If your upper bound is `None` or short of the true size of the input, you will allocate  something that is too small and then overrun the buffer (and panic).  I think that we want an additional constraint: `I::IntoIter: ExactSizeIterator`, so that you can just do:

```suggestion
        let cnt = self.len();
```

martinthomson (approved):
Looks good to me.

akoshelev:
merging it as most feedback has been addressed and we need to bet #587 out of the door

PR #585, opened 2023-04-06
Disallow array joins

martinthomson:
Builds on #582 and disallows joins through a lint.  There are exceptions to that, but this might help keep things under control.


src/query/executor.rs
```diff
@@ -68,7 +69,7 @@ where
 async fn execute_test_multiply<F: Field>(
     ctx: SemiHonestContext<'_>,
     mut input: AlignedByteArrStream,
-) -> Vec<Replicated<F>>
+) -> std::result::Result<Vec<Replicated<F>>, Error>
```
martinthomson, line 72: The redefinition of `Result` as a trait in this file is a bit of a problem.  It really did my head in for a while there.

src/helpers/buffers/ordering_sender.rs
```diff
@@ -385,6 +385,7 @@ impl<B: Borrow<OrderingSender> + Unpin> Stream for OrderedStream<B> {
 
 #[cfg(test)]
 mod test {
+    #![allow(clippy::disallowed_methods)] // It's a test without a context to use.
```
akoshelev, line 388: I'll see if we can disable this lint for all tests, it is a bit tedious

src/seq_join.rs
```diff
@@ -0,0 +1,250 @@
...
+        seq_try_join_all(self.active_work(), iterable)
+    }
+
+    /// Join multiple tasks in parallel.  Only do this if you can't use a sequential join.
+    fn parallel_join<I>(&self, iterable: I) -> futures::future::TryJoinAll<I::Item>
+    where
+        I: IntoIterator,
+        I::Item: futures::future::TryFuture,
+    {
+        #[allow(clippy::disallowed_methods)] // Just in this one place.
```
akoshelev, line 95: wouldn't it be nice if we were able to allow `try_join_all` only here?

akoshelev (approved):

PR #586, opened 2023-04-06
Allow disallowed methods in tests

akoshelev:
Allow array joins in tests

Follow up on #585 - relax this requirement to avoid array joins in prod code only
Also this change proposes to rename `join` to `try_join` - see commit message for details why.

 
## Testing

* removed lint suppression in prod code and verified that clippy is angry about it

```bash
cargo clippy
    Checking num-traits v0.2.15
    Checking byteorder v1.4.3
    Checking curve25519-dalek v3.2.0
    Checking ordered-float v2.10.0
    Checking x25519-dalek v2.0.0-pre.1
    Checking metrics-util v0.14.0
    Checking hpke v0.10.0
    Checking metrics-tracing-context v0.12.0
    Checking ipa v0.1.0 (/Users/koshelev/workspace/raw-ipa)
warning: use of a disallowed method `futures::future::join_all`
   --> src/secret_sharing/replicated/malicious/additive_share.rs:294:22
    |
294 |           let result = join_all(
    |  ______________________^
295 | |             self.into_iter()
296 | |                 .map(|v| async move { v.downgrade().await.access_without_downgrade() }),
297 | |         );
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods
    = note: `#[warn(clippy::disallowed_methods)]` on by default
```

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,4 +1,4 @@
 disallowed-methods = [
   { path = "futures::future::join_all", reason = "We don't have a replacement for this method yet. Consider extending `SeqJoin` trait." },
-  { path = "futures::future::try_join_all", reason = "Use Context.join instead." },
+  { path = "futures::future::try_join_all", reason = "Use Context.try_join instead." },
 ]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,4 +1,4 @@
 disallowed-methods = [
-  "futures::future::join_all",
-  "futures::future::try_join_all",
+  { path = "futures::future::join_all", reason = "We don't have a replacement for this method yet. Consider extending `SeqJoin` trait." },
+  { path = "futures::future::try_join_all", reason = "Use Context.join instead." },
 ]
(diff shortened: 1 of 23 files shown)


.clippy.toml
```diff
@@ -1,4 +1,4 @@
 disallowed-methods = [
-  "futures::future::join_all",
-  "futures::future::try_join_all",
+  { path = "futures::future::join_all", reason = "We don't have a replacement for this method yet. Consider extending `SeqJoin` trait." },
+  { path = "futures::future::try_join_all", reason = "Use Context.try_join instead." },
 ]
```
andyleiserson, line 4: It might be worth a note here that there is a blanket waiver of `disallowed-methods` for tests, just in case somebody in the future intends to add a method to the list that they truly want to be disallowed everywhere.

andyleiserson (approved):

PR #587, opened 2023-04-08
restore MpcHelperClient and partial implementation of HttpTransport

andyleiserson:
First part of these changes. I have an additional set of changes (bigger and still in progress) to restore `MpcHelperServer`.

The original version of this pull request removed support for dispatching `ReceiveQuery` events via the transport, however it was restored after discussion. The following text from the original description referred to the removal of `ReceiveQuery`:
> My thinking is that `PrepareQuery` and `Records`, i.e., the helper-to-helper messages, go via transport. The other requests are made by external clients. We could plumb them through the transport, but I don't think that's necessary -- we can also dispatch them directly to the local `HelperApp`. Somewhat related to this is the initialization order issues between transport, server, and processor, that are not apparent in this PR because it's not dealing with the server yet. In the case where the HTTP server receives step data, it wants to hand that off to the transport. This means the server wants to have a reference to the transport, which I resolved by moving `MpcHelperServer` out of the transport.
>
> Other perspectives on the role of the transport:
> * We could in theory spin up a test setup consisting of 3 helpers on one machine, interconnected with InMemoryNetwork, but serving the create/input/results APIs over HTTP.
> * For the external client messages, the HTTP server provides an optional facade. If we want these operations to dispatch via transport, then we should arrange to have them handled similarly in tests and production. In particular, the `start_query` and `execute_query` methods on `HelperApp` should dispatch via the transport rather than directly to the query processor.


akoshelev:
> I have removed support for dispatching ReceiveQuery events via the transport. My thinking is that PrepareQuery and Records, i.e., the helper-to-helper messages, go via transport. The other requests are made by external clients. We could plumb them through the transport, but I don't think that's necessary -- we can also dispatch them directly to the local HelperApp. 

When we were thinking about the role of `Transport` (#352), we imagined it being the communication gateway into helper app. Lambda-like environments (CF workers that we may need to support one day) often provide their own communication layer that is different from HTTP. So we wanted to have a single entry point that helper can use to communicate with other helpers and clients, so we wouldn't have to re-implement the other parts of IPA

With that approach, unless there is a better alternative, client (RC) commands need to be routed through the transport as well

src/net/transport.rs
```diff
@@ -30,115 +36,134 @@ impl HttpTransport {
...
+                    })?;
+                tokio::spawn(async move {
+                    resp_future
+                        .map_err(Into::into)
+                        .and_then(MpcHelperClient::resp_ok)
+                        .await
+                        .expect("failed to stream records");
+                });
+                // TODO: error handling here needs to be improved.
+                //  * We are supposed to wait until the stream has been successfully opened with the remote
```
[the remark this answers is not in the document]
andyleiserson: I was thinking "100 continue", but it looks like hyper's HTTP2 implementation doesn't support that yet. If the server sends a regular response on seeing just the request headers, is it able to change the status after seeing the body?

src/helpers/transport/mod.rs
```diff
@@ -15,12 +15,12 @@ pub use bytearrstream::{AlignedByteArrStream, ByteArrStream};
...
 
 pub trait ResourceIdentifier: Sized {}
-pub trait QueryIdBinding: Sized
+pub trait QueryIdBinding: Send + Sized
 where
     Option<QueryId>: From<Self>,
 {
 }
-pub trait StepBinding: Sized
+pub trait StepBinding: Send + Sync + Sized
```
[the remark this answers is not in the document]
andyleiserson: I was thinking there's not a lot of harm in POD types being Send + Sync, but as you point out, it's easy to avoid the need entirely, so I'll do that.

src/bin/helper.rs
```diff
@@ -22,40 +31,58 @@ struct Args {
     scheme: Scheme,
 }
 
-fn main() {
-    unimplemented!();
+fn config() -> (NetworkConfig, ServerConfig) {
```
martinthomson, line 34: Can we load this from a file instead?  Having a default value is good, but I get nervous seeing large r# strings.  (Deferring this is OK.)

andyleiserson:
> So we wanted to have a single entry point that helper can use to communicate with other helpers and clients

Total aside, but, another context where an all-encompassing I/O abstraction might come in useful is running in a TEE.

Any thoughts on whether we want helper-to-helper and client-to-helper requests to go through the same server instance? One server is always simpler than two, but one server might not work e.g. if we want to support HTTP1 for clients, or if we want to use different TLS configurations.

What use case do you have in mind for the codepath in `InMemoryTransport::listen` that processes ReceiveQuery messages? I didn't think it was getting exercised at all, other than the one unit test. Other tests access the query processor via the methods on `HelperApp`. It's not a lot of work to have plumbing for 3 or 4 request types through the transport, but I wanted to double-check that it's really something we want before investing in it.

src/net/http_serde.rs
```diff
@@ -541,49 +501,41 @@ pub mod query {
                         self.step.as_ref()
                     ))
                     .build()?;
-                let headers = Headers {
-                    offset: self.offset,
-                };
+                // TODO: this is a misuse of the origin header, and is insecure.
+                // need to authenticate clients with TLS.
```
martinthomson, line 505: Do we have an issue to track this?

src/net/transport.rs
```diff
@@ -30,115 +36,134 @@ impl HttpTransport {
     #[must_use]
     #[allow(clippy::needless_pass_by_value)] // TODO: remove when ServerConfig is used
     pub fn new(
-        id: HelperIdentity,
+        identity: HelperIdentity,
         _server_conf: ServerConfig,
         network_conf: Arc<NetworkConfig>,
+        _callbacks: TransportCallbacks<Arc<Self>>,
     ) -> Arc<Self> {
+        #[cfg(never)]
```
martinthomson, line 44: What is the story here?  Lots of cfg(never) chunks.

src/net/transport.rs
```diff
@@ -30,115 +36,134 @@ impl HttpTransport {
...
+        Option<QueryId>: From<Q>,
+        Option<Step>: From<S>,
+    {
+        match route.resource_identifier() {
+            RouteId::Records => {
+                let resp_future = self.clients[dest]
+                    .step(
+                        dest,
+                        <Option<QueryId>>::from(route.query_id()).expect("TODO"),
+                        <Option<Step>>::from(route.step()).as_ref().expect("TODO"),
```
martinthomson:
```suggestion
                        <Option<QueryId>>::from(route.query_id()).ok_or(Error::Something)?,
                        <Option<Step>>::from(route.step()).as_ref().ok_or(Error::Something)?,
```
Right?

src/net/transport.rs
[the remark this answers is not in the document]
martinthomson: I had a discussion with Alex about this and though 100 continue is a thing, I would prefer that we not use it.  For sending records, we should be able to provide an ACK in the client code that satisfies the API constraints, without waiting for the server to respond.  That means surfacing any errors later during processing instead.  Otherwise, we add a round trip unnecessarily.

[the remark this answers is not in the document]
akoshelev: Yep, agree with Martin - acking every records request (one handshake per step) is suboptimal to say the least.

I have confirmed that for `records`/`step` API we don't really need an ack even for in-memory implementation. That leaves us with the potential dualism of `send` API - some send requests require an ack and some don't.

We could potentially solve it by adding this requirement to the `RouteParams` trait, but it is also an option to have special treatment for `step`/`records` API to return right after `POST` request is sent

I am willing to keep the acks inside the in memory implementation for now

akoshelev:
> Any thoughts on whether we want helper-to-helper and client-to-helper requests to go through the same server instance? 

It is probably easier to assume that they go through the same server endpoint for now, but it also feels like it should be easy to change later and hide the fact that IPA runs more than one server behind a single HTTP transport facade.

I think it is very likely that we end up using different TLS certificates for clients and helper-to-helper comms, just that is probably not necessary for in-market test.

> What use case do you have in mind for the codepath in InMemoryTransport::listen that processes ReceiveQuery messages? 

that is probably a miss on my end - it does require a bit of plumbing to work properly, but ideally we should have an in-memory client that emulates RC behavior. Commands sent from this client should be submitted directly to in-memory transport which currently does not expose a method for that

src/net/http_serde.rs
andyleiserson: Filed #597 (I was already planning to resolve this as part of the TLS work, which is partly done, but is currently on hold pending getting the server back to working condition).

src/net/transport.rs
andyleiserson: These will be cleaned up in the next PR that restores the server.

(Once we've dealt with the known work items, we definitely need to do a pass to scrub any leftover `cfg(never)`s.

src/net/transport.rs
andyleiserson: Certainly `expect("TODO")` is not a keeper. I do think this could be written as an assertion -- this is not processing external requests, it's called by our own code, and we shouldn't write code that tries to send records without an associated query or step ID. A better possibility is changing the API to use an enum like [`TransportCommand`](https://github.com/andyleiserson/ipa/blob/aeaf2cc25c07d63cf3acc0daeab6be570c74cab2/src/helpers/transport/mod.rs#L22) so we can infallibly extract fields that are required for a given request type.

andyleiserson: Or another possibility is splitting `Transport::send` into multiple functions (`send_record_stream`, `prepare_query`, etc.). If the expectation (or not) of an immediate acknowledgement is implied by the request type, using multiple functions could help keep track of that.

andyleiserson:
I restored the `ReceiveQuery` handling and filed a few issues (#596, #597, #600) for remaining cleanup items.

src/net/client/mod.rs
```diff
@@ -148,19 +160,18 @@ impl MpcHelperClient {
     /// If the request has illegal arguments, or fails to deliver to helper
     /// # Panics
     /// If messages size > max u32 (unlikely)
-    pub async fn step(
+    #[allow(clippy::unused_async)] // TODO: will need to be async if we want to wait for server acknowledgement
```
akoshelev, line 163: probably no longer need to be async after our discussion

akoshelev (approved):

src/net/transport.rs
```diff
@@ -30,115 +36,146 @@ impl HttpTransport {
...
+                    })
+            }
+            RouteId::ReceiveQuery => {
+                // TODO(600): Can we eliminate cases for client requests from this function
+                // entirely?
+                Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "can't send ReceiveQuery to another helper",
+                )
+                .into())
```
martinthomson, line 162: An `unreachable!()` should do here.

src/net/transport.rs
```diff
@@ -30,115 +36,146 @@ impl HttpTransport {
...
+                tokio::spawn(async move {
+                    resp_future
+                        .map_err(Into::into)
+                        .and_then(MpcHelperClient::resp_ok)
+                        .await
+                        .expect("failed to stream records");
+                });
+                // TODO(600): We need to do something better than panic if there is an error sending the
+                // data. Note, also, that the caller of this function (`GatewayBase::get_sender`)
+                // currently panics on errors.
```
martinthomson, line 142: A panic isn't good, but we can (and should) collect all of these tasks somewhere and poll them alongside the other work on the task.  Spawning them is probably fine, but maybe we can track them in a `FuturesUnordered` somewhere here that we can periodically on the server end (probably in the task we use to consume input, which is the one we'll use to return the answer to the report collector ?).

src/net/transport.rs
andyleiserson: In that case, do you also want to change the the formerly `expect("TODO")` cases to panic if somebody tries to send records (step data) without specifying query or step ID?

src/net/transport.rs
andyleiserson: I would propose to come back to this later (when we deal with #600), unless you really want to see it addressed in this PR?

akoshelev:
merging

PR #589, opened 2023-04-12
Fix the DP wrapping-add attack

taikiy:
This PR fixes the issue #520.

Opened a new issue #590 to address the cost increase - Find an alternative fix for issue #520.


akoshelev (changes requested):
all good, we should consider returning an error if cap > PRIME / 2 to avoid server crash on bad client inputs.

src/protocol/ipa/mod.rs
```diff
@@ -794,6 +794,73 @@ pub mod tests {
...
+        //   * the credit is attributed to at most one breakdown key.
+        let mut count = 0;
+        for x in &result {
+            assert!(x.trigger_value.as_u128() <= u128::from(PER_USER_CAP));
+
+            if x.trigger_value.as_u128() > 0 {
+                count += 1;
+            }
+        }
+        assert_eq!(count, 1);
```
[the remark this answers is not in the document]
taikiy: Much cleaner. Thanks!

src/protocol/attribution/credit_capping.rs
```diff
@@ -43,8 +43,11 @@ where
     }
     let input_len = input.len();
 
-    // The cap must be 1/2 of the prime number to avoid overflow.
-    assert!(F::PRIME.into() > (cap * 2).into());
+    if (u128::from(cap) * 2) >= F::PRIME.into() {
```
taikiy, line 46: Cast `cap:u32` to u128 first and then multiply by 2 to avoid overflow.

src/protocol/attribution/credit_capping.rs
```diff
@@ -23,6 +23,9 @@ use std::iter::{repeat, zip};
 /// ## Errors
 /// Fails if the multiplication protocol fails.
 ///
+/// ## Panics
```
benjaminsavage, line 26: I'm not sure if we should panic *here*. This seems like something that we would ideally validate at the very beginning of the protocol, not delay until after we have done a bunch of work (like sorting).

src/protocol/attribution/credit_capping.rs
```diff
@@ -196,6 +210,36 @@ where
     .await
 }
 
+async fn report_level_capping<F, C, T>(
```
benjaminsavage, line 213: I would really appreciate a comment that explains how this fixes the root issue we are concerned about. I don't understand it right now.

src/protocol/attribution/credit_capping.rs
```diff
@@ -40,6 +40,12 @@ where
     }
     let input_len = input.len();
 
+    if (u128::from(cap) * 2) >= F::PRIME.into() {
+        return Err(crate::error::Error::InvalidQueryParameter(format!(
+            "The cap {cap} must be less than 1/2 of the prime number to avoid overflow.",
```
benjaminsavage, line 45: Technically, this does not prevent overflow. We can still have it, but it becomes detectable. Perhaps:

```
The cap {cap} must be less than 1/2 of the prime modulus to make overflow detectable, and propagable.
```

src/protocol/attribution/credit_capping.rs
```diff
@@ -48,30 +54,41 @@ where
...
-    // 4. Compute the `final_credit`
+    // Step 5. Compute the reversed prefix-OR of `exceeds_cap_bits`
+    //
+    let prefix_or_exceeds_cap_bits =
+        comp_bits_prefix_or(ctx.clone(), input, exceeds_cap_bits).await?;
+
+    //
+    // Step 6. Compute the `final_credit`
     //
     // We compute capped credits in the method, and writes to `original_credits`.
```
benjaminsavage, line 84: I think the comment is out of date.

src/protocol/attribution/credit_capping.rs
benjaminsavage: This comment is good. I think it would probably be better above:

```
pub async fn credit_capping<F, C, T>(
```

src/protocol/attribution/credit_capping.rs
```diff
@@ -255,6 +322,35 @@ where
         .await
 }
 
+/// Returns a reversed prefix-OR'ed vector of `exceeds_cap_bits`.
+///
+/// This step ensures that once the comparison "credit > cap" is true, then the true value
+/// will be propagated to all the rows that precede the row with the true value. The next
+/// step `compute_final_credits` will then check these bits and set the credit to zero.
+async fn comp_bits_prefix_or<F, C, T>(
```
benjaminsavage, line 330: I don't love this name. What is "comb" short for?

benjaminsavage: Could we call it something like "propagate_overflow_detection"?

src/protocol/ipa/mod.rs
```diff
@@ -1034,6 +1034,69 @@ pub mod tests {
...
+        let trigger_values = result
+            .into_iter()
+            .map(|x| x.trigger_value.as_u128())
+            .collect::<Vec<_>>();
+        assert_eq!(MAX_BREAKDOWN_KEY as usize, trigger_values.len());
+        println!("actual results: {trigger_values:#?}");
+
+        // Check that
+        //   * the contribution never exceeds the cap.
+        //   * the credit is attributed to at most one breakdown key.
```
benjaminsavage, line 1093: I don't understand this part. Why would we need to require this? And where is the code that guarantees this?

src/protocol/ipa/mod.rs
```diff
@@ -1089,13 +1152,13 @@ pub mod tests {
         const NUM_MULTI_BITS: u32 = 3;
         const FIELD_SIZE: u64 = <Fp32BitPrime as Serializable>::Size::U64;
 
-        // empirical value as of Mar 30, 2023.
-        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 18_234;
-        const BYTES_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 63_648;
+        /// empirical value as of Apr 13, 2023.
+        const RECORDS_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 21_936;
```
benjaminsavage, line 1156: 20% increase =(

src/protocol/ipa/mod.rs
```diff
@@ -1089,13 +1152,13 @@ pub mod tests {
...
+        const BYTES_SENT_SEMI_HONEST_BASELINE_CAP_3: u64 = 78_456;
```
benjaminsavage, line 1157: 23% increase =(

src/protocol/ipa/mod.rs
```diff
@@ -1089,13 +1152,13 @@ pub mod tests {
...
 
-        // empirical value as of Mar 30, 2023.
-        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 46_020;
-        const BYTES_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 174_792;
+        /// empirical value as of Apr 13, 2023.
+        const RECORDS_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 55_440;
+        const BYTES_SENT_MALICIOUS_BASELINE_CAP_3: u64 = 212_472;
```
benjaminsavage, line 1161: 21.6% increase =(

Who knew fixing a bug would be so expensive =(

src/protocol/attribution/mod.rs
```diff
@@ -69,6 +69,7 @@ pub async fn prefix_or_binary_tree_style<F, C, S>(
     ctx: C,
     stop_bits: &[S],
     uncapped_credits: &[S],
+    should_add_on_first_iteration: bool,
```
benjaminsavage, line 72: OK, I think I understand this now, but it's poorly documented (my fault, not yours). I think we should address this with a comment.

```
/// If the caller has foreknowledge that there will never be any two adjacent rows, *both* containing a 1, then it is same to pass `true`, which will simply add values on the first iteration (thereby saving one multiplication per row). If the caller does not know of any such guarantee, `false` should be passed.

benjaminsavage (approved):
Thanks for fixing this bug @taikiy!

PR #591, opened 2023-04-12
Reduce the number of parameters to the ipa protocol

taikiy:


martinthomson (commented):
What about the bench ?

src/helpers/transport/query.rs
```diff
@@ -190,14 +190,31 @@ pub struct IpaQueryConfig {
 impl Default for IpaQueryConfig {
     fn default() -> Self {
         Self {
-            per_user_credit_cap: 1,
+            per_user_credit_cap: 3,
```
taikiy, line 193: This is only used for the query processor test. Nonetheless, 1 is a special case and skips many attribution steps. Just to be on the safe side, set it to something other than 1.

akoshelev (approved):

PR #594, opened 2023-04-12
Add security mode parameter to IPA benchmark

akoshelev:
Makes it easier to choose semi-honest vs malicious IPA


akoshelev:
> The conflict here requires some thinking I don't have a strong opinion on whether this should be moved inside `IpaConfig` or not, though I might prefer moving it inside.

yea I decided to add another parameter, relatively easy to do now

PR #601, opened 2023-04-13
Fix the memory spike

martinthomson:
We're still far too exposed to memory allocations, but this should fix the spike we're seeing about 2/3 - 3/4 through a run.

This is safe because the comparison ends with sequential operations, even if earlier operations (RBG) are not.

IPA for 50000 records took 568.861833105s compared to 843.574458344s without this change (same baseline; not rebased).


akoshelev (approved):
nice! backpressure FTW!

PR #602, opened 2023-04-14
Make it run on multiple threads

martinthomson:
This doesn't work.  The work moves between threads in the pool, but it completely fails to use any of them concurrently.  I couldn't work out where the bottleneck was.  This might at least provide a baseline for further tests though.


benches/oneshot/ipa.rs
```diff
@@ -134,3 +139,15 @@ async fn main() -> Result<(), Error> {
...
+fn main() -> Result<(), Error> {
+    let args = Args::parse();
+    let rt = Builder::new_multi_thread()
+        .worker_threads(args.threads)
+        .enable_all()
+        .build()
+        .unwrap();
+    let _guard = rt.enter();
+    let task = rt.spawn(run(args));
+    rt.block_on(task)?
```
martinthomson, line 152: Note that calling `block_on` here will ensure that the join task runs on a single thread, but the [documentation for `spawn`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.spawn) claims that the work will be moved to different threads as necessary, so that won't be the issue.

martinthomson:
My current theory is that we need more calls to [`spawn`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.spawn) in different places, so that the runtime polls more futures.  I was not successful in spawning off more tasks though.

akoshelev:
> My current theory is that we need more calls to `spawn`

same feeling here. We need to make child tasks run in parallel with the parent, currently we're just polling one large future by a single thread. I've done some experiments and it seems there is a deadlock somewhere, even if I add parallelism to `send` and `receive` (with #603 in place, so it is not related to panics inside `seq_join`). 

I think even if I can make it work, it won't be enough as all the expensive stuff (AES and friends) will be happening in one thread. Probably the parallelism need to be at `multiply` or `seq_join` item level.

akoshelev (approved):
We are likely going to see a perf hit with this change. As @martinthomson noticed, this will not force Tokio to drive futures in parallel, so there still going to be only one thread driving IPA protocol, but instead of it being always **the same**, we would use `Args::threads`. Bouncing large task between threads come with a cost - allocations made inside one thread are more expensive to access from another.

Maybe we can set `threads` to 1 for now until we have proper parallelism in place.

martinthomson (commented):
@akoshelev I know that you approved this, but I forgot to land it and it rotted.

Tracing changes mean that this is a bit more involved now.  Can you take another look?

akoshelev (approved):
all good, I had to resolve some conflicts but I think it should be good to go

PR #603, opened 2023-04-16
Let seq_join resolve futures out of order

akoshelev:
As discussed in 582 it is beneficial to let `seq_join` to operate on list of futures that may resolve out of order.

It is also a blocker if we want to increase the parallelism in the system and run IPA on multiple threads - whenever we call `tokio::spawn` we should expect tasks to finish out of order they were spawned


src/seq_join.rs
```diff
@@ -177,10 +220,11 @@ where
 mod test {
     use crate::seq_join::{seq_join, seq_try_join_all};
     use futures::{
-        future::{lazy, pending, BoxFuture},
+        future::{lazy, BoxFuture},
         Future, StreamExt,
     };
-    use std::{convert::Infallible, iter::once, num::NonZeroUsize};
+    use futures_util::stream::poll_immediate;
```
martinthomson, line 226: This is in `futures::stream` too.  Is there something we needed `futures_util` for?  It seems like your IDE keeps picking it up over the canonical version.

akoshelev: yea I guess it is because of order it appears in the auto-import window :(

src/seq_join.rs
```diff
@@ -117,14 +117,55 @@ where
...
+    Done,
+}
+
+impl<F: IntoFuture> ActiveItem<F> {
+    /// Drives this item to resolved state when value is ready to be taken out. Has no effect
+    /// if the value is ready.
+    ///
+    /// ## Panics
+    /// Panics if polled after completion.
+    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<()> {
```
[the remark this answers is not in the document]
akoshelev: yea `#[allow(unused_must_use)]` is ugly. I still like `Poll` interface here because it better represents the status of this operation, but `#[must_use]` on `Poll` kills it. 

Changed it to return `bool` instead.

src/seq_join.rs
```diff
@@ -117,14 +117,55 @@ where
...
+impl<F: IntoFuture> ActiveItem<F> {
+    /// Drives this item to resolved state when value is ready to be taken out. Has no effect
+    /// if the value is ready.
+    ///
+    /// ## Panics
+    /// Panics if this item is completed
+    fn check_ready(&mut self, cx: &mut Context<'_>) -> bool {
+        let mut ready = true;
+        match self {
+            ActiveItem::Pending(f) => {
```
martinthomson, line 134:
```rust
if let ActiveItem::Pending(f) = self {
   ...
}
```

or

```rust
let ActiveItem::Pending(f) = self else { return true; };
if let Poll::Ready(v) = Future::poll(Pin::as_mut(f), cx) {
    *self = ActiveItem::Resolved(v);
    true
} else {
    false
}
```

src/seq_join.rs
```diff
@@ -117,14 +117,55 @@ where
...
+    ///
+    /// ## Panics
+    /// If the value is not ready yet.
+    #[must_use]
+    fn take(self) -> F::Output {
+        if let ActiveItem::Resolved(v) = self {
+            v
+        } else {
+            panic!("No value to take out")
+        }
```
martinthomson, line 157: Is this better?  I don't know any more.

```suggestion
        let ActiveItem::Resolved(v) = self else {
            panic!("No value to take out");
        };
        v
```

martinthomson (approved):

src/seq_join.rs
akoshelev: 4 < 5 so it is strictly better

PR #604, opened 2023-04-19
Enable attribution window in the random tests

taikiy:


src/test_fixture/ipa.rs
```diff
@@ -72,29 +72,47 @@ pub fn generate_random_user_records_in_reverse_chronological_order(
...
+                Some(0)
+            }
+        })
+        .collect::<Vec<_>>();
+
+    // reverse the records so we can iterate from the most recent to the oldest
+    let mut reversed_records = records_for_user.to_vec();
+    reversed_records.reverse();
+
     let mut pending_trigger_value = 0;
```
benjaminsavage, line 107: I don't want to do it this way. I do not like the way that this approach duplicates the logic of searching backwards to find the last-touch. I forsee difficulties when we attempt to extend this to multi-touch attribution, because it will force us to make changes in two places.

I would rather keep it as just a single pass over the data (not two), with no need to reverse the list.

I recommend the following:

1. Replace `pending_trigger_value` with a Vec of `pending_trigger_reports`, and just push trigger reports into it.
2. When encountering a source event, instead of checking if `pending_trigger_value > 0` just check if `pending_trigger_reports.len() > 0`
3. Then, iterate through each of the `pending_trigger_reports`, (for which you'll now have individual timestamps) and compute the delta to this particular source event. If it's longer than the window, just continue. Otherwise, do the current logic.

taikiy: Updated!

src/test_fixture/ipa.rs
```diff
@@ -85,24 +85,51 @@ pub fn update_expected_output_for_user(
     records_for_user: &[TestRawDataRecord],
     expected_results: &mut [u32],
     per_user_cap: u32,
-    _attribution_window_seconds: u32, // TODO(taikiy): compute the output with the attribution window
+    attribution_window_seconds: u32,
 ) {
-    let mut pending_trigger_value = 0;
+    // Calculate time deltas from the nearest source report to the current trigger report.
```
benjaminsavage, line 90: Can you remove the commenter rows?

benjaminsavage (approved):
Awesome! And I’m glad to see that the tests still pass! A good indication that the windowing code functions as intended :)

PR #609, opened 2023-04-21
Fix sequential prss metrics in IPA

akoshelev:
I realized that we no longer print/emit sequential randomness metrics

```bash
RUST_LOG=ipa=DEBUG cargo bench --bench oneshot_ipa --no-default-features --features="enable-benches debug-trace" -- -n 10 | cut -d',' -f5 | sort -nr | head -n10
Sequential PRSS
0
0
0
0
0
0
0
0
0
```

after some debugging, it seems that shuffle now calls `next_u32` method instead of `next_u64`. I think it happened because we switched from `usize` to `u32` in `get_two_of_three_random_permutations`. Static dispatch does not call instrumented `next_u64` and here we are.

Unit tests did not catch that because they use a different protocol that never hit that path. Now it is changed to generate `u32` values too.

It also feels bad that we don't test PRSS metrics for IPA, so I changed the `communication_baseline` test. That required some fiddling with test infrastructure because Sequential PRSS calls are not deterministic: there is a loop inside `gen_range` method that calls `next_u32` several times depending on initial seed value. So baseline tests must use the same seed to correctly estimate the calls to sequential PRSS.

I also got annoyed by the size of baseline test, so I split it in two.

### DIFF
--- a/src/protocol/context/mod.rs
+++ b/src/protocol/context/mod.rs
@@ -81,6 +81,7 @@ mod tests {
         telemetry::metrics::{
             BYTES_SENT, INDEXED_PRSS_GENERATED, RECORDS_SENT, SEQUENTIAL_PRSS_GENERATED,
         },
+        test_fixture::{Reconstruct, Runner, TestWorld, TestWorldConfig},
     };
     use futures_util::{future::join_all, try_join};
     use rand::{
@@ -91,7 +92,6 @@ mod tests {
     use typenum::Unsigned;
 
     use super::*;
-    use crate::test_fixture::{Reconstruct, Runner, TestWorld, TestWorldConfig};
 
     trait AsReplicatedTestOnly<F: Field> {
         fn l(&self) -> F;
@@ -141,7 +141,10 @@ mod tests {
         let (seq_l, seq_r) = {
             let ctx = ctx.narrow(&format!("seq-prss-{index}"));
             let (mut left_rng, mut right_rng) = ctx.prss_rng();
-            (left_rng.gen::<F>(), right_rng.gen::<F>())
+            (
+                left_rng.gen::<F>() + F::truncate_from(left_rng.gen::<u32>()),
+                right_rng.gen::<F>() + F::truncate_from(right_rng.gen::<u32>()),
+            )
         };
 
         let send_channel = ctx.send_channel(left_peer);
@@ -202,17 +205,18 @@ mod tests {
             .total(3 * input_size * field_size)
             .per_step(&metrics_step, 3 * input_size * field_size);
 
-        // each helper generates 2 128 bit values resuling in 4 calls to rng::<gen>() per input row
+        // each helper generates 2 128 bit values and 2 u32 values
+        // resulting in 6 calls to rng::<gen>() per input row
         let seq_prss_assert = snapshot
             .assert_metric(SEQUENTIAL_PRSS_GENERATED)
-            .total(4 * 3 * input_size)
-            .per_step(&metrics_step.narrow("seq-prss-0"), 4 * 3);
+            .total(6 * 3 * input_size)
+            .per_step(&metrics_step.narrow("seq-prss-0"), 6 * 3);
 
         for role in Role::all() {
             records_sent_assert.per_helper(role, input_size);
             bytes_sent_assert.per_helper(role, field_size * input_size);
             indexed_prss_assert.per_helper(role, input_size);
-            seq_prss_assert.per_helper(role, 4 * input_size);
+            seq_prss_assert.per_helper(role, 6 * input_size);
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/protocol/context/prss.rs
```diff
@@ -62,8 +62,9 @@ impl<'a> InstrumentedSequentialSharedRandomness<'a> {
 }
 
 impl RngCore for InstrumentedSequentialSharedRandomness<'_> {
+    #[allow(clippy::cast_possible_truncation)]
     fn next_u32(&mut self) -> u32 {
-        self.inner.next_u32()
+        self.next_u64() as u32
```
akoshelev, line 67: this is the actual fix - route rng through instrumented `next_u64` call

src/protocol/context/mod.rs
```diff
@@ -141,7 +141,10 @@ mod tests {
         let (seq_l, seq_r) = {
             let ctx = ctx.narrow(&format!("seq-prss-{index}"));
             let (mut left_rng, mut right_rng) = ctx.prss_rng();
-            (left_rng.gen::<F>(), right_rng.gen::<F>())
+            (
+                left_rng.gen::<F>() + F::truncate_from(left_rng.gen::<u32>()),
```
martinthomson, line 145: Please comment here about what this is doing.

src/protocol/ipa/mod.rs
```diff
@@ -1137,148 +1139,158 @@ pub mod tests {
...
-                        IpaQueryConfig::new(
-                            per_user_cap,
-                            MAX_BREAKDOWN_KEY,
-                            ATTRIBUTION_WINDOW_SECONDS,
-                            NUM_MULTI_BITS,
-                        ),
-                    )
+        #[tokio::test]
+        async fn semi_honest() {
+            // empirical values per user cap as of Apr 20, 2023.
```
martinthomson, line 1158: More comments here might be necessary to explain what the different values are.

src/protocol/ipa/mod.rs
```diff
@@ -1137,148 +1139,158 @@ pub mod tests {
...
+        #[tokio::test]
+        async fn malicious() {
+            // empirical values per user cap as of Apr 20, 2023.
+            let records_sent: HashMap<u32, u64> = HashMap::from([(1, 36_714), (3, 55_440)]);
+            let bytes_sent: HashMap<u32, u64> = HashMap::from([(1, 137_568), (3, 212_472)]);
+            let indexed_prss: HashMap<u32, u64> = HashMap::from([(1, 76_014), (3, 113_337)]);
+            let seq_prss: HashMap<u32, u64> = HashMap::from([(1, 1108), (3, 1108)]);
+
+            let records = generate_input();
+            for per_user_cap in [1, 3] {
```
martinthomson, line 1218: It might be better to have different tests for the different per-user caps rather than use this construct with `HashMap<u32, u64>` as you do here.  Then you can have a simple expected counts structure.  A function like this would be lovely:

```rust
fn malicious_cap1() {
  malicious(1, ExpectedMetrics {
    records_sent: 36_714,
    bytes_sent: 137_568,
    indexed_prss: 76_014,
    sequential_prss: 1108,
  });
}
```

src/protocol/sort/shuffle.rs
```diff
@@ -167,17 +167,17 @@ pub async fn unshuffle_shares<F: Field, S: SecretSharing<F> + Reshare<C, RecordI
 
 #[cfg(all(test, not(feature = "shuttle")))]
 mod tests {
-
     use crate::{
         protocol::{sort::shuffle::get_two_of_three_random_permutations, Step},
         test_fixture::{make_participants, permutation_valid},
     };
+    use rand::thread_rng;
```
martinthomson, line 174: `crate::rand::thread_rng` ?  Only for consistency reasons.

src/test_fixture/mod.rs
```diff
@@ -43,11 +43,11 @@ pub fn narrow_contexts<C: Context>(contexts: &[C; 3], step: &impl Substep) -> [C
 /// Generate three participants.
 /// p1 is left of p2, p2 is left of p3, p3 is left of p1...
 #[must_use]
-pub fn make_participants() -> [PrssEndpoint; 3] {
-    let mut r = thread_rng();
-    let setup1 = PrssEndpoint::prepare(&mut r);
-    let setup2 = PrssEndpoint::prepare(&mut r);
-    let setup3 = PrssEndpoint::prepare(&mut r);
+pub fn make_participants<R: RngCore + CryptoRng>(r: &mut R) -> [PrssEndpoint; 3] {
+    // let mut r = thread_rng();
```
martinthomson, line 47:
```suggestion
```

src/protocol/prss/mod.rs
```diff
@@ -327,7 +327,7 @@ pub mod test {
     #[test]
     fn three_party_values() {
         const IDX: u128 = 7;
-        let [p1, p2, p3] = make_participants();
+        let [p1, p2, p3] = make_participants(&mut thread_rng());
```
martinthomson, line 330: It might be easier here to wrap this function:

```rust
fn participants() -> [Participant; 3] {
  make_participants(&mut thread_rng())
}
```

akoshelev:
addressed feedback

martinthomson (approved):
Greatly improved! Thanks.

PR #614, opened 2023-04-24
restoring MpcHelperServer

andyleiserson:
There are still a few cleanup items remaining, but this is almost ready. The outstanding work is:

- [x] Restoring tests to working condition
- [ ] Cleaning up error handling (in particular, handling of errors reading step data -- i.e. `HttpRecordsStream`, `unwrap_body_chunks`, and the mapping functionality in `ReceiveRecords` those use)


src/helpers/transport/callbacks.rs
```diff
@@ -1,43 +1,68 @@
...
+callbacks! {
+    /// Called when helper receives a new query request from an external party.
+    (ReceiveQueryCallback, ReceiveQueryResult):
+        async fn(T, QueryConfig) -> Result<QueryId, NewQueryError>;
+
+    /// Called when helper receives a request from the coordinator to start working on a new query.
+    (PrepareQueryCallback, PrepareQueryResult):
+        async fn(T, PrepareQuery) -> Result<(), PrepareQueryError>;
+
+    (QueryInputCallback, QueryInputResult):
```
andyleiserson, line 53: This should have a doc comment as well.

src/helpers/transport/receive.rs
andyleiserson, line 1: This moved from test-fixture, but also changed to support converting stream items from `Result<Bytes, Error>` to `Vec<u8>` for the HTTP case. I want to come back to this and work on something cleaner.

PR #617, opened 2023-04-27
Various metrics improvements

akoshelev:
See commit messages for details

### DIFF
--- a/src/test_fixture/logging.rs
+++ b/src/test_fixture/logging.rs
@@ -1,16 +1,29 @@
 use metrics_tracing_context::MetricsLayer;
-use std::sync::Once;
+use std::{str::FromStr, sync::Once};
 use tracing::Level;
-use tracing_subscriber::{fmt, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
+use tracing_subscriber::{
+    filter::Directive, fmt, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter,
+};
 
+/// Set up logging for IPA
+///
+/// ## Panics
+/// Does not, but compiler cannot be convinced otherwise
 pub fn setup() {
     static INIT: Once = Once::new();
 
     INIT.call_once(|| {
+        let default_directive = if let Some(crate_name) = option_env!("CARGO_CRATE_NAME") {
+            // only print IPA crate logging by default
+            Directive::from_str(&format!("{crate_name}=INFO")).unwrap()
+        } else {
+            Level::INFO.into()
+        };
+
         tracing_subscriber::registry()
             .with(
                 EnvFilter::builder()
-                    .with_default_directive(Level::INFO.into())
+                    .with_default_directive(default_directive)
                     .from_env_lossy(),
             )
             .with(fmt::layer())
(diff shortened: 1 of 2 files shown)


src/test_fixture/metrics.rs
```diff
@@ -57,7 +57,11 @@ impl MetricsHandle {
         // logging is required to import span fields as metric values
         logging::setup();
 
-        let id = thread_rng().gen::<u128>();
+        let id = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(8)
```
martinthomson, line 62: Is 40 bits enough to avoid collisions?  I guess it is.

src/test_fixture/logging.rs
```diff
@@ -1,16 +1,29 @@
...
+///
+/// ## Panics
+/// Does not, but compiler cannot be convinced otherwise
 pub fn setup() {
     static INIT: Once = Once::new();
 
     INIT.call_once(|| {
+        let default_directive = if let Some(crate_name) = option_env!("CARGO_CRATE_NAME") {
+            // only print IPA crate logging by default
+            Directive::from_str(&format!("{crate_name}=INFO")).unwrap()
```
martinthomson, line 18: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#method.with_target seems like a better way to get a directive without having to build and parse a string.

martinthomson (approved):
All good additions.

src/test_fixture/metrics.rs
akoshelev: I was thinking about this id as a string where each character can have at most 62 different values.  We could have $62^8$ possible combinations and to get 50% chance of collision we need to run ~14M unit tests in parallel.

martinthomson: OK, that's probably OK - collisions aren't necessarily fatal either, they just mix things together.

src/test_fixture/logging.rs
akoshelev: Unfortunately `with_target` statically configures the logger and does not allow `RUST_LOG` to override it. [`with_default` ](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#method.with_default) looks almost what we would want but it only takes the level argument which would again print Shuttle output

martinthomson: Oh, I see what is going on now.  I am referring to a `Filter`, which is not a `Directive`.  A `Directive` is not a real filter, it's an intermediate format.  I don't like how that crate works now.  I shouldn't have looked at the code.

akoshelev: yea, I wasn't too concerned about it because this crate provides the right abstraction for layers and filters, so we can implement our own layer that does the filtering the way we like

src/test_fixture/metrics.rs
akoshelev: yea, collisions are "fatal" if metrics test collide with any other test - that will make the metrics test fail. I'll keep an eye on them, any id string that has less than 31 characters in it will be a clear win compared to u128. I thought that using 64 bit identifiers should be fine too, that gets us to 4B unit tests running in parallel. If we get to that number, I will happily retire (and change it back to u128)

PR #618, opened 2023-04-28
Introducing `StepNarrow` trait

taikiy:
Taking a baby step with the `Step` optimization.

New: `trait StepNarrow` which defines `fn narrow(&self, step: &S) -> Self;`
Rename:
* `struct Step` -> `struct GenericStep`
* `trait SubStep` -> `trait Step`

Upcoming next is a new `trait Gate` and make `Context` generic over `G: Gate`.


src/protocol/step.rs
```diff
@@ -0,0 +1,6 @@
+use super::Step;
```
martinthomson, line 1: Want to move all the step related structs/traits here?

src/test_fixture/world.rs
```diff
@@ -150,7 +150,7 @@ impl TestWorld {
     }
 
     #[must_use]
-    pub fn execution_step(execution: usize) -> impl Substep {
+    pub fn execution_step(execution: usize) -> impl Step {
```
martinthomson, line 153: I just realized that this is going to hurt quite a bit when we stop using `GenericStep`.  I think that we're OK with that; we can just not use this feature for non-generic steps, but @akoshelev might not realize his dream of running tests with non-generic steps as a result.  We rely on this to make steps diverse.

src/protocol/mod.rs
```diff
@@ -75,23 +77,22 @@ impl Substep for str {}
     derive(serde::Deserialize),
     serde(from = "&str")
 )]
-pub struct Step {
+pub struct GenericStep {
```
martinthomson, line 80: GenericGate?  Or does that make it too hard to deal with (comments and such).

martinthomson (approved):

src/protocol/mod.rs
akoshelev: DescriptiveStep? StringBackedStep? not sure what is the best name here and what would be the appropriate name for "efficient" step backed by an int

src/test_fixture/world.rs
akoshelev: Looked into it as well - it seems we can make it work if we consume the world to run a test (that shouldn't require narrowing)

```rust
pub trait Runner {
    async fn oneshot<I, A, O, H, R>(self, input: I, helper_fn: H) -> [O; 3];
    ...
}
```

agree with @martinthomson that we don't need this feature for all tests, likely we only need to run IPA with non generic step

akoshelev (approved):

src/protocol/step.rs
taikiy: Yes, will move them in the next PR.

src/protocol/mod.rs
taikiy: strings
* DescriptiveStep
* StringStep
* StringBackedStep
* VerboseStep
* TextualStep

integers
* IntStep
* NumericStep
* CompactStep
* MemoryEfficientStep

akoshelev: I vote for `DescriptiveStep` and `CompactStep`

PR #619, opened 2023-04-29
small MpcHelperServer cleanups

andyleiserson:


src/helpers/transport/receive.rs
andyleiserson, line 1: This change restores `ReceiveRecords` to the original version that was in `test_fixture` before #614. `LogErrors` is new.

src/helpers/transport/receive.rs
```diff
@@ -1,51 +1,74 @@
...
-    #[cfg(feature = "web-app")]
-    pub(crate) fn mapped(key: StreamKey, coll: StreamCollection<T>, f: F) -> Self {
+    type Item = Vec<u8>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match Pin::get_mut(self).inner.poll_next_unpin(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(Ok(chunk))) => Poll::Ready(Some(chunk.into())),
+            Poll::Ready(Some(Err(err))) => {
+                error!("error reading records: {err}");
```
akoshelev, line 47: I wonder if we should panic in this case. This does not seem like a recoverable error and if we just return None, then our protocols eventually hang with no clear indication what went wrong

martinthomson: Are we looking for a panic, or should this return `Poll<Option<Result<Vec<u8>>>>` (i.e., `type Item = Result<Vec<u8>>`), so that the read is fallible?  (Not here, but on ReceiveRecords.)

akoshelev: that is another option, however at some point we need to unwrap this result and it does not seem that we can recover if some part of records stream is missing. so clients are no better positioned to decide because the best they can do is `unwrap` imo

akoshelev: You're right - we could potentially make it work with panics and runtime-per-query approach where panic will just abort the the thread but it is worse than returning `Result` imo

akoshelev (approved):

src/helpers/transport/receive.rs
martinthomson: That outcome works for me, but can we comment to that effect here so that we don't lose this insight?

martinthomson (approved):

PR #621, opened 2023-05-01
Remove test-fixture dependency from CLI

akoshelev:
As @andyleiserson mentioned, we have an issue now where `web-app` and `test-fixture` features are mutually incompatible because of different `Transport` implementations. That bubbles up to the top-level feature `cli` that we use to start HTTP server and to run test circuits. 

Here is what I am proposing to do to resolve this:

* [This PR] Not every CLI needs to depend on `test-fixture`, only those that we use to test IPA. `Reconstruct`/`IntoShares` are two examples. So removing `test-fixture` dependency from `cli` feature and adding it to `test_mpc` should keep our CLI dependency tree clean.
* [Pending] `test_mpc` still needs some of the web components (client) to communicate with IPA helpers. I suggest that we add another feature `web-client` and use it to build `MpcHelperClient` struct and its friends. That feature does not need to depend on transport because it should only build client package.

After making those two changes, I believe we should no longer have issues with `cli` and `web-app`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 default = ["web-app", "tracing/max_level_trace", "tracing/release_max_level_info"]
 # TODO: it feels wrong that CLI uses test-fixture module, but untangling `IntoShares`/`Reconstruct` and inputs
 # from other test components wasn't easy.
-cli = ["comfy-table", "clap", "test-fixture"]
+cli = ["comfy-table", "clap"]
 enable-serde = ["serde", "serde_json"]
 protocol-perf-testing = ["no-prss"]
 no-prss = []
@@ -109,7 +109,7 @@ bench = false
 
 [[bin]]
 name = "test_mpc"
-required-features = ["cli", "web-app"]
+required-features = ["cli", "web-app", "test-fixture"]
 bench = false
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ bench = false
 
 [[bin]]
 name = "helper"
-required-features = ["cli", "web-app"]
+required-features = ["cli", "web-app", "real-world-infra"]
 bench = false
 
 [[bin]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,9 +6,7 @@ edition = "2021"
 
 [features]
 # by default remove all TRACE, DEBUG spans from release builds
-default = ["web-app", "tracing/max_level_trace", "tracing/release_max_level_info"]
-# TODO: it feels wrong that CLI uses test-fixture module, but untangling `IntoShares`/`Reconstruct` and inputs
-# from other test components wasn't easy.
+default = ["web-app", "in-memory-infra", "tracing/max_level_trace", "tracing/release_max_level_info"]
 cli = ["comfy-table", "clap"]
 enable-serde = ["serde", "serde_json"]
 protocol-perf-testing = ["no-prss"]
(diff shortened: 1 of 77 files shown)


andyleiserson:
> I suggest that we add another feature web-client and use it to build MpcHelperClient struct and its friends.

That makes sense to me. A related issue is that we may want to build a client that supports only the report collector APIs, and not the helper-to-helper APIs. There is a TODO about this in `src/net/client/mod.rs`:

```
/// TODO: we need a client that can be used by any system that is not aware of the internals
///       of the helper network. That means that create query and send inputs API need to be
///       separated from prepare/step data etc.
```

I'm not sure how important it is to fully separate things though. The more important things to keep in mind seem like:
1. For helper-to-helper clients, the client will need to be configured with a TLS client certificate, while for report collector clients, it will need to be configured without one.
2. If we are offering a rust client for use by external code, it would be nice to have a minimal set of dependencies.

andyleiserson (approved):

akoshelev:
> If we are offering a rust client for use by external code, it would be nice to have a minimal set of dependencies.

That's exactly what I meant in that TODO. It is important to have minimal set of dependencies, but maybe even more important to try to avoid accidental leakage of server configs/constants. 

I thought the desired configuration for IPA client is a separate crate with no dependency on `ipa` crate (maybe even compilation error if it appears). That may be a lot of work for now, because it requires shuffling things around and making sure that some structs are shared between *ipa client*  and *ipa server*, so we can delay that

akoshelev:
@andyleiserson I pushed another change to this PR that splits apart in memory and HTTP transports. It is a little clunky now but I don't see a better way unfortunately. Things that could help here include using `cfg_aliases` crate and specifying features for `cargo test` which is currently impossible :(

Cargo.toml
```diff
@@ -109,7 +108,7 @@ bench = false
 
 [[bin]]
 name = "test_mpc"
-required-features = ["cli", "web-app"]
+required-features = ["cli", "web-app", "test-fixture"]
```
akoshelev, line 111: one more change is required here - make `ipa-web-client` feature and depend on it here.

andyleiserson (approved):
This looks good to me. Remembering to pass `--no-default-features` can be a hassle, but it's a hassle I can live with, and probably worth it in order to have the HTTP tests covered in CI.

Possibly it's worth putting in a more explicit error when the conflicting features are specified? (There's a suggested method in https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features.) Conflicting definitions of `TransportImpl` will be clear enough most of the time, but a very specific error could be helpful to new developers.

martinthomson (approved):
I'll let @akoshelev decide if the `compile_error!()` thing should be included in this patch before landing.  It seems like it might be worthwhile.

akoshelev:
I can totally add it. Maybe it is time to think again about making IPA into a multi-crate repo. Features set becomes a bit harder to maintain and understand

akoshelev:
yea, it works

```bash
cargo build --no-default-features --features "in-memory-infra real-world-infra web-app"
   Compiling ipa v0.1.0 (/Users/koshelev/workspace/raw-ipa)
error[E0428]: the name `TransportImpl` is defined multiple times
  --> src/helpers/gateway/mod.rs:30:1
   |
27 | pub type TransportImpl = super::transport::InMemoryTransport;
   | ------------------------------------------------------------- previous definition of the type `TransportImpl` here
...
30 | pub type TransportImpl = crate::sync::Arc<crate::net::HttpTransport>;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `TransportImpl` redefined here
   |
   = note: `TransportImpl` must be defined only once in the type namespace of this module

error: feature "in-memory-infra" and feature "real-world-infra" cannot be enabled at the same time
  --> src/lib.rs:83:1
   |
83 | compile_error!("feature \"in-memory-infra\" and feature \"real-world-infra\" cannot be enabled at the same time");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #622, opened 2023-05-02
Eliminate differences between semi-honest and malicious protocols

martinthomson:
I'll need to do a write-up, but I thought that I might share this so that people can see where this might be headed.

protocol::sort::generate_permutation_opt::tests::malicious_sort_in_semi_honest hangs.  Otherwise, this was surprisingly quick once I worked through the baffling and extremely numerous compiler errors.  A lot less code deleted than I had hoped for, or at least the added trait constraints outweighed the removed code.  There are a few places where this is not pretty.


martinthomson:
So this should be ready now.  It's big, because changes at this layer now hit a lot of code, but I am liking the layout.

See comments on [3f4a467](https://github.com/private-attribution/ipa/pull/622/commits/3f4a467ae457fed21d097db1e1ebd65686c84e46) for details of how this is laid out.

The short answer for how this works is that you start with a base context that is effectively semi-honest.  If you instantiate `MaliciousContext`, then you do as we have done so before, except that you started with a context called something we used to only use after upgrading (there is an `UpgradedMaliciousContext` now that does the malicious protocol execution).

A `SemiHonestContext` works exactly the same way (more or less), including upgrades, but when you upgrade it, it does the same as it used to.

I need to double-check why the sequential PRSS is getting invoked a few extra times here, but otherwise, this doesn't change functionality at all.

martinthomson (commented):
Some notes for reviewers.

akoshelev:
I did a quick look and if I got the main idea, we now use "malicious" path everywhere. Which means validators get exposed in both paths, but semi-honest validate is a "no-op". I like this approach and I think this is what we discussed during Feb offsite.

src/protocol/sort/generate_permutation.rs
```diff
@@ -126,38 +97,58 @@ pub(super) async fn malicious_shuffle_and_reveal_permutation<F: Field + Extendab
...
+impl<'a, F: ExtendableField> DowngradeMalicious
+    for ShuffledPermutationWrapper<Replicated<F>, UpgradedSemiHonestContext<'a, F>>
+{
+    type Target = Vec<u32>;
+    async fn downgrade(self) -> UnauthorizedDowngradeWrapper<Self::Target> {
+        let output = self
+            .reveal(self.ctx.narrow(&RevealPermutation), NoRecord)
+            .await
+            .unwrap();
+        UnauthorizedDowngradeWrapper::new(output)
```
akoshelev, line 148: I am a bit surprised that we need to downgrade via wrapper. Isn't it just semi-honest to semi-honest conversion?

src/protocol/sort/generate_permutation.rs
```diff
@@ -44,53 +45,23 @@ pub struct ShuffledPermutationWrapper<T, C: Context> {
...
 /// 3. Validate the accumulated macs - this returns the revealed permutation
-pub(super) async fn malicious_shuffle_and_reveal_permutation<F: Field + ExtendableField>(
-    m_ctx: MaliciousContext<'_, F>,
-    input_permutation: Vec<MaliciousReplicated<F>>,
-    malicious_validator: MaliciousValidator<'_, F>,
-) -> Result<RevealedAndRandomPermutations, Error> {
+pub(super) async fn shuffle_and_reveal_permutation<C, S, F>(
+    m_ctx: C::UpgradedContext<F>,
+    input_permutation: Vec<S>,
+    malicious_validator: C::Validator<F>,
```
akoshelev, line 56: I remember we had an idea of adding/using `ctx.reveal(shares)` that validates before revealing in malicious setting and just reveals shares one by one in semi-honest.

That requires validator to be inside `UpgradedContext` and a runtime check (or another generic argument) that validator is called only once, but does not require carrying the validator along with the context.

Did you decide that this is not the best way to go?

src/test_fixture/circuit.rs
```diff
@@ -14,7 +18,11 @@ use super::join3v;
 ///
 /// # Panics
 /// panics when circuits did not produce the expected value.
-pub async fn arithmetic<F: Field>(width: u32, depth: u8) {
+pub async fn arithmetic<F>(width: u32, depth: u8)
+where
+    F: Field + IntoShares<Replicated<F>>,
+    for<'a> Replicated<F>: SecureMul<SemiHonestContext<'a>>,
```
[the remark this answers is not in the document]
akoshelev: my brain compiler goes to infinite loop on each HRTB it sees

src/test_fixture/world.rs
```diff
@@ -144,6 +143,22 @@ impl TestWorld {
             .unwrap()
     }
 
+    /// Creates malicious protocol contexts for 3 helpers
+    ///
+    /// # Panics
+    /// Panics if world has more or less than 3 gateways/participants
+    #[must_use]
+    pub fn malicious_contexts(&self) -> [MaliciousContext<'_>; 3] {
+        let execution = self.executions.fetch_add(1, Ordering::Release);
```
akoshelev, line 152: my fault, I spelled it wrong inside the similar `contexts` fn. This should be `AcqRel`

src/protocol/sort/generate_permutation.rs
martinthomson: Oh, I meant to follow up on this one and it got lost in the noise.  @richajaindce I think that this is your work.  Is this because we need to reveal and the way we decided to reveal was to do that as part of the downgrade?

src/protocol/sort/generate_permutation.rs
martinthomson: I decided not to make that change here;  I'm still not happy with the way that validators integrate into the process, but maybe we can move it into the upgraded context.  That might be a good follow-up.

src/test_fixture/circuit.rs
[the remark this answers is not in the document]
martinthomson: I still don't really understand them, but I can apply them more reliably now :)

src/test_fixture/world.rs
martinthomson: You are right.  I just copy-pasted.  Will fix.

martinthomson: Hang on.  We should be OK here with simply Relaxed.  We don't care that the value is consistent with the surrounding code, just that we get a unique value each time it is invoked.  Relaxed gives us that.

martinthomson: My read is that this is OK: https://en.cppreference.com/w/cpp/atomic/memory_order#Relaxed_ordering

That is, the action of incrementing will always occur atomically.  It isn't a load+store where two threads might load the same value; the operation is atomic.  https://stackoverflow.com/a/62613988/1375574 agrees.

akoshelev: my reasoning was based on https://doc.rust-lang.org/std/sync/atomic/index.html#portability. Basically not every ARM platform supports atomic CAS and that I believe is reflected in the doc: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU32.html#method.fetch_add - they explicitly mention ordering for store and add operations.

Neither of those matters in this case I think, so we should leave it as is. Just for actual prod code, I feel like I would still prefer `AcqRel` just to be on the safe side on every platform :) Who knows maybe Android phones will be running IPA soon!

martinthomson: I don't see anything in the portability section, other than the note that the implementation might not be performant.  But that is true for any ordering on those platforms.  That is more a question there of availability of primitives, where load+store can cost a lot more if there is contention.

My understanding was that Relaxed has no real benefit on x86, but it does on ARM if there are instructions.

andyleiserson (commented):
I'm inclined to say this is worth it for the deduplication it achieves -- there's only a few places where I'd say the bounds are truly excessive. But, I have a higher tolerance for fighting with the type system than others, and this is also the kind of thing where I can say "sure, seems fine" and then find myself banging my head against the wall over it in a few months.

andyleiserson:
I experimented with one possible way of reducing the need to write trait bounds in https://github.com/martinthomson/raw-ipa/compare/upgraded-context...andyleiserson:ipa:bounds?expand=1. (Naming obviously would need work.)

The idea is, define a trait that captures the types used in the IPA protocol (prime field, sharing of prime field, sharing of binary field, context) and the necessary bounds. Then individual functions can reference this trait and the bounds don't need to be repeated on each one. I also kind of like that when a new requirement is added, it only needs to be added in the trait definition, as opposed to on the signature of every function in the call stack to where it is used.

src/protocol/attribution/mod.rs
```diff
@@ -3,28 +3,159 @@ pub mod aggregate_credit;
...
+    SB: LinearSecretSharing<Gf2> + BasicProtocols<C::UpgradedContext<Gf2>, Gf2> + 'static,
+    Vec<SB>: DowngradeMalicious<Target = Vec<SemiHonestAdditiveShare<Gf2>>>,
+    F: PrimeField + ExtendableField,
+    BK: GaloisField,
+    ShuffledPermutationWrapper<S, C::UpgradedContext<F>>: DowngradeMalicious<Target = Vec<u32>>,
+    MCCappedCreditsWithAggregationBit<F, S>: DowngradeMalicious<
+        Target = MCCappedCreditsWithAggregationBit<F, SemiHonestAdditiveShare<F>>,
+    >,
+    MCAggregateCreditOutputRow<F, S, BK>:
+        DowngradeMalicious<Target = MCAggregateCreditOutputRow<F, SemiHonestAdditiveShare<F>, BK>>,
```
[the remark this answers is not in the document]
martinthomson: The `Target = ` bound on `DowngradeMalicious` turns out to be essential (otherwise, when we try to downgrade, the compiler observes that we could be producing anything.

I did consider adding more letters to the signature, but it's not that much of a gain.  More associated types would help, but I found myself blocked by the lack of defaults for associated types in the language.

src/protocol/basics/mul/mod.rs
```diff
@@ -51,33 +51,33 @@ use {malicious::multiply as malicious_mul, semi_honest::multiply as semi_honest_
 
 /// Implement secure multiplication for semi-honest contexts with replicated secret sharing.
 #[async_trait]
-impl<'a, F: Field> SecureMul<SemiHonestContext<'a>> for Replicated<F> {
+impl<C: Context, F: Field> SecureMul<C> for Replicated<F> {
```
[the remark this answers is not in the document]
martinthomson: Ah, this was a particularly challenging part of the change.  So it's probably worth explaining.

We have four contexts: `SemiHonest`, `Malicious`, and upgraded versions of the two.  This implementation needs to hit three of those four, which all use `Replicated<F>` (`UpgradedMaliciousContext` uses `MaliciousReplicated<F>`).  This could be set to bound `C` to `UpgradableContext`, but then I'd need to add another implementation for `UpgradedSemiHonestContext` as well.  This was easier.

src/protocol/sort/generate_permutation.rs
martinthomson: This seems like something we can iterate on, for sure.  It seems like the biggest question is how we manage reveal.

What we might want to do is only implement Reveal for the two upgraded contexts.  That might be simpler overall.

martinthomson:
@andyleiserson that looks pretty promising.  I considered making an alias trait like this, but I hadn't gotten that far.

The short names (ZF, ZC) are a bit terse.  I might name `Fields` something like `IpaProtocol` though, and try to pull in a few more constraints.  (And use `P` as the constraint alias rather than `Z`.)

Sort and attribution both look like places that this pattern might help.  Each could build on the other, with IPA composing sorting and attribution.  (Can the ShuffledPermutationWrapper constraint on the function be included in the type constraints of a sort type alias?)

I wonder whether the match- and breakdown-key types couldn't be folded in to that type as well.

akoshelev (approved):

PR #625, opened 2023-05-03
Use intel clmul intrinsics

martinthomson:
If there is any performance gain, I wasn't able to detect it, but at least this is clean.

ARM intrinsics are still experimental, so these will have to do for now.

Closes #544.


.github/workflows/check.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     name: Additional Builds and Concurrency Tests
     env:
       CARGO_TERM_COLOR: always
-      RUSTFLAGS: -D warnings
+      RUSTFLAGS: -D warnings -C target-cpu=native
```
akoshelev, line 66: so dynamic feature detection does not work? If so, should we enable it in `cargo.toml` as well?

akoshelev:
> ARM intrinsics are still experimental, so these will have to do for now.

isn't `neon` and `aes` what we need for ARM? https://doc.rust-lang.org/core/arch/aarch64/fn.vmull_p64.html 
it works for my M1 :)

```bash
 rustc --print=cfg | egrep "(neon|aes)"
target_feature="aes"
target_feature="neon"
```

src/ff/galois_field.rs
```diff
@@ -42,6 +42,56 @@ fn assert_copy<C: Copy>(c: C) -> C {
...
+    #[cfg(all(target_arch = "x86_64", target_feature = "pclmulqdq"))]
+    return if BITS <= 32 {
+        unsafe { clmul_x86_64::clmul32(a, b) }
+    } else if BITS <= 64 {
+        unsafe { clmul_x86_64::clmul64(a, b) }
+    } else {
+        unreachable!();
+    };
+
+    debug_assert!(2 * BITS <= u128::BITS);
```
akoshelev, line 85: it is a little worrying that we won't have coverage here in CI but it does not seem easy to add :(
maybe one day we would build a docker image that emulates ARM and test IPA there

src/ff/galois_field.rs
```diff
@@ -42,6 +42,56 @@ fn assert_copy<C: Copy>(c: C) -> C {
...
+
+#[allow(unreachable_code)]
+fn clmul<const BITS: u32>(a: u64, b: u64) -> u128 {
+    #[cfg(all(target_arch = "x86_64", target_feature = "pclmulqdq"))]
+    return if BITS <= 32 {
+        unsafe { clmul_x86_64::clmul32(a, b) }
+    } else if BITS <= 64 {
+        unsafe { clmul_x86_64::clmul64(a, b) }
+    } else {
+        unreachable!();
```
akoshelev, line 82: super nit, but I disagree that it is unreachable :)

src/ff/galois_field.rs
```diff
@@ -42,6 +42,56 @@ fn assert_copy<C: Copy>(c: C) -> C {
...
+    }
+
+    /// clmul with 64-bit inputs (and a 128-bit answer).
+    pub unsafe fn clmul64(a: u64, b: u64) -> u128 {
+        let product = clmul(a, b);
+        extract::<1>(product) << 64 | extract::<0>(product)
+    }
+}
+
+#[allow(unreachable_code)]
```
akoshelev, line 74:
```suggestion
```

akoshelev (approved):

.github/workflows/check.yml
martinthomson: On my machine at least, the clmul feature is not available until I tell rust to enable it.

My understanding is that there might be features that are enabled by default, but the default configuration attempts to compile for any CPU with the target architecture, with features disabled if they are not available too all/most/? CPUs with that architecture.

src/ff/galois_field.rs
martinthomson: Unfortunately, once the hardware acceleration is enabled, the default code at the bottom is unreachable.

src/ff/galois_field.rs
martinthomson: It should be; we don't implement a Galois field with more than 64 bits.

src/ff/galois_field.rs
martinthomson: Yeah, getting access to aarch64 hardware is hard.  We could emulate, but I don't know if the emulator has these features.

martinthomson (commented):
@akoshelev in the absence of CI for aarch64, can I request that you give this a go to see if it works for you?  You might have to use `RUSTFLAGS="-C target-cpu=native"`, though I guess it is possible that on M1 macs the features are enabled by default.

akoshelev:
> in the absence of CI for aarch64, can I request that you give this a go to see if it works for you? You might have to use RUSTFLAGS="-C target-cpu=native", though I guess it is possible that on M1 macs the features are enabled by default.

works for me even without `target-cpu=native`.

```rust
    #[cfg(all(
        target_arch = "aarch64",
        target_feature = "neon",
        target_feature = "aes"
    ))]
    return unsafe {
        println!("vmul_p64");
        core::arch::aarch64::vmull_p64(a, b)
    };
```

```bash
cargo test -- --nocapture | grep vmul_p64 | wc -l
...
  279108
```

.github/workflows/check.yml
martinthomson: Oh yes, a huge difference.  Many more features in the latter than the former.  sse2 is in both, but I need the latter for aes.

src/ff/galois_field.rs
```diff
@@ -46,46 +49,69 @@ fn assert_copy<C: Copy>(c: C) -> C {
...
         }
         _mm_clmulepi64_si128(to_m128i(a), to_m128i(b), 0)
     }
 
-    #[allow(clippy::cast_sign_loss)]
+    #[allow(clippy::cast_sign_loss)] // Thanks Intel.
+    #[inline]
     unsafe fn extract<const I: i32>(v: __m128i) -> u128 {
-        _mm_extract_epi64(v, I) as u128
+        // Note: watch for sign extension that you get from casting i64 to u128 directly.
```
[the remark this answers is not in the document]
martinthomson: The tests we have caught it well enough.

PR #626, opened 2023-05-03
Catch the Gf2 sorting bug

martinthomson:
Constraining sort so that the field needs to implement `PrimeField` prevents it from compiling when someone sets `F` = `Gf2`.  I've added a few additional checks for some of the sillier versions of sort as well.


src/protocol/sort/generate_permutation_opt.rs
```diff
@@ -243,11 +243,52 @@ mod tests {
...
+
+    #[tokio::test]
+    pub async fn malicious() {
+        sortn(10).await;
+    }
+
+    /// Passing 32 records for Fp31 doesn't work.
+    #[tokio::test]
+    #[should_panic]
+    async fn fp31_overflow() {
```
akoshelev, line 262: I love Fp31 field. Do you want to check the panic message too? I've seen cases when tests continue to panic but for a different reason :)

src/protocol/sort/multi_bit_permutation.rs
```diff
@@ -30,16 +31,24 @@ use std::iter::repeat;
...
+        return Ok(vec![ctx.share_known_value(F::ZERO); num_records]);
+    }
 
     let num_multi_bits = (input[0]).len();
     assert!(num_multi_bits > 0);
 
+    if u128::try_from(num_records).unwrap() >= <F as PrimeField>::PRIME.into() {
+        return Err(Error::FieldValueTruncation(String::from(
+            "prime field is too small to sort this many records",
+        )));
```
akoshelev, line 49: it is a nit, but it may be easier to debug later if you say how many records in the error message

```suggestion
        return Err(Error::FieldValueTruncation(format!(
            "prime field {} is too small to sort {} records", std::any::type_name::<F>(), num_records)
        )));
```

akoshelev (approved):

PR #628, opened 2023-05-04
Moving `Step` related code to its own folder

taikiy:
No logic change. Just moving things around and renaming things.

Also renamed `GenericStep` to `Descriptive` without the `Step` postfix now that it's under `step` module. The name sounds too generic on its own though. Will be happy to change it to `DescriptiveStep`.


martinthomson (approved):
OK, this sort of thing is going to rot very badly.  Let's just land this one.

martinthomson (approved):

PR #629, opened 2023-05-04
Format code using group_imports=StdExternalCrate setting

akoshelev:
We have a lot of empty lines in import sections that don't follow any conventions. `rustfmt` tries to be conservative and does not reorder them, leading to multiple `use crate::` statements per module. Some IDE, ahem Jetbrains ahem, also do not respect the order and just add imports to the end of the import section even if `use crate::` already exists.

There is a nightly (again) setting that may work for us:
[group_imports](https://rust-lang.github.io/rustfmt/?version=master&search=group#group_imports) so I decided to try it. It basically creates 3 import sections for std, external crates and crate imports. 

There are no functional changes in this PR, only formatting and adjusting pre-commit hook and CI settings.

### DIFF
--- a/src/helpers/gateway/send.rs
+++ b/src/helpers/gateway/send.rs
@@ -1,17 +1,18 @@
-use crate::sync::Arc;
-use dashmap::DashMap;
-use futures::Stream;
 use std::{
     marker::PhantomData,
     num::NonZeroUsize,
     pin::Pin,
     task::{Context, Poll},
 };
+
+use dashmap::DashMap;
+use futures::Stream;
 use typenum::Unsigned;
 
 use crate::{
     helpers::{buffers::OrderingSender, ChannelId, Error, Message, Role, TotalRecords},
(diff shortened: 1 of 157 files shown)


src/helpers/gateway/send.rs
```diff
@@ -1,17 +1,18 @@
...
     num::NonZeroUsize,
     pin::Pin,
     task::{Context, Poll},
 };
+
+use dashmap::DashMap;
+use futures::Stream;
 use typenum::Unsigned;
 
 use crate::{
```
akoshelev, line 12: here is an example of multiple `use crate::` statements per module

akoshelev:
100%, I wasn't planning to given how many large PRs we have in-flight. I will wait until all of them land because it is really a matter of running `cargo fmt` again

akoshelev:
I am going to land it today before we start our major effort for aggregation circuit

Issue #608, opened 2023-04-20
Server should print info-level logs by default


andyleiserson:
The server translates the number of `-v` options on the command line to a log level, which is clever, and follows the convention of more `-v` meaning more verbose, but it also currently means that no `-v` flags is "quiet" and you have to pass `-vv` to get info level logs. Info logs should be printed by default. I'd propose that adding `-v` increases from there, and adding `-q` decreases, but the exact convention isn't important to me as long as you get info by default.

andyleiserson:
Changed the default log level to info in the linked PR.

PR #630, opened 2023-05-04
TLS server certificate configuration

andyleiserson:
Some possible points for discussion:
* This changes the default log level to info (#608).
* I have added `keygen` and `test-setup` commands as alternate functions of the `helper` binary. I like having utility commands embedded in the server binary since it means you always have the utilities anywhere you have the server. However, this does have the consequence that the utility-only dependencies `rcgen` and `time` become dependencies of `web-app`. If there is a preference to make each of the CLI functions a standalone binary (or to put the CLI utilities together in a `helper-tool` binary), I can do that.
* I add `cli` and `test-fixture` to the feature set for web tests because the new end-to-end test invokes `test_mpc`. I started to do the work to have test_mpc optionally print the results in machine-parsable format but decided it was more trouble than it was worth (mostly due to `print_output` being generic over `T: Debug`), for this test just checking that any output is produced successfully seems sufficient. It would also be possible to implement the client portion of this test inline like `net::transport::e2e_tests::happy_case`, but it seemed useful to have some test coverage of the test_mpc utility.
* The end-to-end test is noisier than I expect -- the server and test_mpc output is printed to the console even if the test passes. Maybe this is because logging is set up to go to stderr by default? Do we want to change that? Related, I moved printing of the "server listening on <address>" from helper main to MpcHelperServer, because it seems like a generally useful thing to do when a server starts. However, that meant it got printed lots of times by unit tests, which I silenced with `#[cfg(not(test))]`.

### DIFF
--- a/src/cli/test_setup.rs
+++ b/src/cli/test_setup.rs
@@ -20,10 +20,11 @@ pub struct TestSetupArgs {
     #[arg(short, long, default_value = "test_data")]
     output_dir: PathBuf,
 
+    /// Write http URLs to the network configuration (certificates will still be generated)
     #[arg(long)]
-    https: bool,
+    disable_https: bool,
 
-    #[arg(long, num_args = 3, value_name = "PORT", default_values = vec!["3000", "3001", "3002"])]
+    #[arg(short, long, num_args = 3, value_name = "PORT", default_values = vec!["3000", "3001", "3002"])]
     ports: Vec<u16>,
 }
 
@@ -59,7 +60,7 @@ pub fn test_setup(args: TestSetupArgs) -> Result<(), Box<dyn Error>> {
             Ok::<_, Box<dyn Error>>(PeerConfig {
                 url: format!(
                     "{}://localhost:{}",
-                    if args.https { "https" } else { "http" },
+                    if args.disable_https { "http" } else { "https" },
                     port
                 )
                 .parse()
--- a/src/cli/test_setup.rs
+++ b/src/cli/test_setup.rs
@@ -1,0 +1,79 @@
+use crate::{
+    cli::{keygen, KeygenArgs},
+    config::{NetworkConfig, PeerConfig},
+};
+use clap::Args;
+use std::{
+    error::Error,
+    fs::{self, DirBuilder},
+    iter::zip,
+    path::PathBuf,
+};
+
+#[derive(Debug, Args)]
+#[clap(
+    name = "test_setup",
+    about = "Prepare a test network of three helpers",
+    next_help_heading = "Test Setup Options"
+)]
+pub struct TestSetupArgs {
+    #[arg(short, long, default_value = "test_data")]
+    output_dir: PathBuf,
+
+    #[arg(long)]
+    https: bool,
+
+    #[arg(long, num_args = 3, value_name = "PORT", default_values = vec!["3000", "3001", "3002"])]
+    ports: Vec<u16>,
+}
+
+/// Prepare a test network of three helpers.
+///
+/// # Errors
+/// If a problem is encountered.
+///
+/// # Panics
+/// If something that shouldn't happen goes wrong.
+pub fn test_setup(args: TestSetupArgs) -> Result<(), Box<dyn Error>> {
+    if args.output_dir.exists() {
+        if !args.output_dir.is_dir() || args.output_dir.read_dir()?.next().is_some() {
+            return Err("output directory already exists and is not empty".into());
+        }
+    } else {
+        DirBuilder::new().create(&args.output_dir)?;
+    }
+
+    let peers = zip([1, 2, 3], args.ports)
+        .map(|(id, port)| {
+            let tls_cert = args.output_dir.join(format!("h{id}.pem"));
+            let tls_key = args.output_dir.join(format!("h{id}.key"));
+
+            keygen(KeygenArgs {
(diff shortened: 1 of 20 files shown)


src/cli/test_setup.rs
```diff
@@ -0,0 +1,79 @@
...
+    next_help_heading = "Test Setup Options"
+)]
+pub struct TestSetupArgs {
+    #[arg(short, long, default_value = "test_data")]
+    output_dir: PathBuf,
+
+    #[arg(long)]
+    https: bool,
+
+    #[arg(long, num_args = 3, value_name = "PORT", default_values = vec!["3000", "3001", "3002"])]
```
martinthomson, line 26:
```suggestion
    #[arg(short, long, num_args = 3, value_name = "PORT", default_values = vec!["3000", "3001", "3002"])]
```

src/config.rs
```diff
@@ -46,19 +47,51 @@ impl NetworkConfig {
...
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct PeerConfig {
+    /// Peer URL
     #[serde(with = "crate::uri")]
-    pub origin: Uri,
-    pub tls: HttpConfig,
+    pub url: Uri,
+
+    /// Peer's TLS certificate or CA
```
martinthomson, line 56:
```suggestion
    /// Peer's TLS certificate or CA in PEM format
```

tests/helper_networks.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    test_mpc
+        .stdin
+        .as_ref()
+        .unwrap()
+        .write_all(b"3,6\n")
+        .unwrap();
+    test_mpc.wait().unwrap_status();
+
+    for mut helper in helpers.into_iter() {
```
martinthomson, line 81:
```suggestion
    for mut helper in helpers {
```

src/cli/test_setup.rs
```diff
@@ -0,0 +1,79 @@
...
+    name = "test_setup",
+    about = "Prepare a test network of three helpers",
+    next_help_heading = "Test Setup Options"
+)]
+pub struct TestSetupArgs {
+    #[arg(short, long, default_value = "test_data")]
+    output_dir: PathBuf,
+
+    #[arg(long)]
+    https: bool,
```
martinthomson, line 24: I might prefer the inverse, so that doing the good thing is easier.  So maybe:

```suggestion
    disable_https: bool,
```

src/net/client/mod.rs
```diff
@@ -245,6 +271,26 @@ pub(crate) mod tests {
         (wrap(&arc_cb), wrap(&arc_cb))
     }
 
+    #[tokio::test]
+    async fn untrusted_certificate() {
+        let echo_data = "asdf";
```
martinthomson, line 276:
```suggestion
        const ECHO_DATA: &str = "asdf";
```

tests/helper_networks.rs
```diff
@@ -0,0 +1,97 @@
...
+    error::Error,
+    io::{self, Write},
+    iter::zip,
+    process::{Command, ExitStatus, Stdio},
+    str,
+};
+use tempfile::tempdir;
+
+const HELPER_BIN: &str = env!("CARGO_BIN_EXE_helper");
+const TEST_MPC_BIN: &str = env!("CARGO_BIN_EXE_test_mpc");
```
martinthomson, line 11: Unfortunately, I don't have a good alternative for this.  I assume that we can't import code from those modules so that we could statically enforce correctness, like:

```rust
use crate::helper;
const HELPER_BIN: &str = env!(concat!("CARGO_BIN_EXE_", stringify!(helper)));
```

(I'm not even sure if that would catch a rename or move of helper...)

martinthomson (approved):
Solid work.  @akoshelev, I will leave you to execute the merge.

src/cli/test_setup.rs
andyleiserson: Thoughts on whether it makes sense to do this on `helper` itself too? (I think I considered doing that, but resisted the temptation to tweak something that already existed even though I agree HTTPS should be the default.)

Aside: do you happen to know how to configure clap for `--foo` / `--no-foo` boolean options from a single specification?

tests/helper_networks.rs
andyleiserson: Yeah, I'm not sure how much better we can do here assuming we want to invoke the binaries. If we structured the tools as functions exposed by the `ipa` crate with trivial `main` wrappers, then we could invoke the functions from this test.

In the event one of these gets renamed, failure to update this shouldn't escape CI -- it's just something that would be caught by tests rather than build.

tests/helper_networks.rs
```diff
@@ -0,0 +1,98 @@
...
+                    Ok(())
+                } else {
+                    Err(status.to_string().into())
+                }
+            })
+            .unwrap()
+    }
+}
+
+fn test_network(ports: &[u16; 3], https: bool) {
```
akoshelev, line 31: this test is simply amazing, covering every command in helper binary!

akoshelev:
This is brilliant, high quality work and tons of useful functionality enabled! Great work @andyleiserson ! 

I am going to merge it in 1 hour

akoshelev (approved):

PR #631, opened 2023-05-06
Add a function that executes IPA in the clear

akoshelev:
This will be used inside `test_mpc` binary to validate the results returned from MPC helpers.

### DIFF
--- a/src/test_fixture/ipa.rs
+++ b/src/test_fixture/ipa.rs
@@ -135,7 +135,7 @@ pub fn generate_random_user_records_in_reverse_chronological_order(
 /// Assumes records all belong to the same user, and are in reverse chronological order
 /// Will give incorrect results if this is not true
 #[allow(clippy::missing_panics_doc)]
-fn update_expected_output_for_user<'a, I: Iterator<Item = &'a TestRawDataRecord>>(
+fn update_expected_output_for_user<'a, I: IntoIterator<Item = &'a TestRawDataRecord>>(
     records_for_user: I,
     expected_results: &mut [u32],
     per_user_cap: u32,
--- a/src/test_fixture/ipa.rs
+++ b/src/test_fixture/ipa.rs
@@ -17,7 +17,7 @@ use rand::{
     distributions::{Distribution, Standard},
     Rng,
 };
-use std::cmp::min;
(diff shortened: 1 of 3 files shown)


src/test_fixture/ipa.rs
```diff
@@ -81,8 +135,8 @@ pub fn generate_random_user_records_in_reverse_chronological_order(
 /// Assumes records all belong to the same user, and are in reverse chronological order
 /// Will give incorrect results if this is not true
 #[allow(clippy::missing_panics_doc)]
-pub fn update_expected_output_for_user(
-    records_for_user: &[TestRawDataRecord],
+fn update_expected_output_for_user<'a, I: Iterator<Item = &'a TestRawDataRecord>>(
```
martinthomson, line 138:
```suggestion
fn update_expected_output_for_user<'a, I: IntoIterator<Item = &'a TestRawDataRecord>>(
```

src/test_fixture/ipa.rs
```diff
@@ -35,6 +35,60 @@ pub struct TestRawDataRecord {
...
+/// ## Panics
+/// Will panic if you run in on Intel 80286 or any other 16 bit hardware.
+pub fn ipa_in_the_clear(
+    input: &[TestRawDataRecord],
+    per_user_cap: u32,
+    attribution_window: u32,
+) -> Vec<u32> {
+    // build a view that is convenient for attribution. match key -> events sorted by timestamp in reverse
+    // that is more memory intensive, but should be faster to compute. We can always opt-out and
+    // execute IPA in place
```
martinthomson, line 55: Rather than executing in place, we could iterate `input` once for each user.  That is, collect user IDs and then iterate `input` using `filter()`.  I seriously doubt that this will be a problem for test inputs.  This function takes many orders of magnitude less time than the MPC version right now.

martinthomson (approved):

src/test_fixture/ipa.rs
akoshelev: yea I was thinking about it too. What steered me away from this is that memory footprint is manageable: 10Mb overhead for 1M records and I did not need to change  `update_expected_output_for_user`.

PR #633, opened 2023-05-09
Match the behavior of IPA in the clear and MPC IPA

akoshelev:
Found this issue while working on some CLI tooling. Currently MPC just skips the attribution window logic if the window is set to be 0. IPA in the clear did not do it and I was observing different results running them one after another. 

It does seem better to use the type system to make this behavior more clear. As 0 truly does not make much sense as window value, `Option<NonZeroU32>>` is used instead.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -12,7 +12,10 @@ use ipa::{
     },
 };
 use rand::{rngs::StdRng, thread_rng, Rng, SeedableRng};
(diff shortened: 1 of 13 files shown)


taikiy:
Thanks for fixing the attribution window param!

taikiy (approved):
lgtm

akoshelev:
I am going to merge it as there are more PRs coming, feel free to review it later and I will address the feedback

PR #637, opened 2023-05-09
cleanup processes spawned by helper network test

andyleiserson:


tests/helper_networks.rs
```diff
@@ -28,6 +29,51 @@ impl UnwrapStatusExt for Result<ExitStatus, io::Error> {
...
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref().unwrap()
+    }
+}
+
+impl Drop for TerminateOnDrop {
+    fn drop(&mut self) {
+        if let Some(mut child) = self.0.take() {
+            eprintln!("killing process {}", child.id());
```
akoshelev, line 71: does it work when `wait` is called before dropping?

martinthomson: Hmm, if the panic is on another thread, then this will fail.

Maybe the answer is to make wait take `&mut self` instead of a value.  Then you wouldn't be calling `into_inner()`, but `self.0.as_mut().wait()`; holding a reference to the process instead.  You could then avoid having to kill by taking the value after the wait concludes.

andyleiserson: I'm not sure I understand the concern. Is this a concern about this primitive as used here? Or as it might be used more generally?

If `TerminateOnDrop::wait` is called before `TerminateOnDrop::drop`, then `wait` moved the value out of the option, and drop will do nothing because the option is `None`.

I don't think you can call `TerminateOnDrop::wait` and `TerminateOnDrop::drop` in parallel on different threads, because that would imply both threads simultaneously own the `Child`?.

This certainly won't clean up the child processes 100% of the time though (e.g. `panic = abort` or abnormal termination of the parent by some external action). If there is some off-the-shelf solution that will do a better job of cleanup I would be happy to use it.

martinthomson: I think that I see it now.  In a test other threads might panic, but those are caught.  So all we really need to do is ensure that - if this thread panics - the call to `drop` cleans up.  If we're waiting on the process, we don't need to kill it, because we're unlikely to be the ones that panic.

PR #639, opened 2023-05-10
Make test_mpc compare MPC results and results in the clear

akoshelev:
Printing secret shared values is great, but humans are not trained to read them unfortunately.
It is probably better to print expected results and actual results received from MPC ring.
CLI should indicate whether there is a difference and highlight it.

```bash
+-----+----------+---------+-------+
| Row | Expected | Actual  | Diff? |
+==================================+
| 0   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 1   | Some(3)  | Some(3) |       |
|-----+----------+---------+-------|
| 2   | Some(3)  | Some(3) |       |
+-----+----------+---------+-------+
```

it is not visible here, but matching rows are green and if there is a mismatch, it will show as red.

It also suppresses `test_mpc` verbose output in integration tests.

### DIFF
--- a/src/bin/test_mpc.rs
+++ b/src/bin/test_mpc.rs
@@ -54,7 +54,7 @@ pub struct CommandInput {
     input_file: Option<PathBuf>,
 
     #[arg(value_enum, long, default_value_t = FieldType::Fp32BitPrime, help = "Convert the input into the given field before sending to helpers")]
-    input_type: FieldType,
+    field: FieldType,
 }
 
 impl From<&CommandInput> for InputSource {
@@ -81,7 +81,7 @@ async fn clients_ready(clients: &[MpcHelperClient; 3]) -> bool {
         && clients[2].echo("").await.is_ok()
 }
 
-fn validate<I, S>(expected: I, actual: I)
+fn validate<I, S>(quiet: bool, expected: I, actual: I)
 where
     I: IntoIterator<Item = S>,
     I::IntoIter: ExactSizeIterator,
@@ -120,7 +120,9 @@ where
         }
     }
 
-    println!("{table}");
+    if !quiet {
+        println!("{table}");
+    }
     assert!(
         mismatch.is_empty(),
         "Expected and actual results don't match: {:?}",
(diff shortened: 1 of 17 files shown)


src/bin/test_mpc.rs
```diff
@@ -83,24 +81,58 @@ async fn clients_ready(clients: &[MpcHelperClient; 3]) -> bool {
...
+    let mut actual = actual.into_iter().fuse();
+    let mut mismatch = Vec::new();
+
+    let mut table = Table::new();
+    table.set_header(vec!["Row", "Expected", "Actual", "Diff?"]);
+
+    let mut i = 0;
+    loop {
+        let next_expected = expected.next();
+        let next_actual = actual.next();
```
martinthomson, line 100:
```rust
for (next_expected, next_actual) = zip(expected, actual).filter(|(e, a)| e.is_some() || a.is_some()) {
  ...
}
```
?

martinthomson (approved):

src/bin/test_mpc.rs
```diff
@@ -83,24 +81,57 @@ async fn clients_ready(clients: &[MpcHelperClient; 3]) -> bool {
...
+
+        match (&next_expected, &next_actual) {
+            (None, None) => break,
```
martinthomson, line 103: This is still better as:

```rust
if next_expected.is_none() && next_actual.is_none() {
  break;
}
```

src/bin/test_mpc.rs
```diff
@@ -123,57 +153,80 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                        input, &clients, query_id,
-                    )
-                    .await;
-                    print_output(&output);
-                }
-                InputType::Int64 => panic!("Only field values are supported"),
-            }
-        }
+        TestAction::Multiply => multiply(args, &clients).await,
+        TestAction::SemiHonestIPA => semi_honest_ipa(args, &clients).await,
```
martinthomson, line 157: No Malicious?

martinthomson (approved):
The approval stands :)

src/bin/test_mpc.rs
akoshelev: would be trivial to add, just not there yet

src/bin/test_mpc.rs
akoshelev: I don't mind doing it this way!

PR #641, opened 2023-05-11
make https the default, but enable running in http mode with the https config

andyleiserson:
Branch includes #637.

@akoshelev observed that `https` URLs in `network.toml` were honored even when running in HTTP mode. With this change, the URL scheme in the configuration is overridden with whatever is selected on the command line. To reduce confusion, I also changed `test-setup` to write URLs in the config with host and port only.

I also made HTTPS the default when running a helper (option is now `--disable-https` instead of `--https`), and in the default configuration for real-infra unit tests.

### DIFF
--- a/src/bin/helper.rs
+++ b/src/bin/helper.rs
@@ -1,4 +1,5 @@
-use clap::{self, builder::ArgPredicate, ArgAction, Parser, Subcommand};
+use clap::{self, Parser, Subcommand};
+use hyper::http::uri::Scheme;
 use ipa::{
     cli::{keygen, test_setup, KeygenArgs, TestSetupArgs, Verbosity},
     config::{NetworkConfig, ServerConfig, TlsConfig},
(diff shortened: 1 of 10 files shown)


src/net/server/mod.rs
```diff
@@ -188,6 +171,37 @@ impl MpcHelperServer {
     }
 }
 
+async fn spawn_server<A>(
```
andyleiserson, line 174: Moved this because `start_on()` crept over the clippy limit for lines of code in a function.

src/net/server/mod.rs
```diff
@@ -50,11 +50,20 @@ impl TracingSpanMaker for () {
 pub struct MpcHelperServer {
     transport: Arc<HttpTransport>,
     config: ServerConfig,
+    _network_config: NetworkConfig,
```
andyleiserson, line 53: Not used here currently, but will be used by the client certificate support.

src/bin/helper.rs
```diff
@@ -87,17 +84,29 @@ async fn server(args: ServerArgs) -> Result<(), Box<dyn Error>> {
     };
     let server_config = ServerConfig {
         port: args.port,
-        https: args.https,
+        disable_https: args.disable_https,
         tls,
     };
```
martinthomson, line 89: `ServerArgs::config()` might be a function that produces a `ServerConfig`, which would simplify this function a little..

martinthomson (approved):

PR #642, opened 2023-05-11
Allow passing 0 to `oneshot_ipa` to disable the attribution window (window = infinite)

taikiy:


benches/oneshot/ipa.rs
```diff
@@ -49,8 +49,13 @@ struct Args {
...
     /// The size of the attribution window, in seconds.
-    #[arg(short = 'w', long, default_value = "86400")]
-    attribution_window: Option<NonZeroU32>,
+    #[arg(
+        short = 'w',
+        long,
+        default_value = "86400",
+        help = "The size of the attribution window, in seconds. Pass 0 for an infinite window."
+    )]
+    attribution_window: Option<u32>,
```
martinthomson, line 58: You could, if you wanted, set this to simply `u32`, then your function below would be:

```rust
fn attribution_window(&self) -> Option<NonZeroU32> {
    NonZeroU32::new(self.attribution_window)
}
```

With `default_value`, the option here never really comes into play anyway.

taikiy: Removed Option<>!

PR #643, opened 2023-05-11
Streaming: Make seq_join take a stream

martinthomson:
I've started to look for places to apply this, and there are fewer than I'd have liked.  But this change was easy enough to execute.

Next step is to start changing the gap between protocol steps so that one produces a stream and the other consumes it.  Not as many options on that front as I had hoped, based on my first look, but we'll see.


akoshelev (approved):
heh, that was easy it seems :)

PR #645, opened 2023-05-12
Avoid a second build for benchmarks

martinthomson:
We build the benchmarks twice for this job, but we don't have to.


akoshelev (approved):
I trust @martinthomson that this does the right thing, because github did not run CI actions for this PR - or it decided not to show it to me. 

I do however see that we compile benchmarks twice: https://github.com/private-attribution/ipa/actions/runs/4954424381/jobs/8862900779

PR #647, opened 2023-05-13
Event generation using test_mpc and IPA integration test + migrate to HTTP2

akoshelev:
There are there major changes that this PR brings:

* Implement event generator as `Iterator<TestRawDataRecord>`. for millions of events that is more efficient than the existing approach. It takes about 3 seconds to generate 1 million events (excluding IO)
* Add new `gen-ipa-inputs` command
* IPA integration test that uses `gen-ipa-inputs` to generate inputs and run semi-honest IPA.

--edit
Unintended change, but IPA integration tests were failing (see #650) on HTTP 1.1, so I had to include changing clients to use HTTP2


tests/helper_networks.rs
```diff
@@ -140,3 +213,13 @@ fn http_network() {
...
 }
+
+#[test]
+fn http_semi_honest_ipa() {
+    test_ipa(&[3003, 3004, 3005], IpaSecurityModel::SemiHonest, false);
+}
+
+#[test]
+fn https_semi_honest_ipa() {
+    test_ipa(&[4433, 4434, 4435], IpaSecurityModel::SemiHonest, true);
```
akoshelev, line 224: this is a bit annoying - I know that there is no guaranteed way to pick an available port (at least on MacOS, on Linux it works pretty reliable with SO_REUSEPORT and rust appears to be using [it](https://github.com/rust-lang-nursery/rust-cookbook/issues/500)), but even best-effort approach with TcpListener should work here. Right now it is not possible to test this in parallel

andyleiserson: I thought about this when I wrote the test... a possible strategy I came up with is to create `TcpListeners` in the test, pass them to the helper processes as /dev/stdin, and add an option to the helper that does `TcpListener::from_raw_fd` rather than bind a socket. I didn't do this at the time because it seemed like a fair amount of work, and it seemed useful to have at least one test that exercises the standard flow for binding the socket on startup. But if we're adding more tests that follow this framework, it becomes more important to do this work. I'll see if I can get this done next week.

akoshelev:
integration tests hanging is not good: https://github.com/private-attribution/ipa/actions/runs/4964162687/jobs/8884058950?pr=647, opening an issue for that

src/bin/test_mpc.rs
```diff
@@ -77,7 +89,29 @@ enum TestAction {
     /// Execute end-to-end multiplication.
     Multiply,
     /// Execute IPA in semi-honest majority setting
-    SemiHonestIPA,
+    SemiHonestIpa(IpaQueryConfig),
+    /// Generate inputs for IPA
+    GenIpaInputs {
+        /// Number of records to generate
+        #[clap(long, short = 'c')]
```
martinthomson, line 96: the oneshot test uses -n (for number of records)

src/cli/csv.rs
```diff
@@ -0,0 +1,22 @@
+use std::{io, io::Write};
+
+pub trait CsvSerializer {
```
martinthomson, line 3: Do we name this simply `Serializer` (and then re-export as `CsvSerializer`?

src/test_fixture/event_gen.rs
```diff
@@ -0,0 +1,299 @@
...
+/// and maximum number of events per user. See [`Config`] for more details
+///
+/// [`Config`]: Config
+pub struct EventGenerator<R: Rng> {
+    config: Config,
+    rng: R,
+    users: Vec<UserStats>,
+    // even bit vector takes too long to initialize for MAX_USER_ID. Need a sparse structure
+    // here
+    used: HashSet<UserId>,
```
martinthomson, line 108: Do we even need to track this?  I mean, the problem here is that we're essentially stateless, but you could create user IDs in monotonically increasing sequence.  A PRP would allow you to have the identifiers that you use appear to be randomly distributed in the target range...

src/test_fixture/event_gen.rs
```diff
@@ -0,0 +1,299 @@
...
+                ..Config::default()
+            },
+        );
+
+        assert!(gen.next().is_some());
+        assert!(gen.next().is_some());
+        assert!(gen.next().is_none());
+    }
+
+    mod proptests {
```
martinthomson, line 241: I would test that timestamps are monotonic as well.

src/test_fixture/event_gen.rs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        })
+    }
+}
+
+impl<R: Rng> Iterator for EventGenerator<R> {
+    type Item = TestRawDataRecord;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        const USERS_IN_FLIGHT: usize = 10;
```
martinthomson, line 188: This seems a little small for larger event counts.  It's arbitrary, but I'd have said that you would expect hundreds of people for a million records.  (It's no extra work for IPA, but I'm not sure how much we care about fidelity.)

martinthomson (approved):
@akoshelev and @andyleiserson should work out who gets to merge and who get to rebase :)

akoshelev:
> @akoshelev and @andyleiserson should work out who gets to merge and who get to rebase :)

@andyleiserson lets merge your changes first

richajaindce:
> > @akoshelev and @andyleiserson should work out who gets to merge and who get to rebase :)
> 
> @andyleiserson lets merge your changes first

There are 2 PRs which would need to be rebased based on what gets landed first (#654 also) :P 
How about, we land this since it is already approved?

andyleiserson:
> How about, we land this since it is already approved?

That's fine with me.

akoshelev:
I am on vacation until Tuesday PT time, so I'll merge it to unblock you folks. Sorry @andyleiserson for the rebase

PR #649, opened 2023-05-15
Add an exact-sized stream trait

martinthomson:
If we're going to use streams, there will come a time where we need to be somewhat more definite about the length of the stream.  Our contexts need to know how many records are processed for a given step, so being able to interrogate a stream to determine its length will help a fair bit.

This change defines a new trait, modeled from `ExactSizeIterator` for streams and a stream wrapper that provides an implementation of that trait.  The idea is that when we get a stream from the network, we'll be able to explicitly mark the size of that stream so that protocol step implementations can just know how many records are being processed.

### DIFF
--- a/src/exact.rs
+++ b/src/exact.rs
@@ -1,0 +1,144 @@
+use futures::stream::Stream;
+use pin_project::pin_project;
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// An analogue of [`ExactSizeIterator`] for [`Stream`].
+/// This behaves exactly as you might expect based on the documentation of
+/// [`ExactSizeIterator`].
+///
+/// [`ExactSizeIterator`]: std::iter::ExactSizeIterator
+/// [`Stream`]: futures::stream::Stream
+pub trait ExactSizeStream: Stream {
+    /// Return the length of the stream that remains.
+    fn len(&self) -> usize {
+        let (lower, upper) = self.size_hint();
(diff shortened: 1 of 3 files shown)


src/exact.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+impl<S: Stream> Stream for FixedLength<S> {
+    type Item = S::Item;
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.project();
+        let res = this.inner.as_mut().poll_next(cx);
+        if let Poll::Ready(v) = &res {
+            if v.is_some() {
+                *this.len = (*this.len).wrapping_sub(1);
```
akoshelev, line 53: I would also add a debug_assert here, it feels the same as providing len that is greater than actual stream length

akoshelev (approved):

src/exact.rs
martinthomson: My theory was that any error will be caught at the end, so saving the assertions is fine.  Are you concerned that short streams (where we think the length is longer than it really is) need to be signaled sooner?  (A check here will never catch long streams.)

akoshelev: yea I was thinking in our setting something is off if streams are short. At least in debug builds/tests we should always provide the exact length, I can't think of a valid use case where the mismatch is expected - although I could be missing something

PR #653, opened 2023-05-17
Client cert authentication for inter-helper communication

### LINKED ISSUE #597: Client identity should not be taken from origin header
This should be addressed as part of the TLS work, but filing an issue to reference from a TODO.

andyleiserson:
Fixes #597


src/config.rs
andyleiserson, line 1: Merged this into `src/net/test.rs`.

src/net/test.rs
```diff
@@ -11,17 +22,134 @@ use axum::{
...
+    pub fn with_disable_https_option(mut self, value: bool) -> Self {
+        self.disable_https = value;
+        self
+    }
+
+    #[must_use]
+    pub fn build(self) -> TestConfig {
+        let mut sockets = None;
+        let ports = self.ports.unwrap_or_else(|| {
+            let socks = array::from_fn(|_| TcpListener::bind("127.0.0.1:0").unwrap());
```
martinthomson, line 112: The IPv6 people I know would be disappointed here.  But that's their problem, not ours :)

src/net/test.rs
```diff
@@ -138,52 +256,117 @@ impl TestServerBuilder {
...
-    pub fn default() -> Self {
-        Self::builder().build()
-    }
-
-    /// Return a test client builder
-    pub fn builder() -> TestClientsBuilder {
-        TestClientsBuilder::default()
-    }
-}
+pub const TEST_CERTS: [&str; 3] = [
```
martinthomson, line 274: Is this cleaner as `const X: [&[u8]; 3] = [b"...", b"...", b"..."];`?

src/net/server/mod.rs
```diff
@@ -210,6 +242,267 @@ where
...
+                Ok(()) => {
+                    return Some(ClientIdentity(id));
+                }
+                Err(webpki::Error::CertNotValidForName) => continue,
+                Err(err) => {
+                    error!("error checking client certificate: {err}");
+                }
+            }
+        }
+        // It might be nice to log something here. We could log the certificate base64?
```
martinthomson, line 385: That seems right.  Logging the certificate seems like a good idea.  I'd do that with `info!()` or lower though. Maybe use `warn!()` here (logspam from clients that offer random certificates probably isn't worth `error!()`, which ends up on the console even when logging is disabled).

src/net/server/mod.rs
```diff
@@ -210,6 +242,267 @@ where
...
+        }
+        // No exact match, look for a trust relationship
+        let end_entity_cert = match EndEntityCert::try_from(cert.as_ref()) {
+            Ok(cert) => cert,
+            Err(err) => {
+                error!("unable to parse client certificate: {err}");
+                return None;
+            }
+        };
+        for (id, peer) in network_config.enumerate_peers() {
```
martinthomson, line 370: I'm not 100% on this idea.  You've allowed the configuration to list trust anchors, rather than end entity certificates.  But in order for that to work, you need to verify a chain AND verify a name.  The code is relatively simple, so I'm not that concerned about carrying it, but I am not sure that we need it.  The exact match code is probably fine, up until the point we need to authenticate clients (and that's a whole new project).

src/net/server/mod.rs
```diff
@@ -210,6 +242,267 @@ where
...
+            let Ok(name) = DnsNameRef::try_from_ascii_str(peer.url.host().unwrap()) else {
+                error!("peer {} has an invalid hostname", peer.url.host().unwrap());
```
martinthomson, line 372: This is a case where configuration is bad, so it would be OK to `panic!()`.  The only reason I might prefer this is that the error tends to appear pretty late in the whole process (after the network is setup and a query comes in).

src/net/transport.rs
```diff
@@ -303,4 +306,18 @@ mod e2e_tests {
...
+            .build();
+        test_three_helpers(conf).await;
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn three_helpers_https() {
+        let conf = TestConfigBuilder::with_open_ports().build();
+        test_three_helpers(conf).await;
+    }
 }
```
martinthomson, line 323: I realize that this is tricky, but maybe we should have tests that validate each of the different client certificate configuration styles.  Except maybe the one where the system trust store is used (which would be too variable and hard to get test certificates for).  The code seems solid, but I can't verify that.

Difficulty in testing might be cause to trim some of the CA-based stuff.

src/net/server/handlers/query/mod.rs
```diff
@@ -30,6 +43,73 @@ pub fn h2h_router(transport: Arc<HttpTransport>) -> Router {
...
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<B>) -> Self::Future {
+        match req.extensions().get() {
+            Some(ClientIdentity(_)) => self.inner.call(req).left_future(),
+            // rustfmt's handling of this may be a bug?
```
martinthomson, line 89: It looks like a formatting bug to me too, but not worth worrying about.

```suggestion
```

martinthomson (commented):
Very nice!

I have a few questions about the functionality, particularly where you are using trust anchors rather than EE certs for authentication.  The barb in the tail here is testing (see the comment toward the end).

src/config.rs
andyleiserson: Oops, this comment was supposed to be on `src/test_fixture/config.rs`.

src/bin/helper.rs
```diff
@@ -13,6 +14,53 @@ use std::{error::Error, fs, path::PathBuf};
...
+        der::{asn1::ObjectIdentifier, Encode, PemWriter},
+        pem,
+        pkcs8::{PrivateKeyInfo, SecretDocument},
+        EcParameters, EcPrivateKey, LineEnding,
+    };
+    use std::str;
+
+    // Well this is annoying. The `native-tls` implementation on Mac OS cannot
+    // import unencrypted PKCS8-format EC keys. So we rewrite the key into a
+    // format that it will accept (OpenSSL format).
```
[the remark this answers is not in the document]
andyleiserson: There is, and I already left a comment on it, but I also intended and forgot to link it here, so thanks for the reminder.

src/config.rs
```diff
@@ -66,58 +78,48 @@ impl NetworkConfig {
...
-    /// If a certificate is specified here, only the specified certificate will be accepted. The system
-    /// truststore will not be used.
-    pub certificate: Option<String>,
+    /// It is possible to rely on the system truststore and omit this configuration if the peers use
+    /// certificates from public CAs, but the client certificate verification is not currently
+    /// configured to work off of the system truststore. (Besides the mechanics of accessing the
+    /// system truststore, there is also the issue of what client CA names the server should send
+    /// in that case.)
+    #[serde(default, deserialize_with = "certificate_from_pem")]
+    pub certificate: Option<Certificate>,
```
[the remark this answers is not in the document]
andyleiserson: They could be distinguished by hostname (except in test environments where they're all on the same host -- in that case I have no solution to offer). Maybe that is not as strong a check as we would like.

There were two cases I had in mind that might justify supporting CA certs here:
1. Since the helpers need TLS server certificates as well, and currently use the same web server to serve both requests both from other helpers and external requests, it might be convenient for the helper to be able to obtain one cert from a public CA and use that for both client and server auth.
2. Having some private CA that issues certificates to all the helpers. This seems quite dubious.

I am happy to drop the support for using CA certs here. As you note, there is a testing gap that ought to be closed if we keep it.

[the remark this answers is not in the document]
martinthomson: I'll leave the question of retaining code to your judgment, with the only stipulation being that if code is in the tree, it is tested somehow.

The question of a private CA doesn't really fit with my expectations here.  My general sense is that with mutually distrustful helpers (an important goal of ours), we don't want to create a single entity that can issue for all of them.  So that even pushes against the idea of having any CA-based system on the off-chance that two helpers ultimately choose the same CA and create the potential for collusion risk via that CA.

What I think is far more likely for client certs is that they are self-signed and sent around manually at the formation of the helper network and when their operational practices dictate.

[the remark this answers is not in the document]
martinthomson: Oh, I should have added: this code could be useful still.

If the helpers jointly operate a CA to authenticate report collectors, that might be an OK way to manage client authentication for requests.  We're not building that now, but it seems like you could have a system like that (each helper could have its own CA. say, with all helpers recognizing the CA so that they can authenticate uploads.

I don't know if that is better than the usual token-based schemes; business needs will ultimately have to influence this somewhat.

src/net/test.rs
andyleiserson: I can try changing this (and a few other places IPv4 localhost is used) to the name `localhost`. I do generally think it's worth doing the right thing for IPv6 when the choice presents itself. I'm not sure I'd want to bet money on whether this code sees a machine without an IPv4 stack or a machine with broken name resolution of localhost first, but we can see.

src/net/test.rs
andyleiserson: I guess so... I tend to think of PEM as string-like and prefer to put it in string types, but it does seem like the stuff we're passing these to prefers bytes.

src/net/server/mod.rs
```diff
@@ -216,6 +246,249 @@ where
...
+    type Error = S::Error;
+    type Future = Either<S::Future, Ready<Result<Response, S::Error>>>;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<B>) -> Self::Future {
+        if let Some(header_value) = req.headers().get(HTTP_CLIENT_ID_HEADER.clone()) {
```
akoshelev, line 473:
```suggestion
        if let Some(header_value) = req.headers().get(&HTTP_CLIENT_ID_HEADER) {
```

src/net/server/mod.rs
```diff
@@ -216,6 +246,249 @@ where
...
+            let id_result = header_value
+                .to_str()
+                .map_err(Into::into)
+                .and_then(|value_str| usize::from_str(value_str).map_err(Into::into))
+                .and_then(|value_int| {
+                    HelperIdentity::try_from(value_int).map_err(|e| {
```
akoshelev, line 479: maybe we can add `HelperIdentity::new(&str)` constructor later. Id does not really have to be an integer

src/net/server/mod.rs
```diff
@@ -102,53 +131,54 @@ impl MpcHelperServer {
...
+                    increment_counter!(RequestProtocolVersion::from(request.version()));
+                    increment_counter!(REQUESTS_RECEIVED);
+                }),
+        );
         let handle = Handle::new();
 
         let task_handle = match (self.config.disable_https, listener) {
             (true, Some(listener)) => {
+                let svc = svc
+                    .layer(layer_fn(SetClientIdentityFromHeader::new))
```
akoshelev, line 147: would it be a valid configuration with HTTPS enabled and `SetClientIdentityFromHeader` used instead of `ClientCertRecognizingAcceptor`? If so, is there any reasonable way we can protect against it?

akoshelev:
this is looking really good, it would be great to have some tests that prove incorrect configuration (using insecure authentication with HTTPS) cannot find its way into prod

akoshelev (approved):

src/net/server/mod.rs
andyleiserson: > would it be a valid configuration with HTTPS enabled and SetClientIdentityFromHeader used instead of ClientCertRecognizingAcceptor? If so, is there any reasonable way we can protect against it?

The current code doesn't allow enable/disable of SetClientIdentityFromHeader middleware other than via the HTTPS disable. So it's not possible to do something bad via configuration.

Whether we want to protect against (intentional or not) changes in behavior as code evolves is also a valid question. One strategy is to remove it from production builds entirely with conditional compilation -- like what `weak-field` does for `Fp31`. This is probably what I would pick, but I'm not sure if we want to add more build variants to the mix right now.

Another option might be to add some sort of interlock with the configuration. e.g. if `network.toml` is marked as production, then we disallow activating insecure modes. This could be done as a redundant check in `SetClientIdentityFromHeader::new`, which conveniently already has a copy of NetworkConfig.

(Side note, I'll also remove `pub(super)` from `SetClientIdentityFromHeader` -- it is left over from an earlier version where I tried to make `HTTP_CLIENT_ID_HEADER` an associated constant.)

src/net/server/mod.rs
andyleiserson: I changed it to use serde consistently for `HelperIdentity`. It seems a little weird to use `serde_json` to do integer-string conversions, but the serde support for this type is necessary anyways since it appears in a `PrepareQuery` message.

src/net/server/mod.rs
akoshelev: Yea, I was thinking along the same lines - if there is only one boolean check that prevents us from getting to use  `SetClientIdentityFromHeader` with HTTPS (in production), more safety nets would be good to have. maybe the interlock is everything what we need for now.

andyleiserson: I am going to merge this so that we can resolve the conflicts with #654, but I opened #660 to add the production interlock.

Issue #596, opened 2023-04-13
Helper config should be loaded from a file


andyleiserson:
The helper service config should be loaded from a file. Currently there is a config snippet hardcoded in src/bin/helper.rs.

PR #654, opened 2023-05-17
Generate and write public and private key using HPKE.

richajaindce:
These will be used to encrypt and decrypt match key shares


src/config.rs
```diff
@@ -96,13 +96,16 @@ pub struct PeerConfig {
     /// If a certificate is specified here, only the specified certificate will be accepted. The system
     /// truststore will not be used.
     pub certificate: Option<String>,
+
+    pub matchkey_encryption_key: Option<String>,
```
martinthomson, line 100: This won't be optional in practice, so maybe just make this a `String` directly.

src/config.rs
```diff
@@ -264,6 +268,16 @@ ZdJtrEnGRc0RGSBwP3N/fkcEAzuPw1ivcqOH5bWXPzSqPqQfADOrd8lK
 -----END PRIVATE KEY-----
 ";
 
+#[cfg(any(test))]
```
martinthomson, line 271: drop the `any()` if you only have one item in the list

src/config.rs
martinthomson, line 100: Can you document this parameter, similar to the comment above (though maybe a lot shorter).

martinthomson (commented):
I assume that this is only part of the whole.  I don't see the use of these keys anywhere.

With that in mind, this is good, but I would avoid using `Option<T>` for these.  It will be easier if we retain a code path where possible.

src/config.rs
richajaindce: Sounds good. Although do we need to pass the matchkey public key to the peers?
I think we should also add it to ServerConfig. Will do that

martinthomson: The other peers won't need this, but it probably makes sense to reuse the same configuration (or at least a subset of it) for client configuration.

richajaindce: Yes, I hope ServerConfig is the right place to put it.

src/config.rs
```diff
@@ -268,13 +296,10 @@ ZdJtrEnGRc0RGSBwP3N/fkcEAzuPw1ivcqOH5bWXPzSqPqQfADOrd8lK
 -----END PRIVATE KEY-----
 ";
 
-#[cfg(any(test))]
 const TEST_MATCHKEY_ENCRYPTION_KEY: &str = "\
```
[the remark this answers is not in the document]
akoshelev: we use `test-fixture` feature extensively in CLI tools. It makes me very nervous that there exists a valid build for helper server that could reference those test keys. I am thinking if we accidentally build and deploy such configuration and someone realizes that IPA uses the test key then they can decrypt all match keys.

I like the setup Andy proposed in the TLS PR - have `test_setup` generate everything we need and put it inside `test_data` folder for default setup.  Generate `network.toml` that references those and pick it up during helper bootstrap. This may work well for HPKE configuration too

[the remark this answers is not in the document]
richajaindce: I was able to hide it behind test-fixture feature flag for now. Some of plumbing for test setup should be done but I will complete it when I start coding the consumption in MPC.

src/config.rs
andyleiserson: I think `ServerConfig` indeed needs both the private key and the public key (in some formats, the private key is stored with a copy of the public key, but if we're just using raw bytes, that won't be the case).

I also think that the public key should go in `NetworkConfig` (`network.toml`). The idea is that `network.toml` gets distributed to clients of the helper network. `test_mpc` already takes an argument `--network network.toml`, and the event generation stuff that @akoshelev is adding to it will soon (possibly immediately, if configuration/use of the report key is mandatory) need to support encrypting the reports.

src/config.rs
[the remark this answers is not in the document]
andyleiserson: In #653 I have made some more changes to the way this test data is handled. `localhost_config` is gone, the `self-signed-certs` feature is gone, and some bits and pieces of code for generating test configs are consolidated into `src/net/test.rs`.

If this PR is ready soon, go ahead and merge it, and I'll resolve conflicts with #653, possibly moving some of the report key test data to match what I've done with the TLS test data.

andyleiserson:
> With that in mind, this is good, but I would avoid using Option<T> for these. It will be easier if we retain a code path where possible.

I agree that these aren't really optional, in the long term. There is a question of whether we want to support bypassing the match key encryption for some kinds of testing / protocol development work, like how we support HTTP for testing.

Another issue that may come up is how the encryption interacts with the choice between `real-world-infra` and `in-memory-infra`. There's no reason that match keys can't be encrypted in both cases, but a couple reasons we might not want that:
1. It may be useful to skip match key encryption for some kinds of MPC protocol development work where it isn't relevant. (Either for ease of interpreting the test data, or to reduce the number of build dependencies.)
2. `network.toml` and `ServerConfig` ([which should soon be `server.toml`](https://github.com/private-attribution/ipa/issues/596)) are currently only used for the real-world-infra. If we do want to use match key encryption with in-memory-infra, that may require some more tinkering with the test setup stuff.

src/config.rs
richajaindce: Within NetworkConfig we have a PeerConfig. Should we add a ClientConfig which will store the public key needed for encryption? There might be other information which would be needed to communicated to clients which can go there. @andyleiserson

src/config.rs
```diff
@@ -167,24 +183,37 @@ pub struct ServerConfig {
 
     /// TLS configuration for helper-to-helper communication
     pub tls: Option<TlsConfig>,
+
+    /// Configuration needed for encrypting and decrypting match keys
+    pub matchkey_encryption_info: Option<MatchKeyEncryptionConfig>,
```
richajaindce, line 188: @andyleiserson I have stored public and private keys as optional as MatchKeyEncryptionConfig in ServerConfig. Lets leave it optional for now to support disabling encryption of match keys

richajaindce: Added a flag in net/test.rs for disable_matchkey_encryption but not adding any test for that as yet. Will add it along with coding decryption handling of the reports

src/config.rs
andyleiserson: I think the public key could go directly in PeerConfig. The only significance that I had in mind for `PeerConfig` was that it was a container for the configuration that needed to be replicated 3x for the helpers.

Another possibility is to represent the key information in the OHTTP key configuration format mentioned in https://github.com/patcg-individual-drafts/ipa/blob/main/details/encryption.md#helper-key-configuration, but that doesn't seem important to do right now, and even using that format I think we could nest it within PeerConfig.

richajaindce: There are a few things in PeerConfig (e.g. port) which advertisers/publishers don't care about. I will move public_key inside PeerConfig and then we can see how it evolves

andyleiserson:
I resolved the merge conflicts with #653 and #647, you can pull the resolved version from the `report_key` branch in my fork. I'm not sure how to get a concise rendering of a merge commit out of github, but if you do `git show`, the lines marked with `++` are the ones that were introduced or changed by my conflict resolutions.

richajaindce:
> I resolved the merge conflicts with #653 and #647, you can pull the resolved version from the `report_key` branch in my fork. I'm not sure how to get a concise rendering of a merge commit out of github, but if you do `git show`, the lines marked with `++` are the ones that were introduced or changed by my conflict resolutions.

Thanks a lot for that! Will take a look

akoshelev:
@richajaindce I made some small changes, lmk what you think, if you are ok with these, we probably can push it

richajaindce:
> @richajaindce I made some small changes, lmk what you think, if you are ok with these, we probably can push it

The changes look great!! Thanks for them. Lets ship it!

PR #655, opened 2023-05-17
Print the step's full path

taikiy:
Let say we have this:

#[Step]
pub enum Step {
    ModulusConversionForMatchKeys,
}

The proc macro expansion function called at the compile time will receive the `Step` token but it doesn't have attributes like its full module path. I believe that path resolution is done at build time and there's no way to interpret that in proc macro.

This change adds a feature that, when enabled, narrowed steps will be prefixed with their full module paths. This information is used in the proc macro to generate StepNarrow implementations for IPA steps.

### DIFF
--- a/src/protocol/step.rs
+++ b/src/protocol/step.rs
@@ -75,7 +75,7 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
         }
 
         let mut id = self.id.clone() + "/";
-        #[cfg(feature = "step-trace")]
+        #[cfg(all(feature = "step-trace", feature = "in-memory-infra"))]
         {
             id += &(std::any::type_name::<S>().to_owned() + "::");
         }
--- a/src/protocol/step.rs
+++ b/src/protocol/step.rs
@@ -74,9 +74,14 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
             assert!(!s.contains('/'), "The string for a step cannot contain '/'");
         }
 
-        Self {
-            id: self.id.clone() + "/" + step.as_ref(),
+        let mut id = self.id.clone() + "/";
+        #[cfg(feature = "step-trace")]
+        {
+            id += &(std::any::type_name::<S>().to_owned() + "::");
         }
+        id += step.as_ref();
+
+        Self { id }
     }
 }
 
--- a/src/protocol/step.rs
+++ b/src/protocol/step.rs
@@ -77,7 +77,7 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
         let mut id = self.id.clone() + "/";
         #[cfg(all(feature = "step-trace", feature = "in-memory-infra"))]
         {
-            id += &(std::any::type_name::<S>().to_owned() + "::");
+            id += [std::any::type_name::<S>(), "::"].concat().as_ref();
         }
         id += step.as_ref();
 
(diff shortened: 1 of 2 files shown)


src/protocol/step.rs
```diff
@@ -74,9 +74,14 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
             assert!(!s.contains('/'), "The string for a step cannot contain '/'");
         }
 
-        Self {
-            id: self.id.clone() + "/" + step.as_ref(),
+        let mut id = self.id.clone() + "/";
+        #[cfg(feature = "step-trace")]
+        {
+            id += &(std::any::type_name::<S>().to_owned() + "::");
```
martinthomson, line 80: Note that ":" is not a valid URL character, so this will only work with the in-memory transport, you should add that as a condition.

src/protocol/step.rs
```diff
@@ -74,9 +74,14 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
...
         }
+        id += step.as_ref();
```
martinthomson, line 82: Did you think about the challenge of recording the enum variant as well, rather than the `AsRef<str>` output?

src/protocol/step.rs
martinthomson, line 80: This will create more garbage than necessary because it creates an owned copy of the type name and an owned copy of the type name plus "::".  I don't see why the following wouldn't work:

```suggestion
            id += std::any::type_name::<S>() + "::";
```

martinthomson (commented):
This seems right to me.

My original thought was that we'd use a different step variant to record this information, but tweaking this implementation also works.

src/protocol/step.rs
taikiy: Yes, initially that was my preferred option and we could use [strum](https://docs.rs/strum/latest/strum/) for that. One challenge would be the `struct` steps (i.e., `struct BitOpStep(usize)`), and tuple like enum variants with `repeat64str!` macro. (i.e., `IpaProtocolStep::Sort(usize)`).

I think we can do something similar to thiserror::Error derive macro where we annotate the step keys and values. With that approach we can generate both Descriptive (as_ref()) and Compact (ints) gates.

So I think we should consider thiserror::Error kind of approach. Since that'll require more consideration, use this simpler `AsRef<str>` output approach for the time being. I think the codegen algorithm shouldn't change that much with either approach.

src/protocol/step.rs
taikiy: How about "#"?

martinthomson: Oh, I don't think that it is worth trying to fix, but it might be worth restricting this feature to use with in-memory to avoid causing explosions.

src/protocol/step.rs
martinthomson: strum IntoStaticStr seems idea for this.  Want to add that to this patch so that you can get the entire thing together?

taikiy: I think I'll make a separate diff for that because we'll need to deal with `struct` steps. I guess the easiest solution is to convert `struct BitOpStep(usize)` to `enum Step { BitOp(usize) }`. For this, proc macro is also useful to enforce all step definitions must be of enum type.

src/protocol/step.rs
andyleiserson: Since the larger goal here is to eliminate use of strings for step representation in performance-sensitive contexts, it's not worth a lot of fiddling, but another possibility is to use [concat](https://doc.rust-lang.org/std/primitive.slice.html#method.concat): `[std::any::type_name::<S>(), "::"].concat()`.

martinthomson: Or even:

```rust
id += std::any::type_name::<S>();
id += "::";
```

Though Andy's point holds :)

PR #658, opened 2023-05-19
Replace `step::Descriptive` with type alias

taikiy:
This is my workaround to the "make Context generic over Gate" attempt which turned out to be a nightmare. Instead of making the context generic, I replaced all references to `Descriptive` concrete type with `GateImpl` type alias.

Using a type alias is much simpler but doesn't give us the flexibility of generics. I hope this is enough to let us switch between `Descriptive` gate and `Compact` gate...

### A little more context behind `Context<G: Gate>` attempt

Changing `Context` to `Context<G: Gate>` affects both protocol and infra layers, and eventually lifetime got involved because of the line below:

`pub type InMemoryTransport = Weak<transport::InMemoryTransport>;`

Once `ChannelId` was made generic over G, it led all the way to `InMemoryTransport<G>`.


akoshelev:
> Changing Context to Context<G: Gate> affects both protocol and infra layers, and eventually lifetime got involved because of the line below:

> pub type InMemoryTransport = Weak<transport::InMemoryTransport>;

> Once ChannelId was made generic over G, it led all the way to InMemoryTransport<G>.

can you please elaborate why `pub type InMemoryTransport = Weak<transport::InMemoryTransport>;` is a problem? there does not seem to be any lifetime here

src/protocol/step/mod.rs
```diff
@@ -1,4 +1,9 @@
-use std::fmt::{Debug, Display, Formatter};
+mod descriptive;
+
+pub use descriptive::Descriptive;
+use std::fmt::Debug;
+
+pub type GateImpl = Descriptive;
```
martinthomson, line 6:
```suggestion
pub type Gate = Descriptive;
```

martinthomson (approved):

PR #659, opened 2023-05-24
Removing match key provider from HPKE for now

richajaindce:
Based on https://github.com/patcg-individual-drafts/ipa/issues/57, there is an attack that a malicious Match Key Provider (MKP) can carry out by creating fake sites and using those sites' privacy budgets to measure events that happened on its own source or trigger site. Until we are able to find a good mitigation to this attack, we have decided to move forward with no match key provider in the IPA design. 
This PR removes mkp from being added to associated data in hpke.

### DIFF
--- a/src/hpke/info.rs
+++ b/src/hpke/info.rs
@@ -74,7 +74,7 @@ impl<'a> Info<'a> {
         let info_len = DOMAIN.len()
             + self.helper_origin.len()
             + self.site_domain.len()
-            + 3 // account for 4 delimiters
+            + 3 // account for 3 delimiters
             + std::mem::size_of_val(&self.key_id)
             + std::mem::size_of_val(&self.epoch)
             + std::mem::size_of_val(&self.event_type);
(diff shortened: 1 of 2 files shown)


akoshelev:
Can you please add an explanation why we are doing it in the PR description, so things don't get lost in history?

akoshelev (approved):

src/hpke/info.rs
```diff
@@ -79,19 +72,16 @@ impl<'a> Info<'a> {
     /// sender or receiver context.
     pub(super) fn into_bytes(self) -> Box<[u8]> {
         let info_len = DOMAIN.len()
-            + self.match_key_provider_origin.len()
             + self.helper_origin.len()
             + self.site_domain.len()
-            + 4 // account for 4 delimiters
+            + 3 // account for 4 delimiters
```
akoshelev, line 77: comment is obsolete now

richajaindce: Oh my bad..

Issue #644, opened 2023-05-12
Create a Dockerfile for IPA


akoshelev:
IPA setup requires some low-level dependencies like LLVM to be installed, along with Rust toolchain. There are too many Linux vendors and that makes it increasingly hard to support installation instructions for some/all of them.

Docker image running Ubuntu should help with this. A Dockerfile with instructions to install LLVM/openssl + Rust toolchain should suffice

andyleiserson:
I'll put in a plug for https://docs.docker.com/build/building/multi-stage/ as a strategy to produce an executable image that doesn't have to carry a full copy of the build toolchain.

martinthomson:
I'll second @andyleiserson's recommendation.  If you need a worked example, we use it here: https://github.com/mozilla/neqo/tree/main/qns

martinthomson:
I see we have one now.  Maybe close this?

I do see that it doesn't have an entrypoint set though, so maybe that's the last thing to do...

akoshelev:
Adding an entrypoint is a good idea. I have a bunch of PRs in flight that I'd like to close before I can add it because they all change the same file. After that I think I can close this

PR #662, opened 2023-05-25
Add Dockerfile to create IPA helper image

akoshelev:
It is not very capable right now, want to collect some early feedback. It does set up an image with helper binary and TLS config.

Addresses #644 

## Testing

```bash
$ scripts/helper-image.sh --identity 1 --hostname localhost
1
[+] Building 181.3s (17/17) FINISHED
 => [internal] load build definition from helper.Dockerfile                                                                                                                                                              0.0s
 => => transferring dockerfile: 798B                                                                                                                                                                                     0.0s
 => [internal] load .dockerignore                                                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                                                          0.0s
 => resolve image config for docker.io/docker/dockerfile:1                                                                                                                                                               0.7s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14                                                                                          0.0s
 => [internal] load .dockerignore                                                                                                                                                                                        0.0s
 => [internal] load build definition from helper.Dockerfile                                                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/debian:bullseye-slim                                                                                                                                                  0.6s
 => [internal] load metadata for docker.io/library/rust:latest                                                                                                                                                           0.8s
 => [internal] load build context                                                                                                                                                                                        1.4s
 => => transferring context: 2.77MB                                                                                                                                                                                      1.4s
 => CACHED [builder 1/4] FROM docker.io/library/rust:latest@sha256:c2dcc19371a976065950915307b03a9c5514e5143bdd833635d5501a0642992e                                                                                      0.0s
 => CACHED [stage-1 1/4] FROM docker.io/library/debian:bullseye-slim@sha256:7606bef5684b393434f06a50a3d1a09808fee5a0240d37da5d181b1b121e7637                                                                             0.0s
 => [stage-1 2/4] RUN apt-get update && rm -rf /var/lib/apt/lists/*                                                                                                                                                      1.8s
 => [builder 2/4] COPY . .                                                                                                                                                                                              15.4s
 => [builder 3/4] RUN set -eux;     cargo build --bin helper --release --no-default-features --features "web-app real-world-infra"                                                                                     162.1s
 => [stage-1 3/4] COPY --from=builder /target/release/helper /usr/local/bin/ipa-helper                                                                                                                                   0.0s
 => [stage-1 4/4] RUN set -eux;     mkdir -p /etc/ipa/pub;     /usr/local/bin/ipa-helper keygen --name localhost --tls-cert /etc/ipa/pub/1.pem --tls-key /etc/ipa/1.key                                                  0.2s
 => exporting to image                                                                                                                                                                                                   0.1s
 => => exporting layers                                                                                                                                                                                                  0.1s
 => => writing image sha256:c2ffb498408018eaa1aa57ffb030b0576e805c98b55358cd7f25c0832560405c                                                                                                                             0.0s
 => => naming to docker.io/private-attribution/ipa:dockerfile                                                                                                                                                            0.0s

```

### DIFF
--- a/scripts/helper-image.sh
+++ b/scripts/helper-image.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eou pipefail
+
 help() {
   echo "Usage: $0 --hostname [hostname] --identity [1|2|3]"
   echo "- hostname: public hostname that will appear on TLS certificate"
@@ -32,12 +34,8 @@ if [[ -z "${identity}" || -z "${hostname}" ]]; then
     exit 1
 fi
 
-branch="$(git rev-parse --abbrev-ref HEAD)"
-if [[ "$branch" == "main" ]]; then
-  tag="private-attribution/ipa:latest.$identity"
-else
-  tag="private-attribution/ipa:$branch.$identity"
-fi
+rev=$(git log -n 1 --format='format:%H' | cut -c1-10)
+tag="private-attribution/ipa:$rev-h$identity"
 
 cd "$(dirname "$0")"/.. || exit 1
 docker build -t "$tag" -f docker/helper.Dockerfile --build-arg IDENTITY="$identity" --build-arg HOSTNAME="$hostname" .
--- a/scripts/helper-image.sh
+++ b/scripts/helper-image.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+help() {
+  echo "Usage: $0 --hostname [hostname] --identity [1|2|3]"
+  echo "- hostname: public hostname that will appear on TLS certificate"
+  echo "- identity: helper identity to be fixed inside the IPA build"
+}
+
+parse_args() {
+  while [ "${1:-}" != "" ]; do
+    case "$1" in
+      --identity)
+        shift
+        identity="$1"
+        ;;
+      --hostname)
+        shift
+        hostname="$1"
+        ;;
+      *)
+      # unknown option
+      help
+      exit 1
+    esac
+    shift
+  done
+}
+
+parse_args "$@"
+if [[ -z "${identity}" || -z "${hostname}" ]]; then
+    help
+    exit 1
+fi
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+if [[ "$branch" == "main" ]]; then
+  tag="private-attribution/ipa:latest"
+else
+  tag="private-attribution/ipa:${branch}"
+fi
+
+cd "$(dirname "$0")"/.. || exit 1
+docker build -t "$tag" -f docker/helper.Dockerfile --build-arg IDENTITY="$identity" --build-arg HOSTNAME="$hostname" .
--- a/scripts/helper-image.sh
+++ b/scripts/helper-image.sh
@@ -34,9 +34,9 @@ fi
 
 branch="$(git rev-parse --abbrev-ref HEAD)"
 if [[ "$branch" == "main" ]]; then
-  tag="private-attribution/ipa:latest"
+  tag="private-attribution/ipa:latest.$identity"
 else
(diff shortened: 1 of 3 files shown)


bmcase:
@akoshelev  I installed docker and built the image,  when I try to run it with `sudo docker run <Image ID>`  nothing seems to happen.  What should I expect when running the docker image?  Should I be able to run some of the tests like `oneshot_ipa`?

bmcase:
Notes: talked with Alex and learned a few things. (inaccuracies my own)
- when the docker images runs it is an IPA Helper server that is waiting for a client to talk to it, the client being the Report Collector implementation. 
- when the docker image builds it generates its private/public key pairs.  They are in `etc/ipa`  you can access them by entering the docker image as: `docker run -it docker.io/private-attribution/ipa:test-docker.1 /bin/bash`  You can then copy them out and exchange in a secure way with the other helpers. 
- we should make a script to upload the public keys of other helpers to the docker image.

richajaindce:
> * `docker run -it docker.io/private-attribution/ipa:test-docker.1 /bin/bash`

I still couldn't make it run :( So will leave setup instructions for u to add @bmcase

scripts/helper-image.sh
```diff
@@ -0,0 +1,43 @@
...
+
+parse_args "$@"
+if [[ -z "${identity}" || -z "${hostname}" ]]; then
+    help
+    exit 1
+fi
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+if [[ "$branch" == "main" ]]; then
+  tag="private-attribution/ipa:latest.$identity"
```
andyleiserson, line 37: `latest` tags can be a source of confusion -- if we have to spin multiple versions of the helper image, it's not always obvious what version somebody has. And in some environments, a locally-stored `latest` image may be used without checking that it matches the latest available in the registry. I generally prefer to use the git SHA or a simple integer version in the tag instead.

Since the helper identity is also an integer, it may get confused with a version indicator. Maybe name the images `ipa-test-h{n}`?

scripts/helper-image.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
```
andyleiserson, line 2: I don't see anywhere that it would matter with the current version of the script, but maybe add `set -eo pipefail` just in case?

scripts/helper-image.sh
```diff
@@ -0,0 +1,43 @@
...
+branch="$(git rev-parse --abbrev-ref HEAD)"
+if [[ "$branch" == "main" ]]; then
+  tag="private-attribution/ipa:latest.$identity"
+else
+  tag="private-attribution/ipa:$branch.$identity"
+fi
+
+cd "$(dirname "$0")"/.. || exit 1
+docker build -t "$tag" -f docker/helper.Dockerfile --build-arg IDENTITY="$identity" --build-arg HOSTNAME="$hostname" .
```
andyleiserson, line 43: If we are going to pregenerate keys, we may want to augment this so that the images are fully prepared -- i.e. build all three images, collect keys, create and add `network.toml` to the images. However, no reason that has to be done in this PR.

andyleiserson (approved):

bmcase:
> I still couldn't make it run :( So will leave setup instructions for u to add @bmcase

This is what I did yesterday that worked for entering with bash and copying keys to/from the image. https://github.com/private-attribution/ipa/pull/663

scripts/helper-image.sh
akoshelev: sounds good to me. Branch is redundant in this case imo

Issue #615, opened 2023-04-24
Implement QueryId generation


andyleiserson:
`src/protocol/mod.rs`:
```
/// TODO: Generating this unique id may be tricky as it may involve communication between helpers and
/// them collaborating on constructing this unique id. These details haven't been flushed out yet,
/// so for now it is just an empty struct. Once we know more about it, we will make necessary
/// amendments to it
```

PR #664, opened 2023-05-28
Make helpers capable of running more than one query

akoshelev:
Make helpers capable of running more than one query

This change allows both in memory and HTTP implementations to run more than one query in sequence.
There were two things preventing it: `QueryProcessor` not properly cleaning up its state and transports keeping the stream
collections after query completion. The former was easy to fix, but I am not super happy with the fix for the latter.

There were a couple of options how to fix the transports:

* Generate a unique query id for each run (#615). We need this anyway, so this could be an incomplete fix.
  Incomplete because one may consider keeping streams for completed query around forever as memleak.
* Teach `Transport` trait to be aware of query completion. That looked as too big of a fix and a violation of the design we have right now
* Implement `reset` for both transports that cleans up their internal state.
  That's the approach taken in this change. HTTP transport is aware of query completion callback,
  so changing it was easy, in memory transport does not directly interact with it, so the fix is implemented at `TestApp` level

### DIFF
--- a/src/query/runner/ipa.rs
+++ b/src/query/runner/ipa.rs
@@ -23,20 +23,18 @@ impl Runner {
         ctx: SemiHonestContext<'_>,
         field: FieldType,
         input: ByteArrStream,
-    ) -> Box<dyn ProtocolResult> {
-        match field {
+    ) -> Result<Box<dyn ProtocolResult>, Error> {
+        Ok(match field {
             #[cfg(any(test, feature = "weak-field"))]
             FieldType::Fp31 => Box::new(
                 self.run_internal::<crate::ff::Fp31, MatchKey, BreakdownKey>(ctx, input)
-                    .await
-                    .expect("IPA query failed"),
+                    .await?,
             ),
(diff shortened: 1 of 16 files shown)


src/query/runner/ipa.rs
```diff
@@ -23,20 +23,18 @@ impl Runner {
         ctx: SemiHonestContext<'_>,
         field: FieldType,
         input: ByteArrStream,
-    ) -> Box<dyn ProtocolResult> {
-        match field {
+    ) -> Result<Box<dyn ProtocolResult>, Error> {
```
andyleiserson, line 26: Can/should this use `QueryResult`? (Even if yes, no need to change in this PR, because I have more changes coming to this code.)

andyleiserson (approved):
Thanks for taking care of this. We'll need to revisit some of it, but it should meet our needs in the near term.

src/query/runner/ipa.rs
akoshelev: oh yea, my bad - missed it. Leaving it as is as suggested

PR #666, opened 2023-05-28
Some integration tests improvements

akoshelev:
Things that I noticed:

* When helper CLI fails to interpret its arguments, it just hangs because tests never check its status. That causes integration tests to hang. Fixed by checking whether the helper CLI exited before allowing to proceed
* temp dir contains useful artifacts to reproduce failures. Unfortunately it drops the directory by default, even if tests failed with a panic. This PR changes that by always preserving the temp directory if panic occurs.

### DIFF
--- a/tests/helper_networks.rs
+++ b/tests/helper_networks.rs
@@ -12,7 +12,10 @@ use std::{
     process::{Child, Command, ExitStatus, Stdio},
     str,
 };
-use tempfile::tempdir;
+use tempdir::TempDir;
+
+#[cfg(all(test, feature = "cli"))]
+pub mod tempdir;
 
 const HELPER_BIN: &str = env!("CARGO_BIN_EXE_helper");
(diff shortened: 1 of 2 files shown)


tests/helper_networks.rs
```diff
@@ -147,13 +150,19 @@ fn spawn_helpers(
...
+            }
+
+            child.terminate_on_drop()
         })
         .collect::<Vec<_>>()
 }
 
 fn test_network(https: bool) {
-    let dir = tempdir().unwrap();
+    let dir = TempDir::new(false); // set to true if tempdir needs to be preserved
```
richajaindce, line 165: nit: set to true if tempdir needs to be preserved ALWAYS. I think we are preserving it in case of a fail anyways by this change. Also, can you point to where the folder with failed run details can be found?

richajaindce (approved):
Looks good

PR #667, opened 2023-05-29
Reduce the size of Docker build context

akoshelev:
Before this change, we had Docker transferring gigabytes (if both release and debug builds exist) of data inside the image and make no use of it. It was happening because docker [expects](https://docs.docker.com/engine/reference/builder/#dockerignore-file) `.dockerignore` file to be inside the root folder of the context.

In IPA folder structure, `.dockerignore` was put right next to the `.Dockerfile` so it wasn't picked up when building the image.

There does not seem to be an option to provide path to `.dockerignore` to Docker CLI, so I build the context manually

### DIFF
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,2 +1,3 @@
 target/
 test_data/
+.git/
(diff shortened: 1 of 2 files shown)


richajaindce (approved):
This fixes the issue. Lets land this :)

PR #669, opened 2023-05-30
Trigger CI actions on integration tests changes

akoshelev:
If a PR includes only integration test changes, it does not trigger the workflow, see https://github.com/private-attribution/ipa/pull/666

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,12 +6,14 @@ on:
       - '**/Cargo.lock'
       - 'src/**/*'
       - 'benches/**/*'
+      - 'tests/**/*'
   pull_request:
       paths:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
       - 'src/**/*'
       - 'benches/**/*'
+      - 'tests/**/*'
 
 jobs:
   basic:


.github/workflows/check.yml
```diff
@@ -6,12 +6,14 @@ on:
...
       - 'src/**/*'
       - 'benches/**/*'
+      - 'tests/**/*'
   pull_request:
       paths:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
       - 'src/**/*'
       - 'benches/**/*'
+      - 'tests/**/*'
```
andyleiserson, line 16: Super nit, `tests/**` would be sufficient, no?

andyleiserson (approved):

.github/workflows/check.yml
akoshelev: hmm, I thought `tests/**` will grab only files inside tests and all the subdirectories (but not the files in them)?

maybe a macos thing

```bash
list=(/var/log/**/*); echo ${list[@]} | tr -s ' ' '\n' | wc -l
61
list=(/var/log/**); echo ${list[@]} | tr -s ' ' '\n' | wc -l
49
```

PR #671, opened 2023-05-31
Command to generate `network.toml`

akoshelev:
In order to streamline our helper image bootstrap, we need a command to generate a valid `network.toml` from a set of public keys and public uris. The flow would be as follows:

* script generates 3 docker images, each with its own configuration (DNS name)
* script copies public configuration from each helper to each other helper
* script runs `docker run` to execute this command to generate `network.toml` on each image

after this, images will be fully ready to be copied to the nodes that will run helpers

## Example
```bash
  cargo run --bin helper --no-default-features --features="cli web-app real-world-infra" -- confgen --keys-dir test_data --output-dir /tmp/ipaconfgen
```

### DIFF
--- a/docker/helper.Dockerfile
+++ b/docker/helper.Dockerfile
@@ -21,11 +21,12 @@ RUN apt-get update && rm -rf /var/lib/apt/lists/*
 COPY --from=builder ${SOURCES_DIR}/target/release/helper $HELPER_BIN_PATH
 
 # generate certificate/private key for TLS
+# make sure these names are consistent with the ones defined in CliPaths trait: src\cli\paths.rs
 RUN set -eux; \
     mkdir -p $CONF_DIR/pub; \
     $HELPER_BIN_PATH keygen \
     --name $HOSTNAME \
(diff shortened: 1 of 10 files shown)


richajaindce:
> In order to streamline our helper image bootstrap, we need a command to generate a valid `network.toml` from a set of public keys and public uris. The flow would be as follows:
> 
> * script generates 3 docker images, each with its own configuration (DNS name)
> * script copies public configuration from each helper to each other helper
> * script runs `docker run` to execute this command to generate `network.toml` on each image
> 
> after this, images will be fully ready to be copied to the nodes that will run helpers

Can you also include an example of how to run the script in the description. Also, in case anyone wants to override defaults (e.g. public/private key), how will they do it?

akoshelev:
> Can you also include an example of how to run the script in the description. Also, in case anyone wants to override defaults (e.g. public/private key), how will they do it?

sure, updated the description. for now CLIs don't support custom names for public/private keys. I don't see an issue adding them in the future though...

benjaminsavage (approved):
Thanks for writing this Alex!

PR #673, opened 2023-06-01
Generate dynamic steps

taikiy:
There are dynamic steps that are generated based on the number of input rows to IPA.

Since we are only executing IPA with 10 input rows, it only generates log2(10) (maybe a few more/less because of the optimizations) dynamic steps. We want to generate all possible steps for code generation without actually executing IPA with 1B rows.

This change also suppresses header lines not needed for step colleciton.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -102,8 +102,7 @@ async fn run(args: Args) -> Result<(), Error> {
     };
 
     let seed = args.random_seed.unwrap_or_else(|| thread_rng().gen());
-    #[cfg(not(feature = "step-trace"))]
-    println!(
+    tracing::trace!(
         "Using random seed: {seed} for {q} records",
         q = args.query_size
     );
@@ -124,8 +123,7 @@ async fn run(args: Args) -> Result<(), Error> {
         ipa_in_the_clear(&raw_data, args.per_user_cap, args.attribution_window());
 
     let world = TestWorld::new_with(config.clone());
-    #[cfg(not(feature = "step-trace"))]
-    println!("Preparation complete in {:?}", _prep_time.elapsed());
+    tracing::trace!("Preparation complete in {:?}", _prep_time.elapsed());
 
     let _protocol_time = Instant::now();
     test_ipa::<BenchField>(
@@ -136,8 +134,7 @@ async fn run(args: Args) -> Result<(), Error> {
         args.mode,
     )
     .await;
-    #[cfg(not(feature = "step-trace"))]
-    println!(
+    tracing::trace!(
         "{m:?} IPA for {q} records took {t:?}",
         m = args.mode,
         q = args.query_size,
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -95,13 +95,14 @@ impl Args {
 async fn run(args: Args) -> Result<(), Error> {
     type BenchField = Fp32BitPrime;
 
-    let prep_time = Instant::now();
+    let _prep_time = Instant::now();
     let config = TestWorldConfig {
         gateway_config: GatewayConfig::new(args.active()),
         ..TestWorldConfig::default()
     };
 
     let seed = args.random_seed.unwrap_or_else(|| thread_rng().gen());
+    #[cfg(not(feature = "step-trace"))]
     println!(
         "Using random seed: {seed} for {q} records",
         q = args.query_size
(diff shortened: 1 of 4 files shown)


scripts/collect_steps.py
```diff
@@ -16,15 +17,22 @@
...
+BREAKDOWN_KEYS = [8, 33]
 SECURITY_MODEL = ["malicious", "semi-honest"]
 ROOT_STEP_PREFIX = "protocol/alloc::string::String::run-0"
-EXPECTED_HEADER_LINES_COUNT = 4
+
+DEPTH_DYNAMIC_STEPS = "ipa::protocol::attribution::InteractionPatternStep"
+BIT_DYNAMIC_STEPS = [
+    "ipa::protocol::attribution::aggregate_credit::Step::compute_equality_checks",
+    "ipa::protocol::attribution::aggregate_credit::Step::check_times_credit",
+]
```
taikiy, line 33: In the future, we want to do the same thing with a macro annotating dynamic steps and generate code. In order to do that, we need to be able to resolve the step's module path (or source file path) but that functionality is only available in nightly.

scripts/collect_steps.py
```diff
@@ -16,15 +17,22 @@
...
     "--",
     "--num-multi-bits",
-    "1",
+    "3",
 ]
 QUERY_SIZE = 10
 PER_USER_CAP = [1, 3]
 ATTRIBUTION_WINDOW = [0, 86400]
-BREAKDOWN_KEYS = [1, 33]
+BREAKDOWN_KEYS = [8, 33]
```
akoshelev, line 25: is there a difference between 1 and 8?

scripts/collect_steps.py
```diff
@@ -66,14 +67,38 @@ def collect_steps(args):
...
         line = proc.stdout.readline()
         if not line or line == "":
             break
 
         if not line.startswith(ROOT_STEP_PREFIX):
-            header += 1
+            print("Unexpected line: " + line)
+            exit(1)
+
+        # There are protocols in IPA that that will generate log(N) steps where N is the number
```
akoshelev, line 79: maybe this comment belongs to DEPTH_DYNAMIC_STEPS definition

scripts/collect_steps.py
```diff
@@ -16,15 +17,22 @@
...
 ROOT_STEP_PREFIX = "protocol/alloc::string::String::run-0"
-EXPECTED_HEADER_LINES_COUNT = 4
+
+DEPTH_DYNAMIC_STEPS = "ipa::protocol::attribution::InteractionPatternStep"
+BIT_DYNAMIC_STEPS = [
+    "ipa::protocol::attribution::aggregate_credit::Step::compute_equality_checks",
+    "ipa::protocol::attribution::aggregate_credit::Step::check_times_credit",
+]
+MAXIMUM_DEPTH = 32
+MAXIMUM_BIT_LENGTH = 32
```
akoshelev, line 35: while it seems valid to only consider inputs with less than 2B rows, I thought we will and already are decomposing match keys of 40 bits (and looking to use 64 bit match keys in the future)

src/telemetry/step_stats.rs
```diff
@@ -35,6 +35,7 @@ impl CsvExporter for Metrics {
         // then dump them to the provided Write interface
         // TODO: include role dimension. That requires rethinking `Metrics` implementation
         // because it does not allow such breakdown atm.
+        #[cfg(not(feature = "step-trace"))]
```
akoshelev, line 38: maybe we can contain this feature flag to be inside the benchmark only, by configuring test world and metrics by optiionally taking an argument not to print the header

scripts/collect_steps.py
taikiy: There are steps that are dynamically generated based on the number of bits used to represent breakdown keys, and I realized that the coverage wasn't enough with "1". "8" is enough to cover all cases for 1-32 (specialized aggregation protocol). >=33 uses a generic protocol so one is enough to cover all.

scripts/collect_steps.py
taikiy: Bit-dynamic steps are from fields which its bit-length are determined at runtime (e.g. breakdown keys). Match key's bit-length is fixed in the code so the sort protocol generates complete steps.

scripts/collect_steps.py
akoshelev: gotcha. I would suggest to maybe follow up with a comment that explains it inside the code as well. Does not have to be in this PR

akoshelev (approved):

benches/oneshot/ipa.rs
```diff
@@ -95,13 +95,14 @@ impl Args {
...
 
-    let prep_time = Instant::now();
+    let _prep_time = Instant::now();
     let config = TestWorldConfig {
         gateway_config: GatewayConfig::new(args.active()),
         ..TestWorldConfig::default()
     };
 
     let seed = args.random_seed.unwrap_or_else(|| thread_rng().gen());
+    #[cfg(not(feature = "step-trace"))]
```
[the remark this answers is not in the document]
taikiy: Actually, I think this is a useful info to print out in debug mode, too. I made a change together with `print_header` field in Metrics.

scripts/collect_steps.py
taikiy: added comments

src/telemetry/step_stats.rs
taikiy: done!

PR #674, opened 2023-06-01
malicious network tests

bmcase:
add ability for client (RC) to request that the servers run a malicious IPA query and tests for this.


bmcase:
@akoshelev can you help review this?   Not sure what is going on with `runner/ipa.rs` tests -- there are warnings to remove the imports but if they are removed their are compile errors.  Also some branch conflict with this part.

src/net/http_serde.rs
```diff
@@ -135,7 +135,29 @@ pub mod query {
...
+                        max_breakdown_key: u32,
+                        attribution_window_seconds: Option<NonZeroU32>,
+                        num_multi_bits: u32,
+                    }
+                    let Query(IPAQueryConfigParam {
+                        per_user_credit_cap,
+                        max_breakdown_key,
+                        attribution_window_seconds,
+                        num_multi_bits,
+                    }) = req.extract().await?;
```
richajaindce, line 158: Can this be combined like for semihonest?

src/helpers/transport/query.rs
```diff
@@ -248,6 +251,6 @@ impl IpaQueryConfig {
 
 impl From<IpaQueryConfig> for QueryType {
     fn from(value: IpaQueryConfig) -> Self {
```
richajaindce, line 253: Where is this used? Dont we need it for malicious also?

src/query/executor.rs
```diff
@@ -73,18 +73,36 @@ pub fn start_query(
         // Negotiate PRSS first
         let step = step::Descriptive::default().narrow(&config.query_type);
         let prss = negotiate_prss(&gateway, &step, &mut rng).await.unwrap();
-        let ctx = SemiHonestContext::new(&prss, &gateway);
+        //let ctx = SemiHonestContext::new(&prss, &gateway);
```
richajaindce, line 76: remove this

src/query/executor.rs
```diff
@@ -73,18 +73,36 @@ pub fn start_query(
...
+                    .run(
+                        SemiHonestContext::new(&prss, &gateway),
+                        config.field_type,
+                        input,
+                    )
+                    .await
+            }
+            QueryType::MaliciousIpa(ipa_query_config) => {
+                IpaRunner(ipa_query_config)
+                    //.run(SemiHonestContext::new(&prss, &gateway), config.field_type, input)
```
richajaindce, line 100: remove the commented line

src/cli/playbook/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod input;
-mod ipa;
+pub mod ipa;
```
akoshelev, line 2: I think we can keep it private and re-export the malicious function like this

```rust
pub use self::ipa::malicious;
```

or if we can make a single function like suggested above, we won't even need that

src/helpers/transport/query.rs
akoshelev: good finding, I wasn't able to locate a single use of this conversion, so we can just try to delete it and see if any check fails. I spot-checked shuttle tests and benches and received no complains about removal

andyleiserson:
This overlaps my upcoming changes for match key decryption, and probably also #664, so we'll need to pay attention to the sequencing.

src/net/http_serde.rs
bmcase: I'm not sure I follow what you mean.  These appear to also be separate from semihonest in lines 125-136.  what am I missing?

src/cli/playbook/ipa.rs
```diff
@@ -102,3 +102,91 @@ where
 
     breakdowns
 }
+
+/// Malicious IPA protocol.
+/// Returns aggregated values per breakdown key represented as index in the returned vector
+#[allow(clippy::missing_panics_doc)]
+pub async fn malicious<F, MK, BK>(
```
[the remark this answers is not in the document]
bmcase: yep, fixed this too

src/cli/playbook/mod.rs
bmcase: yeah, was able to just remove after the change above

bmcase:
Other than the changes @richajaindce  suggested in `http_serde,rs`, I've incorporated this feedback.  cc @akoshelev

bmcase:
okay, I compressed the matched code in `http_serde.rs` as @richajaindce  suggested.

src/net/http_serde.rs
bmcase: okay, updated this to remove some of the duplicate code in the match.

akoshelev:
@bmcase you need         #[allow(clippy::large_futures)] annotation inside the newly created test. I would suggest merging the latest changes from `main` and running `rustup update` - you would need the latest compiler version to reproduce the error. Andy already fixed the similar issue: https://github.com/private-attribution/ipa/blob/2371b14f9fc23f4787d22049f0756d1c6ae0cae6/src/query/runner/ipa.rs#L116

bmcase:
okay merged the main branch into the PR and fixed the clippy errors.

src/cli/playbook/ipa.rs
```diff
@@ -102,3 +102,91 @@ where
...
+//         // I think using u32 is wrong, we should move to u128
+//         let trigger_value = u32::try_from(row.trigger_value.as_u128()).unwrap();
+//         if breakdown_key >= breakdowns.len() {
+//             breakdowns.resize(breakdown_key + 1, 0);
+//             breakdowns[breakdown_key] += trigger_value
+//         }
+//     }
+//
+//     breakdowns
+// }
```
akoshelev, line 192:
```suggestion
```

src/net/http_serde.rs
```diff
@@ -157,11 +171,27 @@ pub mod query {
             match self.query_type {
                 #[cfg(any(test, feature = "test-fixture", feature = "cli"))]
                 QueryType::TestMultiply => write!(f, "query_type={}", QueryType::TEST_MULTIPLY_STR),
-                QueryType::Ipa(config) => {
+                QueryType::SemiHonestIpa(config) => {
```
akoshelev, line 174: Maybe we can replace it with

```rust
                qt @ (QueryType::SemiHonestIpa(config) | QueryType::MaliciousIpa(config)) => {
                    write!(
                        f,
                        "query_type={qt}&per_user_credit_cap={}&max_breakdown_key={}&num_multi_bits={}",
                        config.per_user_credit_cap,
                        config.max_breakdown_key,
                        config.num_multi_bits,
                        qt=qt.as_ref(),
                    )?;

                    if let Some(window) = config.attribution_window_seconds {
                        write!(f, "&attribution_window_seconds={}", window.get())?;
                    }

                    Ok(())
                }
```

this covers both malicious and semi-honest case and does not require code repetition for attribution window

akoshelev (approved):

src/net/http_serde.rs
bmcase: okay, added this

src/cli/playbook/ipa.rs
bmcase: yep, forgot to delete that old code.

PR #678, opened 2023-06-02
Fix clippy warnings from rust update

andyleiserson:


src/query/runner/ipa.rs
```diff
@@ -115,6 +115,7 @@ mod tests {
 
         let world = TestWorld::default();
         let contexts = world.contexts();
+        #[allow(clippy::large_futures)]
```
andyleiserson, line 118: The future is 23024 bytes, vs. a limit of 16 kB. A limit of 16 kB for futures on the stack seems unnecessarily conservative for us, but since this is a test anyways, I just waived it.

akoshelev: I love this new lint, I don't have to guess the future size on the stack anymore!

akoshelev (approved):

PR #682, opened 2023-06-07
Initialize key registry from a set of public/private keys

akoshelev:
I was thinking about taking a file name instead, but it seems that it should be done at the higher level.

@andyleiserson lmk if that's what you expect


andyleiserson:
We will also need something to call this function on the key registry instance with the key pair from `ServerConfig.hpke_config`.

andyleiserson:
I pulled this commit into #692 and made some additional changes on top of it.

akoshelev:
closed by #692

PR #683, opened 2023-06-07
Reduce duplicated code involved in running a query

andyleiserson:
In particular, this makes there be only one copy of the code that processes the input reports for an IPA query.

### DIFF
--- a/src/query/executor.rs
+++ b/src/query/executor.rs
@@ -1,5 +1,5 @@
 use crate::{
-    ff::{Field, GaloisField, Serializable},
+    ff::{Field, FieldType, Fp32BitPrime, GaloisField, Serializable},
     helpers::{
         negotiate_prss,
         query::{QueryConfig, QueryType},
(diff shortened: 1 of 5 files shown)


andyleiserson:
This is scaled back a bit from what I had originally tried to do. I tried to make the query processing generic over the return type of the query, i.e., a query would return `Vec<O: Serializable>` instead of `Box<dyn ProtocolResult>`. But I ran into some issues doing that and in the interest of getting what we need now done, removed those changes.

src/query/runner/ipa.rs
```diff
@@ -1,121 +1,89 @@
...
+    ) -> Result<Vec<MCAggregateCreditOutputRow<F, AdditiveShare<F>, BreakdownKey>>, Error> {
+        let Self(config, _) = self;
+
+        let mut input =
+            input.align(<IPAInputRow<F, MatchKey, BreakdownKey> as Serializable>::Size::USIZE);
+        let mut input_vec = Vec::new();
+        while let Some(data) = input.next().await {
+            input_vec.extend(IPAInputRow::<F, MatchKey, BreakdownKey>::from_byte_slice(
+                &data.unwrap(),
+            ));
```
andyleiserson, line 83: The big win is that there's only one copy of this code that does `input.align` and the subsequent processing of the inputs. It's only a few lines right now, but it's about to get a bit more involved.

src/query/executor.rs
```diff
@@ -63,54 +71,104 @@ where
...
+    gateway: Gateway,
+    input: ByteArrStream,
+    query_impl: F,
+) -> JoinHandle<QueryResult>
+where
+    F: for<'a> FnOnce(
+            &'a PrssEndpoint,
+            &'a Gateway,
+            ByteArrStream,
+        ) -> Pin<Box<dyn Future<Output = QueryResult> + Send + 'a>>
```
andyleiserson, line 150: This box is needed only because having two type parameters for the function and the returned future prevents describing the HRTB for the `'a` lifetime.

martinthomson: I'm still amazed at how you can operate with lifetime traits.  I like rust, but hate having to deal with lifetimes.  It's as though the language was designed to suck you in and then trap you in a Kafkaesque nightmare filled with unbalanced quotes.

src/query/executor.rs
```diff
@@ -63,54 +71,104 @@ where
...
 }
 
-#[allow(unused)]
-pub fn start_query(
+pub fn execute(
     config: QueryConfig,
     gateway: Gateway,
     input: ByteArrStream,
 ) -> JoinHandle<QueryResult> {
+    match (config.query_type, config.field_type) {
```
martinthomson, line 79: This is a bit repetitious, but I think that this is a reasonable amount of repetition to manage.

martinthomson (approved):

akoshelev (approved):

PR #688, opened 2023-06-08
Script for report collector to run the protocol (1/n)

richajaindce:
1. Extracts out report collector functions to generate inputs and run query in a separate file
2. run_all script would be used by RC to run tests for all required configurations


src/bin/report_collector.rs
```diff
@@ -0,0 +1,383 @@
...
+    path::PathBuf,
+    time::{Duration, Instant},
+};
+use tokio::time::sleep;
+
+#[derive(Debug, Parser)]
+#[clap(
+    name = "mpc-client",
+    about = "CLI to execute test scenarios on IPA MPC helpers"
+)]
```
akoshelev, line 38:
```suggestion
#[clap(
    name = "rc",
    about = "Report Collector CLI"
)]
```

src/bin/report_collector.rs
```diff
@@ -0,0 +1,383 @@
...
+        if let Some(ref file_name) = source.input_file {
+            InputSource::from_file(file_name)
+        } else {
+            InputSource::from_stdin()
+        }
+    }
+}
+
+#[derive(Debug, Subcommand)]
+enum TestAction {
```
akoshelev, line 86:
```suggestion
enum ReportCollectorCommand {
```

src/bin/report_collector.rs
```diff
@@ -0,0 +1,383 @@
...
+    breakdowns: u32,
+}
+
+async fn clients_ready(clients: &[MpcHelperClient; 3]) -> bool {
+    clients[0].echo("").await.is_ok()
+        && clients[1].echo("").await.is_ok()
+        && clients[2].echo("").await.is_ok()
+}
+
+fn validate<I, S>(expected: I, actual: I)
```
akoshelev, line 121: can we move it to `playbook` module and re-use in both test_mpc and rc binaries?

in-market-test/run_queries.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
akoshelev, line 1: would be great to have a comment here explaining the usage

Cargo.toml
```diff
@@ -133,6 +133,11 @@ path = "src/bin/ipa_bench/ipa_bench.rs"
 required-features = ["cli", "enable-serde"]
 bench = false
 
+[[bin]]
+name = "report_collector"
+required-features = ["cli", "test-fixture", "web-app", "weak-field"]
```
akoshelev, line 138: not sure if we want to expose `Fp31` to the clients, but I don't think we can fix it easily now. Maybe we can follow up with removing it from `test-fixture` and `rc` binary?

in-market-test/run_queries.sh
```diff
@@ -0,0 +1,42 @@
...
+LOGS_PATH=~/SyntheticTest/logs
+
+
+mkdir -p $DATA_PATH
+mkdir -p $LOGS_PATH
+
+echo "Generating input data"
+cd $IPA_REPO
+for (( n=0; n<${#query_sizes[@]}; n++ )); do
+  cargo run --bin report_collector --features="cli test-fixture" -- gen-ipa-inputs -n ${query_sizes[$n]} --output-file $DATA_PATH/ipa-events-${query_sizes[$n]}.txt
```
akoshelev, line 22: this is likely running in debug mode. You could also build a binary `cargo build <bin> --release` and run it instead

src/bin/report_collector.rs
```diff
@@ -0,0 +1,383 @@
...
+    let args = Args::parse();
+    let _handle = args.logging.setup_logging();
+
+    let scheme = if args.disable_https {
+        Scheme::HTTP
+    } else {
+        Scheme::HTTPS
+    };
+
+    let network_path = args.network.as_deref();
```
akoshelev, line 179: this can be combined with `test_mpc` logic to initialize clients - maybe just a helper function inside the `cli` module?

src/bin/report_collector.rs
richajaindce: n00b qn: What is the purpose of playbook? What kind of code should go there?

Cargo.toml
richajaindce: Removed dependency on Fp31 atleast

src/bin/report_collector.rs
richajaindce: good point, let me do that

akoshelev (approved):

Issue #685, opened 2023-06-07
Malicious IPA hangs in real world setup with 3 helpers


akoshelev:
I have set up 3 machines that represent helper parties and another one that plays the role of report collector (test_mpc). Helpers are using Docker to run IPA code, RC just straight calls cargo to run the test.

The results are disappointing. Semi-honest IPA scales up to 1M event with no issues. Malicious IPA breaks somewhere between 1000 and 5000 input reports. 

This will be a tracking issue to fix this bug

akoshelev:
@martinthomson suggested to use artificially high maximum number of HTTP2 streams (10k from the current 4k)

akoshelev:
First thing I am going to try is to run this without Docker

akoshelev:
fun stuff - forgot that you need to dance around permissions to allow binding to low numbered ports

```bash
./target/release/helper --network network.toml --identity 3 --tls-cert h3.pem --tls-key h3.key --port 443
2023-06-07T23:48:13.006296Z  INFO ipa::cli::verbosity: Logging setup at level info
thread 'main' panicked at 'Failed to bind server to a port', /home/ec2-user/workspace/ipa/src/net/server/mod.rs:189:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'Failed to serve: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', /home/ec2-user/workspace/ipa/src/net/server/mod.rs:243:18
[ec2-user@ip-172-31-56-177 ipa]$ sudo setcap CAP_NET_BIND_SERVICE=+eip ./target/release/helper
```

akoshelev:
having the same issue without docker as well. we hang at user capping

```bash
2023-06-07T23:58:10.857052Z  INFO ipa_query:attribute:apply_window: ipa::protocol::attribution::apply_attribution_window: close time.busy=501µs time.idle=18.9µs
2023-06-07T23:58:10.857068Z  INFO ipa_query:attribute:accumulate_credit: ipa::protocol::attribution::accumulate_credit: new
2023-06-07T23:58:11.804931Z  INFO ipa_query:attribute:accumulate_credit: ipa::protocol::attribution::accumulate_credit: close time.busy=527ms time.idle=421ms
2023-06-07T23:58:11.804955Z  INFO ipa_query:attribute:user_capping: ipa::protocol::attribution::credit_capping: new
...
nothing is happening after
```

[...]

PR #690, opened 2023-06-09
Remove the limit on number of concurrent streams

akoshelev:
See #685 for details. We certainly need more than 256 streams. Having an upper limit close to that number leaves us vulnerable to having this issue again in the future.

So I'd rather not limit it and rely on performance metrics if we start seeing impact of too many streams open.

I also don't see an easy way to test this, so tests are not included :(

### DIFF
--- a/src/net/server/mod.rs
+++ b/src/net/server/mod.rs
@@ -232,11 +232,7 @@ where
         async move {
             server
                 // TODO: configuration
-                .http_config(
-                    HttpConfig::default()
-                        .http2_max_concurrent_streams(Some(256))
-                        .build(),
-                )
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Have you confirmed that there is no limit by default?  The HTTP/2 spec default is unlimited, but hyper might have its own default.

akoshelev:
> Have you confirmed that there is no limit by default? The HTTP/2 spec default is unlimited, but hyper might have its own default.

yea that's what hyper documentation states: https://docs.rs/hyper/latest/hyper/server/conn/struct.Http.html#method.http2_max_concurrent_streams

PR #692, opened 2023-06-09
Decrypt match key shares in input reports

andyleiserson:
This change is on top of #689.

I am going to see if I can get encrypted match keys working with test_mpc.


andyleiserson:
This now has support for submitting encrypted match keys from test_mpc and the `helper_networks` tests are passing. However, some unit tests are failing because of the static key registry.

Cargo.toml
```diff
@@ -44,21 +44,19 @@ axum = { version = "0.5.17", optional = true, features = ["http2"] }
...
-# TODO: the only place where we use bytes is in ByteArrStream. ordering_mpsc is a replacement for it
-# so we can remove this dependency after migrating to it. It is added only to avoid having `web-app` feature enabled
-# by default.
 bytes = "1.4"
 clap = { version = "4.0", optional = true, features = ["derive"] }
 comfy-table = { version = "7.0", optional = true }
 config = "0.13.2"
 criterion = { version = "0.5.1", optional = true, default-features = false, features = ["async_tokio", "plotters", "html_reports"] }
 dashmap = "5.4"
+derivative = "2.2"
```
akoshelev, line 53: is it worth it and does it make our compilation time significantly worse or no?

src/helpers/transport/stream/axum_body.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl Stream for WrappedAxumBodyStream {
+    type Item = <AxumInner as Stream>::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.project().0.poll_next(cx)
+    }
+}
+
+// TODO: I would prefer that this be cfg(test), but it's needed in the CLI tools.
```
akoshelev, line 40: `[cfg(feature = 'test-fixture')]`?

src/helpers/transport/stream/box_body.rs
```diff
@@ -0,0 +1,53 @@
+use crate::helpers::BytesStream;
+
+use futures::Stream;
+use pin_project::pin_project;
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+type BoxInner = Pin<Box<dyn BytesStream>>;
```
akoshelev, line 10: Probably should be moved to `BytesStream` definition and renamed to `BoxedBytesStream`?

src/helpers/transport/stream/box_body.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[pin_project]
+pub struct WrappedBoxBodyStream(#[pin] BoxInner);
```
akoshelev, line 13: it is already pinned, so no need to use pin project here

src/helpers/transport/stream/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+pub use axum_body::WrappedAxumBodyStream;
+pub use box_body::WrappedBoxBodyStream;
+pub use collection::{StreamCollection, StreamKey};
+pub use input::{LengthDelimitedStream, RecordsStream};
+
+use crate::error::BoxError;
+use bytes::Bytes;
+use futures::Stream;
+
+pub trait BytesStream: Stream<Item = Result<Bytes, BoxError>> + Send {
```
akoshelev, line 17: is it worth making it generic over a reference to [u8]?

src/hpke/registry.rs
```diff
@@ -41,12 +42,56 @@ impl KeyPair {
...
+
+impl Deref for PublicKeyOnly {
+    type Target = IpaPublicKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+pub trait PublicKeyRegistry {
```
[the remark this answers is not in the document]
andyleiserson: One nice thing about the key registry abstraction (even for managing public keys) is that it provides some assurance the correct key ID is used. A seal API that passes key ID and `&key` directly has more risk that the passed key ID doesn't match the passed key.

However it may make sense to have the lowest level seal API take key ID and key separately, and then provide higher level wrappers for particular use cases. We may want to add key IDs to `NetworkConfig` and then we could provide a seal API that takes `NetworkConfig` rather than key IDs / keys.

src/helpers/transport/stream/input.rs
```diff
@@ -0,0 +1,834 @@
...
+    future::Ready,
+    io,
+    marker::PhantomData,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use typenum::{Unsigned, U2};
+
+#[derive(Debug)]
+pub struct BufDeque {
```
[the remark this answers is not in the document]
andyleiserson: Yes, this file came from aligned `ByteArrStream` but I refactored it a bit so `BufDeque` could be reused by the fixed- and variable-length stream adapters.

src/hpke/registry.rs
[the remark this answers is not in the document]
akoshelev: agree, I wouldn't want to pass both key id and `&key` -  that defeats the purpose. I am on the same page wrt to configuration, although I might prefer a wrapper around `NetworkConfig` to work with encryption - but essentially it will use just client configuration to encrypt

akoshelev (approved):

src/query/processor.rs
```diff
@@ -38,9 +41,12 @@ use tokio::sync::oneshot;
 /// that initiated this request asks for them.
 ///
 /// [`AdditiveShare`]: crate::secret_sharing::replicated::semi_honest::AdditiveShare
-#[derive(Default)]
+#[derive(Derivative)]
+#[derivative(Default)]
 pub struct Processor {
     queries: RunningQueries,
+    #[derivative(Default(value = "Arc::new(KeyRegistry::<KeyPair>::empty())"))]
```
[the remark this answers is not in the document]
andyleiserson: I understand the general concern about unnecessary dependencies, but I don't understand the specific concern here -- the expression for the default value is a little unwieldy but it still seems easier to me than writing out the full impl (which would also need to include this expression).

I don't have numbers for how much it impacts compile time. I don't think it makes sense to pull in a crate to derive 1 or 2 impls, but I think there are quite a few other manual impls that could be replaced with `derivative`. One case where derivative comes in useful is when there's one field in a struct that is secret so shouldn't be included in the debug representation -- derivative makes it easy to say "skip this field".

I also think it's more defensible to omit tests for a derivative-based representation than it is for a manual implementation of Debug or Display.

src/helpers/transport/stream/mod.rs
andyleiserson: That does seem like it could be useful, but I am inclined not to spend time on it until there is a more specific need. I'm worried that there may be places where handing off our generic version to something that strictly requires `Bytes` is problematic.

Cargo.toml
andyleiserson: I have removed derivative in the latest commit.

src/query/processor.rs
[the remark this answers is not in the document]
akoshelev: I like single-line statements but not when they end up as a large string even if this string is compile-time verified. It breaks IDE syntax formatting and navigation features and I don't see a clear advantage over just `impl Default`. It is a bit annoying to write, but it does not happen for every struct so I think it is tolerable. 

Most of the features `derivatives` bring are not super annoying to implement by hand imo.

PR #696, opened 2023-06-09
Fix a few tests that weren't getting run in CI

andyleiserson:


akoshelev:
Fixing the broken build, this change has nothing to do with it

akoshelev (approved):

PR #703, opened 2023-06-13
Add query size parameter to IPA queries

akoshelev:
IPA needs to know how many records will be submitted inside a query. It is important for multiple reasons, one of them is to be able to tune the size of send/receive buffer


src/helpers/transport/query.rs
```diff
@@ -69,6 +64,46 @@ impl RouteParams<RouteId, NoQueryId, NoStep> for &QueryConfig {
...
+    where
+        <S as TryInto<usize>>::Error: Debug,
+        S: TryInto<usize> + TryInto<u32>,
+    {
+        let sz: usize = size.try_into().expect("query size fits into machine word");
+        let sz = u32::try_from(sz)
+            .and_then(NonZeroU32::try_from)
+            .map_err(|_| QueryConfigError::BadQuerySize {
+                actual_size: sz,
+                max_size: MAX_QUERY_SIZE,
```
martinthomson, line 96: Are you relying on `MAX_QUERY_SIZE` being equal to `NonZeroU32::MAX`? 

... and I just saw that it is not.  We'll need a new type for this, with a `TryFrom` implementation that doesn't accept 2<sup>31</sup>, as this does.

src/net/http_serde.rs
```diff
@@ -109,10 +109,13 @@ pub mod query {
         async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
             #[derive(serde::Deserialize)]
             struct QueryTypeParam {
+                // TODO: serde custom error?
+                size: NonZeroU32,
```
martinthomson, line 113: We might need a new type for this is `MAX_QUERY_SIZE` changes.

martinthomson (approved):
LGTM, though please fix the bug I identified (or open an issue for it, we're not likely to need 2 billion inputs for a while).

src/net/http_serde.rs
akoshelev: added a new type `QuerySize` - made it cleaner

PR #711, opened 2023-06-16
new gate generated by Gate derive macro

taikiy:
This is the base of new memory-efficient gate to be annotated by Gate derive macro.

Step and StepNarrow implementations will be synthesized by Gate macro, but there are other things that need to be defined in addition.


src/protocol/boolean/bitwise_equal.rs
```diff
@@ -107,7 +107,7 @@ where
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-enum Step {
+pub enum Step {
```
akoshelev, line 110: lets make it consistent with other step definitions (pub(crate))

src/protocol/step/compact.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+//
+// steps used in tests
+//
+
+#[cfg(any(feature = "test-fixture", debug_assertions))]
+impl StepNarrow<str> for Compact {
+    fn narrow(&self, step: &str) -> Self {
+        Self(static_state_map(self.0, step))
```
akoshelev, line 99: static allows mapping to root step, fallback for RGB and upgrade. Which branch is picked for test steps?

tests/compact_gate.rs
```diff
@@ -0,0 +1,10 @@
+use std::process::Command;
+
+#[test]
+fn compact_gate() {
```
akoshelev, line 4: this takes ~1m on my machine each time I run `cargo test`. That is probably too slow for a test that runs that often. We should push most of them to the CI and run just a couple locally

src/protocol/step/compact.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[cfg(any(feature = "test-fixture", debug_assertions))]
+impl StepNarrow<String> for Compact {
+    fn narrow(&self, step: &String) -> Self {
+        Self(static_state_map(self.0, step.as_str()))
+    }
+}
+
+#[cfg(any(feature = "test-fixture", debug_assertions))]
+impl StepNarrow<crate::helpers::prss_protocol::PrssExchangeStep> for Compact {
```
akoshelev, line 111: what these are for?

src/protocol/context/semi_honest.rs
```diff
@@ -194,6 +214,31 @@ impl<'a, F: ExtendableField> UpgradedContext<F> for Upgraded<'a, F> {
         Ok(x)
     }
 
+    // Moved from "context/mod.rs" because we want to use the local `UpgradeContext`.
+    // `upgrade` for semi-honest used to be narrowed but never executed. That causes
+    // `Compact` gate to panic because it doesn't know anything about state transitions
+    // that are not executed. To workaround this, we use a dummy step to narrow the
+    // context to let `Compact` gate know that it could be ignored.
```
akoshelev, line 221: do we still need to add `upgrade` to the list of "static" mappings after thins change?

scripts/test_compact_gate.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
taikiy: I agree. I'll work on this immediately after this PR.

tests/compact_gate.rs
taikiy: Did it run in the `pre-commit`? I thought I had this only in CI and not pre-commit.

src/protocol/step/compact.rs
taikiy: Any struct/enum that implements `Step` must implement `StepNarrow`, whether it's executed or not. Actually, I think I should remove `#[cfg]` statement here.

src/protocol/context/semi_honest.rs
taikiy: That is correct.

I haven't made the change we tried when we were in London. That turned out to not work because there are more narrows after `upgrade` (in `UpgradeToMalicious` implementations before the no-op functions `UpgradedContext::upgrade_[one|sparse]` are called).

src/protocol/step/compact.rs
taikiy: The first call to `narrow` from TestWorld will have step string "run-0", and the gate state ID is 0, so it's mapped to the root step. Once this is matched, subsequent `narrow` calls will/should not reach `static_state_map` (unless the step is one of dummy or conditional like semi-honest upgrade and fallback). The narrowed step should match one of "real" protocol steps in one of `StepNarrow` implementations.

Obviously, this won't work in unit tests because but that's expected because we only use descriptive gate with unit tests, and we don't generate full steps for unit tests to begin with. So, the root step mapping in `static_state_map` is there specifically for this case.

src/protocol/boolean/bitwise_equal.rs
taikiy: fixed!

tests/compact_gate.rs
akoshelev: I think I saw this by just running `cargo test`

src/helpers/mod.rs
```diff
@@ -6,7 +6,7 @@ use std::{
 mod buffers;
 mod error;
 mod gateway;
-mod prss_protocol;
+pub mod prss_protocol;
```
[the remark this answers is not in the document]
taikiy: fixed

src/protocol/step/compact.rs
```diff
@@ -0,0 +1,190 @@
...
+        return QUERY_TYPE_MALICIOUS_STATE;
+    } else if step == crate::helpers::prss_protocol::PrssExchangeStep.as_ref() {
+        return PRSS_EXCHANGE_STATE;
+    }
+
+    panic!("cannot narrow with \"{step}\" from state {state}");
+}
+
+// Reverse of `static_state_map` for `Compact::as_ref()`
+fn static_reverse_state_map(state: u16) -> &'static str {
```
akoshelev, line 75: I think you rely on this function somewhere inside the macro, which makes this dependency a bit more tightly coupled that ideal. Not sure if there is a better way though

src/protocol/step/compact.rs
```diff
@@ -0,0 +1,190 @@
...
+
+// steps that do not trigger communications
+
+impl StepNarrow<crate::protocol::context::semi_honest::UpgradeStep> for Compact {
+    // `src/protocol/context/semi_honest.rs`
+    // Semi-honest implementations of `UpgradedContext::upgrade()` and subsequent
+    // `UpgradeToMalicious::upgrade()` narrows but these will end up in `UpgradedContext::upgrade_one()`
+    // or `UpgradedContext::upgrade_sparse()` which both return Ok() and never trigger communications.
+    fn narrow(&self, step: &crate::protocol::context::semi_honest::UpgradeStep) -> Self {
+        Self(static_state_map(self.0, step.as_ref()))
```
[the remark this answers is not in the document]
taikiy: I wanted to hide these inside compact.rs because, as you said, these are somewhat "magic". These are kind of temporary solution until we have a new macro that supports annotations for "not-executed steps". Defining a blanket implementation doesn't sound right to me for the same reason because descriptive gate also implements StepNarrow trait.

src/protocol/step/compact.rs
taikiy: I could move everything inside the proc macro so that the IPA core code looks a bit more cleaner, but the dependency is still there. I hope to find a better solution with a new macro but the synchronization of assigning a unique state ID for these special steps would be tricky.

src/protocol/step/compact.rs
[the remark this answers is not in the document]
akoshelev: oh what I meant was something along these lines

```rust
trait NoCommsStep : AsRef<str> {}

impl <C: Step + NoCommsStep> StepNarrow<C> for Compact {
    fn narrow(&self, step: &C) -> Self {
        Self(static_state_map(self.0, step.as_ref()))
    }
}

impl NoCommsStep for crate::protocol::context::semi_honest::UpgradeStep {}
impl NoCommsStep for crate::protocol::query::QueryType {}
```

just cosmetics but helps keeping `static_state_map` usage in one place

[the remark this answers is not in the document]
taikiy: Ah I see. Thanks! Modified as suggested.

akoshelev (approved):
we need to start moving the compact gate into our main branch.

PR #712, opened 2023-06-17
Add cfg_aliases and two aliases: unit-test and web-test

akoshelev:
For some reason I could not use `test` inside the alias definition, it never enables the alias when `cargo test`. Other than that, it should reduce the repetitive definitions. We are going to have more once `descriptive-gate` will be a required feature for unit tests and we would run some of them with compact as well

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,8 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    cfg_aliases! {
+        unit_test: { all(not(feature = "shuttle"), feature = "in-memory-infra") },
+        web_test: { all(not(feature = "shuttle"), feature = "real-world-infra") },
+    }
+}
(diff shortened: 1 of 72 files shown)


akoshelev:
Testing whether we still run the same set of tests after this change

```bash
test result: ok. 296 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 27.11s
```

after

```bash
test result: ok. 301 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 21.81s

     Running tests/tempdir.rs (target/debug/deps/tempdir-70484b18fb03750d)
```

looks like I've enabled some!

build.rs
```diff
@@ -0,0 +1,8 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    cfg_aliases! {
```
andyleiserson, line 4: Maybe add a comment mentioning that `cfg(test)` doesn't work here?

andyleiserson (approved):

PR #714, opened 2023-06-19
Introduce workspace

taikiy:
Our repo has grown, and with the recent addition of `ipa-macros`, it's a good time to introduce workspace.

Changes:
- Moved (ipa main lib) `/src` to `/ipa/src`
- Moved `/ipa/src/bin/helper` to `/helper`
- Moved integration tests `/tests` to `/helper/tests`
- Fix clippy errors
- Remove unused dependencies

I kept files changes to minimum. Once the code is landed, we could move `web-app` related stuff under `/helper` or on its own crate folder.


ipa/Cargo.toml
```diff
@@ -0,0 +1,141 @@
+[package]
+name = "ipa"
+version = "0.1.0"
+rust-version = "1.64.0"
+edition = "2021"
+
+[features]
+# by default remove all TRACE, DEBUG spans from release builds
+default = ["web-app", "tracing/max_level_trace", "tracing/release_max_level_info"]
```
taikiy, line 9: Removed "in-memory-infra" from `default`. Dependent external crates will specify either `in-memory-infra` or `real-world-infra` in their own Cargo.toml.

akoshelev: this broke most of our unit tests - I checked the CI and we only run 55 tests because of it

.github/workflows/check.yml
```diff
@@ -1,19 +1,19 @@
...
 on:
   push:
-      paths:
-      - '**/Cargo.toml'
-      - '**/Cargo.lock'
-      - 'src/**/*'
-      - 'benches/**/*'
-      - 'tests/**/*'
+    paths:
+      - "**/Cargo.toml"
```
akoshelev, line 5: wow, was it broken because of workspace introduction?

helper/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+clap = { version = "4.0", optional = true, features = ["derive"] }
+comfy-table = { version = "7.0", optional = true }
+hyper = { version = "0.14.26", optional = true, features = ["client", "h2", "stream"] }
+hyper-tls = { version = "0.5.0", optional = true }
+# TODO(640): sec1 only needed until we standardize on rustls
+sec1 = { version = "0.7", features = ["alloc", "pem", "pkcs8"], optional = true }
+time = { version = "0.3", optional = true }
+tokio = { version = "1.28", features = ["rt", "rt-multi-thread", "macros"] }
+tracing = "0.1.37"
+generic-array = "0.14.7"
```
akoshelev, line 24: we could specify common dependencies inside the workspace toml file: https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

Cargo.toml
```diff
@@ -1,107 +1,10 @@
...
-disable-metrics = []
-# TODO move web-app to a separate crate. It adds a lot of build time to people who mostly write protocols
-# TODO Consider moving out benches as well
-web-app = [
-    "axum", "axum-server", "base64", "clap", "comfy-table", "enable-serde", "hyper", "hyper-tls",
-    "rcgen", "rustls-pemfile", "sec1", "time", "tokio-rustls", "toml", "tower", "tower-http",
+members = [
+    "ipa",
+    "ipa-macros",
+    "helper",
```
akoshelev, line 6: I kinda like having it, but curious what @martinthomson thinks. Macros have to be in a separate crate, so that alone justifies the workspace creation, isolating helper service may not be strictly required for now.

helper/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "helper"
```
akoshelev, line 2: maybe something like `ipa-mpc-helper`? Or brainstorm for a better name?

helper/Cargo.toml
taikiy: `ipa` and `helper` have many in common, but `ipa-macros` only depend on syn and quote.

helper/Cargo.toml
taikiy: +1 on `ipa-mpc-helper`. Or we could remove "ipa" and make them just `macros` and `mpc-helper`.

ipa/Cargo.toml
taikiy: oops, thanks for catching. Fixed.

.github/workflows/check.yml
taikiy: It was because I missed specifying `in-memory-infra` to `cargo test`. I updated `check.yml` too because I think it needs `*/` in `paths` now.

akoshelev: that was the question about single-quotes changed to double quotes :)

helper/Cargo.toml
akoshelev: it seems to be common to keep the prefix: https://github.com/rust-random/rand

pre-commit
```diff
@@ -77,25 +77,28 @@ check() {
...
+    cargo clippy -p ipa --tests --features="shuttle in-memory-infra" -- -D warnings
 
 check "Tests" \
     cargo test
 
+check "IPA Tests" \
+    cargo test -p ipa --features in-memory-infra
+
 check "Web tests" \
     cargo test --no-default-features --features "cli web-app real-world-infra test-fixture"
```
akoshelev, line 95: q: does it run both ipa and ipa-helper tests?

helper/Cargo.toml
akoshelev: and maybe ipa should become `ipa-core`

pre-commit
taikiy: This one only runs ipa tests. ipa-helper tests are run in "Web tests".

With "IPA tests", pre-commit covers all unit tests, but there are some overwrap in "Tests" and "IPA Tests". For example, unit tests in `ipa/src/exact.rs` have `#[cfg(all(test, not(feature = "shuttle")))]` which is true in both cases. We could add `not(in-memory-infra)` if we want to avoid the overwrap, but seems a hacky way of fixing this.

taikiy:
Closing this one as it's been several months and it'd be difficult to resolve conflicts. We can revisit this in the future if needed.

Issue #701, opened 2023-06-12
Gracefully handle query initialization errors


akoshelev:
Currently IPA can recover if query execution fails at the protocol stage (i.e. when we actually start running IPA). However, if there was any error before that (that is - create_query or prepare_query failed), any subsequent attempt to run another query fails with "Query is already running error". 

We have a drop guard already that removes the query from the list, we probably just need to change that to support another mode to clean up only if thread is panicking.

richajaindce:
This is a bit annoying while running tests. Lets fix this before the test run @akoshelev. If you point me to an example of how to fix it, I can take it up

akoshelev:
Do you happen to have a stack trace? I am not sure where we are panicking right now

akoshelev:
One scenario I am aware of - when helper configuration is bad (wrong certificate, etc), an  attempt to run a query results in an SSL error. After that, helpers are unable to run queries. This is likely an unrecoverable error because it requires a bounce to read the new configuration, but deserves a quick fix anyway

richajaindce:
It also happened when helpers were started without matchkey encryption keys and I tried to run a query with encryption enabled.

PR #715, opened 2023-06-19
Make new_query recover from communication errors

akoshelev:
Currently if H1 failed to communicate with H2 or H3, it remembers the query forever. This change makes it forget about the query it failed to prepare. Addresses #701


richajaindce (approved):
LGTM

Issue #640, opened 2023-05-10
Pick a TLS stack and use it consistently


andyleiserson:
We are currently using both rustls (for server-side TLS) and native-tls (for client-side TLS). For the helpers, I think we should use one TLS stack consistently. If we distribute a client for the RC APIs, that might use a different stack than the helpers.

It appears that the axum-server crate we are using only supports rustls, but there is a separate issue (#148) for possibly moving away from axum.

akoshelev:
is there anything that may influence our decision towards either of those crates?

martinthomson:
rustls seems to have more momentum.  I'd prefer it if we go with that.

PR #716, opened 2023-06-20
Remove hyper-tls and replace it with hyper-rustls

akoshelev:
Addresses #640

### DIFF
--- a/src/bin/helper.rs
+++ b/src/bin/helper.rs
@@ -1,6 +1,5 @@
 use clap::{self, Parser, Subcommand};
 use hyper::http::uri::Scheme;
-use hyper_tls::native_tls::Identity;
 use ipa::{
     cli::{
         client_config_setup, keygen, test_setup, ConfGenArgs, KeygenArgs, TestSetupArgs, Verbosity,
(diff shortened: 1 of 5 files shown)


martinthomson (approved):
@andyleiserson is closer to this, but I can't see a downside here.  Especially given the removal of PKCS\#8 haxx for macs.

src/bin/helper.rs
```diff
@@ -152,14 +102,9 @@ async fn server(args: ServerArgs) -> Result<(), BoxError> {
...
             let key = fs::read_to_string(&key_file).unwrap().into_bytes();
-            #[cfg(target_os = "macos")]
-            let key = pkcs8::munge_private_key(&key);
-
             (
-                ClientIdentity::Certificate(
-                    Identity::from_pkcs8(fs::read_to_string(&cert).unwrap().as_bytes(), &key)
-                        .unwrap(),
-                ),
+                ClientIdentity::from_pks8(fs::read_to_string(&cert).unwrap().as_bytes(), &key)
```
martinthomson, line 106: Do you want to use `expect()` here to report errors?

src/net/client/mod.rs
```diff
@@ -102,31 +133,46 @@ impl MpcHelperClient {
...
+                    store
+                };
+
+                let builder = builder.with_root_certificates(cert_store);
                 match identity {
-                    ClientIdentity::Certificate(identity) => {
-                        builder.identity(identity);
-                    }
+                    ClientIdentity::Certificate((cert, pk)) => builder
+                        .with_single_cert(vec![cert], pk)
```
martinthomson, line 155: I only just noticed this, but there is a potential risk here in that a PKCS\#8 file containing a certificate path (i.e., more than just the end-entity certificate) is not used.  I think that the right answer is to retain `Vec<Certificate>` instead.  The trust store in `cert_store` is fine (the end-entity certificate is all we need for that).

src/net/server/mod.rs
```diff
@@ -533,6 +540,22 @@ mod e2e_tests {
...
+        fn verify_server_cert(
+            &self,
+            _end_entity: &Certificate,
+            _intermediates: &[Certificate],
+            _server_name: &ServerName,
+            _scts: &mut dyn Iterator<Item = &[u8]>,
+            _ocsp_response: &[u8],
+            _now: SystemTime,
+        ) -> Result<ServerCertVerified, rustls::Error> {
+            Ok(ServerCertVerified::assertion())
```
martinthomson, line 555: Don't we need dangerous_configuration enabled in rustls for this to work?

martinthomson (commented):
A few more things to think about perhaps.

src/net/server/mod.rs
akoshelev: that only brings a special `DangerousClientConfig` with `set_certificate_verifier` only. does not seem to be worth it

andyleiserson: `axum-server` enables it. (Not saying that is right or that we should rely on it, but that's what's going on.)

akoshelev: eh, true I missed that. let me be explicit about it in Cargo.toml

```
#[allow(unreachable_pub)]
#[derive(Debug)]
#[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
pub struct ServerCertVerified(());
```

PR #717, opened 2023-06-20
Timing only the IPA run and not encryption

richajaindce:


src/cli/playbook/ipa.rs
```diff
@@ -115,7 +116,11 @@ where
                 .collect::<Vec<_>>()
         })
         .reconstruct();
-
+    tracing::info!(
+        "Query size {:?} took {t:?}",
+        query_size,
```
martinthomson, line 121:
```suggestion
        "Query size {query_size:?} took {t:?}",
```

martinthomson (approved):

src/cli/playbook/ipa.rs
martinthomson: Or 
```suggestion
        "Running IPA for {query_size:?} records took {t:?}",
```

akoshelev (approved):

PR #718, opened 2023-06-20
Github action to build helper docker base image

richajaindce:
This builds base docker image of a helper using the current IPA code. We will then use this image to create helper node specific images.


richajaindce:
I dont have permissions to run this action on main repo hence the failure. @martinthomson @akoshelev what is the best way to unblock? For reference, it does work on my personal repo : https://github.com/richajaindce/raw-ipa/actions/runs/5324489087

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+        
+    - name: Build and Publish GitHub Actions Image
+      uses: docker/build-push-action@v4
+      with:
+        file: docker/baseimage/Dockerfile
+        push: true
+        tags: ghcr.io/${{ github.actor }}/ipa-helper-base:latest
```
akoshelev, line 34: I would just name it `ipa-helper:latest`. we would have `ipa-client:latest` at some point I guess too.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+
+  build:
+
```
akoshelev, line 12:
```suggestion
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+
```
akoshelev, line 10:
```suggestion
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: Docker Image CI
```
akoshelev, line 1:
```suggestion
name: Publish Helper Image CI
```

docker/baseimage/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# syntax=docker/dockerfile:1
+ARG SOURCES_DIR=/usr/src/ipa
+FROM rust:latest as builder
+ARG SOURCES_DIR
+LABEL maintainer="akoshelev"
+
+# Prepare helper binaries
+WORKDIR "$SOURCES_DIR"
+COPY . .
```
akoshelev, line 9: this may be ok in this place, but this basically transfers the whole directory as docker context, so it may take a while if there are build artifacts

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+      uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+        
+    - name: Build and Publish GitHub Actions Image
+      uses: docker/build-push-action@v4
+      with:
+        file: docker/baseimage/Dockerfile
```
akoshelev, line 32: docker/ci/helper.Dockerfile maybe?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
```
akoshelev, line 6: you probably don't want to publish a new artifact a new image on pull requests. Only valid push to main should constitute a new image build

I also like Martin's suggestion to run this at random time occasionally

```yaml
  schedule:
    # Once monthly at a randomly selected time.
    - cron: "24 2 3,18 * *"
```

martinthomson: I would like to see more restrictions on the push condition so that we don't generate a build for every push to main (which is every merged PR).  Limiting builds to those that change the docker files or Cargo.toml could help.

Maybe add a workflow_dispatch condition so that we can ask for the image to be made manually instead.

akoshelev: don't we need a fresh image each time we change IPA code because it carries the helper binary?

richajaindce: Agree with Alex that we should generate docker image each time helper code changes so we have the latest helper image.

docker/baseimage/Dockerfile
richajaindce: Lets reduce the size in later PRs

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish Helper Image CI
+
+on:
+  push:
```
martinthomson, line 4:
```suggestion
  workflow_dispatch:
  push:
```

martinthomson (commented):
I've enabled the necessary actions on the organization, so this should run now.

.github/workflows/docker.yml
richajaindce: We would need a fresh docker image each time only for testing purpose. This can be done manually using workflow dispatch. When in future, we run a scheduled test at a fixed cadence, we can change the cron to generate a new image each time. This should be sufficient for now and we can revisit it later.

docker/ci/helper.Dockerfile
```diff
@@ -0,0 +1,22 @@
+# syntax=docker/dockerfile:1
+ARG SOURCES_DIR=/usr/src/ipa
+FROM rust:latest as builder
+ARG SOURCES_DIR
```
martinthomson, line 4: Duplicate ARG tag.

docker/ci/helper.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+LABEL maintainer="akoshelev"
```
martinthomson, line 5: You want this on the second image, not this disposable one.

docker/ci/helper.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Prepare helper binaries
+WORKDIR "$SOURCES_DIR"
+COPY . .
+RUN set -eux; \
+    cargo build --bin helper --release --no-default-features --features "web-app real-world-infra"
+
+# Copy them to the final image
+FROM debian:bullseye-slim
+ENV HELPER_BIN_PATH=/usr/local/bin/ipa-helper
+ENV CONF_DIR=/etc/ipa
```
martinthomson, line 16: This isn't created here, so I would instead include that in the entrypoint script (or as a command line flag), once that is available.

martinthomson (approved):
A few nits on the docker image.  The lack of an entrypoint remains an issue, but I want to see if this works properly.  Merging it will help do that.

PR #721, opened 2023-06-21
DP parameters

martinthomson:
This is a sketch of how things might work for configuring DP.  Needs rebasing.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -2,7 +2,10 @@ use clap::Parser;
 use ipa::{
     error::Error,
     ff::Fp32BitPrime,
-    helpers::{query::IpaQueryConfig, GatewayConfig},
+    helpers::{
+        query::{DifferentialPrivacy, IpaQueryConfig},
+        GatewayConfig,
+    },
     test_fixture::{
         ipa::{
             generate_random_user_records_in_reverse_chronological_order, test_ipa,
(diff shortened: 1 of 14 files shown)


andyleiserson:
Maybe this is bikeshed territory, but does it make sense to have a more generic way to specify a privacy mechanism and its parameters? I'm thinking something like `privacy: Option<PrivacyMechanism>` where `PrivacyMechanism` is an enum, instead of `dp: Option<DifferentialPrivacy>`. I feel like we may pretty quickly want to add other options here (e.g. different noise distributions, randomized response, or even simple variants like a specified epsilon combined with an implicit delta, or a way to say "spend x% of my privacy budget for this query").

(I don't totally like my `privacy: Option<PrivacyMechanism>` suggestion because it results in the rather unfortunate `privacy: None`. But I'm not immediately sure how to do better without making it too verbose.)

martinthomson:
I think that you mean `privacy: PrivacyMechanism`.  Then we can do something like:

```rust
enum PrivacyMechanism {
  DifferentialPrivacy(DifferentialPrivacy),
  #[cfg(feature = "no-privacy")]
  None,
}
```

benjaminsavage:
Can we close this PR?

martinthomson:
Yeah, I think we'll need to come back to this, but it's not worth rebasing this now.

PR #723, opened 2023-06-23
Simplify the CI image

martinthomson:
This should suffice.  `COPY . .` caused me real issues locally as well.

I've added an entrypoint too.


richajaindce (approved):
Thanks for helping out!

docker/ci/helper.Dockerfile
```diff
@@ -1,17 +1,13 @@
 # syntax=docker/dockerfile:1
-ARG SOURCES_DIR=/usr/src/ipa
 FROM rust:latest as builder
 
-# Prepare helper binaries
-WORKDIR "$SOURCES_DIR"
-COPY . .
-RUN set -eux; \
-    cargo build --bin helper --release --no-default-features --features "web-app real-world-infra"
+RUN mkdir -p /ipa
```
richajaindce, line 4: Nit: Since /ipa is showing up multiple times, we can consider extracting it into a variable.

martinthomson: I'm not doing that specifically because adding variables creates additional layers in the image, which isn't worth it.  As long as the name is short, typing it is roughly equivalent to typing a variable.

richajaindce (approved):

PR #724, opened 2023-06-23
Make upgrade take an iterator

martinthomson:
The cost of this is that some of the type safety we got from the traits had to be cut (maybe we can restore them, but I'm not smart enough today).


src/protocol/context/upgrade.rs
```diff
@@ -44,17 +41,6 @@ use std::{
...
-///
-/// ```compile_fail
-/// use ipa::protocol::{context::{UpgradeContext, UpgradeToMalicious, UpgradedMaliciousContext as C}, NoRecord, RecordId};
-/// use ipa::ff::Fp32BitPrime as F;
-/// use ipa::secret_sharing::replicated::{
-///     malicious::AdditiveShare as MaliciousReplicated, semi_honest::AdditiveShare as Replicated,
-/// };
-/// // This can't be upgraded with a record-bound context because the record ID
-/// // is used internally for vector indexing.
-/// let _ = <UpgradeContext<C<'_, F>, F, RecordId> as UpgradeToMalicious<Vec<Replicated<F>>, _>>::upgrade;
```
martinthomson, line 56: I couldn't work out how to make this keep failing.  The way it failed before was that we only implemented for `Vec<Vec<T>>`, but now that there is a generic `IntoIterator` version, I'm not sure what we would do to keep this.

src/protocol/context/upgrade.rs
```diff
@@ -431,15 +407,14 @@ where
 // context. This is only used for tests where the protocol takes a single `Replicated<F>` input.
 #[cfg(test)]
 #[async_trait]
-impl<'a, C, F, T, M> UpgradeToMalicious<'a, T, M> for UpgradeContext<'a, C, F, NoRecord>
+impl<'a, C, F, M> UpgradeToMalicious<'a, Replicated<F>, M> for UpgradeContext<'a, C, F, NoRecord>
```
martinthomson, line 410: A smaller implementation worked fine, for all but the modulus conversion tests.  This seemed like a good trade.

src/protocol/context/upgrade.rs
```diff
@@ -220,8 +209,7 @@ where
 /// It assumes the inner vector is much smaller (e.g. multiple bits per record) than the outer vector (e.g. records)
 /// Each inner vector element uses a different context and outer vector shares a context for the same inner vector index
```
andyleiserson, line 210:
```suggestion
```

src/protocol/context/upgrade.rs
andyleiserson: Using a non-`Vec` type that is `IntoIterator` (like `[Replicated<F>; 3]`) would work, but I'm not sure how much value that has. I might be inclined to delete all of these doc tests -- without the negative case, I don't think the positive cases have much value.

andyleiserson (approved):
It makes me a little nervous to have an impl for `Vec<T>` that panics for large vectors -- it appears at the surface like it will work for any Vec, but it won't. `Vec<Vec<T>>` seemed a little safer since that's a less common type to have around.

One easy improvement might be to return an error on attempt to upgrade a long vector rather than panic. Then at least if we end up reaching this case at runtime, the helper doesn't crash.

A fix that I do like but is much more work, is to use a dedicated type for "bit-decomposition of a value that is at most 64 bits". That would also help us get the last bit of storage efficiency for `Vec<MaliciousReplicated<Gf2>>` (although for that case, I'm also hopeful that we can use a larger field).

martinthomson:
@andyleiserson I took a swing at doing the hard thing you mentioned.  It wasn't easy, as you said, but I think that there are real improvements to be had from it.

The main drawback is that I had to move `IntoShares` to take `Iterator` rather than `IntoIterator`.  I'm in two minds about that, but implementing `IntoIterator` for `BitDecomposed` seemed like it was more useful.

src/protocol/context/upgrade.rs
martinthomson: Test restoration incoming.  I forgot to revert this part of the change.

src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -268,19 +277,37 @@ where
...
+impl<T> Downgrade for BitDecomposed<T>
 where
     T: Downgrade,
 {
-    type Target = Vec<<T as Downgrade>::Target>;
+    type Target = BitDecomposed<<T as Downgrade>::Target>;
     async fn downgrade(self) -> UnauthorizedDowngradeWrapper<Self::Target> {
-        // TODO: validate that the vector here is truly small.
         #[allow(clippy::disallowed_methods)]
         let result = join_all(
```
andyleiserson, line 287: This is exactly the kind of thing I was hoping to catch with this change, nice!

andyleiserson (approved):
> The main drawback is that I had to move `IntoShares` to take `Iterator` rather than `IntoIterator`. I'm in two minds about that, but implementing `IntoIterator` for `BitDecomposed` seemed like it was more useful.

Yeah, I agree this is a tough call. If we had both owned and borrowed versions of `BitDecomposed`, then we might not need `BitDecomposed` to deref to a slice, and then maybe we could do without the IntoIterator implementation for `BitDecomposed`. But that's a whole additional set of changes to make.

It might be possible to avoid the `into_iter()` calls in the common case of passing a `Vec` to a test, in exchange for more verbosity in the less common case of passing an iterator to a test, by implementing `IntoShares` for `Vec<U>` and `Box<dyn Iterator<Item = U>>`. (Or for a newtype wrapper around an iterator.) However, the changes to call `into_iter()` are already done, so I don't see a compelling reason to pursue this.

src/secret_sharing/decomposed.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl<S> BitDecomposed<S> {
+    const MAX: usize = 64;
+
+    /// Create a new value from an iterator.
+    /// # Panics
+    /// If the iterator produces more than `Self::MAX` items.
+    pub fn new<I: IntoIterator<Item = S>>(bits: I) -> Self {
+        let bits = bits.into_iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
martinthomson: https://gcc.godbolt.org/z/x9d7We759 seems to support the view that it optimizes out.

src/protocol/sort/secureapplyinv.rs
```diff
@@ -67,10 +95,14 @@ mod tests {
                 .map(u128::from)
                 .map(Fp31::truncate_from);
 
+            // Flatten the input so that it can implement `IntoShares`.
```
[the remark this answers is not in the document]
martinthomson: Nice, I had forgotten that part.

src/secret_sharing/decomposed.rs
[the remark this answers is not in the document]
andyleiserson: [It's nice being able to use unstable features](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/vec/spec_from_iter.rs.html) :)

PR #725, opened 2023-06-23
Breakdown key simplification cleanups

martinthomson:
@benjaminsavage, this might help.  I've added a couple of tweaks (see #724) above yours as well, because I'll be needing those.

### DIFF
--- a/src/protocol/ipa/mod.rs
+++ b/src/protocol/ipa/mod.rs
@@ -26,10 +26,9 @@ use crate::{
     },
 };
 use async_trait::async_trait;
-use futures::future::try_join4;
 use futures_util::future::{try_join, try_join3};
 use generic_array::{ArrayLength, GenericArray};
-use std::{marker::PhantomData, ops::Add};
+use std::{iter::zip, marker::PhantomData, ops::Add};
 use typenum::Unsigned;
 
 use super::context::{UpgradableContext, UpgradedContext};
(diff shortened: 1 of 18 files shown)


src/protocol/ipa/mod.rs
```diff
@@ -322,80 +319,6 @@ where
     }
 }
 
-pub struct IPAModulusConvertedInputRow<F: Field, T: LinearSecretSharing<F>> {
```
martinthomson, line 325: Dead code removal!

src/protocol/ipa/mod.rs
```diff
@@ -492,30 +415,23 @@ where
 
     let arithmetically_shared_values = m_ctx.upgrade(arithmetically_shared_values).await?;
 
-    let binary_shared_values = upgraded_gf2_match_key_bits
-        .iter()
-        .zip(upgraded_gf2_breakdown_key_bits.iter())
-        .map(|(match_key, breakdown_key)| {
-            BinarySharedIPAInputs::new(match_key.clone(), breakdown_key.clone())
-        })
+    let binary_shared_values = zip(upgraded_gf2_match_key_bits, upgraded_gf2_breakdown_key_bits)
```
martinthomson, line 418: Simplifications here.

martinthomson (commented):
Most of this is import cleanups, but I've noted the increments over #725

akoshelev (approved):

src/protocol/ipa/mod.rs
benjaminsavage: Awesome :)

PR #729, opened 2023-06-26
Insecure DP and normal distribution backed by Box-Muller transform

akoshelev:
This builds on top of #702 and contains mostly cosmetic changes, polishing and more tests. The core idea remains the same - use Box-Muller transform to implement normal distribution (alternatively `num_distr`/`statsrs` could've been used) and build a simple DP application mechanism to apply noise in the clear. 

Note that this mechanism is general, i.e. without any specifics related to how aggregates are represented. It does allow negative values after noise is applied. The idea is that the further transformation can be done outside of DP.


src/protocol/dp/distributions.rs
```diff
@@ -0,0 +1,200 @@
...
+    distributions::{uniform::SampleUniform, Distribution, Uniform},
+    Rng,
+};
+
+use std::{
+    fmt::Debug,
+    ops::{Add, Div, Mul, Neg, Sub},
+};
+
+pub trait Float:
```
akoshelev, line 11: I am on the fence on whether we need support something other than `f64`.

martinthomson: I'm not.  `f64` is not just fine, but the only real choice for something like that.  Quad-precision floating point is inefficient and unnecessary and `f32` has too much quantization noise.

src/protocol/dp/distributions.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+#[cfg(all(test, unit_test))]
+mod tests {
+    use super::*;
+    use rand::{distributions::Distribution, thread_rng};
+    use std::iter::Sum;
+
+    fn compute_mean_and_std<F: Float + Sum, A: AsRef<[F]>>(samples: A) -> (F, F) {
```
martinthomson, line 136: It might be sufficient to calculate the variance relative to the expected mean.  In that case, you can make this a one-pass function.  Presumably you are looking to ensure that the value is close.

A one-pass function can take an indefinite iterator and then truncate at an arbitrary count based on whatever it decides.

martinthomson (commented):
A big ticket item: we should be sampling from a discrete distribution, not a (basically) continuous one.  That is how any MPC will report results.  That means rounding.

src/protocol/dp/distributions.rs
```diff
@@ -0,0 +1,200 @@
...
+    ( $( $std_type:ident )+ ) => {
+        $(
+            impl Float for $std_type {
+                const ZERO: Self = 0.0;
+                const ONE: Self = 1.0;
+                const PI: Self = std::$std_type::consts::PI;
+                const MIN_POSITIVE: Self = $std_type::MIN_POSITIVE;
+
+                fn cos(self) -> Self {
+                    $std_type::sin(self)
```
martinthomson, line 58: I predict that this will have a bad outcome at some point.

akoshelev: unless it is always set to be √2/2 which seems to be true for all the tests I've written

bmcase:
> A big ticket item: we should be sampling from a discrete distribution, not a (basically) continuous one. That is how any MPC will report results. That means rounding.

If we want to have noise from a discrete Gaussian or discrete Laplace, this [paper ](https://arxiv.org/pdf/2004.00010.pdf)looks promising.  It has also been [implemented ](https://github.com/IBM/discrete-gaussian-differential-privacy)in python which could make for a helpful reference implementation.

martinthomson:
The implementation of the function looks somewhat involved, at least if we're looking to implement in MPC.  "requires a constant number of operations in expectation" is a term they use in a couple of places.  Their Bernoulli function has a cap on the number of recursions (not iterations, sadly, which will complicate things); but their discrete Laplace implementation does not: "the number N of iterations of the outer loop needed to output a value Z is geometrically distributed".  Note that we operate obliviously, so we have to perform all computations.  That means compromise if we are going to implement that.  It's probably not a big deal, but it introduces error that would need analysis.

martinthomson:
Of course, if you want the simple (non-MPC) implementation, the algorithms are very simple to execute.

bmcase:
For the short term we decided to get to discrete using a rounded Gaussian.  Unless epsilon is very small, the variance of the noise between a rounded Gaussian and continuous Gaussian is not much.  See Figure 2 of the above paper and the following computed values: 
image

src/protocol/dp/distributions.rs
```diff
@@ -0,0 +1,118 @@
+use rand::{
+    distributions::{Distribution, Uniform},
+    Rng,
+};
+
+use std::{f64::consts::PI, fmt::Debug};
+
+#[cfg(all(test, unit_test))]
+pub fn close(a: f64, b: f64, precision: u8) -> bool {
```
martinthomson, line 9: A trait might be better, but I'm going to say at least name this `is_close()` or `is_approx_equal()`

Also, please document `precision`.  I understand that the goal is to test that the difference between values is less than `10^-precision`.

src/protocol/dp/distributions.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+/// Rounded Normal distribution based on [`Box-Muller`] transform.
+///
+/// [`Box-Muller`]: https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
+#[derive(Debug)]
+pub struct RoundedBoxMuller {
+    pub mean: f64,
+    pub std: f64,
```
martinthomson, line 44: This should hold an instance of `BoxMuller` rather than re-implementing the algorithm.

src/protocol/dp/insecure.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        if !(f64::MIN_POSITIVE..=1.0 - f64::MIN_POSITIVE).contains(&delta) {
+            return Err(Error::BadDelta(delta));
+        }
+
+        // for (eps, delta) DP, the variance needs to be sensitivity^2/(eps^2) * 2ln(1.25/delta) see https://arxiv.org/pdf/1702.07476.pdf page 2
+        // sensitivity=L2(max(output_(with user x) - output_(without user x)))=sqrt(breakdown_count * user_contribution_per_breakdown^2)<cap
+        // minimum eps, delta is 1/u64_max, max for delta is 1-min
+        let variance = (cap / epsilon) * f64::sqrt(2.0 * f64::ln(1.25 / delta));
```
martinthomson, line 39: Duplicated code!

src/protocol/dp/insecure.rs
```diff
@@ -0,0 +1,241 @@
...
+    fn follows_normal_distribution(seed: u64, cap: u8, epsilon: u8, delta: f64) {
+        const N: usize = 10000;
+        // The sample_variance from a Gaussian distribution follows a chi square distribution with
+        // bounds:
+        // LB = (n - 1) * std^2 / chi2inv(alpha/2,n - 1)
+        // UB = (n - 1) * std^2 / chi2inv(1 - alpha/2, n - 1)
+        // where N is the size of the sample, alpha - the probability of any value to be outside
+        // of the expected distribution range. For the purpose of this test, alpha is set to 0.02%,
+        // chi2inv(0.0001,10,000 - 1) = 9,482.6
+        // chi2inv(0.9999,10,000 - 1) = 10,535
```
martinthomson, line 167: is this 10_000 - 1 or 10, -1?  Please avoid using ',' in numbers.

martinthomson (commented):
I'm not particularly happy about code repetition here, but I'm OK with proceeding.

martinthomson (approved):

akoshelev:
merging with a thought in mind that `DiscreteDp` may need to operate on integers. I guess I will know better once we have code that applies this DP to aggregates

PR #731, opened 2023-06-28
Query status API

andyleiserson:


src/query/processor.rs
```diff
@@ -517,6 +552,31 @@ mod tests {
...
+            {
+                sleep(Duration::from_millis(1)).await;
+            }
+
+            let results = app.complete_query(query_id).await?.map(|bytes| {
+                semi_honest::AdditiveShare::<Fp31>::from_byte_slice(&bytes).collect::<Vec<_>>()
+            });
+
+            Ok(assert_eq!(
+                vec![Fp31::truncate_from(20u128)],
```
martinthomson, line 575: `Vec<T>` is `PartialEq` to `&[T]` so you can just use a slice here.

martinthomson (approved):
LGTM, but then I'm probably not the best person to review this.

src/query/state.rs
```diff
@@ -66,6 +75,48 @@ impl QueryState {
...
+    /// We could return the result via the JoinHandle, except that we want to check the status
+    /// of the task, and shuttle doesn't implement `JoinHandle::is_finished`.
+    pub join_handle: JoinHandle<()>,
+}
+
+impl RunningQuery {
+    pub fn try_complete(&mut self) -> Option<QueryResult> {
+        match self.result.try_recv() {
+            Ok(result) => Some(result),
+            Err(TryRecvError::Closed) => {
```
akoshelev, line 95: wouldn't it also be in case if `try_complete` is called after the results have been received already? Would a lingering HTTP request that arrives right in between "try_complete".."complete" sequence of calls crash our query?

akoshelev (approved):
LGTM as well, but I'd love to understand whether `receiver::try_recv` is a footgun for our use case

src/query/state.rs
akoshelev: I think we are safe here because `RunningQuery` is behind a mutex, so it is not possible to poll it from two different threads

src/cli/playbook/ipa.rs
```diff
@@ -103,6 +109,23 @@ where
...
+            .unwrap()
+            .into_iter()
+            .all(|status| status == QueryStatus::Completed)
+        {
+            break;
+        }
+
+        sleep(delay).await;
+        delay = min(Duration::from_secs(60), delay * 2);
+        // TODO: Add a timeout of some sort. Possibly, add some sort of progress indicator to
```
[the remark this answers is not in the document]
andyleiserson: #732 

The difficulty of establishing an appropriate timeout when a query can contain an unbounded amount of work is exactly why I didn't try to tackle this now.

When I was considering getting rid of the blocking wait in the complete query API, the poll-for-completion idiom seen here would also have appeared in several other places. For simple test queries that ought to complete "instantly", I wasn't excited about introducing a case where we potentially wait for a longer, higher level timeout in CI to catch the failure.

In the CLI tools, it seems less important to try and set a default timeout.

src/query/state.rs
andyleiserson: Yeah, I think we're okay because of the mutex. Note that `try_complete` and `try_recv` take `&mut self`, so even if something were to change about the mutex, the mutability rules provide some protection.

I think in the long term we'll need some kind of "query manager" component that is free-running in the server. The query manager would move a query from running to completed state soon after it finishes, rather than the query remaining in the running state until somebody happens to call the status API. The query manager would also be responsible for things like garbage collecting abandoned queries after some period of time. In the case where there are multiple things asking for query status (e.g. the client APIs and a management API), it will not make sense for the client status API to be responsible for bookkeeping query status.

I didn't intend `RunningQuery` to be anything complicated -- I called it that because it's the data associated with the `QueryState::Running` enum variant. If we get `JoinHandle::is_finished` in shuttle, then we can get rid of `RunningQuery` and go back to `QueryState::Running(JoinHandle)`. It seems redundant to me to put the QueryId in RunningQuery -- I would rather use `QueryHandle` (which already contains QueryId) for that, possibly augmented with a method or `Deref` to access `QueryState`. (It also seems redundant to me for the AwaitingInputs state to contain QueryId.)

PR #733, opened 2023-07-01
Add seed, report-filter and conversion probability arguments to event generator

akoshelev:
For synthetic test we need to be able to generate uncoordinated impressions and conversions on different machines. That requires sharing a seed for random generator, new argument that tells `EventGenerator` to skip source or trigger events and conversion probability. 

I tested it using the following script

```bash
cargo run --release --bin report_collector --features="clap cli test-fixture" -- gen-ipa-inputs -n 1000 --output-file /tmp/ipa-conversions.txt --max-breakdown-key 20 --report-filter trigger-only --seed 123

cargo run --release --bin report_collector --features="clap cli test-fixture" -- gen-ipa-inputs -n 1000 --output-file /tmp/ipa-impressions.txt --max-breakdown-key 20 --report-filter source-only --seed 123

cat /tmp/ipa-impressions.txt /tmp/ipa-conversions.txt | sort -t ',' -k1 >! /tmp/ipa-gen-test.txt

cargo run --release --bin report_collector --features="cli test-fixture" -- --disable-https --input-file /tmp/ipa-gen-test.txt --network test_data/network.toml semi-honest-ipa --max-breakdown-key 20 --plaintext-match-keys
```

I was looking for the following outcome: some breakdowns must've had conversions occurred and aggregates should be greater than 0

```bash
+-----+----------+---------+-------+
| Row | Expected | Actual  | Diff? |
+==================================+
| 0   | Some(3)  | Some(3) |       |
|-----+----------+---------+-------|
| 1   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 2   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 3   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 4   | Some(4)  | Some(4) |       |
|-----+----------+---------+-------|
| 5   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 6   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 7   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 8   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 9   | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 10  | Some(0)  | Some(0) |       |
|-----+----------+---------+-------|
| 11  | Some(2)  | Some(2) |       |
|-----+----------+---------+-------|
[...]

PR #735, opened 2023-07-03
Add an option to RC binary to produce an output file 

akoshelev:
This builds on top of #733 and adds an option to report collector to produce a file with breakdowns. This will be needed to apply DP noise with different $\epsilon$ values. 

This change causes some re-arrangement of arguments for existing commands (gen-ipa-inputs).


src/bin/report_collector.rs
```diff
@@ -282,5 +279,23 @@ async fn ipa(
 
     tracing::info!("{m:?}", m = ipa_query_config);
 
-    validate(expected, actual)
+    validate(&expected, &actual);
```
richajaindce, line 282: Should we log output irrespective of validation?

src/bin/report_collector.rs
```diff
@@ -282,5 +279,23 @@ async fn ipa(
...
+
+    if let Some(ref path) = args.output_file {
```
richajaindce, line 284: We should also record the QueryConfig and time taken for the run in the output file so that we have everything in a place

src/bin/report_collector.rs
akoshelev: I thought no, we should not because if results are not valid, they shouldn't be recorded

richajaindce (approved):

src/bin/report_collector.rs
richajaindce: In that case, should we log that the results obtained were not as expected? I was thinking this might be useful to debug failures in future

akoshelev: I'll probably leave it out of scope for now - lets see if we get those often enough to have a need for it

src/bin/report_collector.rs
akoshelev: I can do that probably, this requires a scheme for the output file, but it should not be too complicated to do

src/bin/report_collector.rs
```diff
@@ -255,46 +257,53 @@ async fn ipa(
...
             .map_err(|e| format!("Failed to create output file {}: {e}", path.display()))?;
 
-        let mut writer = BufWriter::new(file);
-        for (i, row) in actual.iter().enumerate() {
-            if i > 0 {
-                write!(writer, ",")?;
-            }
-            write!(writer, "{}", row)?;
-        }
+        write!(file, "{}", serde_json::to_string_pretty(&actual)?)?;
```
richajaindce, line 306: Lets also add a print with output file name. 
"The results of query were written in {path}"

richajaindce (approved):
Thanks for adding this change :)

src/bin/report_collector.rs
akoshelev: yea I thought it may be redundant because caller specifies the path via `--output-file` argument

Issue #726, opened 2023-06-26
Logging for synthetic data


richajaindce:
For the test, we need logging in a few places
1. Print config on helper node with the running query
2. Also print the time taken by helper to execute a query.
Since all helpers are in different timezones, the log times arent very helpful.
This will also assist helpers to measure the runtime without asking us and provide proof points also.

richajaindce:
While throwing out errors, we should also log which helper gave that error. This should be added if query returns with 
1. Query is running state 
2. Network configuration is incorrect with error message "invalid peer certificate: BadSignature"
This would help in easier troubleshooting of setup

PR #738, opened 2023-07-04
Add query config and query size in the tracing logs

richajaindce:
This addresses https://github.com/private-attribution/ipa/issues/726

### DIFF
--- a/src/query/runner/ipa.rs
+++ b/src/query/runner/ipa.rs
@@ -67,7 +67,7 @@ where
         DowngradeMalicious<Target = MCAggregateCreditOutputRow<F, AdditiveShare<F>, BreakdownKey>>,
     AdditiveShare<F>: Serializable,
 {
-    #[tracing::instrument("ipa_query", skip_all)]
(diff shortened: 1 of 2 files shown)


src/query/runner/ipa.rs
```diff
@@ -67,7 +67,7 @@ where
         DowngradeMalicious<Target = MCAggregateCreditOutputRow<F, AdditiveShare<F>, BreakdownKey>>,
     AdditiveShare<F>: Serializable,
 {
-    #[tracing::instrument("ipa_query", skip_all)]
+    #[tracing::instrument("ipa_query", skip_all, fields(query_config=%self.config, %query_size))]
```
akoshelev, line 70:
```suggestion
    #[tracing::instrument("ipa_query", skip_all, fields(query_config=?self.config, %query_size))]
```

given that `Display` implementation outputs the `Debug`, we can ask tracing to do it for us. I am not sure however how verbose it is, we may still need a specialized `Display` or customized `Debug` to make it more concise

akoshelev (approved):

PR #739, opened 2023-07-05
Compact gate integration tests are rust native

taikiy:
Current integration tests for the compact gate runs a python script. This diff ports the equivalent tests in rust. I wanted to reuse `test_ipa` in `helper_networks` integrations tests, so I've moved it and others to `common` folder.


src/helpers/transport/query.rs
```diff
@@ -300,7 +300,7 @@ impl Default for IpaQueryConfig {
     fn default() -> Self {
         Self {
             per_user_credit_cap: 3,
-            max_breakdown_key: 64,
+            max_breakdown_key: 20,
```
taikiy, line 303: We only accept max_breakdown_key <= 32 now. Using 20 here which was the value used in helper_network integration tests.

akoshelev: may be good to create a type for it to enforce and document why

akoshelev (approved):

src/helpers/transport/query.rs
taikiy: I chatted with Ben Savage and we think this <=32 limitation is temporary. We have the assert to check the value is within the limit, so it should be okay with u32.

akoshelev:
lets fix the conflict and merge it @taikiy ?

PR #740, opened 2023-07-05
Clean up tests

akoshelev:
* Move shuttle tests to individual modules. For IPA it means just running whatever we have in `ipa/mod.rs` with different executors
* Make some unit tests run once in CI/pre-commit


src/helpers/mod.rs
```diff
@@ -660,3 +660,179 @@ mod tests {
         }
     }
 }
+
+#[cfg(all(test, feature = "shuttle"))]
+mod concurrency_tests {
```
akoshelev, line 665: moved from `src/tests/infra.rs`

richajaindce (approved):
Looks good

PR #748, opened 2023-07-11
Remove no-prss feature and implementation

akoshelev:
We no longer test our protocols without PRSS and I don't think we need to, because it adds necessary cost that we can't avoid in prod.


src/net/client/mod.rs
```diff
@@ -154,7 +154,7 @@ impl MpcHelperClient {
                 let builder = builder.with_root_certificates(cert_store);
                 match identity {
                     ClientIdentity::Certificate((cert_chain, pk)) => builder
-                        .with_single_cert(cert_chain, pk)
+                        .with_client_auth_cert(cert_chain, pk)
```
martinthomson, line 157: Note that you need to update the rustls dependency here.  I had an old version in Cargo.lock that wouldn't build due to this change.

martinthomson (approved):

PR #750, opened 2023-07-11
Fix docker action to push on current repo and not on runner's repo

richajaindce:


akoshelev (approved):
can't run the workflow to check whether it works or no, so yolo'ing it to main

Issue #728, opened 2023-06-26
query execution failed: runtime error


richajaindce:
Not sure whats wrong but running IPA with user cap = 1 in real helpers failed for 100k and failed on AWS cluster for 1M rows

Error : thread 'main' panicked at 'called Result::unwrap() on an Err value: FailedHttpRequest { status: 500, reason: "query execution failed: runtime error" }', /home/ec2-user/workspace/ipa/src/cli/playbook/ipa.rs:109:10

akoshelev:
was able to reproduce it on 1M records

## RC

```bash
2023-06-26T19:08:12.360008Z  INFO ipa::cli::verbosity: Logging setup at level info
2023-06-26T19:51:30.811996Z  INFO ipa::cli::playbook::ipa: Starting query after finishing encryption
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FailedHttpRequest { status: 500, reason: "query execution failed: Infrastructure error: Expected to receive RecordId(999424) but hit end of stream" }', /home/ec2-user/workspace/ipa/src/cli/playbook/ipa.rs:109:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## H1

```bash
2023-06-26T19:51:30.901961Z  INFO ipa_query: ipa::query::runner::ipa: new
2023-06-26T19:58:15.432130Z ERROR ipa_query: ipa::helpers::transport::receive: error reading records: error reading a body from connection: connection reset
2023-06-26T19:58:15.432225Z ERROR ipa_query: ipa::helpers::transport::receive: error reading records: error reading a body from connection: connection reset
2023-06-26T19:58:15.432232Z ERROR ipa_query: ipa::helpers::transport::receive: error reading records: error reading a body from connection: connection reset
2023-06-26T19:58:15.432246Z ERROR ipa_query: ipa::helpers::transport::receive: error reading records: error reading a body from connection: connection reset
2023-06-26T19:58:15.432288Z ERROR ipa_query: ipa::helpers::transport::receive: error reading records: error reading a body from connection: connection reset
2023-06-26T19:58:15.645819Z  INFO ipa_query: ipa::query::runner::ipa: close time.busy=395s time.idle=9.90s
2023-06-26T19:58:15.649718Z ERROR tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error latency=404748 ms
```

## H2

```bash
[...]

PR #751, opened 2023-07-12
Add step/role to receive error

akoshelev:
While dealing with #728 I noticed that end of stream error does not carry the information about the origin and step.


richajaindce (approved):
Looks good!

PR #753, opened 2023-07-12
Add idle tracking

RuiyuZhu:
Add idle tracking in transport layer so gateway could use that to detect if the program has been stuck.


akoshelev:
In general I don't think we can make it work with monitoring transport without duplicating code for HTTP/in-memory implementations. In order for transport to monitor everything, it needs to plumb monitoring into its `send/receive` methods but also inside the streams that are returned from it. Putting the state watchdog above it (Gateway) requires us to implement it once and work everywhere

PR #752, opened 2023-07-12
Readme update

eriktaubeneck:
updating the README and Cargo.toml as I'm getting started. 

A few of these benchmarks aren't working for me, and I added notes in the readme. I can delete those and open issues, if that's preferred, but I just wanted to make sure we documented it somewhere.

PR #754, opened 2023-07-13
Fix arithmetic benchmark

akoshelev:
In #752 it was noticed that it no longer finishes for width values more than 1. It turns out it happens because `world.contexts()` creates a new execution and record identifiers greater than 0 will block forever.

I also included running it in CI so we don't let it rot


akoshelev:
CI is failing because of Rust 1.71 release. #755 should fix it

Issue #749, opened 2023-07-11
DP tests are failing


akoshelev:
https://github.com/akoshelev/raw-ipa/actions/runs/5514048224/jobs/10052858316

```bash
---- protocol::dp::insecure::test::epsilon_variance_table stdout ----
epsilon = 1, rounded_normal_sample_variance = 29.31470711071145, continuous_variance = 28.077308218556965
thread 'protocol::dp::insecure::test::epsilon_variance_table' panicked at 'assertion failed: sample_variance - dp.rounded_normal_dist.std().powi(2) < 1.0', src/protocol/dp/insecure.rs:223:13
```

martinthomson:
Same problem here.

```
$ cat proptest-regressions/protocol/dp/insecure.txt
# Seeds for failure cases proptest has generated in the past. It is
# automatically read and these particular cases re-run before any
# novel cases are generated.
#
# It is recommended to check this file in to source control so that
# everyone who runs the test benefits from these saved cases.
cc a6d6f9aebc8b3e75a06b948d5b0df73fd2718dd54136326c6c1f02856ae4e16f # shrinks to rng_seed = 8299925628559468849, epsilon = 1, delta = 1e-9, cap = 1
```

martinthomson:
It seems like proptest is very good at finding examples on the bounds.  How many standard deviations from the mean was your cutoff?

akoshelev:
this case is probably different (but needs a fix too). There is a fixed probability for `output_differentially_private` to fail, it seems to be high enough for proptest to notice it. 

 for `epsilon_variance_table` test we don't do anything sophisticated I think, I asked @bmcase to take a look

PR #759, opened 2023-07-17
Relax bounds for normal distribution tests

akoshelev:
Proptest was able to find a case to fail, so setting up the probability for variance to be outside the expected range to 2e-4.

was noted in #749


akoshelev:
included in #768

Issue #746, opened 2023-07-07
Revisit bit decomposition and random bit generator implementations


akoshelev:
Currently IPA uses bit decomposition protocol from Damgård's [paper](https://iacr.org/archive/tcc2006/38760286/38760286.pdf) that includes random bit generator. There is a [paper](https://www.iacr.org/archive/pkc2007/44500343/44500343.pdf)  that describes a more optimal implementation of bit decomposition and random bit generator that we could use to improve IPA network utilization.

We've already implemented some of the protocols from this paper, namely [greater than C](https://github.com/private-attribution/ipa/blob/25db2edfc6d6a728e51903b829e76da58923256a/src/protocol/boolean/comparison.rs#L63).

The goal for this task is to implement random bit generator and bit decomposition from the newer paper and compare it with the existing one to decide which one is to keep. 

## Code pointers

* Random bit generator protocol (Damgård): https://github.com/private-attribution/ipa/blob/07414f4bf0fa72b706b2d2bc0e026e40cdfedffb/src/protocol/boolean/solved_bits.rs#L83
* Bit decomposition protocol (Damgård): https://github.com/private-attribution/ipa/blob/07414f4bf0fa72b706b2d2bc0e026e40cdfedffb/src/protocol/boolean/bit_decomposition.rs#L24
* IPA uses bit decomposition to convert match keys and breakdown keys: https://github.com/private-attribution/ipa/blob/8e42a99fb6f340bff02aa24fe0f6eab471d1edc8/src/protocol/ipa/mod.rs#L362

eriktaubeneck:
Looking at section 5 of the [new paper](https://www.iacr.org/archive/pkc2007/44500343/44500343.pdf), it appears that the `solved_bits` function is the same, but the `bit_decomposition` is improved. I'm planning to focus on that, but posting here to set correct expectations.

andyleiserson:
I spent some time on this when I was looking at the comparison protocol in that paper. My conclusion (as best I can remember / reconstruct from my notes) was that the Nishide + Ohta improved bit decomposition protocol may not offer an improvement over what we already have implemented for bit decomposition (which follows the Damgård outline, but has some optimizations).

[...]

PR #760, opened 2023-07-17
[DRAFT] Simplified bit decomposition

eriktaubeneck:
This addresses #746. It isn't yet passing tests, but I wanted to put this up so that some of the rusticans in the group could make sure I'm not doing anything dumb with references, clones, etc.

I'm going to keep hunting down the bug with correctness, so don't worry about that yet.

Also, it's unclear if this will actually offer any performance gains, so don't worry too much yet about the structure. If this does pan out to be more efficient, then I'll definitely want some input on this, because this almost certainly shouldn't just be inline like this.


src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+        // let el = usize::try_from(u128::BITS - F::PRIME.into().leading_zeros()).unwrap();
         let el = u128::BITS - F::PRIME.into().leading_zeros();
-        let x = (1 << el) - F::PRIME.into();
+        let two_exp_el = u128::pow(2, el);
+        let _f_int: u128 = two_exp_el + c.as_u128() - F::PRIME.into();
+        debug_assert!(_f_int < u64::max_value().into());
+        let f_int: u64 = _f_int as u64;
+
+        // Step 4.2. Compute [g]_B = (f_i - c_i) [q]_p + c_i
+        let mut g_bin: Vec<G> = Vec::with_capacity(usize::try_from(el).unwrap());
```
akoshelev, line 76: you probably shouldn't keep these values, rather compute that on the fly because it is very cheap and requires only two 8 byte values to be around

you can still define an Enum to represent those logical values and use it, just compute it on the fly

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+            &r.b_b,
+            p_minus_c.into(),
         )
         .await?;
 
-        // Step 7. a bitwise scalar value `f_B = bits(2^el - p)`
+        // Step 4 has a lot going on. We'd going to break it down into substeps.
+        // Step 4.1. Make a bitwise scalar value of f = 2^el + c - p.
+        // let el = usize::try_from(u128::BITS - F::PRIME.into().leading_zeros()).unwrap();
         let el = u128::BITS - F::PRIME.into().leading_zeros();
```
akoshelev, line 69: is it just `let el = F::BITS`?

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+            // let f_i_minus_c_i: u64 = f_i - c_i;
+            // let g_prime: S = q_p.clone().into() * f_i_minus_c_i;
+            // // c_i ∈ {0, 1}, so F::truncate_from will work
+            // debug_assert!(c_i <= 1_u128);
+            // g_bin.push(S::share_known_value(&ctx, F::truncate_from(c_i)) + &g_prime);
+        }
+
+        // Step 5. Compute BitwiseSum([r]_B, [g]_B)
+        let mut h: Vec<S> = Vec::with_capacity(usize::try_from(el).unwrap());
+        let one_minus_q_p = S::share_known_value(&ctx, F::ONE) - &q_p.clone();
```
akoshelev, line 106:
```suggestion
        let one_minus_q_p = S::share_known_value(&ctx, F::ONE) - &q_p;
```

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+        h.push(result_bit);
 
-        // Step 8, 9. [g_i] = [q] * f_i
-        // Step 10. [h]_B = [d + g]_B, where [h]_B = ([h]_0,...[h]_(el+1))
-        // Step 11. [a]_B = ([h]_0,...[h]_(el-1))
-        let a_b =
-            maybe_add_constant_mod2l(ctx.narrow(&Step::AddDtoG), record_id, &d_b, x, &q_p).await?;
+        for (bit_index, bit) in r.b_b.iter().enumerate().skip(1) {
+            let mult_result = if last_carry_known_to_be_zero {
+                // TODO: this makes me sad
```
akoshelev, line 129: what is going on here?

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+            let last_carry_or_bit = -mult_result.clone() + &last_carry + bit;
+            let next_carry = match &g_bin[bit_index] {
+                G::Zero => mult_result,
+                G::One => {
+                    last_carry_known_to_be_zero = false;
+                    last_carry_or_bit
+                }
+                G::Q => {
+                    last_carry_known_to_be_zero = false;
+                    q_p.multiply(
```
akoshelev, line 149: I am not sure this will work. Values seem to be chosen by rolling a dice, so for record = 1 you may hit this path while record 0 may short-circuit it if bit is zero or one. 

There is a limitation that the number of multiplications (or communications) per protocol must be deterministic

Unless I misunderstood this code, it is not the case here

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+            G::One => (r.b_b[0].clone(), S::share_known_value(&ctx, F::ONE), false),
+            G::Q => (
+                q_p.multiply(&r.b_b[0], ctx.narrow(&Step::AddGtoR), record_id)
+                    .await?,
+                q_p.clone() + &r.b_b[0],
+                false,
+            ),
+            G::NotQ => (
+                one_minus_q_p
+                    .clone()
```
akoshelev, line 118:
```suggestion
```

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+            };
+            g_bin.push(g_i);
+            // let f_i_minus_c_i: u64 = f_i - c_i;
+            // let g_prime: S = q_p.clone().into() * f_i_minus_c_i;
+            // // c_i ∈ {0, 1}, so F::truncate_from will work
+            // debug_assert!(c_i <= 1_u128);
+            // g_bin.push(S::share_known_value(&ctx, F::truncate_from(c_i)) + &g_prime);
+        }
+
+        // Step 5. Compute BitwiseSum([r]_B, [g]_B)
```
akoshelev, line 104: that deserves to be a sub protocol?

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
-        // Step 7. a bitwise scalar value `f_B = bits(2^el - p)`
+        // Step 4 has a lot going on. We'd going to break it down into substeps.
+        // Step 4.1. Make a bitwise scalar value of f = 2^el + c - p.
+        // let el = usize::try_from(u128::BITS - F::PRIME.into().leading_zeros()).unwrap();
         let el = u128::BITS - F::PRIME.into().leading_zeros();
-        let x = (1 << el) - F::PRIME.into();
+        let two_exp_el = u128::pow(2, el);
+        let _f_int: u128 = two_exp_el + c.as_u128() - F::PRIME.into();
+        debug_assert!(_f_int < u64::max_value().into());
+        let f_int: u64 = _f_int as u64;
```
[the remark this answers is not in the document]
eriktaubeneck: still ok to do so while coercing to a new type?

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: yep good call. I was thinking the same thing. the two loops can become one

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: oh well that explains what's going on in line 129! that does make me sad.

this probably means we won't get any wins here, since some of the "speed up" would be from knowing this value ~half the time, randomly

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: what should I read to know when to clone and when not to?

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
+        // Step 2.1: Edge case, if r is coincidentally a bit decomposition of [a]_p, we're done.
+        if c == F::ZERO {
+            return Ok(r.b_b);
+        }
 
-        // Step 6. q = d >=? p (note: the paper uses p <? d, which is incorrect)
-        let q_p = BitwiseLessThanPrime::greater_than_or_equal_to_prime(
-            ctx.narrow(&Step::IsPLessThanD),
+        // Step 3. Compute [q]_p = 1 - [r <_B p - c ]_p. q is 1 iff r + c >= p in the integers.
+        let p_minus_c: u128 = F::PRIME.into() - c.as_u128();
```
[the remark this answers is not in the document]
eriktaubeneck: it is the same, but it's only used with `bitwise_less_than_constant` that expects a `u128`.

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: Maybe? I was just modifying what's used in [add_constant.rs](https://github.com/private-attribution/ipa/blob/main/src/protocol/boolean/add_constant.rs#L144). Looking at the blame, it seems like @martinthomson did this to remove the trailing bits.

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -38,49 +40,170 @@ impl BitDecomposition {
...
 
-        // Step 7. a bitwise scalar value `f_B = bits(2^el - p)`
+        // Step 4 has a lot going on. We'd going to break it down into substeps.
+        // Step 4.1. Make a bitwise scalar value of f = 2^el + c - p.
+        // let el = usize::try_from(u128::BITS - F::PRIME.into().leading_zeros()).unwrap();
         let el = u128::BITS - F::PRIME.into().leading_zeros();
-        let x = (1 << el) - F::PRIME.into();
+        let two_exp_el = u128::pow(2, el);
+        let _f_int: u128 = two_exp_el + c.as_u128() - F::PRIME.into();
+        debug_assert!(_f_int < u64::max_value().into());
```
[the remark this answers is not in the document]
eriktaubeneck: hmm good call out. this probably does need to be a `u128`.

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: also, I'm getting an error that this value has been moved, and it was moved here.

src/protocol/boolean/bit_decomposition.rs
akoshelev: ah no, I think this version is better if PRIME << $2^n-1$

src/protocol/boolean/bit_decomposition.rs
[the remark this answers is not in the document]
akoshelev: yep, perfectly valid

src/protocol/boolean/bit_decomposition.rs
akoshelev: > what should I read to know when to clone and when not to?

Rust standard library docs are pretty good:
https://doc.rust-lang.org/std/clone/trait.Clone.html
https://doc.rust-lang.org/std/marker/trait.Copy.html

akoshelev: >also, I'm getting an error that this value has been moved, and it was moved here.

it is weird because `multiply` should take it by reference

src/protocol/boolean/bit_decomposition.rs
```diff
@@ -133,6 +281,38 @@ mod tests {
...
+        F: PrimeField + ExtendableField + Sized,
+        Standard: Distribution<F>,
+    {
+        let r_f_b = get_bits::<F>(r, num_bits);
+
+        let result = world
+            .semi_honest((q, r_f_b), |ctx, (q_p, r_b_d)| async move {
+                let ctx = ctx.set_total_records(1);
+                let r_b = r_b_d.to_vec();
+                compute_bit_addition(ctx, RecordId::from(0), r_b, g_b, &q_p)
```
eriktaubeneck, line 301: @akoshelev I'm getting an error here about moving `g_b`. Any idea how to fix it? Here's the full error:

```
error[E0507]: cannot move out of `g_b`, a captured variable in an `Fn` closure
   --> src/protocol/boolean/bit_decomposition.rs:298:58
    |
287 |           g_b: GBIterator,
    |           --- captured outer variable
...
298 |               .semi_honest((q, r_f_b), |ctx, (q_p, r_b_d)| async move {
    |  ______________________________________-------------------_^
    | |                                      |
    | |                                      captured by this `Fn` closure
299 | |                 let ctx = ctx.set_total_records(1);
300 | |                 let r_b = r_b_d.to_vec();
301 | |                 compute_bit_addition(ctx, RecordId::from(0), r_b, g_b, &q_p)
    | |                                                                   ---
    | |                                                                   |
    | |                                                                   variable moved due to use in generator
    | |                                                                   move occurs because `g_b` has type `GBIterator`, which does not implement the `Copy` trait
302 | |                     .await
303 | |                     .unwrap()
304 | |             })
    | |_____________^ `g_b` is moved here
```

src/protocol/boolean/bit_decomposition.rs
martinthomson: our code is a little weird in a few ways.

We don't implement `Copy` for shares when we probably could; that is, we require that you explicitly call `clone()` so that the cost of creating a new instance is explicit.

We often do require a new instance: most operations return a new value, so you need the output to be a new instance; but we often only have references (because we want to avoid making copies).  Also, our operations generally have a full instance on the left and a reference on the right.  That seems a bit weird until you realize that what that allows us to do is have the left operand be modified by the various operations and returned as the output.

src/protocol/boolean/bit_decomposition.rs
martinthomson: This is code that Ben and I wrote to get around the fact that we can't have holes in the sequence record indices.  That is, if you multiply record N, you also have to multiply records 0 through N-1 or the infrastructure stalls.

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: this is outdated now, resolving.

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: yea, I have to do even more of this with this PR.

eriktaubeneck: this specific line is outdated, and Alex and I have talked about it, so resolving this comment.

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: it probably does. I've refactored it into a function, but I'm not sure if it should actually go into a new file. the `GBIterator` we create is very contextual to this specific protocol for bit decomposition, and not a generic bitwise addition, so it probably makes sense to keep it "close".

src/protocol/boolean/bit_decomposition.rs
eriktaubeneck: cc @martinthomson too 😀

martinthomson: The callback identified is invoked 3 times.  You only have one copy of `g_b`.  Consider moving `g_b` into the tuple that you pass to `semi_honest()` (you might need to `collect()` it...)

eriktaubeneck: I tried moving it into that tuple, but `semi_honest` modifies the values passed to it (in this case, converting field elements into secret sharing.) When I included `g_b`, it complained about not having an implementation for that type of iterator, which makes sense, since I just made the iterator.

eriktaubeneck:
Ok - officially looking for some feedback here, likely from @akoshelev and @martinthomson:

1. `bit_decomposition` isn't actually used in our protocol at this point. Match keys are received directly as bits, and we do modulo conversion instead of decomposition a field element into bits. **Question**: Do we just want to drop this function, or is it worth keeping around for potential future use?
2. I added a baseline tests to the existing `bit_decomposition` function in [this branch](https://github.com/private-attribution/ipa/tree/main-bit-decomp-perf) as well as in this branch. **Compared to main, this "simplified" protocol is less performant.**
    1. For 20 Fp31 elements (that's group size 31, not 31 bits), performance:
        1. main: `{ records_sent: 1560, bytes_sent: 1560, indexed_prss: 1830, seq_prss: 0 }`
         2. updated: 
3. Just for fun, I also opened [a branch](https://github.com/private-attribution/ipa/tree/simplified-bit-decomposition-variable-mulitplications) with all the "dummy" multiplications removed, which results in a nondeterministic number of multiplications. I expected this to cause the tests to fail, but it didn't. It also resulted in an improvement of performance:
    1. `{ records_sent: 1371, bytes_sent: 1371, indexed_prss: 1641, seq_prss: 0 }`
    2. My assumption here is that this would actually break if used, and the test isn't capturing it properly here, but @akoshelev or @martinthomson, if you have any insight here, that would be helpful.
    
In the end, two decisions:
1. Should we keep this code at all?
2. If yes, is the nondeterministic multiplications somehow viable, or are the tests just not catching it? If it is viable, do we want to keep this update?

martinthomson:
Aren't the non-deterministic multiplications present on both branches?  In which case, we'd get a similar gain from fixing that on both sides.

Assuming yes to the above, I don't see much value in a slower and more complex decomposition algorithm, unless we find that circuit depth or some other metric could be relevant.

akoshelev:
Circuit depth may be interesting to compare, but we don't have tools to do that currently. We could take a look at steps used inside it and see if the depth (`/` in the step path) is comparable. 

fake multiplications is an interesting one - if we really need to fill gaps in infra, we could probably just generate a zero and send/receive it, without going through the expensive PRSS stuff. 


If there is nothing we could do to bring the number of multiplications down, then we probably should stick to the old protocol

eriktaubeneck:
> Aren't the non-deterministic multiplications present on both branches? In which case, we'd get a similar gain from fixing that on both sides.

Great point. I just tested removing those on main, and it results in: `{ records_sent: 1515, bytes_sent: 1515, indexed_prss: 1785, seq_prss: 0 }`. So in order of performance (best to worst) we have:
1. new protocol, no fake multiplications (1371 records sent, 1641 prss)
2. old protocol, no fake multiplications (1515 records sent, 1785 prss)
3. old protocol w/ fake multiplications (1560 records sent, 1830 prss)
4. new protocol w/ fake multiplication (1620 records sent, 1890 prss)

> Assuming yes to the above, I don't see much value in a slower and more complex decomposition algorithm, unless we find that circuit depth or some other metric could be relevant.

So it's only slower if we have to do the fake multiplications (and I don't understand why the tests aren't failing without them.) It's also maybe a bit uglier in its current implementation, but I think the complexity is about the same. It appears more complex right now because I just added the new `bit_addition` and didn't remove the old `add_constant` function, but the two have roughly the same complexity.

> If there is nothing we could do to bring the number of multiplications down, then we probably should stick to the old protocol.
Definitely agree here.

eriktaubeneck:
Ok! I got the [tests to fail due to nondeterministic errors](https://github.com/private-attribution/ipa/commit/5404c95ed3e8fedb12d166e22024ba8a23aa814b):

```
thread 'protocol::boolean::bit_decomposition::tests::bit_decomposition_perf' panicked at 'assertion failed: !total_records.is_unspecified()', src/helpers/gateway/send.rs:125:9
```

So, given the constraint of deterministic multiplications, we shouldn't merge this.

Only remaining question is: Should we remove bit decomposition entirely since it's currently unused dead code?

benjaminsavage:
Can we close this PR?

eriktaubeneck:
We can, but I'd like to decide if we can delete the bit decomposition code which is unused.

PR #761, opened 2023-07-18
fix flaky dp test

bmcase:
giving some more room on the difference between variance of rounded and continuous gaussians.  We don't have a theoretical bound to use here right now, so is a simple test but for small epsilon where these can diverge it was failing often.  Ideally, we get a better bound in the future, but the actual plan is to switch to generating a discrete gaussian (or laplace) and stop using the rounded gaussian.


akoshelev:
Agree that this measure is temporary until we have a proper DP mechanism. Flaky tests are bad

PR #764, opened 2023-07-21
step attribute macro

taikiy:
This new macro `#[step]` adds `#[derive(AsRefStr)]` from strum that expands to `AsRef(str)` implementation, so that we don't have to keep writing that boilerplate `impl AsRef<str>` code. Eventually, this macro would also generate the state transition code (`impl StepNarrow<StepA> for Compact`) in the file where `StepA` is defined, instead of `compact.rs`.

Currently it only supports:
* `enum` steps, although changing `struct` steps to `enum` is super easy
* no "dynamic" or "conditional" steps are supported. Will be adding this feature next

I've also moved the steps.txt file from `ipa-macros/` folder to the core `ipa/src/protocol/step/`.


ipa-macros/src/derive_gate/mod.rs
taikiy, line 1: moved things that will be used commonly to `parser.rs`

ipa-macros/src/parser.rs
taikiy, line 1: Code here is moved from `derive_gate/mod.rs`. Most of the code is unchanged. The only parts are `steps.txt` path related.

ipa-macros/src/parser.rs
```diff
@@ -0,0 +1,211 @@
...
+        })
+        .collect::<Vec<_>>();
+
+    construct_tree(steps)
+}
+
+/// Reads the steps file and returns a vector of strings, where each string represents a line in the file.
+pub(crate) fn read_steps_file(file_path: &str) -> Vec<String> {
+    // construct the path to the steps file saved in STEPS_FILE_PATH relative to this crate's root.
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR").to_owned() + STEPS_FILE_PATH);
```
taikiy, line 70: This is the new code. `CARGO_MANIFEST_DIR` points to `ipa-macros`, so we go up one level, then find the file in `src/` directory ("../" part is in `STEPS_FILE_PATH`).

ipa-macros/tests/derive_gate/common/mod.rs
```diff
@@ -42,3 +42,10 @@ pub fn static_reverse_state_map(state: u16) -> &'static str {
         _ => unreachable!("static_reverse_state_map should not be called"),
     }
 }
+
+pub fn static_deserialize_state_map(s: &str) -> u16 {
```
taikiy, line 46: Forgot to add this previously. I also changed pre-commit to run ipa-macros integration tests.

pre-commit
```diff
@@ -99,3 +99,6 @@ check "IPA benchmark" \
 
 check "Sort benchmark" \
     cargo bench --bench oneshot_sort --no-default-features --features="enable-benches descriptive-gate"
+
+check "ipa-macros tests" \
+    pushd ipa-macros && cargo test --features="trybuild" && popd
```
taikiy, line 104: Is there a better way to do this? ipa-macros isn't a workspace so I can't use `cargo -p ipa-macros`.

src/protocol/ipa/mod.rs
```diff
@@ -1265,7 +1238,7 @@ pub mod tests {
                     records_sent: 36_090,
                     bytes_sent: 133_776,
                     indexed_prss: 74_046,
-                    seq_prss: 1098,
+                    seq_prss: 1132,
```
taikiy, line 1241: This one is weird. I changed `ShuffleRevealStep::*Permutation` to `ShuffleRevealPermutationStep::*` due to clippy errors. That affected as_ref() strings to change and ultimately caused this regression. I verified that reverting these as_ref() strings back fixed the issue. We should be okay since bytes_sent is unchanged, but I'm not sure whether seq_prss (or indexed_prss) is a useful metric to track?

src/protocol/step/steps.txt
taikiy, line 1: Regenerated steps.txt since I had to change some of enum variants.

ipa-macros/src/derive_gate/mod.rs
```diff
@@ -1,12 +1,7 @@
-use crate::tree::Node;
+use crate::parser::*;
```
akoshelev, line 1: I am surprised you got away with this. Isn't clippy configured to flag star imports?

ipa-macros/src/step/mod.rs
```diff
@@ -0,0 +1,32 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::parse_macro_input;
+
+/// Generate a state transition graph and the corresponding `StepNarrow` implementations for the
+/// IPA protocol.
+pub fn expand(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
akoshelev, line 7: as we are basically implementing a trait, I believe `Derive(step)` is what we would want?

ipa-macros/src/step/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    add_derives(&step_enum).into()
+}
+
+/// Add required derives for the step enum to work with the IPA protocol.
+fn add_derives(enum_item: &syn::ItemEnum) -> TokenStream {
+    let enum_ident = &enum_item.ident;
+    quote!(
+        #[derive(AsRefStr, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+        #[strum(serialize_all = "snake_case")]
```
akoshelev, line 26: I see why, but it is also coupling two concepts tightly together (step -> strum -> str). It will be problematic if we want our own custom `AsRef` for any reason (including but not limited to standards).

It may be worth splitting it off and just generating step stuff here

ipa-macros/src/step/mod.rs
```diff
@@ -0,0 +1,32 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::parse_macro_input;
+
+/// Generate a state transition graph and the corresponding `StepNarrow` implementations for the
```
akoshelev, line 5: are we still generating `StepNarrow` here? can't find it in this file

ipa-macros/src/derive_gate/mod.rs
taikiy: Apparently not 😆

ipa-macros/src/step/mod.rs
taikiy: I'm using attribute proc macro here because of two features that are not available in derive proc macro. 1) it replaces itself with any tokens during macro expansion so it allows me to add a derive trait, and 2) supports attributes that are applied to the entire item such as `#[step(obsolete)]`. I have a new diff that this attribute macro adds `derive(Step)`

ipa-macros/src/step/mod.rs
taikiy: Good point. I actually stopped using strum in the next diff because I need to support tuple enum variants. For this diff it should be okay as is.

ipa-macros/src/step/mod.rs
taikiy: I jumped ahead of myself. I'll fix the comment.

ipa-macros/src/step/mod.rs
akoshelev: yea, we should use `derive(Step)` and let end users derive other attributes, even if they are common for all steps.

ipa-macros/src/step/mod.rs
akoshelev: yea we can let people opt-out of strum if they want to, or use it by slapping `derive(AsRefStr)` if they really don't care about the strings

pre-commit
```diff
@@ -99,3 +99,7 @@ check "IPA benchmark" \
...
+
```
akoshelev, line 105: there is a new line warning here

akoshelev (approved):
we can push it to make progress - I expect `step` macro to turn into `derive(Step)` as @taikiy mentioned in the comments

PR #766, opened 2023-07-25
Single producer single consumer channel

akoshelev:
This is a step towards adding more parallelism to IPA. Tasks spawned to execute steps need to communicate with each other. Some that will use internal parallelism need the full power of MPSC channels, others that execute just a few multiplications per row need a more efficient channel implementation. There is always a single downstream consumer that takes a stream from producing task, having a single producer allows to use a more efficient implementation that does not require synchronization on the writer side.

The proposed implementation uses a single ring buffer with two pointers that are updated exclusively by reader/writer. Whenever a pointer is updated, the opposite end of the channel is notified.  

I don't know if there is a better implementation with 100% safe Rust, there are ways to improve it with unsafe pointers but not sure if we want to cross that bridge now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,12 @@ harness = false
 required-features = ["enable-benches"]
 
 [[bench]]
+name = "criterion_spsc"
+path = "benches/ct/spsc.rs"
+harness = false
+required-features = ["enable-benches"]
+
+[[bench]]
 name = "iai_arithmetic"
 path = "benches/iai/arithmetic_circuit.rs"
 harness = false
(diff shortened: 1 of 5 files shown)


benjaminsavage:
Can we close this PR?

benjaminsavage:
Pinging again. This PR is ancient. Can we close it?

akoshelev:
let me close it for now - at some point I want to experiment with task-per-step approach as I believe it will have better performance compared to parallelism via `seq_join`/`parallel_join`. But I don't have time to run this experiment yet

Issue #706, opened 2023-06-15
[CompactGate] Allow `narrow`ing of steps that are conditionally executed


taikiy:
# Goal
Automatically generate code to support `narrow`ing of steps that are conditionally executed.

# Context
`Compact` gate code is generated from the steps file, which only contain steps that are executed in IPA. It doesn't contain any information about steps that are narrowed but not executed. However, some protocols narrow the context in advance for convenience. Such narrowing to unknown state will cause the `Compact` gate to panic.

Examples of these steps:
- crate::protocol::boolean::random_bits_generator::FallbackStep
     This step is only executed when generated random secret shared bits are of a share larger than
     `2^32 - 5`. However, the context is narrowed to this step when a RBG instance is created. This one
     is easy to solve by allowing to narrow to a hard-coded state as there are no branching child steps.
- crate::protocol::attribution::Step::CurrentCreditOrCreditUpdate
     This and other similar steps in the "binary tree prefixed sum" protocols are highly optimized such
     that they are executed (or not executed) depending on the iteration/depth of the loop. To solve this,
     we could change the code to only narrow before actually calling the `multiply` instead of narrowing
     in advance. Code would become a bit messy, but it's doable.
- crate::protocol::context::UpgradeStep
     This steps is executed in both malicious and semi-honest contexts, but the `narrow` call in semi-honest
     context is a dummy; it doesn't trigger `send`. This is a bit tricky to solve because there are preceding/
     succeeding subsequent steps, and the compact gate only knows its current state and the new state.

There are two ways to solve this:

- (a) Allow `narrow`ing from any state to that state. We could do this either by
     1) hard coding or by
     2) annotating the steps that are conditionally executed. We need to take into consideration about a case where
        there are child steps branching off from the conditional step.
- (b) Forcibly execute conditional steps when `step-trace` feature is on to generate the steps file.

Currently, we are doing (a-1). It could let unexpected state transitions happen without panicking, but there isn't a better way to solve this.

taikiy:
fixed with #767

PR #767, opened 2023-07-27
collect step tracing each time a step is narrowed

taikiy:
Current compact gate has an issue (#706) that not all narrowed steps are not executed. This diff adds one metric to our tracing. Each time we narrow a step, we log it and generate metrics output when debug tracing is on. Since `steps.txt` (file we use to generate compact gate) will contain all narrowed steps rather than executed steps, we don't need special handlings for such steps anymore.

This resolves the issue #706, but it also comes with a side effect. The compact gate would now contain state transitions that don't actually trigger communications. I can't think of that being a problem in production, but it is a bit of a concern that we cannot distinguish between comms/no-comms state transitions.

That said, the benefit of being able to get rid of that "magic" in compact.rs surpasses the side effect.

### DIFF
--- a/src/protocol/step/descriptive.rs
+++ b/src/protocol/step/descriptive.rs
@@ -1,4 +1,5 @@
 use super::{Step, StepNarrow};
+#[cfg(feature = "step-trace")]
 use crate::telemetry::{labels::STEP, metrics::STEP_NARROWED};
 use std::fmt::{Debug, Display, Formatter};
 
@@ -54,7 +55,10 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
             id += [std::any::type_name::<S>(), "::"].concat().as_ref();
         }
         id += step.as_ref();
-        metrics::increment_counter!(STEP_NARROWED, STEP => id.clone());
+        #[cfg(feature = "step-trace")]
+        {
+            metrics::increment_counter!(STEP_NARROWED, STEP => id.clone());
+        }
 
         Self { id }
     }
--- a/src/protocol/step/descriptive.rs
+++ b/src/protocol/step/descriptive.rs
@@ -1,4 +1,5 @@
 use super::{Step, StepNarrow};
+use crate::telemetry::{labels::STEP, metrics::STEP_NARROWED};
 use std::fmt::{Debug, Display, Formatter};
 
 /// A descriptive representation of a unique step in protocol execution.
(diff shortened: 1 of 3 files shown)


src/protocol/step/descriptive.rs
```diff
@@ -53,6 +54,7 @@ impl<S: Step + ?Sized> StepNarrow<S> for Descriptive {
             id += [std::any::type_name::<S>(), "::"].concat().as_ref();
         }
         id += step.as_ref();
+        metrics::increment_counter!(STEP_NARROWED, STEP => id.clone());
```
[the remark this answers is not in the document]
taikiy: Sure, I was wondering the same. More I think about it, it makes sense to hide this to save mem.

src/telemetry/mod.rs
```diff
@@ -98,5 +99,11 @@ pub mod metrics {
             Unit::Count,
             "Number of times PRSS is used as CPRNG to generate a random value"
         );
+
+        describe_counter!(
+            STEP_NARROWED,
+            Unit::Count,
+            "Trace steps each time they are narrowed"
```
akoshelev, line 106: maybe indicate that this is a counter in the description

src/telemetry/step_stats.rs
```diff
@@ -38,18 +40,19 @@ impl CsvExporter for Metrics {
         if self.print_header {
             writeln!(
                 w,
-                "Step,Records sent,Bytes sent,Indexed PRSS,Sequential PRSS"
+                "Step,Records sent,Bytes sent,Indexed PRSS,Sequential PRSS,Step narrowed"
```
akoshelev, line 43: depending on whether `step-trace` feature is enabled or no, the value in this column will be either always 0, or I assume in the range [0, 1]. Either way it is not very interesting to look at I think

akoshelev (approved):

src/telemetry/step_stats.rs
taikiy: Actually, this column will have more than just [0, <total_num_rows>] because some protocols skip 1 or 2 rows for optimization. For example, `BitOpStep` in `aggregate_credit::check_times_credit` sends 27 records and narrows 27 times. On the other hand, `compute_stop_bits` sends 27 records but narrows 3 times. (factor of 3 because of 3 helpers). We'll also have numbers like 0,0,0,0,3 which would tell us that the step is "no-comms". I find these quite useful.

PR #768, opened 2023-07-31
Relax normal distribution bounds for DP test even more

akoshelev:
https://github.com/private-attribution/ipa/actions/runs/5694472598


martinthomson (approved):
Yikes.  Is it proptest that is hitting this?

akoshelev:
> Yikes. Is it proptest that is hitting this?

yea, it seems efficient in hitting the corner cases.

PR #769, opened 2023-08-01
Improve logging in IPA

akoshelev:
* Use gate.as_ref() to print step names, rather than gate id
* Log reshare(vec) and shuffle steps
* Do not print IPA query config for every log line

```
ipa_query{query_config=IpaQueryConfig { per_user_credit_cap: 5, max_breakdown_key: 20, attribution_window_seconds: None, num_multi_bits: 3, plaintext_match_keys: false } query_size=500000}:sort_permutation:validate{gate=6939}: ipa::protocol::context::validator: new
```

is hard to read


richajaindce (approved):
Looks great!

PR #770, opened 2023-08-05
Add global panic hook and log destination endpoint

akoshelev:
See commit messages for details

### DIFF
--- a/src/net/transport.rs
+++ b/src/net/transport.rs
@@ -153,16 +153,12 @@ impl Transport for Arc<HttpTransport> {
                 let step =
                     <Option<Gate>>::from(route.gate()).expect("step required when sending records");
                 let resp_future = self.clients[dest].step(query_id, &step, data)?;
-                tokio::spawn(async move {
-                    resp_future
-                        .map_err(Into::into)
-                        .and_then(MpcHelperClient::resp_ok)
-                        .await
-                        .expect("failed to stream records");
-                });
-                // TODO(600): We need to do something better than panic if there is an error sending the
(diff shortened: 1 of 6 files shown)


src/net/transport.rs
```diff
@@ -153,16 +153,12 @@ impl Transport for Arc<HttpTransport> {
                 let step =
                     <Option<Gate>>::from(route.gate()).expect("step required when sending records");
                 let resp_future = self.clients[dest].step(query_id, &step, data)?;
-                tokio::spawn(async move {
```
akoshelev, line 156: @andyleiserson not sure why we had to spawn a task here, but all tests pass when I removed it. That turned out to be beneficial as `resp_future` no longer need to be `'static`

akoshelev:
this is how Hyper error messages look like after this change

```
Err(ConnectError { dest: "localhost:57123", inner: hyper::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) } }) })
```

```
2023-08-05T00:03:46.333834Z ERROR ipa::error: ThreadId(1) "main" panicked at 'called `Result::unwrap()` on an `Err` value: FailedHttpRequest { endpoint: "localhost:3000", status: 500, reason: "test mode to test handlers" }', src/bin/report_collector.rs:252:71
```

richajaindce (approved):
LGTM

src/net/transport.rs
andyleiserson: I can't remember either, seems reasonable to remove.

PR #771, opened 2023-08-08
Better proc-macro for Step enums

taikiy:
`#[step]` attribute macro expands to `#[derive(Step)]`, and the derive-macro will implement all necessary traits. The new macro also adds two attributes `obsolete` and `dynamic`. You can annotate an enum variant as `obsolete` so that the macro will correctly ignore it while macro expansion. For those steps that are dynamically executed (i.e., BitOpSteps), you can use `#[dynamic]` attribute.

Currently, I'm using `#[step]` attribute to generate all required trait derives for the annotated enum. The downside of this is that you can't add new derives unless you change the macro code. That said, I don't think there's a case where we want to add a derive to one specific step enum and not others. Attribute macro allows enum-wide attribute like `#[step(obsolete)]` which is nice. Without this, one needs to annotate each variant in the enum instead.


src/protocol/step/compact.rs
```diff
@@ -31,53 +32,30 @@ impl Debug for Compact {
 }
 
 const ROOT_STATE: u16 = 0;
-const FALLBACK_STATE: u16 = 65535;
-const UPGRADE_SEMI_HONEST_STATE: u16 = 65534;
+const PRSS_EXCHANGE_STATE: u16 = 65531;
 const QUERY_TYPE_SEMIHONEST_STATE: u16 = 65533;
 const QUERY_TYPE_MALICIOUS_STATE: u16 = 65532;
```
taikiy, line 37: still have these until we output `real-world-infra` narrows to steps.txt

akoshelev:
I wonder if making `step` attribute/macro very powerful is the direction we should pursue. It makes it convenient to write in the "happy" case, but makes it inflexible if custom step is required. I am also not in favor of generalizing `#dynamic` annotation - any step that requires less than or more than 64 elements would require a lot of customization or getting rid of step annotation completely. Same applies to automatically derived macros - leaving them out of scope for step derivation is the right thing to do imo.

taikiy:
BTW, I didn't use the `unique-type-id` crate since that uses an external toml file to manually assign IDs. Instead, the macro spits out an error when it detects a name conflict at compile time, letting the user know that the conflict must be resolved.

akoshelev:
chatted in TPAC about it - while I remain concerned about step macro being too powerful and general concern about proc-macros debuggability/readability/maintainability, I think we should push this change. We realized that `[#obsolete]` attribute is no longer required as we track step transitions inside  the`narrow` and the remaining usages are inside the dead code only, so there will be less features for `step` macro. 

I would also prefer to not use `dynamic` as the overhead of using `repeat64str` is pretty low compared to readability, but it is not the hill I want to die on. 

@taikiy is going to clean up some dead code and unless someone objects, push this change

akoshelev (approved):

PR #775, opened 2023-08-11
Generalized hpke

akoshelev:
As we are preparing to create a separate aggregation circuit, we need better names for HPKE primitives used in IPA. We no longer encrypt shares of match keys only, for aggregation we want to encrypt boolean shares of various sizes. We will start with 1 byte shares

This change just moves things around and does a bunch of renames. No functional changes

### DIFF
--- a/src/hpke/mod.rs
+++ b/src/hpke/mod.rs
@@ -4,8 +4,8 @@
 
 use generic_array::ArrayLength;
 use hpke::{
-    aead::AeadTag, generic_array::typenum::Unsigned, single_shot_open_in_place_detached,
-    single_shot_seal_in_place_detached, OpModeR, OpModeS,
+    aead::AeadTag, single_shot_open_in_place_detached, single_shot_seal_in_place_detached, OpModeR,
+    OpModeS,
 };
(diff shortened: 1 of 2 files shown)


src/hpke/mod.rs
```diff
@@ -29,44 +29,32 @@ type IpaKem = hpke::kem::X25519HkdfSha256;
...
 
-/// match key size, in bytes
-const MATCHKEY_LEN: usize = <XorReplicated as IpaSerializable>::Size::USIZE;
-
-/// Total len in bytes for an encrypted matchkey including the authentication tag.
-pub const MATCHKEY_CT_LEN: usize =
-    MATCHKEY_LEN + <AeadTag<IpaAead> as Serializable>::OutputSize::USIZE;
-
-pub trait MatchKeyCrypt: GaloisField + IpaSerializable {
+pub trait FieldShareCrypt: GaloisField + IpaSerializable {
```
richajaindce, line 37: Can we rename it to something better? Just SecretShareCrypt? Not sure if "field" in the name makes it a bit confusing?

richajaindce (approved):

PR #777, opened 2023-08-16
Add aggregate query type

eriktaubeneck:
initial changes to add an `Aggregate` query type, which will do a "prio-like" aggregation, but using modulus conversion instead of proofs.


eriktaubeneck:
right now, the two `QueryType:Aggregate` types don't have any config, but we'll probably want to include the number of histogram bins being aggregated as an query parameter.

taikiy:
We'll also need to take the maximum number of bits in the input bit-decomposed shares (`N` in Alex's document)

src/query/runner/aggregate.rs
```diff
@@ -0,0 +1,84 @@
...
+    ) -> Result<Vec<Replicated<F>>, Error> {
+        todo!()
+    }
+}
+
+/// Helps to convince the compiler that things are `Send`. Like `seq_join::assert_send`, but for
+/// streams.
+///
+/// <https://github.com/rust-lang/rust/issues/102211#issuecomment-1367900125>
+pub fn _assert_stream_send<'a, T>(
```
akoshelev, line 80: we already have it, so this one is probably not needed

akoshelev (approved):
lgtm

PR #779, opened 2023-08-16
Renew test certificates

akoshelev:
They were valid for 90 days and have expired which caused tests to fail

https://github.com/private-attribution/ipa/actions/runs/5872910100/job/15925288484

This change updates test certificates and private keys with certificates being valid for 50 years.


eriktaubeneck:
should we add a test that fails if today's date is greater than Aug 15, 2073? /s

PR #780, opened 2023-08-21
Secure aggregation protocol interface

taikiy:
This PR hooks up the aggregate query request to the internal protocol to be called to perform a simple aggregation of binary shares. The protocol itself is not implemented yet. Coming next.

### DIFF
--- a/src/protocol/aggregation/mod.rs
+++ b/src/protocol/aggregation/mod.rs
@@ -1,0 +1,146 @@
+mod input;
+use std::iter::zip;
+
+use futures::future::try_join;
+pub use input::AggregateInputRow;
+use ipa_macros::step;
+use strum::AsRefStr;
+
+use self::input::BinarySharedAggregateInputs;
+use crate::{
+    error::Error,
+    ff::{Field, GaloisField, Gf2, PrimeField},
+    protocol::{
+        context::{Context, UpgradableContext, UpgradedContext, Validator},
+        BasicProtocols,
+    },
+    secret_sharing::{
+        replicated::{
+            malicious::{DowngradeMalicious, ExtendableField},
+            semi_honest::AdditiveShare as Replicated,
+            ReplicatedSecretSharing,
+        },
+        BitDecomposed, Linear as LinearSecretSharing,
+    },
+};
+
+#[step]
+pub(crate) enum Step {
+    BinaryValidator,
+    UpgradeValueBits,
+    UpgradeBreakdownKeyBits,
+}
+
+/// Binary-share aggregation protocol.
+///
+/// # Errors
+/// Propagates errors from multiplications
+pub async fn aggregate<'a, C, SB, F, V, BK>(
+    sh_ctx: C,
+    input_rows: &[AggregateInputRow<V, BK>],
+) -> Result<Vec<Replicated<F>>, Error>
+where
+    C: UpgradableContext,
+    C::UpgradedContext<Gf2>: UpgradedContext<Gf2, Share = SB>,
+    SB: LinearSecretSharing<Gf2>
+        + BasicProtocols<C::UpgradedContext<Gf2>, Gf2>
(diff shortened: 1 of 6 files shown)


src/protocol/aggregation/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    VB: Validator<C, Gf2>,
+    C: UpgradableContext<Validator<F> = V>,
+    C::UpgradedContext<Gf2>: UpgradedContext<Gf2, Share = SB> + Context,
+    SB: LinearSecretSharing<Gf2>
+        + BasicProtocols<C::UpgradedContext<Gf2>, Gf2>
+        + DowngradeMalicious<Target = Replicated<Gf2>>
+        + 'static,
+    F: PrimeField + ExtendableField,
+{
+    // TODO(taikiy):
```
eriktaubeneck, line 95: what we wrote down last week was:
1. generate `num_contributions` streams (one for each contribution bucket)
2. (fan out) in parallel, for each stream:
    3. mod convert bits
    4. aggregate the bits
    5. validate the aggregate (once per stream, at the end of all aggregation)
    6. return the share of the aggregate
7. (fan in) collect the shares of the aggregate per stream
8. output the vector of shares

with relation to this list, we shouldn't need to apply capping (we get it "for free" in mod convert, so long as the cap is a value of 2^n). also, the breakdown keys aren't specifically provided (not sure if that's implied here).

martinthomson: Yeah, this doesn't prevent someone from dumping 2^n-1 in every bucket.  An alternative encoding might be "bucket number, bucket value".  The cost of that is quite a bit more though because you have to modulus convert the value and produce a value in Fp, which is then multiplied once for each possible bucket.  For B buckets, there are lots of comparisons involved, even if we make those more efficient the Sierpinski approach.

There is also the option where you add up all the bits and do a range check.  A single comparison across all buckets might be easy for one-bit inputs; accumulating all values and range checking that means picking up the solved bits implementation.  I don't know if we can squeeze any extra performance out if we are willing to have out-of-range values revealed...

eriktaubeneck: I'm not sure I follow. As I understood it in this setting, individual reports need to be capped. Assuming that cap falls at a value of 2^n, we can enforce that cap simply by only using n bits. 

Modulus conversion then allows us to move from a space where we can add up lots of events from distinct reports without overflow. 

I'm not even sure how capping would work here, as reports aren't tied to specific people.

martinthomson: A simple noise addition, like the one you describe, is probably fine, but a tighter bound might be achieved if we know that the user can only contribute one bucket's worth.

benjaminsavage: OK, let's take a step back and talk about what the MPC needs to do, and why.

My understanding is this:
- We are *not* relying on the MPC to perform per-user capping across multiple reports. We are relying on the browser / device to protect the privacy of individuals by capping their total contribution across an epoch.
- We are *not* concerned about protecting the *privacy* of individuals who have hacked / malicious clients (we can't). So we do *not* need to worry about checking that the per-user cap is not exceeded on the MPC side.
- We *are* concerned about hacked / invalid clients screwing up measurement results. Our approach follows that of PRIO, which is to ensure the maximum contribution one client can make to the total measurement result is at most X. We are using a "modulus conversion" approach to ensure this instead of a ZKP approach.

Per this reasoning, I do not think the MPC needs to do any form of per-user capping, either across multiple contributions from a single browser / device, or across multiple buckets. 

Do people disagree with these conclusions? If so why?

benjaminsavage (approved):

src/protocol/aggregation/mod.rs
eriktaubeneck: @benjaminsavage this was my understanding

PR #781, opened 2023-08-24
Implement aggregate protocol 

taikiy:
Actual implementation of the simple aggregation protocol.

### DIFF
--- a/src/protocol/aggregation/mod.rs
+++ b/src/protocol/aggregation/mod.rs
@@ -1,17 +1,18 @@
 mod input;
-use std::iter::zip;
 
-use futures::future::try_join;
+use futures::{stream::iter as stream_iter, TryStreamExt};
 pub use input::AggregateInputRow;
 use ipa_macros::step;
 use strum::AsRefStr;
 
-use self::input::BinarySharedAggregateInputs;
+use super::{basics::Reshare, RecordId};
 use crate::{
     error::Error,
-    ff::{Field, GaloisField, Gf2, PrimeField},
+    ff::{Field, GaloisField, Gf2, PrimeField, Serializable},
+    helpers::query::AggregateQueryConfig,
     protocol::{
-        context::{Context, UpgradableContext, UpgradedContext, Validator},
+        context::{UpgradableContext, UpgradedContext, Validator},
+        modulus_conversion::convert_bits,
         BasicProtocols,
     },
     secret_sharing::{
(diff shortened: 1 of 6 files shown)


src/protocol/aggregation/mod.rs
```diff
@@ -26,110 +27,124 @@ use crate::{
...
-        .map(|(value, breakdown_key)| BinarySharedAggregateInputs::new(value, breakdown_key))
-        .collect::<Vec<_>>();
-
-    secure_aggregation(validator, binary_shared_values).await
+
+    // TODO(taikiy):
+    // 1. slice the buckets into N streams and send them to the aggregation protocol
+    // 2. collect the results and return them
+
+    let output = aggregate_values(validator.context(), gf2_value_bits).await?;
```
benjaminsavage, line 71: As @martinthomson points out, we can begin with a simple implementation that matches the IPA Aggregation step. It doesn't scale well with large numbers of breakdown keys but it works like this:

1. Use the bit-decomposed breakdown_key value to compute an "equivalence check" with every possible breakdown (value either 0 or 1). This can be done in less multiplications using the "Sterpinkski Thing" we have already implemented.
2. Multiply the contribution by these "equivalence checks" (one multiplication per breakdown... not awesome scaling in the event there are many breakdown keys.
3. Send these multiplication results to the N streams. Most of the times you'll be sending a sharing of zero.

This is fine for now. We should see if we can come up with a more scalable approach.

src/protocol/aggregation/mod.rs
```diff
@@ -26,110 +27,124 @@ use crate::{
...
+
+    validator.validate(vec![output]).await
 }
 
 /// Performs a set of aggregation protocols on binary shared values.
+/// This protocol assumes that the input rows are from the same match key and
+/// for the same breakdown key, so that we don't have to worry about computing
+/// the helper_bits or capping.
```
benjaminsavage, line 79: We are not assuming they are all from the same "match key". There are no match keys. We aren't assuming it is from the same device either. We are assuming devices / browsers are properly applying per-user capping to protect people who are using legitimate software.

src/protocol/aggregation/mod.rs
```diff
@@ -26,110 +27,124 @@ use crate::{
...
+    let convert_ctx = ctx
+        .narrow(&Step::ConvertValueBits)
+        .set_total_records(record_count);
+    let converted_contribution_values = convert_bits(
+        convert_ctx,
+        stream_iter(contribution_value_bits_gf2),
+        0..u32::try_from(bits).unwrap(),
+    );
+
+    //TODO: retrieve the previous aggregate value from the storage
```
benjaminsavage, line 106: For now, I recommend we keep the design simple by keeping no state on the Aggregation servers. The report collector already has the result of the last query. They can add it to this result if they want to.

src/protocol/aggregation/mod.rs
```diff
@@ -26,110 +27,124 @@ use crate::{
...
+    let aggregate = converted_contribution_values
+        .try_fold(S::ZERO, |mut acc, row| async move {
+            acc += &row.as_decimal();
+            Ok(acc)
+        })
+        .await?;
+
+    //TODO:
+    // * add the new aggregate value to the previous one
+    // * add DP noise relative to cap and max # of reports
```
benjaminsavage, line 117: No. The DP noise is not related in any way to the # of reports. It's only related to the cap.

src/protocol/aggregation/mod.rs
```diff
@@ -26,110 +27,124 @@ use crate::{
...
+    // * store the new aggregate value
```
benjaminsavage, line 118: As above, let's go with no storage for now.

src/protocol/aggregation/mod.rs
```diff
@@ -144,3 +159,44 @@ where
...
+        const INPUT: &[u32] = &[0, 0, 18, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 10, 0, 0, 6, 0];
+
+        let world = TestWorld::default();
+        let result = world
+            .semi_honest(
+                INPUT.iter().map(|&value| {
+                    BitDecomposed::decompose(CONTRIBUTION_BITS, |i| {
+                        Gf2::try_from((u128::from(value) >> i) & 1).unwrap()
+                    })
+                }),
```
benjaminsavage, line 187: Didn't we normally hide this type of test logic within a trait that performs this logic across all tests?

src/protocol/aggregation/mod.rs
```diff
@@ -144,3 +159,44 @@ where
...
+    #[tokio::test]
+    pub async fn aggregate() {
+        const CONTRIBUTION_BITS: u32 = 8;
+        const EXPECTED: u128 = 36;
+
+        const INPUT: &[u32] = &[0, 0, 18, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 10, 0, 0, 6, 0];
+
+        let world = TestWorld::default();
+        let result = world
+            .semi_honest(
```
benjaminsavage, line 182: Would be good to test the Malicious version as well.

src/secret_sharing/decomposed.rs
```diff
@@ -39,6 +39,24 @@ impl<S> BitDecomposed<S> {
...
+
+    pub fn len(&self) -> usize {
+        return self.bits.len();
+    }
+
+    /// Convert this into a base-10 field value.
+    /// The inner vector of this type is a list any field type (e.g. Z2, Zp) and
+    /// each element is (should be) a share of 1 or 0. This function iterates
+    /// over the shares of bits and computes Σ(2^i * b_i).
+    pub fn as_decimal<F>(&self) -> S
```
benjaminsavage, line 57: I'm going to quibble with you on the name of this one. Decimal implies base ten. We aren't using base ten. Can we call it something like: to_additive_sharing_in_large_field().

src/protocol/aggregation/mod.rs
benjaminsavage: To be clear - I'm suggesting this is the next PR, not this one.

PR #782, opened 2023-08-25
Rename "aggregation" to "sparse aggregation" and fix clippy

benjaminsavage:
Just renamed a thing and made clippy happy


taikiy (approved):

src/query/runner/aggregate.rs
taikiy, line 1: Let's change the name of the top-level protocol function name at line 95:
```
- aggregate(ctx, input.as_slice()).await
+ sparse_aggregate(ctx, input.as_slice()).await
```

taikiy (approved):

richajaindce:
The title of the PR doesnt reflect what it does :P Can we rename it?

PR #783, opened 2023-08-25
sparse aggregation protocol

taikiy:
clippy is updated and complaining. To be rebased on top of #782.

This diff implements the sparse aggregation protocol. The input is a vector of `(contribution_value, breakdown_key)` pairs where `breakdown_key` indicates to which bucket the `contribution_value` should be added. Since the breakdown keys are secret shared, we use `check_everything` (Sierpiński triangle thing) for equality checks. A contribution value is multiplied by a share of 1 or 0 so that the value is only contributed when the breakdown_key matches the bucket index.

### DIFF
--- a/src/protocol/aggregation/mod.rs
+++ b/src/protocol/aggregation/mod.rs
@@ -36,18 +36,19 @@ pub(crate) enum Step {
 }
 
 /// Binary-share aggregation protocol for a sparse breakdown key vector input.
-/// It takes a tuple of two vectors, contribution_values and breakdown_keys, and
-/// aggregate each value to the corresponding histogram bucket specified by the
-/// breakdown key. Since breakdown keys are secret shared, we need to create a
-/// vector of Z2 shares for each record indicating which bucket the value should
-/// be aggregated to. The output is a vector of Zp shares - a histogram of the
-/// aggregated values.
+/// It takes a tuple of two vectors, `contribution_values` and `breakdown_keys`,
+/// and aggregate each value to the corresponding histogram bucket specified by
+/// the breakdown key. Since breakdown keys are secret shared, we need to create
+/// a vector of Z2 shares for each record indicating which bucket the value
+/// should be aggregated to. The output is a vector of Zp shares - a histogram
+/// of the aggregated values.
 ///
 /// # Errors
 /// Propagates errors from multiplications
 pub async fn sparse_aggregate<'a, C, S, SB, F, CV, BK>(
     sh_ctx: C,
     input_rows: &[SparseAggregateInputRow<CV, BK>],
+    num_buckets: usize,
 ) -> Result<Vec<Replicated<F>>, Error>
 where
     C: UpgradableContext,
@@ -93,9 +94,13 @@ where
     )
     .await?;
 
-    let output =
-        aggregate_values_per_bucket(ctx, converted_value_bits, converted_breakdown_key_bits)
-            .await?;
+    let output = sparse_aggregate_values_per_bucket(
+        ctx,
+        converted_value_bits,
+        converted_breakdown_key_bits,
+        num_buckets,
+    )
+    .await?;
 
     validator.validate(output).await
 }
@@ -106,10 +111,11 @@ where
 /// # Errors
 /// propagates errors from multiplications
 #[tracing::instrument(name = "aggregate_values_per_bucket", skip_all)]
-pub async fn aggregate_values_per_bucket<F, C, S>(
+pub async fn sparse_aggregate_values_per_bucket<F, C, S>(
     ctx: C,
     contribution_values: Vec<BitDecomposed<S>>,
     breakdown_keys: Vec<BitDecomposed<S>>,
+    num_buckets: usize,
 ) -> Result<Vec<S>, Error>
 where
     F: PrimeField,
@@ -118,14 +124,12 @@ where
 {
     debug_assert!(contribution_values.len() == breakdown_keys.len());
     let num_records = contribution_values.len();
-    // for now, we assume that the bucket count is 2^BK::BITS
-    let num_buckets = 1 << breakdown_keys[0].len();
 
     let equality_check_ctx = ctx
         .narrow(&Step::ComputeEqualityChecks)
         .set_total_records(num_records);
 
-    // Generate N streams for each bucket specified by the breakdown key (N = |breakdown_keys|).
+    // Generate N streams for each bucket specified by the `num_buckets`.
     // A stream is pipeline of contribution values multiplied by the "equality bit". An equality
     // bit is a bit that is a share of 1 if the breakdown key matches the bucket, or 0 otherwise.
     let streams = seq_join(
@@ -180,12 +184,8 @@ where
                 let step = BitOpStep::from(check_idx);
                 let c = check_times_value_ctx.narrow(&step);
                 let record_id = RecordId::from(record_id);
-                let v = &value_bits;
-                async move {
-                    check
-                        .multiply(&v.to_additive_sharing_in_large_field(), c, record_id)
-                        .await
-                }
+                let v = value_bits.to_additive_sharing_in_large_field();
+                async move { check.multiply(&v, c, record_id).await }
             }),
     )
     .await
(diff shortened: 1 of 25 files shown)


src/protocol/aggregation/mod.rs
```diff
@@ -21,19 +24,23 @@ use crate::{
...
 pub(crate) enum Step {
     Validator,
     ConvertValueBits,
+    ConvertBreakdownKeyBits,
+    ComputeEqualityChecks,
+    CheckTimesValue,
 }
 
-/// Binary-share aggregation protocol.
+/// Binary-share aggregation protocol for a sparse breakdown key vector input.
```
richajaindce, line 39: Since this is the interface of this protocol, Can you also add a comment about the input and output format i.e. an array of [conversion value, breakdown key] and also output

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +58,229 @@ where
...
+            CV::BITS,
+            |row, i| {
+                Replicated::new(
+                    Gf2::truncate_from(row.contribution_value.left()[i]),
+                    Gf2::truncate_from(row.contribution_value.right()[i]),
+                )
+            },
+        )
+        .await?,
+        upgrade_bit_shares(
```
benjaminsavage, line 77: This can be done in parallel with the line above.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +58,229 @@ where
...
+        .try_fold(vec![S::ZERO; num_buckets], |mut acc, bucket| async move {
+            for (i, b) in bucket.into_iter().enumerate() {
+                acc[i] += &b;
+            }
             Ok(acc)
         })
-        .await?;
-
-    Ok(aggregate)
+        .await
```
benjaminsavage, line 173: I might be misunderstanding this, but this looks like it's operating across all of the streams, but I thought each stream was able to progress independently.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +58,229 @@ where
...
+                                                &value.to_additive_sharing_in_large_field(),
+                                                c,
+                                                record_id,
+                                            )
+                                            .await
+                                    }
+                                }),
+                        )
+                        .await
+                }
```
benjaminsavage, line 162: I find this really difficult to read / digest / understand.

Is there anything we can do to improve the code readability here? Perhaps something that leads to less indentation? Maybe factor something out into another function?

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +58,229 @@ where
...
+    let streams = seq_join(
+        ctx.active_work(),
+        stream_iter(breakdown_keys)
+            .zip(stream_iter(contribution_values))
+            .enumerate()
+            .map(|(i, (bk, v))| {
+                let eq_ctx = &equality_check_ctx;
+                let mul_ctx = &check_times_value_ctx;
+                async move {
+                    let equality_checks = check_everything(eq_ctx.clone(), i, &bk).await?;
```
benjaminsavage, line 138: Wait a minute... doesn't "i" refer to the index of the record in the stream of values? Why would we check for equality against that?

benjaminsavage (changes requested):
The tests pass, so the logic is clearly accurate, but I find it quite hard to follow. I would like to try to find a way to make it easier to read / digest / understand.

src/protocol/aggregation/mod.rs
taikiy: Changed to use `try_join` on the tuples.

src/protocol/aggregation/mod.rs
taikiy: I think `try_fold` applies async computations on each stream independently. cc: @martinthomson

src/protocol/aggregation/mod.rs
taikiy: `i` passed to `check_everything` is used as a record ID. `check_everything` uses the record ID to execute `pregenerate_all_combinations` once per record.

src/protocol/aggregation/mod.rs
taikiy: Extracted the multiplication part.

src/protocol/aggregation/mod.rs
benjaminsavage: Ah, my bad. Makes sense.

src/protocol/aggregation/mod.rs
martinthomson: I'm pretty sure that this will work sequentially.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +58,229 @@ where
...
     pub async fn aggregate() {
-        const CONTRIBUTION_BITS: u32 = 8;
-        const EXPECTED: u128 = 36;
+        type CV = Gf8Bit;
+        type BK = Gf3Bit;
 
-        const INPUT: &[u32] = &[0, 0, 18, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 10, 0, 0, 6, 0];
+        const EXPECTED: &[u128] = &[28, 0, 0, 6, 1, 0, 0, 8];
+
+        const INPUT: &[(u32, u32)] = &[
```
martinthomson, line 257: I had trouble reading the ordering here.  I read this as a k-v map, where it is a v-k map instead.

src/protocol/aggregation/mod.rs
benjaminsavage: @akoshelev and @martinthomson - can you please advise on how to write this code for maximum streaming performance?

src/protocol/aggregation/mod.rs
benjaminsavage: I also got caught up on the same thing.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +64,241 @@ where
...
+        stream_iter(breakdown_keys)
+            .zip(stream_iter(contribution_values))
+            .enumerate()
+            .map(|(i, (bk, v))| {
+                let eq_ctx = &equality_check_ctx;
+                let c = ctx.clone();
+                async move {
+                    let equality_checks = check_everything(eq_ctx.clone(), i, &bk).await?;
+                    equality_bits_times_value(&c, equality_checks, num_buckets, v, num_records, i)
+                        .await
```
benjaminsavage, line 142: My assumption is that we want to send work over to other streams right here (as opposed to in a second loop).

Right at this point (after this await) we should have N secret-shares. I think this is the place to just iterate over all of them and essentially just Send them over to other threads for processing.

I think creating a second loop *might* work, but it's confusing because it requires you to understand that the loop is being evaluate in a lazy way...

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +64,241 @@ where
...
+        .narrow(&Step::CheckTimesValue)
+        .set_total_records(num_records);
+
+    ctx.try_join(
+        check_bits
+            .into_iter()
+            .take(num_buckets)
+            .enumerate()
+            .map(|(check_idx, check)| {
+                let step = BitOpStep::from(check_idx);
```
benjaminsavage, line 180: We are kind of abusing "BitOpStep" now. This was designed for bitwise operations and cannot go above 64. There might be thousands of breakdowns - so maybe we should think more about this.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +64,241 @@ where
...
+                let c = check_times_value_ctx.narrow(&step);
+                let record_id = RecordId::from(record_id);
+                let v = &value_bits;
+                async move {
+                    check
+                        .multiply(&v.to_additive_sharing_in_large_field(), c, record_id)
```
benjaminsavage, line 186: We should not be doing this in a loop. We should do it only once outside of the loop.

src/protocol/aggregation/mod.rs
```diff
@@ -52,139 +64,241 @@ where
...
+    .await
 }
 
-fn get_gf2_breakdown_key_bits<CV, BK>(
+async fn upgrade_bit_shares<F, C, S, H, CV, BK>(
+    ctx: C,
     input_rows: &[SparseAggregateInputRow<CV, BK>],
-) -> Vec<BitDecomposed<Replicated<Gf2>>>
+    num_bits: u32,
+    f: H,
```
benjaminsavage, line 198: Not clear from the name (f) or the comment (non-existing) what this param does

src/protocol/aggregation/mod.rs
taikiy: I have a PR in review that could solve this. We can create a step like BitOpStep but with a number of our choice.

src/protocol/aggregation/mod.rs
taikiy: Added the comment.

martinthomson (commented):
OK, I got thinking...

src/protocol/aggregation/mod.rs
martinthomson: Funny.  I had the same comment, but I hadn't even thought of the performance implication, just that this function had no business dealing with the conversion.

src/protocol/aggregation/mod.rs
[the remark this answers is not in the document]
benjaminsavage: I like `binary_to_onehot()` or `bitwise_to_onehot()`

taikiy:
Thanks for reviewing!
I updated the PR according to the comments but now I'm getting this error and don't know how to resolve...

```
higher-ranked lifetime error
could not prove `[async block@src/protocol/aggregation/mod.rs:271:63: 279:14]: std::marker::Send`
```
[code](https://github.com/private-attribution/ipa/pull/783/commits/477878fbc62c57ca9dd52e7679a6e87ae59630db#diff-6877f2b84a12df745aab6837412ad7cd69b0ceccb89b32730ffb67291e0f9737R271-R279)

I found a post on the web that solved the similar error by wrapping the function and simply returning it with `Send` bound signature, but that didn't seem to work. @martinthomson, @akoshelev, any idea how I can solve this? Thanks in advance.

akoshelev:
> Thanks for reviewing! I updated the PR according to the comments but now I'm getting this error and don't know how to resolve...
> 
> ```
> higher-ranked lifetime error
> could not prove `[async block@src/protocol/aggregation/mod.rs:271:63: 279:14]: std::marker::Send`
> ```
> 
> [code](https://github.com/private-attribution/ipa/pull/783/commits/477878fbc62c57ca9dd52e7679a6e87ae59630db#diff-6877f2b84a12df745aab6837412ad7cd69b0ceccb89b32730ffb67291e0f9737R271-R279)
> 
> I found a post on the web that solved the similar error by wrapping the function and simply returning it with `Send` bound signature, but that didn't seem to work. @martinthomson, @akoshelev, any idea how I can solve this? Thanks in advance.

you could try https://github.com/private-attribution/ipa/blob/e3d974848542131ad298ef590335a8dd3bb903f9/src/seq_join.rs#L17 - seems to be the same issue

benjaminsavage (approved):

PR #785, opened 2023-09-20
Attribution and Capping for OPRF approach

benjaminsavage:
This PR correctly implements the new OPRF approach to attribution and per-user contribution capping.

Aggregation is left for another day.

This uses a bitwise saturating sum approach to perform the contribution capping, where the caller specifies how many bits to use.

All the logic is reflected in this Google sheet which implements the same oblivious algorithm: https://docs.google.com/spreadsheets/d/1-6cvFe7GbhdzuiaTuiZ9ojq36rA4AVMjBvTqvvpJaHw/edit?usp=sharing

While multiple users can be run in parallel (using a try_join_all for now), each user's circuit is rather deep, with each row of data depending on the preceding row. This can lead to very deep circuits for users with many rows of data.

### DIFF
--- a/src/protocol/boolean/saturating_sum.rs
+++ b/src/protocol/boolean/saturating_sum.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::Error,
-    ff::Gf2,
+    ff::{Field, Gf2},
     protocol::{boolean::or::or, context::Context, step::BitOpStep, BasicProtocols, RecordId},
     secret_sharing::{BitDecomposed, Linear as LinearSecretSharing},
 };
@@ -59,6 +59,39 @@ impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
             is_saturated,
         ))
     }
+
+    ///
+    /// NOTE: ignores the `is_saturated` flag. The return value is non-sensical if `is_saturated` is true
+    ///
+    /// Only returns the least significant `num_bits` of the delta.
+    ///
+    pub async fn truncated_delta_to_saturation_point<C>(
+        &self,
+        ctx: C,
+        record_id: RecordId,
+        num_bits: usize,
+    ) -> Result<BitDecomposed<S>, Error>
+    where
+        C: Context,
+        S: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+    {
+        assert!(num_bits <= self.sum.len());
+
+        let mut carry_in = S::share_known_value(&ctx, Gf2::ONE);
+        let mut output = vec![];
+        for (i, bit) in self.sum.iter().enumerate().take(num_bits) {
+            let c = ctx.narrow(&BitOpStep::from(i));
+
+            let compute_carry_out = i < num_bits - 1;
+            let (difference_bit, carry_out) =
+                one_bit_subtractor(c, record_id, &S::ZERO, bit, &carry_in, compute_carry_out)
+                    .await?;
+
+            output.push(difference_bit);
+            carry_in = carry_out;
+        }
+        Ok(BitDecomposed::new(output))
+    }
 }
 
 ///
@@ -103,6 +136,51 @@ where
         + carry_in;
 
     Ok((sum_bit, carry_out))
+}
+
+///
+/// This improved one-bit subtractor that only requires a single multiplication was taken from:
+/// "Improved Garbled Circuit Building Blocks and Applications to Auctions and Computing Minima"
+/// `https://encrypto.de/papers/KSS09.pdf`
+///
+/// Section 3.1 Integer Addition, Subtraction and Multiplication
+///
+/// For each bit, the `difference_bit` can be efficiently computed as just `d_i = x_i ⊕ !y_i ⊕ c_i`
+/// This can be computed "for free" in Gf2
+///
+/// The `carry_out` bit can be efficiently computed with just a single multiplication as:
+/// `c_(i+1) = c_i ⊕ ((x_i ⊕ c_i) ∧ !(y_i ⊕ c_i))`
+///
+/// Returns (`difference_bit`, `carry_out`)
+///
+async fn one_bit_subtractor<C, SB>(
+    ctx: C,
+    record_id: RecordId,
+    x: &SB,
+    y: &SB,
+    carry_in: &SB,
+    compute_carry_out: bool,
+) -> Result<(SB, SB), Error>
+where
+    C: Context,
+    SB: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    // compute difference bit as not_y XOR x XOR carry_in
+    let difference_bit = SB::share_known_value(&ctx, Gf2::ONE) - y + x + carry_in;
+    if compute_carry_out {
+        let x_xor_carry_in = x.clone() + carry_in;
+        let y_xor_carry_in = y.clone() + carry_in;
+        let not_y_xor_carry_in = SB::share_known_value(&ctx, Gf2::ONE) - &y_xor_carry_in;
+
+        let carry_out = x_xor_carry_in
+            .multiply(&not_y_xor_carry_in, ctx, record_id)
+            .await?
+            + carry_in;
+
+        Ok((difference_bit, carry_out))
+    } else {
+        Ok((difference_bit, SB::ZERO))
+    }
 }
 
 #[cfg(all(test, unit_test))]
--- a/src/protocol/boolean/saturating_sum.rs
+++ b/src/protocol/boolean/saturating_sum.rs
@@ -40,11 +40,9 @@ impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
             // we still must compute the carries, which still requires a single multiplication
             // so there is no savings
             let x = value.get(i).unwrap_or(&zero);
-            let (sum_bit, carry_out) =
-                one_bit_adder(c, record_id, x, &self.sum[i], &carry_in).await?;
+            let sum_bit = one_bit_adder(c, record_id, x, &self.sum[i], &mut carry_in).await?;
 
             output_sum.push(sum_bit);
-            carry_in = carry_out;
         }
         let is_saturated = or(
             ctx.narrow(&BitOpStep::from(self.sum.len())),
@@ -83,12 +81,17 @@ impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
             let c = ctx.narrow(&BitOpStep::from(i));
 
             let compute_carry_out = i < num_bits - 1;
-            let (difference_bit, carry_out) =
-                one_bit_subtractor(c, record_id, &S::ZERO, bit, &carry_in, compute_carry_out)
-                    .await?;
+            let difference_bit = one_bit_subtractor(
+                c,
+                record_id,
+                &S::ZERO,
+                bit,
+                &mut carry_in,
+                compute_carry_out,
+            )
+            .await?;
 
             output.push(difference_bit);
-            carry_in = carry_out;
         }
         Ok(BitDecomposed::new(output))
     }
@@ -107,15 +110,17 @@ impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
 /// The `carry_out` bit can be efficiently computed with just a single multiplication as:
 /// `c_(i+1) = c_i ⊕ ((x_i ⊕ c_i) ∧ (y_i ⊕ c_i))`
 ///
-/// Returns (`sum_bit`, `carry_out`)
+/// Returns `sum_bit`
+///
+/// The mutable refernce to `carry_in` is mutated to take on the value of the `carry_out` bit
 ///
 async fn one_bit_adder<C, SB>(
     ctx: C,
     record_id: RecordId,
     x: &SB,
     y: &SB,
-    carry_in: &SB,
-) -> Result<(SB, SB), Error>
+    carry_in: &mut SB,
+) -> Result<SB, Error>
 where
     C: Context,
     SB: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
@@ -130,12 +135,12 @@ where
     // (1) When the `carry_in` bit is 0 and both `x` and `y` are 1
     // (2) When the `carry_in` bit is 1 and both `x` and `y` are 0
     // So by computing `(x ⊕ c) ∧ (y ⊕ c)` we isolate those cases with a single multiplication
-    let carry_out = x_xor_carry_in
+    *carry_in = x_xor_carry_in
         .multiply(&y_xor_carry_in, ctx, record_id)
         .await?
         + carry_in;
 
-    Ok((sum_bit, carry_out))
+    Ok(sum_bit)
 }
 
 ///
@@ -151,16 +156,18 @@ where
 /// The `carry_out` bit can be efficiently computed with just a single multiplication as:
 /// `c_(i+1) = c_i ⊕ ((x_i ⊕ c_i) ∧ !(y_i ⊕ c_i))`
 ///
-/// Returns (`difference_bit`, `carry_out`)
+/// Returns `difference_bit`
+///
+/// If `compute_carry_out` is set to `true`, then the mutable refernce to `carry_in` is mutated to take on the value of the `carry_out` bit
 ///
 async fn one_bit_subtractor<C, SB>(
     ctx: C,
     record_id: RecordId,
     x: &SB,
     y: &SB,
-    carry_in: &SB,
+    carry_in: &mut SB,
     compute_carry_out: bool,
-) -> Result<(SB, SB), Error>
+) -> Result<SB, Error>
 where
     C: Context,
     SB: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
@@ -172,15 +179,12 @@ where
         let y_xor_carry_in = y.clone() + carry_in;
         let not_y_xor_carry_in = SB::share_known_value(&ctx, Gf2::ONE) - &y_xor_carry_in;
 
-        let carry_out = x_xor_carry_in
+        *carry_in = x_xor_carry_in
             .multiply(&not_y_xor_carry_in, ctx, record_id)
             .await?
             + carry_in;
-
-        Ok((difference_bit, carry_out))
-    } else {
-        Ok((difference_bit, SB::ZERO))
     }
+    Ok(difference_bit)
 }
 
 #[cfg(all(test, unit_test))]
--- a/src/protocol/boolean/saturating_sum.rs
+++ b/src/protocol/boolean/saturating_sum.rs
@@ -1,0 +1,171 @@
+use crate::{
+    error::Error,
+    ff::Gf2,
+    protocol::{context::Context, step::BitOpStep, BasicProtocols, RecordId},
+    secret_sharing::{BitDecomposed, Linear as LinearSecretSharing},
+};
+
+#[derive(Debug)]
+pub struct SaturatingSum<S: LinearSecretSharing<Gf2>> {
+    pub sum: BitDecomposed<S>,
+    pub is_saturated: S,
+}
+
+impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
+    pub fn new(value: BitDecomposed<S>, is_saturated: S) -> SaturatingSum<S> {
+        SaturatingSum {
+            sum: value,
+            is_saturated,
+        }
+    }
+
+    pub async fn add<C>(
+        &self,
+        ctx: C,
+        record_id: RecordId,
+        value: &BitDecomposed<S>,
+    ) -> Result<SaturatingSum<S>, Error>
+    where
+        C: Context,
+        S: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+    {
+        assert!(self.sum.len() >= value.len());
+
+        let mut output_sum = Vec::with_capacity(self.sum.len());
+        let mut carry_in = S::ZERO;
+        for i in 0..self.sum.len() {
+            let c = ctx.narrow(&BitOpStep::from(i));
+            let (sum_bit, carry_out) = if i < value.len() {
+                one_bit_adder(c, record_id, &value[i], &self.sum[i], &carry_in).await?
+            } else {
+                one_bit_adder(c, record_id, &S::ZERO, &self.sum[i], &carry_in).await?
+            };
+
+            output_sum.push(sum_bit);
+            carry_in = carry_out;
+        }
+        let is_saturated = -carry_in
+            .clone()
+            .multiply(
+                &self.is_saturated,
+                ctx.narrow(&BitOpStep::from(self.sum.len())),
+                record_id,
+            )
+            .await?
+            + &carry_in
+            + &self.is_saturated;
+
+        Ok(SaturatingSum::new(
+            BitDecomposed::new(output_sum),
+            is_saturated,
+        ))
+    }
+}
+
+///
+/// Returns (`sum_bit`, `carry_out`)
+///
+async fn one_bit_adder<C, SB>(
+    ctx: C,
+    record_id: RecordId,
+    x: &SB,
+    y: &SB,
+    carry_in: &SB,
+) -> Result<(SB, SB), Error>
+where
+    C: Context,
+    SB: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    // compute sum bit as x XOR y XOR carry_in
+    let sum_bit = x.clone() + y + carry_in;
+
+    let x_xor_carry_in = x.clone() + carry_in;
+    let y_xor_carry_in = y.clone() + carry_in;
+    let carry_out = x_xor_carry_in
+        .multiply(&y_xor_carry_in, ctx, record_id)
+        .await?
+        + carry_in;
+
+    Ok((sum_bit, carry_out))
+}
+
+#[cfg(all(test, unit_test))]
+mod tests {
+    use super::SaturatingSum;
+    use crate::{
+        ff::{Field, Gf2},
+        protocol::{context::Context, RecordId},
+        secret_sharing::{
+            replicated::semi_honest::AdditiveShare as Replicated, BitDecomposed, SharedValue,
+        },
+        test_fixture::{get_bits, Reconstruct, Runner, TestWorld},
+    };
+
+    impl Reconstruct<u128> for [SaturatingSum<Replicated<Gf2>>; 3] {
+        fn reconstruct(&self) -> u128 {
+            let [s0, s1, s2] = self;
+
+            let sum_bits: BitDecomposed<Gf2> = BitDecomposed::new(
+                s0.sum
+                    .iter()
+                    .zip(s1.sum.iter())
+                    .zip(s2.sum.iter())
+                    .map(|((a, b), c)| [a, b, c].reconstruct()),
+            );
+
+            let is_saturated = [&s0.is_saturated, &s1.is_saturated, &s2.is_saturated].reconstruct();
+
+            if is_saturated == Gf2::ZERO {
+                sum_bits
+                    .iter()
+                    .map(Field::as_u128)
+                    .enumerate()
+                    .fold(0_u128, |acc, (i, x)| acc + (x << i))
+            } else {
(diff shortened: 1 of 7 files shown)


src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    ComputeDifferenceToCap(usize),
+    ComputedCappedAttributedTriggerValueNotSaturatedCase(usize),
+    ComputedCappedAttributedTriggerValueJustSaturatedCase(usize),
+}
+
+impl crate::protocol::step::Step for Step {}
+
+impl AsRef<str> for Step {
+    fn as_ref(&self) -> &str {
+        const EVER_ENCOUNTERED_SOURCE_EVENT: [&str; 64] = repeat64str!["eese_row"];
```
richajaindce, line 61: This boilerplate code is gone now, thanks to Taiki's recent change. I think you need to mark the step as dynamic and use that instead. Sample example: https://github.com/private-attribution/ipa/blob/main/src/protocol/sort/apply_sort/shuffle.rs#L21

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    Ok(BitDecomposed::new(output))
+}
+
+#[allow(clippy::too_many_lines, clippy::too_many_arguments)]
+async fn compute_row_with_previous<C, BK, TV>(
+    ctx: C,
+    record_id: RecordId,
+    row_for_user: usize,
+    input_row: &PrfShardedIpaInputRow<BK, TV>,
+    inputs_required_from_previous_row: &InputsRequiredFromPrevRow,
```
akoshelev, line 347: I would also take it by mutable reference and remove an extra return value  - that removes the need for the reassignment inside the callee's loop and makes it more obvious what is going on

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    num_breakdown_key_bits: usize,
```
akoshelev, line 348: taking those from the type should help with clippy warnings

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+            ctx.narrow(&Step::DidTriggerGetAttributed(row_for_user)),
+            record_id,
+        )
+        .await?;
+
+    let narrowed_ctx = ctx.narrow(&Step::AttributedTriggerValue(row_for_user));
+    let attributed_trigger_value = BitDecomposed::new(
+        try_join_all(
+            tv.iter()
+                .zip(repeat(did_trigger_get_attributed.clone()))
```
akoshelev, line 435:
```suggestion
                .zip(repeat(&did_trigger_get_attributed))
```

and then you don't need an extra `&` on the line 441

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+                        trigger_value_bit
+                            .multiply(&did_trigger_get_attributed, c, record_id)
+                            .await
+                    }
+                }),
+        )
+        .await?,
+    );
+
+    let (saturating_sum, is_saturated) = compute_saturating_sum(
```
akoshelev, line 449: saturation could follow the same pattern, mutating saturated sum flag in place

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+        row_for_user,
+    ));
+    let narrowed_ctx2 =
+        ctx.narrow(&Step::ComputedCappedAttributedTriggerValueJustSaturatedCase(row_for_user));
+    let capped_attributed_trigger_value = BitDecomposed::new(
+        try_join_all(
+            attributed_trigger_value
+                .iter()
+                .zip(inputs_required_from_previous_row.difference_to_cap.iter())
+                .zip(repeat(is_saturated.clone()))
```
akoshelev, line 490:
```suggestion
                .zip(repeat(&is_saturated)))
```

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+            one_bit_subtractor(c, record_id, &SB::ZERO, bit, &carry_in).await?;
+
+        output.push(difference_bit);
+        carry_in = carry_out;
+    }
+    Ok(BitDecomposed::new(output))
+}
+
+#[allow(clippy::too_many_lines, clippy::too_many_arguments)]
+async fn compute_row_with_previous<C, BK, TV>(
```
akoshelev, line 342: I know that we have an excel file that does this calculation, but it would be nice to put a paragraph or two, describing the algorithm at the very high level

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    };
+    let outputs_for_aggregation = CappedAttributionOutputs {
+        did_trigger_get_attributed,
+        attributed_breakdown_key_bits,
+        capped_attributed_trigger_value,
+    };
+    Ok((inputs_required_for_next_row, outputs_for_aggregation))
+}
+
+#[cfg(all(test, any(unit_test, feature = "shuttle")))]
```
akoshelev, line 534: why do you need to run these tests under the randomized scheduler? there seem to be no obvious point where we call shuttle executor here

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    repeat64str,
+    secret_sharing::{
+        replicated::{semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing},
+        BitDecomposed, Linear as LinearSecretSharing,
+    },
+};
+
+pub struct PrfShardedIpaInputRow<BK: GaloisField, TV: GaloisField> {
+    prf_of_match_key: u64,
+    is_trigger_bit: Replicated<Gf2>,
```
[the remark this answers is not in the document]
benjaminsavage: It was much trickier to get it to compile, so I decided to start with Semi Honest. Also, we will likely need to rewrite our malicious approach to use entirely new techniques - so I didn't want to spend too much time making the old approach work.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+        saturating_sum: BitDecomposed::new(vec![Replicated::ZERO; num_saturating_sum_bits]),
+        is_saturated: Replicated::ZERO,
+        // This is incorrect in the case that the CAP is less than the maximum value of "trigger value" for a single row
+        // Not a problem if you assume that's an invalid input
+        difference_to_cap: BitDecomposed::new(vec![Replicated::ZERO; num_trigger_value_bits]),
+    }
+}
+
+///
+/// Returns (`sum_bit`, `carry_out`)
```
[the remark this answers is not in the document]
benjaminsavage: Yes, and I should probably put it in a separate file with tests too.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+///
+async fn one_bit_adder<C, SB>(
+    ctx: C,
+    record_id: RecordId,
+    x: &SB,
+    y: &SB,
+    carry_in: &SB,
```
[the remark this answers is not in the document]
benjaminsavage: I guess I don't like mutable anything, so I try to avoid it when possible.

src/protocol/prf_sharding/mod.rs
benjaminsavage: Yes :). I'll do that. Also the excel file is not public - so a comment is necessary.

src/protocol/prf_sharding/mod.rs
benjaminsavage: I did that because...
...copy paste without understanding what I'm doing? 

:)

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+            one_bit_adder(c, record_id, &cur_value[i], &prev_sum[i], &carry_in).await?
+        } else {
+            one_bit_adder(c, record_id, &SB::ZERO, &prev_sum[i], &carry_in).await?
+        };
+
+        output.push(sum_bit);
+        carry_in = carry_out;
+    }
+    let updated_is_saturated = -carry_in
+        .clone()
```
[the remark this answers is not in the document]
andyleiserson: I concur with @akoshelev's suggestion to encapsulate the saturating sum. I might suggest using `protocol::boolean::or` instead of the inline logic, but I think instead it's possible to compute `is_saturated` and `just_saturated` using only one multiply:

```
just_saturated = carry & !prev_saturated
is_saturated = just_saturated + prev_saturated
```

(Where the `+` on the second line is addition, not logical OR, because they can't both be true.)

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    },
+    repeat64str,
+    secret_sharing::{
+        replicated::{semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing},
+        BitDecomposed, Linear as LinearSecretSharing,
+    },
+};
+
+pub struct PrfShardedIpaInputRow<BK: GaloisField, TV: GaloisField> {
+    prf_of_match_key: u64,
```
andyleiserson, line 22: Possibly we don't want to replicate the match key PRF with each row record -- we may not need to carry it here at all, as long as we have something like a "match key same as last row" bit.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+/// The output of this circuit is the input to the next stage: Aggregation.
+///
+/// # Errors
+/// Propagates errors from multiplications
+/// # Panics
+/// Propagates errors from multiplications
+pub async fn attribution_and_capping<C, BK, TV>(
+    sh_ctx: C,
+    input_rows: &[PrfShardedIpaInputRow<BK, TV>],
+    num_breakdown_key_bits: usize,
```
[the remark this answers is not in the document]
andyleiserson: And when we have a type for the saturating sum, something like `SS:BITS` too.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    assert!(num_saturating_sum_bits > num_trigger_value_bits);
+    assert!(num_trigger_value_bits > 0);
+    assert!(num_breakdown_key_bits > 0);
+
+    let binary_validator = sh_ctx.narrow(&Step::BinaryValidator).validator::<Gf2>();
+    // TODO: fix num total records to be not a hard-coded constant, but variable per step
+    // based on the histogram of how many users have how many records a piece
+    let binary_m_ctx = binary_validator.context().set_total_records(1);
+
+    let mut output = vec![];
```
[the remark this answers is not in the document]
andyleiserson: I think `input_rows.len()` is suitable for `set_total_records`, too.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+        attributed_breakdown_key_bits: BitDecomposed::decompose(num_breakdown_key_bits, |i| {
+            Replicated::new(
+                Gf2::truncate_from(input_row.breakdown_key.left()[i]),
+                Gf2::truncate_from(input_row.breakdown_key.right()[i]),
+            )
+        }),
+        saturating_sum: BitDecomposed::new(vec![Replicated::ZERO; num_saturating_sum_bits]),
+        is_saturated: Replicated::ZERO,
+        // This is incorrect in the case that the CAP is less than the maximum value of "trigger value" for a single row
+        // Not a problem if you assume that's an invalid input
```
andyleiserson, line 207: Seems like a reasonable restriction, but probably worth an assertion or some other form of check?

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    let mut num_users_encountered = 0;
+    let mut row_for_user = 0;
+    while i < input_rows.len() {
+        let cur_row = &input_rows[i];
+        if prev_prf == cur_row.prf_of_match_key {
+            // Do some actual computation
+            let (inputs_required_for_next_row, capped_attribution_outputs) =
+                compute_row_with_previous(
+                    binary_m_ctx.clone(),
+                    RecordId(num_users_encountered),
```
andyleiserson, line 154: Maybe I'm missing something about the structure, but I would have expected `RecordId(i)`.

Also, it may be necessary to arrange for `compute_row_with_previous` to be called for every row, even the first row for each match key, to satisfy the restrictions on contiguous record IDs for exchanged data.

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
andyleiserson: I would prefer a struct for the return type rather than a tuple or a mutable argument.

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
andyleiserson: It might be worth noting the restriction that this only works in `Gf2` to discourage anyone from changing the type parameters.

andyleiserson (commented):
As `num_saturating_sum_bits` increases, at some point the non-bitwise implementation might become cheaper than the bitwise implementation, so there might be a case for parameterizing the credit accumulation algorithm, depending on how much dynamic range we anticipate wanting for credit.

(I do think there may be some additional optimization possible in the comparison protocol when the cap is a power of two, which would change the performance crossover point.)

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
benjaminsavage: Woah! Nice! That's a slick optimization that I will definitely put in. Thanks for the careful review Andy!

src/protocol/prf_sharding/mod.rs
benjaminsavage: Makes sense to me. I think we can return to this when we implement the first part of the OPRF approach. Since we will be generating the inputs to this stage ourselves - we are free to change this to anything we want.

src/protocol/prf_sharding/mod.rs
benjaminsavage: I'm in the process of refactoring this and all of this will change to be better

benjaminsavage:
I have not yet incorporated all of the excellent feedback here, but now that I finally eliminated all of the infra hangs, and I was able to extend my test case to support a FAR more complex scenario I thought I would update it. 

Expect additional updates to incorporate all of the excellent feedback here.

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
akoshelev: it may be a suitable case for a mutable argument here, because the following code is a poor replacement for it imo

```rust
   loop {
       let mut x = 1;
       let y = do_foo(&x);
       x = y;
   }
 ```

in C++ I would also advocate for return struct because it is hard to see that argument is getting mutated, Rust handles this well by forcing you to write `&mut x` at the callee's site.

File: src/protocol/boolean/saturating_sum.rs [hunk 18]
```diff
@@ -0,0 +1,170 @@
...
+        assert_eq!(63, saturating_add(60, 6, 3, 3).await);
+        assert_eq!(64, saturating_add(60, 6, 4, 3).await);
+        assert_eq!(64, saturating_add(60, 6, 5, 3).await);
+    }
+
+    async fn saturating_add(a: u32, num_a_bits: u32, b: u32, num_b_bits: u32) -> u128 {
+        let world = TestWorld::default();
+
+        let a_bits = get_bits::<Gf2>(a, num_a_bits);
+        //let a_saturated = Gf2::ZERO;
```
martinthomson, line 153:
```suggestion
```

src/protocol/boolean/saturating_sum.rs
```diff
@@ -0,0 +1,170 @@
...
+            let is_saturated = [&s0.is_saturated, &s1.is_saturated, &s2.is_saturated].reconstruct();
+
+            if is_saturated == Gf2::ZERO {
+                sum_bits
+                    .iter()
+                    .map(Field::as_u128)
+                    .enumerate()
+                    .fold(0_u128, |acc, (i, x)| acc + (x << i))
+            } else {
+                2_u128.pow(s0.sum.len() as u32)
```
martinthomson, line 124: You should assert that all of the bits are also 1 when saturation occurs.

src/protocol/boolean/saturating_sum.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+
+#[derive(Debug)]
+pub struct SaturatingSum<S: LinearSecretSharing<Gf2>> {
+    pub sum: BitDecomposed<S>,
+    pub is_saturated: S,
+}
+
+impl<S: LinearSecretSharing<Gf2>> SaturatingSum<S> {
+    pub fn new(value: BitDecomposed<S>, is_saturated: S) -> SaturatingSum<S> {
```
martinthomson, line 15: Is there ever a case where you would construct one of these with is_saturated set to anything other than ZERO?  I ask, because it seems like you might want to avoid the case where `sum` is less than the maximum, say:

```rust
   let x = SaturatingSum::new(get_bits::<Gf2>(20, 5), ctx.share_known_value(1));
```

It's generally not a good idea to allow objects to be constructed in a way that allows invariants like this to be violated.  I can't see a concrete problem in this case, but there you go.

martinthomson (commented):
Partial review.  The easy bits first.

src/protocol/boolean/saturating_sum.rs
```diff
@@ -0,0 +1,170 @@
...
+        SaturatingSum {
+            sum: value,
+            is_saturated,
+        }
+    }
+
+    pub async fn add<C>(
```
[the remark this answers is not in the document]
benjaminsavage: Yes, I've been meaning to do that. Will add it now.

src/protocol/boolean/saturating_sum.rs
benjaminsavage: But they are not. That would be more costly. The less significant bits continue to change as you keep accumulating more numbers - but the proper interpretation is to ignore them when the "is_saturated" bit is a 1

src/protocol/boolean/saturating_sum.rs
benjaminsavage: The only case would probably be in tests, but I'm not currently using that.

As I note in my comment above, it's not an invariant violation to construct a SaturatingSum where the lower bits are random stuff and the "is_saturated" is a 1. That is just interpreted as being saturated.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    num_trigger_value_bits: usize,
+    num_saturating_sum_bits: usize,
+) -> InputsRequiredFromPrevRow
+where
+    BK: GaloisField,
+    TV: GaloisField,
+{
+    InputsRequiredFromPrevRow {
+        ever_encountered_a_source_event: share_of_one - &input_row.is_trigger_bit,
+        attributed_breakdown_key_bits: BitDecomposed::decompose(num_breakdown_key_bits, |i| {
```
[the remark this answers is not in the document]
benjaminsavage: OK, I'll do that.

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
benjaminsavage: Yeah, now that I've refactored the code into a separate SaturatingSum class, I think the mutable reference is a good solution. I will implement that.

src/protocol/prf_sharding/mod.rs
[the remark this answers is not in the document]
benjaminsavage: OK, now that I have encapsulated this in a SaturatingSum class, I no longer have access to `prev_saturated` and I'm not sure how to put this optimization in.

I suggest leaving it out for now, and saving it for a rainy day when we need a little boost in our performance =).

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+
+    let mut carry_in = SB::ZERO;
+    let mut output = vec![];
+    for i in 0..num_saturating_sum_bits {
+        let c = ctx.narrow(&BitOpStep::from(i));
+        let (sum_bit, carry_out) = if i < num_trigger_value_bits {
+            one_bit_adder(c, record_id, &cur_value[i], &prev_sum[i], &carry_in).await?
+        } else {
+            one_bit_adder(c, record_id, &SB::ZERO, &prev_sum[i], &carry_in).await?
+        };
```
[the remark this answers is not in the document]
benjaminsavage: @martinthomson had the even slicker suggestion of: 
```
let x = value.get(i).unwrap_or(&zero);
let sum_bit = one_bit_adder(c, record_id, x, &self.sum[i], &mut carry_in).await?;
```

src/protocol/prf_sharding/mod.rs
benjaminsavage: This is now encapsulated within the SaturatingSum class

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,704 @@
...
+    // based on the histogram of how many users have how many records a piece
+    let binary_m_ctx = binary_validator.context().set_total_records(1);
+
+    let mut output = vec![];
+
+    assert!(!input_rows.is_empty());
+    let first_row = &input_rows[0];
+    let mut prev_prf = first_row.prf_of_match_key;
+    let mut prev_row_inputs = initialize_new_device_attribution_variables(
+        Replicated::share_known_value(&binary_m_ctx, Gf2::ONE),
```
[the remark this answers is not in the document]
benjaminsavage: Yes, to avoid passing in the context.

src/protocol/prf_sharding/mod.rs
benjaminsavage: I've added:
```
assert!(num_saturating_sum_bits > num_trigger_value_bits);
```

So it should be impossible to encounter this error.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    num_saturating_sum_bits: usize,
+) -> Result<CappedAttributionOutputs, Error>
+where
+    C: UpgradedContext<Gf2, Share = Replicated<Gf2>>,
+    BK: GaloisField,
+    TV: GaloisField,
+{
+    let bd_key = BitDecomposed::decompose(BK::BITS, |i| {
+        input_row.breakdown_key.map(|v| Gf2::truncate_from(v[i]))
+    });
```
martinthomson, line 431: We might want to consider a function for this conversion on GfXXX types.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    let narrowed_ctx1 = ctx.narrow(&Step::ComputedCappedAttributedTriggerValueNotSaturatedCase);
+    let narrowed_ctx2 = ctx.narrow(&Step::ComputedCappedAttributedTriggerValueJustSaturatedCase);
+
+    let one = &Replicated::share_known_value(&narrowed_ctx1, Gf2::ONE);
+
+    Ok(BitDecomposed::new(
+        ctx.parallel_join(
+            attributed_trigger_value
+                .iter()
+                .zip(prev_row_diff_to_cap.iter())
```
martinthomson, line 374:
```suggestion
            zip(attributed_trigger_value.iter(), prev_row_diff_to_cap.iter())
```
I tend to prefer this formulation.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+            cur_row_breakdown_key_bits
+                .iter()
+                .zip(prev_row_breakdown_key_bits.iter())
+                .enumerate()
+                .map(|(i, (cur_bit, prev_bit))| {
+                    let c = ctx.narrow(&BitOpStep::from(i));
+                    async move {
+                        let maybe_diff = is_trigger_bit
+                            .multiply(&(prev_bit.clone() - cur_bit), c, record_id)
+                            .await?;
```
martinthomson, line 290: Isn't this prime for our if_else() protocol?

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    let mut output = Vec::with_capacity(rows_for_user.len() - 1);
+    for (i, row) in rows_for_user.iter().skip(1).enumerate() {
+        let ctx_for_this_row_depth = ctx_for_row_number[i].clone(); // no context was created for row 0
+        let record_id_for_this_row_depth = record_id_for_each_depth[i + 1]; // skip row 0
+
+        let capped_attribution_outputs = compute_row_with_previous(
+            ctx_for_this_row_depth,
+            record_id_for_this_row_depth,
+            row,
+            &mut prev_row_inputs,
```
martinthomson, line 222: This looks like a method on the mutable `prev_row_inputs` struct.

martinthomson (approved):

src/protocol/boolean/saturating_sum.rs
```diff
@@ -0,0 +1,318 @@
...
+/// "Improved Garbled Circuit Building Blocks and Applications to Auctions and Computing Minima"
+/// `https://encrypto.de/papers/KSS09.pdf`
+///
+/// Section 3.1 Integer Addition, Subtraction and Multiplication
+///
+/// For each bit, the `sum_bit` can be efficiently computed as just `s_i = x_i ⊕ y_i ⊕ c_i`
+/// This can be computed "for free" in Gf2
+///
+/// The `carry_out` bit can be efficiently computed with just a single multiplication as:
+/// `c_(i+1) = c_i ⊕ ((x_i ⊕ c_i) ∧ (y_i ⊕ c_i))`
```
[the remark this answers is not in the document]
benjaminsavage: Yeah... too subtle. I can use either `AND` or `&`... what do you like better?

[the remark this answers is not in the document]
martinthomson: ¯\\\_(ツ)\_/¯

src/protocol/boolean/saturating_sum.rs
```diff
@@ -0,0 +1,318 @@
...
+    y: &SB,
+    carry_in: &SB,
+    compute_carry_out: bool,
+) -> Result<(SB, SB), Error>
+where
+    C: Context,
+    SB: LinearSecretSharing<Gf2> + BasicProtocols<C, Gf2>,
+{
+    // compute difference bit as not_y XOR x XOR carry_in
+    let difference_bit = SB::share_known_value(&ctx, Gf2::ONE) - y + x + carry_in;
```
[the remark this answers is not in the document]
benjaminsavage: As you point out, in Gf2 it's all the same :). I'll change it to addition.

src/protocol/prf_sharding/mod.rs
benjaminsavage: That makes sense.

src/protocol/prf_sharding/mod.rs
benjaminsavage: Yes, I can use that in a few places.

src/protocol/prf_sharding/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    BK: GaloisField,
+    TV: GaloisField,
+{
+    assert!(!rows_for_user.is_empty());
+    if rows_for_user.len() == 1 {
+        return Ok(vec![]);
+    }
+    let first_row = &rows_for_user[0];
+    let mut prev_row_inputs = initialize_new_device_attribution_variables(
+        Replicated::share_known_value(&ctx_for_row_number[0], Gf2::ONE),
```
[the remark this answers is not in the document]
benjaminsavage: I think you still need to know the total size at creation time.

src/protocol/boolean/saturating_sum.rs
benjaminsavage: OK, so there is one place that I call it with ONE, and that's in the `add()` method itself. It returns a new SaturatingSum, and it might be saturated.

martinthomson: It seems like you might have separate `new` and `new_saturated` functions for that then.

PR #787, opened 2023-09-26
Allow addition and multiplication for references to secret shared values

akoshelev:
This change brings an extra trait bound for linear secret sharing trait, to allow addition and multiplication on references, rather than owned types. Currently it is only possible to do `a + b` or `a + &b`, but not `&a + b` or `&a + &b`. This change makes it possible.

As usual, it is not all rosy here, because of the compiler issue: https://github.com/rust-lang/rust/issues/20671. There is a need currently to specify the additional trait bound on every function, but I think it is worth it because not every `.clone()` call can be eliminated by the compiler, as it can be seen [here](https://rust.godbolt.org/z/MsoTnaeoc). We're better off helping optimizer here. So it cost us additional trait bound, but gives a much nicer addition/multiplication w/o extra clones and makes us not worry about compiler not being able to optimize it. 

Once (if?) that gnarly issue is fixed, we could probably just add `for <'a, 'b> &'a Self: Add<'b Self>`bound to `ArithmeticOps` trait and eliminate all extra helper traits along with extra bounds on functions. 

## Implementation notes

* I could've probably used a macro to synthesize all extra `Add`/`Sub`/`Mul` implementations, but decided it is not worth it for just two types. Maybe if we had more secret sharing types ....

* I tried hard to mimic `num` trait definition (https://github.com/rust-num/num/pull/283), but ran into an error complaining that `Add` operation is not general enough:

```
error: implementation of `Add` is not general enough
  --> src/secret_sharing/mod.rs:97:49
   |
97 |         assert_eq!(AdditiveShare::<Fp31>::ZERO, sum_ref_ref(&AdditiveShare::<Fp31>::ZERO, &AdditiveShare::ZERO));
   |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `Add` is not general enough
   |
   = note: `Add<replicated::semi_honest::additive_share::AdditiveShare<fp31::Fp31>>` would have to be implemented for the type `&'0 replicated::semi_honest::additive_share::AdditiveShare<fp31::Fp31>`, for any lifetime `'0`...
   = note: ...but `Add<replicated::semi_honest::additive_share::AdditiveShare<fp31::Fp31>>` is actually implemented for the type `&'1 replicated::semi_honest::additive_share::AdditiveShare<fp31::Fp31>`, for some specific lifetime `'1`
```

it remains a mystery to me, why I was getting it. The same function worked just fine for `u32` and primitive types had the same 4 `Add` overloads with reference ones defined exactly as ours for additive secret shares.

### DIFF
--- a/src/ff/mod.rs
+++ b/src/ff/mod.rs
@@ -63,11 +63,15 @@ impl<T, Rhs, Output> ArithmeticOps<Rhs, Output> for T where
 /// The trait for references which implement local arithmetic operations, taking the
 /// second operand either by value or by reference.
 ///
+/// The need for this trait is dictated by [`rust-issue`] that causes us not being able to constrain
+/// references to `Self`. Once this issue is fixed, we can simply get rid of it.
+///
 /// This is automatically implemented for types which implement the operators. The idea is borrowed
 /// from [`RefNum`] trait, but I couldn't really make it work with HRTB and secret shares. Primitive
 /// types worked just fine though, so it is possible that it is another compiler issue.
 ///
 /// [`RefNum`]: https://docs.rs/num/0.4.1/num/traits/trait.RefNum.html
+/// [`rust-issue`]: https://github.com/rust-lang/rust/issues/20671
 pub trait RefOps<'a, Base: 'a, R: 'a>:
     LocalArithmeticOps<Base, Base>
     + LocalArithmeticOps<&'a Base, Base>
--- a/src/ff/mod.rs
+++ b/src/ff/mod.rs
@@ -15,59 +15,77 @@ use generic_array::{ArrayLength, GenericArray};
 pub use prime_field::Fp31;
 pub use prime_field::{Fp32BitPrime, PrimeField};
 
-use crate::secret_sharing::SharedValue;
-
 #[derive(Debug, thiserror::Error, PartialEq, Eq)]
 pub enum Error {
     #[error("unknown field type {type_str}")]
     UnknownField { type_str: String },
 }
 
-pub trait ArithmeticOps:
-    Add<Output = Self>
-    + AddAssign
-    + Sub<Output = Self>
-    + SubAssign
-    + Mul<Output = Self>
-    + MulAssign
-    + Neg<Output = Self>
-    + Sized
+/// Arithmetic operations that do not require communication in our MPC setting and can be performed
+/// locally.
+pub trait LocalArithmeticOps<Rhs = Self, Output = Self>:
+Add<Rhs, Output = Output>
++ AddAssign<Rhs>
++ Sub<Rhs, Output = Output>
++ SubAssign<Rhs>
++ Neg<Output = Output>
++ Sized
 {
 }
 
-impl<T> ArithmeticOps for T where
-    T: Add<Output = Self>
-        + AddAssign
-        + Sub<Output = Self>
-        + SubAssign
-        + Mul<Output = Self>
-        + MulAssign
-        + Neg<Output = Self>
-        + Sized
+/// TODO: add docs
+/// May or may not require communication, depending on the value. Multiplying field values is a
+/// local operation, while multiplying secret shares is not.
+pub trait ArithmeticOps<Rhs = Self, Output = Self>: LocalArithmeticOps<Rhs, Output>
+// TODO: make mul a trait and implement it for secret sharing
+    + Mul<Rhs, Output = Output>
+    + MulAssign<Rhs>
 {
 }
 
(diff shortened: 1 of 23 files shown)


src/secret_sharing/replicated/semi_honest/additive_share.rs
```diff
@@ -73,20 +73,35 @@ where
...
 
-    fn add(mut self, rhs: &Self) -> Self::Output {
-        self += rhs;
-        self
+    fn add(self, rhs: Self) -> Self::Output {
+        Add::add(&self, &rhs)
+    }
+}
+
+impl<'a, V: SharedValue> Add<AdditiveShare<V>> for &'a AdditiveShare<V> {
```
martinthomson, line 92: Do you need to name the lifetime in this case?

src/ff/mod.rs
```diff
@@ -15,57 +15,76 @@ use generic_array::{ArrayLength, GenericArray};
...
+/// The need for this trait is dictated by [`rust-issue`] that causes us not being able to constrain
+/// references to `Self`. Once this issue is fixed, we can simply get rid of it.
+///
+/// This is automatically implemented for types which implement the operators. The idea is borrowed
+/// from [`RefNum`] trait, but I couldn't really make it work with HRTB and secret shares. Primitive
+/// types worked just fine though, so it is possible that it is another compiler issue.
+///
+/// [`RefNum`]: https://docs.rs/num/0.4.1/num/traits/trait.RefNum.html
+/// [`rust-issue`]: https://github.com/rust-lang/rust/issues/20671
+pub trait RefOps<'a, Base: 'a, R: 'a>:
```
martinthomson, line 75: I think that `ArithmeticOps` is the better choice of name here, with the trait above being for `ArithmeticAssignOps` or something like that.

src/ff/mod.rs
```diff
@@ -15,57 +15,76 @@ use generic_array::{ArrayLength, GenericArray};
...
-    for<'a> Add<&'a Self, Output = Self>
-    + for<'a> AddAssign<&'a Self>
-    + Neg<Output = Self>
-    + for<'a> Sub<&'a Self, Output = Self>
-    + for<'a> SubAssign<&'a Self>
-    + Mul<V, Output = Self>
+pub trait LocalAssignOps<Rhs = Self>: AddAssign<Rhs> + SubAssign<Rhs> {}
+impl<T, Rhs> LocalAssignOps<Rhs> for T where T: AddAssign<Rhs> + SubAssign<Rhs> {}
+
+/// Arithmetic operations that may or may not require communication.
```
martinthomson, line 43: Can these really involve communication?  None of these traits are async, so this can't do that.  That suggests that these are still local operations, but only those that involve value-based l-values.

martinthomson (commented):
Some clarity about naming is probably worth trying to achieve.

src/ff/mod.rs
akoshelev: I'd love to keep `Ref` somewhere in the name to indicate that it only makes sense to implement/bound if you need l-value references

martinthomson: `ArithmeticRefOps` then, I guess.

src/protocol/basics/if_else.rs
```diff
@@ -30,6 +31,6 @@ where
     //   = false_value
     Ok(false_value.clone()
```
martinthomson, line 32: Missed a clone opportunity?

src/ff/mod.rs
```diff
@@ -15,57 +15,76 @@ use generic_array::{ArrayLength, GenericArray};
...
-    + MulAssign
-    + Neg<Output = Self>
-    + Sized
+/// Arithmetic operations that do not require communication in our MPC setting and can be performed
+/// locally.
+///
+/// Note: Neg operation is also local, but is causing issues when added as a bound to this trait.
+/// The problem is that it does not use `Rhs` generic and rustc overflows trying to compile functions
+/// that use HRTB generics bounded by `LocalArithmeticOps`.
+pub trait LocalArithmeticOps<Rhs = Self, Output = Self>:
```
akoshelev, line 30: addsub?

src/protocol/basics/if_else.rs
akoshelev: oh my...

akoshelev:
@martinthomson I think it is ready for review now

martinthomson (approved):

PR #790, opened 2023-10-03
remove obsolete synthetic steps generation code from the script

taikiy:
Our current code base and with the new steps derive macro, we can remove most of the synthetic steps generation code from `collect_steps.py` script.


akoshelev (approved):
👍

PR #791, opened 2023-10-03
Feature vector label vector dot product

benjaminsavage:
Attributes trigger events to source events (cap of at most one attributed conversion per user).

Then, aggregates the "feature vector" of all source events receiving attribution across all users.

This is useful for performing Logistic Regression, as outlined here: https://github.com/patcg-individual-drafts/ipa/blob/main/logistic_regression.md


src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+        prime_field_ctx
+            .narrow(&Step::ModulusConvertFeatureVectorBits)
+            .set_total_records(outputs_chunked_by_user.len()),
+        stream_iter(outputs_chunked_by_user),
+        0..FV::BITS,
+    );
+
+    converted_feature_vector_bits
+        .try_fold(
+            vec![S::ZERO; 1 << FV::BITS],
```
akoshelev, line 271:
```suggestion
            vec![S::ZERO; usize::try_from(FV::BITS).unwrap()],
```

makes allocator's life a bit easier  :)

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+                .map(|x| RecordId(x - 1))
+                .collect(),
+            rows_for_user,
+        ));
+    }
+    let outputs_chunked_by_user = seq_try_join_all(sh_ctx.active_work(), futures)
+        .await?
+        .into_iter()
+        .flatten()
+        .collect::<Vec<BitDecomposed<Replicated<Gf2>>>>();
```
martinthomson, line 255: This collects twice, which isn't ideal.  Ideally this would use `seq_join(...).flatten()`, but that doesn't work as this is a stream of iterable objects.  You can do this to fix that:

```rust
  seq_join(sh_ctx.active_work(), futures)
    .map(|x| futures::steam::iter(x))
    .flatten()
    .collect::<Vec<_>>()
```

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+    let mut futures = Vec::with_capacity(rows_chunked_by_user.len());
+    for rows_for_user in rows_chunked_by_user {
+        for i in 0..rows_for_user.len() {
+            if i >= num_users_who_encountered_row_depth.len() {
+                num_users_who_encountered_row_depth.push(0);
+            }
+            num_users_who_encountered_row_depth[i] += 1;
+        }
+
+        futures.push(evaluate_per_user_attribution_circuit(
```
martinthomson, line 241: Rather than collect all of the futures, this should really use a stream or other generator function.    Then you aren't creating unnecessary memory pressure.

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+    let binary_validator = sh_ctx.narrow(&Step::BinaryValidator).validator::<Gf2>();
+    let binary_m_ctx = binary_validator.context();
+    let mut num_users_who_encountered_row_depth = Vec::with_capacity(histogram.len());
+    let ctx_for_row_number = set_up_contexts(&binary_m_ctx, &histogram);
+    let mut futures = Vec::with_capacity(rows_chunked_by_user.len());
+    for rows_for_user in rows_chunked_by_user {
+        for i in 0..rows_for_user.len() {
+            if i >= num_users_who_encountered_row_depth.len() {
+                num_users_who_encountered_row_depth.push(0);
+            }
```
martinthomson, line 237: Rather than adding one-by-one, you can do:

```rust
if num_users_who_encountered_row_depth.len() < rows_for_user.len() {
  num_users_who_encountered_row_depth.resize(rows_for_user.len(), 0);
}
```

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+            num_users_who_encountered_row_depth[i] += 1;
+        }
+
+        futures.push(evaluate_per_user_attribution_circuit(
+            &ctx_for_row_number,
+            num_users_who_encountered_row_depth
+                .iter()
+                .take(rows_for_user.len())
+                .map(|x| RecordId(x - 1))
+                .collect(),
```
martinthomson, line 247: If you did the incrementing *after* this invocation, you could replace this with `&num_users_who_encountered_row_depth[..rows_for_user.len()]`, which would be much better.

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,469 @@
...
+                .set_total_records(*num_users_having_that_row_number);
+            context_per_row_depth.push(ctx_for_row_number);
+        }
+    }
+    context_per_row_depth
+}
+
+fn chunk_rows_by_user<FV>(
+    input_rows: Vec<PrfShardedIpaInputRow<FV>>,
+) -> Vec<Vec<PrfShardedIpaInputRow<FV>>>
```
martinthomson, line 173: This is a bit unfortunate.  Is it possible to change this to be stream -> stream?  It's OK if the output stream returns a Vec for each user (those should be small), but this won't be nice as it is.

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: Thank you Alex! Good find =).

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: I can't. That would take an immutable borrow, but it's mutable...

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,470 @@
...
+                        _,
+                        Gf32Bit,
+                        Fp32BitPrime,
+                        Replicated<Fp32BitPrime>,
+                    >(ctx, input_rows)
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct();
```
benjaminsavage, line 466:
```
implementation of `std::marker::Send` is not general enough
`std::marker::Send` would have to be implemented for the type `&'0 [context::semi_honest::Upgraded<'_, bit_array_1::Gf2>]`, for any lifetime `'0`...
...but `std::marker::Send` is actually implemented for the type `&'1 [context::semi_honest::Upgraded<'_, bit_array_1::Gf2>]`, for some specific lifetime `'1`
```

I need help understanding what this error means.

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: I need some kind of function to operate on the stream which affects state as it goes along. Specifically I need to update the `num_users_who_encountered_row_depth`.  Should I use unfold and manually invoke `next()` on the incoming stream?

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: OK, I found a way to do it!

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: Done!

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: Thanks for telling me about this trick! I've used it.

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: Got it working!

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+        for count in record_id_for_row_depth.iter_mut().take(rows_for_user.len()) {
+            *count += 1;
+        }
+        async move { evaluate_per_user_attribution_circuit(contexts, record_ids, rows_for_user) }
+    }));
+
+    // Execute all of the async futures (sequentially), and flatten the result
+    let flattenned_stream = seq_join(sh_ctx.active_work(), stream_of_per_user_circuits)
+        .map(|x| x.unwrap().into_iter())
+        .flatten_iters();
```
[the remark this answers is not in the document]
akoshelev: ideally we should get rid of `x.unwrap()` too, but that would require writing our own stream combinator

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+    let mut record_id_for_row_depth = vec![0_u32; histogram.len()];
+    let ctx_for_row_number = set_up_contexts(&binary_m_ctx, histogram);
+
+    // Chunk the incoming stream of records into stream of vectors of records with the same PRF
+    let mut input_stream = stream_iter(input_rows);
+    let first_row = input_stream.next().await.unwrap();
+    let rows_chunked_by_user = chunk_rows_by_user(input_stream, first_row);
+
+    // Convert to a stream of async futures that represent the result of executing the per-user circuit
+    let stream_of_per_user_circuits = pin!(rows_chunked_by_user.then(|rows_for_user| {
```
akoshelev, line 259: I am trying to convince myself that this will work with multi-threading too, but I trust Rust compiler to say so....

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+        let num_user_rows = rows_for_user.len();
+        let contexts = ctx_for_row_number[..num_user_rows - 1].to_owned();
+        let record_ids = record_id_for_row_depth[..num_user_rows].to_owned();
+
+        for count in record_id_for_row_depth.iter_mut().take(rows_for_user.len()) {
```
akoshelev, line 264: can be a little shorter and shouldn't cost an extra bounds check because line 262 has the same check

```suggestion
        for count in &mut record_id_for_row_depth[..num_user_rows] {
```

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+    let first_row = &rows_for_user[0];
+    let mut prev_row_inputs = initialize_new_device_attribution_variables(first_row);
+
+    let mut output = Vec::with_capacity(rows_for_user.len() - 1);
+    // skip the first row as it requires no multiplications
+    // no context was created for the first row
+    for (i, (row, ctx)) in
+        zip(rows_for_user.iter().skip(1), ctx_for_row_number.into_iter()).enumerate()
+    {
+        let record_id_for_this_row_depth = RecordId(record_id_for_each_depth[i + 1]); // skip row 0
```
akoshelev, line 320: Constructing `RecordId` this way shouldn't be possible (we are bypassing the check that num of records < 1B). It is a bug that it is allowed - can you use `RecordId::from` here?

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+    input_stream: IS,
+    first_row: PrfShardedIpaInputRow<FV>,
+) -> impl Stream<Item = Vec<PrfShardedIpaInputRow<FV>>>
+where
+    FV: GaloisField,
+    IS: Stream<Item = PrfShardedIpaInputRow<FV>> + Unpin,
+{
+    unfold(Some((input_stream, first_row)), |state| async move {
+        state.as_ref()?;
+        let (mut s, last_row) = state.unwrap();
```
akoshelev, line 175:
```suggestion
        let (mut s, last_row) = state?;
```

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+///     - the second having 4 records
+///     - the third having 6 records
+///   Then the histogram that should be provided is:
+///     - [3, 3, 2, 2, 1, 1]
+///
+/// # Errors
+/// Propagates errors from multiplications
+/// # Panics
+/// Propagates errors from multiplications
+#[allow(clippy::async_yields_async)]
```
akoshelev, line 224: nit but I would move it down to the violation, so we don't accidentally introduce a bug inside this function.

```rust
        #[allow(clippy::async_yields_async)] // this is ok, because seq join wants a stream of futures
        async move { evaluate_per_user_attribution_circuit(contexts, record_ids, rows_for_user) }
```

src/protocol/prf_sharding/feature_label_dot_product.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    Ok(output)
+}
+
+///
+/// Upon encountering the first row of data from a new user (as distinguished by a different OPRF of the match key)
+/// this function encapsulates the variables that must be initialized. No communication is required for this first row.
+///
+fn initialize_new_device_attribution_variables<FV>(
```
akoshelev, line 336: this is now just a `impl From<&PrfShardedIpaInputRow> for InputsRequiredFromPrevRow` maybe?

Cargo.toml
```diff
@@ -95,6 +95,7 @@ tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 typenum = "1.16"
 # hpke is pinned to it
 x25519-dalek = "2.0.0-pre.0"
+stream-flatten-iters = "0.2.0"
```
akoshelev, line 98:
```suggestion
```

see below

akoshelev (approved):
I would get rid of that extra crate and there are a couple of nits that may be good to take a look at. 

This is good and maybe this infra overhead + allocations for each is not that bad. Eventually we may want to fix it, but if number of rows per user is reasonable, we can live with it for quite a while, especially with jemalloc.

src/protocol/prf_sharding/feature_label_dot_product.rs
benjaminsavage: I can't figure out how to do this.

PR #792, opened 2023-10-06
Fix some (all?) concurrency issues inside infrastructure

akoshelev:
So far I've discovered three:

1) Contention when creating a sender
2) Contention when creating a receiver

Those were easy to fix, just use the entry API that provides atomic `check-and-insert`. It costs us a clone but correctness is more important.

3) Race inside `OrderingSender`

This was not that easy to find and fix. Essentially what was happening is two threads (write to `i` and write to `i+1`) both read the same value from `next` atomic ( = `i`). Then, thread `i` won the race and blocked `waiting` shard trying to wake the waker for `i+1`. Obviously there was no waker yet, so it released the lock to let `i+1` thread put its waker there. After that, there was no action that could lead to calling that waker and the whole program stalled. 

I considered a brute force fix (to extend the critical section), but it seems that there was a more efficient way to deal with it. It works similarly to MVCC and versioning checks. We keep a version (last number seen) on each shard that gets updated each time someone tries to wake `i`. This version is a monotonically increasing number. 

Whenever a thread attempts to add its waker, it must supply its own view of the `next` atomic value. If it is lower than the value this shard has seen so far, this operation will get rejected and thread will read the `next` value again and retry. In this case, it will discover that `next` actually moved and it does not need to store a waker anymore.

I tested it with both shuttle (discovered that bug) and flaky tester - both pass.

### DIFF
--- a/src/helpers/buffers/ordering_sender.rs
+++ b/src/helpers/buffers/ordering_sender.rs
@@ -403,8 +403,11 @@ mod test {
         stream::StreamExt,
         FutureExt,
     };
+    use futures_util::future::try_join_all;
     use generic_array::GenericArray;
     use rand::Rng;
+    #[cfg(feature = "shuttle")]
+    use shuttle::future as tokio;
     use typenum::Unsigned;
 
     use super::OrderingSender;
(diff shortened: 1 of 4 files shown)


akoshelev:
failures are due to Rust update

martinthomson (approved):
I assume that the use of the entry API is to manage the concurrency risks associated with separate calls to get() and insert().  Is this something specific to dashmap?

If that is the case, can you add a comment about that to both instances.

src/helpers/buffers/ordering_sender.rs
```diff
@@ -124,37 +123,87 @@ struct WakerItem {
 }
 
 /// A collection of saved wakers.
-#[derive(Default)]
+#[derive(Default, Debug)]
 struct WaitingShard {
+    /// The maximum index that was used to wake a task that belongs to this shard.
+    /// Updates to this shard will be rejected if th supplied index is less than this value.
```
benjaminsavage, line 129: %s/th/the/

akoshelev:
I am merging it because the failure on CI is unrelated: https://github.com/private-attribution/ipa/actions/runs/6488382926/job/17627926247

If it persists, we would need to either split the additional step run into multiple or figure out what takes the most space on the drive

PR #794, opened 2023-10-06
Robust + Move breakdown to bucket generic to take any size even if not a power of 2 

richajaindce:
adding tests for aggregation + move bucket


src/protocol/prf_sharding/mod.rs
```diff
@@ -669,6 +670,7 @@ async fn move_single_value_to_bucket<BK, C, S, F>(
     record_id: RecordId,
     bd_key: BitDecomposed<S>,
     value: S,
+    breakdown_count: Option<usize>,
```
benjaminsavage, line 673: I don't like this type signature. I'd rather just pass in a usize and have the function check if it's a power of two or not.

src/protocol/prf_sharding/mod.rs
```diff
@@ -677,27 +679,33 @@ where
...
+    };
+    assert!(total_breakdowns <= 1 << BK::BITS);
+    let mut row_contribution = vec![value; total_breakdowns];
 
     for (tree_depth, bit_of_bdkey) in bd_key.iter().rev().enumerate() {
         let depth_c = ctx.narrow(&BinaryTreeDepthStep::from(tree_depth));
         let span = step >> 1;
-        let mut futures = Vec::with_capacity((1 << BK::BITS) / step);
-        for i in (0..1 << BK::BITS).step_by(step) {
+        let mut futures = Vec::with_capacity(bd_key.len() / step);
```
benjaminsavage, line 692: This doesn't seem right. 
Consider the loop when `step = 4`;
`bd_key.len() / step` is something like `5 / 4`, which is ~= 1. Previously we initialized the number of futures to something like `2^5 / 4`, which is 8.

It works because the vector resizes, but the idea is to avoid that.

src/protocol/prf_sharding/mod.rs
```diff
@@ -677,27 +679,33 @@ where
...
         let contributions = ctx.parallel_join(futures).await?;
 
         for (index, bdbit_contribution) in contributions.into_iter().enumerate() {
             let left_index = index * step;
-            let right_index = left_index + span;
-
-            row_contribution[left_index] -= &bdbit_contribution;
-            row_contribution[right_index] = bdbit_contribution;
+            if left_index + span < total_breakdowns {
+                let right_index = left_index + span;
```
benjaminsavage, line 705: We do this twice. I'd recommend a local var.

```
let right_index = left_index + span;
if right_index < total_breakdowns {
```

src/protocol/prf_sharding/mod.rs
```diff
@@ -677,27 +679,33 @@ where
...
+                row_contribution[left_index] -= &bdbit_contribution;
```
benjaminsavage, line 706: Shouldn't this be moved outside of the `if` statement? As in, shouldn't this always be done?

benjaminsavage: I'm almost certain this is a bug. If the tests didn't catch this, we should add a test that would fail here until this is resolved.

src/protocol/prf_sharding/mod.rs
```diff
@@ -860,17 +913,17 @@ pub mod tests {
...
+                test_output::<Gf5Bit, Gf3Bit>(20, 0),
+                test_output::<Gf5Bit, Gf3Bit>(20, 3),
+                test_output::<Gf5Bit, Gf3Bit>(12, 5),
+                test_output::<Gf5Bit, Gf3Bit>(20, 7),
+                test_output::<Gf5Bit, Gf3Bit>(18, 0),
+                test_output::<Gf5Bit, Gf3Bit>(12, 0),
+                test_output::<Gf5Bit, Gf3Bit>(12, 7),
+                test_output::<Gf5Bit, Gf3Bit>(12, 7),
+                test_output::<Gf5Bit, Gf3Bit>(12, 7),
+                test_output::<Gf5Bit, Gf3Bit>(12, 4),
```
benjaminsavage, line 926: This doesn't seem like an improvement...

src/protocol/prf_sharding/mod.rs
```diff
@@ -742,20 +759,30 @@ pub mod tests {
...
-            attributed_breakdown_key,
-            capped_attributed_trigger_value,
+            attributed_breakdown_key: get_bits::<Gf2>(
+                attributed_breakdown_key.try_into().unwrap(),
+                BK::BITS,
+            ),
+            capped_attributed_trigger_value: get_bits::<Gf2>(
+                capped_attributed_trigger_value.try_into().unwrap(),
+                TV::BITS,
+            ),
```
benjaminsavage, line 778: I don't understand why this was necessary. I'd rather not make this change if we can help it.

src/protocol/prf_sharding/mod.rs
```diff
@@ -799,6 +826,40 @@ pub mod tests {
         }
     }
 
+    impl IntoShares<CappedAttributionOutputs> for PreAggregationTestOutput {
+        fn share_with<R: Rng>(self, rng: &mut R) -> [CappedAttributionOutputs; 3] {
```
benjaminsavage, line 830: Why is this now required?

src/protocol/prf_sharding/mod.rs
```diff
@@ -940,4 +993,119 @@ pub mod tests {
...
+            let breakdown_key = rng.gen_range(0..32);
+
+            let value = Fp32BitPrime::truncate_from(10_u128);
+            let mut expected = [Fp32BitPrime::truncate_from(0_u128); 32];
+            expected[breakdown_key] = value;
+
+            let result: Vec<_> = world
+                .semi_honest(
+                    (
+                        get_bits::<Fp32BitPrime>(breakdown_key.try_into().unwrap(), Gf5Bit::BITS),
```
benjaminsavage, line 1050: This is very hard to read. Can you make this easier to understand?

src/protocol/prf_sharding/mod.rs
```diff
@@ -940,4 +993,119 @@ pub mod tests {
...
+                )
+                .await
+                .reconstruct();
+            assert_eq!(result, &expected);
+        });
+    }
+
+    #[test]
+    fn semi_honest_move_value_to_single_bucket_not_power_two() {
+        const BREAKDOWN_COUNT: usize = 50;
```
benjaminsavage, line 1075: There is too much duplication between `semi_honest_move_value_to_single_bucket` and `semi_honest_move_value_to_single_bucket_not_power_two`.

I recommend just merging them into a single test that loops many times, trying many combinations of "num_breakdowns" (some of which are powers of two, and some of which are not), and value.

We do not have good test coverage as it is, since you're only testing 50, and only for a single value.

benjaminsavage (changes requested):
Left some comments, mainly I'm not content with the tests.

src/protocol/prf_sharding/mod.rs
richajaindce: my bad, this got added unnecessarily, fixing it

src/protocol/prf_sharding/mod.rs
richajaindce: Changed this because do_aggregation takes breakdown keys and trigger values in BitDecomposed format. For that we need to know BK and TV to get those number of bits. Earlier we were just reconstructing u128 from bits which didnt need this transformation

src/protocol/prf_sharding/mod.rs
richajaindce: Needed for do_aggregation and move_bucket test which splits BitDecomposed<Gf2> into its shares

src/protocol/prf_sharding/mod.rs
```diff
@@ -677,16 +685,18 @@ where
...
+    let mut row_contribution = vec![value; breakdown_count];
 
     for (tree_depth, bit_of_bdkey) in bd_key.iter().rev().enumerate() {
         let depth_c = ctx.narrow(&BinaryTreeDepthStep::from(tree_depth));
         let span = step >> 1;
-        let mut futures = Vec::with_capacity((1 << BK::BITS) / step);
-        for i in (0..1 << BK::BITS).step_by(step) {
-            let bit_c = depth_c.narrow(&BitOpStep::from(i));
+        let mut futures = Vec::with_capacity(breakdown_count / step);
+        for (step_no, i) in (0..breakdown_count).step_by(step).enumerate() {
```
benjaminsavage, line 696: `for (i, tree_index) in (0..breakdown_count).step_by(step).enumerate() {`

src/protocol/prf_sharding/mod.rs
```diff
@@ -695,7 +705,6 @@ where
         for (index, bdbit_contribution) in contributions.into_iter().enumerate() {
             let left_index = index * step;
             let right_index = left_index + span;
-
             row_contribution[left_index] -= &bdbit_contribution;
             row_contribution[right_index] = bdbit_contribution;
```
benjaminsavage, line 709: I'm confident that this will fail in certain cases.

src/protocol/prf_sharding/mod.rs
```diff
@@ -742,20 +760,30 @@ pub mod tests {
         }
     }
 
-    fn test_output(
+    fn test_output<BK, TV>(
```
benjaminsavage, line 763: This can be a new function called: `bitwise_bd_key_and_value`

The old thing can remain as it was just get renamed to: `decimal_bd_key_and_value`

src/protocol/prf_sharding/mod.rs
```diff
@@ -820,16 +882,8 @@ pub mod tests {
...
-                attributed_breakdown_key: attributed_breakdown_key_bits
-                    .iter()
-                    .map(Field::as_u128)
-                    .enumerate()
-                    .fold(0_u128, |acc, (i, x)| acc + (x << i)),
-                capped_attributed_trigger_value: capped_attributed_trigger_value_bits
-                    .iter()
-                    .map(Field::as_u128)
-                    .enumerate()
-                    .fold(0_u128, |acc, (i, x)| acc + (x << i)),
```
benjaminsavage, line 832: This code can remain (within the `decimal_xxx` variant)

src/protocol/prf_sharding/mod.rs
benjaminsavage: This can become:
```
binary_bd_key_and_value(17, 7),
...
```

src/protocol/prf_sharding/mod.rs
```diff
@@ -940,4 +994,83 @@ pub mod tests {
...
         });
     }
+
+    #[test]
+    fn semi_honest_aggregation() {
+        run(|| async move {
+            let world = TestWorld::default();
+
+            let records: Vec<PreAggregationTestOutput> = vec![
+                test_output::<Gf5Bit, Gf3Bit>(17, 7),
```
benjaminsavage, line 1004: This can become `binary_bd_key_and_value`

benjaminsavage (approved):
I love it. Ship it.

src/protocol/prf_sharding/mod.rs
```diff
@@ -677,17 +692,20 @@ where
     F: PrimeField + ExtendableField,
 {
     let mut step: usize = 1 << BK::BITS;
-    let mut row_contribution = vec![value; 1 << BK::BITS];
+
+    assert!(breakdown_count <= 1 << BK::BITS);
+    let mut row_contribution = vec![value; breakdown_count];
```
martinthomson, line 697: It *might* be more efficient to initialize the vector with zero values instead, but that means putting "value" at index 0 as an extra step.

src/protocol/prf_sharding/mod.rs
```diff
@@ -940,4 +1028,131 @@ pub mod tests {
...
+                assert_eq!(result, expected);
+            });
+        }
+    }
+
+    #[test]
+    fn semi_honest_move_value_to_single_bucket_out_of_range() {
+        const MAX_BREAKDOWN_COUNT: usize = 127;
+        for robust_for_breakdown_key_gt_count in [true, false] {
+            for _ in 1..10 {
```
martinthomson, line 1116: We should not be running multiple iterations of a test like this.

PR #795, opened 2023-10-06
Prf eval

danielmasny:
implementation of DY evaluation protocol based on replicated secret sharing


src/ff/curve_points.rs
```diff
@@ -0,0 +1,282 @@
...
+
+impl Block for CompressedRistretto {
+    type Size = U32;
+}
+
+///ristretto point for curve 25519
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct RP25519(<Self as SharedValue>::Storage);
+
+//how to add const ONE: Self = Self(Scalar::ONE); within the struct, do I need to do it via a trait?
```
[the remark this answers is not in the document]
danielmasny: Alex resolved the comment. regarding the zero and one point, the one point is just a generator on the curve, i.e. the base point. To the best of my knowledge, the zero point is actually the point at infinity. The dalek library hasn't defined a constant for the zero point ideally we also shouldn't define it. I don't think we need it, but shared value seems to need it. (I guess to be able to generate a share of zero, or zerorize certain shares?)

src/ff/curve_points.rs
```diff
@@ -0,0 +1,282 @@
...
+        RP25519(RistrettoPoint::from_uniform_bytes(&scalar_bytes).compress())
+    }
+}
+
+
+impl std::ops::Add for RP25519 {
+type Output = Self;
+
+fn add(self, rhs: Self) -> Self::Output {
+    Self((self.0.decompress().unwrap()+rhs.0.decompress().unwrap()).compress())
```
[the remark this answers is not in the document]
danielmasny: Probably need to look more detailed into the tradeoff between memory cost vs cpu. I think decompressed points take 3 times more memory (I haven't checked how costly the compression is). Currently we do not do much arithmetics on curve points, just one scalar multiplication and one reveal which reconstructs the shares using two additions per party.

src/ff/curve_points.rs
```diff
@@ -0,0 +1,282 @@
...
+fn sub_assign(&mut self, rhs: Self) {
+    *self = *self - rhs;
+}
+}
+
+
+///Scalar Multiplication
+///<'a, 'b> std::ops::Mul<&'b Fp25519> for &'a
+impl RP25519 {
+    pub const ONE: Self = Self(constants::RISTRETTO_BASEPOINT_COMPRESSED);
```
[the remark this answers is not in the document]
danielmasny: Yes, I am also not happy. Ideally we would not define a zero point (which is the point at infinity on the curve). We also don't need it for shares of curve points. see comment above

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+///Scalar Multiplication
+///<'a, 'b> `std::ops::Mul<&'b"` Fp25519 for &'a
+impl RP25519 {
+
+    /// # Errors
+    /// Propagates errors from decompressing invalid curve point
+    pub fn s_mul(self, rhs: Fp25519) -> Result<RP25519, Error> {
```
akoshelev, line 98: I would probably go with `impl <'a, 'b> Mul<Rhs = &'b Fp25519> for &'a RP25519`, given in how many places we panic on failing to decompress

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+        self.0
+            .decompress()
+            .map_or(Err(Error::DecompressingInvalidCurvePoint), |x| {
+                Ok((x * Scalar::from(rhs)).compress().into())
+            })
+    }
+}
+
+///do not use
+impl std::ops::Mul for RP25519 {
```
akoshelev, line 108: Should we split `SharedValue` trait into two to support cases where shared values are multiplicative and where they are not? We can't get malicious security with those anyway

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+    };
+}
+
+cp_hash_impl!(u64, 8);
+
+cp_hash_impl!(u32, 4);
+
+/// Daniel had to implement this since Reveal wants it, prefer not to, I dont understand why it is
+/// actually needed there, maybe to upgrade it to malicious? but it still shouldn't be needed
+impl Field for RP25519 {
```
akoshelev, line 177: I think we don't need it anymore after #796

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+///do not use
+impl std::ops::MulAssign for RP25519 {
+    fn mul_assign(&mut self, _rhs: RP25519) {
+        panic!("Two curve points cannot be multiplied! Do not use *, *= for RP25519 or secret shares of RP25519");
+    }
+}
+
+impl From<Scalar> for RP25519 {
```
akoshelev, line 123:
```suggestion
impl From<&Scalar> for RP25519 {
```

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn from(s: Scalar) -> Self {
+        RP25519(RistrettoPoint::mul_base(&s).compress())
+    }
+}
+
+impl From<Fp25519> for RP25519 {
```
akoshelev, line 129:
```suggestion
impl From<&Fp25519> for RP25519 {
```

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+impl From<RP25519> for CompressedRistretto {
+    fn from(s: RP25519) -> Self {
+        s.0
+    }
+}
+
+macro_rules! cp_hash_impl {
+    ( $u_type:ty, $byte_size:literal) => {
```
akoshelev, line 148: you can get the size using `std::mem::size_of::<$u_type>()` or even simpler 

```rust
                let mut okm = <$u_type>::MIN.to_le_bytes();
```

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+                hk.expand(&[], &mut okm).unwrap();
+                RP25519::deserialize(&okm.into())
+            }
+        }
+    };
+}
+
+cp_hash_impl!(u64, 8);
+
+cp_hash_impl!(u32, 4);
```
akoshelev, line 173: do you want to use it in tests only?

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+            fn from(s: RP25519) -> Self {
+                let hk = Hkdf::<Sha256>::new(None, s.0.as_bytes());
+                let mut okm = [0u8; $byte_size];
+                //error invalid length from expand only happens when okm is very large
+                hk.expand(&[], &mut okm).unwrap();
+                <$u_type>::from_le_bytes(okm)
+            }
+        }
+
+        impl From<$u_type> for RP25519 {
```
akoshelev, line 159: do you need this?

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn truncate_from<T: Into<u128>>(v: T) -> Self {
+        let hk = Hkdf::<Sha256>::new(None, &v.into().to_le_bytes());
+        let mut okm = [0u8; 32];
+        //error invalid length from expand only happens when okm is very large
+        hk.expand(&[], &mut okm).unwrap();
+        RP25519::deserialize(&okm.into())
+    }
+}
+
+impl TryFrom<u128> for RP25519 {
```
akoshelev, line 200: this does not seem to be fallible, `From` should be sufficient here

src/ff/curve_points.rs
```diff
@@ -0,0 +1,272 @@
...
+    type Error = crate::error::Error;
+
+    fn try_from(v: u128) -> Result<Self, Self::Error> {
+        let mut bits = [0u8; 32];
+        bits[..].copy_from_slice(&v.to_le_bytes());
+        let f: RP25519 = RP25519::ONE;
+        f.serialize((&mut bits).into());
```
akoshelev, line 207: what is going on here? aren't we returning `ONE` no matter what?

akoshelev (commented):
good start, I haven't looked at everything, can make another pass later

src/ff/curve_points.rs
danielmasny: need them to convert curve points into pseudonyms, i.e. u64 (maybe u128 is better since there are less collisions)

src/ff/curve_points.rs
danielmasny: was needed by field trait which we dont need anymore

src/ff/curve_points.rs
akoshelev: I am asking because I saw usages of `u64` but `u32` seems to be only used in tests. If that's true, it is better to move this implementation behind `test` feature

src/ff/ec_prime_field.rs
```diff
@@ -0,0 +1,262 @@
...
+    type Size = U32;
+}
+
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct Fp25519(<Self as SharedValue>::Storage);
+
+impl Fp25519 {
+    pub const ONE: Self = Self(Scalar::ONE);
+
+    //must not use with ZERO
```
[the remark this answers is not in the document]
danielmasny: I have added an assert. It will be possible to generate zero values, however when not altering the code, it will only happen with negligible probability for input values for invert. Comparing two Fp25519 is cheap since its basically a [u8;32]

Cargo.toml
```diff
@@ -55,6 +55,7 @@ clap = { version = "4.3.2", optional = true, features = ["derive"] }
 comfy-table = { version = "7.0", optional = true }
 config = "0.13.2"
 criterion = { version = "0.5.1", optional = true, default-features = false, features = ["async_tokio", "plotters", "html_reports"] }
+curve25519-dalek = "4.1.1"
```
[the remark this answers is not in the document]
danielmasny: It seems that curve25519_dalek is better maintained than dalek_ng.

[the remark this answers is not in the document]
akoshelev: @martinthomson https://github.com/zkcrypto/curve25519-dalek-ng/ seems to be long forgotten, is this the one that you meant to use?

src/ff/curve_points.rs
```diff
@@ -0,0 +1,222 @@
...
+
+impl Serializable for RP25519 {
+    type Size = <<RP25519 as WeakSharedValue>::Storage as Block>::Size;
+
+    fn serialize(&self, buf: &mut GenericArray<u8, Self::Size>) {
+        *buf.as_mut() = self.0.to_bytes();
+    }
+
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Self {
+        RP25519(CompressedRistretto((*buf).into()))
```
akoshelev, line 38: can we panic here instead of panicking on decompress?

Cargo.toml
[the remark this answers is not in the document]
martinthomson: OK, that was based on my assessment the last time I looked.  It looks like the differences have been reconciled now.

src/ff/curve_points.rs
akoshelev: One reason why it may be useful to do is that ensures that data that enters our system is valid at the point of creation. Practically speaking it is much harder to debug if this system panics at the negation or multiplication. Validating assumptions at construction is generally considered good practice.

However the cost of this check is significant as you pointed out - decompressing a point is quite involved, so we could start with `debug_assert!(point.decompress().is_some())`

src/protocol/ipa_prf/prf_eval/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+) -> Result<Vec<u64>, Error>
+where
+    C: Context,
+{
+    let ctx = sh_ctx.set_total_records(input_match_keys.len());
+    let futures = input_match_keys
+        .iter()
+        .enumerate()
+        .map(|(i, x)| eval_dy_prf(ctx.clone(), i.into(), &prf_key, x));
+    seq_try_join_all(sh_ctx.active_work(), futures).await
```
akoshelev, line 40:
```suggestion
    ctx.try_join(futures).await
```

src/protocol/ipa_prf/prf_eval/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// generates PRF key k as secret sharing over Fp25519
+pub fn gen_prf_key<C>(ctx: &C) -> AdditiveShare<Fp25519>
+where
+    C: Context,
+{
+    ctx.narrow(&Step::PRFKeyGen)
+        .prss()
+        .generate_replicated(u128::MAX - 100u128)
```
akoshelev, line 56: this is very a suspicious code - I hope it is not used anywhere and can be deleted

src/ff/curve_points.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn from(s: RP25519) -> Self {
+        s.0
+    }
+}
+
+macro_rules! cp_hash_impl {
+    ( $u_type:ty) => {
+        impl From<RP25519> for $u_type {
+            fn from(s: RP25519) -> Self {
+                let hk = Hkdf::<Sha256>::new(None, s.0.as_bytes());
```
akoshelev, line 139: when using macros, it is better to import the types inside it, so you can freely use it in different modules if you need to. Here, I suggest you move an implementation for `u32` inside the `test` module and do this inside this macro


```rust
 impl From<RP25519> for $u_type {
            fn from(s: RP25519) -> Self {
                use hkdf::Hkdf;
                use sha2::Sha256;

                let hk = Hkdf::<Sha256>::new(None, s.0.as_bytes());
                ...
```

src/ff/ec_prime_field.rs
```diff
@@ -0,0 +1,255 @@
...
+    };
+}
+
+#[cfg(test)]
+sc_hash_impl!(u64);
+
+#[cfg(test)]
+sc_hash_impl!(u32);
+
+/// Daniel had to implement this since PRSS wants it, prefer not to
```
akoshelev, line 153: can you open a quick issue to our repo and reference it here so we don't forget about it?

src/secret_sharing/mod.rs
```diff
@@ -48,6 +51,16 @@ pub trait Block: Sized + Copy + Debug {
     type Size: ArrayLength<u8>;
 }
 
+pub trait WeakSharedValue:
```
akoshelev, line 54: just to debate it a bit - what would be a good name for this type? does it look like a group element under addition?

akoshelev (commented):
this is looking pretty good, just some cleanup and docs would be great to do before merging

src/secret_sharing/mod.rs
danielmasny: dont have a strong opinion, we have SharedValue that also implements mult.. For a group you technically need that you have inverse elements (-) and a neutral element (0). It seems that this is the case for all our use cases. However, would we call SharedValue then a ring or field element?

akoshelev: well, neutral element is here (ZERO). Negate is here too, SharedValue could be ring and field can get `inv` if we need to :) I can help if we all agree that this is what needs to be done - I don't think we should make it part of this PR though.

akoshelev (approved):
lgtm, I'll give it a day to let others take a look too and then merge it if there are no objections

src/ff/curve_points.rs
```diff
@@ -0,0 +1,239 @@
...
+cp_hash_impl!(u64);
+
+/// implementing random curve point generation for testing purposes,
+/// in the actual IPA protocol, we generate them from scalars, i.e. Fp25519
+#[cfg(test)]
+impl rand::distributions::Distribution<RP25519> for rand::distributions::Standard {
+    fn sample<R: crate::rand::Rng + ?Sized>(&self, rng: &mut R) -> RP25519 {
+        let mut scalar_bytes = [0u8; 64];
+        rng.fill_bytes(&mut scalar_bytes);
+        RP25519(RistrettoPoint::from_uniform_bytes(&scalar_bytes).compress())
```
martinthomson, line 172: I would still prefer that we not compress unless transmitting data.  The extra size is negligible and the uncompressed form is more efficient to use.

danielmasny: I could implement the traits we need for a decompressed version of curve points. I would probably still need to define arithmetic operations on top of compressed points to satisfy the traits needed for secret shares & sending messages. 

I think we should definitely store compressed points in memory to not waste any memory since it is the main bottleneck for running larger queries. I dont think operating on decompressed points would significantly reduce the amount of compress and decompress operations since currently I am only generate random exponents and then generate a curve point from these exponents. After that, there is one reveal operation which requires three additions  over curvepoints (in the clear, not secret shared!) and then another exponentiation on top of the revealed curve point. So there are two unnecessary compress decompress operations in total per record (plus 2 decompress when you don't store your own shares in decompressed format, the compress you still need to do since you need to send your shares ).

danielmasny:
closed, see #833 (same functionality, cleaned up commit history)

PR #796, opened 2023-10-07
Relax semi-honest reveal bounds

akoshelev:
It shouldn't require fields, semi-honest reveal works with shared values just fine.

noticed in #795

### DIFF
--- a/src/protocol/basics/reveal.rs
+++ b/src/protocol/basics/reveal.rs
@@ -18,7 +18,7 @@ use crate::{
             malicious::{AdditiveShare as MaliciousReplicated, ExtendableField},
             semi_honest::AdditiveShare as Replicated,
         },
-        SecretSharing,
+        SecretSharing, SharedValue,
     },
 };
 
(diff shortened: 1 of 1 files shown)


akoshelev:
@danielmasny

andyleiserson (approved):

danielmasny:
we need the malicious reveal for malicious security. The current protocol is targeted towards achieving malicious security in the sense that it only needs a maliciously secure multiplication (and reveal). If we just care about semi honest security, there is a more efficient protocol. There seem to be some changes that we could make to the reveal protocol such that semi honest and malicious has basically the same cost. Then we would not need to distinguish between them and just always use the malicious reveal.

akoshelev:
I believe our OPRF effort is currently focused around semi-honest security (at least attribution and aggregation steps). It is definitely possible to fix our type system to support both, but for now it could be cleaner if OPRF types that we add did not have to implement traits just for the purpose of conforming to the interface.

It is also possible that the malicious security base types we use will no longer be in place for OPRF approach

PR #799, opened 2023-10-12
#[dynamic] attribute takes an integer literal to specify the max number of steps

taikiy:
Unblocking @richajaindce for her sparse attribution-aggregation code testing. Dynamic steps can be extended beyond 64. I set the upper limit to 1024 to prevent the IDE from freezing by accidentally writing a huge number and the macro generating gigantic code.

### DIFF
--- a/ipa-macros/src/derive_step/mod.rs
+++ b/ipa-macros/src/derive_step/mod.rs
@@ -47,6 +47,8 @@ use crate::{
     tree::Node,
 };
 
+const MAX_DYNAMIC_STEPS: usize = 1024;
+
 trait CaseStyle {
     fn to_snake_case(&self) -> String;
 }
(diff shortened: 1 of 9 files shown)


martinthomson (approved):

ipa-macros/src/derive_step/mod.rs
```diff
@@ -112,20 +114,30 @@ pub fn expand(input: TokenStream) -> TokenStream {
...
         let ident_snake_case = ident.to_string().to_snake_case();
         let ident_upper_case = ident_snake_case.to_uppercase();
 
         if is_dynamic_step(v) {
-            // create an array of 64 strings and use the variant index as array index
-            let steps = (0..64)
+            let num_steps = match get_dynamic_step_count(v) {
+                Ok(n) => n,
+                Err(e) => {
+                    // we can't return from a closure, so we need to set the result and break
```
akoshelev, line 128: I'd probably convert to `for` loop here to avoid this quirkiness

akoshelev (approved):

ipa-macros/src/derive_step/mod.rs
taikiy: done!

PR #802, opened 2023-10-13
Simplify FromRequest::from_request for QueryConfigParams

cryo28:
It seems that the code introduces a few unnecessary local structs to parse http query parameters into. But these structs are the same as the final QueryType-nested structs.

So, we can parse the params into those structs directly and make the code simpler.


akoshelev (approved):
love the simplification here! 👍

akoshelev:
#801 should fix the failure, merging

PR #804, opened 2023-10-16
Restructure the bucketing code

### LINKED ISSUE #803: Empty range in bucketing tests
```
---- protocol::prf_sharding::tests::semi_honest_move_value_to_single_bucket_in_range stdout ----
thread 'protocol::prf_sharding::tests::semi_honest_move_value_to_single_bucket_in_range' panicked at /home/martin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/rng.rs:134:9:
cannot sample empty range
[...]
   4: ipa::protocol::prf_sharding::tests::semi_honest_move_value_to_single_bucket_in_range::{{closure}}::{{closure}}
             at ./src/protocol/prf_sharding/mod.rs:1082:37
```

The range is 1..1, see:

https://github.com/private-attribution/ipa/blob/d8e195de8b9da91dda1305330e9daafd3e33372d/src/protocol/prf_sharding/mod.rs#L1081-L1082

martinthomson:
Move the function to a file on its own.
Skip iterations when the number of buckets is less than half the number of bits and `robust` is false.
Refactor tests.
Add an overflow test and guard against having more than 128 buckets (which we can't support).
Fix gen_range() invocations.

Closes #803.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,7 @@ cfg_aliases = "0.1.1"
 command-fds = "0.2.2"
 hex = "0.4"
 permutation = "0.4.1"
-proptest = "1.3"
+proptest = "1"
 tempfile = "3"
 tokio-rustls = { version = "0.24.0", features = ["dangerous_configuration"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,7 @@ cfg_aliases = "0.1.1"
 command-fds = "0.2.2"
 hex = "0.4"
 permutation = "0.4.1"
-proptest = "1.0.0"
+proptest = "1.3"
 tempfile = "3"
 tokio-rustls = { version = "0.24.0", features = ["dangerous_configuration"] }
 
--- a/src/protocol/prf_sharding/bucket.rs
+++ b/src/protocol/prf_sharding/bucket.rs
@@ -1,0 +1,234 @@
+use embed_doc_image::embed_doc_image;
+
+use crate::{
+    error::Error,
+    ff::{GaloisField, PrimeField, Serializable},
+    protocol::{
+        basics::SecureMul, context::UpgradedContext, prf_sharding::BinaryTreeDepthStep,
+        step::BitOpStep, RecordId,
+    },
+    secret_sharing::{
+        replicated::malicious::ExtendableField, BitDecomposed, Linear as LinearSecretSharing,
+    },
+};
+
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -106,7 +106,7 @@ cfg_aliases = "0.1.1"
 command-fds = "0.2.2"
 hex = "0.4"
 permutation = "0.4.1"
-proptest = "1.0.0"
+proptest = "1.3"
```
martinthomson, line 109: I wasn't able to get a clean result without this.  I'll try reverting this.

src/protocol/prf_sharding/bucket.rs
```diff
@@ -0,0 +1,234 @@
...
+{
+    let mut step: usize = 1 << BK::BITS;
+
+    assert!(
+        breakdown_count <= 1 << BK::BITS,
+        "Asking for more buckets ({breakdown_count}) than bits in the key ({}) allow",
+        BK::BITS
+    );
+    assert!(
+        breakdown_count <= 128,
```
richajaindce, line 56: With #799 this would change to 2047. Should we change it already here?

src/protocol/prf_sharding/bucket.rs
```diff
@@ -0,0 +1,234 @@
...
+    #[should_panic]
+    fn move_out_of_range_too_many_buckets_type() {
+        run(move || async move {
+            _ = move_to_bucket(MAX_BREAKDOWN_COUNT + 1, 0, false).await;
+        });
+    }
+
+    #[test]
+    #[should_panic]
+    fn move_out_of_range_too_many_buckets_steps() {
```
richajaindce, line 208: Arent both these functions checking max breakdown count > MAX_NUMBER?

martinthomson: The first should fail on the first check (`count > (1<< BK::BITS)`).

The second should fail on the second check (`count > 128`).  We can increase that once #799 lands)

richajaindce: oh ok. Sounds good

richajaindce (approved):
Looks good :)

src/protocol/prf_sharding/bucket.rs
martinthomson: Just noting that #799 itself doesn't fix this, but might allow this to change.  We'll need to keep this synchronized manually.

PR #805, opened 2023-10-16
UnorderedReceiver can be closed now

akoshelev:
This is required to properly implement resource cleanup and install debugging infrastructure to Gateway.

One issue we noticed while working on it, is that it is impossible to distinguish between messages awaiting receive and streams that have been fully exhausted.

After this change, observers can check `is_closed` status and exclude this stream.

In the future, when resource cleanup will be a thing, such receivers can be safely removed from the channel map.

### DIFF
--- a/src/helpers/buffers/unordered_receiver.rs
+++ b/src/helpers/buffers/unordered_receiver.rs
@@ -110,6 +110,7 @@ impl Spare {
         };
         Some(m)
     }
+
     /// Returns `true` if there are no bytes currently awaiting a read.
     fn is_empty(&self) -> bool {
         self.offset == self.buf.len()
@@ -153,6 +154,8 @@ where
     /// that easing load on this mechanism.  There might also need to be some
     /// end-to-end back pressure for tasks that do not involve sending at all.
     overflow_wakers: Vec<Waker>,
+    /// If this receiver is closed and no longer capable of receiving data.
+    closed: bool,
     _marker: PhantomData<C>,
 }
 
@@ -212,53 +215,78 @@ where
 
     /// Poll for the next record.  This should only be invoked when
     /// the future for the next message is polled.
+    ///
+    /// ## Errors
+    /// If buffer capacity is not enough to read `M` and the underlying stream does not have
+    /// more data. This may lead to different behavior depending on the order of issued reads. See
+    /// [`read_order`] test for an example.
+    ///
+    /// [`read_order`]: test::read_order
     fn poll_next<M: Message>(&mut self, cx: &mut Context<'_>) -> Poll<Result<M, Error>> {
         // If spare has enough data for us, poll it first
         // otherwise, poll the underlying stream until it returns pending or it provides enough
         // data to return a value.
-
         let message = self.spare.read();
-        let next = match message {
-            Some(m) => {
-                // this check exists to make sure the inner stream is eventually moved to
-                // the closed state. We don't want to poll it too often, but we also need to know
-                // when it is done and `UnorderedReceiver` can be dropped.
-                if self.spare.is_empty() {
-                    // we don't want to be woken up here, control loop is driven by the client.
-                    // They decide when they want the next message and must issue a `poll` for it.
+        let next = if let Some(m) = message {
+            // this check exists to make sure the inner stream is eventually moved to
+            // the closed state. We don't want to poll it too often, but we also need to know
+            // when it is done and `UnorderedReceiver` can be dropped.
+            if self.spare.is_empty() {
+                // we don't want to be woken up here, control loop is driven by the client.
+                // They decide when they want the next message and must issue a `poll` for it.
 
-                    // TODO: https://github.com/rust-lang/rust/issues/98286
-                    let mut cx = Context::from_waker(noop_waker_ref());
-                    match self.stream.as_mut().poll_next(&mut cx) {
-                        Poll::Ready(Some(bytes)) => {
-                            // Spare is empty because of the check above.
-                            self.spare.replace(bytes.as_ref());
-                        }
-                        Poll::Ready(None) | Poll::Pending => {}
+                // TODO: https://github.com/rust-lang/rust/issues/98286
+                let mut cx = Context::from_waker(noop_waker_ref());
+                match self.stream.as_mut().poll_next(&mut cx) {
+                    Poll::Ready(Some(bytes)) => {
+                        // Spare is empty because of the check above.
+                        self.spare.replace(bytes.as_ref());
                     }
+                    Poll::Ready(None) => {
+                        self.close();
+                    }
+                    Poll::Pending => {}
                 }
-
-                self.wake_next();
-                Poll::Ready(Ok(m))
             }
-            None => loop {
-                match ready!(self.stream.as_mut().poll_next(cx)) {
-                    Some(bytes) => {
-                        if let Some(m) = self.spare.extend(bytes.as_ref()) {
-                            self.wake_next();
-                            break Poll::Ready(Ok(m));
-                        }
+
+            self.wake_next();
+            Poll::Ready(Ok(m))
+        } else {
+            loop {
+                if let Some(bytes) = ready!(self.stream.as_mut().poll_next(cx)) {
+                    if let Some(m) = self.spare.extend(bytes.as_ref()) {
+                        self.wake_next();
+                        break Poll::Ready(Ok(m));
                     }
-                    None => {
-                        break Poll::Ready(Err(Error::EndOfStream {
-                            record_id: RecordId::from(self.next),
-                        }));
-                    }
+                } else {
+                    self.spare.replace(&[]);
+                    self.close();
+                    break Poll::Ready(Err(Error::EndOfStream {
+                        record_id: RecordId::from(self.next),
+                    }));
                 }
-            },
+            }
         };
 
         next
+    }
+
+    /// Returns `true` if this receiver is closed.
+    fn is_closed(&self) -> bool {
+        self.closed
+    }
+
+    /// Close this receiver, so it can no longer be used to poll messages.
+    /// ## Errors
+    /// If the underlying stream is not closed or if there are unread bytes inside the buffer.
+    fn close(&mut self) {
+        if !self.closed {
+            assert!(self.stream.is_done());
+            assert!(self.spare.is_empty());
+
+            self.closed = true;
+            self.spare = Spare::default();
+        }
     }
 }
 
@@ -297,6 +325,7 @@ where
                 spare: Spare::default(),
                 wakers,
                 overflow_wakers: Vec::new(),
+                closed: false,
                 _marker: PhantomData,
             })),
         }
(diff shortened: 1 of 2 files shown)


src/helpers/buffers/unordered_receiver.rs
```diff
@@ -105,6 +106,10 @@ impl Spare {
         };
         Some(m)
     }
+    /// Returns `true` if there are no bytes currently awaiting a read.
```
martinthomson, line 109:
```suggestion

    /// Returns `true` if there are no bytes currently awaiting a read.
```

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -204,29 +209,54 @@ where
...
+                // this check exists to make sure the inner stream is eventually moved to
+                // the closed state. We don't want to poll it too often, but we also need to know
+                // when it is done and `UnorderedReceiver` can be dropped.
+                if self.spare.is_empty() {
+                    // we don't want to be woken up here, control loop is driven by the client.
+                    // They decide when they want the next message and must issue a `poll` for it.
+
+                    // TODO: https://github.com/rust-lang/rust/issues/98286
+                    let mut cx = Context::from_waker(futures::task::noop_waker_ref());
+                    match self.stream.as_mut().poll_next(&mut cx) {
```
martinthomson, line 228: `if let` ...

martinthomson, line 228: Does this even need to poll the stream if it is using spare bytes?  The next call will then hit the `None` arm of the outer match, which polls the stream properly.

I don't think that it is a good idea to read proactively like this.  We've already got what we need, so this will only inflate our buffer and fail to provide proper back pressure.  Given that we're waking the next item as noted, I don't see why the next poll won't work.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -204,29 +209,54 @@ where
...
+        // otherwise, poll the underlying stream until it returns pending or it provides enough
+        // data to return a value.
+
+        let message = self.spare.read::<M>();
+        let next = match message {
+            Some(m) => {
+                // this check exists to make sure the inner stream is eventually moved to
+                // the closed state. We don't want to poll it too often, but we also need to know
+                // when it is done and `UnorderedReceiver` can be dropped.
+                if self.spare.is_empty() {
```
martinthomson, line 222: This will fail to capture the case where you have an incomplete message buffered.  In that case, you will get `None` from an attempt to read, but you won't have added more data to the buffer.

Because this function cannot know how big the next message is, you can't really do anything different here, but I would argue that topping up the spare is not helpful.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -204,29 +209,54 @@ where
...
+                        break Poll::Ready(Err(Error::EndOfStream {
+                            record_id: RecordId::from(self.next),
+                        }));
+                    }
+                }
+            },
+        };
+
+        if next.is_ready() {
+            self.wake_next();
```
martinthomson, line 256: There are only two places that you need to call this.  Line 243 and line 237.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -284,6 +314,19 @@ where
...
+    /// there is no more data inside receiver's buffers.
+    ///
+    /// Calling `poll_next` on closed receivers will result in [`EOS`] error.
+    ///
+    /// [`EOS`]: crate::helpers::Error::EndOfStream
+    pub fn is_closed(&self) -> bool {
+        // If this function is ever called on the hot path, consider caching closed status.
+        // Closed streams cannot move back to open.
+        let inner = self.inner.lock().unwrap();
+        inner.stream.is_done() && inner.spare.is_empty()
```
martinthomson, line 328: Consider what might happen if you have spare bytes that are too small for a single message.  I think that tends to end up with an error, but you won't return `is_closed() == true` in that case.

I think that you want to flush spare in that case.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -204,29 +209,54 @@ where
...
             }
+            None => loop {
+                match ready!(self.stream.as_mut().poll_next(cx)) {
+                    Some(bytes) => {
+                        if let Some(m) = self.spare.extend(bytes.as_ref()) {
+                            break Poll::Ready(Ok(m));
+                        }
+                    }
+                    None => {
+                        break Poll::Ready(Err(Error::EndOfStream {
```
martinthomson, line 247: You should flush `self.spare` if you hit this arm, see below.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -510,4 +553,32 @@ mod test {
...
+    fn end_of_stream() {
+        const DATA: &[u8] = &[1u8, 2, 3, 4, 5];
+        run(|| async move {
+            let recv = receiver([DATA]);
+            let _: Fp32BitPrime = recv.recv(0u8).await.unwrap();
+
+            assert!(matches!(
+                recv.recv::<Fp32BitPrime, _>(1u8).await,
+                Err(EndOfStream { .. })
+            ));
```
martinthomson, line 581: Test is_closed() here as well.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -113,7 +118,7 @@ where
     C: AsRef<[u8]>,
 {
     /// The stream we're reading from.
-    stream: Pin<Box<S>>,
+    stream: Pin<Box<Fuse<S>>>,
```
martinthomson, line 121: 👍

martinthomson (commented):
I'm not seeing the justification for some of these changes.

src/helpers/buffers/unordered_receiver.rs
akoshelev: oh, that's right. end of stream does not need that

src/helpers/buffers/unordered_receiver.rs
akoshelev: This is the intended behavior. If there is something left in the spare, there must be a receive request for it later. This will correctly poll the stream if spare does not have enough bytes. The only edge case, where we don't move the fused stream to the closed state, is when spare is empty after the last read.

src/helpers/buffers/unordered_receiver.rs
akoshelev: Now I am convinced that `M` must be on `OperatingState` rather than on `poll_next`. Will make this change

akoshelev: I take this back - moving `M` up is not possible because we need heterogeneous channels and we keep them all in the same map. 

I will still close the receiver, but will add a note to the `poll_next` as this behavior may be confusing in cases when we read different messages off the same channel

src/helpers/buffers/unordered_receiver.rs
akoshelev: restructured the code and now I need `Poll::Ready(None)`

src/helpers/buffers/unordered_receiver.rs
akoshelev: I believe now it is handled - next poll will return an error and stream will be properly closed

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -203,30 +215,78 @@ where
...
+    ///
+    /// [`read_order`]: test::read_order
     fn poll_next<M: Message>(&mut self, cx: &mut Context<'_>) -> Poll<Result<M, Error>> {
-        if let Some(m) = self.spare.read() {
+        // If spare has enough data for us, poll it first
+        // otherwise, poll the underlying stream until it returns pending or it provides enough
+        // data to return a value.
+        let message = self.spare.read();
+        let next = if let Some(m) = message {
+            // this check exists to make sure the inner stream is eventually moved to
```
martinthomson, line 231:
```suggestion
            // This check exists to make sure the inner stream is eventually moved to
```

I don't know if this comment is still accurate.

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -203,30 +215,78 @@ where
...
+    /// Returns `true` if this receiver is closed.
+    fn is_closed(&self) -> bool {
+        self.closed
+    }
+
+    /// Close this receiver, so it can no longer be used to poll messages.
+    /// ## Errors
+    /// If the underlying stream is not closed or if there are unread bytes inside the buffer.
+    fn close(&mut self) {
+        if !self.closed {
```
martinthomson, line 283: Shouldn't we just assert instead?

martinthomson (approved):

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -203,30 +215,78 @@ where
...
+                    break Err(Error::EndOfStream {
+                        record_id: RecordId::from(self.next),
+                    });
+                }
+            }
+        };
+
+        if next.is_ok() && self.spare.is_empty() {
+            // we don't want to be woken up here, control loop is driven by the client.
+            // They decide when they want the next message and must issue a `poll` for it.
```
akoshelev, line 255: this is a bug - at the end of the stream if stream is not ready, nothing will be polled

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -203,30 +215,78 @@ where
...
+        self.closed
+    }
+
+    /// Close this receiver, so it can no longer be used to poll messages.
+    /// ## Errors
+    /// If the underlying stream is not closed or if there are unread bytes inside the buffer.
+    fn close(&mut self) {
+        if !self.closed {
+            assert!(self.stream.is_done());
+            assert!(self.spare.is_empty());
```
akoshelev, line 285: assert that wakers are empty?

src/helpers/buffers/unordered_receiver.rs
```diff
@@ -203,30 +215,78 @@ where
...
+            }
+        };
+
+        if next.is_ok() && self.spare.is_empty() {
+            // we don't want to be woken up here, control loop is driven by the client.
+            // They decide when they want the next message and must issue a `poll` for it.
+
+            // TODO: https://github.com/rust-lang/rust/issues/98286
+            let mut cx = Context::from_waker(noop_waker_ref());
+            match self.stream.as_mut().poll_next(&mut cx) {
```
akoshelev, line 259: @martinthomson We discussed that a real waker should be used here if there are no active wakers in the receiver and it is the last record. When the underlying stream returns `Poll::Ready` everything is fine, however there seems to be a subtle violation of `Stream` contract - we already have an item in our hand to return (line 271 `next`) but if underlying stream is not ready, the executor is not be obligated to poll us again, because we returned `Poll::Ready`. 

I could probably make it work by holding onto `next` and returning `Poll::Pending`, but that's not great in case when we are racing with receivers, i.e. it is not the end of the stream yet.

Trying to think if there is a good solution here.

martinthomson: Hmm, that doesn't sound great.

We could do as you say and always be one item ahead.  That would ensure that we deliver the last item and close at the same time.  But I think that the core problem is a lack of anything interested in the close.

Would it make sense for us to have the context/gateway infrastructure be interested in the close of each receiver?  That is, for each receiver that we register, we have that hooked into the context, with a future that resolves when the receiver closed.  Before we close a query, we would await those futures to ensure that they were driven to completion (you could also spawn a task for that purpose that is in turn awaited before the transaction resolves).

benjaminsavage:
@akoshelev - can we abandon this PR now?

akoshelev:
yea lets close it for now

PR #808, opened 2023-10-18
OPRF attribution windowing

taikiy:
Adding attribution window to the OPRF attribution protocol.

### DIFF
--- a/src/protocol/prf_sharding/mod.rs
+++ b/src/protocol/prf_sharding/mod.rs
@@ -97,7 +97,6 @@ impl InputsRequiredFromPrevRow {
     ///         - `did_trigger_get_attributed` - a secret-shared bit indicating if this row corresponds to a trigger event
     ///           which was attributed. Might be able to reveal this (after a shuffle and the addition of dummies) to minimize
     ///           the amount of processing work that must be done in the Aggregation stage.
-    #[allow(clippy::too_many_lines)]
     pub async fn compute_row_with_previous<C, BK, TV, TS>(
         &mut self,
         ctx: C,
@@ -152,29 +151,15 @@ impl InputsRequiredFromPrevRow {
         )
         .await?;
 
-        let (did_trigger_get_attributed, is_trigger_within_window) = try_join(
-            input_row.is_trigger_bit.multiply(
-                &ever_encountered_a_source_event,
-                ctx.narrow(&Step::DidTriggerGetAttributed),
-                record_id,
-            ),
-            is_trigger_event_within_attribution_window(
-                ctx.narrow(&Step::CheckAttributionWindow),
-                record_id,
-                attribution_window_seconds,
-                &timestamp,
-                &source_event_timestamp,
-            ),
-        )
-        .await?;
-
         let attributed_trigger_value = zero_out_trigger_value_unless_attributed(
             ctx.narrow(&Step::AttributedTriggerValue),
             record_id,
-            &did_trigger_get_attributed,
-            &is_trigger_within_window,
+            &input_row.is_trigger_bit,
+            &ever_encountered_a_source_event,
             &tv,
             attribution_window_seconds,
+            &timestamp,
+            &source_event_timestamp,
         )
         .await?;
 
@@ -218,7 +203,6 @@ impl InputsRequiredFromPrevRow {
         self.source_event_timestamp = source_event_timestamp;
 
         let outputs_for_aggregation = CappedAttributionOutputs {
-            did_trigger_get_attributed,
             attributed_breakdown_key_bits,
             capped_attributed_trigger_value,
         };
@@ -228,7 +212,6 @@ impl InputsRequiredFromPrevRow {
 
 #[derive(Debug)]
 pub struct CappedAttributionOutputs {
-    pub did_trigger_get_attributed: Replicated<Gf2>,
     pub attributed_breakdown_key_bits: BitDecomposed<Replicated<Gf2>>,
     pub capped_attributed_trigger_value: BitDecomposed<Replicated<Gf2>>,
 }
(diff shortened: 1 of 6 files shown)


src/protocol/prf_sharding/mod.rs
```diff
@@ -117,14 +160,24 @@ impl InputsRequiredFromPrevRow {
...
+        let mut attributed_trigger_value = zero_out_trigger_value_unless_attributed(
             ctx.narrow(&Step::AttributedTriggerValue),
             record_id,
             &did_trigger_get_attributed,
             &tv,
         )
         .await?;
 
+        attributed_trigger_value = zero_out_trigger_value_if_outside_attribution_window(
+            ctx.narrow(&Step::CheckAttributionWindow),
```
benjaminsavage, line 172: I recommend saving a few multiplications by lifting out the "Is this within the attribution window" and storing that as a local variable. Then multiply that with `did_trigger_get_attributed`, then pass the result of that multiplication into a method that multiplies with all the bits of the trigger value.

src/protocol/prf_sharding/mod.rs
```diff
@@ -480,6 +590,66 @@ where
...
+            record_id,
+            source_event_timestamp,
+        )
+        .await?;
+
+    // The result is true if the time delta is `[0, attribution_window_seconds)`
+    // If we want to include the upper bound, we need to use `bitwise_greater_than_constant()`
+    // and negate the result.
+    let is_trigger_event_inside_attribution_window = bitwise_less_than_constant(
+        ctx.narrow(&Step::ConvertTimeDeltaBits),
```
benjaminsavage, line 627: I'd suggest a different step name like `CompareTimeDeltaToAttributionWindow`

src/secret_sharing/decomposed.rs
```diff
@@ -79,6 +83,46 @@ impl<S> BitDecomposed<S> {
...
+    {
+        assert!(self.len() >= rhs.len());
+
+        let mut output = vec![];
+        let mut carry_in = S::share_known_value(&ctx, Gf2::ONE);
+        let zero = S::ZERO;
+        for i in 0..self.len() {
+            let c = ctx.narrow(&BitOpStep::from(i));
+            let compute_carry_out = i < self.len() - 1;
+            let lfs = &self[i];
```
benjaminsavage, line 115: lhs

benjaminsavage (changes requested):
Looks pretty good. Please save on those multiplications (TV::BITS mults saved!). That's my main comment.

src/protocol/prf_sharding/mod.rs
```diff
@@ -108,6 +137,20 @@ impl InputsRequiredFromPrevRow {
         )
         .await?;
 
+        let source_event_timestamp = match attribution_window_seconds {
+            None => self.source_event_timestamp.clone(),
+            Some(_) => {
+                timestamp_of_most_recent_source_event(
```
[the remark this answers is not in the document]
taikiy: done

src/protocol/prf_sharding/mod.rs
taikiy: oops, done

src/secret_sharing/decomposed.rs
taikiy: "left fand side"

taikiy: fixed :)

martinthomson: is this like having a variable called "variable"?

src/protocol/prf_sharding/mod.rs
```diff
@@ -68,30 +97,37 @@ impl InputsRequiredFromPrevRow {
     ///         - `did_trigger_get_attributed` - a secret-shared bit indicating if this row corresponds to a trigger event
     ///           which was attributed. Might be able to reveal this (after a shuffle and the addition of dummies) to minimize
     ///           the amount of processing work that must be done in the Aggregation stage.
-    pub async fn compute_row_with_previous<C, BK, TV>(
+    #[allow(clippy::too_many_lines)]
```
benjaminsavage, line 100: Does it still exceed the max number of lines now, after this latest refactoring?

src/protocol/prf_sharding/mod.rs
```diff
@@ -408,32 +475,82 @@ where
...
+        }
+    }
+}
+
+/// Move a field of the most recent source event down to all of trigger events that follow it.
+///
+/// The logic here is extremely simple. For each row:
+/// (a) if it is a source event, take the current field.
+/// (b) if it is a trigger event, take the field from the preceding line
+async fn field_of_most_recent_source_event<C>(
```
benjaminsavage, line 539: Instead of "Field" perhaps call it "Attribute" since the former is kind of an overloaded term.

src/protocol/prf_sharding/mod.rs
```diff
@@ -480,6 +604,49 @@ where
...
+    source_event_timestamp: &BitDecomposed<Replicated<Gf2>>,
+) -> Result<Replicated<Gf2>, Error>
+where
+    C: UpgradedContext<Gf2, Share = Replicated<Gf2>>,
+{
+    // if there is no attribution window, then all trigger events are attributed
+    if attribution_window_seconds.is_none() {
+        let share_of_one = Replicated::share_known_value(&ctx, Gf2::ONE);
+        return Ok(share_of_one);
+    }
```
benjaminsavage, line 625: It's unfortunate that we are still going to multiply this sharing of ONE with the "did trigger get attributed" in the case of no windowing.

src/protocol/prf_sharding/mod.rs
```diff
@@ -105,22 +141,38 @@ impl InputsRequiredFromPrevRow {
...
+                &source_event_timestamp,
+            ),
+        )
+        .await?;
 
         let attributed_trigger_value = zero_out_trigger_value_unless_attributed(
             ctx.narrow(&Step::AttributedTriggerValue),
             record_id,
             &did_trigger_get_attributed,
+            &is_trigger_within_window,
```
benjaminsavage, line 175: An alternative approach, which avoids the unnecessary multiplication in the case of no windowing, is to just lift that multiplication out of the `zero_out_trigger_value_unless_attributed` function and do it here, like this:

```
let zero_out_flag = if Some(window) = attribution_window_seconds {
    let c = ctx.narrow(&Step::AttributedEventCheckFlag);
    &did_trigger_get_attributed
        .multiply(is_trigger_within_window, c, record_id)
        .await?
} else {
    did_trigger_get_attributed
}

let attributed_trigger_value = zero_out_trigger_value_unless_attributed(
    ctx.narrow(&Step::AttributedTriggerValue),
    record_id,
    &zero_out_flag,
    &tv,
)
.await?;
```

benjaminsavage (changes requested):
Looks good! Just a few more suggestions.

src/protocol/prf_sharding/mod.rs
taikiy: by 2 lines :(

src/protocol/prf_sharding/mod.rs
taikiy: I added the `attribution_window_seconds` to `zero_out_...()` and do it inside the function instead. I wanted to move both `did_trigger_get_attributed` and `is_trigger_within_window` flags inside the function too, but we need `did_trigger_get_attributed` to mutate `self` at the end of `compute_row_with_previous`.

@benjaminsavage, do we plan to use `did_trigger_get_attributed` in the output? Currently we set it in `CappedAttributionOutputs` but never reference it.

src/protocol/prf_sharding/mod.rs
```diff
@@ -557,22 +540,41 @@ where
 /// another secret-shared bit indicating if a given row is within the attribution window. We multiply these two bits together and
 /// multiply it with the bits of the `trigger_value` in order to zero out contributions from unattributed trigger events.
 ///
+#[allow(clippy::too_many_arguments)]
```
taikiy, line 543: `too_many_lines` is no more, but now `too_many_arguments`... (8 vs. 7)
I could create a struct for timestamps + attribution window, but that requires either clone()'s or a life-time. I think this is more readable.

benjaminsavage (approved):
I love it. Thanks for doing this @taikiy

PR #809, opened 2023-10-18
[RFC][oprf][shuffle] OPRF Shuffle using a 2-round 4-message shuffle protocol

cryo28:
1. New query type
2. Implementation of the protocol (not sharded)

This is not a final code I'd like to land. But more a preview to gather feedback and comments. In a few places I don't really know what I am doing so I resorted to copy-pasta. 

If there is not major problems with the code I'll proceed to writing a bunch of unit tests, docs to package the commit for merging.

image


src/protocol/context/oprf.rs
```diff
@@ -0,0 +1,88 @@
...
+            Base, InstrumentedIndexedSharedRandomness, InstrumentedSequentialSharedRandomness,
+        },
+        prss::Endpoint as PrssEndpoint,
+        step::{Gate, Step, StepNarrow},
+    },
+    seq_join::SeqJoin,
+};
+
+#[derive(Clone)]
+pub struct Context<'a> {
```
martinthomson, line 16: What is the purpose of this code?

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,610 @@
...
+            batch_size,
+        )
+    )?;
+
+    let c_hat = add_single_shares(c_hat_1.iter(), c_hat_2.iter());
+    let res = combine_shares(b_hat, c_hat);
+    Ok(res)
+}
+
+async fn run_h3<C, L, R>(
```
martinthomson, line 366: There is a lot of code duplication between these functions.

src/helpers/transport/query/oprf_shuffle.rs
```diff
@@ -0,0 +1,17 @@
...
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+pub struct QueryConfig {
+    pub bk_size: u8, // breakdown key size bits
+    pub tv_size: u8, // trigger value size bits
+}
+
+impl Default for QueryConfig {
+    fn default() -> Self {
+        Self {
+            bk_size: 40,
```
martinthomson, line 13: This could probably be smaller.

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    let x_1_arg = add_single_shares(add_single_shares(a, b), z_12);
+    let x_1 = permute(&pi_12, x_1_arg);
+    let x_2_arg = add_single_shares(x_1.iter(), z_31.iter());
+    let x_2 = permute(&pi_31, x_2_arg);
+
+    send_to_peer(
+        ctx,
+        &OPRFShuffleStep::TransferX2,
+        Direction::Right,
+        x_2.clone(),
```
akoshelev, line 229: clone here feels wrong

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        Direction::Left => rng.0,
+        Direction::Right => rng.1,
+    };
+
+    let mut permutation = (0..batch_size).collect::<Vec<_>>();
+    permutation.shuffle(&mut rng);
+    permutation
+}
+
+fn permute(
```
akoshelev, line 495: this relies on the `IntoInterator` specialization to avoid extra allocations, but I don't see why it can't be a one liner `apply(permutation, &mut data)` on the caller's side

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+use bitvec::bitvec;
+use embed_doc_image::embed_doc_image;
+
+#[embed_doc_image("apply", "images/sort/apply.png")]
+/// Permutation reorders (1, 2, . . . , m) into (σ(1), σ(2), . . . , σ(m)).
+/// For example, if σ(1) = 2, σ(2) = 3, σ(3) = 1, and σ(4) = 0, an input (A, B, C, D) is reordered into (C, D, B, A) by σ.
+///
+/// ![Apply steps][apply]
+fn apply<T>(permutation: &[u32], values: &mut [T]) {
+    // NOTE: This is copypasta from crate::protocol::sort
```
akoshelev, line 513: should we lift it out of sort and reuse?

akoshelev (commented):
Haven't finished the whole thing, some initial feedback in place

src/protocol/context/mod.rs
```diff
@@ -8,6 +9,7 @@ use std::{num::NonZeroUsize, sync::Arc};
 
 use async_trait::async_trait;
 pub use malicious::{Context as MaliciousContext, Upgraded as UpgradedMaliciousContext};
+pub use oprf::Context as OPRFContext;
```
akoshelev, line 12: leftover?

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    sample_iter(rng).take(batch_size as usize)
+}
+
+fn sample_iter<R: Rng>(rng: R) -> impl Iterator<Item = OPRFShare> {
+    rng.sample_iter(Standard)
+}
+
+// ---------------------------- helper communication ------------------------------------ //
+
+async fn send_to_peer<C: Context, I: IntoIterator<Item = OPRFShare>>(
```
danielmasny, line 283: Do we need separate functions for send and receive here? It seems that ctx.send are simple enough to be used directly. Then you don't need to pass all these inputs. @akoshelev what are the best practices here?

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    for record_id in 0..batch_size {
+        let msg = receive_channel.receive(RecordId::from(record_id)).await?;
+        output.push(msg);
+    }
+
+    Ok(output)
+}
+
+// ------------------ Pseudorandom permutations functions -------------------- //
+
```
danielmasny, line 316: similar as above, do we need separate functions here? I think it makes the code less clean.

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    let mut rng = narrow_context.prss_rng();
+
+    let with_left = generate_pseudorandom_permutation(batch_size, &mut rng.0);
+    let with_right = generate_pseudorandom_permutation(batch_size, &mut rng.1);
+    (with_left, with_right)
+}
+
+fn generate_pseudorandom_permutation<R: Rng>(batch_size: u32, rng: &mut R) -> Vec<u32> {
+    let mut permutation = (0..batch_size).collect::<Vec<_>>();
+    permutation.shuffle(rng);
```
danielmasny, line 328: Do we use the same function: .shuffle in our sorting protocol for generating random permutations from a seed? It seems to me that this could be generated on the fly rather than storing this huge decompressed permutation in memory.

src/protocol/oprf/oprf_share.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct OPRFShare {
+    pub timestamp: OprfF,
+    pub mk: OprfMK,
+    pub is_trigger_bit: OprfF,
+    pub breakdown_key: OprfBK,
+    pub trigger_value: OprfF,
+}
+
```
danielmasny, line 24: Could we implement (Weak)SharedValue for OPRFShare? Then we could operate on SharedValue during the shuffle. (WeakSharedValue is part of https://github.com/private-attribution/ipa/pull/795#top and only requires to implement additions).

src/protocol/oprf/oprf_share.rs
```diff
@@ -0,0 +1,148 @@
...
+            timestamp,
+            mk,
+            is_trigger_bit,
+            breakdown_key,
+            trigger_value,
+        }
+    }
+}
+
+impl Message for OPRFShare {}
```
danielmasny, line 148: We dont need that when implementing SharedValue

src/protocol/sort/mod.rs
```diff
@@ -1,9 +1,9 @@
+pub mod apply;
```
danielmasny, line 1: see above, we should consider remove this from sort and add it e.g. as a basic protocol

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    permutation
+}
```
danielmasny, line 330: There is not a single test. We should test whether this is actually correct, by applying it e.g. to a sorted list, and checking the output for inequality with the input list and then sorting the output again and checking the list with the input list for equality. It should be easy to do by using TestWorld

danielmasny (commented):
Shuffle looks great to me! I have a couple of suggestions and left them as comments. I haven't checked the new query type aspects of the pr. 

Thanks for all your work!

src/protocol/oprf/mod.rs
akoshelev: we don't have to. If it is common to send and receive all records sequentially, we could add a trait to this module that extends iterators and provides `seq_send` method that takes an iterable of messages and sends them all

cryo28:
Closing in favor of #816 816

PR #811, opened 2023-10-20
Upgrade dependencies

akoshelev:
Notable bumps

* GenericArray to 1.0
* HPKE changed a lot, moving from 0.10 to 0.11
* Explicit dependency on 'rust-webpki' due to https://rustsec.org/advisories/RUSTSEC-2023-0053
* Use native `rustls` types rather than their proxies defined in `tokio-rustls`


martinthomson (approved):
FORWARD!

Thanks for doing the drudge work.

PR #813, opened 2023-10-24
Minor changes/fixes in OPRF

richajaindce:
1. Increasing max number of breakdowns we can support with OPRF to 512
2. In case no conversion is attributed to any bucket, do_aggregation would return vector with zeroes

### DIFF
--- a/src/protocol/prf_sharding/bucket.rs
+++ b/src/protocol/prf_sharding/bucket.rs
@@ -1,16 +1,34 @@
 use embed_doc_image::embed_doc_image;
+use ipa_macros::Step;
 
 use crate::{
     error::Error,
     ff::{GaloisField, PrimeField, Serializable},
     protocol::{
-        basics::SecureMul, context::UpgradedContext, prf_sharding::BinaryTreeDepthStep,
-        step::BitOpStep, RecordId,
+        basics::SecureMul, context::UpgradedContext, prf_sharding::BinaryTreeDepthStep, RecordId,
     },
     secret_sharing::{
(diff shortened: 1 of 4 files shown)


src/protocol/prf_sharding/bucket.rs
```diff
@@ -221,7 +239,7 @@ pub mod tests {
                             RecordId::from(0),
                             breakdown_key_share,
                             value_share,
-                            129,
+                            513,
```
martinthomson, line 242: Up on line 228, you need to find a different GfXBit type for this function.  Otherwise you'll never get past 256 buckets.

src/protocol/prf_sharding/mod.rs
```diff
@@ -592,6 +592,13 @@ where
     F: PrimeField + ExtendableField,
 {
     let num_records = user_level_attributions.len();
+
+    // in case no attributable conversion is found, return 0.
+    // as anyways the helpers know that no attributions resulted.
+    if num_records == 0 {
```
martinthomson, line 598: Do you have a test for this?

richajaindce: Adding now

src/protocol/prf_sharding/bucket.rs
```diff
@@ -1,17 +1,35 @@
...
     },
 };
 
+#[derive(Step)]
+pub enum BucketStep {
+    #[dynamic(256)]
+    Bit(usize),
+}
+
+impl From<u32> for BucketStep {
```
akoshelev, line 21: this is a fallible conversion, want to implement `TryFrom` for it?

benjaminsavage (approved):
LGTM

PR #814, opened 2023-10-24
Oneshot for OPRF + enabling compact-gate for it

richajaindce:
Sample command to run
cargo bench --bench oneshot_ipa --features="enable-benches descriptive-gate" --no-default-features -- --per-user-cap 64 --oprf --max-trigger-value 4 -n 100


src/test_fixture/ipa.rs
```diff
@@ -191,3 +195,80 @@ pub async fn test_ipa<F>(
...
+    records.sort_by(|a, b| b.user_id.cmp(&a.user_id));
+
+    let result: Vec<F> = world
+        .semi_honest(
+            records.into_iter(),
+            |ctx, input_rows: Vec<OprfReport<Timestamp, BreakdownKey, TriggerValue>>| async move {
+                let sharded_input = input_rows
+                    .into_iter()
+                    .map(|single_row| {
+                        let is_trigger_bit_share = if single_row.event_type == EventType::Trigger {
```
benjaminsavage, line 239: Consider adding a comment here to explain this is temporary and will go away once we have a real
1. Decrypt
2. Add dummy records
3. Shuffle
4. Compute OPRF

src/test_fixture/ipa.rs
```diff
@@ -191,3 +195,80 @@ pub async fn test_ipa<F>(
...
+            },
+        )
+        .await
+        .reconstruct();
+
+    let mut result = result
+        .into_iter()
+        .map(|v| u32::try_from(v.as_u128()).unwrap())
+        .collect::<Vec<_>>();
+    let _ = result.split_off(expected_results.len());
```
benjaminsavage, line 272: What is this?

scripts/collect_steps.py
```diff
@@ -171,8 +169,42 @@ def extract_intermediate_steps(steps):
                     print(" ".join(args), file=sys.stderr)
                     steps.update(collect_steps(args))
 
+OPRF_USER_CAP = [16, 64, 128]
+OPRF_SECURITY_MODEL = ["semi-honest"]
+OPRF_TRIGGER_VALUE = [4]
```
benjaminsavage, line 174: consider 7?

src/test_fixture/input/sharing.rs
```diff
@@ -354,3 +354,33 @@ where
...
+{
+    fn share_with<R: Rng>(self, rng: &mut R) -> [OprfReport<TS, BK, TV>; 3] {
+        let event_type = if self.is_trigger_report {
+            EventType::Trigger
+        } else {
+            EventType::Source
+        };
+        let timestamp = TS::truncate_from(self.timestamp).share_with(rng);
+        let breakdown_key = BK::truncate_from(self.breakdown_key).share_with(rng);
+        let trigger_value = TV::truncate_from(self.trigger_value).share_with(rng);
```
benjaminsavage, line 372: Should probably not use `truncate_from` as this will hide errors (like we encountered today!) we should fail spectatularly.

steps1.txt
taikiy, line 1: Remove this file

benjaminsavage (approved):
Thanks Richa, this is great.

scripts/collect_steps.py
```diff
@@ -171,6 +169,40 @@ def extract_intermediate_steps(steps):
...
+                            str(tv),
+                            "-o"
+                        ]
+                    print(" ".join(args), file=sys.stderr)
+                    steps.update(collect_steps(args))
+
+if __name__ == "__main__":
+    steps = set()
+    ipa_steps(steps)
+    oprf_steps(steps)
```
taikiy, line 204: Ideally, each function should return a new set and update the `steps`, rather than mutating the argument.

scripts/collect_steps.py
```diff
@@ -171,6 +169,40 @@ def extract_intermediate_steps(steps):
                     print(" ".join(args), file=sys.stderr)
                     steps.update(collect_steps(args))
 
+OPRF_BREAKDOWN_KEYS = [256]
+OPRF_USER_CAP = [16, 64, 128]
+OPRF_SECURITY_MODEL = ["semi-honest"]
+OPRF_TRIGGER_VALUE = [7]
```
taikiy, line 175: If we are ever testing the OPRF protocol with a single value parameter (i.e., breakdown keys, security model, and trigger value), you don't need to define them as lists nor use `for` loops. Instead of the deep `for` chains, replace them with e.g., `ARGS + [..., "-m", OPRF_SECURITY_MODEL, ...]?

benjaminsavage:
Now that the checks are passing, I'm going to merge this to fix the build. @richajaindce - please follow up with another small PR to address @taikiy 's comments.

scripts/collect_steps.py
richajaindce: originally I was thinking of keeping the array so that we can add different values but for now, will revert to your suggestion. In future if we see a need, we can always add one more for loop

PR #815, opened 2023-10-25
Supporting the opposite capping logic

benjaminsavage:
Updating IPA in the clear logic to be capable of supporting either type of capping, either you cap the OLDEST stuff first, or you cap the MOST RECENT stuff first.


richajaindce (approved):
Looks great!

PR #816, opened 2023-10-25
[oprf][shuffle] OPRF Shuffle using a 2-round 4-message shuffle protocol

cryo28:
1. Implementation of the protocol (not sharded)
2. New query type
3. Record structs and implt for the new query type row: probably not needed or should be changed. I had to implement them without knowing what they should actually be, in order to implement a new query type.

screenshot_20231018_140210


src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
+use std::ops::{Add, AddAssign};
```
martinthomson, line 1: I think that Daniel pointed this out, but this has nothing to do with the OPRF.  It should replace our existing shuffle.

danielmasny: I agree that this should be the shuffle that we use in both approaches. Alex and I discussed this. Alex seemed to prefer that we still put it in an OPRF related module even when it is a basic protocol (this is also the case for operations on Boolean shares). The main reason is that we have a more clear separation between old and new components in case we want to migrate away from the old IPA.

src/protocol/sort/apply_sort/mod.rs
```diff
@@ -5,12 +5,9 @@ pub use shuffle::shuffle_shares;
 use crate::{
     error::Error,
     protocol::{
-        basics::Reshare,
+        basics::{apply_permutation::apply_inv, Reshare},
         context::Context,
```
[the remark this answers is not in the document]
danielmasny: we could potentially leave it if we want to make the new shuffle compatible with radix sort at some point

src/query/runner/oprf_shuffle/share.rs
```diff
@@ -0,0 +1,244 @@
...
+    fn mul_assign(&mut self, rhs: Self) {
+        self.timestamp *= rhs.timestamp;
+        self.mk *= rhs.mk;
+        self.is_trigger_bit *= rhs.is_trigger_bit;
+        self.breakdown_key *= rhs.breakdown_key;
+        self.trigger_value *= rhs.trigger_value;
+    }
+}
+
+impl SharedValue for ShuffleShare {
```
danielmasny, line 96: great!

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    GenerateBHat,
+    GenerateZ,
+    TransferCHat,
+    TransferX2,
+    TransferY1,
+}
+
+/// # Errors
+/// Will propagate errors from transport and a few typecasts
+pub async fn shuffle<C, I, S>(
```
danielmasny, line 30: Awesome, now we can use the shuffle with any AdditiveShares of SharedValues!

src/query/runner/oprf_shuffle/share.rs
```diff
@@ -0,0 +1,244 @@
...
+        mk: ShuffleShareMK::ZERO,
+        is_trigger_bit: ShuffleShareF::ZERO,
+        breakdown_key: ShuffleShareBK::ZERO,
+        trigger_value: ShuffleShareF::ZERO,
+    };
+}
+
+impl ShuffleShare {
+    #[must_use]
+    pub fn from_input_row(input_row: &ShuffleInputRow, shared_with: Direction) -> Self {
```
danielmasny, line 116: I am not sure about this, could you explain me why we need these functions?

danielmasny (commented):
Awesome, thanks! Looks good to me (ignoring the query parts). I haven't checked any of the shuffling parts except that you are using sharedvalue. Is there a way to see only your changes with respect to my previous review?

src/protocol/oprf/shuffle/mod.rs
akoshelev: I don't think shuffle implementations are at the state where we can switch between them. The new one requires special input format (moving from struct { secret_share1, secret_share2, ... } to secret_share { struct1, struct2 }) while the old one requires them to be reshareable only. It would be ideal if shuffle is generalized over some sort of a trait, but I don't think it is atm.

So we may want to keep OPRF-specific things inside a different module

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    let a_add_b_iter = shares
+        .into_iter()
+        .map(|s: AdditiveShare<S>| s.left().add(s.right()));
+    let mut x_1: Vec<S> = add_single_shares(a_add_b_iter, z_12).collect();
+
+    let ctx_perm = ctx.narrow(&OPRFShuffleStep::ApplyPermutations);
+    let (mut rng_perm_l, mut rng_perm_r) = ctx_perm.prss_rng();
+    apply_permutation(&mut rng_perm_r, &mut x_1);
+
+    let mut x_2: Vec<S> = add_single_shares(x_1, z_31).collect();
```
akoshelev, line 85: can we use `x_1` allocation to store the result of this operation?

src/protocol/oprf/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod shuffle;
```
akoshelev, line 1: suggestion to  move `oprf/shuffle/mod.rs` to `oprf/shuffle.rs`

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+        let msg = receive_channel.receive(RecordId::from(record_id)).await?;
+        output.push(msg);
+    }
+
+    Ok(output)
+}
+
+// ------------------ Pseudorandom permutations functions -------------------- //
+
+fn apply_permutation<R: Rng, S>(rng: &mut R, items: &mut [S]) {
```
akoshelev, line 337: I see no reason for having this function

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+fn add_single_shares<A, B, S, L, R>(l: L, r: R) -> impl Iterator<Item = S>
+where
+    A: Add<B, Output = S>,
+    L: IntoIterator<Item = A>,
+    R: IntoIterator<Item = B>,
+{
+    l.into_iter().zip(r).map(|(a, b)| a + b)
+}
+
+fn add_single_shares_in_place<S, R>(mut items: Vec<S>, r: R) -> Vec<S>
```
akoshelev, line 227: this implies `()` as output of this function

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+async fn run_h2<C, I, S, Zl, Zr>(
+    ctx: &C,
+    batch_size: usize,
+    shares: I,
+    (z_12, z_23): (Zl, Zr),
+) -> Result<Vec<AdditiveShare<S>>, Error>
+where
+    C: Context,
+    I: IntoIterator<Item = AdditiveShare<S>>,
+    S: SharedValue + Add<Output = S> + Message,
```
akoshelev, line 103:
```suggestion
    S: SharedValue + Add<Output = S>,
```

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    rng.sample_iter(Standard).take(batch_size)
+}
+
+// ---------------------------- helper communication ------------------------------------ //
+
+async fn send_to_peer<C, S>(
+    ctx: &C,
+    step: &OPRFShuffleStep,
+    direction: Direction,
+    items: Vec<S>,
```
akoshelev, line 292: this contract causes extra clones of the size `N` on the caller's side. It is possible to avoid by taking an iterator of `S: SharedValue`.

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    let c_hat_2: Vec<S> = add_single_shares(y_3.iter(), a_hat.iter()).collect();
+    let ((), c_hat_1): ((), Vec<S>) = future::try_join(
+        send_to_peer(
+            ctx,
+            &OPRFShuffleStep::TransferCHat,
+            Direction::Left,
+            c_hat_2.clone(),
+        ),
+        receive_from_peer(
```
akoshelev, line 204: you should receive this to `y_3` allocation.

src/protocol/oprf/shuffle/mod.rs
martinthomson: This isn't OPRF-specific, it's just a different shuffle.  It's OK if we have two shuffles for now, but this shouldn't have to be bundled in with the OPRF.

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+            Direction::Left,
+        ),
+    )
+    .await?;
+
+    let mut x_3 = x_2;
+    add_single_shares_in_place(&mut x_3, z_23);
+    x_3.shuffle(&mut rng_perm_r);
+
+    let mut c_hat_1 = repurpose_allocation(y_1);
```
akoshelev, line 140: This may be cheap but it has a non-trivial cost of zeroing out memory. I don't see a need to do that, you can just replace elements in that vector from the iterator.

src/protocol/oprf/shuffle/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+
+async fn send_to_peer<C, S>(
+    items: &[S],
+    ctx: &C,
+    step: &OPRFShuffleStep,
+    direction: Direction,
+) -> Result<(), Error>
+where
+    C: Context,
+    S: Copy + Message,
```
akoshelev, line 312: please use `SharedValue`, `Message` is not an appropriate abstraction for it

akoshelev (commented):
I think this is very close. I pushed some polishing and added a TODO - lmk if they don't make sense to you

akoshelev (approved):
lgtm, I think we can address other things in a separate PR

PR #819, opened 2023-10-26
Cargo audit seems busted

martinthomson:


akoshelev:
I think it works now: https://github.com/private-attribution/ipa/actions/runs/6681656268/job/18156004970

we could switch to https://github.com/rustsec/audit-check if it happens again - I spent some time investigating it and I am not sure it is cargo audit to blame here

danielmasny:
outdated

PR #822, opened 2023-10-30
Real world test

richajaindce:

PR #823, opened 2023-10-31
Max out the number of rows for OPRF step generation

taikiy:
OPRF protocol step generation needed to run #822. `ipa::protocol::prf_sharding::UserNthPerStep` depends on the maximum number of rows of any users in the input. In order to generate all possible combinations of the steps, we need to generate the max allowable number of rows for a user. The max allowable number should be the number specified by `#[dynamic(...)]` in `UserNthRowStep` enum. For now, we use `records_per_user` which can be specified when running the benchmark.

Note that this is a temporary solution. If the max number of rows become too large, we need a better way to generate the steps, likely by synthesizing the steps in the `Step` derive macro.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -115,16 +115,10 @@ async fn run(args: Args) -> Result<(), Error> {
     // Note that this is a temporary solution. If the max number of rows become too
     // large, we need a better way to generate the steps, likely by synthesizing the
     // steps in the `Step` derive macro.
-    let (max_number_of_users, query_size) = if cfg!(feature = "step-trace") && args.oprf {
-        (
-            NonZeroU32::new(1).unwrap(),
-            usize::try_from(args.records_per_user).unwrap(),
-        )
+    let (single_user_mode, query_size) = if cfg!(feature = "step-trace") && args.oprf {
+        (true, usize::try_from(args.records_per_user).unwrap())
     } else {
-        (
-            EventGeneratorConfig::default().max_number_of_users,
-            args.query_size,
-        )
+        (false, args.query_size)
     };
 
     let seed = args.random_seed.unwrap_or_else(|| random());
@@ -139,7 +133,7 @@ async fn run(args: Args) -> Result<(), Error> {
             max_trigger_value: NonZeroU32::try_from(args.max_trigger_value).unwrap(),
             max_breakdown_key: NonZeroU32::try_from(args.breakdown_keys).unwrap(),
             max_events_per_user: NonZeroU32::try_from(args.records_per_user).unwrap(),
-            max_number_of_users,
+            single_user_mode,
             ..Default::default()
         },
     )
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -105,6 +105,28 @@ async fn run(args: Args) -> Result<(), Error> {
         ..TestWorldConfig::default()
     };
 
+    // OPRF protocol step generation.
+    // `ipa::protocol::prf_sharding::UserNthPerStep` depends on the maximum number
+    // of rows of any users in the input. In order to generate all possible
+    // combinations of the steps, we need to generate the max allowable number of
+    // rows for a user. The max allowable number should be the number specified
+    // by `#[dynamic(...)]` in `UserNthRowStep` enum. For now, we use
+    // `records_per_user` which can be specified when running the benchmark.
+    // Note that this is a temporary solution. If the max number of rows become too
+    // large, we need a better way to generate the steps, likely by synthesizing the
+    // steps in the `Step` derive macro.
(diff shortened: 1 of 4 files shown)


benches/oneshot/ipa.rs
```diff
@@ -105,6 +105,22 @@ async fn run(args: Args) -> Result<(), Error> {
...
+    // by `#[dynamic(...)]` in `UserNthRowStep` enum. For now, we use
+    // `records_per_user` which can be specified when running the benchmark.
+    // Note that this is a temporary solution. If the max number of rows become too
+    // large, we need a better way to generate the steps, likely by synthesizing the
+    // steps in the `Step` derive macro.
+    let (single_user_mode, query_size) = if cfg!(feature = "step-trace") && args.oprf {
+        (true, usize::try_from(args.records_per_user).unwrap())
+    } else {
+        (false, args.query_size)
+    };
```
benjaminsavage, line 122: Rather than use a boolean, you can use the IpaQueryStyle enum.

benches/oneshot/ipa.rs
```diff
@@ -117,10 +133,11 @@ async fn run(args: Args) -> Result<(), Error> {
             max_trigger_value: NonZeroU32::try_from(args.max_trigger_value).unwrap(),
             max_breakdown_key: NonZeroU32::try_from(args.breakdown_keys).unwrap(),
             max_events_per_user: NonZeroU32::try_from(args.records_per_user).unwrap(),
+            single_user_mode,
             ..Default::default()
         },
     )
-    .take(args.query_size)
+    .take(query_size)
```
benjaminsavage, line 140: This seems wrong... If you're running things in OPRF mode, and you specify a query_size, it will now be totally ignored? I think we want to use the oneshot test for more than just step generation, right?

benches/oneshot/ipa.rs
benjaminsavage: Oh wait, it looks like I misunderstood the meaning of "single_user_mode".

benches/oneshot/ipa.rs
taikiy: This is only the case when `cfg!(feature = "step-trace")` is true, so all other tests will be run as usual.

src/test_fixture/event_gen.rs
```diff
@@ -51,6 +51,13 @@ pub struct Config {
     pub report_filter: ReportFilter,
     #[cfg_attr(feature = "clap", arg(long, required_if_eq("report_filter", "TriggerOnly"), default_value = "0.02", value_parser = validate_probability))]
     pub conversion_probability: Option<f32>,
+    #[cfg_attr(
+        feature = "clap",
+        arg(long, default_value = "10", conflicts_with("single_user_mode"))
+    )]
+    pub number_of_users_in_flight: NonZeroU32,
+    #[cfg_attr(feature = "clap", arg(long, default_value = "false"))]
+    pub single_user_mode: bool,
```
benjaminsavage, line 60: Do we need both? Can we just use a single param for this?

src/test_fixture/event_gen.rs
```diff
@@ -51,6 +51,13 @@ pub struct Config {
     pub report_filter: ReportFilter,
     #[cfg_attr(feature = "clap", arg(long, required_if_eq("report_filter", "TriggerOnly"), default_value = "0.02", value_parser = validate_probability))]
     pub conversion_probability: Option<f32>,
+    #[cfg_attr(
+        feature = "clap",
+        arg(long, default_value = "10", conflicts_with("single_user_mode"))
+    )]
+    pub number_of_users_in_flight: NonZeroU32,
```
benjaminsavage, line 58: I recommend instead an interface of just:
```pub num_users```

File: src/test_fixture/event_gen.rs [hunk 5]
```diff
@@ -241,8 +254,12 @@ impl<R: Rng> Iterator for EventGenerator<R> {
     type Item = TestRawDataRecord;
 
     fn next(&mut self) -> Option<Self::Item> {
-        const USERS_IN_FLIGHT: usize = 10;
```
benjaminsavage, line 244: let users_in_flight = std::min(10, config.num_users);

src/test_fixture/event_gen.rs
```diff
@@ -241,8 +254,12 @@ impl<R: Rng> Iterator for EventGenerator<R> {
...
-        while self.users.len() < USERS_IN_FLIGHT {
+        let number_of_users_in_flight = if self.config.single_user_mode {
+            1
+        } else {
+            usize::try_from(self.config.number_of_users_in_flight.get()).unwrap()
+        };
+        while self.users.len() < number_of_users_in_flight {
```
benjaminsavage, line 262: Maybe we move this into the constructor, and make it possible to configure it. Then this "replace" logic can just maintain self.users at a constant length.

taikiy:
#834 fixes the issue I wanted to address with this PR. Closing

PR #824, opened 2023-10-31
Logging improvements

akoshelev:
See individual commits for details


akoshelev:
yep, after this change you could use `RUST_LOG` to control the debug output

```bash
RUST_LOG=ipa=TRACE cargo run ...
```

this will filter out the noise from Hyper and Axum

PR #825, opened 2023-10-31
Fix OPRF attribution and capping circuit

akoshelev:
It was using sequential join that limits number of futures driven concurrently to completion. This circuit requires all user-level circuits to be polled, so `parallel_join` is more appropriate

There may be other places where parallel join must be used, I did not check those, this got me to 10k rows.

```
     Running `target/debug/report_collector --network test_data/network.toml --input-file /tmp/events-10000.txt --output-file output.txt --disable-https oprf-ipa --max-breakdown-key 256 --per-user-credit-cap 16 --plaintext-match-keys`
2023-10-31T21:04:37.643906Z  INFO ipa::cli::verbosity: Logging setup at level info
2023-10-31T21:04:38.406898Z  INFO ipa::cli::playbook::ipa: Starting query for OPRF
2023-10-31T21:07:11.873166Z  INFO ipa::cli::playbook::ipa: Running IPA for 10000 records took 153.46518s
2023-10-31T21:07:11.873315Z  WARN report_collector: output.txt file exists, renaming to "output-e13qo"
2023-10-31T21:07:11.875135Z  INFO report_collector: IpaQueryConfig { per_user_credit_cap: 16, max_breakdown_key: 256, attribution_window_seconds: None, num_multi_bits: 3, plaintext_match_keys: true }
```

### DIFF
--- a/src/protocol/prf_sharding/mod.rs
+++ b/src/protocol/prf_sharding/mod.rs
@@ -29,7 +29,7 @@ use crate::{
         },
         BitDecomposed, Linear as LinearSecretSharing, SharedValue,
     },
-    seq_join::{seq_join, seq_try_join_all},
+    seq_join::seq_join,
 };
 
 pub mod bucket;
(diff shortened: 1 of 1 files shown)


benjaminsavage (approved):
Makes sense. I guess each user is totally independent of every other, so we can (in principle) evaluate these circuits in any order. I guess `parallel_join` expresses that concept more cleanly.

It's a little unfortunate that we have this hidden dependency between the logic here and details of the infrastructure though. We should look to try to resolve that. Asking for the records to be evaluated sequentially should (in theory) work. 

I guess the biggest issue here is that the buffer doesn't flush until we reach the total number of records, and that the infra layer is able to stall itself out by limiting the number of futures.

PR #826, opened 2023-11-01
modulus conversion with retention

martinthomson:
If you only want to convert some fields of a struct, this would allow you to do that.  This is incomplete in the sense that it produces tuples.  You might want to have a type that can be reconstructed from the converted and retained fields.  Maybe that could be part of this functionality also, but I'm leaving that for a future refactoring.


benjaminsavage (approved):
Thanks martin!

PR #827, opened 2023-11-01
Stream oprf ipa

benjaminsavage:
Converting the new OPRF approach of IPA to use streaming.

Can no longer separate out the aggregation stage into a separate function because streams are just super hard to use. But there was no other place in the code that we were calling that, so OK.

Also had to eliminate a bunch of tests (that arguably were not good tests) that were testing just the first step or just the second step in isolation.


benjaminsavage:
This branch was landed as a result of landing #829

benjaminsavage:
Actually it wasn't landed... how confusing.

PR #828, opened 2023-11-02
Hack to avoid stall

benjaminsavage:
There is an issue where sequential join limits the number of Futures that are spawned...

...this interacts poorly with our infra which is waiting for a sufficiently large batch to perform any communication...

...which leads to this deadlock where you're just stuck, and no progress is made...

This is a hack which allows the code to actually run for input sizes > 1000. I do not like this code, but I'll put it up here just to stimulate a conversation about how to move forward. I think I'd prefer a different solution, where we increase the number of futures that can spawn to more than the minimum batch size, so that progress can be made. 

- Running IPA for 10000 records took 27.914203637s
- Running IPA for 100000 records took 308.087942399s


benjaminsavage:
wrong base commit

PR #829, opened 2023-11-02
Hack to avoid stall

benjaminsavage:
There is an issue where sequential join limits the number of Futures that are spawned...

...this interacts poorly with our infra which is waiting for a sufficiently large batch to perform any communication...

...which leads to this deadlock where you're just stuck, and no progress is made...

This is a hack which allows the code to actually run for input sizes > 1000. I do not like this code, but I'll put it up here just to stimulate a conversation about how to move forward. I think I'd prefer a different solution, where we increase the number of futures that can spawn to more than the minimum batch size, so that progress can be made. 

- Running IPA for 10000 records took 27.914203637s
- Running IPA for 100000 records took 308.087942399s
- Running IPA for 1000000 records took 2865.641394015s

One more hack: the random events generator is making timestamps that are larger than 2^20 so I can't use the correct type.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -21,13 +21,13 @@ pub use basics::BasicProtocols;
 
 use crate::{
     error::Error,
-    ff::{Gf20Bit, Gf3Bit, Gf40Bit, Gf8Bit},
+    ff::{Gf3Bit, Gf40Bit, Gf8Bit, Gf32Bit},
 };
 
 pub type MatchKey = Gf40Bit;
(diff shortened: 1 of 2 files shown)


akoshelev:
I agree that collecting a stream just to pass it to `parallel_join` is suboptimal. We may want another combinator that works with streams. 

`seq_join` is not an appropriate join operation for circuits that are dependent on each other. If my understanding is correct, here we need all circuits to be running in parallel in order to make progress, so we need to make sure all of them are running.

akoshelev (approved):

PR #833, opened 2023-11-03
Prf eval two

danielmasny:
PR #795 with cleaned up history


akoshelev:
sanitizers are broken in #837, not related to this change

PR #830, opened 2023-11-02
Add OPRF root step to compact

akoshelev:
was getting an error running OPRF IPA with compact steps enabled

```
cannot narrow from the invalid step oprf_ipa
stack trace:
```

there is still another issue that I don't know how to fix yet

```
thread 'tokio-runtime-worker' panicked at src/protocol/prf_sharding/mod.rs:219:10:
Invalid state transition: 8815 -> row3
```

PR #834, opened 2023-11-06
make random event generator generate "min_events_per_user..=max_events_per_user" number of events

taikiy:
OPRF steps depend on the maximum number of rows of all given users. In order to generate steps for the `Compact` gate, we need to be able to specify exactly how many **user(s)** and **events** we want to generate from the `EventGenerator`.

The current `EventGenerator` randomly generates events between `[1..=max_events_per_user]` per user, for up to `max_user_id` users (the name is confusing. I renamed it to `user_count`). It allows us to control the total number of events to generate, but not the number of events per user.

This is a quick change to fixate the number of events users can produce. We lose the randomness, but for the testing purpose I don't think that is a big deal as the randomness doesn't follow a real-world distribution anyway. Ideally, callers of the generator should be able to pass in a distribution of events-per-user.

This fixes other issues mentioned in #830. There's another issue with the generated timestamps exceeding the current 20 bits limit. I'll make another PR for that.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -1,5 +1,5 @@
 use std::{
-    num::{NonZeroU32, NonZeroUsize},
+    num::{NonZeroU32, NonZeroU64, NonZeroUsize},
     time::Instant,
 };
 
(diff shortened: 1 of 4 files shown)


akoshelev:
Can you support both use cases by adding `min_events_per_user`? I don't want to give up on IPA functionality just yet, where having different number of events per user matter. It is likely the case for OPRF too.

benches/oneshot/ipa.rs
```diff
@@ -111,16 +111,36 @@ async fn run(args: Args) -> Result<(), Error> {
...
     let rng = StdRng::seed_from_u64(seed);
+    let (user_count, min_events_per_user, max_events_per_user, query_size) =
+        if args.oprf && cfg!(feature = "step-trace") {
+            // For the steps collection, OPRF mode requires a single user with the same number
+            // of dynamic steps as defined for `UserNthRowStep::Row`.
+            (
+                NonZeroU64::new(1).unwrap(),
+                NonZeroU32::new(64).unwrap(),
+                NonZeroU32::new(64).unwrap(),
+                64,
```
benjaminsavage, line 122: One cannot tell by looking at this how these numbers are used, and what they mean.

taikiy: Could you tell me which part is unclear? I think the tuple variable names are self-explanatory and the comment describes why we need 1 and 64. Maybe change the integer literals to constants?

PR #839, opened 2023-11-07
Remove compact gate tests from basic checks

akoshelev:
Its great that we run them twice, but it is probably [too much](https://github.com/private-attribution/ipa/actions/runs/6789188348/job/18458850996). Run them once as part of additional checks step

[Here is the step where we run it](https://github.com/private-attribution/ipa/blob/main/.github/workflows/check.yml#L140)


eriktaubeneck (approved):
lgtm

PR #840, opened 2023-11-08
Set timestamp upper limit in EventGenerator's config

taikiy:
Our protocols expect timestamps to be within a fixed bit length (i.e., Gf20Bit), but the event generator may generate events that are outside this bound. This change adds a new config parameter to bind the timestamps to be within a given value.

### DIFF
--- a/benches/oneshot/ipa.rs
+++ b/benches/oneshot/ipa.rs
@@ -129,7 +129,7 @@ async fn run(args: Args) -> Result<(), Error> {
                 args.query_size,
             )
         };
-    let raw_data = EventGenerator::with_config(
+    let mut raw_data = EventGenerator::with_config(
         rng,
         EventGeneratorConfig {
             user_count,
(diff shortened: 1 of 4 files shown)


benjaminsavage (commented):
I like this. I like how it simplifies things. I recommend we go one step further and eliminate the "num users in flight" concept now that we are no longer generating events in timestamp order.

taikiy:
> I like this. I like how it simplifies things. I recommend we go one step further and eliminate the "num users in flight" concept now that we are no longer generating events in timestamp order.

I just realized that the default value of the `user_count` is 1 trillion. If we remove the `USERS_IN_FLIGHT`, the generator will try to sample a user from 1 trillion - we'll almost never see source-trigger pairs for the same users in the test input.

benjaminsavage (approved):

benjaminsavage:
I still think it should now be possible to eliminate the USERS_IN_FLIGHT and just have a single "user in flight" at a given time. It can be one, not 10.

Can you please change timestamps to use GF20Bit now?

Issue #812, opened 2023-10-20
Field trait cleanup


danielmasny:
currently prss needs the field trait for the truncate function (i.e. generate a uniform value from a u128). This is counter intuitive since this has nothing to do with a mathematical field.

akoshelev:
As part of this task, we probably should also fix the PRSS and let it generate a given number of bytes, rather than constraining it to 128 bits

benjaminsavage:
Previously, there was a hack in the BooleanArray where the "try_from" and "truncate_from" functions would *hash* inputs rather than accepting the bits they were given.

I reverted this behavior in #853. This caused two tests to start failing (ABY share conversion and boolean_array_256 tests), so I disabled them.

Once this task is complete those tests should be re-enabled.

danielmasny:
https://github.com/private-attribution/ipa/pull/868 fixes the issue with truncate and reenables the test. Truncate needs to be removed from PRSS. Truncate does not lead to a secure PRSS for BA256 and any secret sharing for values of more than 128 bit (unless we use hashing or a PRG within truncate).

andyleiserson:
This is partially addressed (see the description of #877 for detail), but should remain open. I have marked some of the remaining cleanup with `TODO(812)` in the code.

PR #844, opened 2023-11-14
Boolean + Boolean Arrays + Basic Operations + Conversion to Fp25519

danielmasny:
adds support for Boolean Shares plus basic operations:
- addition, saturated addition
- subtraction, saturated subtraction
- comparison
- conversion to mod 2^k
- conversion to Fp25519


src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,264 @@
...
+        result.set(
+            i,
+            bit_adder(
+                ctx.narrow(&BitOpStep::from(i)),
+                record_id,
+                &v,
+                y.get(i).as_ref(),
+                carry,
+            )
+            .await?,
```
benjaminsavage, line 113: What makes this "low communication" exactly? It seems to me that this perfectly replicates existing code: https://github.com/private-attribution/ipa/blob/main/src/protocol/boolean/saturating_sum.rs#L131 without any additional communication reduction.

This still performs the multiplications per-bit in a loop, without bit-packing. So this will necessarily send one byte of communication per bit being added (just like the code currently does).

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+};
+
+#[cfg(all(test, unit_test))]
+#[derive(Step)]
+pub(crate) enum Step {
+    SaturatedSubtraction,
+    MultiplyWithCarry,
+}
+
+///Comparison operation
```
benjaminsavage, line 19: Add space

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    C: Context,
+    for<'a> &'a AdditiveShare<XS>: IntoIterator<Item = AdditiveShare<XS::Element>>,
+    YS: WeakSharedValue + CustomArray<Element = XS::Element>,
+    XS: WeakSharedValue + CustomArray + Field,
+    XS::Element: Field + std::ops::Not<Output = XS::Element>,
+{
+    let mut carry = AdditiveShare(XS::Element::ONE, XS::Element::ONE);
+    // we don't care about the subtraction, we just want the carry
+    let _ = subtraction_circuit(ctx, record_id, x, y, &mut carry).await;
+    Ok(carry)
```
benjaminsavage, line 40: slick! I like it.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    C: Context,
+    for<'a> &'a AdditiveShare<XS>: IntoIterator<Item = AdditiveShare<XS::Element>>,
+    YS: WeakSharedValue + CustomArray<Element = XS::Element>,
+    XS: WeakSharedValue + CustomArray + Field,
+    XS::Element: Field + std::ops::Not<Output = XS::Element>,
+{
+    let mut carry = AdditiveShare::<XS::Element>::ZERO;
+    // we don't care about the subtraction, we just want the carry
+    let _ = subtraction_circuit(ctx, record_id, x, y, &mut carry).await;
+    Ok(carry)
```
benjaminsavage, line 64: Very cool that you're able to offer both!

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+///non-saturated unsigned integer subtraction
```
benjaminsavage, line 67: add space

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    XS: WeakSharedValue + CustomArray + Field,
+    XS::Element: Field + std::ops::Not<Output = XS::Element>,
+{
+    let mut carry = AdditiveShare(XS::Element::ONE, XS::Element::ONE);
+    // we don't care about the subtraction, we just want the carry
+    let _ = subtraction_circuit(ctx, record_id, x, y, &mut carry).await;
+    Ok(carry)
+}
+
+///Comparison operation
```
benjaminsavage, line 43: add space

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    y: &AdditiveShare<YS>,
+) -> Result<AdditiveShare<XS>, Error>
+where
+    C: Context,
+    for<'a> &'a AdditiveShare<XS>: IntoIterator<Item = AdditiveShare<XS::Element>>,
+    YS: WeakSharedValue + CustomArray<Element = XS::Element>,
+    XS: WeakSharedValue + CustomArray + Field,
+    XS::Element: Field + std::ops::Not<Output = XS::Element>,
+{
+    let mut carry = AdditiveShare(XS::Element::ONE, XS::Element::ONE);
```
benjaminsavage, line 86: I'm a bit confused by this syntax:
```
AdditiveShare(XS::Element::ONE, XS::Element::ONE);
```
Doesn't that mean the (left, right) are *both* one for all 3 helpers? I guess that's one way to express a secret sharing of 1 in a field of size 2... Is that a correct interpretation of what's going on?

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    subtraction_circuit(ctx, record_id, x, y, &mut carry).await
+}
+
+///saturated unsigned integer subtraction
```
benjaminsavage, line 90: add space

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    )
+    .await?;
+
+    //carry computes carry=(x>=y)
+    //if carry==0 {all 0 array, i.e. Array[carry]} else {result}:
+    //compute (1-carry)*Array[carry]+carry*result =carry*result
+    let sat = AdditiveShare::<S>::expand(&carry)
+        .multiply(&result, ctx.narrow(&Step::MultiplyWithCarry), record_id)
+        .await?;
+    Ok(sat)
```
benjaminsavage, line 124: Once again, while this is a really nice thing to have (everything is all zeroes when you go negative) this isn't required for the current OPRF circuit, and we can save a multiplication by skipping this.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+        record_id,
+        x,
+        y,
+        &mut carry,
+    )
+    .await?;
+
+    //carry computes carry=(x>=y)
+    //if carry==0 {all 0 array, i.e. Array[carry]} else {result}:
+    //compute (1-carry)*Array[carry]+carry*result =carry*result
```
benjaminsavage, line 120: add spaces

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    let sat = AdditiveShare::<S>::expand(&carry)
+        .multiply(&result, ctx.narrow(&Step::MultiplyWithCarry), record_id)
+        .await?;
+    Ok(sat)
+}
+
+///subtraction using bit subtractor
+/// subtracts y from x, Output has same length as x (carries and indices of y too large for x are ignored)
+///implementing `https://encrypto.de/papers/KSS09.pdf` from Section 3.1/3.2
```
benjaminsavage, line 129: add spaces

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    Ok(result)
+}
+
+///bit subtractor
+///implementing `https://encrypto.de/papers/KSS09.pdf` from Section 3.1/3.2
+///output = x + !(c + y)
+///update carry to carry = ( x + carry)(!(y + carry)) + carry
+/// # Errors
+/// propagates errors from multiply
```
benjaminsavage, line 170: I don't want to replicate the code here: https://github.com/private-attribution/ipa/blob/main/src/protocol/boolean/saturating_sum.rs#L180

The current code has better comments at the very least.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+                compare_geq, compare_gt, integer_sat_sub, integer_sub,
+            },
+        },
+        rand::thread_rng,
+        secret_sharing::{replicated::semi_honest::AdditiveShare, SharedValue},
+        test_executor::run,
+        test_fixture::{Reconstruct, Runner, TestWorld},
+    };
+
+    ///testing correctness of Not
```
benjaminsavage, line 220: add space

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+
+            assert_eq!(result, <Boolean>::from(expected));
+
+            let result2 = world
+                .semi_honest(records.into_iter(), |ctx, x_y| async move {
+                    compare_geq::<_, BA64, BA64>(
+                        ctx.set_total_records(1),
+                        protocol::RecordId(0),
+                        &x_y[0],
+                        &x_y[0],
```
benjaminsavage, line 288: Add a comment here. I originally thought this was a mistake until I realized you were just exercizing the "equals" case.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct();
+            assert_eq!(result2, <Boolean>::from(true));
+        });
+    }
+
+    ///testing comparisons gt
```
benjaminsavage, line 299: `/// testing comparisons gt`

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+    #[test]
+    fn semi_honest_compare_gt() {
```
benjaminsavage, line 301: you're actually testing gte not gt

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct();
+            assert_eq!(result2, <Boolean>::from(false));
+        });
+    }
+
+    ///testing correctness of subtraction
```
benjaminsavage, line 346: add a space after the ///

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+        run(|| async move {
+            let world = TestWorld::default();
+
+            let mut rng = thread_rng();
+
+            let records: Vec<BA64> = vec![rng.gen::<BA64>(), rng.gen::<BA64>()];
+            let x = records[0].as_u128();
+            let y = records[1].as_u128();
+
+            let expected = ((x + 1 + u128::from(u64::MAX)) - y) % (1 + u128::from(u64::MAX));
```
benjaminsavage, line 358: I think this is another good opportunity to use the 
```
let z = 1_u128 << 64;
```

Then this line becomes:
```
let expected = ((x + z) - y) % z;
```

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
```diff
@@ -0,0 +1,439 @@
...
+        run(|| async move {
+            let world = TestWorld::default();
+
+            let mut rng = thread_rng();
+
+            let records = (rng.gen::<BA64>(), rng.gen::<BA32>());
+            let x = records.0.as_u128();
+            let y = records.1.as_u128();
+
+            let expected = ((x + 1 + u128::from(u64::MAX)) - y) % (1 + u128::from(u64::MAX));
```
benjaminsavage, line 420: same as above. define z to shorten.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+    GenerateSecretSharing,
+    IntegerAddBetweenMasks,
+    IntegerAddMaskToX,
+    RevealY,
+}
+
+/// share conversion from Boolean array of size n to integer mod 2^n
+/// See ABY3 (`https://eprint.iacr.org/2018/403.pdf`):
+///
+/// currently not used and also not tested
```
benjaminsavage, line 33: hmmmm....

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+/// new shares are H1: (-s, y), H2: (y, -r), H3: (-r,-s)
+/// r+s+y = r+s+x-r-s = x
+///
+/// unfortunately, we cannot exploit that x might have much lower bit length than the Arithmetic share
+/// by sampling masks r+s from the same bit length (and setting higher order bits to 0)
+/// this is due to the fact that y needs to be revealed (which requires to reveal at least 2 bits more
+/// than the bit length of x even when r+s has 0 as higher order bits, due to carries)
+/// r+s needs to mask these additional bits which cause additional carries that need to be masked as well
+/// # Errors
+/// Propagates Errors from Interger Subtraction and Partial Reveal
```
benjaminsavage, line 49: Integer

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+            sh_r.0 = <B as WeakSharedValue>::ZERO;
+        }
+        Role::H2 => {
+            sh_s.1 = <B as WeakSharedValue>::ZERO;
+        }
+        Role::H3 => {
+            sh_r.1 = <B as WeakSharedValue>::ZERO;
+            sh_s.0 = <B as WeakSharedValue>::ZERO;
+        }
+    }
```
benjaminsavage, line 92: I really find this difficult to follow and vet for accuracy and compare against the protocol description. I urge you to please re-write this without any mutable data structures. Please just initialize r and s based on the role.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+/// y{m} := rs{m} xor x{m} xor `carry_x`
+/// where x{m} is 0 and `carry_x` is carry{m-1} which contains information about x
+/// carry is defined as carry{i} = carry{i-1} xor (x{i-1} xor carry{i-1})(y{i-1} xor carry{i-1})
+/// for all i where x{i-1}=0: carry{i} = carry{i-1}y{i-1}
+/// therefore for j>=0,
+/// y{m+j} := rs{m+j} xor (`carry_x` * `product_(k=m)^(m+j-1)rs{k})`
+/// this will result in leakage (rs{255}=0, rs{254}=0)
+/// y{255} := `carry_x` * `product_(k=m)^(255)rs{k})`
+/// y{254} := `carry_x` * `product_(k=m)^(254)rs{k})`
+/// however, these terms are only non-zero when all rs{k} terms are non-zero
```
benjaminsavage, line 153: This explanation could use some wordsmithing to make it easier to understand.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+    record_id: RecordId,
+    x: &AdditiveShare<B>,
+) -> Result<AdditiveShare<Fp25519>, Error>
+where
+    C: Context,
+    for<'a> &'a AdditiveShare<B>: IntoIterator<Item = AdditiveShare<B::Element>>,
+    B: WeakSharedValue + CustomArray<Element = Boolean> + Field,
+{
+    // we generate r <- (r1,r2,r3) and redefine r = (0,0,r3), and s = (r1,0,0)
+    let mut sh_r: AdditiveShare<BA256> = ctx
```
benjaminsavage, line 169: Once again, I urge you to avoid mutability to make things less error prone and easier to follow.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+    match ctx.role() {
+        Role::H1 => sh_r.1 = <BA256 as WeakSharedValue>::ZERO,
+        Role::H2 => sh_r.0 = <BA256 as WeakSharedValue>::ZERO,
+        Role::H3 => (),
+    }
+
+    //generate s from r
+    let mut sh_s = AdditiveShare::<BA256>::ZERO;
+
+    //redefine (r, s): H1: ((0,0),(r1,0)), H2: ((0,r3),(0,0)), H3: ((r3,0),(0,r1))
```
benjaminsavage, line 188: please add spaces after // in all of these comments.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+    match ctx.role() {
+        Role::H1 => sh_r.0 = <BA256 as WeakSharedValue>::ZERO,
+        Role::H2 => sh_s.1 = <BA256 as WeakSharedValue>::ZERO,
+        Role::H3 => {
+            sh_r.1 = <BA256 as WeakSharedValue>::ZERO;
+            sh_s.0 = <BA256 as WeakSharedValue>::ZERO;
+        }
+    }
+
+    // addition r+s might cause carry, so we need 253 bits
```
benjaminsavage, line 198: This comment is not obviously connected to the code. I expected to see the number 253 appear somewhere.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+/// y{m+j} := rs{m+j} xor (`carry_x` * `product_(k=m)^(m+j-1)rs{k})`
+/// this will result in leakage (rs{255}=0, rs{254}=0)
+/// y{255} := `carry_x` * `product_(k=m)^(255)rs{k})`
+/// y{254} := `carry_x` * `product_(k=m)^(254)rs{k})`
+/// however, these terms are only non-zero when all rs{k} terms are non-zero
+/// this happens with probability 1/(2^(256-m)) which is negligible for small m
+/// # Errors
+/// Propagates Errors from Interger Subtraction and Partial Reveal
+#[cfg(all(test, unit_test))]
+async fn convert_to_fp25519<C, B>(
```
benjaminsavage, line 158: This is a lot of duplicate code between this and `async fn convert<C, B, A>`. I wonder if there's a better way here to support both without so much duplication.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -0,0 +1,361 @@
...
+        Role::H3 => Ok(AdditiveShare::<Fp25519>(
+            Fp25519::from(sh_r.0).neg(),
+            Fp25519::from(sh_s.1).neg(),
+        )),
+    }
+}
+
+/// inserts a smaller array into a larger
+/// allows conversion between Boolean Array types like 'BA64' and 'BA256'
+/// we don't use it right except for testing purposes
```
benjaminsavage, line 239: We don't use it _right_? What do you mean? Do we use it wrong?

benjaminsavage (changes requested):
I think it might be best to split this up into 2 diffs. The ABY stuff isn't tested or used - so that seems like a good candidate to split out into a follow-up PR. I'm mainly concerned about the duplication of code with the addition and subtraction code. I'd like to have just one implementation.

src/ff/boolean_array.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+//macro for implementing Boolean array, only works for a byte size for which Block is defined
+macro_rules! boolean_array_impl {
+    ( $modname:ident, $name:ident, $bits:expr, $bytes:expr, $one:expr ) => {
+        #[allow(clippy::suspicious_arithmetic_impl)]
+        #[allow(clippy::suspicious_op_assign_impl)]
+        mod $modname {
+            use hkdf::Hkdf;
+            use sha2::Sha256;
```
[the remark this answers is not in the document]
danielmasny: I use this in the truncate function

src/ff/boolean_array.rs
```diff
@@ -0,0 +1,408 @@
...
+                }
+
+                /// uses hashing in order to be compatible with larger array sizes
+                fn truncate_from<T: Into<u128>>(v: T) -> Self {
+                    let hk = Hkdf::<Sha256>::new(None, &v.into().to_le_bytes());
+                    let mut okm = [0u8; $bytes];
+                    //error invalid length from expand only happens when okm is very large
+                    hk.expand(&[], &mut okm).unwrap();
+                    <$name>::deserialize(&okm.into())
+                }
```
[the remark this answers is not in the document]
danielmasny: The issue is that we initially thought that shared values would be at most 128 bit values so the PRSS creates pseudo uniform 128 bit from which the pseudo uniform shares can be created. For Boolean arrays, we might have much larger values like 256bit, so we need to expand the uniform 128 bit to uniform 256 bits. The hash function here functions like a PRG. Since we already use HKDF, I wanted to reuse it. Alternatively, we could use any cryptographic PRG that expands a 128 bit seed to the size of the shared value.

src/ff/boolean_array.rs
```diff
@@ -0,0 +1,408 @@
...
+
+            impl TryFrom<u128> for $name {
+                type Error = crate::error::Error;
+
+                /// Fallible conversion from `u128` to this data type. The input value must
+                /// be at most `Self::BITS` long. That is, the integer value must be less than
+                /// or equal to `2^Self::BITS`, or it will return an error.
+                fn try_from(v: u128) -> Result<Self, Self::Error> {
+                    if u128::BITS - v.leading_zeros() <= Self::BITS {
+                        Ok(Self::truncate_from(v))
```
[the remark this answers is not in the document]
danielmasny: I this function is used nowhere, However I need to implement it for field in order to be able to use PRSS. I already opened an issue to fix this: https://github.com/private-attribution/ipa/issues/812

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,264 @@
...
+    //compute carry*Array[carry]+(1-carry)*result = result+carry(Array[carry]-result)
+    let carry_array = AdditiveShare::<S>::expand(&carry);
+    let sat = result.clone()
+        + carry_array
+            .multiply(
+                &(carry_array.clone() - result),
+                ctx.narrow(&Step::MultiplyWithCarry),
+                record_id,
+            )
+            .await?;
```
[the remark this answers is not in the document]
danielmasny: We can discuss this in our next meeting. I agree that for the attribution, there might be an easier way to do what you need. For a saturated addition in general, i.e. where you want to have the result of the addition, I don't see how to do it more efficiently.

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
danielmasny: Yes so this is just the straightforward depth N addition circuit with N bits of communication per party. The underlying structure is a Boolean Array (BA) and not a bit decomposed GF. We could at some point implement the parallel prefix adder to reduce the depth to log N at the cost of 3N bits of communication per party. So there is no communication saving to the previous addition it only changes the underlying structure from a bit decomposed GF to a BA.

When multiplying a single bit in a BA it gets converted into a bool which is also 1 byte. However, a BA is more compact in memory and it can also be send more efficiently to other parties than a bit decomposed GF. For example when revealing it or shuffling it.

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,264 @@
...
+    Ok(result)
+}
+
+///bit adder
+///implementing `https://encrypto.de/papers/KSS09.pdf` from Section 3.1
+///output = x + (c + y)
+///update carry to carry = ( x + carry)(y + carry) + carry
+/// # Errors
+/// propagates errors from multiply
+async fn bit_adder<C, S>(
```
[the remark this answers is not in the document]
danielmasny: I could add the support for BA in the exising code. However we did not want to change code that might make old IPA not compile anymore. We might split IPA old and new into different crates.

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,264 @@
...
+    record_id: RecordId,
+    x: &AdditiveShare<S>,
+    y: Option<&AdditiveShare<S>>,
+    carry: &mut AdditiveShare<S>,
+) -> Result<AdditiveShare<S>, Error>
+where
+    C: Context,
+    S: Field,
+{
+    let output = x + y.unwrap_or(&AdditiveShare::<S>::ZERO) + &*carry;
```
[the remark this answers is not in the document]
danielmasny: I think this is called a reborrow. I creates an immutable reference from a mutable reference.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
danielmasny: Yes! We need to set secret shared value (named carry) to 1.

src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_low_com.rs
danielmasny: I improved the comments, I have put BA vs GF2 on the agenda for our next meeting.

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
danielmasny: lets remove this for now since we dont use it and I dont even test it

src/ff/boolean_array.rs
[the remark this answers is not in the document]
benjaminsavage: I still do not understand. Let's schedule a 1:1 to talk through this in person.

src/ff/boolean_array.rs
```diff
@@ -0,0 +1,408 @@
...
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0
```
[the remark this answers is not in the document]
benjaminsavage: This is marked as resolved, but there are still not 256 values in this array.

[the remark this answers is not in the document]
benjaminsavage: Sorry, nevermind. I was looking at a particular commit... Github UI is confusing... I see now that you've updated it in a subsequent commit.

src/ff/ec_prime_field.rs
```diff
@@ -266,4 +288,17 @@ mod test {
...
+    /// test conversion BA256 to Fp25519
+    #[test]
+    fn test_ba256_to_fp25519() {
+        let mut rng = thread_rng();
+        let mut b = rng.gen::<BA256>();
+        b.set(255, Boolean::ZERO);
+        b.set(254, Boolean::ZERO);
+        b.set(253, Boolean::ZERO);
+        b.set(252, Boolean::ZERO);
+        b.set(251, Boolean::ZERO);
```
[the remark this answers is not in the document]
benjaminsavage: This still does not make sense to me.

Isn't the prime 2^255 - 19? It would seem that you only need to zero out the highest two bits in that case to ensure that the value being expressed is less than p. Why are you zeroing out the 5 highest bits?

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
benjaminsavage: I don't like this comment. Do you mean:

/// we don't use it except for testing purposes

Because this sounds like you're saying:

/// we use this code incorrectly, except in our tests, where it is used correctly.

src/secret_sharing/replicated/semi_honest/additive_share.rs
```diff
@@ -237,6 +248,71 @@ where
     }
 }
 
+///Implement `ArrayAccess` for `AdditiveShare` over `WeakSharedValue` that implements `ArrayAccess`
```
benjaminsavage, line 251: Please add a space between the /// and the comment

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,268 @@
...
+                        &x_y[0],
+                        &x_y[1],
+                    )
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct()
+                .as_u128();
+            assert_eq!((x, y, z, result), (x, y, z, expected));
```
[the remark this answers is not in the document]
danielmasny: this is for debugging purposes such that when result does not equal expected, I want to see what x,y,z is. Notice that x,y are uniform. Alex even suggested to put the seed for the prg in there for debugging test with random inputs. In this case x,y,z is sufficient for debugging purposes.

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
```diff
@@ -0,0 +1,264 @@
...
+                        &x_y[0],
+                        &x_y[1],
+                    )
+                    .await
+                    .unwrap()
+                })
+                .await
+                .reconstruct()
+                .as_u128();
+            assert_eq!((x, y, result), (x, y, expected));
```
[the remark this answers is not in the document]
danielmasny: I add x,y here for debugging purposes

src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
danielmasny: right now, we only use it for testing purposes. I forgot to write "now". We will need this functionality later for a secure quicksort where we need to add a counter to make each element unique.

src/ff/ec_prime_field.rs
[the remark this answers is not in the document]
martinthomson: This doesn't seem right to me either.  Bernstein does recommend fixing some bits to either zero or one, but as follows:

> Inside your program, to generate a 32-byte Curve25519 secret key, start by generating 32 secret random bytes from a cryptographically safe source: mysecret[0], mysecret[1], ..., mysecret[31]. Then do
>
>```
>     mysecret[0] &= 248;
>     mysecret[31] &= 127;
>     mysecret[31] |= 64;
>```

That amounts to clearing three low bits, clearing bit 256 (obviously), and setting bit 255.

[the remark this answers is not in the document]
benjaminsavage: @danielmasny - why was this comment marked as resolved? I do not see any changes or updated comments or response from you.

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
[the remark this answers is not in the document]
benjaminsavage: Daniel and I connected. We will address this in a follow up PR:
- Convert OPRF IPA to use BinaryArray
- Convert OPRF IPA to use this code for saturating sum
- Delete the old implementation of the one-bit-adder / subtractor

src/protocol/ipa_prf/boolean_ops/addition_low_com.rs
benjaminsavage: Resolved: This is low communication compared to another protocol that's not implemented.

benjaminsavage (approved):
Thanks for doing this Daniel, and for all these changes!

PR #846, opened 2023-11-16
OPRF IPA attribution window bug fix

taikiy:
It was a stupid mistake of mine. I initialized the very first row with a share of zero, so we always subtracted zero from the trigger event timestamp, causing the event to get ignored. The bug happens if the first row for a user is a source event and the succeeding trigger event's timestamp is > `ATTRIBUTION_WINDOW_SECONDS`.

### DIFF
--- a/src/protocol/ipa_prf/prf_sharding/mod.rs
+++ b/src/protocol/ipa_prf/prf_sharding/mod.rs
@@ -589,7 +589,9 @@ where
         // This is incorrect in the case that the CAP is less than the maximum value of "trigger value" for a single row
         // Not a problem if you assume that's an invalid input
         difference_to_cap: BitDecomposed::new(vec![Replicated::ZERO; TV::BITS as usize]),
-        source_event_timestamp: BitDecomposed::new(vec![Replicated::ZERO; TS::BITS as usize]),
+        source_event_timestamp: BitDecomposed::decompose(TS::BITS, |i| {
+            input_row.timestamp.map(|v| Gf2::truncate_from(v[i]))
+        }),
     }
 }
 
@@ -834,6 +836,8 @@ where
 
 #[cfg(all(test, unit_test))]
 pub mod tests {
+    use std::num::NonZeroU32;
+
     use super::{CappedAttributionOutputs, PrfShardedIpaInputRow};
     use crate::{
         ff::{Field, Fp32BitPrime, GaloisField, Gf2, Gf20Bit, Gf3Bit, Gf5Bit},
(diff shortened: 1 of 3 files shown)


src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1062,4 +1066,68 @@ pub mod tests {
             assert_eq!(result, &expected);
         });
     }
+
+    #[test]
+    fn semi_honest_aggregation_capping_attribution_with_attribution_window() {
+        // In OPRF, the attribution window's upper bound is excluded - i.e., [0..ATTRIBUTION_WINDOW_SECONDS).
```
benjaminsavage, line 1072: Once @danielmasny's PR lands (https://github.com/private-attribution/ipa/pull/844) we will have access to both "greater than" and "greater than or equal to" comparisons, and it need not be this way.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1062,4 +1066,68 @@ pub mod tests {
...
+                oprf_test_input_with_timestamp(234, true, 0, 5, 199), // tsΔ = 199, attributed to 12
+                /* Third User */
+                oprf_test_input_with_timestamp(345, false, 20, 0, 0),
+                oprf_test_input_with_timestamp(345, true, 0, 7, 100), // tsΔ = 100, attributed to 20
+                oprf_test_input_with_timestamp(345, false, 18, 0, 200),
+                oprf_test_input_with_timestamp(345, false, 12, 0, 300),
+                oprf_test_input_with_timestamp(345, true, 0, 7, 400), // tsΔ = 100, attributed to 12
+                oprf_test_input_with_timestamp(345, true, 0, 7, 499), // tsΔ = 199, attributed to 12
+                oprf_test_input_with_timestamp(345, true, 0, 7, 500), // tsΔ = 200, not attributed
+                oprf_test_input_with_timestamp(345, true, 0, 7, 700), // tsΔ = 400, not attributed
```
benjaminsavage, line 1097: Recommend reducing the trigger values from 7 down to something smaller like 3, so that capping does not kick in.

src/protocol/mod.rs
```diff
@@ -21,13 +21,13 @@ pub use basics::BasicProtocols;
 
 use crate::{
     error::Error,
-    ff::{Gf32Bit, Gf3Bit, Gf40Bit, Gf8Bit},
```
benjaminsavage, line 24: potentially need to rebase / merge

benjaminsavage (approved):

PR #852, opened 2023-11-22
Vectorization for prime fields

andyleiserson:
Semi-honest AdditiveShare can hold a vector of sharings instead of just one sharing. The semi-honest multiply can operate on these vectors.

There are some comments in `src/secret_sharing/mod.rs` that may help provide an overall picture.

## Notes

* This includes the changes from #908; once that is merged the diff will be smaller (but still big).
* Earlier versions introduced a `Gf2Array` type, I have removed it in favor of using the BA types.
* I did not end up merging `SharedValueArray` with `ArrayAccess` / `CustomArray`, because there's actually a significant difference between the two: protocol code shouldn't need to access individual elements of a `SharedValueArray` since the computation for each record is exactly the same.

## Performance

Results from `cargo bench --no-default-features -F enable-benches,descriptive-gate --bench criterion_arithmetic` (with additional benchmark changes not in this PR). Throughput measured in Melem/s (millions of multiplies per second).

|                   | 4096x256x8 | 16384x256x8 | 4096x256x32 | 4096x256x128 | 4096x256x512 |
|-------------------|------------|-------------|-------------|--------------|--------------|
| x1 ser            | 0.56       |             | 0.57        | 0.57         | 0.57         |
| x1 par3           | 0.45       |             | 0.46        | 0.47         | 0.47         |
| x1 par3 tri       | 1.11       |             | 1.12        | 1.09         | 1.02         |
| x1 par8 tri       | 0.96       |             | 1.03        | 1.07         | 1.17         |
| x32 ser           | 8.44       |             | 8.43        | 8.00         |              |
| x256 ser          | 13.6       |             | 13.1        | 11.7         |              |
| x1024 ser         | 12.5       |             | 11.6        | 8.44         |              |
| x32 par3          | 7.49       |             | 7.48        | 7.42         |              |
| x256 par3         | 14.6       |             | 13.3        | 12.2         |              |
| x1024 par3        | 15.5       |             | 14.3        | 9.79         |              |
| x32 par3 tri      | 20.1       | 20.8        | 18.6        | 18.4         |              |
| x256 par3 tri     | 36.6       | 37.8        | 36.3        | 36.0         |              |
| x1024 par3 tri    | 36.8       | 38.6        | 37.1        | 36.8         |              |
| x32 par8 tri      | 22.1       | 23.9        | 22.7        | 23.1         |              |
| x256 par8 tri     | 61.4       | 74.3        | 62.3        | 62.1         |              |
| x1024 par8 tri    | 51.5       | 88.9        | 51.4        | 51.2         |              |

Configs are Width x Depth x Active work.

xN: Vectorized N-wide
ser: Serial
parN: Parallel w/ N tokio worker threads
tri: TotalRecords::Indeterminate (no buffering in OrderingSender)

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -462,12 +462,25 @@ macro_rules! boolean_array_impl {
                     use crate::secret_sharing::replicated::ReplicatedSecretSharing;
                     let bits = AdditiveShare::new(ONE, ONE);
                     let iter = bits.into_iter();
+                    assert_eq!(iter.len(), $bits);
                     for (i, j) in iter.enumerate() {
                         if i == 0 {
                             assert_eq!(j, AdditiveShare::new(Boolean::ONE, Boolean::ONE));
                         } else {
                             assert_eq!(j, AdditiveShare::<Boolean>::ZERO);
                         }
+                    }
+                }
+
+                #[test]
+                fn iterate_secret_shared_boolean_array_len() {
+                    use crate::secret_sharing::replicated::ReplicatedSecretSharing;
+                    let bits = AdditiveShare::new(ONE, ONE);
+                    let mut iter = bits.into_iter();
+                    assert_eq!(iter.len(), $bits);
+                    for b in (0..$bits).rev() {
+                        iter.next().unwrap();
+                        assert_eq!(iter.len(), b);
                     }
                 }
 
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -8,7 +8,7 @@ use typenum::{U14, U2, U32, U8};
 use crate::{
     ff::{boolean::Boolean, ArrayAccess, Field, Serializable},
     protocol::prss::{FromRandom, FromRandomU128},
-    secret_sharing::{Block, SharedValue},
+    secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
 };
 
 /// The implementation below cannot be constrained without breaking Rust's
@@ -39,6 +39,12 @@ impl<'a> Iterator for BAIterator<'a> {
 
     fn next(&mut self) -> Option<Self::Item> {
         self.iterator.next().map(|v| Boolean::from(*v))
+    }
+}
+
+impl<'a> ExactSizeIterator for BAIterator<'a> {
+    fn len(&self) -> usize {
+        self.iterator.len()
     }
 }
 
@@ -95,6 +101,8 @@ macro_rules! boolean_array_impl_small {
 
         // TODO(812): remove this impl; BAs are not field elements.
         impl Field for $name {
+            const NAME: &'static str = stringify!($name);
+
             const ONE: Self = Self(bitarr_one!($bits));
 
             fn as_u128(&self) -> u128 {
@@ -152,6 +160,10 @@ macro_rules! boolean_array_impl_small {
             fn from_random_u128(src: u128) -> Self {
                 Field::truncate_from(src)
             }
+        }
+
+        impl FieldVectorizable<1> for $name {
+            type ArrayAlias = StdArray<$name, 1>;
         }
     };
 }
@@ -358,6 +370,10 @@ macro_rules! boolean_array_impl {
                 }
             }
 
+            impl Vectorizable<1> for $name {
+                type Array = StdArray<$name, 1>;
+            }
+
             impl std::ops::Mul for $name {
                 type Output = Self;
                 fn mul(self, rhs: Self) -> Self::Output {
@@ -394,7 +410,7 @@ macro_rules! boolean_array_impl {
             #[allow(clippy::into_iter_without_iter)]
             impl<'a> IntoIterator for &'a AdditiveShare<$name> {
                 type Item = AdditiveShare<Boolean>;
-                type IntoIter = ASIterator<BAIterator<'a>>;
+                type IntoIter = ASIterator<'a, $name>;
 
                 fn into_iter(self) -> Self::IntoIter {
                     self.iter()
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -272,6 +272,7 @@ macro_rules! boolean_array_impl {
             use super::*;
             use crate::{
                 ff::{boolean::Boolean, ArrayAccess, Expand, Serializable},
+                impl_shared_value_common,
                 secret_sharing::{
                     replicated::semi_honest::{ASIterator, AdditiveShare},
                     SharedValue,
@@ -317,6 +318,8 @@ macro_rules! boolean_array_impl {
                 type Storage = Store;
                 const BITS: u32 = $bits;
                 const ZERO: Self = Self(<Store>::ZERO);
+
+                impl_shared_value_common!();
             }
 
             impl_serializable_trait!($name, $bits, Store, $deser_type);
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -284,6 +284,11 @@ macro_rules! boolean_array_impl {
             #[derive(Clone, Copy, PartialEq, Eq, Debug)]
             pub struct $name(pub(super) Store);
 
+            impl $name {
+                #[cfg(all(test, unit_test))]
+                const STORE_LEN: usize = bitvec::mem::elts::<u8>($bits);
+            }
+
             impl ArrayAccess for $name {
                 type Output = Boolean;
                 type Iter<'a> = BAIterator<'a>;
@@ -427,12 +432,26 @@ macro_rules! boolean_array_impl {
 
             #[cfg(all(test, unit_test))]
             mod tests {
+                use proptest::{
+                    prelude::{prop, Arbitrary, Strategy},
+                    proptest,
+                };
                 use rand::{thread_rng, Rng};
 
                 use super::*;
 
-                // Only small BAs expose this via `Field`.
-                const ONE: $name = $name(bitarr_one!($bits));
+                impl Arbitrary for $name {
+                    type Parameters = <[u8; $name::STORE_LEN] as Arbitrary>::Parameters;
+                    type Strategy = prop::strategy::Map<
+                        <[u8; $name::STORE_LEN] as Arbitrary>::Strategy,
+                        fn([u8; $name::STORE_LEN]) -> Self,
+                    >;
+
+                    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+                        <[u8; $name::STORE_LEN]>::arbitrary_with(args)
+                            .prop_map(|arr| $name(Store::from(arr)))
+                    }
+                }
 
                 #[test]
                 fn set_boolean_array() {
@@ -444,43 +463,28 @@ macro_rules! boolean_array_impl {
                     assert_eq!(ba.get(i), Some(a));
                 }
 
-                #[test]
-                fn iterate_boolean_array() {
-                    let bits = ONE;
-                    let iter = bits.iter();
-                    for (i, j) in iter.enumerate() {
-                        if i == 0 {
-                            assert_eq!(j, Boolean::ONE);
-                        } else {
-                            assert_eq!(j, Boolean::ZERO);
+                proptest! {
+                    #[test]
+                    fn iterate_boolean_array(a: $name) {
+                        let mut iter = a.iter().enumerate();
+                        assert_eq!(iter.len(), $bits);
+                        while let Some((i, b)) = iter.next() {
+                            assert_eq!(bool::from(b), a.0[i]);
+                            assert_eq!(iter.len(), $bits - 1 - i);
                         }
                     }
-                }
 
-                #[test]
-                fn iterate_secret_shared_boolean_array() {
+                    #[test]
+                    fn iterate_secret_shared_boolean_array(a: AdditiveShare<$name>) {
                     use crate::secret_sharing::replicated::ReplicatedSecretSharing;
-                    let bits = AdditiveShare::new(ONE, ONE);
-                    let iter = bits.into_iter();
-                    assert_eq!(iter.len(), $bits);
-                    for (i, j) in iter.enumerate() {
-                        if i == 0 {
-                            assert_eq!(j, AdditiveShare::new(Boolean::ONE, Boolean::ONE));
-                        } else {
-                            assert_eq!(j, AdditiveShare::<Boolean>::ZERO);
+                        let mut iter = a.iter().enumerate();
+                        assert_eq!(iter.len(), $bits);
+                        while let Some((i, sb)) = iter.next() {
+                            let left = Boolean::from(a.left().0[i]);
+                            let right = Boolean::from(a.right().0[i]);
+                            assert_eq!(sb, AdditiveShare::new(left, right));
+                            assert_eq!(iter.len(), $bits - 1 - i);
                         }
-                    }
-                }
-
-                #[test]
-                fn iterate_secret_shared_boolean_array_len() {
-                    use crate::secret_sharing::replicated::ReplicatedSecretSharing;
-                    let bits = AdditiveShare::new(ONE, ONE);
-                    let mut iter = bits.into_iter();
-                    assert_eq!(iter.len(), $bits);
-                    for b in (0..$bits).rev() {
-                        iter.next().unwrap();
-                        assert_eq!(iter.len(), b);
                     }
                 }
 
--- a/ipa-core/src/helpers/gateway/send.rs
+++ b/ipa-core/src/helpers/gateway/send.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Borrow,
     marker::PhantomData,
     num::NonZeroUsize,
     pin::Pin,
@@ -52,7 +53,11 @@ impl GatewaySender {
         }
     }
 
-    pub async fn send<M: Message>(&self, record_id: RecordId, msg: M) -> Result<(), Error> {
+    pub async fn send<M: Message, B: Borrow<M>>(
+        &self,
+        record_id: RecordId,
+        msg: B,
+    ) -> Result<(), Error> {
         debug_assert!(
             self.total_records.is_specified(),
             "total_records cannot be unspecified when sending"
@@ -109,7 +114,7 @@ impl<M: Message> SendingEnd<M> {
     ///
     /// [`set_total_records`]: crate::protocol::context::Context::set_total_records
     #[tracing::instrument(level = "trace", "send", skip_all, fields(i = %record_id, total = %self.inner.total_records, to = ?self.channel_id.role, gate = ?self.channel_id.gate.as_ref()))]
-    pub async fn send(&self, record_id: RecordId, msg: M) -> Result<(), Error> {
+    pub async fn send<B: Borrow<M>>(&self, record_id: RecordId, msg: B) -> Result<(), Error> {
         let r = self.inner.send(record_id, msg).await;
         metrics::increment_counter!(RECORDS_SENT,
             STEP => self.channel_id.gate.as_ref().to_string(),
--- a/ipa-core/src/protocol/basics/mul/semi_honest.rs
+++ b/ipa-core/src/protocol/basics/mul/semi_honest.rs
@@ -8,8 +8,9 @@ use crate::{
         prss::SharedRandomness,
         RecordId,
     },
-    secret_sharing::replicated::{
-        semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing,
+    secret_sharing::{
+        replicated::semi_honest::AdditiveShare as Replicated, FieldSimd, SharedValueArray,
+        Vectorizable,
     },
 };
 
@@ -26,16 +27,16 @@ use crate::{
 /// ## Errors
 /// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
 /// back via the error response
-pub async fn multiply<C, F>(
+pub async fn multiply<C, F, const N: usize>(
     ctx: C,
     record_id: RecordId,
-    a: &Replicated<F>,
-    b: &Replicated<F>,
+    a: &Replicated<F, N>,
+    b: &Replicated<F, N>,
     zeros: MultiplyZeroPositions,
-) -> Result<Replicated<F>, Error>
+) -> Result<Replicated<F, N>, Error>
 where
     C: Context,
-    F: Field,
+    F: Field + FieldSimd<N>,
 {
     let role = ctx.role();
     let [need_to_recv, need_to_send, need_random_right] = zeros.work_for(role);
@@ -43,19 +44,26 @@ where
     zeros.1.check(role, "b", b);
 
     // Shared randomness used to mask the values that are sent.
-    let (s0, s1) = ctx.prss().generate(record_id);
+    let (s0, s1) = ctx
+        .prss()
+        .generate::<(<F as Vectorizable<N>>::Array, _), _>(record_id);
 
-    let mut rhs = a.right() * b.right();
+    let mut rhs = a.right_arr().clone() * b.right_arr();
+
     if need_to_send {
         // Compute the value (d_i) we want to send to the right helper (i+1).
-        let right_d = a.left() * b.right() + a.right() * b.left() - s0;
+        let right_d =
+            a.left_arr().clone() * b.right_arr() + a.right_arr().clone() * b.left_arr() - &s0;
 
-        ctx.send_channel(role.peer(Direction::Right))
-            .send(record_id, right_d)
+        ctx.send_channel::<<F as Vectorizable<N>>::Array>(role.peer(Direction::Right))
+            .send(record_id, &right_d)
             .await?;
         rhs += right_d;
     } else {
-        debug_assert_eq!(a.left() * b.right() + a.right() * b.left(), F::ZERO);
+        debug_assert_eq!(
+            a.left_arr().clone() * b.right_arr() + a.right_arr().clone() * b.left_arr(),
+            <<F as Vectorizable<N>>::Array as SharedValueArray<F>>::ZERO
+        );
     }
     // Add randomness to this value whether we sent or not, depending on whether the
     // peer to the right needed to send.  If they send, they subtract randomness,
@@ -65,9 +73,9 @@ where
     }
 
     // Sleep until helper on the left sends us their (d_i-1) value.
-    let mut lhs = a.left() * b.left();
+    let mut lhs = a.left_arr().clone() * b.left_arr();
     if need_to_recv {
-        let left_d = ctx
+        let left_d: <F as Vectorizable<N>>::Array = ctx
             .recv_channel(role.peer(Direction::Left))
             .receive(record_id)
             .await?;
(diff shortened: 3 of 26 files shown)


akoshelev:
> The first commit merges SharedValue into Field and then renames WeakSharedValue back to SharedValue. It could be its own PR: https://github.com/andyleiserson/ipa/commit/79f56589e2b78b3e5fb117c06c5565c5f29d1436.

great idea, we should do it because this change is needed on its own even if vectorization strategy changes

src/secret_sharing/array.rs
```diff
@@ -0,0 +1,245 @@
...
+impl<V: SharedValue, const N: usize> Sub<Self> for &StdArray<V, N> {
+    type Output = StdArray<V, N>;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        StdArray(
+            self.0
+                .iter()
+                .zip(rhs.0.iter())
+                .map(|(a, b)| *a - *b)
+                .collect::<Vec<_>>()
```
akoshelev, line 132: do we need a todo here?

src/secret_sharing/gf2_array.rs
```diff
@@ -0,0 +1,221 @@
...
+use generic_array::GenericArray;
+use typenum::U1;
+
+use crate::{
+    ff::{Gf2, Serializable},
+    secret_sharing::{SharedValue, SharedValueArray},
+};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Gf2Array<const N: usize>([usize; N]);
```
akoshelev, line 16: yea I agree, `const WORDS: usize` at least makes my brain stop and think before treating `N` as number of bits here.

src/secret_sharing/array.rs
```diff
@@ -0,0 +1,245 @@
...
+use typenum::U32;
+
+use crate::{
+    ff::{Field, Serializable},
+    helpers::Message,
+    secret_sharing::{SharedValue, SharedValueArray},
+};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StdArray<V, const N: usize>([V; N]);
```
akoshelev, line 17: is it worth exploring an idea of using one array type for both and adding interim structs that provide semantic to it?

apologies for stupid names, but something like `ByteValue<V, N>(Array<V, N>)` and `BitValue<WORDS>(Array<usize, WORDS>)`. It feels cleaner to implement `Add`/`Mul` on those rather than having two different array types

src/secret_sharing/mod.rs
```diff
@@ -51,44 +58,126 @@ pub trait Block: Sized + Copy + Debug {
...
+// similar. Doing so would require either (1) a generic impl of `Serializable` for any `N`, which
+// is hard to write, or (2) additional trait bounds of something like `F::Array<1>: Message`
+// throughout many of our protocols.
+//
+// Writing `impl<F: Field> Vectorized<1> for F` means that the compiler will always see that it
+// is available anywhere an `F: Field` trait bound is effective.
 
-    const BITS: u32;
+pub trait Vectorized<const N: usize>: SharedValue {
+    type Message: Message + Clone;
```
akoshelev, line 86: curious what drives the requirement for it to be `Clone`?

akoshelev (commented):
I think this is the right direction

src/secret_sharing/mod.rs
andyleiserson: Probably I had a `Clone` somewhere in one of my working implementations of vectorized multiply. I'll try to get rid of this.

src/secret_sharing/array.rs
andyleiserson: Are you proposing to implement `Add` and `Mul` on `Array`, or on `BitValue` / `ByteValue`?

The problem with implementing them on `Array` is that we want to specialize them to bitwise operations for GF2 and arithmetic operations for any other field.

If they are on `BitValue` and `ByteValue`, what functionality do you see `Array` having? The other significant functionality on arrays right now is serialization and `FromRandom`. Serialization could be done generically, but I'm not sure deserialization can be (if we want to reduce values mod the prime), nor FromRandom.

andyleiserson: I'm actually not sure that we need both Gf2Array and the boolean arrays -- but I haven't tried to reconcile them yet.

src/secret_sharing/array.rs
andyleiserson: I assume to avoid the allocation? Yes, I agree that is a good idea. I need to scrub all of the arithmetic impls. I have already changed a few to use `array::from_fn` instead of allocating.

danielmasny (commented):
I have not reviewed everything. I first wanted to understand the general approach better. I got a bit confused about sharedvaluearray and why we implement our own arraytype with operations on them. I was also wondering whether Gf2Array<N> would play a similar role to e.g. the BA40 or a Gf40bit type. See details below.

src/ff/curve_points.rs
```diff
@@ -25,18 +25,20 @@ impl Block for CompressedRistretto {
...
-#[derive(Clone, Copy, PartialEq, Debug)]
+// TODO: should not be Eq!
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
 pub struct RP25519(CompressedRistretto);
 
 /// Implementing trait for secret sharing
-impl WeakSharedValue for RP25519 {
+impl SharedValue for RP25519 {
     type Storage = CompressedRistretto;
+    type Array<const N: usize> = StdArray<RP25519, N>;
```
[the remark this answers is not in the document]
andyleiserson: We have to use our own array type in order to provide an implementation of the arithmetic operation traits -- the coherence rules don't let us implement them for `[RP25519; N]`.

I think the restriction imposed by the coherence rules here is actually a good thing -- the danger in being able to define a fundamental operation like `Add` for arrays is that not everyone may agree on what the operation means for arrays. If we wrap the array, then the behavior of the operations is part of the definition of our wrapper type.

src/ff/galois_field.rs
```diff
@@ -131,7 +131,7 @@ fn clmul<GF: GaloisField>(a: GF, b: GF) -> u128 {
 }
```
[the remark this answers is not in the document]
andyleiserson: Maintain compatibility is what I was going for here -- I don't foresee any immediate vectorization of the stuff that uses Galois fields larger than GF2, but if we want to keep the types in the codebase as `SharedValue`s, then they need to fulfill the trait definition.

(In my latest version, the `Array` associated type is no longer attached to `SharedValue`, for unrelated reasons -- but I still expect small changes to some of the types that new IPA doesn't use.)

src/ff/galois_field.rs
```diff
@@ -567,6 +568,7 @@ bit_array_impl!(
     bit_array_40,
     Gf40Bit,
     U8_5,
+    crate::secret_sharing::StdArray<Gf40Bit, N>,
```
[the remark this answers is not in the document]
andyleiserson: Implementing a parameterized `GfNBit` type for bit-level resolution of `N` is hard -- which is why the Galois fields and boolean arrays are implemented with macros.

But more fundamentally I agree with the point that there are use cases for both (N, 40) and (40, N). This is the horizontal vs. vertical vectorization distinction.

src/protocol/basics/mul/semi_honest.rs
```diff
@@ -65,34 +101,46 @@ where
...
 
     // Sleep until helper on the left sends us their (d_i-1) value.
-    let mut lhs = a.left() * b.left();
+    let mut lhs = FieldArray::<F, N>::mul_elements(a.left_arr(), b.left_arr());
     if need_to_recv {
-        let left_d = ctx
-            .recv_channel(role.peer(Direction::Left))
-            .receive(record_id)
-            .await?;
+        let left_d: F::Array<N> = <F as Vectorized<N>>::from_message(
```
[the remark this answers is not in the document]
andyleiserson: I don't think there's anything about 8-16 kB messages that makes them unusable. But I do think that the incremental gains of vectorization will mostly disappear at some fraction of that.

Right now, it looks like each supported vectorization config (data type and width) will need to be listed explicitly in at least in a macro invocation, which means it won't be possible for derived parameters to inadvertently drift outside the sensible range.

src/secret_sharing/array.rs
```diff
@@ -0,0 +1,245 @@
...
+        value.try_into().map(Self).map_err(|_| ())
+    }
+}
+
+impl<'a, 'b, V: SharedValue, const N: usize> Add<&'b StdArray<V, N>> for &'a StdArray<V, N> {
+    type Output = StdArray<V, N>;
+
+    fn add(self, rhs: &'b StdArray<V, N>) -> Self::Output {
+        //StdArray(self.0.iter().zip(rhs.0.iter()).map(|(a, b)| *a + *b).collect::<Vec<_>>().try_into().unwrap())
+        StdArray(array::from_fn(|i| self.0[i] + rhs.0[i]))
```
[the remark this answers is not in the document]
andyleiserson: This has the same role as the corresponding operation implemented on the BA* types. The BA types implement elementwise operations on an array of bits using bitwise AND, XOR. This is an implementation of elementwise operations on an array of Fp32BitPrime or Fp25519 or similar.

src/secret_sharing/gf2_array.rs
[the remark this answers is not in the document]
andyleiserson: I think it ought to be possible to use BAs rather than have a separate Gf2Array type. I didn't put particular thought into using `[usize; N]` here, I probably did it because it was closest to the `[T; N]` used by StdArray. Bitslices ought to do.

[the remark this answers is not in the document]
danielmasny: Do you plan to store a single bit in each usize then? or do you pack them and split them in usize chunks? We need access each bit individually but at the same time want to have a compact representation in memory and network messages.

src/secret_sharing/array.rs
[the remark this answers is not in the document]
danielmasny: I assumed that we would pass an array without boolean operations to a function and if we need to add within the function, we would implement this explicitly. I am wondering, if I e.g. add x,y,z together, this would need two iteration. Would a compiler be able to optimize this and merge it into a single iteration? At least for Fp25519 types, there are actually not that many places where we need basic operations like additions. However there are other operations like mapping them to curve points where we also would need to explicitly iterate over the vector within a the function.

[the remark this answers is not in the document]
danielmasny: related, does the compiler currently unroll all the loops and rearrange assembly instructions for cache optimization?

src/secret_sharing/array.rs
```diff
@@ -0,0 +1,245 @@
...
+    helpers::Message,
+    secret_sharing::{SharedValue, SharedValueArray},
+};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StdArray<V, const N: usize>([V; N]);
+
+// Introducing the `Eq` bound here is a wart. For some reason AdditiveShare is `Eq`, but
+// `SharedValue` is not. Probably all of these should be Eq, or none of them.
+impl<V: SharedValue + Eq, const N: usize> SharedValueArray<V> for StdArray<V, N> {
```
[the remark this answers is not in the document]
andyleiserson: There are a bunch of related issues here, and I'm not sure `SharedValueArray` is the key element of all of them -- I thought the major design decision was between something like `AdditiveShare<[T; N]>` and `AdditiveShare<T, N>`.

The purpose of `SharedValueArray` is to serve as the trait bound for the `Array` associated type. The associated type `SharedValue::Array` can't be constrained as a `SharedValue`, since that creates a cycle when trying to determine if something is a `SharedValue`.

In the current version, the `Array` associated type is no longer on `SharedValue`, so I guess it might be possible to merge the `SharedValueArray` type with `SharedValue`. I'm not sure I want to do that though -- it seems nice to have a distinction between "I'm working with a value" and "I'm working with a vector of values". There's also an analogous set of traits `Field` and `FieldArray` -- in that case there is the added concern that the array is not a field.

There is a broader issue of whether we want the APIs used to implement protocols to make vectorization explicit (`SharedValue<V, N>` and `*Array` traits that need to be referenced in vectorized code), or implicit (`SharedValue<[V; N]>` and the regular operations on `SharedValue` and `Field` become vectorized). I believe that making it explicit is less error-prone. Any time you use the API vectorization, it's definitely vectorization across records (vertical) -- and you can apply this vectorization to anything. Vectorization across values within a record (horizontal) needs to be coded explicitly in the protocol, since it requires knowledge of the structure of the data.

Some of this is also related to the question of what the differences are between BA*, Gf2Array, and GfNBit. I am hoping to have a proposal for that soon.

src/secret_sharing/array.rs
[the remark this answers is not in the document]
andyleiserson: Yes, the compiler will unroll these, map them to SIMD intrinsics, etc.

src/secret_sharing/array.rs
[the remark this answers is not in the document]
danielmasny: For the BA type, I was adding an Array trait (CustomArray, ArrayAccess). Would it be enough to distinguish a shared value of a single value from an array by checking whether the array trait has been implemented?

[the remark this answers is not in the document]
andyleiserson: This is a circuitous answer, but we might be able to merge `ArrayAccess` and `SharedValueArray` into a single trait, and then I would say that yes, whatever the merged trait is can distinguish arrays from single values.

A difference between `ArrayAccess` and `SharedValueArray` is that `SharedValueArray` inherits a bunch of the arithmetic operation traits. Right now I believe some arithmetic operations are invoked on BAs via `Field` -- if the BAs stop implementing `Field`, what trait do you think they should get the arithmetic operations from? (There are also some explicit `std::ops::Not` bounds in the PRF IPA protocols, but naming individual arithmetic operations everywhere we need them gets tedious.)

src/secret_sharing/mod.rs
andyleiserson: A couple things that make it hard to remove `Clone`:
* Currently the `send` function takes its argument by value, not by reference.
* We `clone` one of the arguments when what we really want is a ref-times-ref multiply.

Both of these are fixable, but not sure if we want to combine it with these changes.

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -66,6 +82,162 @@ pub trait SharedValue:
...
+}
+
+// The purpose of this trait is to avoid placing a `Message` trait bound on `SharedValueArray`, or
+// similar. Doing so would require either (1) a generic impl of `Serializable` for any `N`, which
+// is hard to write, or (2) additional trait bounds of something like `F::Array<1>: Message`
+// throughout many of our protocols.
+//
+// Writing `impl<F: Field> Vectorized<1> for F` means that the compiler will always see that it
+// is available anywhere an `F: Field` trait bound is effective.
+
```
andyleiserson, line 160:
```suggestion
```

This text is stale, disregard.

src/secret_sharing/mod.rs
akoshelev: would it make it easier for now for `send` to take `B: Borrow<Message>`?

src/secret_sharing/array.rs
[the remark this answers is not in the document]
danielmasny: Almost all arithmetic operations come from sharedvalue except multiplication (which currently comes from field) and std::ops::Not. I think it makes sense to bundle them all together in an existing trait. I am fine moving forward with SharedValueArray as a merge of the previous sharedvalue and arrayaccess/customarray. For the multiplication and maybe the not operation we might still need separate solutions since we might not want to implement them for all sharedvaluearrays. On the other side we could also optimize for bitarray since most of the operations are boolean arithmetics (so we could ask for sharedvaluearray to implement std::ops::Not).

ipa-core/src/protocol/basics/mul/semi_honest.rs
```diff
@@ -182,4 +201,145 @@ mod test {
...
+                                RecordId::from(MANYMULT_WIDTH * (i - 1) + j),
+                            ));
+                        }
+                        val = ctx.parallel_join(res).await.unwrap();
+                    }
+                    val
+                },
+            )
+            .await;
+        tracing::debug!("Protocol execution time: {:?}", begin.elapsed());
```
akoshelev, line 295: we should benchmark it again after #873 is merged

ipa-core/src/secret_sharing/array.rs
```diff
@@ -0,0 +1,422 @@
...
+
+use crate::{
+    ff::{Field, Fp32BitPrime, Serializable},
+    helpers::Message,
+    protocol::prss::{FromRandom, FromRandomU128},
+    secret_sharing::{FieldArray, SharedValue, SharedValueArray},
+};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StdArray<V: SharedValue, const N: usize>([V; N]);
```
akoshelev, line 19: can you please document why we need to wrap a standard array? I assume because we need to implement some traits, but it would be great to have it written somewhere

ipa-core/src/secret_sharing/array.rs
```diff
@@ -0,0 +1,422 @@
...
+impl<V: SharedValue, const N: usize> FromIterator<V> for StdArray<V, N>
+where
+    Self: Serializable,
+{
+    fn from_iter<T: IntoIterator<Item = V>>(iter: T) -> Self {
+        let mut res = Self::ZERO;
+        let mut iter = iter.into_iter();
+
+        for i in 0..N {
+            res.0[i] = iter.next().unwrap();
```
akoshelev, line 86:
```suggestion
            res.0[i] = iter.next().unwrap_or_else(|| panic!("Expected iterator to produce {N} items, got only {i+1}");
```

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -1,14 +1,62 @@
...
+//!     execution, etc. Vectorization can result in a greater amount of arithmetic work being
+//!     performed for a given amount of overhead work, thus increasing the efficiency of the
+//!     implementation.
+//!
+//! ## Vectorization traits
+//!
+//! There are two sets of traits related to vectorization.
+//!
+//! If you are writing protocols, the trait of interest is `FieldSimd<N>`, which can be specified in
+//! a trait bound, something like `F: Field + FieldSimd<N>`.
```
akoshelev, line 22: why `Field + FieldSimd<N>` and not just `FieldSimd<N>`?

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -57,13 +109,140 @@ pub trait Block: Sized + Copy + Debug {
...
+///     with a trait name.
+///  3. `F: Vectorizable<N>`. This is implied by the previous two, because `FieldArray`
+///     is a sub-trait of `SharedValueArray`.
+pub trait FieldVectorizable<const N: usize>: SharedValue + Sized {
+    type ArrayAlias: FieldArray<Self>;
+}
+
+// We could define a `SharedValueSimd` trait that is the analog of this for `SharedValue`s, but
+// there are not currently any protocols that need it.
+pub trait FieldSimd<const N: usize>:
```
akoshelev, line 191: I am not sure I understand the purpose of this trait, seems to exist for the purpose of grouping various trait bounds. Is it correct?

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -57,13 +109,140 @@ pub trait Block: Sized + Copy + Debug {
...
+}
+
+// Portions of the implementation treat non-vectorized operations as a vector with `N = 1`. This
+// blanket impl (and the fact that `F: Field` is the only trait bound) is important in allowing code
+// that writes `F: Field` to continue working without modification.
+impl<F: Field> FieldSimd<1> for F {}
+
+// Supported vectorizations
+
+impl FieldSimd<32> for Fp32BitPrime {}
```
akoshelev, line 203: clarifying question - is it for modulus conversion to be able to process all bits in parallel?

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -57,13 +109,140 @@ pub trait Block: Sized + Copy + Debug {
...
+    + Add<Self, Output = Self>
+    + for<'a> Add<&'a Self, Output = Self>
+    + AddAssign<Self>
+    + for<'a> AddAssign<&'a Self>
+    + Neg<Output = Self>
+    + Sub<Self, Output = Self>
+    + for<'a> Sub<&'a Self, Output = Self>
+    + SubAssign<Self>
+    + for<'a> SubAssign<&'a Self>
+    + Message
```
akoshelev, line 224: I would imagine the reverse to be a better model to avoid cyclic dependencies. Secret sharing is a core module and infrastructure stuff already depends on it.

akoshelev (dismissed):
This is pretty much good to go imo, I would just love to see a bit more documentation around arrays (sharedvalue, field).

src/secret_sharing/array.rs
akoshelev: for some reason I missed this message. The core idea was to avoid confusion and accidental misuse of `StdArray` as a trait bound for some things that do not require all the arithmetic, just because it is convenient and used everywhere.

Keeping the semantic of arrays as is (i.e. implementing traits for wrappers) feels cleaner but I don't think I strongly oppose the proposed approach

ipa-core/src/secret_sharing/replicated/semi_honest/additive_share.rs
```diff
@@ -86,138 +122,161 @@ where
...
+impl<'a, 'b, V: SharedValue + Vectorizable<N>, const N: usize> Add<&'b AdditiveShare<V, N>>
+    for &'a AdditiveShare<V, N>
+{
+    type Output = AdditiveShare<V, N>;
 
-    fn add(self, rhs: &'b AdditiveShare<V>) -> Self::Output {
-        AdditiveShare(self.0 + rhs.0, self.1 + rhs.1)
+    fn add(self, rhs: &'b AdditiveShare<V, N>) -> Self::Output {
+        AdditiveShare(
+            Add::add(self.0.clone(), &rhs.0),
```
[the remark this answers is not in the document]
andyleiserson: You may be thinking of the discussion about the clones in the `multiply` implementation, and the issue is the same -- this is dispatching via the `SharedValueArray` trait, which requires `Array: Sub<Array>` and `Array: Sub<&Array>`, but can't require `&Array: Sub<&Array>`, because it's not possible to express a bound on a reference to the type in a supertrait bound.

If we really wanted to avoid these clones, we could implement a custom trait that provides ref-by-ref arithmetic operations as an inherent method accepting two arguments (`trait MyAdd { fn add(a: &Self, b: &Self) -> Self; }`. I feel like the compiler is probably doing a pretty good job of optimizing these clones so that it may not be worth the trouble to do that, but I don't have solid data to back that up.

We could also drop the ref-by-ref impls here, which would make it more explicit in the caller when a clone is taking place.

ipa-core/src/secret_sharing/mod.rs
andyleiserson: It's an overlapping trait impl problem. There can't be wildcard `Message` impls for both `V: SharedValue` and `A: SharedValueArray`. (There's also a problem with the type parameter of `SharedValueArray` being unconstrained in that impl, but that might be resolvable by changing it to an associated type.)

One strategy would be to implement `Message` for anything that is serializable, i.e. `impl<T: Debug + Send + Serializable + 'static> Message for T {}`. That makes it easier to inadvertently create a channel for some type that wasn't really meant to be sent over the network -- not sure how serious a concern that is.

Another possibility would be moving the `Message` trait down to the bottom layer and implementing it explicitly on each applicable type (fields, BAs, etc.), like `Vectorizable`.

Or maybe implementing `SharedValue` for `SharedValueArray`, if we're okay getting rid of `trait Block` and associated type `SharedValue::Store`, neither of which is used very extensively.

ipa-core/src/secret_sharing/mod.rs
andyleiserson: Exactly. It is like a convenience trait alias. Meaning something like `trait MyTrait: LongListOfTraits {}` / `impl<T: LongListOfTraits> MyTrait for T`, which lets you write `T: MyTrait` instead of `T: LongListOfTraits`.

But in this case `LongListOfTraits` is not so much a long list of traits, as a short list of traits with a complicated associated type bound. But the point is the same -- without `FieldSimd`, lots of protocols would need to say `F: Field + Vectorizable<N, Array = <Self as FieldVectorizable<N>>::ArrayAlias> + FieldVectorizable<N>`.

ipa-core/src/secret_sharing/mod.rs
andyleiserson: We discussed today but to capture the answer, no, it's not to support any particular application, it's just the vectorization dimension I picked for the initial test cases in this PR. And, the vectorization dimension const parameter of `FieldSimd` being 32 is unrelated to the fact that it's `Fp32BitPrime` -- the dimension could be anything within reason.

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -143,7 +148,7 @@ impl GatewaySenders {
         match self.inner.entry(channel_id.clone()) {
             Entry::Occupied(entry) => (Arc::clone(entry.get()), None),
             Entry::Vacant(entry) => {
-                const SPARE: Option<NonZeroUsize> = NonZeroUsize::new(64);
+                const SPARE: Option<NonZeroUsize> = NonZeroUsize::new(256);
```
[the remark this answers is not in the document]
andyleiserson: Created https://github.com/private-attribution/ipa/pull/932 for this.

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -57,13 +108,150 @@ pub trait Block: Sized + Copy + Debug {
...
+impl FieldSimd<32> for Fp32BitPrime {}
+
+pub trait SharedValueArray<V>:
+    Clone
+    + Eq
+    + Debug
+    + Send
+    + Sync
+    + Sized
+    + TryFrom<Vec<V>, Error = ()>
```
akoshelev, line 227: oh this is probably a leftover, do we want actual error message here?

ipa-core/src/secret_sharing/mod.rs
akoshelev: do you mind if I just create another trait: `Sendable` and implement it for `SharedValue` and `SharedValueArray`? Then we can implement `Message` just for `Sendable` and be done with it. 

Any other hierarchy I feel may be confusing

andyleiserson: I'm not sure that is sufficient? `SharedValue` and `SharedValueArray` are traits, so `Sendable` would need to be implemented for `T: SharedValue(Array)?` , and the compiler will reject them as potentially overlapping like it does with `Message`.

If the idea is to add `Sendable` to core rather than moving `Message` to core, then maybe implementing it explicitly on fields, BAs, etc. rather than for a type parameter helps?

akoshelev: I did some experiments with `Sendable` here: https://github.com/private-attribution/ipa/commit/f1a3b5205e5331b1f4fe26bbc1e070a8fc5fb374

I think we can make it work relatively easy but the purpose of `Message` trait in this case becomes vague

andyleiserson: That seems like a reasonable approach to me. I don't think `Sendable` and `Message` are distinguished by purpose, but one is used in the protocol code and the other is used in infrastructure code. I tweaked your commit a little bit and added it to this branch.

ipa-core/src/protocol/basics/mul/semi_honest.rs
andyleiserson: I modified the arithmetic circuit benchmark to use seq_join, and ran a bunch configurations of the modified version. It seems like parallelization does not help this benchmark much (either with or without vectorization), and in some cases even makes it slightly slower.

[vectorization-perf-2.txt](https://github.com/private-attribution/ipa/files/14103387/vectorization-perf-2.txt)

(In all configurations, these results have PRSS used index tracking disabled, which gives a ~2x speed up compared to the results in the description. Increasing the vectorization width from 32 to 256 gives another ~50% on top of that.)

andyleiserson: It looks like this test case isn't using metrics because it is not using `TestWorld::semi_honest` or `TestWorld::malicious` (it is using `TestWorld`, but not the convenience wrapper for sharing/reconstructing I/O, which is what registers a tracing span). I wasn't running with `disable-metrics`, but adding it doesn't change the results. Turning off stall detection improves performance 5-10% in the novec-par case, doesn't make much difference in the vec-par case.

akoshelev: If turning off stall detection helps, it could be explained by contention on accessing send and receive channels.

danielmasny (approved):
I looks good to me.

ipa-core/src/ff/boolean_array.rs
```diff
@@ -95,6 +101,8 @@ macro_rules! boolean_array_impl_small {
 
         // TODO(812): remove this impl; BAs are not field elements.
```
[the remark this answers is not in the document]
andyleiserson: It will be resolved by the two branches we looked at on Friday:

https://github.com/andyleiserson/ipa/compare/bit-vectorization...andyleiserson:ipa:u128?expand=1
https://github.com/andyleiserson/ipa/compare/u128...andyleiserson:ipa:field-cleanup?expand=1

PR #853, opened 2023-11-22
Use BooleanArray in OPRF IPA

benjaminsavage:
Eliminates the use of BitDecomposed<Gf2> instead using Boolean Arrays as introduced in #844


src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -158,40 +143,40 @@ impl InputsRequiredFromPrevRow {
...
             ),
-            updated_sum.truncated_delta_to_saturation_point(
+            integer_sub(
                 ctx.narrow(&Step::ComputeDifferenceToCap),
                 record_id,
-                TV::BITS,
-            ),
+                &Replicated::<TV>::expand(Boolean::ZERO),
+                &updated_sum,
+            )
```
benjaminsavage, line 172: I think that this will do the same thing as what I had before...

The only time that I need to use the "difference_to_cap" is if this row is the one that put us over the limit. So I know that the delta to cap is going to fit in the number of bits used to express a trigger value (no need for the larger quantity used for the cumulative sum).

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -48,21 +44,21 @@ where
...
-#[cfg(all(test, unit_test))]
-pub async fn integer_sat_add<C, S>(
+pub async fn integer_sat_add<C, XS, YS>(
     ctx: C,
     record_id: RecordId,
-    x: &AdditiveShare<S>,
-    y: &AdditiveShare<S>,
-) -> Result<AdditiveShare<S>, Error>
+    x: &AdditiveShare<XS>,
+    y: &AdditiveShare<YS>,
```
danielmasny, line 51: Do you really need different lengths here? I didn't support different lengths since I thought that we probably don't need it.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -602,63 +584,58 @@ where
...
-                .enumerate()
-                .map(|(i, (cur_bit, prev_bit))| {
-                    let c = ctx.narrow(&BitOpStep::from(i));
-                    async move { if_else(c, record_id, is_trigger_bit, prev_bit, cur_bit).await }
-                }),
-        )
-        .await?,
-    ))
+    // expand is_trigger_bit to array
+    let is_trigger_bit_array = Replicated::<BK>::expand(&is_trigger_bit);
```
danielmasny, line 599: nice :D

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
benjaminsavage: Yes, I think we need different lengths.

In my current test I'm having the trigger values be 3-bit numbers, with a cap of 32 (5 bits). So I'm adding a smaller 3-bit value to the running total which is a 5-bit number.

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -73,17 +69,19 @@ where
...
         ctx.narrow(&Step::IfElse),
         record_id,
         &carry_array,
         &carry_array,
         &result,
     )
-    .await
+    .await?;
+
+    Ok((carry, saturated_bits))
```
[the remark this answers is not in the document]
benjaminsavage: oh, ok, you mean just move this code over to the other non-saturating function? Happy to do that.

src/ff/boolean.rs
```diff
@@ -14,6 +16,14 @@ impl Block for bool {
...
 
+impl ExtendableField for Boolean {
+    type ExtendedField = Gf32Bit;
+
+    fn to_extended(&self) -> Self::ExtendedField {
+        Gf32Bit::try_from(self.as_u128()).unwrap()
+    }
+}
+
 ///trait for secret sharing
```
martinthomson, line 26: Not your problem, but this doc comment is not very good.  Do comments on impl even have any effect?

src/ff/boolean_array.rs
```diff
@@ -138,13 +135,14 @@ macro_rules! boolean_array_impl {
...
                 fn truncate_from<T: Into<u128>>(v: T) -> Self {
-                    let hk = Hkdf::<Sha256>::new(None, &v.into().to_le_bytes());
-                    let mut okm = [0u8; $bytes];
-                    //error invalid length from expand only happens when okm is very large
-                    hk.expand(&[], &mut okm).unwrap();
-                    <$name>::deserialize(&okm.into())
+                    let v = v.into();
+                    let mut val = Self::ZERO;
+                    for i in 0..$bits {
+                        val.set(i, Boolean::from((v >> i & 1) == 1));
```
martinthomson, line 142: Much better :)

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -19,13 +19,12 @@ pub(crate) enum Step {
 /// adds y to x, Output has same length as x (carries and indices of y too large for x are ignored)
```
martinthomson, line 19: Can you update the documentation here?

The original here didn't use sentences (it should) and didn't describe what it returned.  That meant that you didn't need to update it, but we can do better.

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -183,7 +195,7 @@ mod test {
 
             let expected = (x + y) % (1_u128 << 64);
 
-            let result = world
+            let (_, result) = world
                 .semi_honest(records.into_iter(), |ctx, x_y| async move {
```
martinthomson, line 199: It is a little odd that we need to put the two BA64 values into a Vec then produce an iterator.  Is there some reason this can't use `(x, y)`?

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -195,9 +207,8 @@ mod test {
                     .unwrap()
                 })
                 .await
-                .reconstruct()
-                .as_u128();
-            assert_eq!((x, y, result), (x, y, expected));
+                .reconstruct();
+            assert_eq!((x, y, result.as_u128()), (x, y, expected));
```
martinthomson, line 211: Please check the carry as well.

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -34,7 +33,8 @@ where
     XS::Element: Field,
 {
     let mut carry = AdditiveShare::<XS::Element>::ZERO;
-    addition_circuit(ctx, record_id, x, y, &mut carry).await
+    let sum = addition_circuit(ctx, record_id, x, y, &mut carry).await?;
+    Ok((carry, sum))
```
martinthomson, line 37: The ordering of these two threw me here.  Most other addition functions return (sum, carry), see i8::carrying_add or any of the other standard library functions.

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -258,9 +269,8 @@ mod test {
...
                 })
                 .await
-                .reconstruct()
-                .as_u128();
-            assert_eq!((x, y, result), (x, y, expected));
+                .reconstruct();
+            assert_eq!((x, y, result.as_u128()), (x, y, expected));
         });
     }
 }
```
martinthomson, line 276: I see a test for `integer_add::<_, BA64, BA32>` but not `integer_add::<_, BA32, BA64>`.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -475,7 +471,7 @@ where
             .narrow(&Step::ModulusConvertBreakdownKeyBitsAndTriggerValues)
             .set_total_records(num_outputs),
         flattenned_stream,
-        0..BK::BITS + TV::BITS,
+        0..<BK as WeakSharedValue>::BITS + <TV as WeakSharedValue>::BITS,
```
martinthomson, line 470:
```suggestion
        0..(<BK as WeakSharedValue>::BITS + <TV as WeakSharedValue>::BITS),
```

Too complex to track now.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -602,63 +591,58 @@ where
...
-                            .map(|(i, (cur_bit, prev_bit))| {
-                                let c = ctx.narrow(&BitOpStep::from(i));
-                                async move {
-                                    if_else(c, record_id, is_trigger_bit, prev_bit, cur_bit).await
-                                }
-                            }),
-                    )
-                    .await?,
-            ))
+            // expand is_trigger_bit to array
```
martinthomson, line 631:
```suggestion
```
If you want a comment here, it would be: we need to replicate is_trigger_bit so that contains as many values as the timestamp.

But really, do we need that?

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -213,30 +196,42 @@ impl InputsRequiredFromPrevRow {
...
+            self.attributed_breakdown_key_bits.0.get(i).unwrap() == Boolean::ONE
         } else {
-            let i = i - self.attributed_breakdown_key_bits.len();
-            &self.capped_attributed_trigger_value[i]
+            let i = i - bk_bits;
+            self.capped_attributed_trigger_value.0.get(i).unwrap() == Boolean::ONE
         };
-        BitConversionTriple::new(role, bit.left() == Gf2::ONE, bit.right() == Gf2::ONE)
+        let bit_right = if i < bk_bits {
+            self.attributed_breakdown_key_bits.1.get(i).unwrap() == Boolean::ONE
```
martinthomson, line 225: Isn't this cleaner?

```suggestion
            self.attributed_breakdown_key_bits.right()[i] == Boolean::ONE
```

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -673,18 +657,22 @@ where
...
-    trigger_event_timestamp: &BitDecomposed<Replicated<Gf2>>,
-    source_event_timestamp: &BitDecomposed<Replicated<Gf2>>,
-) -> Result<BitDecomposed<Replicated<Gf2>>, Error>
+    trigger_event_timestamp: &Replicated<TS>,
+    source_event_timestamp: &Replicated<TS>,
+) -> Result<Replicated<TV>, Error>
 where
-    C: UpgradedContext<Gf2, Share = Replicated<Gf2>>,
+    C: Context,
+    TV: WeakSharedValue + CustomArray<Element = Boolean> + Field, // + ToBitConversionTriples,
```
martinthomson:
```suggestion
    TV: WeakSharedValue + CustomArray<Element = Boolean> + Field,
```

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -712,64 +700,62 @@ where
...
-                ctx.narrow(&Step::ComputeTimeDelta),
-                record_id,
-                source_event_timestamp,
-            )
-            .await?;
+        let mut constant_bits = <TS as WeakSharedValue>::ZERO;
+        for i in 0..<TS as WeakSharedValue>::BITS {
+            let bit = Boolean::from((attribution_window_seconds.get() >> i) & 1 == 1);
+            constant_bits.set(i.try_into().unwrap(), bit);
+        }
```
martinthomson, line 742: This looks like a `From` implementation or similar conversion function that belongs on TS.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -792,47 +778,45 @@ where
...
-                            ),
-                        )
-                        .await?;
-                        Ok::<_, Error>(not_saturated_case + &just_saturated_case)
-                    }
-                }),
-        )
-        .await?,
-    ))
+    // expand is_saturated to array
```
martinthomson, line 792: See previous comments.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -953,64 +941,44 @@ pub mod tests {
...
+            let capped_attributed_tv = [
+                s0.capped_attributed_trigger_value.clone(),
+                s1.capped_attributed_trigger_value.clone(),
+                s2.capped_attributed_trigger_value.clone(),
+            ]
+            .reconstruct();
+
+            let mut bd_key_decimal = 0_u128;
+            for i in 0..<BK as WeakSharedValue>::BITS {
+                bd_key_decimal += bk_key_bits.get(i.try_into().unwrap()).unwrap().as_u128() << i;
```
martinthomson, line 964: Implementing `Index` for u32 and usize seems like it might go a long way here.

src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -953,64 +941,44 @@ pub mod tests {
...
+            }
+
+            let mut tv_decimal = 0_u128;
+            for i in 0..<TV as WeakSharedValue>::BITS {
+                tv_decimal += capped_attributed_tv
```
martinthomson, line 969: Don't you want to iterate over the reconstructed values like this:

```rust
let value = value_bits.into_iter().enumerate().fold(0, |b, (i, v)| b | u128::from(v) << i);
```
I would use "breakdown_key_bits" and "value_bits" for the variables too.  We already know that they are capped attribution outputs.  Similarly for the field names.
```

File: src/protocol/ipa_prf/prf_sharding/mod.rs [hunk 20]
```diff
@@ -1020,16 +988,16 @@ pub mod tests {
...
 
-            let records: Vec<PreShardedAndSortedOPRFTestInput<Gf5Bit, Gf3Bit, Gf20Bit>> = vec![
+            let records: Vec<PreShardedAndSortedOPRFTestInput<BA5, BA3, BA20>> = vec![
                 /* First User */
                 oprf_test_input(123, false, 17, 0),
                 oprf_test_input(123, true, 0, 7),
                 oprf_test_input(123, false, 20, 0),
                 oprf_test_input(123, true, 0, 3),
-                /* Second User */
+                // /* Second User */
```
martinthomson, line 994: Check comment style.

src/ff/boolean_array.rs
benjaminsavage: But this is definitely needed somewhere else. It shouldn't be here for sure though. I'm going to assume @danielmasny is working on moving this into a different location.

src/protocol/ipa_prf/prf_sharding/mod.rs
benjaminsavage: I agree it would be... but there is no indexing operation available.

src/protocol/ipa_prf/prf_sharding/mod.rs
benjaminsavage: Actually, I should just call "as_u128()" here and skip all of this nonsense.

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
benjaminsavage: Added a test for this! Luckily it passes =).

src/protocol/ipa_prf/prf_sharding/mod.rs
martinthomson: Maybe you can interpret the comment as a request to add `impl Index<usize> for $name` to the macro then.

benjaminsavage:
Woah! Vectorization FTW! This approach has already reduced the number of "Steps" in the steps.txt file by 4,851! There are only 4324 steps related to the OPRF sharded approach, a more than 50% reduction! 

Even better, there are only 63 steps per *row*. Once the new infra eliminates the need to scope these steps by the *row* there will be another massive reduction.

ipa-core/src/ff/boolean_array.rs
```diff
@@ -255,19 +253,19 @@ macro_rules! boolean_array_impl {
...
-                }
+                // #[test]
+                // fn set_boolean_array() {
+                //     let mut rng = thread_rng();
+                //     let i = rng.gen::<usize>() % usize::try_from(<$name>::BITS).unwrap();
+                //     let a = rng.gen::<Boolean>();
+                //     let mut ba = rng.gen::<$name>();
+                //     ba.set(i, a);
+                //     assert_eq!(ba.get(i), Some(a));
+                // }
```
martinthomson, line 268: I think that the accepted approach for this is [`#[ignore]`](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested)

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -258,9 +283,32 @@ mod test {
...
-                .reconstruct()
-                .as_u128();
-            assert_eq!((x, y, result), (x, y, expected));
+                .reconstruct();
+            assert_eq!(
+                (x, y, result.as_u128(), carry.as_u128()),
+                (x, y, expected, expected_carry)
+            );
+
+            let x = x & ((1 << 32) - 1);
```
martinthomson, line 292: Isn't this a second test?

martinthomson (approved):
If the tests are simply failing, put them back with `#[ignore]`.  If they don't even compile, we'll need to work out what to do to restore them, or if they even can be ignored.

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
benjaminsavage: It's still testing addition of different lengths, just in the opposite order.

ipa-core/src/protocol/boolean/saturating_sum.rs
```diff
@@ -1,340 +0,0 @@
...
-
-            output_sum.push(sum_bit);
-        }
-        let is_saturated = or(
-            ctx.narrow(&BitOpStep::from(self.sum.len())),
-            record_id,
-            &carry_in,
-            &self.is_saturated,
-        )
-        .await?;
```
benjaminsavage, line 60: Whoops. This was important functionality which was lost. This is the cause of issue #863

src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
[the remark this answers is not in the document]
danielmasny: Yes, if you don't need the saturated addition, just the bit that indicates an overflow, the other one will be cheaper.

src/ff/boolean_array.rs
danielmasny: Is this the truncate function? It works this way however it does not lead to a secure PRSS/Multiply for arrays of more than 128 bits. We need to fix this as indicated in https://github.com/private-attribution/ipa/issues/812.

PR #854, opened 2023-11-23
Removing test config for all functions needed for convert_to_fp25519

richajaindce:


src/protocol/sort/mod.rs
```diff
@@ -29,7 +29,7 @@ pub(crate) enum SortStep {
 }
 
 #[derive(Step, Clone, Copy)]
-pub(crate) enum ShuffleStep {
+pub enum ShuffleStep {
```
richajaindce, line 32: Not sure why I needed to move it out of the crate but if I dont, I get an error
#[tracing::instrument(name = "shuffle_once", skip_all, fields(to = ?shuffle_for_helper(which_step)))]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private type

akoshelev: you can make `shuffle_once` function `pub(crate)` and this error should go away

akoshelev (approved):

PR #855, opened 2023-11-23
Quicksort

danielmasny:
not ready to merge yet. Sort works, however not clear how to best handle associated data, I have set it to () for now


.github/workflows/check.yml
```diff
@@ -141,8 +141,6 @@ jobs:
 
   # sanitizers currently require nightly https://github.com/rust-lang/rust/issues/39699
   sanitize:
-    # disabled due to https://github.com/private-attribution/ipa/issues/837
```
akoshelev, line 144: this seems to be from another PR (#842)

danielmasny: I will clean this up, I think it was included through a merge

danielmasny:
Ok it is ready for review now. quicksort wants a key extraction function similar to sort_by_key. The implementation is insecure for non-unique keys (since non-unique elements increase the expected running time). Further, it does not include shuffle. Quicksort on "unshuffled" data leaks information about the data.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+) -> Result<(), Error>
+where
+    C: Context,
+    S: Send + std::marker::Sync,
+    F: FnMut(&S) -> &AdditiveShare<K> + std::marker::Send + Clone + std::marker::Sync,
+    for<'a> &'a AdditiveShare<K>: IntoIterator<Item = AdditiveShare<K::Element>>,
+    K: WeakSharedValue + Field + CustomArray<Element = Boolean>,
+{
+    // create stack
+    let mut stack: Vec<(C, usize, usize)> = vec![];
```
akoshelev, line 61: you can likely predict the depth of the circuit here, so `Vec::with_capacity(...)` should be better here

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+            // precompute comparison against pivot and reveal result in parallel
+            let comp = seq_join(
+                ctx.active_work(),
+                stream_iter(iterator.enumerate().map(|(n, x)| {
+                    let mut fc = f.clone();
+                    async move {
+                        // compare current element against pivot
+                        let sh_comp = compare_gt(
+                            pctx.narrow(&Step::Compare)
+                                .set_total_records(b_r - (b_l + 1)),
```
akoshelev, line 86: I suspect that in practice this will be very inefficient because it recursively creates steps with tiny number of records to be sent through each. Is there a way to use a finite (const) number of steps for quicksort?

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+pub async fn quicksort_by_key_insecure<C, K, F, S>(
+    ctx: C,
+    list: &mut [S],
+    desc: bool,
+    f: F,
+) -> Result<(), Error>
+where
+    C: Context,
+    S: Send + std::marker::Sync,
+    F: FnMut(&S) -> &AdditiveShare<K> + std::marker::Send + Clone + std::marker::Sync,
```
akoshelev, line 56: `FnMut` in `Vec` is used to allow more closure types, including those that change the captured environment. This is probably not the case here, so if `Fn` makes your life easier here, we can stick with it

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+            let mut iterator = list[b_l..b_r].iter();
+            // first element is pivot, apply key extraction function f
+            let pivot = f.clone()(iterator.next().unwrap());
+            // create pointer to context for moving into closure
+            let pctx = &ctx;
+            // precompute comparison against pivot and reveal result in parallel
+            let comp = seq_join(
+                ctx.active_work(),
+                stream_iter(iterator.enumerate().map(|(n, x)| {
+                    let mut fc = f.clone();
```
akoshelev, line 81: wondering if you could just move a pointer inside `async`

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+                            .reveal(
+                                pctx.narrow(&Step::Reveal)
+                                    .set_total_records(b_r - (b_l + 1)),
+                                RecordId::from(n),
+                            )
+                            .await
+                    }
+                })),
+            )
+            .try_collect::<Vec<_>>()
```
akoshelev, line 105: there is likely an optimization to keep those bits inside a bitvec and/or reuse it between iterations

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        let mut list_mut = list.to_vec();
+        quicksort_by_key_insecure(ctx, &mut list_mut[..], desc, |x| x).await?;
+        let mut result: Vec<AdditiveShare<BA64>> = vec![];
+        list_mut.iter().for_each(|x| result.push(x.clone()));
+        Ok(result)
+    }
+
+    #[test]
+    fn test_quicksort_insecure_semi_honest() {
```
akoshelev, line 166: do we also want to test on edge cases (all equal, sorted in opposite direction, empty list) etc?

akoshelev (commented):
Did a first pass, haven't looked at everything closely, maybe others can chime in too

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,210 @@
...
+            // set up iterator
+            let mut iterator = list[b_l..b_r].iter();
+            // first element is pivot, apply key extraction function f
+            let pivot = f.clone()(iterator.next().unwrap());
+            // create pointer to context for moving into closure
+            let pctx = &ctx;
+            // precompute comparison against pivot and reveal result in parallel
+            let comp = seq_join(
+                ctx.active_work(),
+                stream_iter(iterator.enumerate().map(|(n, x)| {
```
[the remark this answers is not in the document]
danielmasny: so I should only use seq_join when working with streams? For iterators there is no advantage over try_join?

ipa-core/src/protocol/ipa_prf/quicksort.rs
danielmasny: the minimum length with be log length of input slice, in the worst case it will be length of the input slice (which is unlikely to happen). Is it still better to use vec::with_capacity even if I need to allocate exponentially more memory (log N vs N)? It probably depends on N which is the amount of events per users which we estimated to be at most 100k. so it would be 2*log_2(100k)< 34 vs 100k elements (notice that this stack implements depth first tree transgression)

ipa-core/src/protocol/ipa_prf/quicksort.rs
danielmasny: This is related to the issue above. It is unclear how the pivot will split the slice into two halves. In the best case, it will be evenly split. We could do a breadth first tree transgression and execute the comparisons against the different pivots in parallel. Even in this case the amount of records will be input array length - amount of pivots (i.e. amount of nodes of this level of the tree). The implementation would be more involved and harder to follow.

ipa-core/src/protocol/ipa_prf/quicksort.rs
danielmasny: Using a bitvec here by unwrapping Boolean(bool) would make sense to save some memory costs. We would need to generate bitvec with lengths that are only known at run time. We wouldn't save any communication costs since these bits are not used in any MPC computation. Further, these bits are not reused across iterations but we could store them in the same memory location.

ipa-core/src/protocol/ipa_prf/quicksort.rs
danielmasny: added all these test cases

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,356 @@
...
+                            // apply key extraction function f to element x
+                            pf(x),
+                            pivot,
+                        )
+                        .await?;
+
+                        // reveal outcome of comparison
+                        sh_comp
+                            .reveal(
+                                pctx.narrow(&Step::Reveal)
```
benjaminsavage, line 91: I think it'd be better to move these calls to `set_total_records` up, outside of the iterator. They only need to be set once per recursion. Subsequent calls to `narrow` will inherit the number of records.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,356 @@
...
+        if b_l + 1 < b_r {
+            // set up iterator
+            let mut iterator = list[b_l..b_r].iter();
+            // first element is pivot, apply key extraction function f
+            let pivot = f(iterator.next().unwrap());
+            // create pointer to context for moving into closure
+            let pctx = &ctx;
+            let pf = &f;
+            // precompute comparison against pivot and reveal result in parallel
+            let comp = seq_join(
```
benjaminsavage, line 73: I don't understand why you're using seq_join here instead of a parallel join. As far as I can tell, there is no dependency between these computations, so they should all be able to be done in parallel.

danielmasny: To my understanding seq_join is preferred when operating in parallel over multiple records of the same step while parallel join is preferred when operating in parallel over multiple steps of the same record. In this example, I need to finish a specific step (recursion) for all events (records) before I can move on to the next step (recursion). Alex suggested using try_join, I will sync with him.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -0,0 +1,356 @@
...
+            .try_collect::<Vec<_>>()
+            .await?;
+
+            // swap elements based on comparisons
+            // i is index of first element larger than pivot
+            let mut i = b_l + 1;
+            for j in b_l + 1..b_r {
+                // desc = true will flip the order
+                if comp[j - (b_l + 1)] == Boolean::from(false ^ desc) {
+                    list.swap(i, j);
```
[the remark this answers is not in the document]
danielmasny: This is a bit confusing since I wanted to do all comparisons in parallel and then in a second step reorder the elements based on the comparison. Element list[b_l] is the pivot element and comp[b_l] stores the outcome of the comparison of list[b_l+1] with the pivot. comp has only length b_r-(b_l+1) rather than b_r-b_l since we do not include the pivot element (and don't compare the pivot to itself).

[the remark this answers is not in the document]
benjaminsavage: Still, the first time through this loop, this code potentially calls:
```
list.swap(1, 1);
```

I assume that's an optimized case and nothing actually happens (I hope), and we need to increment `i` in that case, so there isn't an obvious way to avoid this.

benjaminsavage (approved):
Landing this to unblock work on the IPA protocol. 

We can continue follow up work in parallel.

PR #862, opened 2023-11-24
OPRF flow part 1

richajaindce:
This includes calling OPRF pseudonym code and also some minor refactoring

Next PR will add shuffle and quick sort


src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,216 @@
...
+        .await?;
+    let pseudonymed_inputs = input_rows
+        .into_iter()
+        .zip(pseudonymed_user_ids.into_iter())
+        .map(|(input, pseudonym)| PrfShardedIpaInputRow {
+            prf_of_match_key: pseudonym,
+            is_trigger_bit: input.is_trigger_bit,
+            breakdown_key: input.breakdown_key,
+            trigger_value: input.trigger_value,
+            timestamp: input.timestamp,
```
richajaindce, line 118: Wasn't able to put this within the above parallel_join. Compiler couldn't understand the Error type.

src/protocol/ipa_prf/boolean_ops/mod.rs
```diff
@@ -1,3 +1,3 @@
 mod addition_sequential;
 mod comparison_and_subtraction_sequential;
-mod share_conversion_aby;
+pub mod share_conversion_aby;
```
akoshelev, line 3: would it be better if we just re-export a single function out of this module?

```rust
mod share_conversion_aby;

pub use share_conversion_aby::convert_to_fp25519;
```

src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,222 @@
+#[cfg(feature = "descriptive-gate")]
```
akoshelev, line 1: Could be easier if you use this flag in `protocol/mod.rs` on `ipa_prf` module

src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,222 @@
...
+        pseudonymed_inputs,
+        user_cap.ilog2().try_into().unwrap(),
+        config.attribution_window_seconds,
+        &histogram,
+    )
+    .await
+}
+
+#[cfg(all(test, any(unit_test, feature = "shuttle")))]
+
```
akoshelev, line 142:
```suggestion
```

src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,222 @@
...
+        user_cap & (user_cap - 1) == 0,
+        "This code only works for a user cap which is a power of 2"
+    );
+
+    // TODO (richaj): Add shuffle either before the protocol starts or, after converting match keys to elliptical curve.
+    // We might want to do it earlier as that's a cleaner code
+
+    let convert_ctx = ctx
+        .narrow(&Step::ConvertFp25519)
+        .set_total_records(input_rows.len());
```
akoshelev, line 94: It should be possible to set total record once and then narrow

src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,222 @@
...
+    let eval_ctx = ctx
+        .narrow(&Step::EvalPrf)
+        .set_total_records(input_rows.len());
+
+    let prf_key = gen_prf_key(&convert_ctx);
+
+    let pseudonymed_user_ids = ctx
+        .parallel_join(input_rows.iter().zip(repeat(prf_key)).enumerate().map(
```
akoshelev, line 102: consider zipping both contexts in too, so you don't need to make the closure long

src/protocol/ipa_prf/mod.rs
```diff
@@ -1,7 +1,222 @@
...
+                    let record_id = RecordId::from(idx);
+                    let prf_of_match_key =
+                        convert_to_fp25519::<_, BA64>(convert_ctx, record_id, &record.match_key)
+                            .await?;
+                    eval_dy_prf(eval_ctx, record_id, &prf_key, &prf_of_match_key).await
+                }
+            },
+        ))
+        .await?;
+    let pseudonymed_inputs = input_rows
```
akoshelev, line 116: this second iteration over the input looks unnecessary. At the end of PRF evaluation circuit, the input row is ready and we could probably yield `PrfShardedIpaInputRow`

richajaindce: Agree on this but unable to get past a compile error. https://github.com/richajaindce/raw-ipa/pull/33

src/protocol/ipa_prf/mod.rs
richajaindce: We stumbled upon an issue while doing this. If a context has set_total_records once and then if set_total_records is called on a narrowed context, we throw an error. Is this an expected behaviour? Due to this, sticking to set_total_records on narrowed context itself

akoshelev: I would suggest to move PRF protocol into its own subroutine. This function is expected to grow as we add shuffle, shards etc. It is probably worth keeping its structure at high level. 

wrt to the issue you're facing - it is expected because attribution needs to use different total records constraint. 
Creating a top-level context for PRF should work

```rust
    let prf_ctx = ctx.set_total_records(input_rows.len());

    let convert_ctx = prf_ctx
        .narrow(&Step::ConvertFp25519);
    let eval_ctx = prf_ctx
        .narrow(&Step::EvalPrf);
```

akoshelev (approved):
lgtm, although I would love to add some guardrails around `truncate_from` so we don't miss fixing it after #812

ipa-core/src/report.rs
```diff
@@ -418,19 +420,11 @@ where
     TV: WeakSharedValue,
     TS: WeakSharedValue,
 {
-    pub timestamp: Replicated<TS>,
-    pub mk_oprf: u64,
-    pub event_type: EventType,
+    pub match_key: Replicated<BA64>,
```
richajaindce, line 423: I will need to revisit these types to reflect correct input but wil handle in next PR.
I think we need 
match_key: Replicated<BA64>,
is_trigger: Boolean,
breakdown_key: Replicated<BK>,
trigger_value: Replicated<TV>
timestamp: Replicated<TS>

or should bk and ts be sent in clear?

benjaminsavage (approved):
Looks good!

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -1,6 +1,232 @@
...
+    EvalPrf,
+    ConvertInputRowsToPrf,
+}
+
+/// IPA OPRF Protocol
+///
+/// We return `Replicated<F>` as output.
+/// This protocol does following steps
+/// 1. Shuffles the input (TBD)
+/// 2. Converts boolean arrays of match keys to elliptical values
```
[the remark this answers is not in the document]
richajaindce: That makes more sense actually. I think when we were discussing earlier, we didnt consider adding dummies hence thought it could be done in beginning or here. Updated

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -1,6 +1,232 @@
...
+                    trigger_value: 2,
+                },
+            ];
+
+            let mut result: Vec<_> = world
+                .semi_honest(records.into_iter(), |ctx, input_rows| async move {
+                    oprf_ipa::<_, BA8, BA3, BA20, BA5, Fp31>(
+                        ctx,
+                        input_rows,
+                        IpaQueryConfig::no_window(PER_USER_CAP, MAX_BREAKDOWN_KEY, NUM_MULTI_BITS),
```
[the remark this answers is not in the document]
richajaindce: we just use attribution window from config for now. We can just pass that for now.

ipa-core/src/query/executor.rs
```diff
@@ -224,7 +223,7 @@ pub fn execute(
             move |prss, gateway, config, input| {
                 let ctx = SemiHonestContext::new(prss, gateway);
                 Box::pin(
-                    OprfIpaQuery::<_, Fp32BitPrime>::new(ipa_config)
+                    OprfIpaQuery::<_, crate::ff::Fp31>::new(ipa_config)
```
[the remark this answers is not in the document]
richajaindce: no we need it similar to every other call in this function

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -1,6 +1,232 @@
...
+
+    let prf_key = gen_prf_key(&convert_ctx);
+
+    ctx.parallel_join(
+        input_rows
+            .into_iter()
+            .zip(zip(
+                repeat(prf_key),
+                zip(repeat(convert_ctx), repeat(eval_ctx)),
+            ))
```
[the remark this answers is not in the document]
benjaminsavage: I like this approach better. I think there is a relatively weaker connection between the order of the arguments way up on previous lines, and the complex, deeply nested parenthetical argument lists below.

ipa-core/src/protocol/ipa_prf/mod.rs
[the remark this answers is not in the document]
benjaminsavage: I don't see any reason to define the constand up there and pass it along if it's going to be totally ignored. It was the `truncate` line that caught me. It indicates that the output is not respecting this argument.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -1,6 +1,232 @@
...
+    F: PrimeField + ExtendableField,
+    Replicated<F>: Serializable,
+{
+    let ctx = ctx.set_total_records(input_rows.len());
+    let convert_ctx = ctx.narrow(&Step::ConvertFp25519);
+    let eval_ctx = ctx.narrow(&Step::EvalPrf);
+
+    let prf_key = gen_prf_key(&convert_ctx);
+
+    ctx.parallel_join(
```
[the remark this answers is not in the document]
richajaindce: will leave it as is for now, and change in next PR

ipa-core/src/protocol/ipa_prf/mod.rs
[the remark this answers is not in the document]
richajaindce: agree on that, but we use it for older IPA. Ideally all these configs should flow from the variable but due to type issues we hit, we are unble to use this Config. For now just passing attribution window to oprf function

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -1,6 +1,224 @@
...
+    F: PrimeField + ExtendableField,
+    Replicated<F>: Serializable,
+{
+    let ctx = ctx.set_total_records(input_rows.len());
+    let convert_ctx = ctx.narrow(&Step::ConvertFp25519);
+    let eval_ctx = ctx.narrow(&Step::EvalPrf);
+
+    let prf_key = gen_prf_key(&convert_ctx);
+
+    ctx.parallel_join(input_rows.into_iter().enumerate().map(|(idx, record)| {
```
richajaindce, line 127: From performance POV, these PRfs should be calculated in parallel vs sequentially  right?

ipa-core/src/query/executor.rs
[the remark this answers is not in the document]
richajaindce: Oh this branch is for "weak-field" which is supposed to work for Fp31. Earlier it was wrong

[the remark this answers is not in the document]
benjaminsavage: oh interesting. I didn't notice this config before.

benjaminsavage (approved):

PR #864, opened 2023-11-28
Adding test case for wrapping bug

benjaminsavage:
Figured it out. Capping wasn't working correctly. 
Once a user reaches the cap, it should stay "saturated" forever. 
So the logic should be "this row is saturated if EITHER this row just overflowed, OR if the previous row was already saturated".
I accidentally removed this important functionality when I migrated the code to use Boolean Array.
I utilized the clever trick that @andyleiserson proposed here: https://github.com/private-attribution/ipa/pull/785#discussion_r1333647004 which allows me to compute this "OR" without any additional communication.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
@@ -844,12 +844,15 @@ pub mod tests {
         timestamp: TS,
     }
 
-    fn oprf_test_input(
+    fn oprf_test_input<BK>(
         prf_of_match_key: u64,
         is_trigger: bool,
         breakdown_key: u8,
         trigger_value: u8,
-    ) -> PreShardedAndSortedOPRFTestInput<BA5, BA3, BA20> {
+    ) -> PreShardedAndSortedOPRFTestInput<BK, BA3, BA20>
+    where
+        BK: WeakSharedValue + Field,
+    {
         oprf_test_input_with_timestamp(
             prf_of_match_key,
             is_trigger,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1068,4 +1071,54 @@ pub mod tests {
...
+                oprf_test_input_with_timestamp(10251308645, true, 0, 2, 471555),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 530458),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 536295),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 600360),
+            ];
+
+            let mut expected = [0_u128; 256];
+            expected[218] = 32;
+
+            let histogram = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1];
```
martinthomson, line 1097: could be a const

like `const HISTOGRAM: usize = [1; 10];`

But I guess that you really want `let histogram = [1; records.len()];`

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1068,4 +1071,54 @@ pub mod tests {
...
         });
     }
+
+    #[test]
+    fn capping_bugfix() {
+        run(|| async move {
+            let world = TestWorld::default();
+
+            let records: Vec<PreShardedAndSortedOPRFTestInput<BA8, BA3, BA20>> = vec![
+                /* First User */
```
martinthomson, line 1081:
```suggestion
```

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1068,4 +1071,54 @@ pub mod tests {
...
+                oprf_test_input_with_timestamp(10251308645, false, 218, 0, 29135),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 3, 179334),
```
martinthomson, line 1083: It would be more helpful if the timestamps were ordered.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1068,4 +1071,54 @@ pub mod tests {
...
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 368968),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 1, 453715),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 2, 471555),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 530458),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 536295),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 600360),
+            ];
+
+            let mut expected = [0_u128; 256];
+            expected[218] = 32;
```
martinthomson, line 1095: Connecting this with the use of BA5 below is not immediately obvious.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
benjaminsavage: What do you mean? They are in increasing time order.

martinthomson: Oops, digit count miss on my end :(

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -1077,24 +1082,26 @@ pub mod tests {
...
-                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 536295),
-                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 600360),
+                oprf_test_input_with_timestamp(10251308645, true, 0, 3, 179334), // running-sum = 3
+                oprf_test_input_with_timestamp(10251308645, true, 0, 3, 313530), // running-sum = 6
+                oprf_test_input_with_timestamp(10251308645, true, 0, 5, 327956), // running-sum = 11
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 368968), // running-sum = 17
+                oprf_test_input_with_timestamp(10251308645, true, 0, 1, 453715), // running-sum = 18
+                oprf_test_input_with_timestamp(10251308645, true, 0, 2, 471555), // running-sum = 20
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 530458), // running-sum = 26
+                oprf_test_input_with_timestamp(10251308645, true, 0, 6, 536295), // running-sum = 32
```
martinthomson: Did you want to pick a number that doesn't perfectly saturate the sum?

martinthomson (commented):
Nice.  Free.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
benjaminsavage: Clippy made me add some underscores, that should improve readability.

PR #868, opened 2023-11-29
quick fix for overflow

danielmasny:
quick fix for overflow issue when using BA256 with new truncate function. 2 tests failed (doesn't seem related to PR): 

test protocol::ipa_prf::prf_sharding::tests::capping_bugfix ... FAILED
test protocol::ipa_prf::prf_sharding::tests::semi_honest_aggregation_capping_attribution_with_attribution_window ... FAILED

new truncate leads to insecure PRSS for BA256, will get fixed when issue https://github.com/private-attribution/ipa/issues/812 is addressed

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -138,7 +138,7 @@ macro_rules! boolean_array_impl {
                     let v = v.into();
                     let mut val = Self::ZERO;
                     for i in 0..$bits {
-                        val.set(i, Boolean::from(v <= 128 && (v >> i & 1) == 1));
+                        val.set(i, Boolean::from(i < 128 && (v >> i & 1) == 1));
                     }
 
                     val
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -137,8 +137,8 @@ macro_rules! boolean_array_impl {
                 fn truncate_from<T: Into<u128>>(v: T) -> Self {
                     let v = v.into();
                     let mut val = Self::ZERO;
-                    for i in 0..$bits {
-                        val.set(i, Boolean::from(i < 128 && (v >> i & 1) == 1));
+                    for i in 0..std::cmp::min(128, $bits) {
+                        val.set(i, Boolean::from((v >> i & 1) == 1));
                     }
 
                     val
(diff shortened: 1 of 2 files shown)


martinthomson (commented):
v <= 128 will only convey very small values. Don’t you at least want i <= 128 ?

danielmasny:
> v <= 128 will only convey very small values. Don’t you at least want i <= 128 ?

Yes :D that was a typo. I actually need i<128, i=128 still leads to an overflow.

ipa-core/src/ff/boolean_array.rs
```diff
@@ -138,7 +138,7 @@ macro_rules! boolean_array_impl {
                     let v = v.into();
                     let mut val = Self::ZERO;
                     for i in 0..$bits {
-                        val.set(i, Boolean::from((v >> i & 1) == 1));
+                        val.set(i, Boolean::from(i < 128 && (v >> i & 1) == 1));
```
akoshelev, line 141: can you put a comment explaining why and when we are going to fix it here?

akoshelev: all right, I think the semantic is not changed and we accept only 128 bit values. there is no reason to iterate over all of them, you can just take first 128 or build a mask and do a single AND instruction - compiler likely can do it too, but I haven't looked at assembly code, so I can't be sure

akoshelev (approved):

ipa-core/src/ff/boolean_array.rs
```diff
@@ -138,7 +138,7 @@ macro_rules! boolean_array_impl {
                     let v = v.into();
                     let mut val = Self::ZERO;
                     for i in 0..$bits {
```
martinthomson, line 140: This is better, though I note that `min` is not `const`.  I expect that the compiler is capable of dealing with that though.

```suggestion
                    for i in 0..std::cmp::min(128,$bits) {
```

Note that this is faster because it is a single comparison, not `$bits` comparisons.  I think that it also makes intent clearer, to the point Alex makes below.

ipa-core/src/ff/boolean_array.rs
martinthomson, line 141: I can't suggest a change to revert this line, but you should.

martinthomson (changes requested):

martinthomson (approved):

PR #869, opened 2023-11-30
Fixing error when running collect steps

benjaminsavage:
I created another regression in landing #853. 

When running the collect steps script, the outputs do not match the expected values due to differences in how per-user capping is applied.

Unfortunately, now that the per-user cap is a compile-time constant, it's harder to deal with this. I'm not sure how to dynamically select a `type` based on the value of a u32. I'm hopeful that someone can suggest a much better way to do this.

This issue is bigger than just the per-user cap. The number of breakdown keys, the maximum size of a trigger value, everything is a compile time constant now, and we potentially want to choose these things dynamically based on the values provided at runtime.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -318,6 +318,17 @@ boolean_array_impl!(
     )
 );
 
+//impl BA4
+boolean_array_impl!(
+    boolean_array_4,
+    BA4,
+    4,
+    1,
+    bitarr ! ( const u8, Lsb0;
+        1, 0, 0, 0
+    )
+);
+
 //impl BA5
 boolean_array_impl!(
     boolean_array_5,
@@ -326,6 +337,28 @@ boolean_array_impl!(
     1,
     bitarr ! ( const u8, Lsb0;
         1, 0, 0, 0, 0
+    )
+);
+
+//impl BA6
+boolean_array_impl!(
+    boolean_array_6,
+    BA6,
+    6,
+    1,
+    bitarr ! ( const u8, Lsb0;
+        1, 0, 0, 0, 0, 0
+    )
+);
+
+//impl BA7
+boolean_array_impl!(
+    boolean_array_7,
+    BA7,
+    7,
+    1,
+    bitarr ! ( const u8, Lsb0;
+        1, 0, 0, 0, 0, 0, 0
     )
 );
 
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -40,7 +40,7 @@ impl<'a> Iterator for BAIterator<'a> {
 
 //macro for implementing Boolean array, only works for a byte size for which Block is defined
 macro_rules! boolean_array_impl {
-    ($modname:ident, $name:ident, $bits:expr, $bytes:expr, $one:expr) => {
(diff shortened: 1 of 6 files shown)


ipa-core/src/test_fixture/ipa.rs
```diff
@@ -298,22 +298,97 @@ pub async fn test_oprf_ipa<F>(
...
-                    F,
-                >(
-                    ctx,
-                    sharded_input,
-                    config.attribution_window_seconds,
-                    ref_to_histogram,
-                )
-                .await
-                .unwrap()
+                if config.per_user_credit_cap == 8 {
```
benjaminsavage, line 301: This is horrible. I apologize. I would *love* a suggestion for how to do this in a less terrible way.

@martinthomson, @akoshelev

ipa-core/src/ff/boolean_array.rs
```diff
@@ -318,6 +318,17 @@ boolean_array_impl!(
...
 
+//impl BA4
+boolean_array_impl!(
+    boolean_array_4,
+    BA4,
+    4,
+    1,
+    bitarr ! ( const u8, Lsb0;
+        1, 0, 0, 0
+    )
```
martinthomson, line 329: I find the formatting here disturbing.  Is there any reason that we can't have this be "[1, 0, 0, 0]" with the common part (`bitarr![const u8, Lsb0; ...]`) supplied by the macro?

ipa-core/src/test_fixture/ipa.rs
martinthomson: I'm not aware of any.  Basically, what you are really saying here is that based on the value of this variable (`per_user_credit_cap`), run different code.  That the difference between the arms is in many ways a variable, you are now into the realm of metaprogramming.  A macro is one option, but even that carries risks.

The only thing that might make this a miniscule amount easier is to use `match` rather than this if/else business.

martinthomson (approved):

ipa-core/src/ff/boolean_array.rs
benjaminsavage: I can't figure out how to do this.

The last thing is supposed to be an "expression". I tried an array, a tuple, and neither of them can be used in that macro.  I'm going to leave this as is for now.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -393,7 +393,7 @@ where
 /// Propagates errors from multiplications
 /// # Panics
 /// Propagates errors from multiplications
-pub async fn attribution_and_capping_and_aggregation<C, BK, TV, TS, SS, S, F>(
+pub async fn attr_cap_aggr<C, BK, TV, TS, SS, S, F>(
```
richajaindce, line 396: Can we name it better?

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -297,23 +296,67 @@ pub async fn test_oprf_ipa<F>(
...
-                    F,
-                >(
-                    ctx,
-                    sharded_input,
-                    config.attribution_window_seconds,
-                    ref_to_histogram,
-                )
-                .await
-                .unwrap()
+                match config.per_user_credit_cap {
```
richajaindce, line 301: we would need the same code in report collector as well.

PR #874, opened 2023-11-30
Rename WeakSharedValue

andyleiserson:
Long ago, `SharedValue` was a base trait common to `Field` and `BitArray`. At some point, `BitArray` became `GaloisField` and was changed to require `Field`. At that point, I think `SharedValue` and `Field` became effectively synonymous.

Recently, the need arose for SharedValues that do not have multiplicative properties. This was introduced as a new `WeakSharedValue`.

This change merges the notion of `SharedValue` into `Field` and then renames `WeakSharedValue` to `SharedValue`.


akoshelev (approved):
lgtm!

benjaminsavage (approved):

danielmasny:
I think using the name field is confusing because it could also be a vector or bitarray which is not a field but also defines a multiplication (e.g. component wise multiplication rather than field multiplication, i.e. polynomial multiplication). I would prefer a different name.

PR #875, opened 2023-12-01
metaprogramming for boolean array values of 1

martinthomson:
I don't know if this is much better than repeating "0, " 255 times, but it was an interesting exercise.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -82,7 +82,7 @@ macro_rules! boolean_array_impl {
 
     type Store = BitArr!(for $bits, in u8, Lsb0);
 
-            /// A Bootean array with $bits bits.
+            /// A Boolean array with $bits bits.
             #[derive(Clone, Copy, PartialEq, Eq, Debug)]
             pub struct $name(pub Store);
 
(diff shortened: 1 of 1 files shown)


martinthomson:
See also https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=699491dd1c8558554987671e10a834fc

ipa-core/src/ff/boolean_array.rs
```diff
@@ -307,73 +334,14 @@ store_impl!(U8, 64);
 store_impl!(U32, 256);
 
 //impl BA3
```
akoshelev, line 336:
```suggestion
```

or something more applicable to multiple implementations

akoshelev (approved):
this is cool. I hope I'll never need to change this code. Is there any impact on build time with this macro in place?

ipa-core/src/ff/boolean_array.rs
```diff
@@ -55,7 +82,7 @@ macro_rules! boolean_array_impl {
 
     type Store = BitArr!(for $bits, in u8, Lsb0);
 
-            ///
+            /// A Bootean array with $bits bits.
```
benjaminsavage, line 85: Bootean is cute.

PR #877, opened 2023-12-04
Remove `Field` trait from PRSS

andyleiserson:
This (mostly) removes the `Field` trait from PRSS. Only mostly, because a compatibility wrapper `generate_fields` is kept, however, if desired, it could easily be removed at the cost of slightly more change to existing code. (I already got rid of a similar wrapper, `generate_replicated`, which had fewer uses.)

This change introduces two new traits:
* `FromPrss`, which is a trait for things that can be generated from PRSS. I don't see it as having any fundamental difference from rand's `Distribution<Standard>`, but since the PRSS doesn't operate over `Distribution<Standard>`, I think it is useful to have a dedicated trait here.
* `FromRandomU128`. The ability to produce random instances of a type from a random u128 was previously an implicit guarantee of implementing `truncate_from`.

Unfortunately, there is additional cleanup that needs to happen that couldn't be done as part of this PR without significantly expanding the scope:
* Removing `impl Field` from the BooleanArray types. This is not possible right now because the OPRF IPA code is relying on BooleanArrays being `impl Field` in order to pass them to our MPC multiply.
* Removing the supertrait bound `Field: FromRandomU128` (because large fields, like `Fp25519`, cannot properly be sampled from a random U128). Doing this would require a significant amount of our existing code be updated to add `FromRandomU128` to trait bounds. Since we have so much code that writes `Field` meaning "small field", I think what makes the most sense is to define a new trait `ArbitrarilyLargeField` that can be used when we know we want to relax the "small" requirement. `Field` would retain the `FromRandomU128` bound; `ArbitrarilyLargeField` would not have it. `ArbitrarilyLargeField` is an unwieldy name, but I don't think it will appear very often. 

I plan to address the additional cleanup in upcoming vectorization work.

I am also keeping in mind @akoshelev's [suggestion](https://github.com/private-attribution/ipa/issues/812#issuecomment-1825019292):
> As part of this task, we probably should also fix the PRSS and let it generate a given number of bytes, rather than constraining it to 128 bits

I agree this makes sense, and I may decide to tackle it as part of PRSS generation of vectors. Just for BA256 and Fp25519, we can do without, and just get two u128s from PRSS.


ipa-core/src/ff/boolean_array.rs
```diff
@@ -377,3 +389,33 @@ boolean_array_impl!(
...
+        (BA256::from((l0, l1)), BA256::from((r0, r1)))
+    }
+}
+
+impl rand::distributions::Distribution<BA256> for rand::distributions::Standard {
+    fn sample<R: crate::rand::Rng + ?Sized>(&self, rng: &mut R) -> BA256 {
+        BA256::from((rng.gen(), rng.gen()))
+    }
+}
+*/
```
benjaminsavage, line 421: Do you want to land these comments? Or remove them before landing?

ipa-core/src/ff/ec_prime_field.rs
```diff
@@ -125,15 +127,19 @@ impl From<Scalar> for Fp25519 {
...
 impl From<BA256> for Fp25519 {
     fn from(s: BA256) -> Self {
         let mut buf: GenericArray<u8, U32> = [0u8; 32].into();
         s.serialize(&mut buf);
+        // Reduces mod order
         Fp25519::deserialize(&buf)
     }
 }
 
+/*
```
benjaminsavage, line 142: Can we just delete this code instead?

benjaminsavage (approved):
Seems like a good first step.

ipa-core/src/protocol/prss/crypto.rs
```diff
@@ -8,12 +8,67 @@ use sha2::Sha256;
...
+///     with the left helper, and one with the randomness that is shared with the right helper.
+///     Thus, one of the generated values is known to both us and the left helper, and likewise for
+///     the right helper.
+///  2. Secret sharings: In this case, a single secret-shared random value is generated. The value
+///     returned by `FromPrss` is our share of that sharing. Within `FromPrss`, the randomness shared
+///     with the left and right helpers is used to construct the sharing.
+///
+/// In the first case, `FromPrss` is implemented for a tuple type, while in the second case,
+/// `FromPrss` is implemented for a secret-shared type.
+pub trait FromPrss: Sized {
```
akoshelev, line 42: I don't know how, but it would be great if we could capture the fact that values need to be sampled uniformly. Right now, someone could just naively sample a single byte from PRSS and then expand it to any `T` by padding, masking, etc.

Feels like `Distribution<Standard>` at least puts a warning sign, but I don't know if we could put it here. Maybe a rustdoc comment for `FromPrss`?

akoshelev (approved):

ipa-core/src/ff/boolean_array.rs
andyleiserson: These I want to keep -- I believe this is the eventual form of what they should look like for BA256. One of the changes I tried to make before realizing that it's hard to remove `Field` from the BooleanArrays right now was to have the macro-generated boolean array implementation only generate its version of these for sizes up to 128 bits. Unfortunately, needing to retain Field means needing to retain FromRandomU128, which has its own FromPrss that conflicts with this one.

ipa-core/src/ff/ec_prime_field.rs
andyleiserson: Yes, I meant to remove this -- unless @danielmasny has a reason why it should stay.

ipa-core/src/protocol/prss/crypto.rs
andyleiserson: Yeah. I originally had something in the comment like "don't implement this if it could be a footgun", but took it out because it seemed unhelpful. It's a bit wordy, but here's what I have now:

```
/// The exact semantics of the generation depend on the value being generated, but like
/// `rand::distributions::Standard`, a uniform distribution is typical. When implementing
/// this trait, consider the consequences if the implementation were to be used in
/// an unexpected way. For example, an implementation that draws from a subset of the
/// possible values could be dangerous, if used in an unexpected context where
/// security relies on sampling from the full space.
```

ipa-core/src/ff/ec_prime_field.rs
danielmasny: I think I have only used this to debug. Since we do not use it, we can delete it. I think I only use from<BA256> for Fp25519. I use that in the share conversion from BA256 to Fp25519.

ipa-core/src/protocol/prss/crypto.rs
danielmasny: I was also thinking of requiring Distribution<Standard> being implemented for PRSS and then generating a seed using AES and using the seed to define an rng which then generates the desired random value. It seems that you only solve the problem for BA256, but we need PRSS also for small values like BA8 and so on. I would prefer to put it in the macro and solve the issue in general, for any BA type.

ipa-core/src/ff/boolean_array.rs
danielmasny: Do you plan to implement the trait from PRSS for any type that needs an MPC multiplication?

ipa-core/src/protocol/prss/crypto.rs
andyleiserson: I feel like another level of AES can only slow things down, but maybe I'm not understanding.

I'm still looking at this stuff in the context of PRSS generation of an array. There are a bunch of issues to figure out. I am exploring the path of supporting random generation from an integer multiple of u128s. (Rationale being that we can always vectorize to get things that wide.) But there are a bunch of issues (trait bounds, limited const generic support) that I haven't figured out yet. Generating from a multiple of u128s at least keeps the PRSS indices reasonably well-defined. If we want to generate below the level of u128s, we may have to revise what a PRSS index means.

> we need PRSS also for small values like BA8 and so on

We have it for the small BAs currently via `FromRandomU128` and `Field::truncate_from`. When we make them not Fields, we can move the contents of `Field::truncate_from` into `FromRandomU128`. The generation is not efficient in terms of PRSS bits consumed, but neither is generation of Fp32BitPrimes.

andyleiserson: To offer something more concrete, this is what I'm working with right now:

```
pub trait FromRandom {
    type Source; // == [u128; N]
    fn len() -> usize; // returns N

    /// Generate a random value of `Self` from a uniformly-distributed `Self::Source`.
    fn from_random(src: Self::Source) -> Self;
}
```

This seems like it can work, but is a bit unwieldy. If we wanted to use `[u8; N]` instead of `[u128; N]`, we could.

danielmasny: I think that would make sense. I am a bit worried about adding more traits around PRSS. Ideally generate value would give you the N random values.

ipa-core/src/ff/boolean_array.rs
andyleiserson: That was my plan. Does that seem like a problem?

danielmasny: isn't that reimplementing Distribution<Standard> with the only difference using a seed instead of a RNG? We could also consider having PRSS returning an RNG (and implement an RNG that panics when we request too many bits). This would allow us to use the standard rust conventions.

danielmasny: Here with panic, I mean debug panic such that we could identify when we requesting AES evaluations for identical inputs twice.

danielmasny (approved):
We can address best naming and changes to PRSS in a separate PR.

ipa-core/src/protocol/prss/crypto.rs
andyleiserson: I opened #884 to continue this discussion.

Issue #867, opened 2023-11-29
OPRF stall


benjaminsavage:
# Starting Helpers:
```
btsavage@btsavage-mbp raw-ipa % cargo run --release --no-default-features --features "cli web-app real-world-infra compact-gate stall-detection" --bin helper -- --identity 3 --network test_data/network.toml --port 3002 --disable-https
```
# Running Script:
```
btsavage@btsavage-mbp raw-ipa % SIZE=1000000; cargo run --release --bin report_collector --features="clap cli test-fixture" -- gen-ipa-inputs -n $SIZE --max-breakdown-key 256 --report-filter all --max-trigger-value 7 --seed 123 > /tmp/events-${SIZE}.txt                                                 
```

Helper 1
```
2023-11-29T00:36:37.268350Z  WARN stall_detector{role=H1}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=10657796 state=
{
"gate[11422]=binary_validator/row9/computed_capped_attributed_trigger_value_just_saturated_case", from=H3. Waiting to receive records ["[2163..3071]"].
"gate[11436]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit0/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11444]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit1/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11452]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit10/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11461]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit2/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11469]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit3/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11477]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit4/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
"gate[11485]=prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values/convert_bit5/xor2", from=H3. Waiting to receive records ["[18432..19455]"].
[...]

PR #878, opened 2023-12-04
Use sequential join to compute PRF

akoshelev:
Running it with tracing enabled to debug #867 made it quite obvious that we attempt to send 1M records without having a chance to succeed.

I also included some minor fixes to improve diagnostics for event type deserialization


richajaindce (approved):

benjaminsavage (approved):
I don't really understand why this wasn't working before, but I'm glad you fixed it!

PR #881, opened 2023-12-05
updates suggested by Alex on quicksort

danielmasny:
Using bitvec and other minor changes

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/quicksort.rs
+++ b/ipa-core/src/protocol/ipa_prf/quicksort.rs
@@ -1,3 +1,4 @@
+use bitvec::prelude::{BitVec, Lsb0};
 use futures::stream::{iter as stream_iter, TryStreamExt};
 use ipa_macros::Step;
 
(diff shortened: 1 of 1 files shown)


danielmasny:
This is an update for https://github.com/private-attribution/ipa/pull/855

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -47,12 +48,17 @@ pub async fn quicksort_by_key_insecure<C, K, F, S>(
 where
     C: Context,
     S: Send + Sync,
-    F: Fn(&S) -> &AdditiveShare<K> + Send + Sync,
+    F: Fn(&S) -> &AdditiveShare<K> + Sync,
```
akoshelev, line 51: 👍

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -47,12 +48,17 @@ pub async fn quicksort_by_key_insecure<C, K, F, S>(
...
     for<'a> &'a AdditiveShare<K>: IntoIterator<Item = AdditiveShare<K::Element>>,
     K: WeakSharedValue + Field + CustomArray<Element = Boolean>,
 {
+    // expected amount of recursion: compute ceil(log_2(list.len()))
+    let expected_depth = std::mem::size_of::<usize>() * 8 - list.len().leading_zeros() as usize;
     // create stack
-    let mut stack: Vec<(C, usize, usize)> = vec![];
+    let mut stack: Vec<(C, usize, usize)> = Vec::with_capacity(expected_depth);
+    // vector for comparison (against pivot) outcomes
+    #[allow(unused_assignments)]
```
akoshelev, line 60: Rustc is right - you are still reallocating those bit vectors on each iteration. There is no `try_collect(container)`, so you would probably do it by hand

```rust
comp.clear();
while let Some(bool_value) = seq_join(ctx.active_work(), stream_iter(...)).next().await {
    comp.push(bool_value);
}
```

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -85,22 +94,28 @@ where
...
+                                    .await?,
+                        )
                     }
                 })),
             )
-            .try_collect::<Vec<_>>()
+            .try_for_each(|x| {
+                comp.push(x);
+                future::ready(Ok(()))
+            })
```
akoshelev, line 110:
```suggestion
            .for_each(|x| {
                comp.push(x);
                future::ready(())
            })
```

danielmasny: I think I need try since x is a result (i.e. mpc multiplication can fail)

akoshelev: oh that's probably right, it will return early if one of the multiplication fails.

:shipit:

PR #882, opened 2023-12-06
Plumbing in Shuffle to oprf protocol

richajaindce:
This is still not called in the protocol as it would break the logic without sort


ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -225,4 +262,61 @@ pub mod tests {
...
+            ];
+
+            let mut result: Vec<TestRawDataRecord> = world
+                .semi_honest(records.clone().into_iter(), |ctx, input_rows| async move {
+                    shuffle_inputs::<_, BA8, BA3, BA20>(ctx, input_rows)
+                        .await
+                        .unwrap()
+                })
+                .await
+                .reconstruct();
```
richajaindce, line 316: Compiler unable to compile this reconstruct 
"the method `reconstruct` exists for array `[Vec<OprfReport<BA8, BA3, BA20>>; 3]`, but its trait bounds were not satisfied
items from traits can only be used if the trait is implemented and in scoperustc
(rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2FUsers%2Frichaj%2FProposals%2FIPA%2Fraw-ipa%2Fipa-core%2Fsrc%2Fprotocol%2Fipa_prf%2Fmod.rs)
mod.rs(309, 54):
sharing.rs(82, 27): trait bound `[&'v [OprfReport<boolean_array_8::BA8, boolean_array_3::BA3, boolean_array_20::BA20>]; 3]: test_fixture::sharing::Reconstruct<Vec<_>>` was not satisfied"

akoshelev: done. please note that we need to increase the size of the input in order to assert random order of shuffle

richajaindce: thanks a ton as always :)

ipa-core/src/ff/boolean_array.rs
```diff
@@ -358,6 +382,20 @@ boolean_array_impl!(
...
 );
 
+// impl BA112
+boolean_array_impl!(
+    boolean_array_112,
+    BA112,
+    112,
+    14,
+    [
+        1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
benjaminsavage, line 392: No need for this if we land @martinthomson's crazy metaprogramming thingy first.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+) -> AdditiveShare<YS>
+where
+    YS: CustomArray<Element = <BA112 as CustomArray>::Element> + SharedValue,
+    BK: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TV: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+{
+    let (mut y_left, mut y_right) = (YS::ZERO, YS::ZERO);
+
+    for i in 0..BA64::BITS as usize {
```
benjaminsavage, line 61: This for loop over a range that calls set on left and right is what I thought Expand did. If not, probably worth defining a function to do this.

benjaminsavage (changes requested):

ipa-core/src/ff/boolean_array.rs
martinthomson: You'll need `[0,][0,][0,][0,][][0,][0,]` for that.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    YS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    BK: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TV: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+{
+    let match_key = extract_from_shared_array::<YS, BA64>(input, 0);
+
+    let mut offset = BA64::BITS as usize;
+
+    let is_trigger = AdditiveShare::<Boolean>::new(
```
richajaindce, line 97: i can create a BA1 to encapsulate is_trigger so that all code looks uniform. But it would need refactoring is_trigger in more places. lmk if that looks better

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    input: &OprfReport<BK, TV, TS>,
+) -> AdditiveShare<YS>
+where
+    YS: CustomArray<Element = <BA112 as CustomArray>::Element> + SharedValue,
+    BK: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TV: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+    TS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+{
+    let (mut y_left, mut y_right) = (YS::ZERO, YS::ZERO);
+    expand_array_in_place(&mut y_left, input.match_key.left(), 0);
```
richajaindce, line 61: was trying to avoid multiple allocations for y. Is there a way to pass mutable tuples and change their values inside a function?

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
danielmasny: Boolean seems to me the right choice here. Using BA1 just adds another layer around it that probably makes it harder to use and doesn't seem to offer any advantages (beyond that you could have a generic concatenate functionality that can concatenate any BAx and BAy to a BAx+y)

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    offset += TV::BITS as usize;
+    expand_array_in_place(&mut y_left, input.timestamp.left(), offset);
+    expand_array_in_place(&mut y_right, input.timestamp.right(), offset);
+
+    AdditiveShare::<YS>::new(y_left, y_right)
+}
+
+// This function converts AdditiveShare obtained from shuffle protocol to OprfReport
+pub fn shuffled_to_oprfreport<YS, BK, TV, TS>(input: &AdditiveShare<YS>) -> OprfReport<BK, TV, TS>
```
danielmasny, line 86: I have noticed that we have a lot of structs that are more or less representing the same data. I was wondering whether we can get rid of some of the structs like OPRF report. I was thinking that the decryption just outputs a bunch of u128s (or [u8, 16] from AES) we can convert that in a large BA containing mk, tb, bk, tv, ts. We could directly shuffle the BA without converting them to a oprfreport. Then for evaluating the prf, you just call it on a stream over the BAs that extracts the mk from the BA and overwrites the mk with its pseudonym (we could make sure mk and pseudonym use the same BA type. For the sort according to pseudonym you also use the BA type as input. After all of that you could finally drop the pseudonym and extract the rest in the BA, i.e. tb, bk, tv, ts, and pass it into the attribution circuit. This way we wouldn't need all the conversions between the different struct types which might cost reallocations (maybe the compiler is able to optimize them away?).

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+                        .await
+                        .unwrap()
+                })
+                .await
+                .reconstruct();
+            assert_ne!(result, records);
+            records.sort();
+            result.sort();
+            assert_eq!(result, records);
+        });
```
danielmasny, line 165: I think that is a good test :)

danielmasny (commented):
thanks for your work! The PR looks good to me. I added some comments where I think we could improve/compress the code. I am not sure whether it ends up being the same in terms of efficiency due to compiler optimizations. Some of the comments might also be beyond the scope of this PR so feel free to ignore them but please let me know what you think!

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -183,6 +183,45 @@ where
...
 }
 
+/// inserts smaller array in the larger array starting from location offset
+#[allow(dead_code)]
+pub fn expand_array_in_place<YS, XS>(y: &mut YS, x: XS, offset: usize)
+where
+    YS: CustomArray<Element = Boolean> + SharedValue,
+    XS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+{
+    for i in 0..XS::BITS as usize {
```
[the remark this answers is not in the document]
richajaindce: Oh I thought BITS is what is the length of the array. Cant locate a length function in custom array. So will leave it as is for now (probably something for followup) 
CC: @benjaminsavage

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    let shuffled = shuffle(ctx, shuffle_input).await?;
+
+    Ok(shuffled
+        .into_iter()
+        .map(|item| shuffled_to_oprfreport(&item))
+        .collect::<Vec<_>>())
+}
+
+// This function converts OprfReport to an AdditiveShare needed for shuffle protocol
+pub fn oprfreport_to_shuffle_input<YS, BK, TV, TS>(
```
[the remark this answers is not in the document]
richajaindce: I was also thinking of creating a generic function for this but we dont have all BAs here (is_trigger is a boolen :/ ). Macro is a good idea - maybe good as a follow up?

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
richajaindce: thats true but it would help keep everything in same format as commented above

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
richajaindce: worked!

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
benjaminsavage: I agree. This seems like a potential optimization that may come at the cost of readbility. Let's revisit that later in the optimization stage.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
benjaminsavage: Kind of. It doesn't actually test that the inputs were shuffled though. This just ensures that the same set of elements appears after shuffling.

benjaminsavage (approved):
Looks good! Thanks @richajaindce!

PR #883, opened 2023-12-07
A nicer interface to bitarr_one

martinthomson:
This might make @richajaindce less put off by the 112-bit stuff.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -40,29 +40,45 @@ impl<'a> Iterator for BAIterator<'a> {
 
 /// A value of ONE has a one in the first element of the bit array, followed by `$bits-1` zeros.
 /// This macro uses a bit of recursive repetition to produce those zeros.
-/// The binary value of `$bits-1` is expanded in LSB order,
-/// One-valued bits are represented as `[0,]`, zero-valued bits as `[]`.
+///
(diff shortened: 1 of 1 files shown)


ipa-core/src/ff/boolean_array.rs
```diff
@@ -40,29 +40,45 @@ impl<'a> Iterator for BAIterator<'a> {
...
-    (7) => { bitarr_one!([][0,][0,]) };
-    (8) => { bitarr_one!([0,][0,][0,]) };
-    (20) => { bitarr_one!([0,][0,][][][0,]) };
-    (32) => { bitarr_one!([0,][0,][0,][0,][0,]) };
-    (64) => { bitarr_one!([0,][0,][0,][0,][0,][0,]) };
-    (256) => { bitarr_one!([0,][0,][0,][0,][0,][0,][0,][0,]) };
-    // This is where recursion ends.
-    ([$($x:tt)*]) => { bitarr![const u8, Lsb0; 1, $($x)*] };
+
+    // The binary value of `$bits-1` is expanded in LSB order for each of the values we care about.
```
richajaindce, line 47: can you also give an example of the conversion in comments for dumb ones like me. i.e. 20 will be expanded to (20-1=)19 = 10011 in reverse order resulting to 11001

richajaindce:
This is exactly how I decoded it.

> This might make @richajaindce less put off by the 112-bit stuff.

PR #885, opened 2023-12-08
Require that `SharedValue: Eq`

andyleiserson:
The compiler is smart enough to conditionally derive `Eq` on `AdditiveShare<V: SharedValue>` depending on whether `V: Eq`. However, when the picture gets more complicated for vectorization, there are cases where the compiler can't figure out the Eq / non-Eq variants. Requiring that `SharedValue`s are always `Eq` will simplify things.

(I don't really think there's anywhere that the Eq vs. PartialEq distinction for these types really matters, so we could probably address this by removing any use of `Eq` `AdditiveShare`s, but my impression is that would be a bigger change.)

The only case where I can see this being potentially controversial is `RP25519(CompressedRistretto)`. However, since `CompressedRistretto` (which comes from the curve25519_dalek crate) is `Eq`, it seems safe to make our wrapper type `Eq` as well.

### DIFF
--- a/ipa-core/src/ff/curve_points.rs
+++ b/ipa-core/src/ff/curve_points.rs
@@ -25,7 +25,7 @@ impl Block for CompressedRistretto {
 /// since we always generate curve points from scalars (elements in Fp25519) and
 /// only deserialize previously serialized valid points, panics will not occur
 /// However, we still added a debug assert to deserialize since values are sent by other servers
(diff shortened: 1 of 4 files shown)


danielmasny (approved):
Makes sense to me. I guess the biggest difference between PartialEq and Eq is when a something is not a valid number. Converting an bytearray to an invalid CompressedRistretto point will result in an error.

PR #887, opened 2023-12-08
OPRF flow with all protocols plumbed in

richajaindce:
This plumbs in shuffle and sort
For sort, for now I have implemented a "sort key" which sorts based on timestamp, is_trigger and counter
Counter runs from 0 to num of rows per user which addresses #888

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/mod.rs
@@ -12,7 +12,7 @@ use crate::{
             boolean_ops::convert_to_fp25519,
             prf_eval::{eval_dy_prf, gen_prf_key},
             prf_sharding::{
-                attribute_cap_aggregate, compute_histogram_of_users_with_row_count,
+                attribute_cap_aggregate, compute_histogram_with_row_count_and_ranges_of_users,
                 PrfShardedIpaInputRow,
             },
         },
@@ -90,7 +90,7 @@ where
 
     prfd_inputs.sort_by(|a, b| a.prf_of_match_key.cmp(&b.prf_of_match_key));
 
-    let histogram = compute_histogram_of_users_with_row_count(&prfd_inputs);
+    let (histogram, ranges) = compute_histogram_with_row_count_and_ranges_of_users(&prfd_inputs);
     attribute_cap_aggregate::<C, BK, TV, TS, SS, Replicated<F>, F>(
         ctx,
         prfd_inputs,
--- a/ipa-core/src/protocol/ipa_prf/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/mod.rs
@@ -5,22 +5,28 @@ use ipa_macros::Step;
 use self::{quicksort::quicksort_ranges_by_key_insecure, shuffle::shuffle_inputs};
 use crate::{
     error::Error,
-    ff::{boolean::Boolean, boolean_array::BA64, CustomArray, Field, PrimeField, Serializable},
+    ff::{
+        boolean::Boolean,
+        boolean_array::{BA32, BA64},
+        CustomArray, Field, PrimeField, Serializable,
+    },
     protocol::{
         context::{UpgradableContext, UpgradedContext},
         ipa_prf::{
             boolean_ops::convert_to_fp25519,
             prf_eval::{eval_dy_prf, gen_prf_key},
             prf_sharding::{
-                attribute_cap_aggregate, compute_histogram_with_row_count_and_ranges_of_users,
-                PrfShardedIpaInputRow,
+                attribute_cap_aggregate, computations_post_shuffle, PrfShardedIpaInputRow,
             },
         },
         RecordId,
     },
     report::OprfReport,
     secret_sharing::{
-        replicated::{malicious::ExtendableField, semi_honest::AdditiveShare as Replicated},
+        replicated::{
+            malicious::ExtendableField, semi_honest::AdditiveShare as Replicated,
+            ReplicatedSecretSharing,
+        },
         SharedValue,
     },
 };
@@ -91,12 +97,12 @@ where
 
     prfd_inputs.sort_by(|a, b| a.prf_of_match_key.cmp(&b.prf_of_match_key));
 
-    let (histogram, ranges) = compute_histogram_with_row_count_and_ranges_of_users(&prfd_inputs);
+    let (histogram, ranges) = computations_post_shuffle(&mut prfd_inputs);
     quicksort_ranges_by_key_insecure(
         ctx.narrow(&Step::SortByTimestamp),
         &mut prfd_inputs,
         false,
-        |x| &x.timestamp,
+        |x| &x.sort_key,
         ranges,
     )
     .await?;
@@ -152,6 +158,7 @@ where
                 breakdown_key: record.breakdown_key,
                 trigger_value: record.trigger_value,
                 timestamp: record.timestamp,
+                sort_key: Replicated::new(BA32::ZERO, BA32::ZERO),
             })
         }
     }))
--- a/ipa-core/src/protocol/ipa_prf/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/mod.rs
@@ -1,9 +1,15 @@
 use std::num::NonZeroU32;
 
+use futures::stream::iter as stream_iter;
+use futures_util::StreamExt;
+use ipa_macros::Step;
+
+use self::{
+    prf_sharding::chunk_rows_by_user, quicksort::quicksort_by_key_insecure, shuffle::shuffle_inputs,
+};
 use crate::{
     error::Error,
     ff::{boolean::Boolean, boolean_array::BA64, CustomArray, Field, PrimeField, Serializable},
-    helpers::Role,
     protocol::{
         context::{UpgradableContext, UpgradedContext},
         ipa_prf::{
@@ -21,13 +27,6 @@ use crate::{
         replicated::{malicious::ExtendableField, semi_honest::AdditiveShare as Replicated},
         SharedValue,
     },
-};
-use futures::stream::iter as stream_iter;
-use futures_util::StreamExt;
-use ipa_macros::Step;
-
-use self::{
-    prf_sharding::chunk_rows_by_user, quicksort::quicksort_by_key_insecure, shuffle::shuffle_inputs,
 };
 
 mod boolean_ops;
@@ -95,8 +94,6 @@ where
     let mut prfd_inputs =
         compute_prf_for_inputs(ctx.narrow(&Step::ConvertInputRowsToPrf), shuffled).await?;
 
-    let histogram = compute_histogram_of_users_with_row_count(&prfd_inputs);
-
     prfd_inputs.sort_by(|a, b| a.prf_of_match_key.cmp(&b.prf_of_match_key));
     // Chunk the incoming stream of records into stream of vectors of records with the same PRF
     let mut input_stream = stream_iter(prfd_inputs);
@@ -108,29 +105,18 @@ where
     let rows_chunked_by_user = chunk_rows_by_user(input_stream, first_row);
 
     let batches = rows_chunked_by_user.collect::<Vec<_>>().await;
-    if ctx.role() == Role::H1 {
-         println!("num_batches = {:?}", batches.len());
-    }
+
     let mut sorted = Vec::new();
     let mut ctx_ts = ctx.clone();
     for mut batch in batches.into_iter() {
         ctx_ts = ctx_ts.narrow(&Step::SortByTimestamp);
         quicksort_by_key_insecure(ctx_ts.clone(), &mut batch, false, |x| &x.timestamp).await?;
-        // if ctx.role() == Role::H1 {
-            // println!("batch: {:?}", batch.iter().map(|i| i.timestamp.clone()).collect::<Vec<_>>());
-        // }
         sorted.push(batch);
     }
-    // if ctx.role() == Role::H1 {
-        println!("Before flatten num records {:?} {:?}", sorted.len(), ctx.role());
-    // }
 
-    let sorted : Vec<PrfShardedIpaInputRow<BK, TV, TS>> = sorted.into_iter().flatten().collect();
-    println!("sorted num records {:?}", sorted.len());
-    if ctx.role() == Role::H1 {
-        println!("sorted {:?}", sorted);
-    }
-    // TODO (richaj) : Call quicksort on match keys followed by timestamp before calling attribution logic
+    let sorted: Vec<PrfShardedIpaInputRow<BK, TV, TS>> = sorted.into_iter().flatten().collect();
+
+    let histogram = compute_histogram_of_users_with_row_count(&sorted);
     attribute_cap_aggregate::<C, BK, TV, TS, SS, Replicated<F>, F>(
         ctx,
         sorted,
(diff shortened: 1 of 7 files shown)


ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -84,18 +90,36 @@ where
...
+    let batches = rows_chunked_by_user.collect::<Vec<_>>().await;
 
-    let histogram = compute_histogram_of_users_with_row_count(&prfd_inputs);
+    let mut sorted = Vec::new();
+    let mut ctx_ts = ctx.clone();
+    for mut batch in batches.into_iter() {
+        ctx_ts = ctx_ts.narrow(&Step::SortByTimestamp);
+        quicksort_by_key_insecure(ctx_ts.clone(), &mut batch, false, |x| &x.timestamp).await?;
+        sorted.push(batch);
+    }
```
richajaindce, line 115: This part is really lame for now I am doing a sequential sort per pseudonym. Ideally we should be able to spawn multiple sorts per pseudonym maybe limit the number of sorts spawned to a LIMIT. Any thoughts on how to do that? @akoshelev @benjaminsavage @martinthomson

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -84,18 +89,36 @@ where
...
+    let rows_chunked_by_user = chunk_rows_by_user(input_stream, first_row);
 
-    let prfd_inputs =
-        compute_prf_for_inputs(ctx.narrow(&Step::ConvertInputRowsToPrf), input_rows).await?;
+    let batches = rows_chunked_by_user.collect::<Vec<_>>().await;
 
-    let histogram = compute_histogram_of_users_with_row_count(&prfd_inputs);
+    let mut sorted = Vec::new();
+    let mut ctx_ts = ctx.clone();
+    for mut batch in batches {
```
richajaindce, line 110: This part is really lame for now I am doing a sequential sort per pseudonym. Ideally we should be able to spawn multiple sorts per pseudonym maybe limit the number of sorts spawned to a LIMIT.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -84,18 +89,36 @@ where
...
+
+    prfd_inputs.sort_by(|a, b| a.prf_of_match_key.cmp(&b.prf_of_match_key));
+    // Chunk the incoming stream of records into stream of vectors of records with the same PRF
+    let mut input_stream = stream_iter(prfd_inputs);
+    let first_row = input_stream.next().await;
+    if first_row.is_none() {
+        return Ok(vec![]);
+    }
+    let first_row = first_row.unwrap();
+    let rows_chunked_by_user = chunk_rows_by_user(input_stream, first_row);
```
benjaminsavage, line 104: There is a bunch of identical looking code within `attribute_cap_aggregate`. Maybe you can just move the call to quicksort to be the first thing that happens within `evaluate_per_user_attribution_circuit`

richajaindce: I did this but fighting with set_total_records here. How should that be set?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -302,6 +305,12 @@ pub(crate) enum Step {
     MoveValueToCorrectBreakdown,
 }
 
+#[derive(Step)]
+pub(crate) enum SortByTimestamp {
+    #[dynamic(1024)]
```
richajaindce, line 310: Limits number of users to 1024

benjaminsavage (changes requested):
I don't want to land an approach that only scales to N users. We need a better long-term solution.

benjaminsavage:
OK, I've updated Quicksort: https://github.com/private-attribution/ipa/pull/892/files

This should work better for your purposes.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -84,15 +91,21 @@ where
...
-    let histogram = compute_histogram_of_users_with_row_count(&prfd_inputs);
-
-    // TODO (richaj) : Call quicksort on match keys followed by timestamp before calling attribution logic
+    let shuffled = shuffle_inputs(ctx.narrow(&Step::Shuffle), input_rows).await?;
+    let mut prfd_inputs =
+        compute_prf_for_inputs(ctx.narrow(&Step::ConvertInputRowsToPrf), shuffled).await?;
+
+    prfd_inputs.sort_by(|a, b| a.prf_of_match_key.cmp(&b.prf_of_match_key));
+
+    let (histogram, ranges) = compute_histogram_range_sort_key_per_user(&mut prfd_inputs);
```
benjaminsavage, line 100: This name of this function is too long

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -145,6 +158,7 @@ where
                 breakdown_key: record.breakdown_key,
                 trigger_value: record.trigger_value,
                 timestamp: record.timestamp,
+                sort_key: Replicated::new(BA32::ZERO, BA32::ZERO),
```
benjaminsavage, line 161: Can be written easier as Replicated::<BA32>::ZERO

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -178,7 +192,7 @@ pub mod tests {
                     trigger_value: 0,
                 },
                 TestRawDataRecord {
-                    timestamp: 0,
+                    timestamp: 5,
```
benjaminsavage, line 195: I don't think we should change the timestamps.

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -306,27 +341,42 @@ pub trait GroupingKey {
...
     let mut last_prf = input[0].get_grouping_key() + 1;
     let mut cur_count = 0;
-    for row in input {
+    let mut start = 0;
+    let mut ranges = vec![];
+    for (idx, row) in input.iter_mut().enumerate() {
         if row.get_grouping_key() == last_prf {
             cur_count += 1;
         } else {
+            // Don't push upto single rows of users
```
benjaminsavage, line 362: It's fine to do so as they'll just get filtered out. I would just remove this to keep the logic simpler.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -306,27 +341,42 @@ pub trait GroupingKey {
...
+    BK: SharedValue,
+    TV: SharedValue,
+    TS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
 {
     let mut histogram = vec![];
     let mut last_prf = input[0].get_grouping_key() + 1;
     let mut cur_count = 0;
-    for row in input {
+    let mut start = 0;
+    let mut ranges = vec![];
```
benjaminsavage, line 357: Can we allocate a vector with capacity? Do we have any way to estimate the total number of users?

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -46,6 +46,7 @@ pub(crate) enum Step {
 /// It terminates once the stack is empty.
 /// # Errors
 /// Will propagate errors from transport and a few typecasts
+#[tracing::instrument(name = "quicksort_by_key_insecure", skip_all)]
```
benjaminsavage, line 49: Can you update this to match the new function name?

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
martinthomson: Does this have to be done all at once?  It seems like it could be sharded, even within a shard.  (Starting with a single vector is probably fine for now, of course.)

richajaindce: The sort code which @benjaminsavage rewrote assumes that ALL the rows will come together to be sorted. Unfortunately, this takes away the opportunity to shard the sorting process

ipa-core/src/protocol/ipa_prf/mod.rs
richajaindce: This test says 2 sources have same timestamp and it should be attributed to 2nd breakdown which sounds a bit absurd

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
benjaminsavage: No, it's totally possible to shard this. You could make multiple calls to quicksort- and just provide each one with a different set of ranges. This could either be done across multiple machines, or multiple sequential calls from the same machine. We can probably convert this to a stream of ranges even since I just get an iterator and map it.

ipa-core/src/protocol/ipa_prf/mod.rs
benjaminsavage: Fair enough! Invalid input.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -915,20 +967,23 @@ pub mod tests {
                     breakdown_key: breakdown_key0,
                     trigger_value: trigger_value0,
                     timestamp: timestamp0,
+                    sort_key: Replicated::new(BA32::ZERO, BA32::ZERO),
```
benjaminsavage, line 970: Same thing can be done here in tests with the Replicated::<BA32>::ZERO

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
richajaindce: wont it also need having a different step name (if it happens on the same machine) since it might conflict with what infra has already seen?

ipa-core/src/protocol/ipa_prf/mod.rs
richajaindce: trying histograms_ranges_sortkeys :D

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
richajaindce: > Can we allocate a vector with capacity? Do we have any way to estimate the total number of users?

Worst case it could = number of input rows (if all rows are for unique users). No way of doing it until this part of the code.

martinthomson: Yes, you would need to have a step that sits above the sharding.  Or even completely new contexts (which I'd avoid).  But you aren't creating an unbounded number of shards, so that works within our current framework.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -43,6 +51,40 @@ pub struct PrfShardedIpaInputRow<BK: SharedValue, TV: SharedValue, TS: SharedVal
...
+impl<BK: SharedValue, TS, TV: SharedValue> PrfShardedIpaInputRow<BK, TV, TS>
+where
+    TS: SharedValue + ArrayAccess<Output = Boolean> + Expand<Input = Boolean>,
+{
+    /// This function defines the sort key.
+    /// The order of sorting is `timestamp`, `is_trigger_bit`, `counter`.
+    /// We sort by `is_trigger_bit` to ensure source events come before trigger in case there
+    /// is a tie in timestamp
+    /// Counter is added to ensure each sorting key is unique to avoid privacy leakage
+    /// NOTE: the order of members to be sorted by
```
benjaminsavage, line 66: NOTE: the sort key will be interpreted in Little endian format, so the order in which things are appended is important.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -43,6 +51,40 @@ pub struct PrfShardedIpaInputRow<BK: SharedValue, TV: SharedValue, TS: SharedVal
...
+    /// should be opposite to what you want as the bits are stored in Little Endian format.
+    /// We still need to add epoch which will be added later
+    pub fn compute_sort_key(&mut self, counter: u64) {
+        let mut y: Replicated<BA32> = Replicated::ZERO;
```
benjaminsavage, line 70: I think you can just assign to `self.sort_key = Replicated::ZERO`

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -43,6 +51,40 @@ pub struct PrfShardedIpaInputRow<BK: SharedValue, TV: SharedValue, TS: SharedVal
...
+        let mut offset = BA7::BITS as usize;
+
+        y.0.set(offset, self.is_trigger_bit.left());
+        y.1.set(offset, self.is_trigger_bit.right());
+
+        offset += 1;
+        expand_shared_array_in_place(&mut y, &self.timestamp, offset);
+        // TODO(richaj): add epoch to sort key computation
+
+        self.sort_key = y;
```
benjaminsavage, line 86: OK, this is pretty cool. Only downside is that we are now storing the information about the timestamp, epoch, and is_trigger_bit in two places on this struct. In a future PR we can probably consolidate this and only store it once.

ipa-core/src/protocol/step/steps.txt
```diff
@@ -7978,6 +7978,386 @@ ipa_core::protocol::ipa_prf::Step::convert_input_rows_to_prf/ipa_core::protocol:
...
+ipa_core::protocol::ipa_prf::Step::sort_by_timestamp/ipa_core::protocol::ipa_prf::quicksort::Step::quicksort_pass10/ipa_core::protocol::ipa_prf::quicksort::Step::compare/ipa_core::protocol::step::BitOpStep::bit15
+ipa_core::protocol::ipa_prf::Step::sort_by_timestamp/ipa_core::protocol::ipa_prf::quicksort::Step::quicksort_pass10/ipa_core::protocol::ipa_prf::quicksort::Step::compare/ipa_core::protocol::step::BitOpStep::bit16
+ipa_core::protocol::ipa_prf::Step::sort_by_timestamp/ipa_core::protocol::ipa_prf::quicksort::Step::quicksort_pass10/ipa_core::protocol::ipa_prf::quicksort::Step::compare/ipa_core::protocol::step::BitOpStep::bit17
+ipa_core::protocol::ipa_prf::Step::sort_by_timestamp/ipa_core::protocol::ipa_prf::quicksort::Step::quicksort_pass10/ipa_core::protocol::ipa_prf::quicksort::Step::compare/ipa_core::protocol::step::BitOpStep::bit18
+ipa_core::protocol::ipa_prf::Step::sort_by_timestamp/ipa_core::protocol::ipa_prf::quicksort::Step::quicksort_pass10/ipa_core::protocol::ipa_prf::quicksort::Step::compare/ipa_core::protocol::step::BitOpStep::bit19
```
benjaminsavage, line 8026: I don't understand why there are only 20 steps for comparison. That seems like it's only comparing BA20s, but our sort key is bigger now right? Shouldn't there be 32?

benjaminsavage (approved):
Hooray! The new OPRF based IPA protocol is now functional!

PR #889, opened 2023-12-11
Low hanging quicksort improvements

martinthomson:


benjaminsavage (approved):

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -71,43 +68,28 @@ where
...
+                    // reveal outcome of comparison
+                    // desc = true will flip the order of the sort
+                    Ok::<_, Error>(Boolean::from(false ^ desc) == comparison)
                 })),
             )
-            .try_for_each(|x| {
-                comp.push(x);
-                future::ready(Ok(()))
-            })
+            .try_collect()
```
danielmasny, line 92: @martinthomson I agree this is simpler. I have been working with @akoshelev on this. The reason why we went with the more complicated version is that it saves allocations. To my understanding Try_collect causes a reallocation of the bitvec (that's why I implemented it by hand and allocated the bitvec outside of the iterator).

martinthomson: Given that your `BitVec` is truly tiny, I don't see this as a cost that needs to be avoided.  And log_2(N) allocations here is hardly the largest cost we have to worry about.  And in time, perhaps, the compiler can learn that multiple calls to collect() in a loop can use the same memory allocation.

danielmasny: I agree in this case. Ben Savage actually merged an improvement to the quicksort implementation (using a breadth first quicksort parallelized over the recursion tree level and all user pseudonyms. I hope this resolves the main concerns with quicksort since it does not require that many different contexts with tiny record numbers anymore.) In that case, the bitvec will actually have up to query size many bits, so it might be in the range of a megabyte (for query size ~ 10 million)

PR #892, opened 2023-12-12
Quicksort with friendlier approach for our infra

benjaminsavage:
Previously, quicksort was setting the total number of records within the recursive steps, and narrowing multiple times. This made it impossible to use this method to sort, in parallel, many small lists.

This PR changes the Quicksort function to accept a vec of ranges that all need to be sorted. Now, there is only one call to "narrow" per quicksort pass (expected number of passes is just log(max_rows_per_user), with worst-case being `max_rows_per_user`. It's nice to have this upper bound on the total number of calls to "narrow".

Also, in this approach, we are able to process all of the records in parallel, across all users, in the same step.


ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -41,78 +37,117 @@ pub(crate) enum Step {
...
+                        Ok::<_, Error>(Boolean::from(desc) == comparison)
+                    }
+                });
+                counter += 1;
+            }
+        }
+
+        let comp = c.parallel_join(futures).await?;
+
+        let mut offset = 0;
```
benjaminsavage, line 115: Maybe comp_offset

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -345,4 +392,143 @@ pub mod tests {
...
+
+                // compute mpc sort
+                let result: Vec<_> = world
+                    .semi_honest(records.into_iter(), |ctx, mut records| async move {
+                        quicksort_ranges_by_key_insecure(
+                            ctx,
+                            &mut records[..],
+                            desc,
+                            |x| &x.timestamp,
+                            vec![
```
benjaminsavage, line 505: Generate this with a loop

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +51,101 @@ where
...
+        let cmp_ctx = c.narrow(&Step::Compare);
+        let rvl_ctx = c.narrow(&Step::Reveal);
+
+        let mut futures = Vec::with_capacity(num_comparisons_needed);
+        let mut counter = 0;
 
-    // iterate through quicksort recursions
-    while let Some((ctx, b_l, b_r)) = stack.pop() {
-        // start of quicksort function
         // check whether sort is needed
```
richajaindce, line 78: This comment should move down to line 82

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
-        if b_l + 1 < b_r {
+        for range in &ranges_to_sort {
+            if range.len() <= 1 {
```
richajaindce, line 82: So nice!

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
                     i += 1;
                 }
             }
+            n += range.len() - 1;
 
             // put pivot to index i-1
-            list.swap(i - 1, b_l);
+            list.swap(i - 1, range.start);
 
             // push recursively calls to quicksort function on stack
```
richajaindce, line 134:
```suggestion
            // push recursively calls to quicksort function 
```

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
+        let c = ctx
+            .narrow(&BitOpStep::from(quicksort_pass))
+            .set_total_records(num_comparisons_needed);
+        let cmp_ctx = c.narrow(&Step::Compare);
+        let rvl_ctx = c.narrow(&Step::Reveal);
 
-    // initialize stack
-    stack.push((ctx, 0usize, list.len()));
+        let mut futures = Vec::with_capacity(num_comparisons_needed);
+        let mut counter = 0;
```
richajaindce, line 78: should we rename this to current_record_idx?

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
-                ctx.active_work(),
-                stream_iter(iterator.enumerate().map(|(n, k)| async move {
-                    // Compare the current element against pivot and reveal the result.
-                    let comparison =
-                        compare_gt(pctx.narrow(&Step::Compare), RecordId::from(n), k, pivot)
+            for k in iterator {
+                futures.push({
+                    let record_id = RecordId::from(counter);
+                    let c1 = cmp_ctx.clone();
+                    let c2 = rvl_ctx.clone();
```
richajaindce, line 94:
```suggestion
                    let cmp_ctx = cmp_ctx.clone();
                    let cmp_ctx = rvl_ctx.clone();
```

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
+                        // reveal outcome of comparison
+                        // desc = true will flip the order of the sort
+                        Ok::<_, Error>(Boolean::from(desc) == comparison)
+                    }
+                });
+                counter += 1;
+            }
+        }
+
+        let comp = c.parallel_join(futures).await?;
```
richajaindce, line 112: Leave a TODO to convert this to seq_join later?

andyleiserson (approved):

danielmasny (approved):
This seems to be basically breadth first quicksort but parallelized over multiple pseudonym groups. It looks good to me. When we use vectorizations, we probably need to rethink this. I think there are a couple of allocations that might not be necessary but otherwise I did not notice anything significant.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -41,11 +39,12 @@ pub(crate) enum Step {
 /// It terminates once the stack is empty.
 /// # Errors
 /// Will propagate errors from transport and a few typecasts
-pub async fn quicksort_by_key_insecure<C, K, F, S>(
+pub async fn quicksort_ranges_by_key_insecure<C, K, F, S>(
     ctx: C,
     list: &mut [S],
     desc: bool,
     get_key: F,
+    mut ranges_to_sort: Vec<Range<usize>>,
```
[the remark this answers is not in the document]
benjaminsavage: Yes, that's how we will use this parameter. But I also re-use this param for each subsequent pass of quicksort as the logic is the same.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
+        // compute the total number of comparisons that will be needed
+        let mut num_comparisons_needed = 0;
+        for range in &ranges_to_sort {
+            if range.len() > 1 {
+                num_comparisons_needed += range.len() - 1;
+            }
+        }
+
+        let c = ctx
+            .narrow(&BitOpStep::from(quicksort_pass))
```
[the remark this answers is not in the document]
benjaminsavage: Yeah, I can use a new step type instead of re-using `BitOpStep`, that's fine with me. I'll specify a high maximum just in case, like 1000.

ipa-core/src/protocol/ipa_prf/quicksort.rs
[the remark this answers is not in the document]
benjaminsavage: That's true, but I need to pass in a real context to the `compare_gt` and `reveal` protocols. So I need to either call `.narrow()` on the pointer, or clone. I'm assuming that `clone` is cheaper than `narrow`.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
         // check whether sort is needed
-        if b_l + 1 < b_r {
+        for range in &ranges_to_sort {
+            if range.len() <= 1 {
+                continue;
+            }
             // set up iterator
-            let mut iterator = list[b_l..b_r].iter().map(get_key);
+            let mut iterator = list[range.clone()].iter().map(get_key);
             // first element is pivot, apply key extraction function f
```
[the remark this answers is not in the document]
benjaminsavage: `range` is a reference. I had to do this because removing the call to clone says:

```
the type `[S]` cannot be indexed by `&std::ops::Range<usize>`
```

This made me sad, but a `Range` is just a struct made of two `usize`, so I think it's a fairly cheap clone.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +53,96 @@ where
...
+
+        let c = ctx
+            .narrow(&BitOpStep::from(quicksort_pass))
+            .set_total_records(num_comparisons_needed);
+        let cmp_ctx = c.narrow(&Step::Compare);
+        let rvl_ctx = c.narrow(&Step::Reveal);
 
-    // initialize stack
-    stack.push((ctx, 0usize, list.len()));
+        let mut futures = Vec::with_capacity(num_comparisons_needed);
```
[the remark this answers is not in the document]
benjaminsavage: I managed to eliminate this vector entirely! yay!

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +60,93 @@ where
...
-                            .await?;
-
-                    // reveal outcome of comparison
-                    // desc = true will flip the order of the sort
-                    Ok::<_, Error>(Boolean::from(false ^ desc) == comparison)
-                })),
-            )
-            .try_collect()
-            .await?;
+    let mut ranges_for_next_pass = Vec::with_capacity(ranges_to_sort.len() * 2);
```
martinthomson, line 63: The size of this list needs to be longer than this, or are you choosing x2 on the basis of an expected value analysis?  Or do you just not care about depth.

ipa-core/src/protocol/ipa_prf/quicksort.rs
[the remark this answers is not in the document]
martinthomson: It's not clear to me whether these ranges are indices in the full list or not, see below.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +60,93 @@ where
...
+    let mut quicksort_pass = 1;
+
+    // iterate through all of the potentially incorrectly ordered ranges
+    // make one pass, comparing each element to the pivot and splitting into two more
+    // potentially incorrectly ordered ranges
+    while !ranges_to_sort.is_empty() {
+        // compute the total number of comparisons that will be needed
+        let mut num_comparisons_needed = 0;
+        for range in &ranges_to_sort {
+            if range.len() > 1 {
```
martinthomson, line 73: Is there *ever* a case where a range is zero length?  Because if you assume that ranges are all >= 1, you could do `ranges.map(|x| x.len()).sum() - ranges.len()`.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -54,65 +60,93 @@ where
...
+            if range.len() <= 1 {
+                continue;
+            }
 
             // swap elements based on comparisons
             // i is index of first element larger than pivot
-            let mut i = b_l + 1;
-            for (j, b) in comp.iter().enumerate() {
+            let mut i = range.start + 1;
+            for (j, b) in comp[n..(n + range.len() - 1)].iter().enumerate() {
```
martinthomson, line 127: I'm not clear on why you are iterating this way rather than having an iterator for the entirety of `comp` and doing: `for i in range.iter().skip(1) { if comp_iter.next() == Some(true) { swap } }` or similar.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -345,4 +386,148 @@ pub mod tests {
...
+                // compute mpc sort
+                let result: Vec<_> = world
+                    .semi_honest(records.into_iter(), |ctx, mut records| {
+                        let ranges_clone = ranges_to_sort.clone();
+                        async move {
+                            quicksort_ranges_by_key_insecure(
+                                ctx,
+                                &mut records[..],
+                                desc,
+                                |x| &x.timestamp,
```
martinthomson, line 511: You could make this a method on SillyStruct and use SillyStruct::timestamp, which is a pattern I would encourage real code to use.

ipa-core/src/protocol/ipa_prf/quicksort.rs
benjaminsavage: Depth is not relevant here. I am only allocating a vector of ranges that will need to be sorted in the next pass. That's expected to be <= the number of ranges in the current pass.

I perform this allocation once per pass, so this will happen in expectation log(max_rows_per_user) times.

ipa-core/src/protocol/ipa_prf/quicksort.rs
benjaminsavage: That would be invalid input. We can assume no zero length ranges. For callers of this API, ranges correspond to sequential records with the same PRF of the match key. So even if there is just one row with a given PRF at index i- the range to describe this is (i..i+1)

ipa-core/src/protocol/ipa_prf/quicksort.rs
benjaminsavage: That's a great idea! I think i can make that work! I'll give it a try.

ipa-core/src/protocol/ipa_prf/quicksort.rs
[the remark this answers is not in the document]
benjaminsavage: These are indices into the main list. Start of range is inclusive and end of range is exclusive, as per usual for Range. This is true not only for the first pass (user input) but also in subsequent passes which are sorting smaller sub-ranges of the list.

ipa-core/src/protocol/ipa_prf/quicksort.rs
benjaminsavage: What makes this very tricky is that the ranges are not necessarily adjacent, or tightly packed. Also, the indices of "comp" *are* tightly packed going from 0 upwards (they needed to be record_ids...).

ipa-core/src/protocol/ipa_prf/quicksort.rs
benjaminsavage: The computation you shared there is the number of comparisons in the first pass. For all subsequent passes, the ranges are no longer densely packed, and the count is different.

benjaminsavage: That's a great suggestion, I'll make that change.

ipa-core/src/protocol/ipa_prf/quicksort.rs
martinthomson: Hmm, so maybe you can iterate comp all at once, then co-iterate the ranges to find the indices you need.

```rust
let mut ranges = ranges.iter();
let mut range_it = ranges.next().unwrap(); // has to be at least one
let mut i = range_it.next().unwrap() + 1;  // first element of the range
for comparison in comp {
   let index = if let Some(n) = range_it.next() { n } else {
      range_it = ranges.next().unwrap().iter(); // has to be more here too
      // some other stuff to setup `i`
   };
   if comparison { swap }
}
```

benjaminsavage: I'm attempting to do this here: https://github.com/private-attribution/ipa/pull/895/files

Still not working yet, but do you like the way this code looks better? I think it's slightly longer now.

PR #895, opened 2023-12-14
Quicksort improvements

benjaminsavage:
Making some changes suggested by @martinthomson on #892. 

Currently tests do not pass... still debugging...

PR #896, opened 2023-12-15
Remove HRTB from OPRF functions

akoshelev:
Clippy was giving us hints but we ignored them.

The idea is to require `ArrayAccess` to implement `iter()` and leverage GAT to pass the lifetime of self.

This lets us remove all the `for <'a> &'a AdditiveShare<>: IntoIterator`.

Build is broken after #895


ipa-core/src/ff/boolean_array.rs
```diff
@@ -294,39 +301,29 @@ macro_rules! boolean_array_impl {
                 }
             }
 
-            // complains that no iter method exists, suppressed warnings
-            #[allow(clippy::into_iter_on_ref)]
             impl<'a> IntoIterator for &'a $name {
```
danielmasny, line 304: I was wondering why there is no default implementation for IntoIterator if all that we do is just calling self.iter()?

ipa-core/src/secret_sharing/replicated/semi_honest/additive_share.rs
```diff
@@ -255,6 +255,7 @@ where
     <S as ArrayAccess>::Output: SharedValue,
 {
     type Output = AdditiveShare<<S as ArrayAccess>::Output>;
+    type Iter<'a> = ASIterator<S::Iter<'a>>;
```
danielmasny, line 258: I love this. Why didn't we consider this earlier when we ran into the issue with trait bounds for references (i.e when I tried to require Into_Iterator for a reference to S)?

danielmasny (approved):
Looks good to me. I have two questions.

ipa-core/src/ff/boolean_array.rs
akoshelev: I think some tests still want it and I don't see harm adding it. Clippy is suggesting that calling `into_iter` on references may be confusing as it does not move the values out of the container. In our case, values are trivially copyable, so we can get away with it

ipa-core/src/secret_sharing/replicated/semi_honest/additive_share.rs
akoshelev: I think the combined effort to push boolean array and vectorization added some generic traits like `ArrayAccess`, `StdArray` that we can now leverage to provide a generic `iter` implementation for all shares over values that implement `ArrayAccess`

ipa-core/src/ff/galois_field.rs
```diff
@@ -132,12 +135,24 @@ fn clmul<GF: GaloisField>(a: GF, b: GF) -> u128 {
...
+    }
+}
+
 macro_rules! bit_array_impl {
     ( $modname:ident, $name:ident, $store:ty, $bits:expr, $one:expr, $polynomial:expr, $({$($extra:item)*})? ) => {
         #[allow(clippy::suspicious_arithmetic_impl)]
         #[allow(clippy::suspicious_op_assign_impl)]
         mod $modname {
-            use super::*;
+        use super::*;
```
akoshelev, line 155: formatting inside macros is broken

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -55,7 +54,6 @@ pub async fn integer_sat_add<C, S>(
 ) -> Result<AdditiveShare<S>, Error>
 where
     C: Context,
-    for<'a> &'a AdditiveShare<S>: IntoIterator<Item = AdditiveShare<S::Element>>,
```
benjaminsavage, line 58: I love this.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -74,13 +74,6 @@ where
     TV: SharedValue + CustomArray<Element = Boolean> + Field,
     TS: SharedValue + CustomArray<Element = Boolean> + Field,
     SS: SharedValue + CustomArray<Element = Boolean> + Field,
-    for<'a> &'a Replicated<SS>: IntoIterator<Item = Replicated<Boolean>>,
-    for<'a> &'a Replicated<TS>: IntoIterator<Item = Replicated<Boolean>>,
-    for<'a> &'a Replicated<TV>: IntoIterator<Item = Replicated<Boolean>>,
-    for<'a> &'a Replicated<BK>: IntoIterator<Item = Replicated<Boolean>>,
-    for<'a> <&'a Replicated<SS> as IntoIterator>::IntoIter: Send,
-    for<'a> <&'a Replicated<TV> as IntoIterator>::IntoIter: Send,
-    for<'a> <&'a Replicated<TS> as IntoIterator>::IntoIter: Send,
```
benjaminsavage, line 83: A thing of beauty! Thanks @akoshelev!

PR #898, opened 2023-12-18
non-negative DP (truncated double geometric) 

bmcase:
Adding initial code to generate non-negative DP noise from a truncated double geometric distribution which we will need for OPRF padding.


ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +58,92 @@ impl Distribution<f64> for RoundedBoxMuller {
...
 impl From<BoxMuller> for RoundedBoxMuller {
     fn from(value: BoxMuller) -> Self {
         Self { inner: value }
     }
 }
-
+/// Double Geometric Distribution
+///
+/// Generates a sample from a geometric distribution with the given success probability.
+fn generate_geometric<R: Rng + ?Sized>(probability: f64, rng: &mut R) -> isize {
```
akoshelev, line 69: this is a bit different than a geometric distribution implemented [here](https://docs.rs/rand_distr/latest/src/rand_distr/geometric.rs.html#32-37). If this is intentional, I would still recommend to follow their pattern, i.e. define a struct and implement `Distribution<u64>` for it

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +58,92 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    // Create a Bernoulli distribution with the specified success probability
+    let bernoulli = Bernoulli::new(probability).expect("Invalid probability");
+    // Generate Bernoulli random numbers until the first success
+    // let mut rng = rand::thread_rng();
+    let mut attempts = 0;
+    // let sample = ud.sample(rng);
+    while !bernoulli.sample(rng) {
+        attempts += 1;
+    }
+    attempts
```
akoshelev, line 79: this can't be negative, so maybe a fixed size integer is more appropriate here

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +58,92 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+}
+/// Truncated Double Geometric distribution.
+#[derive(Debug)]
+pub struct TruncatedDoubleGeometric {
+    s: f64,
+    shift: isize,
+}
+impl TruncatedDoubleGeometric {
+    /// Creates a new `TruncatedDoubleGeometric` distribution with the given success probability and shift parameter.
+    pub fn new(s: f64, shift: isize) -> Self {
```
akoshelev, line 108: likewise, it is better to reject invalid values during the construction phase, rather than panicking at `sample` call

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +58,92 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    attempts
+}
+/// Generates a sample from a double geometric distribution with the given success probability and shift parameter.
+fn generate_double_geometric<R: Rng + ?Sized>(s: f64, shift: isize, rng: &mut R) -> isize {
+    let success_probability = 1.0 - E.powf(-1.0 / s);
+    let attempts1 = generate_geometric(success_probability, rng);
+    let attempts2 = generate_geometric(success_probability, rng);
+    shift + attempts1 - attempts2
+}
+/// Generates a sample from a double geometric distribution with the given success probability and shift parameter.
```
akoshelev, line 88: code style suggestion (I don't know why format is not enforcing it) to leave a blank line between declarations - makes it a bit easier to read

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +166,116 @@ mod tests {
...
+        for _ in 0..num_samples {
+            let sample = generate_geometric(p, &mut rng);
+            *histogram.entry(sample).or_insert(0) += 1;
+        }
+        for x in 0..100 {
+            let observed_probability = histogram
+                .get(&x)
+                .map_or(0.0, |count| f64::from(*count) / f64::from(num_samples));
+            let expected_probability = (1.0 - p).powf(x as f64) * p;
+            // println!("x = {}, Observed Probability = {}, Expected Probability = {}", x, observed_probability, expected_probability);
```
akoshelev, line 187: you should probably uncomment those, by default cargo test suppresses stdout if test passes. So you won't see noise in your terminal

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +166,116 @@ mod tests {
...
+                "Observed probability is not within 1% of expected probability"
+            );
+        }
+    }
+    #[test]
+    fn test_truncated_double_geometric() {
+        let mut rng = rand::thread_rng();
+        let distribution = TruncatedDoubleGeometric { s: 1.0, shift: 25 };
+        distribution.sample(&mut rng);
+    }
```
akoshelev, line 280: you should probably test that invalid values for shift/probabilities are rejected

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +101,63 @@ impl DiscreteDp {
...
+    for n in big_delta.. {
+        if small_delta >= right_hand_side(n, big_delta, epsilon) {
+            return n as isize;
+        }
+    }
+    panic!("No smallest n found for OPRF padding DP");
+}
+
+impl OPRFPaddingDp {
+    pub fn new(new_epsilon: f64, new_delta: f64) -> Result<Self, Error> {
```
akoshelev, line 132: 👍

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +101,63 @@ impl DiscreteDp {
     }
 }
 
+///  Non-negative DP noise for OPRF padding
+///  Samples from a Truncated Double Geometric
+#[derive(Debug)]
+pub struct OPRFPaddingDp {
```
akoshelev, line 107: can you please also describe the functionality of it, i.e. how it samples and how it can be used to obtain the number of dummies to add (if that's what it is intended for)

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +101,63 @@ impl DiscreteDp {
...
+    epsilon: f64,
+    delta: f64,
+    truncated_double_geometric: TruncatedDoubleGeometric,
+}
+
+fn right_hand_side(n: i32, big_delta: i32, epsilon: f64) -> f64 {
```
akoshelev, line 113: same here, some documentation would be great to have here

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -239,4 +296,54 @@ mod test {
...
+        assert_eq!(find_smallest_n(1000,0.1,1e-08),1177);
+        assert_eq!(find_smallest_n(1000,0.01,1e-08),2773);
+    }
+    #[test]
+    fn test_oprf_padding_dp() {
+        let oprf_padding = OPRFPaddingDp::new(1.0, 1e-6);
+
+        let mut rng = rand::thread_rng();
+
+        oprf_padding.expect("REASON").sample(&mut rng);
```
akoshelev, line 347:
```suggestion
        oprf_padding.unwrap().sample(&mut rng);
```

ipa-core/src/protocol/dp/insecure.rs
bmcase: added documentation here and in dp/README.md

ipa-core/src/protocol/dp/insecure.rs
bmcase: added documentation in dp/README.md

ipa-core/src/protocol/dp/insecure.rs
bmcase: made other constructors like this

ipa-core/src/protocol/dp/distributions.rs
bmcase: added more tests like that

ipa-core/src/protocol/dp/distributions.rs
bmcase: updated to return Result type

ipa-core/src/protocol/dp/distributions.rs
bmcase: changed to u32

ipa-core/src/protocol/dp/distributions.rs
bmcase: defined a struct and implemented Disitribution<u32> for it

bmcase:
all checks and relevant tests passing.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+        // generate_truncated_double_geometric(self.s, self.shift, rng)
+        let mut reject = 1;
+        let mut sample = 0; // Declare sample here
+        while reject == 1 {
+            sample = self.double_geometric.sample(rng); // Assign a value to sample inside the loop
+            if sample >= 0 && sample <= (2 * self.shift).try_into().unwrap() {
+                reject = 0;
+            }
+        }
+        sample.try_into().unwrap() // Return the final value of sample
```
akoshelev, line 166: this can be just a loop I think

```rust
loop {
   let s = <compute sample>
   if s >= 0 ... {
     return s;
   }
}
```

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+            double_geometric: DoubleGeometric::new(s, shift)?,
+        })
+    }
+    /// Generates a sample from the `TruncatedDoubleGeometric` distribution.
+    pub fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> u32 {
+        // generate_truncated_double_geometric(self.s, self.shift, rng)
+        let mut reject = 1;
+        let mut sample = 0; // Declare sample here
+        while reject == 1 {
+            sample = self.double_geometric.sample(rng); // Assign a value to sample inside the loop
```
akoshelev, line 161: this comment is moot as code is self explanatory. It is recommended to avoid such comments because they distract readers

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+impl Distribution<i32> for DoubleGeometric {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i32 {
+        self.sample(rng)
+    }
+}
+
+/// Truncated Double Geometric distribution.
+#[derive(Debug, PartialEq)]
+pub struct TruncatedDoubleGeometric {
+    shift: u32, // is truncated to [0, 2*shift]
```
akoshelev, line 141: I am not sure I see where it is truncated, but I like the idea of doing the math once and inside the constructor

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +111,95 @@ impl DiscreteDp {
...
+    // See dp/README.md
+    pub fn new(new_epsilon: f64, new_delta: f64, new_sensitivity: u32) -> Result<Self, Error> {
+        // make sure delta and epsilon are in range, i.e. >min and delta<1-min
+        if new_epsilon < f64::MIN_POSITIVE {
+            return Err(Error::BadEpsilon(new_epsilon));
+        }
+
+        if !(f64::MIN_POSITIVE..=1.0 - f64::MIN_POSITIVE).contains(&new_delta) {
+            return Err(Error::BadDelta(new_delta));
+        }
```
akoshelev, line 181: I would probably add some tests that cover rejections too

ipa-core/src/protocol/dp/README.md
```diff
@@ -0,0 +1,62 @@
...
+To determine how many dummy elements to add we implement some of the non-negative DP distributions from this [paper](https://arxiv.org/abs/2110.08177) focusing initially on the Truncated Double Geometric.
+
+
+## Definition of Truncated Double Geometric
+From Section 3.2 of the [paper](https://arxiv.org/abs/2110.08177) we recall the definition of the Truncated Double Geometric mechanism and distribuiton. Consider a mechanism using this distribution:
+
+$$M(X) = f(X) + z; \quad z \sim p_{DoubleGeometric}(n) $$
+$$\textrm{Pr}_{DoubleGeometric}(x|n) = Ae^{-\epsilon|n-x|}; \qquad x\in \{0,\ldots,2n\}$$
+
+For some normalizing constant $0<A<1$ and some $n \in \mathbb{N}$.
```
akoshelev, line 17: this markdown is broken I think but huge +1 for adding this doc

akoshelev (changes requested):
minor feedback, very close to be pushed I think

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    pub fn new(probability: f64) -> Result<Self, Error> {
+        if probability < f64::MIN_POSITIVE {
+            return Err(Error::BadGeometricProb(probability));
+        }
+        Ok(Self { probability })
+    }
+    /// Generates a sample from the `Geometric` distribution.
+    pub fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> u32 {
+        // Create a Bernoulli distribution with the specified success probability
+        let bernoulli = Bernoulli::new(self.probability).expect("Invalid probability");
```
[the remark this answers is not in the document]
bmcase: yes, that will make more parallel with the rest

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+            shift,
+            geometric: Geometric::new(success_probability)?,
+        })
+    }
+    /// Generates a sample from the `DoubleGeometric` distribution
+    pub fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i32 {
+        // Self::generate_double_geometric(self.s, self.shift, rng)
+        let attempts1 = self.geometric.sample(rng);
+        let attempts2 = self.geometric.sample(rng);
+        (self.shift + attempts1 - attempts2).try_into().unwrap()
```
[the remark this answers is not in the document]
bmcase: do you mean putting in some actual checks on the size of shift and attemps1/2 or handling any such overflow differently that with `try_into()`?

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,142 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    }
+}
+impl Distribution<i32> for DoubleGeometric {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i32 {
+        self.sample(rng)
```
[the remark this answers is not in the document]
bmcase: ah that is nice and makes it more concise

ipa-core/src/protocol/dp/dp_tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
bmcase: thanks! yeah I wanted a quicker way to run just the relevant tests but didn't know about this easier way.

ipa-core/src/protocol/dp/distributions.rs
bmcase: ah, this comment is a bit confusing. It means the later samples will be truncated to within [0,2 * shift]. I'll move it to the sample implementation and make more clear.

ipa-core/src/protocol/dp/insecure.rs
bmcase: good call. added some more here

bmcase:
@akoshelev  made updates from your feedback.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,130 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+///  process samples from Double Geometric and Geometric distributions.
+
+///  Geometric Distribution
+/// Generates a sample from a geometric distribution with the given success probability.
+
+#[derive(Debug, PartialEq)]
+pub struct Geometric {
+    bernoulli: Bernoulli,
+}
+impl Geometric {
```
akoshelev, line 80: we seem to prefer empty lines before definitions - I don't know why formatter is not flagging this, but it would be good to follow to common style

akoshelev (approved):

ipa-core/src/protocol/dp/README.md
```diff
@@ -0,0 +1,62 @@
...
+$$ \delta \geq A \sum_{k=n-\Delta+1}^n  e^{-k \epsilon} $$
+
+
+For the common case of $\Delta=1$, such as in counting queries of users, at equality this simplifies (see appendix A of the cited paper) to:
+$$ \delta = Ae^{-n \epsilon} = Ar^n = \frac{r^n(1-r)}{1 + r -2r^{n+1}} $$
+$$\Rightarrow  \quad n = \Big\lceil -\frac{1}{\epsilon}\ ln \Big(\frac{\delta (1 + r)}{1-r + 2r\delta}\Big) \Big\rceil $$
+
+
+The geometric distribution is unwieldy analytically beyond $\Delta=1$ so we instead compute $n$ numerically.  We do this by letting
+$$ RHS = A \sum_{k=n-\Delta+1}^n  e^{-k \epsilon} $$
```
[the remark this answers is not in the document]
bmcase: went through and removed any other spaces like this.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,130 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+            geometric: Geometric::new(success_probability)?,
+        })
+    }
+}
+impl Distribution<i32> for DoubleGeometric {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i32 {
+        // self.sample(rng)
+        let attempts1 = self.geometric.sample(rng);
+        let attempts2 = self.geometric.sample(rng);
+        (self.shift + attempts1 - attempts2).try_into().unwrap()
```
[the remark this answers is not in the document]
bmcase: okay, added in the constructor of DoubleGeometric so we're not checking with every sample.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,130 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    }
+}
+impl Distribution<u32> for TruncatedDoubleGeometric {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> u32 {
+        // samples are truncated to be within [0, 2*shift]
+        loop {
+            let s = self.double_geometric.sample(rng);
+            if s >= 0 && s <= (2 * self.shift).try_into().unwrap() {
+                return s.try_into().unwrap();
+            }
```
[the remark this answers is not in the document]
bmcase: the probability of executing more times will depend on the parameters. I think if we have a small shift and also small success probability there will be a higher chance of looping here.  I can add another test with parameters more extreme in that direction.

bmcase:
@akoshelev fixed those last few things you mentioned.

ipa-core/src/protocol/dp/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+The geometric distribution is unwieldy analytically beyond $\Delta=1$ so we instead compute $n$ numerically.  We do this by letting
+$$RHS = A \sum_{k=n-\Delta+1}^n  e^{-k \epsilon}$$
+and then for a fixed set of DP parameters $\varepsilon$, $\delta$, $\Delta$ we find the smallest $n$ that makes this RHS larger than $\delta$. This gives the narrowest Truncated Double Geometric that provides this level of DP for the given parameters.
+
+
+## Sampling from a Truncated Double Geometric
+The process of drawing a sample from a Truncated Double Geometric will be done by sampling from a series of distributions
+1. We will assume access to samples from a Bernoulli distribiton as provided by the `rand` crate
```
martinthomson, line 41:
```suggestion
1. We will assume access to samples from a Bernoulli distribution as provided by the `rand` crate
```

martinthomson (commented):
Did you review the algorithm in https://arxiv.org/pdf/2004.00010.pdf for this?  The reason I suggest that is that it seems like there are shortcomings in some of the other implementations that might need to be accounted for and the algorithm they describe is pretty straightforward.

bmcase:
@martinthomson, I did review this paper earlier.  It seemed to me that the main shortcomings it addressed were with sampling from continuous distributions of DP noise and their main solution was to sample directly from discrete distributions (e.g. discrete Gaussian instead of continuous Gaussian).  For us we care about a different final discrete distribution (this truncated double geometric which is shifted to be non-negative).

I could have used their algorithm 2 to sample from a discrete laplace, which would (aside from shifting) be the double geometric here. But instead I found what I think is a simpler relation to use for transforming Bernoulli samples into Discrete Laplace samples.  It is a property that a double geometric random variable can be represented as $Y=\theta + X_1 - X_2$ where $X_1$ and $X_2$ are iid geometric variables with success probability $p = 1 - e^{-1/s}$.  So we can sample from the double geometric by first drawing two independent samples from $X_1$ and $X_2$ and then computing their difference plus the shift by $\theta$.

bmcase:
after discussing in the meeting, sounds like we are ready to merge this?  cc @akoshelev @martinthomson

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +126,98 @@ impl DiscreteDp {
...
+        if small_delta >= right_hand_side(n, big_delta, epsilon) {
+            return n;
+        }
+    }
+    panic!("No smallest n found for OPRF padding DP");
+}
+
+impl OPRFPaddingDp {
+    // See dp/README.md
+    pub fn new(new_epsilon: f64, new_delta: f64, new_sensitivity: u32) -> Result<Self, Error> {
```
martinthomson, line 188: You don't need to call your arguments new_whatever.  It's easier if the names match the field names.

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +126,98 @@ impl DiscreteDp {
...
+            return Err(Error::BadSensitivity(new_sensitivity));
+        }
+
+        // compute smallest shift needed to achieve this delta
+        let smallest_n = find_smallest_n(new_sensitivity, new_epsilon, new_delta);
+
+        Ok(Self {
+            epsilon: new_epsilon,
+            delta: new_delta,
+            sensitivity: new_sensitivity,
```
martinthomson, line 207: Do you need to keep epsilon and delta once you have the distribution?

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +126,98 @@ impl DiscreteDp {
...
+    for k in n - big_delta + 1..=n {
+        result += pow_u32(r, k);
+    }
+    a * result
+}
+fn find_smallest_n(big_delta: u32, epsilon: f64, small_delta: f64) -> u32 {
+    // for a fixed set of DP parameters, finds the smallest n that satisfies equation (11)
+    // of https://arxiv.org/pdf/2110.08177.pdf.  This gives the narrowest TruncatedDoubleGeometric
+    // that will satisify the disired DP parameters.
+    for n in big_delta.. {
```
martinthomson, line 178: This algorithm seems pretty crude.  If there isn't a simple inversion, could this at use a search?

I'm particularly concerned at this level because you have a lot of nested loops here when you trace this all the way down.

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -101,6 +126,98 @@ impl DiscreteDp {
...
+    // Computes the right hand side of equation (11) in https://arxiv.org/pdf/2110.08177.pdf
+    let r = E.powf(-epsilon);
+    let a = (1.0 - r) / (1.0 + r - 2.0 * (pow_u32(r, n + 1)));
+    let mut result = 0.0;
+    for k in n - big_delta + 1..=n {
+        result += pow_u32(r, k);
+    }
+    a * result
+}
+fn find_smallest_n(big_delta: u32, epsilon: f64, small_delta: f64) -> u32 {
```
martinthomson, line 174: Oh, BIG delta is $\Delta$ or the sensitivity, not a larger value of delta.  Maybe call it "sensitivity" throughout and stick with delta for $\delta$.

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -238,4 +355,72 @@ mod test {
...
+        assert_eq!(find_smallest_n(1000, 0.01, 1e-06), 2312);
+        assert_eq!(find_smallest_n(1000, 1.0, 1e-07), 1015);
+        assert_eq!(find_smallest_n(1000, 0.1, 1e-07), 1154);
+        assert_eq!(find_smallest_n(1000, 0.01, 1e-07), 2542);
+        assert_eq!(find_smallest_n(1000, 1.0, 1e-08), 1018);
+        assert_eq!(find_smallest_n(1000, 0.1, 1e-08), 1177);
+        assert_eq!(find_smallest_n(1000, 0.01, 1e-08), 2773);
+    }
+    #[test]
+    fn test_oprf_padding_dp() {
```
martinthomson, line 408: Test functions don't need to be called test_whatever, because they are always in a module called 'test'.

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -238,4 +355,72 @@ mod test {
...
+        let oprf_padding = OPRFPaddingDp::new(1.0, 1e-6, 10);
+
+        let mut rng = rand::thread_rng();
+
+        oprf_padding.unwrap().sample(&mut rng);
+    }
+    fn test_oprf_padding_dp_constructor() {
```
martinthomson, line 415: Did you intend to do this?

```suggestion
    #[test]
    fn test_oprf_padding_dp_constructor() {
```

ipa-core/src/protocol/dp/insecure.rs
```diff
@@ -1,18 +1,43 @@
...
+    )]
+    BadGeometricProb(f64),
+    #[error(
+        "Shift value over 1M -- likely don't need it that large and preventing to avoid any chance of overflow
+        in Double Geometric sample",
+    )]
+    BadShiftValue(u32),
+    #[error(
+        "Sensitivity value over 1M -- likely don't need it that large and preventing to avoid any chance of overflow
+        in Double Geometric sample",
```
martinthomson, line 33: I didn't see any risk of overflow until i32::MAX.  I would instead say "Sensitivity is higher than supported"  (Putting a 1M in this sort of message is a recipe for getting it wrong.  If you want the value in the message, I would use the constant instead, as above.)

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +216,191 @@ mod tests {
...
+                .get(&x)
+                .map_or(0.0, |count| f64::from(*count) / f64::from(num_samples));
+            // let expected_probability =
+            // normalizing_factor * E.powf(-epsilon * ((n - x) as i32 ).abs() as f64);
+            let expected_probability = normalizing_factor
+                * (if x <= n {
+                    E.powf(-epsilon * f64::from(n - x))
+                } else {
+                    E.powf(-epsilon * f64::from(x - n))
+                });
```
martinthomson, line 398:
```suggestion
                * (E.powf(-epsilon) + E.powf(abs(f64::from(n) - f64::from(x))));
```

If you upconvert the raw values, you can just use abs, just like in the comment.  And delete the comment.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +216,191 @@ mod tests {
...
+        let mut histogram = HashMap::new();
+        for value in samples {
+            *histogram.entry(value).or_insert(0) += 1;
+        }
+        let mut sorted_keys: Vec<u32> = histogram.keys().copied().collect();
+        sorted_keys.sort_unstable();
+        // Compute the expected probability for each value in the range [0, 2*n]
+        #[allow(clippy::cast_precision_loss)]
+        let normalizing_factor = (1.0 - E.powf(-epsilon))
+            / (1.0 + E.powf(-epsilon) - 2.0 * E.powf(-epsilon * (f64::from(n + 1)))); // 'A' in paper
```
martinthomson, line 384: Lots of use of `E.powf(-epsilon)` here.  Consider creating a temporary to hold that value.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +216,191 @@ mod tests {
...
+            *histogram.entry(sample).or_insert(0) += 1;
+        }
+        #[allow(clippy::cast_precision_loss)]
+        for x in 0..100 {
+            let observed_probability = histogram
+                .get(&x)
+                .map_or(0.0, |count| f64::from(*count) / f64::from(num_samples));
+            let expected_probability = (1.0 - p).powf(f64::from(x)) * p;
+            println!("x = {x}, Observed Probability = {observed_probability}, Expected Probability = {expected_probability}");
+            assert!((observed_probability - expected_probability) <= 0.01);
```
martinthomson, line 248: You should set your tolerance value and number of iterations as constants.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +216,191 @@ mod tests {
...
+    fn test_truncated_double_geometric_constructor() {
+        // should fail for negative s parameter
+        let mut s = -1.0;
+        let mut n = 25;
+        let mut expected = Err(Error::BadS(s));
+        let mut actual = TruncatedDoubleGeometric::new(s, n);
+        assert_eq!(expected, actual);
+        s = 2.0;
+        n = 3_000_000;
+        expected = Err(Error::BadShiftValue(n));
```
martinthomson, line 272: Use literals for expectations.  (Generally, you should avoid mut variables, get too accustomed to mutability and you will make mistakes.)

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -125,4 +216,191 @@ mod tests {
...
+        actual = TruncatedDoubleGeometric::new(s, n);
+        assert_eq!(expected, actual);
+    }
+    #[test]
+    fn test_truncated_double_geometric() {
+        let mut rng = rand::thread_rng();
+        let s = 1.0;
+        let n = 25;
```
martinthomson, line 280: constants can use const rather than let.

martinthomson (commented):
The general shape of this is reasonable, though I think it could use another iteration of polish.

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -1,18 +1,22 @@
...
+
 /// Returns `true` iff `a` and `b` are close to each other. `a` and `b` are considered close if
 /// |a-b| < 10^(-precision).
 #[cfg(all(test, unit_test))]
 pub fn is_close(a: f64, b: f64, precision: u8) -> bool {
     (a - b).abs()
         < (2.0_f64.powf((a.abs() + 1.0).log2().ceil()) / 10.0_f64.powi(i32::from(precision)))
 }
-
 /// Normal distribution based on [`Box-Muller`] transform.
```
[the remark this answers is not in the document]
bmcase: I don't think I touched this part of the code unless the fmt script changed something, but I think what you're saying is consistent with what Alex mention of wanting a new line between declarations.  Will try to add those wherever lacking

ipa-core/src/protocol/dp/distributions.rs
```diff
@@ -63,44 +61,139 @@ impl Distribution<f64> for RoundedBoxMuller {
...
+    }
+}
+
+impl Distribution<i32> for DoubleGeometric {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i32 {
+        let attempts1 = self.geometric.sample(rng);
+        let attempts2 = self.geometric.sample(rng);
+        <u32 as TryInto<i32>>::try_into(self.shift).unwrap()
+            + <u32 as TryInto<i32>>::try_into(attempts1).unwrap()
+            - <u32 as TryInto<i32>>::try_into(attempts2).unwrap()
```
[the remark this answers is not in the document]
bmcase: that's an interesting idea.  I'd have to think a bit more how to prove what this ends up being but I plotted it and it doesn't look the same.  Even if I let in this way `self.geometric.sample(rng)` have a success probably that is half as much as the success probability in each of X_1 - X_2. 

Here's a python to look at the differences of the two sample distributions: 

`import numpy as np
import random


p1 = 0.3
p2 = 0.15
numsamples = 1000000

x_1 = np.random.geometric(p1,numsamples)
x_2 = np.random.geometric(p1,numsamples)

# compute first as X_1 - X_2
first = x_1 - x_2

firstcounts = {}
for d in first:
    if d in firstcounts:
        firstcounts[d] +=1
    else:
        firstcounts[d] = 1


# compute second as X_3 * (1 - 2 * Z)
second = []
for i in range(numsamples):
    z = bool(random.getrandbits(1))
    x_3 = np.random.geometric(p2)

    second.append(x_3 * (1 - 2 * z))

secondcounts = {}
for d in second:
    if d in secondcounts:
        secondcounts[d] +=1
    else:
        secondcounts[d] = 1

bound = 80
for i in range(-bound,bound+1):
    if i not in firstcounts:
        firstcounts[i] = 0
    if i not in secondcounts:
        secondcounts[i] = 0

for i in range(-bound,bound+1):
    # print(i,firstcounts[i],secondcounts[i],firstcounts[i] - secondcounts[i])
    print(i,firstcounts[i] - secondcounts[i])
`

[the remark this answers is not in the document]
bmcase: you can see the difference is not random but structured. 
image

ipa-core/src/protocol/dp/distributions.rs
bmcase: ah, that nice.  had various type conversion troubles before which led to the very manual way of doing it.

ipa-core/src/protocol/dp/distributions.rs
bmcase: indeed, in fact in the paper we used r = e^{-epsilon} to clean up the presentation.  don't know why I didn't keep that here

ipa-core/src/protocol/dp/distributions.rs
bmcase: okay

ipa-core/src/protocol/dp/distributions.rs
bmcase: but then I would have to make it upper case to make the warning go away... and that would make it look different than the parameters it matches to in the function call.

ipa-core/src/protocol/dp/distributions.rs
bmcase: not quite sure what you're looking for by literals here. Something like `assert_eq!(Err(Error::BadS(s)), actual);`

ipa-core/src/protocol/dp/distributions.rs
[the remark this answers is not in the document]
martinthomson: Interesting result.  I had a little trouble working out what the concrete differences were here.  Because your difference of geometric distributions is not a symmetric version of the same.  And I couldn't work out if this was really a discrete Laplace or not either.

ipa-core/src/protocol/dp/distributions.rs
martinthomson: I'm suggesting that instead of

```rust
let expected = foo;
assert_eq!(expected, bar);
```

You just do this:

```rust
assert_eq!(foo, bar);
```

ipa-core/src/protocol/dp/distributions.rs
martinthomson: Yes, but that's how you do these things in languages with naming conventions.

The way we generally avoid that style of problem is to give things names rather than relying on letters.

ipa-core/src/protocol/dp/distributions.rs
bmcase: okay, I'll update to that here and these related tests

ipa-core/src/protocol/dp/distributions.rs
bmcase: alright, that makes sense.  maybe what I'll do here is what I have in the other test functions of testing more combinations of parameters and update `s` and `n`

ipa-core/src/protocol/dp/distributions.rs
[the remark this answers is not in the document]
bmcase: yeah, it it interesting. I have a reference that `X_1 - X_2` for iid geometrics will have an equivalent probability density to a discrete Laplace, but there the proof is left as an exercise :)  Be nice to understand but I haven't sat down to try and prove it directly.  I created a test that the sample distribution of this `X_1 - X_2` matches the expected probability distribution.

akoshelev:
> tests are failing in unrelated places. @akoshelev do you know what is going on? I rebased to have the latest but still errors from other places. `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.7.7/src/lib.rs:33:42`

yea I think main branch is broken. https://github.com/private-attribution/ipa/pull/942 should fix it

akoshelev:
I think it helps only with lint errors. It seems that nightly is broken again

```bash
error[E0635]: unknown feature `stdsimd`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.7.7/src/lib.rs:33:42
   |
33 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
```

akoshelev:
I fixed the issue in #945, CI should be green now

bmcase:
thanks @akoshelev , yep that fixed it.  @martinthomson  is there anything else you would like to see done here at this point? Do you think we need to prove the exercise that X_1 - X_2 for iid geometrics will be a double geometric?

martinthomson:
I don't think that we need that proof to get some of the structure down.  After all, if we do some more analysis and decide that the distribution doesn't work for us, that's a fairly simple change.  Maintaining this as a PR indefinitely isn't simple.  (I'm still not entirely satisfied, but I don't really have the time to work through the different PMFs to determine where the remaining discrepancies are.)

akoshelev:
what is the conclusion here @martinthomson, @bmcase ? Should we merge it as is and create follow-up issues?

martinthomson:
I'm happy to merge and iterate.

bmcase:
@akoshelev  resolved the merge conflicts that came up here.  Can we go ahead and merge now?

PR #902, opened 2023-12-20
fixing busted stuff

benjaminsavage:
While #900 unblocked the build, it did so in a way that broke the way the "histogram" was computed. This reverts those changes, while continuing to ensure that it doesn't generate any invalid ranges like 0..0 - so now the tests pass.


akoshelev (approved):
👍

akoshelev:
this fixes #900

Issue #884, opened 2023-12-07
Support variable-size generation from PRSS


andyleiserson:
Currently PRSS generation is fixed size in that it always consumes 128 bits of entropy. The actual value generated can be smaller (by discarding some of the entropy), and with effort larger (by instantiating another PRG to expand the entropy). Generation of small values is currently wasteful, in that it pays to generate bits from an AES PRG that it doesn't need. Generation of larger values is becoming necessary to support Curve25519 and vectorization.

This was originally suggested in https://github.com/private-attribution/ipa/issues/812#issuecomment-1825019292, and there is some discussion in https://github.com/private-attribution/ipa/pull/877/files#r1416577716.

andyleiserson:
Copying @danielmasny's comment from https://github.com/private-attribution/ipa/pull/877#discussion_r1423045616:

> isn't that reimplementing Distribution with the only difference using a seed instead of a RNG? We could also consider having PRSS returning an RNG (and implement an RNG that panics when we request too many bits). This would allow us to use the standard rust conventions.
>
> Here with panic, I mean debug panic such that we could identify when we requesting AES evaluations for identical inputs twice.

andyleiserson:
> Isn't that reimplementing Distribution with the only difference using a seed instead of a RNG?

It never does any expansion of the entropy provided. i.e. if we're generating a BA256, the input type is `[u128; 2]`. If we're generating a hypothetical BA1024, the input type would be `[u128; 8]`, and so on.

> This would allow us to use the standard rust conventions.

Yeah, it would be nice to do that. That's what I was thinking about when I wrote "I don't see it as having any fundamental difference from rand's `Distribution<Standard>`" in the description of #877. Although I later reassessed and decided that I do see a fundamental difference: stream-style consumption or entropy (APIs that fill an arbitrary slice with random bits or that can be called any number of times) vs. consumption of a fixed amount of entropy. I guess we could implement in terms of "panic if it requests too many bits", but I think I would prefer that the amount of entropy be exposed explicitly (via the size of the array) rather than just taking an index as an argument, which seems like it has the same drawbacks as the `2*index` approach, but with less visibility of the number of indexes that are going to be used.

andyleiserson:
[...]

PR #903, opened 2023-12-20
Framework for variable-size PRSS generation

andyleiserson:
There is some relevant background discussion on #884 and #852.

This addresses some, but not all, of the remaining cleanup for #812 -- this makes BA256 not a `Field`, which happens to be doable because it's not used as one anywhere. We also need to make the smaller BAs not `Field`, but that requires addressing places where they are used as `Field`s.

BA256 provides a sample of generating something with PRSS that needs more than 128 random bits, but this will also be needed for vectorization.

This doesn't address unnecessary generation of 128 random bits even when the value being generated is smaller. Once things are vectorized that will be less of an issue.


ipa-core/src/protocol/prss/crypto.rs
```diff
@@ -24,16 +27,38 @@ pub trait FromRandomU128 {
...
 /// `rand::distributions::Standard`, a uniform distribution is typical. When implementing
 /// this trait, consider the consequences if the implementation were to be used in
 /// an unexpected way. For example, an implementation that draws from a subset of the
 /// possible values could be dangerous, if used in an unexpected context where
 /// security relies on sampling from the full space.
+pub trait FromRandom: Sized {
+    type SourceLength: ArrayLength;
+
+    /// Generate a random value of `Self` from `SourceLength` uniformly-distributed u128s.
+    fn from_random(src: GenericArray<u128, Self::SourceLength>) -> Self;
```
akoshelev, line 48: curious about the choice of `u128` for container? why not `u8`?

akoshelev (approved):

ipa-core/src/protocol/prss/crypto.rs
andyleiserson: I used `u128` because it matches what we get from PRSS. I'd be fine with using `u8`. The only hesitation I have about `u8` is that it hides the fact that the randomness is generated in multiples of 128 bits, i.e., an implementation that takes a single `u8` is much less efficient than it looks.

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -32,7 +32,7 @@ pub async fn integer_add<C, XS, YS>(
 where
     C: Context,
     YS: SharedValue + CustomArray<Element = XS::Element>,
-    XS: SharedValue + CustomArray + Field,
+    XS: SharedValue + CustomArray + SharedValue,
```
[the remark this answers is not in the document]
andyleiserson: Not sure if this is why you left the comment, but I just noticed this now says `SharedValue + CustomArray + SharedValue`. Oops.

ipa-core/src/protocol/prss/crypto.rs
akoshelev: agree, it is better to indicate that it is a multiple of 16 bytes

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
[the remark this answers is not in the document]
akoshelev: yea I was impressed that Rust allows to specify that :)

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -60,6 +64,65 @@ impl Debug for UsedSet {
...
+impl From<u32> for PrssIndex {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+// It would be nice for this to be TryFrom, but there's a lot of places where we use u128s as PRSS indexes.
+impl From<u128> for PrssIndex {
+    fn from(value: u128) -> Self {
+        Self(value.try_into().unwrap())
```
martinthomson, line 109: This would be a great place to use .expect() instead of .unwrap().

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -60,6 +64,65 @@ impl Debug for UsedSet {
...
+impl From<RecordId> for PrssIndex {
+    fn from(value: RecordId) -> Self {
+        Self(u32::from(value))
+    }
+}
+
+impl PrssIndex {
+    fn offset(self, offset: usize) -> PrssIndex128 {
+        assert!(offset <= u16::MAX.into());
+        PrssIndex128::from(u128::from((u64::from(self.0) << 16) + (offset as u64)))
```
martinthomson, line 122: Do you really want to only use 16 bits for the offset, given that this is a 32-bit value?

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -569,7 +638,7 @@ pub mod test {
     #[test]
     #[cfg(debug_assertions)]
     #[should_panic(
-        expected = "Generated randomness for index '100' twice using the same key 'protocol/test'"
+        expected = "Generated randomness for index '6553600' twice using the same key 'protocol/test'"
```
martinthomson, line 641: any sense in implementing Display for PrssIndexU128 to make this less of a pain to use?

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -75,14 +138,20 @@ pub struct IndexedSharedRandomness {
...
+        &self,
+        index: I,
+    ) -> (GenericArray<u128, N>, GenericArray<u128, N>) {
         let index = index.into();
         #[cfg(debug_assertions)]
         {
-            self.used.insert(index);
+            for i in 0..N::USIZE {
+                self.used.insert(index.offset(i));
+            }
```
martinthomson, line 150: I might suggest that we only need to track against `PrssIndex` for this.  It will greatly reduce the tracking overhead.

ipa-core/src/protocol/prss/mod.rs
andyleiserson: Do you think 32 makes more sense? There was also @danielmasny's suggestion of 64 in https://github.com/private-attribution/ipa/issues/884#issuecomment-1854926221.

Implementing `Display` will eliminate one reason for minimizing it, but it also seemed unlikely that any single invocation of PRSS would generate a value with more than 2^23 bits of entropy? I considered as little as 2^15. However, 2^15 bits is a 4,096-byte vector. That is the upper end of the range I foresee using, but I wanted to allow flexibility to go a bit bigger than that.

ipa-core/src/protocol/prss/mod.rs
andyleiserson: It's probably sufficient, but aren't we already planning to disable this tracking in production configurations? In which case, I'd rather be cautious and do the more extensive check, unless we determine it's a huge hit to our dev runtimes.

Vectorization will shift usage from the index bits to the offset bits, but the overall number of entries added to the map will stay the same. i.e. this change does not increase the size requirement for the tracking structure.

ipa-core/src/protocol/prss/mod.rs
martinthomson: Keeping the values within a 64-bit range is fine, possibly even a good idea.

ipa-core/src/protocol/prss/mod.rs
martinthomson: Just a thought.  The API doesn't let you take an index plus a range of offsets that don't start at zero, so I figured... why not.

You might get similar coverage by only checking 0 and N-1, FWIW.

PR #904, opened 2023-12-21
Add Encryption and Decryption for OPRF Report Structs

danielmasny:


danielmasny:
I will use slices instead of generic arrays to fix the trait bounds. So don't review it yet.

danielmasny:
> I will use slices instead of generic arrays to fix the trait bounds. So don't review it yet.

I fixed it, its ready for review now.

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+    Replicated<BK>: Serializable,
+    Replicated<TV>: Serializable,
+    Replicated<TS>: Serializable,
+{
+    const ENCAP_KEY_MK_OFFSET: usize = 0;
+    const CIPHERTEXT_MK_OFFSET: usize = Self::ENCAP_KEY_MK_OFFSET + EncapsulationSize::USIZE;
+    // need to round up ciphertext length to nearest multiple of 16 since AES has block length 16
+    const ENCAP_KEY_BTT_OFFSET: usize = (Self::CIPHERTEXT_MK_OFFSET
+        + U16::USIZE
+        + ((<Replicated<BA64> as Serializable>::Size::USIZE) + 15))
```
andyleiserson, line 471: I'd suggest formalizing the assumption that the match key is `BA64` by defining something like `pub type OprfIpaMatchKey = BA64`.

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+    BK: SharedValue,
+    TV: SharedValue,
+    TS: SharedValue,
+{
+    data: B,
+    phantom_data: PhantomData<(BK, TV, TS)>,
+}
+
+/// follows the outline of the implementation of `EncryptedReport`
+impl<B, BK, TV, TS> EncryptedOprfReport<BK, TV, TS, B>
```
andyleiserson, line 456: We eventually want at least BK, and maybe also TV and TS to be runtime parameters, right? Are we tracking that work somewhere? (I think it will be a fair amount of work -- certainly not something that makes sense to do in this PR.)

ipa-core/src/report.rs
```diff
@@ -110,6 +110,25 @@ impl From<&EventType> for u8 {
...
+impl From<&EventType> for Replicated<Boolean> {
+    fn from(value: &EventType) -> Self {
+        Replicated(
+            Boolean::from(*value == EventType::Trigger),
+            Boolean::from(*value == EventType::Trigger),
+        )
+    }
+}
+
+impl From<&Replicated<Boolean>> for EventType {
```
andyleiserson, line 122: Is this for tests? It doesn't seem safe to make this generally available.

ipa-core/src/report.rs
```diff
@@ -468,73 +671,205 @@ where
...
+    ) -> Result<Vec<u8>, InvalidReportError> {
+        let mut out = Vec::new();
+        self.encrypt_to(key_id, key_registry, rng, &mut out)?;
+        debug_assert_eq!(out.len(), usize::from(self.encrypted_len()));
+        Ok(out)
+    }
+
+    /// # Errors
+    /// If there is a problem encrypting the report.
+    pub fn encrypt_to<R: CryptoRng + RngCore, B: BufMut>(
```
andyleiserson, line 798: This can also benefit from `GenericArray` type inference and `BA64` type alias.

ipa-core/src/report.rs
```diff
@@ -468,73 +671,205 @@ where
...
+
+    /// # Errors
+    /// If there is a problem encrypting the report.
+    pub fn encrypt<R: CryptoRng + RngCore>(
+        &self,
+        key_id: KeyIdentifier,
+        key_registry: &impl PublicKeyRegistry,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, InvalidReportError> {
+        let mut out = Vec::new();
```
andyleiserson, line 790: I missed it in the original, we could do `Vec::with_capacity(self.encrypted_len())` here to possibly avoid some reallocation overhead.

ipa-core/src/report.rs
```diff
@@ -56,8 +56,8 @@ impl Serializable for EventType {
 
     fn serialize(&self, buf: &mut GenericArray<u8, Self::Size>) {
         let raw: &[u8] = match self {
-            EventType::Trigger => &[0],
-            EventType::Source => &[1],
+            EventType::Trigger => &[1],
+            EventType::Source => &[0],
```
andyleiserson, line 60: I think it will be an explicit merge conflict, but just a heads up that this may interact with the fallible serialization PR.

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+    BK: SharedValue,
+    TV: SharedValue,
+    TS: SharedValue,
+    Replicated<BK>: Serializable,
+    Replicated<TV>: Serializable,
+    Replicated<TS>: Serializable,
+{
+    const ENCAP_KEY_MK_OFFSET: usize = 0;
+    const CIPHERTEXT_MK_OFFSET: usize = Self::ENCAP_KEY_MK_OFFSET + EncapsulationSize::USIZE;
+    // need to round up ciphertext length to nearest multiple of 16 since AES has block length 16
```
andyleiserson, line 468: Our current AEAD is AES-GCM, which does not expand the ciphertext to a multiple of the block size, but coincidentally `Replicated<BA64>` uses exactly one block. To work with other sizes something like `BTT_END` would be needed in the encryption routine.

If we do have padding, my preference would be to reject inputs where the padding is not all zero. Given that we want to make some of these values runtime-configurable in the future, padding does give us a bit of flexibility to tweak the format, although my preference would be to have a version indicator at the very start of the report.

andyleiserson (changes requested):
Looks good overall, I left a few suggestions.

andyleiserson:
Oh, and I'd suggest either cleaning up the commit history locally (I use `git rebase -i`), or using the squash-merge option in github.

ipa-core/src/report.rs
akoshelev: I agree. I think the size of the ciphertext is simply ENC_KEY + PAYLOAD + AAD_TAG. I would prefer to get the tag size and key size from HPKE module rather than hardcoding them here

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+    const EVENT_TYPE_OFFSET: usize = (Self::CIPHERTEXT_BTT_OFFSET
+        + U16::USIZE
+        + ((<Replicated<BK> as Serializable>::Size::USIZE
+            + <Replicated<TV> as Serializable>::Size::USIZE
+            + <Replicated<TS> as Serializable>::Size::USIZE)
+            + 15))
+        & (!15);
+    const SITE_DOMAIN_OFFSET: usize = Self::EVENT_TYPE_OFFSET + 4;
+
+    const TS_OFFSET: usize = 0;
```
akoshelev, line 485: is the offset actually 0 here?

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
 }
 
+/// A binary report as submitted by a report collector, containing encrypted `OprfReport`
+/// An `EncryptedOprfReport` consists of:
+///     `ct_mk`: Enc(`match_key`)
+///     `ct_btt`: Enc(`breakdown_key`, `trigger_value`, `timestamp`)
+///     associated data of `ct_mk`: `key_id`, `epoch`, `event_type`, `site_domain`,
+
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct EncryptedOprfReport<BK, TV, TS, B>
```
akoshelev, line 444: I would strongly recommend to add the report structure to the description similar to IPA report

```rust
// Report structure:
//  * 0..4: `timestamp`
//  * 4: `breakdown_key`
//  * 5..a: `trigger_value`
//  * a..b: `encap_key`
//  * b..c: `mk_ciphertext`
//  * c: `event_type`
//  * c+1: `key_id`
//  * c+2..c+4: `epoch`
//  * c+4..: `site_domain`
```

reading through the code to understand report structure is not ideal

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+        &self.data[Self::CIPHERTEXT_BTT_OFFSET..Self::EVENT_TYPE_OFFSET]
+    }
+
+    /// ## Panics
+    /// Only if a `Report` constructor failed to validate the contents properly, which would be a bug.
+    pub fn event_type(&self) -> EventType {
+        EventType::try_from(self.data[Self::EVENT_TYPE_OFFSET]).unwrap() // validated on construction
+    }
+
+    pub fn key_id(&self) -> KeyIdentifier {
```
akoshelev, line 512: you should encode key identifier offset as const too I think.

akoshelev: Just looked down below - and epoch as well.

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+
+    /// ## Panics
+    /// Only if a `Report` constructor failed to validate the contents properly, which would be a bug.
+    pub fn site_domain(&self) -> &str {
+        std::str::from_utf8(&self.data[Self::SITE_DOMAIN_OFFSET..]).unwrap() // validated on construction
+    }
+
+    /// ## Errors
+    /// If the report contents are invalid.
+    pub fn from_bytes(bytes: B) -> Result<Self, InvalidReportError> {
```
akoshelev, line 534: maybe we can validate the size as well?

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+        let info = Info::new(
+            self.key_id(),
+            self.epoch(),
+            self.event_type(),
+            HELPER_ORIGIN,
+            self.site_domain(),
+        )
+        .unwrap(); // validated on construction
+
+        let mut ct_mk = self.mk_ciphertext().to_vec();
```
akoshelev, line 565: can we keep it on the stack without requiring an extra allocation per report?

ipa-core/src/report.rs
```diff
@@ -413,6 +434,186 @@ where
...
+        let plaintext_mk = open_in_place(key_registry, self.encap_key_mk(), &mut ct_mk, &info)?;
+        let mut ct_btt = self.btt_ciphertext().to_vec();
+        let plaintext_btt = open_in_place(key_registry, self.encap_key_btt(), &mut ct_btt, &info)?;
+
+        Ok(OprfReport::<BK, TV, TS> {
+            timestamp: Replicated::<TS>::deserialize(GenericArray::<
+                u8,
+                <Replicated<TS> as Serializable>::Size,
+            >::from_slice(
```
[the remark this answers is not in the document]
akoshelev: I would also say this is hard to read. I prefer the structure of IPA report where there is a method `timestamp` that returns what we want

ipa-core/src/report.rs
```diff
@@ -425,6 +626,8 @@ where
     pub breakdown_key: Replicated<BK>,
     pub trigger_value: Replicated<TV>,
     pub timestamp: Replicated<TS>,
+    pub epoch: Epoch,
+    pub site_domain: String,
```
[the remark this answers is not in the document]
akoshelev: I would be cautious about site domain as well. We are adding it here as owned value but it is very likely that for real inputs there will be a lot of events for the same site domain. Adding lifetimes to the report struct may be very hard later. 

I wonder if we should just have a hashset as string interner and let report borrow from it? Not necessarily in this PR but if we want to do it, we should do it sooner than later

ipa-core/src/report.rs
```diff
@@ -468,73 +671,205 @@ where
...
         let is_trigger = Replicated::<Boolean>::deserialize(GenericArray::from_slice(
-            &buf[sizeof_matchkey + ts_sz + bk_sz + tv_sz
-                ..sizeof_matchkey + ts_sz + bk_sz + tv_sz + sizeof_eventtype],
+            &buf[mk_sz + ts_sz + bk_sz + tv_sz..mk_sz + ts_sz + bk_sz + tv_sz + it_sz],
         ));
+        let epoch = u16::from_le_bytes([
+            buf[mk_sz + ts_sz + bk_sz + tv_sz + it_sz],
+            buf[mk_sz + ts_sz + bk_sz + tv_sz + it_sz + 1usize],
+        ]);
+        // unwrap is safe since it deserializes serialized valid strings
```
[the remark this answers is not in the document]
akoshelev: what is the reason to carry site domain with the deserialized report? if it is for DP budgeting, then maybe drop it as we don't have it yet. materializing this has a non-trivial cost and we could think how to do it more efficiently as we approach the DP problem

ipa-core/src/report.rs
```diff
@@ -564,6 +899,36 @@ mod test {
         assert_eq!(dec_report, report);
     }
 
+    #[test]
+    fn enc_dec_roundtrip_oprf() {
```
akoshelev, line 903: we should also add tests (prop tests maybe) that reject invalid reports

ipa-core/src/query/runner/oprf_ipa.rs
```diff
@@ -69,7 +74,19 @@ where
             v.truncate(sz);
             v
         } else {
-            panic!("Encrypted match key handling is not handled for OPRF flow as yet");
+            LengthDelimitedStream::<EncryptedOprfReport<BA8, BA3, BA20, _>, _>::new(input_stream)
```
akoshelev, line 77: all of these types will likely be runtime parameters. we should probably add a todo/issue to fix it later

ipa-core/src/report.rs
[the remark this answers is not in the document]
danielmasny: I have added your suggestion, Andy and it works. In this case, I extract timestamp from a vector (which is the outcome of a decryption) and not from an OPRF report. Therefore I didn't add a method returning the timestamp.

ipa-core/src/report.rs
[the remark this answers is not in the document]
danielmasny: I think Its best to remove it for now and once we include it in a PR, we can think of how to best include epoch and site_domain. Probably epoch will be included in timestamp.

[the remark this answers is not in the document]
danielmasny: Maybe we should also separate the struct into two structs by adding OPRFInputRow. Then we would not need to serialize and deserialize epoch and site domain.

ipa-core/src/report.rs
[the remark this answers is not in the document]
danielmasny: We need to carry it in OPRF report since it is part of the decryption process, see previous comment. Lets discuss it there.

ipa-core/src/report.rs
danielmasny: I agree we should consider different length match keys, especially a BA128. I would prefer to add this as a generic like TS, BK or TV rather than a pub type. I have added an issue, see: https://github.com/private-attribution/ipa/issues/916

ipa-core/src/report.rs
danielmasny: Would we compile all possible choices of BA types for them? I am not aware that we are tracking this somewhere.

ipa-core/src/report.rs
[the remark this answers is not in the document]
akoshelev: this struct comes into play when the report has been decrypted already, so we shouldn't have a need for things related to crypto. Separating data required to perform MPC computation and everything else is important I think. 

It is likely because we use the same struct on the client and server side and it is really unnecessary imo, especially because there is a cost we have to pay on the server side for it. So it deserves at least a comment/issue so we can come back to it later.

ipa-core/src/report.rs
danielmasny: I agree. This is again caused that we don't separate OprfReport from InputRow. Report has event_type in the clear while InputRow requires it to be secret shared. Therefore, OprfReport has it secret shared. Right now the From function is only used to generate EncryptedOprfReports from OprfReports in a convenient way. It is called by the encrypt function from OprfReport. The encrypt function is only used in test and playbook_oprf_ipa when the plaintext_match_keys flag is set false. Encrypt is done by the device in real IPA, so for our server based implementation it is only used for testing purposes. However, I dont think I should add a test flag in front of playbook_oprf_ipa.

ipa-core/src/report.rs
danielmasny: AES-GCM uses a counter mode, it also uses a 128 bit multiplication over a Galois-Field to compute the tag (see https://en.wikipedia.org/wiki/Galois/Counter_Mode). I agree that the ciphertext wouldn't need to be a multiple of 128 bit block.  However, for the ciphertext encrypting trigger value, breakdown, ... the plaintext is not coincidentally a multiple of 128 bit blocks, nevertheless its ciphertext is (or at least seems to be). Therefore the ciphertext size seems to be ENC_KEY + PAYLOAD rounded up to a multiple of 128 bits + AAD_TAG. Previously, I did not round it up and it caused a panic in the assertion debug_assert_eq!(out.len(), usize::from(self.encrypted_len())); in the encrypt functionality of OprfReport.

ipa-core/src/report.rs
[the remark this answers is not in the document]
danielmasny: Yes, I agree that it would make sense to separate report and input. There are several issue related to it. Initially Richa used the Report struct as input for the MPC computation. I was building on top of it. If we want to separate it, we probably need to remove the OprfReport struct from all the plumbing in the MPC process and replace it with something like OprfInput.

ipa-core/src/report.rs
danielmasny: yes

ipa-core/src/query/runner/oprf_ipa.rs
danielmasny: Agreed, OprfIpaQuery doesnt have generics for it yet. I have opened an issue https://github.com/private-attribution/ipa/issues/917

ipa-core/src/report.rs
danielmasny: Following Andys function here. Don't we validate the size already implicitly due to the fact that we ask for EventType::try_from(bytes[Self::EVENT_TYPE_OFFSET])? so if EVENT_TYPE_OFFSET leads to an out of bounds, doesn't it already return an error? Further, site_domain does not seem to have a bound on the size.

ipa-core/src/report.rs
danielmasny: You mean I randomly generate ciphertexts (or encrypted reports) and try do decrypt them and assert that it fails? I agree that we could concerned here that we don't implement the encryption decryption correctly. The roundtrip only checks for correctness, not for security. However, I think generating random ciphertexts and trying to decrypt them would not ensure that we encrypt/decrypt correctly.

ipa-core/src/report.rs
danielmasny: Yes we don't need that, again was following Andy here. I assume he made the extra copy since during the next step, open_in_place, we would destroy the original mk_ciphertext. I don't think we need it for anything, though.

akoshelev:
This PR hasn't seen any updated in 2 weeks, are we still planning to work on it?

ipa-core/src/report.rs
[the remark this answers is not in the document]
danielmasny: I have split the struct into OprfReport and OPRFIPAInputRow

ipa-core/src/report.rs
danielmasny: we would need to change the function from &self to &mut self. I suggest to leave it for now.

akoshelev: Can you point me to the heap allocation in the old code? I am struggling to find it - it does seem that it works w/o a roundtrip to jemalloc.

The reason I think it matters is because this will be called 50 million (or whatever shard size is) times - we got a decent chance for memory allocator to do its job properly, but this is a game I think we don't have to play as it should be possible to avoid in this code

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -40,6 +41,119 @@ pub(crate) enum Step {
...
+        let raw = &self.to_le_bytes()[..buf.len()];
+        buf.copy_from_slice(raw);
+    }
+
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Result<Self, Self::DeserializationError> {
+        let mut buf_to = [0u8; 8];
+        buf_to[..buf.len()].copy_from_slice(buf);
+        Ok(u64::from_le_bytes(buf_to))
+    }
+}
```
akoshelev, line 68: deserialization for u64 is straightforward - you can use `from_le_bytes`  from std. I'd suggest to remove this

ipa-core/src/report.rs
akoshelev: yes, likely we would use a macro for it

akoshelev (commented):
some minor feedback

ipa-core/src/report.rs
danielmasny: I have added a comment, the TS_OFFSET is for within the BTT Ciphertext

ipa-core/src/report.rs
danielmasny: https://github.com/private-attribution/ipa/issues/953

ipa-core/src/report.rs
akoshelev: @danielmasny pointed that w/o const generic arithmetic we can't really create a generic array of size MK+PK+TAG w/o extra trait bounds `Sum<Add>`.... Before we let this code to go, I wonder if @andyleiserson has any suggestions here? One optimization I can think of is to allocate outside and pass the buffer inside the `decrypt` method, but jemalloc is probably doing it already for us

ipa-core/src/report.rs
andyleiserson: Is the ciphertext coming out a multiple of 16 bytes not explained by the size of the plaintext we're supplying?

```
let mut plaintext_btt = vec![0u8; Self::BTT_END];
```

I think the following two changes would make the code work for any size match key up to 64 bits (after updating the necessary references to `BA64`):

Line 636:
```
            match_key: Replicated::<BA64>::deserialize(GenericArray::from_slice(
                &plaintext_mk[0..(<Replicated<BA64> as Serializable>::Size::USIZE)],
            ))
```

Line 825:
```
        let mut plaintext_mk = vec![0u8; 16];
        self.match_key.serialize(GenericArray::from_mut_slice(
            &mut plaintext_mk[0..(<Replicated<BA64> as Serializable>::Size::USIZE)],
        ));
```

andyleiserson (approved):

ipa-core/src/report.rs
```diff
@@ -424,136 +422,457 @@ where
...
+            key_id,
+            self.epoch,
+            self.event_type,
+            HELPER_ORIGIN,
+            self.site_domain.as_ref(),
+        )?;
+
+        if Self::TV_OFFSET + <Replicated<TV> as Serializable>::Size::USIZE > U16::USIZE {
+            return Err(PlaintextLengthError());
+        }
```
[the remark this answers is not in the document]
danielmasny: Good catch, we don't need this check anymore since it supports arbitrary lengths

ipa-core/src/report.rs
danielmasny: I dont think so but there might be something odd going on. Right now, in 
```
#[test]
    fn enc_dec_roundtrip_oprf()
```
I use 
```
let report = OprfReport::<BA8, BA3, BA20>
```
and compute
```
let enc_report_bytes = report.encrypt(key_id, &key_registry, &mut rng).unwrap();
let enc_report = EncryptedOprfReport::from_bytes(enc_report_bytes.as_slice()).unwrap();
let dec_report: OprfReport<BA8, BA3, BA20> = enc_report.decrypt(&key_registry).unwrap();
```
I get an error message that the ciphertext length doesn't match when I don't round up $2\cdot (8+3+20)=62$ bit to  $16$ byte

danielmasny: I dont think so, in 
```
    #[test]
    fn test_decryption_fails() 
```
I get and error if I dont round up to a multiple of 16 even though the plaintext contains 62 bit:
```
let report = OprfReport::<BA8, BA3, BA20> 
let enc_report_bytes = report.encrypt(key_id, &key_registry, &mut rng).unwrap();
let enc_report = EncryptedOprfReport::from_bytes(enc_report_bytes.as_slice()).unwrap();
let dec_report: OprfReport<BA8, BA3, BA20> = enc_report.decrypt(&key_registry).unwrap();

assert_eq!(dec_report, report);
```

ipa-core/src/report.rs
danielmasny: I fixed the ciphertext length. The trait bounds are not too bad anymore since you were right, I don't need to round up to a multiple of 16 bytes.

ipa-core/src/report.rs
danielmasny: I have added an new error and check the length

ipa-core/src/report.rs
```diff
@@ -424,136 +421,384 @@ where
...
+            self.epoch(),
+            self.event_type(),
+            HELPER_ORIGIN,
+            self.site_domain(),
+        )
+        .unwrap(); // validated on construction
+
+        let mut ct_mk: GenericArray<u8, CTMKLength> =
+            *GenericArray::from_slice(self.mk_ciphertext());
+        // let mut ct_mk = self.mk_ciphertext().to_vec();
```
akoshelev, line 595: don't want to hold this change for too long, but lets make sure we clean up commented code. Fine with a follow up PR to address those

ipa-core/src/report.rs
```diff
@@ -424,136 +421,384 @@ where
...
+        let plaintext_mk = open_in_place(key_registry, self.encap_key_mk(), &mut ct_mk, &info)?;
+        let mut ct_btt: GenericArray<u8, CTBTTLength<BK, TV, TS>> =
+            GenericArray::from_slice(self.btt_ciphertext()).clone();
```
akoshelev, line 598: you can use the same dereferencing trick `*GenericArray` instead of `GenericArray(...).clone()`

akoshelev (approved):

PR #907, opened 2023-12-31
Oh well... cyclic type of infinite size

benjaminsavage:
In an attempt to mitigate #867, I tried to change the code to no longer generate on future per *per* which then blows up into many more futures, but instead directly generate one future per input row.

I got pretty far, but then I hit this error `cyclic type of infinite size` that I do not know how to resolve.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
@@ -8,6 +8,7 @@ use futures::{
     stream::{iter as stream_iter, unfold},
     Stream, StreamExt, TryStreamExt,
 };
+use futures_util::{Future, future::ready};
 use ipa_macros::Step;
 
 use super::boolean_ops::expand_shared_array_in_place;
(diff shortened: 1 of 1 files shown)


akoshelev:
The cyclic type error makes sense to me. `compute_output` is an async function, so its return type is a `Future`. 

```rust
   let previous_row = if let Some(fut) = previous_row_future {
        fut.await?
    } else {
        initialize_new_device_attribution_variables::<BK, TV, TS, SS>(first_row)
    };
```

this block makes the return type of `compute_output` include the type of `previous_row_future`. because we call this function in a loop, we build a list like `Future<RowN<Future<RowN-1<...Future<Row1>>` and the depth of this is not known at compile time (because it depends on how many iterations the loop contains).

akoshelev:
even if we build a type like this, it will likely have the same observable behavior, unless I misunderstand the intent

andyleiserson:
As far as a direct solution to this error, can you box the future?

The change in https://github.com/private-attribution/ipa/compare/main...andyleiserson:ipa:oprf-stall-7 (note this is a different branch from the one with the buffer workaround that I linked in the chat) is kind of like your change in spirit, but different in how it is implemented in a way that avoids the compound future problem.

I don't think that your change in this PR, or my change that I just mentioned, resolve the underlying work unit problem. Maintaining 1:1 correspondence of futures throughout the protocol only helps if it means that we'll reliably reach the 1,024 record flush threshold *for each step* -- these changes still have 1/records_per_user less data in internal steps of the attribution circuit, than at the output. It may be possible to remedy that by scaling the work factor up by ~records_per_user, but then there's a risk that execution will fail for an unusual distribution of records_per_user. Buffering or parallel execution seem like more robust fixes.

PR #908, opened 2024-01-04
Remove public access to `AdditiveShare` members

andyleiserson:
Making the members of `AdditiveShare` private reduces the number of things that will need to be updated for vectorization.

In cases where there is both horizontal and vertical vectorization, my plan is that the vertical (across records) is the inner vectorization, and the horizontal (vector-type values within one record) is the outer. My plan is that values of this sort will be stored in `BitDecomposed<AdditiveShare<BitVec>>`. (The most likely candidate for `BitVec` is a Boolean array, although for vertical vectorization, it should rarely be necessary to access individual elements of the vector, and disallowing such access except through special pack/unpack primitives may catch some bugs.)

### DIFF
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -16,6 +16,16 @@ impl Block for bool {
 #[derive(Clone, Copy, PartialEq, Debug, Eq)]
 pub struct Boolean(bool);
 
+impl Boolean {
+    pub const TRUE: Boolean = Self(true);
+    pub const FALSE: Boolean = Self(false);
+
+    #[must_use]
+    pub fn as_u128(&self) -> u128 {
+        bool::from(*self).into()
+    }
+}
+
 impl ExtendableField for Boolean {
     type ExtendedField = Gf32Bit;
 
@@ -139,7 +149,7 @@ impl Field for Boolean {
     const ONE: Boolean = Boolean(true);
 
     fn as_u128(&self) -> u128 {
-        bool::from(*self).into()
+        Boolean::as_u128(self)
     }
 
     fn truncate_from<T: Into<u128>>(v: T) -> Self {
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -22,7 +22,7 @@ impl Boolean {
 
     #[must_use]
     pub fn as_u128(&self) -> u128 {
-        bool::from(*self).into()
+        u128::from(bool::from(*self))
     }
 }
 
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
@@ -6,7 +6,12 @@ use crate::ff::Expand;
 use crate::{
     error::Error,
     ff::{ArrayAccess, CustomArray, Field},
-    protocol::{basics::SecureMul, context::Context, step::BitOpStep, RecordId},
+    protocol::{
+        basics::{SecureMul, ShareKnownValue},
+        context::Context,
+        step::BitOpStep,
+        RecordId,
+    },
     secret_sharing::{replicated::semi_honest::AdditiveShare, SharedValue},
 };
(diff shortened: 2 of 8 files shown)


ipa-core/src/ff/boolean.rs
```diff
@@ -16,6 +16,16 @@ impl Block for bool {
 #[derive(Clone, Copy, PartialEq, Debug, Eq)]
 pub struct Boolean(bool);
 
+impl Boolean {
+    pub const TRUE: Boolean = Self(true);
+    pub const FALSE: Boolean = Self(false);
+
+    #[must_use]
+    pub fn as_u128(&self) -> u128 {
+        bool::from(*self).into()
```
martinthomson, line 25: I prefer explicit types unless the source type is specified as Into in a trait.

```rust
    u128::from(bool::from(*self))

**Review by martinthomson — APPROVED** [review]
LGTM overall.

File: ipa-core/src/ff/mod.rs [hunk 2]
```diff
@@ -93,6 +93,29 @@ pub trait ArrayAccess {
...
+// references: Galois fields, which have `ArrayAccess::Output = bool`. It probably would be possible
+// to provide by-reference access for Galois fields using the bit-reference functionality in the
+// bitvec crate.
+//
+// No reason this shouldn't have iter too, but deferring that because it's not needed yet, and will
+// be messy to support for Galois fields because of the `bool` value type.
+pub trait ArrayAccessReplicated {
+    type Ref<'a>: BorrowReplicated
+    where
+        Self: 'a;
```
akoshelev, line 112: for my own curiosity, why is it better than `BorrowReplicated<'a>` ?

ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
```diff
@@ -32,11 +37,11 @@ where
     C: Context,
     YS: SharedValue + CustomArray<Element = XS::Element>,
     XS: SharedValue + CustomArray + Field,
-    XS::Element: Field + std::ops::Not<Output = XS::Element>,
+    XS::Element: Field,
+    AdditiveShare<XS::Element>: std::ops::Not<Output = AdditiveShare<XS::Element>>,
```
akoshelev, line 41: should we consider implementing `Not` for `AdditiveShare<CustomArray>`?

akoshelev (approved):

andyleiserson:
I am going to remove the `BorrowReplicated` and `ArrayAccessReplicated` changes from this PR, because it seems like they're not important for the near-term vectorization changes -- if the only type of array element indexed within AdditiveShare is a single bit, then there may be small or even negative performance benefit of accessing by reference. It may be needed eventually, but I'll defer adding the code until the need is clear.

ipa-core/src/ff/mod.rs
andyleiserson: I think the `BorrowReplicated` trait could have a lifetime parameter, but it's not necessary -- the implicit lifetime parameter in the reference argument to the trait methods is sufficient. Even if the trait were `BorrowReplicated<'a>`, the lifetime parameter would still need to propagate to something here (either the `Ref<'a>` GAT or a lifetime parameter on `ArrayAccessReplicated`. Just saying `BorrowReplicated<'a>` isn't meaningful and wouldn't work, because the `'a` lifetime would be undefined.

The `where Self: 'a` is something the compiler insists on -- it ensures that an impl like the one for `BitDecomposed<S>` with `Ref<'a>` of `BitDecomposedIndex<'a, S>(&'a S)` is possible. Using the type `&'a S` creates an implied bound `S: 'a` for that implementation. Absent `where S: 'a` in the trait definition, the implementation would have stricter bounds than the definition. This is discussed in more detail in rust issue 87479.

ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
akoshelev: yea i was suggesting to implement Not for shares of boolean arrays only. If we don't need that, it is even better

PR #910, opened 2024-01-04
Improve error messages related to PRSS indexes

andyleiserson:


ipa-core/src/protocol/prss/mod.rs
```diff
@@ -75,17 +76,32 @@ impl Debug for UsedSet {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}:{}", self.index.0, self.offset)
+    }
+}
+
+#[cfg(debug_assertions)]
+impl From<u64> for PrssIndex128 {
+    fn from(value: u64) -> Self {
+        Self {
+            index: u32::try_from((value >> 32) & u64::from(u32::MAX))
```
akoshelev, line 94: value is unsigned so >> must be a logical shift. I am missing something apparently what is the reason for doing `& u32::MAX`?

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -106,7 +122,7 @@ impl From<u32> for PrssIndex {
 // It would be nice for this to be TryFrom, but there's a lot of places where we use u128s as PRSS indexes.
 impl From<u128> for PrssIndex {
     fn from(value: u128) -> Self {
-        Self(value.try_into().unwrap())
+        Self(value.try_into().expect("PRSS index out of range"))
```
akoshelev, line 125: I know it is annoying but it would simplify the debugging if we print the actual value that is out of range

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -41,17 +44,15 @@ impl UsedSet {
...
-        let index = u128::from(index);
-        if index > usize::MAX as u128 {
-            tracing::warn!(
-                "PRSS verification can validate values not exceeding {}, index {index} is greater.",
-                usize::MAX
-            );
+        let raw_index = u128::from(index);
+        if raw_index > usize::MAX as u128 {
+            // This is unreachable with the current PRSS index encoding.
+            tracing::warn!("Index '{index}' is not supported by PRSS verification.");
```
akoshelev, line 50: I may prefer the previous message that specifies the bounds because if I ever see it in the logs, I wouldn't need to go to the code to see why it was rejected

martinthomson:
```suggestion
            tracing::warn!("PRSS index is too large: {index} > usize::MAX");
```

ipa-core/src/protocol/prss/mod.rs
martinthomson:
```suggestion
        let Some(v) = u32::try_from(value) else {
            panic!("PRSS indices need to be smaller: {index} > u32::MAX");
        };
        Self(v)
```

martinthomson (approved):
I'm OK with the changes here.  It's nice that this ensures that we stay within u64::MAX.

ipa-core/src/protocol/prss/mod.rs
andyleiserson: I just wrote the two cases (this and the one on the next line) the same way, even though masking isn't necessary in this one. I may have been rewriting it in a hurry to appease clippy.

PR #915, opened 2024-01-12
Add code coverage to the CI

akoshelev:
tested on my repo: https://app.codecov.io/github/akoshelev/raw-ipa/commit/1a593623b8bd0c8f205c17ebbcb46e2ce267f4c4/tree

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -161,8 +161,6 @@ jobs:
   coverage:
     name: Measure coverage
     runs-on: ubuntu-latest
-    env:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
@@ -183,6 +181,7 @@ jobs:
       - name: Report to codecov.io
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           file: ipa.cov
           fail_ci_if_error: false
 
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -155,3 +157,32 @@ jobs:
         run: rustup component add rust-src
       - name: Run tests with sanitizer
         run: RUSTFLAGS="-Z sanitizer=${{ matrix.sanitizer }} -Z sanitizer-memory-track-origins" cargo test -Z build-std --target $TARGET --no-default-features --features "cli web-app real-world-infra test-fixture descriptive-gate"
+
+  coverage:
+    name: Measure coverage
+    runs-on: ubuntu-latest
+    env:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
martinthomson, line 165: You should be able to pass this directly to the action rather than putting it in the environment.  See below.

.github/workflows/check.yml
```diff
@@ -155,3 +157,32 @@ jobs:
...
+        run: cargo install cargo-llvm-cov
+
+      - name: Measure coverage
+        run: ./scripts/coverage-ci --lcov --output-path ipa.cov
+
+      - name: Report to codecov.io
+        uses: codecov/codecov-action@v3
+        with:
+          file: ipa.cov
+          fail_ci_if_error: false
```
martinthomson, line 187:
```suggestion
          fail_ci_if_error: false
          token: ${{ secrets.CODECOV_TOKEN }}
```

martinthomson (approved):

.github/workflows/check.yml
akoshelev: Indeed, it is even marked as [required](https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments)

martinthomson: In practice, it's not, but I think that they must fall back to env.

PR #919, opened 2024-01-16
Make Galois multiplication constant time 

martinthomson:
...sort of.  I haven't tried very hard to confound the optimization
routines in the compiler.

This could slow things down, but I believe that it is essential as a
baseline to ensure that we don't leak.

This adds a new test, because I discovered that our Gf20 polynomial was
incorrectly transcribed.

### DIFF
--- a/ipa-core/src/ff/galois_field.rs
+++ b/ipa-core/src/ff/galois_field.rs
@@ -331,7 +331,7 @@ macro_rules! bit_array_impl {
             //
             // Since we know that x^8 + x^4 + x^3 + x + 1 = 0, we can distribute out terms of this structure and replace them with zero.
             //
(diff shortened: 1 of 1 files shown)


martinthomson:
Thanks codecov for highlighting why tools like this are sometimes stupid.

akoshelev:
Did you have a chance to look at the generated assembly for it using `cargo asm` or Godbolt?

martinthomson:
https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=586daa21839c9040470036b68a079aa4 looks good.  I also checked 4 bits and that seems equally good.

akoshelev (approved):

Issue #921, opened 2024-01-16
Compact gate for the malicious IPA is broken


akoshelev:
Currently failing with

```
thread 'tokio-runtime-worker' panicked at ipa-core/src/protocol/attribution/mod.rs:429:10:
Invalid state transition: 2182 -> current_credit_or_credit_update
```

reproducible in integration tests only, when setting the input size to 100. 10 is too small to hit the bad path, but we can't run anything at scale

akoshelev:
steps to reproduce:

1. Change [this](https://github.com/private-attribution/ipa/blob/fd307243498fdbe889fa51d05a22b2732bd56382/ipa-core/tests/common/mod.rs#L206) to 100
2. Run compact gate tests: `cargo test --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"`

benjaminsavage:
I think #927 addresses this.

akoshelev:
Now it works for 100 rows, but fails for 100k

```
thread 'tokio-runtime-worker' panicked at ipa-core/src/protocol/basics/mul/malicious.rs:18:10:
Invalid state transition: 2274 -> duplicate_multiply
```

benjaminsavage:
This makes sense.

The steps of old IPA are dependent upon the total number of rows. To be specific, the "interaction pattern" stuff that controls the binary tree approach is the part that is dependent.
So in general, the number of passes is O(log(n)) records.

richajaindce:
Based on "ancient" commit we had upto depth31 and that must be the reason for it to work. Will add it back. https://raw.githubusercontent.com/private-attribution/ipa/2763bbf76fd2e7f11e2641c2e66d08c0af3d9cb8/ipa-core/src/protocol/step/steps.txt

akoshelev:
I ran this on 100k input and it worked so I think that this issue can be closed now.
One thing that I noticed is that the order of step is different depending on whether we run `collect_steps` with single thread or multiple threads. `steps.txt` generated using only one thread causes mutli-thread IPA to crash

PR #925, opened 2024-01-18
Make compact-gate tests work for n=100

richajaindce:
Fix for #921


ipa-core/tests/common/mod.rs
```diff
@@ -203,7 +203,7 @@ pub fn test_ipa(mode: IpaSecurityModel, https: bool) {
 }
 
 pub fn test_ipa_with_config(mode: IpaSecurityModel, https: bool, config: IpaQueryConfig) {
-    const INPUT_SIZE: usize = 10;
+    const INPUT_SIZE: usize = 100;
```
richajaindce, line 206: @akoshelev not sure if you wanted to change this SIZE to 100 but 100 shouldn't be too big a number to cause slowdown. Happy to revert if needed

scripts/collect_steps.py
```diff
@@ -19,7 +19,7 @@
     "--num-multi-bits",
     "3",
 ]
-QUERY_SIZE = 10
+QUERY_SIZE = 100
```
richajaindce, line 22: Changed this here to avoid issues in future

akoshelev: would it be sufficient to run malicious IPA for 1M rows? It seems that number of steps synthesized depends on the depth of the circuit - in this case would we be better generating all of them, similarly to what we did before?

richajaindce:
#927 is a better fix

PR #927, opened 2024-01-19
Backport of improved aggregation logic to old IPA

benjaminsavage:
The new approach to aggregation (binary tree, moving value obliviously) is just more efficient than our old code. No reason not to use this improved algorithm in old IPA as well.

### DIFF
--- a/ipa-core/src/protocol/attribution/aggregate_credit.rs
+++ b/ipa-core/src/protocol/attribution/aggregate_credit.rs
@@ -10,8 +10,7 @@ use crate::{
         context::{UpgradableContext, UpgradedContext, Validator},
         ipa_prf::prf_sharding::bucket::move_single_value_to_bucket,
         modulus_conversion::convert_bits,
-        sort::{bitwise_to_onehot, generate_permutation::ShuffledPermutationWrapper},
-        step::BitOpStep,
+        sort::generate_permutation::ShuffledPermutationWrapper,
         BasicProtocols, RecordId,
     },
     secret_sharing::{
@@ -75,6 +74,10 @@ where
     let record_count = breakdown_keys.len();
     let valid_bits_count = u32::BITS - (max_breakdown_key - 1).leading_zeros();
 
+    let move_value_to_bucket_context = ctx
+        .narrow(&Step::MoveValueToBucket)
+        .set_total_records(record_count);
+
     let converted_bk = convert_bits(
         ctx.narrow(&Step::ModConvBreakdownKeyBits)
             .set_total_records(record_count),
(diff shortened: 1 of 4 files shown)


ipa-core/src/protocol/attribution/aggregate_credit.rs
```diff
@@ -110,23 +88,20 @@ where
...
-                    ))
-                    .await
+                    move_single_value_to_bucket(
+                        ctx,
+                        RecordId::from(i),
+                        bk.unwrap(),
+                        cred,
+                        usize::try_from(max_breakdown_key).unwrap(),
+                        true,
+                    )
```
martinthomson, line 100:
```suggestion
                    ).await
```

This?

scripts/collect_steps.py
```diff
@@ -19,7 +19,7 @@
     "--num-multi-bits",
     "3",
 ]
-QUERY_SIZE = 10
+QUERY_SIZE = 100
```
benjaminsavage, line 22: @akoshelev - this line is the key. If you want to run for a larger number of input rows, you need to update this, then regenerate the steps. This is due to the "interaction pattern" stuff which is a binary tree. So there are log(n) steps for various things (like attribution). If you want to run on 20 million rows, you'll need this number to be a lot higher to generate more than "depth=6" which is what the steps currently have. That should support ~2^6 rows, but you'll need a lot more.

richajaindce: is it 2^6 per user or overall that depth=6 will support?

richajaindce (approved):
Looks good. I dont think we need to re-run for bigger size queries (atleast for synthetic data we generate) as # of rows per user is max 50 by default. Unless someone runs with a bigger value. Real test can definitely fail if max users > 64

ipa-core/src/protocol/ipa/mod.rs
```diff
@@ -1115,9 +1115,9 @@ pub mod tests {
                 cap_one(),
                 SemiHonest,
                 PerfMetrics {
-                    records_sent: 14_421,
-                    bytes_sent: 47_100,
-                    indexed_prss: 19_137,
+                    records_sent: 14_397,
```
richajaindce, line 1118: yay some minor gains :D

scripts/collect_steps.py
benjaminsavage: For old IPA, the steps depend upon the total number of rows (and not upon the number of rows per user).

That's no longer the case for OPRF-IPA, but the config to generate those steps is lower down in this file.

ipa-core/src/protocol/ipa/mod.rs
benjaminsavage: It'll be more significant when using a larger number of breakdowns. This test is only using 4 breakdowns =). The savings is like ~O(num_breakdowns)

scripts/collect_steps.py
akoshelev: Interesting - presumably we have been running the step generation process with 10 rows for a while. It was working fine for up to 20M rows.

Issue #928, opened 2024-01-19
bitvec performance


andyleiserson:
`bitvec` appears to not be particularly efficient for operations on large numbers of bits. A benchmark that takes 200 ms on my Gf2Array vectorized implementation takes 320 ms when using bitvec.

akoshelev:
can you post the benchmark code here? curious where the bottleneck may be

andyleiserson:
Updated the description with detail of the benchmarks. The time is consumed in [`sp_bitop_assign`](https://github.com/ferrilab/bitvec/blob/5fb855073acc2ed045094ed89d8daf8c765f0135/src/slice/specialization/lsb0.rs#L38). I tried adding some additional `#[inline]` directives for `word_op`, `load_le` / `store_le`, which did seem to have some effect on the generated code, but for the load/store portion still produced something that looked far from speed-of-light. I didn't see any benchmarks in the bitvec crate that cover this case, and didn't get as far as benchmarking ipa with modified versions of the bitvec crate.

PR #929, opened 2024-01-19
Vectorization of boolean arrays

andyleiserson:
Builds on #852.

Adds support for vectorization of bitwise protocols, and adds vectorized version of `compare_gt` (used by quicksort).

Unvectorized bitwise protocols operate on `AdditiveShare<BA>`, as before (bits of the BA are different bits of multi-bit values associated with a particular record). Vectorized bitwise protocols operate on `BitDecomposed<AdditiveShare<BA>>`, where the elements of the outer `BitDecomposed` are different bits of multi-bit values associated with a particular record, and bits of the inner BA are the values of a particular bit position for a stripe of records. Adds an `ArrayAccessRef` variant of the `ArrayAccess` trait that is implemented for both kinds of inputs, and an `ArrayBuilder` trait that is implemented for both kinds of outputs.

## Performance

`compare_gt` benchmark. Generates 131,072 64-bit values and compares each of them to a single 64-bit threshold. This is a proxy for quicksort.

* Unvectorized: 39.6 s 🐢
* Vectorized x256, bitvec version (what is in this PR): 320 ms
* Vectorized x256, raw array version (see #928): 200 ms
* Vectorized x1024, raw array version: 100 ms
* Vectorized x4096, raw array version: 84 ms 🔥

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -6,7 +6,7 @@ use generic_array::GenericArray;
 use typenum::{U14, U2, U32, U8};
 
 use crate::{
-    ff::{boolean::Boolean, ArrayAccess, Field, Serializable},
+    ff::{boolean::Boolean, ArrayAccess, ArrayBuilder, Field, Serializable},
     protocol::prss::{FromRandom, FromRandomU128},
     secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
 };
@@ -28,9 +28,14 @@ macro_rules! store_impl {
     };
 }
 
-/// iterator for Boolean arrays
+/// Iterator returned by `.iter()` on Boolean arrays
 pub struct BAIterator<'a> {
     iterator: std::iter::Take<Iter<'a, u8, Lsb0>>,
+}
+
+/// Iterator returned by `.into_iter()` on Boolean arrays
+pub struct BAOwnedIterator<S: IntoIterator> {
+    iterator: std::iter::Take<S::IntoIter>,
 }
 
 ///impl Iterator for all Boolean arrays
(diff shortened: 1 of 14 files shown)


akoshelev:
> Vectorized x4096, raw array version: 84 ms 🔥

this is just awesome

ipa-core/src/ff/boolean_array.rs
```diff
@@ -405,20 +478,84 @@ macro_rules! boolean_array_impl {
...
+
+                fn builder() -> Self::Builder {
+                    BooleanArrayBuilder::new()
+                }
+            }
+
+            impl SharedValueArray<Boolean> for $name {
+                const ZERO_ARRAY: Self = <$name as SharedValue>::ZERO;
+
+                fn from_fn<F: FnMut(usize) -> Boolean>(mut f: F) -> Self {
```
danielmasny, line 512: why do we need this function?

ipa-core/src/ff/boolean_array.rs
```diff
@@ -556,3 +703,116 @@ impl rand::distributions::Distribution<BA256> for rand::distributions::Standard
...
+
+#[cfg(all(test, unit_test))]
+mod tests {
+    use rand::{thread_rng, Rng};
+
+    use super::*;
+
+    // It does not seem worth running these tests for every BA type, although
+    // it would be worth writing a version for the largest BA type (which
+    // requires replacing truncate_from).
```
danielmasny, line 752: We could add the test to the macro if we want to test it for any BA type. I would also test it for BA types that are not powers of 2, i.e. BA3. Otherwise I am fine with not testing it for any BA type.

danielmasny (approved):
It looks good to me. I have only minor comments. Maybe write a test for the last few lines that are not covered by a test.

ipa-core/src/ff/boolean_array.rs
andyleiserson: For `StdArray` the implementation uses https://doc.rust-lang.org/std/array/fn.from_fn.html to avoid zero-initializing the array. Otherwise, it's just a different way of writing a for loop to fill the array.

It does appear this is unused in this PR (probably due to either revision of the changes or splitting things into multiple PRs). The function is used in my branch for vectorizing share conversion.

ipa-core/src/ff/boolean_array.rs
andyleiserson: I will move it into the macro, it should be fast and that seems easier than introducing a different macro to generate the tests for particular configs.

PR #930, opened 2024-01-23
draft of quadratic proofs

danielmasny:
draft of proofs for quadratic relations


ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
+use crate::ff::{Field, Invert};
+
+/// function for generating canonical evaluation points
+/// overwrites `evaluation_points`
+pub fn generate_evaluation_points<F>(evaluation_points: &mut [F]) -> ()
```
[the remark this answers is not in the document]
akoshelev: I don't see a reason why you need to keep 0..N points in memory. You can always compute them for a given `i` with simple check that $i \lt p$.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+        if i + 1 < evaluation_points.len() {
+            for j in i + 1..evaluation_points.len() {
+                denominators[i] *= evaluation_points[i] - evaluation_points[j];
+            }
+        }
+        denominators[i] = denominators[i].invert();
+    }
+}
+
+/// computes the lagrange base polynomials
```
akoshelev, line 58: I don't understand the nomenclature here, can you be a bit more specific what we are trying to achieve?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+        .for_each(|(result, base)| {
+            *result *= base
+                .iter()
+                .zip(y.iter())
+                .fold(F::ZERO, |acc, (base, y)| acc + *base * *y)
+        });
+}
+
+/// lagrange evaluation
+/// given points `(x_0,y_0),...,(x_n,y_0)` on a polynomial `p`
```
akoshelev, line 128:
```suggestion
/// given points `(x_0,y_0),...,(x_n,y_n)` on a polynomial `p`
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+/// and points `p_0,...,p_m`
+/// compute `result_0=result_0*p(p_0),...,result_m=result_m*p(p_m)`
```
akoshelev, line 130: what is `result_0` initial value?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+                    }
+                }
+            }
+        });
+}
+
+/// precomputed lagrange evaluation
+/// evaluates polynomial using base polynomials
+/// multiplies `result` with input `result`, i.e. use input `F::ONE`
+pub fn lagrange_evaluation_precomputed<F>(y: &[F], base: &Vec<Vec<F>>, result: &mut [F]) -> ()
```
akoshelev, line 105: I am more in favor of using strongly typed objects here. We implemented Lagrange interpolation a while ago, see if that can be a fit here

https://github.com/private-attribution/ipa/pull/32

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/// lagrange evaluation
+/// given points `(x_0,y_0),...,(x_n,y_0)` on a polynomial `p`
+/// and points `p_0,...,p_m`
+/// compute `result_0=result_0*p(p_0),...,result_m=result_m*p(p_m)`
+/// It uses the lagrange method to compute the points `https://en.wikipedia.org/wiki/Lagrange_polynomial`
+/// further, rather than outputting `p(p_k)`, we set `result_k=result_k*p(p_k)`
+/// which fits better to how we use `p(p_k)`
+pub fn lagrange_evaluation<F>(x: &[F], y: &[F], p: &[F], result: &mut [F]) -> ()
```
akoshelev, line 134: I think this is something more than just evaluating a polynomial at given points. It is probably tied to what malicious security needs to do, so it could be broken up into two parts - polynomial evaluation and everything else

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+
+        // check equality
+        assert_eq!(y_values[evaluation_points_size], evaluated_y[0]);
+    }
+
+    #[test]
+    fn lagrange_precomputed_test() {
+        let mut rng = thread_rng();
+        let evaluation_points_size = rng.gen::<usize>() % 100;
+        let mut evaluation_points_random = vec![Fp25519::ONE; evaluation_points_size];
```
akoshelev, line 255: these tests are large and hard to read. Consider moving repetitive code into helper function and using type system as suggested above.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,291 @@
...
+where
+    F: Field + Invert,
+{
+    debug_assert_eq!(x.len(), y.len());
+    debug_assert_eq!(p.len(), result.len());
+    #[cfg(debug_assertions)]
+    {
+        for i in 0..x.len() {
+            for j in 0..x.len() {
+                debug_assert_eq!((i, j, x[i] - x[j] != F::ZERO || j == i), (i, j, true))
```
akoshelev, line 144: it is already under `cfg[(debug_assertions)]` so `assert_eq` would work better here

akoshelev (commented):
I only looked at lagrange module

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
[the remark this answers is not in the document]
martinthomson:
```suggestion
pub fn generate_evaluation_points<F>(evaluation_points: &mut [F]) {
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,293 @@
...
+                        evaluation_points[i] - evaluation_points[j] != F::ZERO || j == i
+                    ),
+                    (i, j, true)
+                )
+            }
+        }
+    }
+    // compute denominators:
+    for i in 0..evaluation_points.len() {
+        if i > 0 {
```
martinthomson, line 44: The if statement bothers me slightly.  You want $d_i = \left\(\prod_{i\neq j} i\cdot(p_j-p_i)\right\)^{-1}$ yes?

ipa-core/src/ff/ec_prime_field.rs
```diff
@@ -289,4 +303,18 @@ mod test {
...
+        let mut rng = thread_rng();
+        let a = rng.gen::<u128>();
+        let b = rng.gen::<u128>();
+        if a < b {
+            assert!(Fp25519::try_from(a).unwrap() < Fp25519::try_from(b).unwrap())
+        } else if a > b {
+            assert!(Fp25519::try_from(a).unwrap() > Fp25519::try_from(b).unwrap())
+        } else {
+            assert_eq!(Fp25519::try_from(a).unwrap(), Fp25519::try_from(b).unwrap())
+        }
```
akoshelev, line 318:
```suggestion
   assert_eq!(a.cmp(b), FpXXX:try_from(a).cmp(FpXXX::try_from(b))
```

ipa-core/src/helpers/mod.rs
```diff
@@ -245,6 +245,17 @@ pub enum Direction {
     Right,
 }
 
+/// implements negation for `Direction`
+impl std::ops::Neg for Direction {
```
akoshelev, line 249: Oh that may be too magical, maybe just add `fn other(&self) -> Self` to the `Direction` impl?

ipa-core/src/protocol/ipa_prf/malicious_security/fiat_shamir.rs
```diff
@@ -0,0 +1,208 @@
...
+/// and computes a vector of random points by hashing the individual proofs parts
+/// this function only computes `r` for a single proof part
+fn fiat_shamir<F>(proof: &[F]) -> F
+where
+    F: Field,
+{
+    // serialize proof const SIZE: usize = <F as Serializable>::Size::USIZE;
+    let mut ikm = Vec::<u8>::with_capacity(proof.len() * <F as Serializable>::Size::USIZE);
+    proof.iter().for_each(|f| {
+        let mut buf = vec![0u8; <F as Serializable>::Size::USIZE];
```
akoshelev, line 90:
```suggestion
        let mut buf = [0u8; <F as Serializable>::Size::USIZE];
```

ipa-core/src/protocol/ipa_prf/malicious_security/fiat_shamir.rs
```diff
@@ -0,0 +1,208 @@
...
+where
+    F: Field,
+{
+    fiat_shamir(proof_left) + fiat_shamir(proof_right)
+}
+
+/// the Fiat-Shamir core function
+/// it takes a commitment, which is for `NIDZKP` the actual proof
+/// and computes a vector of random points by hashing the individual proofs parts
+/// this function only computes `r` for a single proof part
```
akoshelev, line 82: what is `r`?

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,889 @@
...
+    lagrange_evaluation_precomputed(&final_share_1, &base_r, std::slice::from_mut(&mut output_1));
+
+    return (output_0, output_1);
+}
+
+/// computes `summand_of_g = p*q` and adds it to `g`
+/// polynomials `p` and `q` are represented by value `u`, `v`
+/// which are the evaluations of `p,q` at `evaluation_points[0..recursion_factor]`
+/// `g` is represented by its evaluation at `evaluation_points`
+fn compute_summand_of_g<F>(u: &[F], v: &[F], r_f: usize, base: &Vec<Vec<F>>, g: &mut [F]) -> ()
```
martinthomson, line 525: This should take in iterators and return a value.  Having it mutate an argument makes it a lot less flexible.  Also, this requires additional and potentially unnecessary allocations.

Given that `g` is only added to, could you make this function produce a "map" operand?  That is, you would do something like this:

```rust
g = g.into_iter().map(Summand::new(u, v, precomputed).as_f()).collect::<Vec<_>>();
```

or just
```rust
g = Summand::new(u, v, precomputed) + g;
```

Which could implement `Add<Rhs = X> where X: AsRef<F: Field + Invert>` or similar (excuse the abuse of notation).

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,889 @@
...
+
+    // compute first recursion factor many points of summand_of_g
+    summand_of_g
+        .iter_mut()
+        .take(r_f)
+        .enumerate()
+        .for_each(|(i, x)| *x = u[i] * v[i]);
+
+    // compute summand_of_g = 1*u for rest of points
+    lagrange_evaluation_precomputed(u, &base, &mut summand_of_g[r_f..2 * r_f]);
```
martinthomson, line 545: You should wrap the precomputed lagrange coefficients into a struct.

Given the way that the precomputed items are applied, you could have this in the form:

```rust
g[r_f..2*r_f] = base.evaluate(u) * g[r_f..2*r_f];
```

(This doesn't have to be Mul, as above, use a method on the returned value, if you prefer.)

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,889 @@
...
+        // check whether there are too many random points,
+        // which leads to unnecessary compression
+        debug_assert!(share.0.len() >= r_f);
+
+        // fill u and v with zeros such that it is a multiple of the recursion factor
+        let coset = share.0.len() % r_f;
+        if coset != 0 {
+            share.0.extend(std::iter::repeat(F::ZERO).take(coset));
+            share.1.extend(std::iter::repeat(F::ZERO).take(coset));
+        }
```
martinthomson, line 445: This padding function should be extracted if possible.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,889 @@
...
+
+        // compute secret share g_left (i.e. part of the proof recovered by right party) using prss_right
+        let mut g_left = vec![F::ZERO; g.len()];
+        g_left.iter_mut().for_each(|g| {
+            // get prss
+            // todo replace `F::ONE` with actual prss value
+
+            // compute `g_left`
+            *g = F::ONE;
+        });
```
martinthomson, line 118: You will want to use `slice::fill_with` for this.

martinthomson (commented):
I can see a bunch of stuff here that could be split out into components that could be landed in pieces.  This is a fairly major project and trying to land it all at once will cause it to take a lot longer.

benjaminsavage:
I agree with Martin. The key thing here will be to split this PR up into a few smaller PRs.

I would recommend adding the ability to invert a field element to be a separate PR all on its own. Could be a good place to start.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+            if i > 0 {
+                for j in x.iter().take(i) {
+                    basis_polynomial *= p[k] - *j;
+                }
+            }
+            if i + 1 < x.len() {
+                for j in x.iter().skip(i + 1) {
+                    basis_polynomial *= p[k] - *j;
+                }
+            }
```
benjaminsavage, line 177: Again, seems like it would be simpler to write:
```
for i in 0..x.len() {
  for j in 0..len() {
    if i != j {
      basis_polynomial *= p[k] - *j
    }
  }
}
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+/// given points `(x_0,y_0),...,(x_n,y_n)` on a polynomial `p` and points `p_0,...,p_m`
+/// the function computes `p` evaluated on points `p_0,...,p_m`
+///
+/// It follows the lagrange method to compute the points `https://en.wikipedia.org/wiki/Lagrange_polynomial`
+/// further, rather than outputting `p(p_k)`, we set `result_k=result_k*p(p_k)`
+/// which fits better to how we use `p(p_k)`
+///
+/// # Panics
+/// Panics when there are identical evaluation points, i.e. division by zero
+pub fn lagrange_evaluation<F>(x: &[F], y: &[F], p: &[F], result: &mut [F])
```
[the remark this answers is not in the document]
benjaminsavage: I think we should just have one of the two approaches, presumably the precomputed version. Otherwise we have two independent implementations to keep in sync and maintain.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+                }
+            }
+            if i + 1 < x.len() {
+                for j in x.iter().skip(i + 1) {
+                    basis_polynomial *= p[k] - *j;
+                }
+            }
+            sum += basis_polynomial;
+        }
+        // compute "result_k = result_k * p(p_k)"
```
benjaminsavage, line 180: Remind me why we want to compute this? Is it worth this optimization?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+        },
+        secret_sharing::SharedValue,
+    };
+
+    #[test]
+    fn consistency_at_evaluation_points_test() {
+        // use random evaluation points
+        let mut rng = thread_rng();
+        let e_p_size = rng.gen::<usize>() % 20;
+        let mut e_p = vec![Fp25519::ONE; e_p_size];
```
benjaminsavage, line 203: Why use Fp25519 for testing? I think we will actually want to use a Prime Field like the integers modulo the Mersenne Prime 2^61 - 1 in practice. Why not use that?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+        let mut y_values = vec![Fp25519::ZERO; e_p_size + 1];
+        for x_value in &e_p {
+            // monomial base, i.e. `x^k`
+            let mut base = Fp25519::ONE;
+            // evaluate p via `sum coefficient_k * x^k`
+            for coefficient in &p {
+                y_values.iter_mut().for_each(|y| *y = *coefficient * base);
+                base *= *x_value;
+            }
+        }
```
benjaminsavage, line 243: You lost me here. What is going on?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+        let mut evaluated_q = vec![Fp25519::ONE; r_p_size];
+        lagrange_evaluation(&e_p, &q, &r_p, &mut evaluated_q);
+
+        // precomputed
+        let mut evaluated_q_precomputed = vec![Fp25519::ONE; r_p_size];
+        let mut denominators = vec![Fp25519::ONE; e_p_size];
+        compute_lagrange_denominator(&e_p, &mut denominators);
+        let mut base = vec![vec![Fp25519::ONE; e_p_size]; r_p_size];
+        compute_lagrange_base(&r_p, &e_p, &denominators, &mut base);
+        lagrange_evaluation_precomputed(&q, &base, &mut evaluated_q_precomputed);
```
benjaminsavage, line 286: An alternative approach, instead of just comparing the output of two implementations of lagrange_evaluation, would be to generate a random polynomial (i.e. generate random coefficients), and sample the evaluation points and the test points by just evaluating the polynomial, then to compare the test points with the values obtained via lagrange interpolation.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        evaluated_q
+            .iter()
+            .zip(evaluated_q_precomputed.iter())
+            .for_each(|(x, y)| assert_eq!(*x, *y));
+    }
+
+    #[test]
+    fn lagrange_precomputed_random_test() {
```
benjaminsavage, line 295: How does this test differ from the "canonical" test above?

benjaminsavage (changes requested):
Just reviewed lagrange interpolation so far.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+#[derive(Step)]
+pub(crate) enum Step {
+    HashFromLeft,
+    HashFromRight,
+    PQFromLeft,
+    TwoOutOfTwoZeroCheck,
+}
+
+/// Non-Interactive Distributed Zero Knowledge Proof
+/// a `NIDZKP` which is computed over multiple recursions , is a vector of `g` polynomials
```
benjaminsavage, line 34:
```
/// a `NIDZKP` which is computed over multiple recursions, is a vector of `g` polynomials
```

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+/// the length is the amount of proof recursions
+/// `g` is secret shared using PRSS: left share is stored in struct, right share is generated via PRSS
+/// the length of the left share, `g_left`, determines the degree of the polynomial
+/// for a low communication rate, i.e. `log (amount of multiplications)`,
+/// the length needs to be constant in the amount of multiplications
+/// length might be different across recursions (therefore used `Vec` instead of `GenericArray`)
```
benjaminsavage, line 40: How are these two sentences not in conflict with one another? I understand that the degree of the polynomial can vary in each recursion, so what do you mean by "the length needs to be constant in the amount of multiplications"? What does that mean?

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+    F: Field,
+{
+    pub(crate) proofs: Vec<Vec<F>>,
+    pub(crate) mask_p: F,
+    pub(crate) mask_q: F,
+}
+
+/// generates recursive proof for `sum_i u_i*v_i = out` over field `F`
+///
+/// `r_f`, i.e. recursion factor, determines how much the inputs are compressed during each recursion
```
benjaminsavage, line 57: Since `r` usually is associated with the random point, and the paper we are following calls this either `lambda` or the `compression_factor`, can we use one of those terms?

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+
+#[derive(Step)]
+pub(crate) enum Step {
+    HashFromLeft,
+    HashFromRight,
+    PQFromLeft,
+    TwoOutOfTwoZeroCheck,
+}
+
+/// Non-Interactive Distributed Zero Knowledge Proof
```
benjaminsavage, line 33: Please add a reference to the paper. I'd prefer to use `https://eprint.iacr.org/2023/909.pdf` and point to section 4.1. 

It may also be worth pointing (additionally) to the paper that introduced this technique: section 3 of `https://eprint.iacr.org/2020/1451.pdf`

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+    debug_assert!(F::BITS > usize::BITS - r_f.leading_zeros() + 1);
+    // record, used as counter for PRSS
+    let mut record = 0usize;
+
+    // proof length
+    #[allow(
+        clippy::cast_possible_truncation,
+        clippy::cast_sign_loss,
+        clippy::cast_precision_loss
+    )]
```
benjaminsavage, line 98: Kind of scary... Do we need to ignore all of these warnings?

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+    let proof_length = (u.len() as f64).log(r_f as f64).ceil() as usize + 1usize;
+    // get masks from PRSS
+    let mask_p = ctx.prss().generate_fields(RecordId::from(record)); //(F::ONE, F::ONE);
+    record += 1;
+    let mask_q = ctx.prss().generate_fields(RecordId::from(record));
+    // proof from the prover on the left, i.e. generated via `PRSS.left` rather than receiving it from the left
```
benjaminsavage, line 104: We are inside the function `generate_proof` so it's kind of strange to talk about "receiving it" from anyone. This code is run by the prover, not received from anyone else.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+    };
+    // right part of this helpers own proof, to be sent to the left
+    let mut own_proof_right = NIDZKP {
+        proofs: Vec::with_capacity(proof_length),
+        mask_p: F::ZERO,
+        mask_q: F::ZERO,
+    };
+
+    // precomputation for Lagrange
+    let mut e_p = vec![F::ZERO; 2 * r_f];
```
benjaminsavage, line 120: I don't think you actually need to double it.

Let's take the example of `r_f = 4`, like I did in this Google Sheet: https://docs.google.com/spreadsheets/d/10tCSgiZyQbt4bMB3KqhAUOCfgfO8xIni83J2saipFCg/edit?usp=sharing, see the tab called "All Lagrange Interpolation".

In that case, the length of both `u` and `v` is 4. In all but the final round of recursion, we do not generate any random weights, meaning that each element of `u` and `v` can be interpolated with a degree 3 polynomial (maximum term is x^3).

So the Polynomial that is defined as `u /cdot v` is a degree 6 polynomial (maximum term is x^6). The uniquely define a degree 6 polynomial requires 7 points. We can easily compute 4 (by simply multiplying the 4 values of `u` and `v` provided), so we only need to compute 3 more. 

This is described in the 909 paper. In section 4.1 it says:

```
Next P_i computes a degree-2(λ − 1) polynomial...
```

I've worked it all out in the sheet. This is a nice optimization that saves one interpolation per iteration.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+    compute_lagrange_denominator(&e_p[0..r_f], &mut ds);
+    let mut base = vec![vec![F::ONE; r_f]; r_f];
+    compute_lagrange_base(&e_p[r_f..2 * r_f], &e_p[0..r_f], &ds, &mut base);
+
+    // iterate over recursions
+    while u.len() >= r_f {
+        // fill u and v with zeros such that it is a multiple of the recursion factor
+        let coset = u.len() % r_f;
+        if coset != 0 {
+            u.extend(std::iter::repeat(F::ZERO).take(coset));
```
benjaminsavage, line 133: An alternative approach, when `u.len() % r_f != 0` is to adjust `r_f` for that iteration.

Yes, this incurs the additional cost of precomputing additional lagrange coefficients, but the authors even used different values per step. They say:

```
Since the first round computation has been significantly improved, our protocol uses the compression parameter λ = 32 in the first round and λ′ = 8 for the rest of rounds
```

In my spreadsheet, I used λ = 4 for all the rounds until the final round when I used λ = 2, since that's all I needed (given the length of the input vector).

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+            v.extend(std::iter::repeat(F::ZERO).take(coset));
+        }
+
+        // amount of polynomials `p`, `q`, i.e. amount of `summand_of_g`
+        let length = u.len() / r_f;
+
+        // define `g` which is represented by `2*recursion_factor` many points.
+        let mut g = vec![F::ZERO; 2 * r_f];
```
benjaminsavage, line 141: Should be one less point.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+            record += 1;
+            let (left, right) = ctx.prss().generate_fields(RecordId::from(record));
+            // push `PRSS.left` to `output_g_left`
+            output_g_left.push(left);
+            // own_g_left:
+            own_g_left.push(right);
+            // compute secret share `own_g_right` by computing `own_g_right=g-own_g_left`
+            // we just need to subtract `own_g_left`, which comes from `PRSS.right`
+            *g -= right;
+        }
```
benjaminsavage, line 163: Let's discuss this over a call - I don't really follow right now and I'd like to understand.

ipa-core/src/protocol/ipa_prf/malicious_security/quadratic_proofs.rs
```diff
@@ -0,0 +1,933 @@
...
+{
+    // check lengths
+    debug_assert_eq!(u.len(), v.len());
+    debug_assert_eq!(base.len(), u.len());
+
+    // define summand of g, g = sum of summand_of_g
+    // summand_of_g is represented by 2*recursion_factor many points.
+    let mut summand_of_g = vec![F::ONE; 2 * r_f];
+
+    // compute first recursion factor many points of summand_of_g
```
benjaminsavage, line 262: You can compute one less than the recursion factor many points.

benjaminsavage (changes requested):
Next batch of comments. Will revisit this again tomorrow.

benjaminsavage:
I think we can close this PR now.

danielmasny:
done! : https://github.com/private-attribution/ipa/pull/1164

PR #932, opened 2024-01-25
Allow zero spare space in `OrderingSender`

andyleiserson:
Prompted by https://github.com/private-attribution/ipa/pull/852/files#r1464197048.

I compromised the quality of the assertion in `State::write` a little. It may not be as effective in catching broken configurations, but it should still prevent reaching a deadlock.

There are two ways I can see revising this:
* Get rid of the notion of spare entirely, and just require that messages are uniformly sized (less work).
* Use an enum to track the mode (spare capacity vs. uniform message size), and the observed message size in the uniform case. This would restore the quality of the assertion, but would be a bit more work to implement.

### DIFF
--- a/ipa-core/src/helpers/buffers/ordering_sender.rs
+++ b/ipa-core/src/helpers/buffers/ordering_sender.rs
@@ -28,10 +28,10 @@ use crate::{
 struct State {
     /// A store of bytes to write into.
     buf: Vec<u8>,
-    /// The portion of the buffer that is marked "spare".
-    /// Once `written + spare` is greater than the buffer capacity,
-    /// data is available to the stream.
-    spare: NonZeroUsize,
+    /// The portion of the buffer that is marked "spare". Once `written + spare` is greater than
+    /// the buffer capacity, data is available to the stream. May be zero if messages are uniformly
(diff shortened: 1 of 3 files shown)


ipa-core/src/helpers/buffers/ordering_sender.rs
```diff
@@ -71,31 +71,41 @@ impl State {
...
+            Self::save_waker(&mut self.write_ready, cx);
+            return Poll::Pending;
+        }
+
+        let b = &mut self.buf[self.written..];
         assert!(
-            M::Size::USIZE < self.spare.get(),
-            "expect message size {:?} to be less than spare {:?}",
+            M::Size::USIZE <= b.len(),
+            "expect message size {:?} to fit in available buffer; only {:?} of {:?} available",
```
martinthomson, line 93: A `#[should_panic]` test might be a good idea here.

martinthomson (approved):
The simplification alone justifies this.

I wonder if allowing for more buffering than what is sent in each write is warranted.  Though that would make this a lot more complicated.

akoshelev (approved):

PR #933, opened 2024-01-26
Generate valid field elements, and other fixes for RecordsStream proptest

### LINKED ISSUE #931: Fix flaky test `test_records_stream_works_with_any_chunks`
#905 made it an error to convert a number larger than prime to a field element. Before, it was converted using `mod PRIME` operation.

We have evidence that occasionally proptest generates a value between $2^{31}-5..2^{31}-1$: https://github.com/private-attribution/ipa/actions/runs/7620669438/job/20755775510, so we need to reject such inputs

andyleiserson:
Fixes a bug in `random_chunks` that showed up with `TestField = Fp31`, makes empty streams a candidate input, and fixes the greater-than-prime error in #931. The combination of `arb_size_in_bytes` and `arb_aligned_bytes` into `arb_expected_and_chunked_body` is not to fix a bug, it just seemed unnecessarily complicated. 

I got failures in Fp32BitPrime twice within 10-15 minutes, but I think I was lucky and O(hours) is a more realistic time frame. You can also change the TestField to Fp31 to get a failure immediately.

Fixes #931.

### DIFF
--- a/ipa-core/src/helpers/transport/stream/input.rs
+++ b/ipa-core/src/helpers/transport/stream/input.rs
@@ -692,7 +692,6 @@ mod test {
 
     mod prop_test {
         use futures::TryStreamExt;
-        use generic_array::GenericArray;
         use proptest::prelude::*;
         use rand::{rngs::StdRng, SeedableRng};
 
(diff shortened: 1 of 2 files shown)


ipa-core/src/helpers/transport/stream/input.rs
```diff
@@ -805,7 +794,7 @@ mod test {
 
         prop_compose! {
             fn random_items(max_len: usize, with_large: bool)
-                           (len in 1..=max_len)
+                           (len in 0..=max_len)
```
akoshelev, line 797: nice catch

akoshelev (approved):

PR #936, opened 2024-02-01
A draft API for validation of replicated shares

martinthomson:
This is a simple stream adapter that performs inline validation of anything that is replicated.  It will error at the end of the stream if it finds a mismatch.  (Future work might include periodic checkpoints.)

The reason I'm putting this up is that it is a great example of how our borrowed contexts are a giant pain in the posterior.  I'm starting to think that we need to think about moving to something with Arc so we can avoid this lifetime business.  It's a real drag.

### DIFF
--- a/ipa-core/src/protocol/basics/validate.rs
+++ b/ipa-core/src/protocol/basics/validate.rs
@@ -58,7 +58,7 @@ impl From<HashFunction> for HashValue {
 }
 
 struct ReplicatedValidatorFinalization {
-    f: Pin<Box<dyn Future<Output = Result<(), Error>>>>,
+    f: Pin<Box<(dyn Future<Output = Result<(), Error>> + Send)>>,
 }
 
 impl ReplicatedValidatorFinalization {
--- a/ipa-core/src/protocol/basics/validate.rs
+++ b/ipa-core/src/protocol/basics/validate.rs
@@ -23,6 +23,7 @@ use crate::{
     helpers::{Direction, Message},
     protocol::{context::Context, RecordId},
     secret_sharing::{replicated::ReplicatedSecretSharing, SharedValue},
+    seq_join::assert_send,
 };
 
 type HashFunction = Sha256;
@@ -37,7 +38,7 @@ impl Serializable for HashValue {
     type DeserializationError = Infallible;
 
     fn serialize(&self, buf: &mut GenericArray<u8, Self::Size>) {
-        buf.copy_from_slice(self.0.as_slice())
+        buf.copy_from_slice(self.0.as_slice());
     }
 
     fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Result<Self, Self::DeserializationError> {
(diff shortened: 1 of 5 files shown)


ipa-core/src/protocol/basics/validate.rs
```diff
@@ -0,0 +1,218 @@
...
+        // Ugh: The version of sha2 we currently use doesn't use the same GenericArray version as we do.
+        let left_hash = HashValue(GenericArray::from(<HashOutputArray>::from(
+            left_hash.finalize_fixed(),
+        )));
+        let right_hash = HashValue(GenericArray::from(<HashOutputArray>::from(
+            right_hash.finalize_fixed(),
+        )));
+        let left_peer = ctx.role().peer(Direction::Left);
+        let right_peer = ctx.role().peer(Direction::Left);
+        let ctx_ref = &ctx;
```
akoshelev, line 71: do you want to borrow here or it is acceptable to clone the context and move it inside the box?

If we want to borrow from `C`, we would need to create a self-referential struct, so maybe we need an inner struct that is pinned and holds `f` that borrows from `ctx` and `ctx`


however, I was able to compile it just by `let ctx = ctx.clone()` but that's probably not what you want here

ipa-core/src/protocol/basics/validate.rs
```diff
@@ -0,0 +1,218 @@
...
+        Self {
+            ctx,
+            left_hash: HashFunction::new(),
+            right_hash: HashFunction::new(),
+        }
+    }
+
+    fn update<S, V>(&mut self, s: &S)
+    where
+        S: ReplicatedSecretSharing<V>,
```
akoshelev, line 119: not sure if it is worth it, but you may just require `V: Serializable` here rather than `SharedValue`

ipa-core/src/protocol/basics/validate.rs
```diff
@@ -0,0 +1,218 @@
...
+impl Message for HashValue {}
+
+struct ReplicatedValidatorFinalization<C> {
+    f: Pin<Box<dyn Future<Output = Result<(), Error>>>>,
+    ctx: C,
+}
+
+impl<C: Context + 'static> ReplicatedValidatorFinalization<C> {
+    fn new(active: ReplicatedValidatorActive<C>) -> Self {
+        let ReplicatedValidatorActive {
```
akoshelev, line 57: it may be nicer from API's perspective to let `ReplicatedValidatorActive` to turn itself into a pair of hashes

ipa-core/src/protocol/basics/validate.rs
martinthomson: Yeah, I managed to get this to compile with a similar trick.  Though the tests are now failing because the `TestWorld` doesn't have a `'static` lifetime.  I put the `'static` condition on to avoid problems creating the future, ...

ipa-core/src/protocol/basics/validate.rs
martinthomson: `ReplicatedSecretSharing` requires that `V: SharedValue` so that doesn't really do anything.

ipa-core/src/protocol/basics/validate.rs
martinthomson: As this is just internal, I didn't do that.  Though I did implement `From` for `HashValue`, which made this function a little less ugly.

ipa-core/src/protocol/basics/validate.rs
```diff
@@ -0,0 +1,284 @@
...
+impl<C: Context + 'static> ReplicatedValidatorActive<C> {
+    fn new(ctx: C) -> Self {
+        Self {
+            ctx,
+            left_hash: HashFunction::new(),
+            right_hash: HashFunction::new(),
+        }
+    }
+
+    fn update<S, V>(&mut self, s: &S)
```
andyleiserson, line 118: For debugging, it might be useful if the update API takes both a `Step` and data, and creates a trace of the validator inputs. Then we can diagnose where things went wrong if there is a mismatch. (Obviously, we would want a flag so we only pay the cost of the detailed tracing when needed.)

martinthomson: The context carries a step, so we could smuggle that in somewhere.

I'm more concerned that I've been unable to instantiate this code :)

ipa-core/src/protocol/basics/validate.rs
```diff
@@ -0,0 +1,344 @@
...
+                let res = f.poll(cx);
+                *self = ReplicatedValidatorState::Finalizing(f);
+                res
+            }
+            Self::Finalizing(f) => f.poll(cx),
+        }
+    }
+}
+
+/// A `ReplicatedValidator` takes a stream of replicated shares of anything
```
danielmasny, line 177: does a stream ensure that the state of Sha2 is updated in the same order? A different order will result in a different hash.

danielmasny (commented):
Do you plan to run this validator just after decrypting the shares (to abort when user devices generate bad shares) or do you want to run it during the whole protocol e.g. for debugging purposes?

PR #940, opened 2024-02-04
Fix Arc imports

akoshelev:


martinthomson (approved):

martinthomson:
Is there some way to prevent this from happening again?  Autocomplete tends not to notice our overrides.

Also, I think we just hit a disk space limitation on the coverage build.

akoshelev:
> Is there some way to prevent this from happening again? Autocomplete tends not to notice our overrides.

I tried to use clippy's `disallowed-types` for it, but it didn't work because it flagged all usages of `crate::sync::Arc` as disallowed too.

I searched a bit and it seems the only way is to create a separate crate and re-export Arc and friends from it

akoshelev:
> I searched a bit and it seems the only way is to create a separate crate and re-export Arc and friends from it

yep, #941 should fix it

PR #944, opened 2024-02-07
Fix collect_steps DEPTH_DYNAMIC_STEPS string

richajaindce:
Solve for #921


akoshelev:
wow, thanks @richajaindce for catching that

akoshelev (approved):

PR #947, opened 2024-02-08
Enable Tokio tasks profiler

akoshelev:
Requires `logging::setup()`, but otherwise works smoothly by running [`tokio-console`](https://github.com/tokio-rs/console) client that consumes metrics emitted by IPA benchmarks and shows per-task and combined stats

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -63,6 +63,10 @@ descriptive-gate = []
 compact-gate = ["ipa-macros/compact-gate"]
 # Enable using more than one thread for protocol execution. Most of the parallelism occurs at parallel/seq_join operations
 multi-threading = ["async-scoped"]
+# Enable tokio task profiling. Requires tokio_unstable flag to be passed to the compiler.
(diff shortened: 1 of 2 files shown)


ipa-core/Cargo.toml
```diff
@@ -135,7 +140,7 @@ sha2 = "0.10"
 shuttle-crate = { package = "shuttle", version = "0.6.1", optional = true }
 thiserror = "1.0"
 time = { version = "0.3", optional = true }
-tokio = { version = "1.35", features = ["fs", "rt", "rt-multi-thread", "macros"] }
+tokio = { version = "1.35", features = ["fs", "rt", "rt-multi-thread", "macros", "tracing"] }
```
andyleiserson, line 143: Could there be a cost to enabling this feature?

andyleiserson (approved):

ipa-core/Cargo.toml
akoshelev: for sure, I am going to take it only if `tokio-console` is enabled

PR #954, opened 2024-02-15
Move u128 conversions out of Field trait

andyleiserson:
There are a lot of changed files, but most of them are only changed to update references to `Field` or `U128Conversions` traits. The core change here is that the `truncate_from` and `as_u128` methods are removed from the `Field` trait and put in the new `U128Conversions` trait, because the u128 conversions do not make sense on `Fp25519`.

There are some [more substantial upcoming changes](https://github.com/andyleiserson/ipa/compare/u128...andyleiserson:ipa:field-cleanup?expand=1) that will resolve the remaining issues with the `Field` trait and clean up the TODOs in this PR.


akoshelev:
oof merging it with ipa v2 removal change will be fun

akoshelev (approved):

PR #956, opened 2024-02-16
Delete the old IPA implementation

akoshelev:
I had to add more clippy checks because I noticed we don't run it when `web-app` feature is enabled, so we've accumulated quite a bit of debt there.

Maybe controversial, but I removed the radix sort implementation, mainly because it is not compatible with the new shuffle protocol. I think if we want to use it, we need to rewrite it anyway.


ipa-core/src/bin/report_collector.rs
```diff
@@ -181,7 +154,7 @@ fn gen_inputs(
 ) -> io::Result<()> {
     let rng = seed
         .map(StdRng::seed_from_u64)
-        .unwrap_or_else(|| StdRng::from_entropy());
+        .unwrap_or_else(StdRng::from_entropy);
```
akoshelev, line 157: here and below - clippy error fixes

ipa-core/src/protocol/basics/mul/semi_honest.rs
```diff
@@ -89,6 +91,26 @@ where
     Ok(Replicated::new_arr(lhs, rhs))
 }
 
+/// Implement secure multiplication for semi-honest contexts with replicated secret sharing.
+#[async_trait]
+impl<C, F, const N: usize> super::SecureMul<C> for Replicated<F, N>
```
akoshelev, line 96: moved things around to avoid slapping descriptive-gate feature flag on too many things

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -32,6 +27,16 @@ pub trait Reveal<C: Context, B: RecordBinding>: Sized {
     async fn reveal<'fut>(&self, ctx: C, record_binding: B) -> Result<Self::Output, Error>
     where
         C: 'fut;
+
+    /// partial reveal protocol to open a shared secret to all helpers except helper `left_out` inside the MPC ring.
+    async fn partial_reveal<'fut>(
```
akoshelev, line 32: `PartialReveal` trait is removed too, I don't see a why `partial_reveal` should be on its own trait.

ipa-core/src/query/processor.rs
```diff
@@ -602,26 +604,55 @@ mod tests {
         }
 
         async fn ipa_query(app: &TestApp) -> Result<(), BoxError> {
```
akoshelev, line 606: `GenericReportTestInput` and associated macro `ipa_test_input!` are gone

danielmasny (approved):
Looks good to me. I noticed you also changed some inputs to use a value rather than reference. There might be other parts that we can remove (e.g. some of the files in ff). We can always removed them later once we have a clear picture what we still need.

ipa-core/src/protocol/basics/partial_reveal.rs
```diff
@@ -1,37 +1,13 @@
...
-        },
-        SharedValue,
-    },
+    protocol::{context::Context, RecordBinding, RecordId},
+    secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
 };
 
-/// Trait for partial reveal protocol to open a shared secret to all helpers except helper `left_out` inside the MPC ring.
 #[async_trait]
 pub trait PartialReveal<C: Context, B: RecordBinding>: Sized {
```
andyleiserson, line 9: Did you mean to delete this entirely, or are you thinking that can wait for a follow-up?

andyleiserson (approved):
Could you explain the issue with descriptive-gate? Compact gate doesn't work with malicious security any more for some reason?

ipa-core/src/protocol/basics/partial_reveal.rs
akoshelev: oh thanks for catching that - I meant to completely remove it but left some stuff behind. Going to delete the definition as well

akoshelev:
> Could you explain the issue with descriptive-gate? Compact gate doesn't work with malicious security any more for some reason?

Yes that is our common source of pain - anything that we don't execute as part of `collect_steps` cannot be compiled with `compact-gate` feature enabled. We don't run OPRF IPA with malicious security enabled, hence the need to gate the malicious implementation behind a feature flag

PR #957, opened 2024-02-17
Precomputed Lagrange

danielmasny:
cleaned up version of Lagrange

### DIFF
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -1,4 +1,4 @@
-use std::fmt::Display;
+use std::{fmt::Display, mem};
 
 use generic_array::GenericArray;
 
@@ -14,6 +14,43 @@ pub trait PrimeField: Field {
     type PrimeInteger: Into<u128>;
 
     const PRIME: Self::PrimeInteger;
+
+    /// Invert function that returns the multiplicative inverse
+    /// the default implementation uses the extended Euclidean algorithm,
+    /// follows inversion algorithm in
+    /// (with the modification that it works for unsigned integers by keeping track of `sign`):
+    /// `https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm`
+    ///
+    /// The function operates on `u128` rather than field elements since we need divisions
+    ///
+    /// ## Panics
+    /// When `self` is `Zero`
+
+    #[must_use]
+    fn invert(&self) -> Self {
+        assert_ne!(*self, Self::ZERO);
+
+        let mut t = 0u128;
+        let mut newt = 1u128;
+        let mut r = Self::PRIME.into();
+        let mut newr = self.as_u128();
+        let mut sign = 1u128;
+
+        while newr != 0u128 {
+            let quotient = r / newr;
+            mem::swap(&mut t, &mut newt);
+            mem::swap(&mut r, &mut newr);
+            newt += quotient * t;
+            newr -= quotient * r;
+
+            // flip sign
+            sign = 1 - sign;
+        }
+
+        // when sign is negative, output `PRIME-t` otherwise `t`
+        // unwrap is safe
+        Self::try_from((1 - sign) * t + sign * (Self::PRIME.into() - t)).unwrap()
+    }
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -293,6 +330,14 @@ macro_rules! field_impl {
                     buf.copy_from_slice(&v.to_le_bytes());
                     let err = $field::deserialize(&buf).unwrap_err();
                     assert!(matches!(err, GreaterThanPrimeError(..)))
+                }
+
+                    #[test]
+                fn test_invert(element: $field) {
+                    if element != $field::ZERO
+                    {
+                        assert_eq!($field::ONE,element * element.invert() );
+                    }
                 }
             }
         }
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -25,7 +25,6 @@ pub trait PrimeField: Field {
     ///
     /// ## Panics
     /// When `self` is `Zero`
-
     #[must_use]
     fn invert(&self) -> Self {
         assert_ne!(*self, Self::ZERO);
@@ -36,7 +35,7 @@ pub trait PrimeField: Field {
         let mut newr = self.as_u128();
         let mut sign = 1u128;
 
-        while newr != 0u128 {
+        while newr != 0 {
             let quotient = r / newr;
             mem::swap(&mut t, &mut newt);
             mem::swap(&mut r, &mut newr);
@@ -332,8 +331,8 @@ macro_rules! field_impl {
                     assert!(matches!(err, GreaterThanPrimeError(..)))
                 }
 
-                    #[test]
-                fn test_invert(element: $field) {
+                #[test]
+                fn invert(element: $field) {
                     if element != $field::ZERO
                     {
                         assert_eq!($field::ONE,element * element.invert() );
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
@@ -21,7 +21,7 @@ pub struct CanonicalLagrangeDenominator<F: Field, N: ArrayLength> {
 
 impl<F, N> CanonicalLagrangeDenominator<F, N>
 where
-    F: PrimeField,
+    F: PrimeField + TryFrom<u128>,
     N: ArrayLength,
 {
     /// generates canonical Lagrange denominators
@@ -70,7 +70,7 @@ pub struct LagrangeTable<F: Field, N: ArrayLength, M: ArrayLength> {
 
 impl<F, N> LagrangeTable<F, N, U1>
 where
-    F: Field,
+    F: Field + TryFrom<u128>,
     N: ArrayLength,
 {
     /// generates a `CanonicalLagrangeTable` from `CanoncialLagrangeDenominators` for a single output point
@@ -122,7 +122,7 @@ where
     /// When the field size is too small for `N` evaluation points
     fn compute_table_row(x_output: &F, table_row: &mut GenericArray<F, N>)
     where
-        F: Field,
+        F: Field + TryFrom<u128>,
         N: ArrayLength,
     {
         for (entry, i) in table_row.iter_mut().zip(0u64..) {
@@ -135,7 +135,7 @@ where
 
 impl<F, N, M> From<CanonicalLagrangeDenominator<F, N>> for LagrangeTable<F, N, M>
 where
-    F: Field,
+    F: Field + TryFrom<u128>,
     N: ArrayLength,
     M: ArrayLength,
 {
@@ -209,7 +209,7 @@ mod test {
 
     impl<F, N> From<MonomialFormPolynomial<F, N>> for Polynomial<F, N>
     where
-        F: Field,
+        F: Field + TryFrom<u128>,
         N: ArrayLength,
     {
         fn from(value: MonomialFormPolynomial<F, N>) -> Self {
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
@@ -1,9 +1,9 @@
 use std::iter;
 
 use generic_array::{ArrayLength, GenericArray};
-use typenum::U1;
+use typenum::{Unsigned, U1};
 
-use crate::ff::{Field, PrimeField};
+use crate::ff::{Field, PrimeField, Serializable};
 
 /// A degree `N-1` polynomial is stored as `N` points `(x,y)`
 /// where the "x coordinates" of the input points are `x_0` to `x_N` are `F::ZERO` to `(N-1)*F::ONE`
@@ -32,8 +32,16 @@ where
     /// When the field size is too small for `N` evaluation points
     pub fn new() -> Self {
         // assertion that field is large enough
-        // also checks that `try_from` for conversions from sufficiently small `u128` to `F` do not panic
-        debug_assert!(F::BITS > usize::BITS - N::USIZE.leading_zeros());
+        // when it is large enough, `F::try_from().unwrap()` below does not panic
+        assert!(
+            F::BITS > usize::BITS - N::USIZE.leading_zeros(),
+            "Field size {} is not large enough to hold {} points",
+            F::BITS,
+            N::USIZE
+        );
+
+        // assertion that table is not too large for the stack
+        assert!(<F as Serializable>::Size::USIZE * N::USIZE < 2024);
 
         let mut denominator = iter::repeat(F::ONE)
             .take(N::USIZE)
@@ -48,10 +56,18 @@ where
     }
 }
 
-/// `LagrangeTable` is a precomputation table for the Lagrange evaluation.
-/// The "x coordinates" of the input points are `x_0` to `x_(N-1)` are `F::ZERO` to `(N-1)*F::ONE`.
-/// The `LagrangeTable` also specifies `M` "x coordinates" for the output points
-/// The "x coordinates" of the output points are `x_N` to `x_(N+M-1)` are `N*F::ONE` to `(N+M-1)*F::ONE`.
+/// `LagrangeTable` is a precomputed table for the Lagrange evaluation.
+/// Allows to compute points on the polynomial, i.e. output points,
+/// given enough points on the polynomial, i.e. input points,
+/// by using the `eval` function.
+/// The "x coordinates" are implicit.
+/// The "y coordinates" of the input points are inputs to `eval`.
+/// The output of `eval` are the "y coordinates" of the output points .
+/// The "x coordinates" of the input points `x_0` to `x_(N-1)` are `F::ZERO` to `(N-1)*F::ONE`.
+/// The `LagrangeTable` also specifies `M` "x coordinates" for the output points.
+/// The "x coordinates" of the output points `x_N` to `x_(N+M-1)` are `N*F::ONE` to `(N+M-1)*F::ONE`
+/// when generated using `from(denominator)`
+/// unless generated using `new(denominator, x_output)` for a specific output "x coordinate" `x_output`.
 pub struct LagrangeTable<F: Field, N: ArrayLength, M: ArrayLength> {
     table: GenericArray<GenericArray<F, N>, M>,
 }
@@ -64,6 +80,9 @@ where
     /// generates a `CanonicalLagrangeTable` from `CanoncialLagrangeDenominators` for a single output point
     /// The "x coordinate" of the output point is `x_output`.
     pub fn new(denominator: CanonicalLagrangeDenominator<F, N>, x_output: &F) -> Self {
+        // assertion that table is not too large for the stack
+        assert!(<F as Serializable>::Size::USIZE * N::USIZE < 2024);
+
         let mut table = denominator.denominator;
         Self::compute_table_row(x_output, &mut table);
         LagrangeTable::<F, N, U1> {
@@ -103,7 +122,7 @@ where
         }
     }
 
-    /// helper function to compute a single row of `CanonicalLagrangeTable`
+    /// helper function to compute a single row of `LagrangeTable`
     ///
     /// ## Panics
     /// When the field size is too small for `N` evaluation points
@@ -128,8 +147,17 @@ where
 {
     fn from(value: CanonicalLagrangeDenominator<F, N>) -> Self {
         // assertion that field is large enough
-        // also checks that `try_from` for conversions from sufficiently small `u128` to `F` do not panic
-        debug_assert!(F::BITS > usize::BITS - (N::USIZE + M::USIZE).leading_zeros());
+        // when it is large enough, `F::try_from().unwrap()` below does not panic
+        assert!(
+            F::BITS > usize::BITS - (N::USIZE + M::USIZE).leading_zeros(),
+            "Field size {} is not large enough to hold {} + {} points",
+            F::BITS,
+            N::USIZE,
+            M::USIZE
+        );
+
+        // assertion that table is not too large for the stack
+        assert!(<F as Serializable>::Size::USIZE * N::USIZE * M::USIZE < 2024);
 
         let mut table = iter::repeat(value.denominator.clone())
             .take(M::USIZE)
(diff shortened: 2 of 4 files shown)


ipa-core/src/ff/prime_field.rs
```diff
@@ -294,6 +331,14 @@ macro_rules! field_impl {
                     let err = $field::deserialize(&buf).unwrap_err();
                     assert!(matches!(err, GreaterThanPrimeError(..)))
                 }
+
+                    #[test]
```
martinthomson, line 335: Indentation looks off here.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut denominator = iter::repeat(F::ONE)
+            .take(N::USIZE)
+            .collect::<GenericArray<_, N>>();
+        for (i, d) in denominator.iter_mut().enumerate() {
+            for j in (0usize..N::USIZE).filter(|&j| i != j) {
+                *d *= F::try_from(i as u128).unwrap() - F::try_from(j as u128).unwrap();
+            }
+            *d = d.invert();
+        }
+        CanonicalLagrangeDenominator { denominator }
```
martinthomson, line 47:
```suggestion
        Self { denominator }
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl<F, N> LagrangeTable<F, N, U1>
+where
+    F: Field,
+    N: ArrayLength,
+{
+    /// generates a `CanonicalLagrangeTable` from `CanoncialLagrangeDenominators` for a single output point
+    /// The "x coordinate" of the output point is `x_output`.
+    pub fn new(denominator: &CanonicalLagrangeDenominator<F, N>, x_output: &F) -> Self {
```
martinthomson, line 66: You should accept denominator by value if you are just going to clone it.  That way, the caller is able to choose whether or not to pass a value or a reference.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut result = iter::repeat(F::ONE)
+            .take(M::USIZE)
+            .collect::<GenericArray<_, _>>();
+        self.mult_result_by_evaluation(polynomial, &mut result);
+        result
+    }
+
+    /// This function uses the `LagrangeTable` to evaluate `polynomial` on the specified output "x coordinates"
+    /// the "y coordinates" of the evaluation are multiplied to `result`
+    pub fn mult_result_by_evaluation(
```
martinthomson, line 93: Does this need to be public?  It looks like it assumes a particular form for `result` that a caller isn't obligated to guarantee.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn from(value: CanonicalLagrangeDenominator<F, N>) -> Self {
+        // assertion that field is large enough
+        // also checks that `try_from` for conversions from sufficiently small `u128` to `F` do not panic
+        debug_assert!(F::BITS > usize::BITS - N::USIZE.leading_zeros() - M::USIZE.leading_zeros());
+
+        let mut table = iter::repeat(value.denominator.clone())
+            .take(M::USIZE)
+            .collect::<GenericArray<_, _>>();
+        table.iter_mut().enumerate().for_each(|(i, row)| {
+            compute_table_row(&F::try_from((i + N::USIZE) as u128).unwrap(), row);
```
martinthomson, line 122:
```suggestion
            compute_table_row(&F::try_from(u128::from(i + N::USIZE)).unwrap(), row);
```
I believe that `u128` implements `From<usize>`.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+/// ## Panics
+/// When the field size is too small for `N` evaluation points
+fn compute_table_row<F, N>(x_output: &F, table_row: &mut GenericArray<F, N>)
+where
+    F: Field,
+    N: ArrayLength,
+{
+    for (i, entry) in table_row.iter_mut().enumerate() {
+        for j in (0usize..N::USIZE).filter(|&j| j != i) {
+            *entry *= *x_output - F::try_from(j as u128).unwrap();
```
martinthomson, line 139:
```suggestion
            *entry *= *x_output - F::try_from(u128::from(j)).unwrap();
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+        });
+        LagrangeTable { table }
+    }
+}
+
+/// helper function to compute a single row of `CanonicalLagrangeTable`
+///
+/// ## Panics
+/// When the field size is too small for `N` evaluation points
+fn compute_table_row<F, N>(x_output: &F, table_row: &mut GenericArray<F, N>)
```
martinthomson, line 132: This can be a function on the struct rather than an orphan.  Invoke it with `Self::compute_row`.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+    F: Field,
+    N: ArrayLength,
+    M: ArrayLength,
+{
+    /// This function uses the `LagrangeTable` to evaluate `polynomial` on the specified output "x coordinates"
+    /// outputs the "y coordinates" such that `(x,y)` lies on `polynomial`
+    pub fn eval(&self, polynomial: &Polynomial<F, N>) -> GenericArray<F, M> {
+        let mut result = iter::repeat(F::ONE)
+            .take(M::USIZE)
+            .collect::<GenericArray<_, _>>();
```
martinthomson, line 86: Can't this use:

```suggestion
        let mut result = arr![F::ONE; M::USIZE];
```

or maybe:


```suggestion
        let mut result = arr![F::ONE; M];
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+                .collect::<GenericArray<_, _>>();
+            Polynomial {
+                y_coordinates: value.eval(&canonical_points),
+            }
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn test_lagrange_single_output_point_using_new(output_point: TestField, input_points in prop::array::uniform32(any::<TestField>())){
```
martinthomson, line 211:
```suggestion
        fn lagrange_single_output_point_using_new(output_point: TestField, input_points in prop::array::uniform32(any::<TestField>())){
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+            let polynomial = Polynomial::from(polynomial_monomial_form.clone());
+            let denominator = CanonicalLagrangeDenominator::<TestField,U32>::new();
+            // generate table using new
+            let lagrange_table = LagrangeTable::<TestField,U32,U1>::new(&denominator,&output_point);
+            let output = lagrange_table.eval(&polynomial);
+            assert_eq!(output,output_expected);
+        }
+
+        #[test]
+        fn test_lagrange_cannonical_using_from(input_points in prop::array::uniform8(any::<TestField>()))
```
martinthomson, line 225:
```suggestion
        fn lagrange_cannonical_using_from(input_points in prop::array::uniform8(any::<TestField>()))
```

ipa-core/src/ff/prime_field.rs
```diff
@@ -294,6 +331,14 @@ macro_rules! field_impl {
...
+                fn test_invert(element: $field) {
```
martinthomson, line 336:
```suggestion
                fn invert(element: $field) {
```

martinthomson (commented):
I like the restructuring.

akoshelev:
you seem to have added a flaky test, we need to fix it before we can check the coverage 

thread 'protocol::ipa_prf::malicious_security::lagrange::test::test_lagrange_cannonical_using_from' panicked at ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs:209:5:
Test failed: attempt to subtract with overflow.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: There are some parts in the DZKP (within the provers computation) where I want to compute q(x')*p(x') for some x coordinate x' so it is convenient do use this function and pass q(x') as result which is mutated to q(x')*p(x'). There is the more general function eval for the general case.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: I first tried that but get:

```
error[E0277]: the trait bound `u128: From<usize>` is not satisfied
   --> ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs:122:44
    |
122 |             compute_table_row(&F::try_from(u128::from(i + N::USIZE)).unwrap(), row);
    |                                            ^^^^ the trait `From<usize>` is not implemented for `u128`
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+    F: PrimeField,
+    N: ArrayLength,
+{
+    /// generates canonical Lagrange denominators
+    ///
+    /// ## Panics
+    /// When the field size is too small for `N` evaluation points
+    pub fn new() -> Self {
+        // assertion that field is large enough
+        // also checks that `try_from` for conversions from sufficiently small `u128` to `F` do not panic
```
akoshelev, line 35: I am not sure I follow what you're trying to say here, specifically about sufficiently small `u128` - you seem to check `F` against usize

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+        debug_assert!(F::BITS > usize::BITS - N::USIZE.leading_zeros());
```
akoshelev, line 36:
```suggestion
        assert!(F::BITS > usize::BITS - N::USIZE.leading_zeros(), "Field size {x} is not large enough to hold {N} points`);
```

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        let mut denominator = iter::repeat(F::ONE)
+            .take(N::USIZE)
+            .collect::<GenericArray<_, N>>();
+        for (i, d) in denominator.iter_mut().enumerate() {
+            for j in (0usize..N::USIZE).filter(|&j| i != j) {
+                *d *= F::try_from(i as u128).unwrap() - F::try_from(j as u128).unwrap();
+            }
+            *d = d.invert();
```
akoshelev, line 45: what is the reason for caching denominator values only? For polynomial of degree $k$, isn't $l_j(k)$ are all the same if we assume $x_0,...,x_{k-1}$ to be $[0,k)$?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// `LagrangeTable` is a precomputation table for the Lagrange evaluation.
+/// The "x coordinates" of the input points are `x_0` to `x_(N-1)` are `F::ZERO` to `(N-1)*F::ONE`.
+/// The `LagrangeTable` also specifies `M` "x coordinates" for the output points
+/// The "x coordinates" of the output points are `x_N` to `x_(N+M-1)` are `N*F::ONE` to `(N+M-1)*F::ONE`.
+pub struct LagrangeTable<F: Field, N: ArrayLength, M: ArrayLength> {
+    table: GenericArray<GenericArray<F, N>, M>,
+}
+
+impl<F, N> LagrangeTable<F, N, U1>
```
akoshelev, line 59: why do we need this specialization?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+            *d = d.invert();
+        }
+        CanonicalLagrangeDenominator { denominator }
+    }
+}
+
+/// `LagrangeTable` is a precomputation table for the Lagrange evaluation.
+/// The "x coordinates" of the input points are `x_0` to `x_(N-1)` are `F::ZERO` to `(N-1)*F::ONE`.
+/// The `LagrangeTable` also specifies `M` "x coordinates" for the output points
```
akoshelev, line 53: I don't understand this - what are the output points?

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+            .take(M::USIZE)
+            .collect::<GenericArray<_, _>>();
+        table.iter_mut().enumerate().for_each(|(i, row)| {
+            compute_table_row(&F::try_from((i + N::USIZE) as u128).unwrap(), row);
+        });
+        LagrangeTable { table }
+    }
+}
+
+/// helper function to compute a single row of `CanonicalLagrangeTable`
```
akoshelev:
```suggestion
/// helper function to compute a single row of [`LagrangeTable`]
/// [`LagrangeTable`]: LagrangeTable
```

in this case you won't miss updating comments after renaming

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+    M: ArrayLength,
+{
+    fn from(value: CanonicalLagrangeDenominator<F, N>) -> Self {
+        // assertion that field is large enough
+        // also checks that `try_from` for conversions from sufficiently small `u128` to `F` do not panic
+        debug_assert!(F::BITS > usize::BITS - N::USIZE.leading_zeros() - M::USIZE.leading_zeros());
+
+        let mut table = iter::repeat(value.denominator.clone())
+            .take(M::USIZE)
+            .collect::<GenericArray<_, _>>();
```
akoshelev, line 136: I worry about us accidentally thrashing on stack creating those arrays. If `N` or `M` is sufficiently large, this may become too large to be stack-allocated. In this form we don't have any protection against it. 

We could either limit `N`, `M` and `F::SIZE` values such that the product is less than 256 bytes (or 1k) or using heap allocations here. The latter may work well if we only need to support a few different degrees $k$

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    impl<F, N> From<MonomialFormPolynomial<F, N>> for Polynomial<F, N>
+    where
+        F: Field,
+        N: ArrayLength,
+    {
+        fn from(value: MonomialFormPolynomial<F, N>) -> Self {
+            let canonical_points: GenericArray<F, N> = (0..N::USIZE)
+                .map(|i| F::try_from(i as u128).unwrap())
```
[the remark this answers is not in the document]
danielmasny: The issue is that from usize is not implemented for a u128. The same for u64 and u32.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+        CanonicalLagrangeDenominator { denominator }
+    }
+}
+
+/// `LagrangeTable` is a precomputation table for the Lagrange evaluation.
+/// The "x coordinates" of the input points are `x_0` to `x_(N-1)` are `F::ZERO` to `(N-1)*F::ONE`.
+/// The `LagrangeTable` also specifies `M` "x coordinates" for the output points
+/// The "x coordinates" of the output points are `x_N` to `x_(N+M-1)` are `N*F::ONE` to `(N+M-1)*F::ONE`.
+pub struct LagrangeTable<F: Field, N: ArrayLength, M: ArrayLength> {
```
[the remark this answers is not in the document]
danielmasny: Its more general this way. If you want to load something precomputed for a field for which we dont have invert implemented, e.g. from a file, you could. Invert is only needed to generate the denominator, otherwise we don't need inversion.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct Polynomial<F: Field, N: ArrayLength> {
+    y_coordinates: GenericArray<F, N>,
+}
+
+/// The Canonical Lagrange denominator is defined as the denominator of the Lagrange base polynomials
+/// `https://en.wikipedia.org/wiki/Lagrange_polynomial`
+/// where the "x coordinates" of the input points are `x_0` to `x_N` are `F::ZERO` to `(N-1)*F::ONE`
+/// the degree of the polynomials is `N-1`
+pub struct CanonicalLagrangeDenominator<F: Field, N: ArrayLength> {
```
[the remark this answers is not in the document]
danielmasny: we first need to compute the denominators, then we can compute the table. It makes sense to separate the two since the denominator only needs the input points while Lagrange table also needs the output points. So we can use the same denominator for several Lagrange tables. In the proofs, we need a table for the output points N,...2*N-1 as well as a table for each random challenge r (which is generated as the hash of the "subproof" of the previous iteration). The size of the denominator is likely to be around 8-32 field elements.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: We can make it more general, however we won't need it. M is the amount of output x-coordinates and this implementation allows to generate a Lagrange table for a single output point using `new(x_output, denominator)`. In the proofs, we need to evaluate the polynomials on a single x-coordinate r which is random. Otherwise we always evaluate polynomials at fixed x-coordinates N...N+M in which case we can just use the implementation of From(denominator) below

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: N,M will range from 1 up to 32. So the max would be 32*31*F::SIZE but this is an extreme case that we might not implement this way. The common case is probably 8*7*F::SIZE which is about 448 bytes. At any point, there will be at most one such table plus another 8*1*F::SIZE table plus a 8*F::SIZE denominator. (F::SIZE will be 8 bytes, at most 16bytes)

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
benjaminsavage: It makes sense to cache the denominators as they are the same across all the Lagrange coefficients used to evaluate the polynomial at any point. This helps us avoid re-computing them. This is the same approach I used in my spreadsheet: https://docs.google.com/spreadsheets/d/10tCSgiZyQbt4bMB3KqhAUOCfgfO8xIni83J2saipFCg/edit?usp=sharing

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
benjaminsavage: I would prefer not to embed this logic into the Lagrange module. I understand it might even be slightly more efficient this way - but it feels like we are violating an important separation of concerns.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,282 @@
...
+        let output = lagrange_table.eval(&polynomial);
+        assert_eq!(output, output_expected);
+    }
+
+    proptest! {
+    #[test]
+    fn proptest_lagrange_single_output_point_using_new(output_point: TestField, input_points in prop::array::uniform32(any::<TestField>())){
+        lagrange_single_output_point_using_new(output_point,input_points);
+    }
+    }
```
benjaminsavage, line 256: I'm not sure what is going on here... but it looks wrong....

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    fn lagrange_canonical_using_from(input_points: [TestField; 8]) {
+        let polynomial_monomial_form = MonomialFormPolynomial {
+            coefficients: GenericArray::<TestField, U8>::from_array(input_points),
+        };
+        // the canonical x coordinates are 0..15, the outputs use coordinates 8..15:
+        let x_coordinates_output = GenericArray::<_, U7>::generate(|i| {
+            TestField::try_from(u128::try_from(i).unwrap() + 8).unwrap()
+        });
```
benjaminsavage, line 265: Honestly, I find the generics here(e.g. U7, U8) really confusing. I wish there was some better way.

benjaminsavage (commented):
@danielmasny is out on paternity leave. How should we take this forward? Should I just land this PR as is and make a few minor changes in a follow up PR? That seems like the simplest approach. @martinthomson ?

martinthomson:
Forwards is a good direction to move and this code is close.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: I am fine with removing it since you seem to prefer it that way.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
danielmasny: I just use the proptest macro here. I added a function since the formatting is not checked within macro calls.

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn proptest_lagrange_single_output_point_using_new(output_point: TestField, input_points in prop::array::uniform32(any::<TestField>())){
+        lagrange_single_output_point_using_new(output_point,input_points);
+    }
+    }
+
+    fn lagrange_canonical_using_from(input_points: [TestField; 8]) {
+        let polynomial_monomial_form = MonomialFormPolynomial {
+            coefficients: GenericArray::<TestField, U8>::from_array(input_points),
+        };
+        // the canonical x coordinates are 0..15, the outputs use coordinates 8..15:
```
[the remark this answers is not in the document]
danielmasny: Yes

Issue #951, opened 2024-02-09
Consider changing `FromRandom` to operate on `u8` arrays


andyleiserson:
Our AES-based PRSS produces outputs in multiples of 16 bytes, so in that context, `u128` is a natural choice for the `FromRandom` input. (Also because it evolved from the `FromRandomU128` trait.) But `u128`s are not always easy to work with, so it may make sense to use `u8` arrays here instead.

PR #958, opened 2024-02-21
Clean up `Field` trait

andyleiserson:
* Remove `Field` trait impl from Boolean arrays.
* Remove u128 conversions from `Fp25519`.
* Add `select` variant of `if_else`. `select` is bus multiplexer with a single-bit control input. `if_else` is a vectorizable multiplexer, with condition input the same width as the data inputs.
* Remove `Field` (or replace with `SharedValue`) in various trait bounds.
* Implement `Vectorizable` and related traits for more boolean arrays.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -180,8 +180,6 @@ macro_rules! impl_serializable_trait {
 
         #[cfg(all(test, unit_test))]
         mod fallible_serialization_tests {
-            use rand::{thread_rng, Rng};
-
             use super::*;
 
             /// [`https://github.com/private-attribution/ipa/issues/911`]
@@ -197,8 +195,6 @@ macro_rules! impl_serializable_trait {
                     "Padding only makes sense for lengths that are not multiples of 8."
                 );
 
-                let mut rng = thread_rng();
-
                 let mut non_zero_padding = $name::ZERO.0;
                 non_zero_padding.set($bits, true);
                 assert_eq!(
@@ -212,10 +208,6 @@ macro_rules! impl_serializable_trait {
                 let mut max_value = $name::ZERO.0;
                 max_value[..$bits].fill(true);
                 deserialize(max_value).unwrap();
-
-                let mut rnd_value = $name::ZERO.0;
-                rnd_value[..$bits].fill_with(|_| rng.gen());
-                deserialize(rnd_value).unwrap();
             }
         }
     };
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -1,5 +1,5 @@
 use bitvec::{
-    prelude::{bitarr, BitArr, Lsb0},
+    prelude::{BitArr, Lsb0},
     slice::Iter,
 };
 use generic_array::GenericArray;
@@ -9,7 +9,7 @@ use crate::{
     error::LengthError,
     ff::{boolean::Boolean, ArrayAccess, ArrayBuilder, Field, Serializable, U128Conversions},
     protocol::prss::{FromRandom, FromRandomU128},
-    secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
+    secret_sharing::{Block, SharedValue, StdArray, Vectorizable},
 };
 
 /// The implementation below cannot be constrained without breaking Rust's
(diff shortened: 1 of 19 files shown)


ipa-core/src/secret_sharing/mod.rs
```diff
@@ -211,9 +240,39 @@ impl<F: Field> FieldSimd<1> for F {}
...
+}
 
-impl FieldSimd<256> for Boolean {}
+boolean_vector!(3, BA3);
+boolean_vector!(5, BA5);
+boolean_vector!(8, BA8);
+boolean_vector!(20, BA20);
+boolean_vector!(32, BA32);
+boolean_vector!(64, BA64);
+boolean_vector!(256, BA256);
```
martinthomson, line 275: The set of sizes we generate values for isn't consistent.  I hope that is OK.  Will someone be surprised if they try to invoke something for which there is one implementation and not another?

ipa-core/src/ff/boolean_array.rs
```diff
@@ -78,64 +78,11 @@ where
     }
 }
 
-/// A value of ONE has a one in the first element of the bit array, followed by `$bits-1` zeros.
-/// This macro uses a bit of recursive repetition to produce those zeros.
-///
-/// The longest call is 8 bits, which involves `2(n+1)` macro expansions in addition to `bitarr!`.
-macro_rules! bitarr_one {
```
akoshelev, line 85: so we don't need this awesome macro anymore? :)

ipa-core/src/ff/boolean_array.rs
```diff
@@ -262,12 +209,13 @@ macro_rules! impl_serializable_trait {
...
                 deserialize(min_value).unwrap();
 
-                let one = $name::ONE.0;
-                deserialize(one).unwrap();
-
                 let mut max_value = $name::ZERO.0;
                 max_value[..$bits].fill(true);
                 deserialize(max_value).unwrap();
+
+                let mut rnd_value = $name::ZERO.0;
```
akoshelev, line 216: does it check something that [this](https://github.com/private-attribution/ipa/blob/7e9432f7bfebb30bb57c0b181e8445f369049873/ipa-core/src/ff/boolean_array.rs#L748) test doesn't already?

ipa-core/src/protocol/basics/if_else.rs
```diff
@@ -34,3 +49,47 @@ where
...
+    true_value: &B,
+    false_value: &B,
+) -> Result<B, Error>
+where
+    C: Context,
+    B: Clone + BooleanArrayMul,
+{
+    let false_value = false_value.clone().into();
+    let true_value = true_value.clone().into();
+    let condition = B::expand(condition).into();
```
akoshelev, line 80: maybe use `From` so the types for these values become obvious?

ipa-core/src/protocol/basics/mul/mod.rs
```diff
@@ -39,3 +50,71 @@ pub trait SecureMul<C: Context>: Send + Sync + Sized {
...
+        + for<'a> Add<&'a Self::Vectorized, Output = Self::Vectorized>
+        + for<'a> Sub<&'a Self::Vectorized, Output = Self::Vectorized>
+        + 'static;
+
+    fn multiply<'fut, C>(
+        ctx: C,
+        record_id: RecordId,
+        a: &'fut Self::Vectorized,
+        b: &'fut Self::Vectorized,
+    ) -> impl Future<Output = Result<Self::Vectorized, Error>> + Send + 'fut
```
akoshelev, line 75: 👍

ipa-core/src/ff/ec_prime_field.rs
```diff
@@ -193,43 +191,17 @@ impl Field for Fp25519 {
...
-impl TryFrom<u128> for Fp25519 {
-    type Error = crate::error::Error;
-
-    fn try_from(v: u128) -> Result<Self, Self::Error> {
-        let mut bits = [0u8; 32];
-        bits[..].copy_from_slice(&v.to_le_bytes());
-        let f: Fp25519 = Fp25519::ONE;
-        f.serialize((&mut bits).into());
-        Ok(f)
+    fn from_random(src: GenericArray<u128, Self::SourceLength>) -> Self {
```
danielmasny, line 197: Would it make sense to use a u8 as a base and increase the SourceLength? I think even Aes uses a GenericArray over u8

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -171,10 +172,13 @@ where
...
     C::UpgradedContext<F>: UpgradedContext<F, Share = Replicated<F>>,
-    BK: SharedValue + U128Conversions + CustomArray<Element = Boolean> + Field,
-    TV: SharedValue + U128Conversions + CustomArray<Element = Boolean> + Field,
-    TS: SharedValue + U128Conversions + CustomArray<Element = Boolean> + Field,
-    SS: SharedValue + U128Conversions + CustomArray<Element = Boolean> + Field,
+    BK: SharedValue + U128Conversions + CustomArray<Element = Boolean>,
+    TV: SharedValue + U128Conversions + CustomArray<Element = Boolean>,
+    TS: SharedValue + U128Conversions + CustomArray<Element = Boolean>,
+    SS: SharedValue + U128Conversions + CustomArray<Element = Boolean>,
+    Replicated<BK>: BooleanArrayMul,
```
akoshelev, line 179: I guess you tried it already, but I am going to ask anyway - have you tried to provide blanket implementation `BooleanArrayMul` for all boolean shares?

danielmasny (approved):
Thanks for cleaning this up. I just have one small comment.

ipa-core/src/ff/ec_prime_field.rs
akoshelev: yea we have #951 for that

ipa-core/src/protocol/ipa_prf/mod.rs
andyleiserson: It's another consequence of const generic limitations -- see [this comment](https://github.com/private-attribution/ipa/pull/958/files#diff-17a4607905850bd30e0208dce5fc0540e1ec37fbfce53cdb60aa5858de93d807R56-R60).

Part of why things get weird here is that we're using the record-wise vectorization support for bit-wise vectorization. To make a fully vectorized version of these protocols we'd need to change that, and could probably get rid of BooleanArrayMul in the process.

ipa-core/src/secret_sharing/mod.rs
andyleiserson: For vectorization in general, because rust's const generic support is limited, the supported vectorizations need to be enumerated explicitly in the source. I did the same thing here (listing only the ones that we use). Also, arbitrary vectorization dimensions are only needed where we're using the record-wise vectorization support to implement bit-wise vectorization within a record. Implementing only the necessary vectorizations helps to discourage that.

ipa-core/src/ff/boolean_array.rs
andyleiserson: No, I don't think it does. I added this one to replace the test case in this function that deserialized the ONE value, but I can drop it in favor of the other test.

akoshelev (approved):

PR #959, opened 2024-02-22
Rewritten arithmetic circuit benchmark

andyleiserson:
The previous version of the benchmark did not use seq_join. Using seq_join is necessary for parallel execution, and also provides a fairer baseline for unvectorized serial operation.

This change also generates test data outside the timed region (except under the iai framework, where that is not supported).

### DIFF
--- a/ipa-core/src/test_fixture/circuit.rs
+++ b/ipa-core/src/test_fixture/circuit.rs
@@ -105,7 +105,7 @@ pub async fn arithmetic<F, const N: usize>(
         .collect::<Vec<_>>()
         .try_into()
     else {
-        panic!("infallible try_into array");
+        unreachable!("infallible try_into array");
     };
 
     let (res0, res1, res2) = join3(fut0, fut1, fut2).await;
--- a/ipa-core/src/test_fixture/circuit.rs
+++ b/ipa-core/src/test_fixture/circuit.rs
@@ -1,10 +1,11 @@
-use futures_util::future::join_all;
+use std::{array, num::NonZeroUsize};
+
+use futures::{future::join3, stream, StreamExt};
 use rand::distributions::{Distribution, Standard};
 
-use super::join3v;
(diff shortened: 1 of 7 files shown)


martinthomson (commented):
Coverage is reporting that the new code isn't covered.  Are we running benches as tests for the purposes of getting good coverage?

ipa-core/src/test_fixture/circuit.rs
```diff
@@ -1,89 +1,141 @@
...
-    }
+    let [inp0, inp1, inp2] = input_data;
+
+    let Ok([fut0, fut1, fut2]): Result<[_; 3], _> = contexts
+        .into_iter()
+        .zip([inp0, inp1, inp2])
+        .map(|(ctx, col_data)| {
+            // Setting TotalRecords::Indeterminate causes OrderingSender to make data available to
+            // the channel immediately, instead of doing so only after active_work records have
+            // accumulated. This gives the best performance for vectorized operation.
```
martinthomson, line 93: This is a little bit of a concern.

akoshelev:
> Coverage is reporting that the new code isn't covered. Are we running benches as tests for the purposes of getting good coverage?

that's a good idea, we should run it with small width and depth, just to make sure the benchmark is written correctly

ipa-core/src/test_fixture/circuit.rs
andyleiserson: We probably want to do one or more of the following:

1. Decouple OrderingSender capacity from send threshold.
2. Decouple active_work for seq_join from OrderingSender capacity.
3. Allow out-of-order writes into the OrderingSender buffer.

Setting TotalRecords::Indeterminate is a crude way of doing #&#x2060;1. I'd like to get some data from a more realistic test case before doing more than that.

akoshelev (approved):

PR #961, opened 2024-02-27
Compact step reprise

martinthomson:
A working version of a new `CompactStep` and `CompactGate` implementation, automatically derived from annotations on `XxxStep` types.

This works in two phases: build.rs and main.

## build.rs phase
Here, code is generated during the execution of `build.rs`.  This is incomplete code, but only for `CompactGate`, which only includes stubbed implementations of `AsRef<str>` and `From<&str>`.

`Step` implementations generate complete implementations of `CompactStep`, which includes a bunch of helpful functions that allow for traversing the tree of steps.  Some of these functions are not used in the main code, but they are used in a `build.rs` file.

The `build.rs` file includes (using the `include!()` macro, hidden by a convenience macro) all of the files that contain steps.  It then runs a build process that generates a file for each `CompactGate` target.  This will create implementations of `AsRef<str>` and `From<&str>` that the main build can pick up.

## main build phase

Here, the same code is generated, but an `include!()` call will be made to pick up the implementations generated by `build.rs`.

## Usage

The derivation of `CompactStep` is a proc-macro.  This is moderately complex, but not overtly so.  Here is a complete example of what an annotated `Step` might look like:

```rust
use ipa_step_derive::{CompactStep, CompactGate};
#[derive(CompactStep, CompactGate)]
enum MyStep {
  Empty,
  #[step(count = 10)]
  Int(u8),
  #[step(child = OtherStep)]
  OperationWithOtherSteps,
  #[step(child = OtherStep, count = 2, name = "short-name")]
  SetOfOperationsWithOtherSteps(u8),
}
```

This is a top-level step as well, so it requires that it is put in a file on its own, so that it can be included in `build.rs`.

The `build.rs` file looks fairly simple (I tried to compress this further, but ran into issues building a macros-by-example wrapper for this, so this tiny bit of repetition will have to do for now):

```rust
use ipa_step::{build_gate, load_steps, track_steps};
load_steps!(path::to::my_step);
fn main() {
  track_steps!(path::to::my_step);
  build_gate::<path::to::my_step::MyStep>();
}
```

The two macros list modules (or modules and files, `load_steps!(foo @ "src/foo/mod.rs")` might be necessary, though I doubt it will be common).  `load_steps!()` ensures that the modules are loaded with a module path that works for generating files.  `track_steps!()` ensures that `build.rs` is re-run if anything changes in those modules.  Finally, `build_gate()` generates the code necessary for each gate and puts that in a file in `$OUT_DIR`, where the proc-macro for the corresponding `CompactGate` generation will be able to find and `include!()` it. The build process uses the complete implementation of `CompactStep` on the target type to guide the building of the implementation.

## Example

There is an `ipa-step-test` crate in the branch that demonstrates how this might be used.  That includes some tests that demonstrate that the the `StepNarrow` implementation is working and correct.

## TODO

This currently duplicates the `Step`-related machinery in `ipa-core`.  The goal is to migrate that code out and start to use this instead.  Unfortunately, there is no easy path to transition, so I'll have to just go in and annotate and move everything around.

I'm not 100% sure that each step only has a single child step type.  If we have any places where two different types are found, some tweaking might be needed.

I haven't implemented anything for `struct Foo(u8)`, only `enum Foo { Bar(u8), }`.  That might require a small amount of adjustment.


akoshelev (approved):
This looks awesome, I can't wait to start using it!

ipa-step-derive/src/variant.rs
```diff
@@ -0,0 +1,359 @@
...
+        step_narrow_arms: &mut TokenStream,
+    ) -> ExtendedSum {
+        // Unpack so that we can use `quote!()`.
+        let VariantAttribute {
+            ident: step_ident,
+            name: step_name,
+            integer: None,
+            child: step_child,
+        } = self
+        else {
```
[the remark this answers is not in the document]
akoshelev: oh, didn't know you can do that

ipa-core/src/helpers/transport/query/oprf_shuffle.rs
martinthomson, line 1: What is going on here?

ipa-core/src/protocol/boolean/mod.rs
```diff
@@ -1,2 +1,53 @@
...
+/// A step generator for bitwise secure operations.
+///
+/// For each record, we decompose a value into bits (i.e. credits in the
+/// Attribution protocol), and execute some binary operations like OR'ing each
+/// bit. For each bitwise secure computation, we need to "narrow" the context
+/// with a new step to make sure we are using an unique PRSS.
+///
+/// This is a temporary solution for narrowing contexts until the infra is
+/// updated with a new step scheme.
+pub trait BitStep: Step + From<usize> {
```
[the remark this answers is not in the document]
martinthomson: Maybe we can parameterize this one further...  `pub struct NBitStep<const N: u32>`.  Then we can do away with the bit depth bit...

ipa-core/src/protocol/boolean/step.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl From<usize> for TwoHundredFiftySixBitOpStep {
+    fn from(v: usize) -> Self {
+        Self::Bit(v)
+    }
+}
+
+#[cfg(test)]
+#[derive(CompactStep)]
```
martinthomson, line 58: I guess we need to look into what the codecov warnings mean here.  Maybe we need to suppress them.

ipa-core/src/protocol/context/mod.rs
```diff
@@ -150,13 +140,15 @@ pub trait UpgradedContext<F: ExtendableField>: Context {
         T: Send,
         for<'a> UpgradeContext<'a, Self, F>: UpgradeToMalicious<'a, T, M>,
     {
-        #[cfg(feature = "descriptive-gate")]
+        #[cfg(descriptive_gate)]
```
martinthomson, line 143: I guess we'll have to deal with upgrades later.

ipa-core/src/protocol/ipa_prf/aggregation/step.rs
```diff
@@ -0,0 +1,36 @@
...
+    #[step(child = BucketStep)]
+    MoveToBucket,
+    #[step(count = 32, child = AggregateValuesStep)]
+    Aggregate(usize),
+}
+
+#[derive(CompactStep)]
+pub enum BucketStep {
+    /// should be equal to `MAX_BREAKDOWNS`
+    #[step(count = 512, child = crate::protocol::boolean::step::BoolAndStep)]
```
martinthomson, line 14:
```suggestion
    #[step(count = 512, child = crate::protocol::boolean::step::BoolAndStep), name = "b"]
```

@benjaminsavage was sad that this said "bit" rather than "bucket", so "bucket" is also fine.  I just prefer to use shorter names when possible.

We could also make this `struct BucketStep(usize)`, which could be even shorter.

ipa-core/src/protocol/ipa_prf/boolean_ops/step.rs
```diff
@@ -0,0 +1,31 @@
+use ipa_step_derive::CompactStep;
+
+/// FIXME: This step is not generic enough to be used in the `saturated_addition` protocol.
+/// It constraints the input to be at most 2 bytes and it will panic in runtime if it is greater
+/// than that. The issue is that compact gate requires concrete type to be put as child.
+/// If we ever see it being an issue, we should make a few implementations of this similar to what
+/// we've done for bit steps
+#[derive(CompactStep)]
+pub(crate) enum SaturatedAdditionStep {
+    #[step(child = crate::protocol::boolean::step::SixteenBitStep)]
```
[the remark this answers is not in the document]
martinthomson: Yeah, this seems like it is something we can work on.  Issue?

ipa-core/src/protocol/step/steps.txt
martinthomson, line 1: \o/

I won't miss this file.

ipa-core/src/query/executor.rs
```diff
@@ -147,6 +145,18 @@ where
     }
 }
 
+#[cfg(descriptive_gate)]
+fn prss_gate() -> Gate {
+    ipa_step::descriptive::Descriptive::default().narrow("prss")
```
martinthomson, line 150: Is this split necessary?  ProtocolStep should still be available here.

ipa-core/src/test_fixture/test_gate.rs
```diff
@@ -0,0 +1,64 @@
...
+use ipa_step::StepNarrow;
+
+use crate::{protocol::Gate, test_fixture::step::TestExecutionStep};
+
+/// This manages the gate information for test runs. Most unit tests want to have multiple runs
+/// using the same instance of [`TestWorld`], but they don't care about the name of that particular
+/// gate.
+///
+/// A few tests and some benchmarks want to execute the actual protocol, but they also care about
+/// performance they use compact gate to manage step transitions.
```
martinthomson, line 12:
```suggestion
/// performance, so they use compact gate to manage step transitions.
```

ipa-core/src/test_fixture/world.rs
```diff
@@ -249,10 +254,20 @@ impl<S: ShardingScheme> TestWorld<S> {
...
             _phantom: PhantomData,
         }
     }
+
+    #[must_use]
+    pub(crate) fn gate(&self) -> Gate {
+        self.gate_vendor.current()
+    }
+
+    fn next_gate(&self) -> Gate {
```
martinthomson, line 268:
```suggestion
    #[must_use]
    fn next_gate(&self) -> Gate {
```

pre-commit
```diff
@@ -79,7 +79,7 @@ check() {
 }
 
 check "Benchmark compilation" \
-    cargo build --benches --no-default-features --features "enable-benches descriptive-gate"
+    cargo build --benches --no-default-features --features "enable-benches compact-gate"
```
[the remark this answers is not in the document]
martinthomson: I'm happy leaving this as mutually exclusive.  We can run both.  As long as we don't have too many `#[cfg...` bits, we're not risking too much.

scripts/coverage-ci
```diff
@@ -10,14 +10,14 @@ cargo build --all-targets
...
 # Need to be kept in sync manually with tests we run inside check.yml.
 cargo test
-# disabled until 961 is merged
-# for gate in "compact" "descriptive"; do
-#  cargo test --no-default-features --features "cli web-app real-world-infra test-fixture $gate-gate"
-# done
-cargo test --no-default-features --features "cli web-app real-world-infra test-fixture descriptive-gate"
 
-cargo test --bench oneshot_ipa --no-default-features --features "enable-benches descriptive-gate" -- -n 62 -c 16
+# descriptive-gate is enabled by default
```
martinthomson, line 14:
```suggestion
# descriptive-gate is enabled when compact-gate is not
```

scripts/coverage-ci
```diff
@@ -10,14 +10,14 @@ cargo build --all-targets
...
+for gate in "compact-gate" ""; do
+  cargo test --no-default-features --features "cli web-app real-world-infra test-fixture $gate"
+done
 
-cargo test --bench criterion_arithmetic --no-default-features --features "enable-benches descriptive-gate"
+cargo test --bench oneshot_ipa --no-default-features --features "enable-benches" -- -n 62 -c 16
```
martinthomson, line 19: This might benefit from a comment:

```suggestion
# Test that our benchmark works when compact-gate is disabled
cargo test --bench oneshot_ipa --no-default-features --features "enable-benches" -- -n 62 -c 16
```

ipa-core/src/query/executor.rs
```diff
@@ -147,6 +145,18 @@ where
...
+}
+
+#[cfg(compact_gate)]
+fn prss_gate() -> Gate {
+    use crate::protocol::step::{ProtocolGate, ProtocolStep};
+
+    ProtocolGate::default().narrow(&ProtocolStep::Prss)
+}
```
benjaminsavage, line 158: Kind of strange that there is no test coverage for this.

ipa-step-derive/src/sum.rs
```diff
@@ -0,0 +1,57 @@
+use std::ops::Add;
+
+use proc_macro2::{Literal, Punct, Spacing, TokenStream};
+use quote::ToTokens;
+
+/// A complex sum, comprised of multiple statements, plus a simple counter.
+/// This implements `ToTokens` so that it can be dropped into `quote!()` easily.
+#[derive(Default, Clone)]
+pub struct ExtendedSum {
```
benjaminsavage, line 9: Where is this used?

benjaminsavage (approved):
I will admit I do not really understand the parsing / generating / proc_macro code. But I do understand the step names and annotations. I'm happy for this to be merged.

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -139,7 +131,7 @@ mod test {
                         .collect::<Vec<_>>();
 
                     validate_three_two_way_sharing_of_zero(
-                        ctx.narrow(&Step::Correctness),
+                        ctx.narrow("correctness"),
```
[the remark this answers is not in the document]
danielmasny: I will change it back once it is needed. I will need it for the proof validation which I will implement next week.

ipa-core/src/helpers/transport/query/oprf_shuffle.rs
akoshelev: good point, I also noticed a lot of empty files in main. Will follow up with a PR to remove it

ipa-core/src/protocol/boolean/mod.rs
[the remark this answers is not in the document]
akoshelev: this requires some changes inside the proc macro to support const generics. We can't use constants in macro attributes, so the way to make it work could be compact step reading the generic parameters and deriving the number of unique steps from it. Not sure if we can make it work, but it is worth experimenting with this

ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
```diff
@@ -6,36 +6,28 @@
...
 };
 
 /// Comparison operation
 ///
 /// Outputs x>=y for length(x) >= log2(y).
 /// # Errors
 /// Propagates errors from multiply
-#[cfg(all(test, unit_test))]
+#[allow(dead_code)]
 pub async fn compare_geq<C, S>(
```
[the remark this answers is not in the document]
akoshelev: maybe @danielmasny had some plans for it?

ipa-step-derive/src/sum.rs
akoshelev: this is used inside the compact gate module to keep track of number of steps in a tree

PR #962, opened 2024-02-28
Vectorized reveal

andyleiserson:
This change also removes the `B: RecordBinding` type parameter on the `Reveal` trait, which is no longer used after removal of IPAv2.

The complexity with vectorized reveal is converting the reveal output from `[T; 1]` to `T` in unvectorized protocols. For general handling of vectorized shares, there are two sets of methods (on the `AdditiveShare` type) -- the existing `new`, `left`, and `right` methods only work for unvectorized shares; the `new_arr`, `left_arr`, and `right_arr` methods return vector shares (even in the case where the vector dimension is one). We could do something similar with reveal, but it would be more complicated because it's traits instead of types. It doesn't seem worth it for the relatively small number of places that we call reveal.

### DIFF
--- a/ipa-core/src/protocol/basics/reveal.rs
+++ b/ipa-core/src/protocol/basics/reveal.rs
@@ -215,7 +215,7 @@ mod tests {
     use crate::{
         error::Error,
         ff::{Field, Fp31, Fp32BitPrime},
-        helpers::Direction,
+        helpers::{Direction, Role},
         protocol::{
             basics::Reveal,
             context::{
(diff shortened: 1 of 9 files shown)


ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -50,26 +52,33 @@ pub trait Reveal<C: Context, B: RecordBinding>: Sized {
...
 #[embed_doc_image("reveal", "images/reveal.png")]
-impl<C: Context, V: SharedValue> Reveal<C, RecordId> for Replicated<V> {
-    type Output = V;
+impl<C: Context, V: SharedValue + Vectorizable<N>, const N: usize> Reveal<C, N>
+    for Replicated<V, N>
+{
+    type Output = <V as Vectorizable<N>>::Array;
 
-    async fn reveal<'fut>(&self, ctx: C, record_id: RecordId) -> Result<V, Error>
+    async fn reveal<'fut>(
```
akoshelev, line 60: want to try your `Cow` trick for RPITIT here? we can always do it later though

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -190,7 +198,7 @@ impl<'a, F: ExtendableField> Reveal<UpgradedMaliciousContext<'a, F>, RecordId>
             .await?;
 
             if share_from_left == share_from_right {
-                Ok(Some(left + right + share_from_left))
+                Ok(Some((left + right + share_from_left).into_array()))
```
akoshelev, line 201: we probably want to use this version of malicious reveal, so it would be worth covering this line with tests

akoshelev (approved):

ipa-core/src/protocol/basics/reveal.rs
andyleiserson: I ended up with something [a bit different from the `Cow` solution](https://github.com/private-attribution/ipa/blob/57b4d4952534bdcd0c7d481a5d070596486b5068/ipa-core/src/protocol/basics/mul/mod.rs#L79).

I'm still not confident that either solution is reasonable to use 100% of the time, but I can try to use it here. I'd rather do it as a follow-up though.

Issue #909, opened 2024-01-04
Remove dependency on `async_trait` and migrate to RPITIT


akoshelev:
[Rust 1.75](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html) brought RPITIT which is great as it removes the need to box futures and dynamic dispatch. For traits like `SecureMul` this could bring performance benefits because we use multiplications a lot. 

I tried to remove the use of `async_trait` from our core traits (my [branch](https://github.com/private-attribution/ipa/compare/main...akoshelev:raw-ipa:rust-175-allow-boxed?expand=1)) and ran into https://github.com/rust-lang/rust/issues/100013. It would be great to try again once this issue is fixed

andyleiserson:
Capturing some analysis/discussion here:

### `Cow` workaround

https://github.com/private-attribution/ipa/blob/122c6c9bd34a6fdf49c0160af6a70265c0082ef9/ipa-core/src/protocol/basics/mul/mod.rs#L73-L80

Full diff: https://github.com/andyleiserson/ipa/commit/122c6c9bd34a6fdf49c0160af6a70265c0082ef9

When there is code like this:

```rust
fn inner<'fut>(arg: &'fut SomeType) -> impl Future + 'fut { ... }

fn outer(...) -> impl Future {
    let local = ...;

    inner(&local)
}
``` 

In the async fn version, it seems like the compiler is smart enough to let you do that -- something like local is moved into the future so that the lifetime on the reference can be sufficient. Or maybe it's just that in the async fn version, you don't have to write the 'fut lifetime explicitly, and the compiler infers something that is less strict than the version with explicit 'fut lifetimes on the references. Making inner take a Cow means you can pass Cow::Owned(local) in the case above (this passing owned local into the function, avoiding lifetime issues), but still pass Cow::Borrowed(ref_arg) when you have outer(ref_arg: &'fut SomeType).

(Possibly there is something different about RPITIT vs. regular RPIT, in which case I'm really talking about RPITIT in this example -- I haven't examined that specifically.)

The trickier part is figuring out where to add explicit 'fut lifetime parameters. I don't know of any better way to do that than trial and error looking for which piece of code makes the HRTB errors go away, and then fiddle with that piece of code.

[...]

PR #965, opened 2024-03-06
RPITIT and address a TODO for reveal

andyleiserson:
Two changes here (separated into two commits) related only in that they both affect reveal.

The first change eliminates async_trait for the `Reveal` trait, using the same wrapper function workaround for HRTB issues that I used for `BooleanArrayMul`.

The second change uses a single generic reveal implementation to support both full and partial reveals, reducing duplication. 

I have a different version of the second change that specifies whether a reveal is full or partial (and if partial, which helper is excluded) as a type parameter to the generic implementation. It would offer the following:
* Avoid the `unwrap` in the trait-provided implementation of `reveal`. (The unwrap is perfectly safe as long as the code stays as-is, just unsightly.)
* Prevent writing code that decides which helper to exclude from a partial reveal at runtime, since such code is likely to break TotalRecords determinism restrictions.

However, it ends up being more than a few lines of type-generic (i.e. hard to read) code, so I think it's probably not worth it. But I can share it if there's interest.


ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -206,6 +210,20 @@ impl<'a, F: ExtendableField> Reveal<UpgradedMaliciousContext<'a, F>, 1> for Mali
     }
 }
 
+// Workaround for https://github.com/rust-lang/rust/issues/100013. Calling this wrapper function
+// instead of `Reveal::reveal` seems to hide the `impl Future` GAT.
+pub fn reveal<'fut, C, S>(
```
akoshelev, line 178: just curious if using `pub async fn reveal(...) -> Result<S::Output,..>` works too

akoshelev:
> However, it ends up being more than a few lines of type-generic (i.e. hard to read) code, so I think it's probably not worth it. But I can share it if there's interest.

if that's easy, you could post a link to this PR and we can see if it ends up being not that hard. I agree that it is probably an overkill, but we could leave it here for educational purposes

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -182,13 +148,14 @@ impl<'a, F: ExtendableField> Reveal<UpgradedMaliciousContext<'a, F>, 1> for Mali
...
-        }
-        if ctx.role().peer(Direction::Right) != left_out {
-            right_sender.send(record_id, left).await?;
-        }
-        if ctx.role() == left_out {
+        let send_left_fut = (Some(ctx.role().peer(Direction::Left)) != left_out)
+            .then(|| left_sender.send(record_id, right));
+        let send_right_fut = (Some(ctx.role().peer(Direction::Right)) != left_out)
+            .then(|| right_sender.send(record_id, left));
+        ctx.parallel_join(send_left_fut.into_iter().chain(send_right_fut))
```
martinthomson, line 155: A parallel join involves a bunch of machinery.  Is there a cheaper way to join two `Option<impl Future>` ?

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -119,53 +124,14 @@ impl<C: Context, V: SharedValue + Vectorizable<N>, const N: usize> Reveal<C, N>
...
-    }
-
-    async fn partial_reveal<'fut>(
-        &self,
+    async fn generic_reveal<'fut>(
+        &'fut self,
         ctx: UpgradedMaliciousContext<'a, F>,
         record_id: RecordId,
-        left_out: Role,
+        left_out: Option<Role>,
```
martinthomson, line 134: An observation, unrelated to this change:

"left" out is a poor name in this context because of the use of "left" and "right" as subjective labels for helpers.  "skipped" or "excluded" perhaps?

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -50,60 +80,35 @@ pub trait Reveal<C: Context, const N: usize>: Sized {
...
     ) -> Result<Option<<V as Vectorizable<N>>::Array>, Error>
     where
         C: 'fut,
     {
         let left = self.left_arr();
         let right = self.right_arr();
 
-        // send except to left_out
-        if ctx.role().peer(Direction::Right) != left_out {
+        // send except to excluded helper (if any)
```
martinthomson, line 101:
```suggestion
        // Send shares, unless the target helper is excluded
```

andyleiserson:
> > However, it ends up being more than a few lines of type-generic (i.e. hard to read) code, so I think it's probably not worth it. But I can share it if there's interest.
> 
> if that's easy, you could post a link to this PR and we can see if it ends up being not that hard. I agree that it is probably an overkill, but we could leave it here for educational purposes

Here it is: https://github.com/private-attribution/ipa/commit/56b3de68a72c2fa866dd5930978cfb484da9b6a4

Or as a diff vs. the version in this PR: https://github.com/andyleiserson/ipa/compare/e7793d60bd22d6c5298ee54cb020378d06cd84af..reveal-alt?expand=1

akoshelev:
> Here it is: [56b3de6](https://github.com/private-attribution/ipa/commit/56b3de68a72c2fa866dd5930978cfb484da9b6a4)

I kinda like it - it does look heavy on the trait implementation side, but using it is quite easy. Not insisting on using it though

ipa-core/src/protocol/basics/reveal.rs
andyleiserson: It does not... my hypothesis is it's related to this issue noted in my comment on #909:

> The compiler is smart enough to see that if you have &local in the async move future, it can make that work by moving local into the future. However, this often results in the HRTB errors when the compiler needs to prove a statement of the form "for any lifetime 'local describing &local in the anonymous future, some other anonymous lifetime like 'ctx must outlive 'local.", which looks like `for <'local> 'ctx: 'local`, which the compiler can only reduce to `'ctx: 'static`.

ipa-core/src/protocol/basics/reveal.rs
andyleiserson: What I really wanted here is the native `try_join_all`. To avoid the parallel join overhead without trying to use the disallowed method, I added a `MaybeFuture` helper so that we can treat each of the left and right actions as futures whether or not they're actually sending.

PR #966, opened 2024-03-07
add command line flag to collect_steps.py for multi-threading

eriktaubeneck:
adding a small update to the `collect_steps.py` script to have the option for multithreading.

### DIFF
--- a/scripts/collect_steps.py
+++ b/scripts/collect_steps.py
@@ -170,9 +170,9 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
-    features = "enable-benches debug-trace step-trace"
+    features = ["enable-benches", "debug-trace", "step-trace"]
     if args.multi_threading:
-        features = f"{features} multi-threading"
+        features.append("multi-threading")
 
     ARGS = [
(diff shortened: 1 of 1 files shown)


scripts/collect_steps.py
```diff
@@ -165,14 +155,39 @@ def ipa_steps():
...
+    parser.add_argument(
+        "-m",
+        "--multi-threading",
+        action="store_true",
+    )
+    args = parser.parse_args()
+
+    features = "enable-benches debug-trace step-trace"
+    if args.multi_threading:
+        features = f"{features} multi-threading"
```
martinthomson, line 175: Shouldn't this use a list and `" ".join()`?

eriktaubeneck: if we wanted to make this more flexible (e.g., add other flags too), that's how I'd do it. @akoshelev suggested the steps concept was going away soon, so I just did this quick an dirty.

akoshelev: we are probably still going to use steps, but this script should go away after #961

akoshelev (approved):

Issue #969, opened 2024-03-09
In-memory Horizonal scaling API design proposal


akoshelev:
This issue exists to describe the proposed approach to adopt horizontal sharding in IPA. I will refer to it when submitting pull requests because the overall change is large enough to warrant a PR split.

## Current state

Some hierarchies are intentionally simplified as being not relevant to the proposed design. 

```mermaid
---
title: Current state
---
classDiagram
    direction LR
    class Message
    <<interface>> Message 
    class ChannelId{
         +role: Role,
         +gate: Gate
    }

    SemiHonestContext o-- Gateway
    Gateway o-- Transport
    class Transport{
       <<interface>>
       +send(dest: HelperIdentity, route: Route, data: Stream)
       +receive(from: HelperIdentity, route: Route)
    }
    class Gateway{ 
        +get_sender~Message~(dest: ChannelId)
        +get_receiver~Message~(origin: ChannelId)
    }
    class SemiHonestContext{
        + Gate gate
        +send_channel~Message~(dest: Role)
        +recv_channel~Message~(origin: Role)
    }
       InMemoryTransport  ..|> Transport
    HttpTransport  ..|> Transport
```

We use channel identifier to uniquely identify a communication channel between two MPC helpers. `Message` trait is used to restrict data that can be sent between helpers (field values - OK, shares - Not OK). 

### Channels

Channels become indexed by *TransportIdentity*. `HelperIdentity` and `ShardIndex` implement it, to allow data to be sent to a specific helper and specific shard. `Role` implements it too, to make it convenient for `Gateway` to open channels indexed by a role.

```mermaid
---
title: Channels
---

classDiagram
    direction LR

    class ChannelId~I: TransportIdentity~ {
         +entity: I,
         +gate: Gate
    }

    class TransportIdentity
    <<interface>>TransportIdentity
    Role ..|> TransportIdentity
    HelperIdentity ..|> TransportIdentity
    ShardIndex ..|> TransportIdentity
  
    class ShardIndex {
        int index
    } 
    class Role {
          <<enumeration>>
          H1,
          H2,
          H3
    }
    class HelperIdentity {
         int id
    }
```

### Transport

`Transport` trait becomes generic over *TransportIdentity* but its interface stays the same

```mermaid
---
title: Transport changes
---

classDiagram
    class Transport ~I: TransportIdentity~ {
       <<interface>>
       +send(dest: I, route: Route, data: Stream)
       +Stream receive(from: I, route: Route)
    }
[...]

PR #970, opened 2024-03-09
First building blocks for Sharding API: ShardIndex and TransportIdentity

akoshelev:
See #969 for overall approach.

This change introduces two fundamental concepts for horizontal scaling - `ShardIndex` and `Transportidentity`. 

The former will be used the same way as `HelperIdentity` is used nowadays - unique identifier of a peer we are communicating with. Having in mind 200B events, we may need more than 65k shards, thus 4 byte identifier usage. 

`TransportIdentity` is a generalization for peer identifiers used at transport layer.  Two fundamental structs implement it: `HelperIdentity` and `ShardIndex` to allow communication via MPC and shard channels.

`Role` is used at the Gateway/Context layer for communication purposes and `RoleResolvingTransport` is a convenient abstraction, so it also implements `TransportIdentity`.

`Transport` trait became generic over `TransportIdentity` - I tested it in my private branch where I have a complete implementation, so this approach works. 

I also did a bit of a refactoring and made `RoleResolvingTransport` to implement the `Transport` trait because it can now.

### DIFF
--- a/ipa-core/src/helpers/gateway/transport.rs
+++ b/ipa-core/src/helpers/gateway/transport.rs
@@ -1,64 +1,94 @@
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use async_trait::async_trait;
+use futures::Stream;
+
 use crate::{
     helpers::{
-        buffers::UnorderedReceiver,
-        gateway::{receive::UR, send::GatewaySendStream},
-        ChannelId, GatewayConfig, Role, RoleAssignment, RouteId, Transport, TransportImpl,
+        HelperIdentity, NoResourceIdentifier, QueryIdBinding, Role, RoleAssignment, RouteId,
+        RouteParams, StepBinding, Transport, TransportImpl,
     },
-    protocol::QueryId,
+    protocol::{step::Gate, QueryId},
 };
+
+#[derive(Debug, thiserror::Error)]
+#[error("Failed to send to {0:?}: {1:?}")]
+pub struct SendToRoleError(Role, <TransportImpl as Transport<HelperIdentity>>::Error);
+
+/// This struct exists to hide the generic type used to index streams internally.
+#[pin_project::pin_project]
+pub struct RoleRecordsStream(#[pin] <TransportImpl as Transport<HelperIdentity>>::RecordsStream);
 
 /// Transport adapter that resolves [`Role`] -> [`HelperIdentity`] mapping. As gateways created
 /// per query, it is not ambiguous.
 ///
 /// [`HelperIdentity`]: crate::helpers::HelperIdentity
 #[derive(Clone)]
-pub(super) struct RoleResolvingTransport {
-    pub query_id: QueryId,
-    pub roles: RoleAssignment,
-    pub config: GatewayConfig,
-    pub inner: TransportImpl,
+pub struct RoleResolvingTransport {
+    pub(super) roles: RoleAssignment,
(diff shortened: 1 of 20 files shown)


ipa-core/src/helpers/gateway/transport.rs
```diff
@@ -1,64 +1,94 @@
...
+pub struct RoleResolvingTransport {
+    pub(super) roles: RoleAssignment,
+    pub(super) inner: TransportImpl,
+}
+
+impl Stream for RoleRecordsStream {
+    type Item = Vec<u8>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.project().0.poll_next(cx)
```
benjaminsavage, line 39: What is `project()` and where is it coming from? Is it defined in `Stream`?

ipa-core/src/helpers/mod.rs
```diff
@@ -405,23 +406,25 @@ impl TryFrom<[Role; 3]> for RoleAssignment {
 /// Combination of helper role and step that uniquely identifies a single channel of communication
 /// between two helpers.
 #[derive(Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub struct ChannelId {
-    pub role: Role,
+pub struct ChannelId<I: transport::Identity> {
+    pub peer: I,
```
benjaminsavage, line 410: This name: `peer` confuses me. Why did you change this from `role`? What does it mean here? And can you add a comment?

ipa-core/src/sharding.rs
```diff
@@ -0,0 +1,116 @@
...
+pub trait ShardConfiguration {
+    /// Returns the index of the current shard.
+    fn shard_index(&self) -> ShardIndex;
+
+    /// Total number of shards present on this helper. It is expected that all helpers have the
+    /// same number of shards.
+    fn shard_count(&self) -> ShardIndex;
+
+    /// Returns an iterator that yields shard indices for all shards present in the system, except
+    /// this once. Shards are yielded in ascending order.
```
benjaminsavage, line 24: Do you mean "except this one?"

benjaminsavage (approved):
This looks pretty straightforward to me. Most of the lines are just replacing ChanelId with HelperChanelId.

ipa-core/src/helpers/mod.rs
akoshelev: the reason I had to rename it is that this struct is now used to communicate between shards too and there is no notion of helper role for such communications. `ChannelId<ShardIndex>` would have `role: ShardIndex` field which looks even more confusing, hence a neutral `peer` that can be anything that wants to communicate with this instance

ipa-core/src/sharding.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// if current shard index is greater or equal to the total number of shards.
+    fn peer_shards(&self) -> impl Iterator<Item = ShardIndex> {
+        let this = self.shard_index();
+        let max = self.shard_count();
+        assert!(
+            this < max,
+            "Current shard index '{this}' >= '{max}' (total number of shards)"
+        );
+
+        max.iter().filter(move |&v| v != this)
```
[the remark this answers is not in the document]
akoshelev: it moves `this` to the closure, so Rust compiler can see that `filter` closure cannot outlive it. We return an iterator and `filter` closure is part of it, so Rust wants to know that this closure does not borrow from local context

PR #973, opened 2024-03-13
Bit matrix transposes for converting to/from vectorized shares

andyleiserson:


ipa-core/src/secret_sharing/vector/transpose.rs
```diff
@@ -0,0 +1,800 @@
...
+        dst.transpose_from(src);
+        dst
+    }
+}
+
+/// 8x8 bit matrix transpose.
+//
+// From Hacker's Delight (2nd edition), Figure 7-6.
+#[cfg(any(all(test, unit_test), feature = "enable-benches"))]
+#[allow(clippy::trivially_copy_pass_by_ref)] // Keeps interface consistent with other functions.
```
akoshelev, line 70: I usually just use `B: Borrow<[u8; 8]>` to silence clippy in such cases

ipa-core/src/secret_sharing/vector/transpose.rs
```diff
@@ -0,0 +1,800 @@
...
+        replicated::{semi_honest::AdditiveShare, ReplicatedSecretSharing},
+        BitDecomposed, SharedValue, StdArray,
+    },
+};
+
+/// Trait for overwriting a value with the transpose of a source value.
+pub trait TransposeFrom<T> {
+    fn transpose_from(&mut self, src: T);
+
+    fn transposed_from(src: T) -> Self
```
akoshelev, line 56: you don't seem to use it anywhere, are the plans to use it outside of this module?

ipa-core/src/secret_sharing/vector/transpose.rs
```diff
@@ -0,0 +1,800 @@
...
+        #[cfg(all(test, unit_test))]
+        #[test]
+        fn $test_fn() {
+            tests::test_transpose_ba_to_ba::<$dst_row, $src_row, $src_rows, $src_cols>();
+        }
+
+        impl TransposeFrom<&BitDecomposed<$src_row>> for Vec<$dst_row> {
+            fn transpose_from(&mut self, src: &BitDecomposed<$src_row>) {
+                self.resize($src_cols, <$dst_row>::ZERO);
+                let src = <&[$src_row; $src_rows]>::try_from(&**src).unwrap();
```
akoshelev, line 217: these `unwrap` calls - can they ever fail? if so, should we make `transpose_from` fallible or use `expect` if invariants should always hold?

ipa-core/src/secret_sharing/vector/transpose.rs
```diff
@@ -0,0 +1,800 @@
...
+/// `[AdditiveShare<Boolean, N>; <M>]` into a NxM bit matrix represented as `[AdditiveShare<BA<M>>; N]`.
+///
+/// For MxN = 256x64, the invocation looks like `impl_transpose_bool_to_ba!(BA256, 256, 64)`.
+macro_rules! impl_transpose_shares_bool_to_ba {
+    ($dst_row:ty, $src_rows:expr, $src_cols:expr, $test_fn:ident) => {
+        impl TransposeFrom<&[AdditiveShare<Boolean, $src_cols>; $src_rows]>
+            for [AdditiveShare<$dst_row>; $src_cols]
+        {
+            fn transpose_from(&mut self, src: &[AdditiveShare<Boolean, $src_cols>; $src_rows]) {
+                // Transpose left share
```
akoshelev, line 248: there might be some opportunities for reducing code duplication here, although I am not sure if it is going to make the code look more complex or no

akoshelev (approved):
mostly lgtm

ipa-core/src/secret_sharing/vector/transpose.rs
andyleiserson: Yes, it is used in the vectorized share conversion protocol.

(I actually think it's pretty rare that we want to transpose on top of existing data vs. allocate new storage containing transposed data -- but a bunch of our types are not `Default`, which makes it hard to implement the `transposed_from` API in safe rust.)

ipa-core/src/secret_sharing/vector/transpose.rs
andyleiserson: Some of them can fail. I am adding an `Error` type to the transpose trait to handle those cases.

The unwraps that are left should only be for infallible `array::try_from`. It might be nice to annotate these cases (and similar conversions like `<usize as TryFrom<u32>>`) somehow, but it feels like something more structured than `expect("string")` is appropriate. Maybe more unwrap traits like `UnwrapInfallible`? (These would not be quite the same though, since `UnwrapInfallible` can truly never fail.)

ipa-core/src/secret_sharing/vector/transpose.rs
andyleiserson: I agree, it is repetitive. I'm not sure the best strategy to reduce duplication. Structuring the closures to read a single element rather than a block would make it less repetitive, but the generated code was substantially worse when I experimented with that.

There might be something with macros if the hygiene rules don't get in the way.

I don't think there will be too much need to modify this code (adding new transposes should mostly be possible by adding macro invocations, rather than by editing this code), but if you feel strongly that it's worth trying to make this less repetitive, I can work on it.

andyleiserson:
There is some code that is missing coverage right now. The critical functionality is covered, and the remaining code should be covered after the next few PRs are merged. I will double-check once that has happened to make sure the coverage looks good.

PR #974, opened 2024-03-15
Fix collect_steps script

akoshelev:
PR #963 changed the root step name, now our script is failing.

Would be great to add it to CI, but I have high hopes that #961 is not too far ahead.

Tested by manually running this script


akoshelev:
@eriktaubeneck this is probably the culprit for the issues you're dealing with

eriktaubeneck (approved):
working for me!

PR #982, opened 2024-03-18
TestWorld for sharded environments

akoshelev:
This change introduces the ability to run very simple circuits on multiple shards in parallel. It is not possible to communicate between shards yet, but it is possible to use the same test infrastructure to create multiple shards and use PRSS inside them as well as provide the input for each shard and consume their output.

The next and hopefully final change will bring the ability to communicate across shards.

### DIFF
--- a/ipa-core/src/helpers/transport/in_memory/handlers.rs
+++ b/ipa-core/src/helpers/transport/in_memory/handlers.rs
@@ -1,0 +1,102 @@
+use std::{collections::HashSet, future::Future};
+
+use crate::{
+    helpers::{
+        query::{PrepareQuery, QueryConfig},
+        transport::in_memory::{routing::Addr, transport::Error, InMemoryTransport},
+        HelperIdentity, RouteId, Transport, TransportCallbacks, TransportIdentity,
+    },
+    protocol::QueryId,
+    sharding::ShardIndex,
+};
+
+///
+pub trait RequestHandler<I: TransportIdentity>: Send {
+    fn handle(
+        &mut self,
+        transport: InMemoryTransport<I>,
+        addr: Addr<I>,
+    ) -> impl Future<Output = Result<(), Error<I>>> + Send;
+}
+
+/// Helper trait to bind in-memory request handlers to transport identity.
+pub trait IdentityHandlerExt: TransportIdentity {
+    type Handler: RequestHandler<Self>;
+}
+
+impl IdentityHandlerExt for HelperIdentity {
+    type Handler = HelperRequestHandler;
+}
+
+impl IdentityHandlerExt for ShardIndex {
+    type Handler = ();
+}
+
+impl RequestHandler<ShardIndex> for () {
+    async fn handle(
+        &mut self,
+        _transport: InMemoryTransport<ShardIndex>,
+        addr: Addr<ShardIndex>,
+    ) -> Result<(), Error<ShardIndex>> {
+        panic!(
+            "Shards can only process {:?} requests, got {:?}",
+            RouteId::Records,
+            addr.route
+        )
+    }
+}
+
+/// Handler that keeps track of running queries and
+/// routes [`RouteId::PrepareQuery`] and [`RouteId::ReceiveQuery`] requests to the stored
+/// callback instance. This handler works for MPC networks, for sharding network see
+/// [`RequestHandler<ShardIndex>`]
+pub struct HelperRequestHandler {
+    active_queries: HashSet<QueryId>,
+    callbacks: TransportCallbacks<InMemoryTransport<HelperIdentity>>,
+}
+
+impl From<TransportCallbacks<InMemoryTransport<HelperIdentity>>> for HelperRequestHandler {
+    fn from(callbacks: TransportCallbacks<InMemoryTransport<HelperIdentity>>) -> Self {
+        Self {
+            active_queries: HashSet::default(),
+            callbacks,
+        }
+    }
+}
+
+impl RequestHandler<HelperIdentity> for HelperRequestHandler {
+    async fn handle(
+        &mut self,
+        transport: InMemoryTransport<HelperIdentity>,
+        addr: Addr<HelperIdentity>,
+    ) -> Result<(), Error<HelperIdentity>> {
+        let dest = transport.identity();
+        match addr.route {
+            RouteId::ReceiveQuery => {
+                let qc = addr.into::<QueryConfig>();
+                (self.callbacks.receive_query)(Transport::clone_ref(&transport), qc)
+                    .await
+                    .map(|query_id| {
+                        assert!(
+                            self.active_queries.insert(query_id),
+                            "the same query id {query_id:?} is generated twice"
+                        );
+                    })
+                    .map_err(|e| Error::Rejected {
+                        dest,
+                        inner: Box::new(e),
+                    })
+            }
+            RouteId::PrepareQuery => {
+                let input = addr.into::<PrepareQuery>();
+                (self.callbacks.prepare_query)(Transport::clone_ref(&transport), input)
+                    .await
+                    .map_err(|e| Error::Rejected {
+                        dest,
+                        inner: Box::new(e),
+                    })
+            }
+            RouteId::Records => unreachable!(),
+        }
+    }
+}
--- a/ipa-core/src/helpers/transport/in_memory/handlers.rs
+++ b/ipa-core/src/helpers/transport/in_memory/handlers.rs
@@ -10,7 +10,10 @@ use crate::{
     sharding::ShardIndex,
 };
 
+/// Trait for in-memory request handlers. MPC handlers need to be able to process query requests,
+/// while shard traffic does not need to and therefore does not make use of it.
 ///
+/// See [`HelperRequestHandler`].
 pub trait RequestHandler<I: TransportIdentity>: Send {
     fn handle(
         &mut self,
--- a/ipa-core/src/helpers/transport/in_memory/handlers.rs
+++ b/ipa-core/src/helpers/transport/in_memory/handlers.rs
@@ -19,19 +19,6 @@ pub trait RequestHandler<I: TransportIdentity>: Send {
     ) -> impl Future<Output = Result<(), Error<I>>> + Send;
 }
 
-/// Helper trait to bind in-memory request handlers to transport identity.
-pub trait IdentityHandlerExt: TransportIdentity {
-    type Handler: RequestHandler<Self>;
-}
-
-impl IdentityHandlerExt for HelperIdentity {
-    type Handler = HelperRequestHandler;
-}
-
-impl IdentityHandlerExt for ShardIndex {
-    type Handler = ();
-}
-
 impl RequestHandler<ShardIndex> for () {
     async fn handle(
         &mut self,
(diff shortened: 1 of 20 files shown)


ipa-core/src/helpers/transport/in_memory/handlers.rs
```diff
@@ -0,0 +1,102 @@
...
+        &mut self,
+        transport: InMemoryTransport<I>,
+        addr: Addr<I>,
+    ) -> impl Future<Output = Result<(), Error<I>>> + Send;
+}
+
+/// Helper trait to bind in-memory request handlers to transport identity.
+pub trait IdentityHandlerExt: TransportIdentity {
+    type Handler: RequestHandler<Self>;
+}
```
andyleiserson, line 25: Do these new traits really need to be different for in-memory vs. network transport?

One strategy to simplify might be to move `TransportCallbacks` up a level. Instead of having five special purpose `Box<dyn>` handlers in `TransportCallbacks`, use `Box<dyn Handler>` (or `Box<dyn TransportMessageHandler>` for a more specific name) as the type of the `callbacks: I::Handler` argument.

The relationship amongst these traits also seems complicated to me. It might be possible to define something like:

```
trait Transport {
   type Identity;
   type Handler;
}

impl Transport for ShardTransport {
   type Identity = ShardIndex;
   type Handler = ...;
}

impl Transport for MpcTransport {
   type Identity = HelperIdentity;
   type Handler = HelperRequestHandler;
}
```

(This doesn't by itself do anything about the duplication between net and memory transports, however.)

ipa-core/src/sharding.rs
```diff
@@ -1,9 +1,28 @@
...
+    fmt::{Debug, Display, Formatter},
+    num::TryFromIntError,
+};
 
 /// A unique zero-based index of the helper shard.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct ShardIndex(u32);
 
+#[derive(Debug, Copy, Clone)]
+pub struct Shard {
```
andyleiserson, line 11: I might call this `Sharded`? I like how `SomeType<Sharded>` reads better than `SomeType<Shard>`. Right now it's purely shard identity information, but `Context<Sharded>` feels like a better home for a get-channel-to-shard method than `Context<Shard>`.

(`NoSharding` could be renamed to `NotSharded` for symmetry, but that's a super nit.)

ipa-core/src/protocol/context/mod.rs
```diff
@@ -157,37 +160,43 @@ pub trait SpecialAccessToUpgradedContext<F: ExtendableField>: UpgradedContext<F>
 /// Context for protocol executions suitable for semi-honest security model, i.e. secure against
 /// honest-but-curious adversary parties.
 #[derive(Clone)]
-pub struct Base<'a> {
+pub struct Base<'a, B: ShardBinding = NoSharding> {
     inner: Inner<'a>,
     gate: Gate,
     total_records: TotalRecords,
+    sharding: B,
```
andyleiserson, line 167: Does this make more sense as part of `Gateway`, to avoid frequent cloning of static information?

andyleiserson (approved):
It might not be necessary to provide type safety for shard index vs. helper identity in the transport layer, since we have the context layer above it to provide safer APIs to protocols.

Do you imagine it will be possible to cast a sharded context into a non sharded context? Because of the potential of shard locality to leak information, I feel like protocols written against a sharded context are more dangerous, and call for more scrutiny, than protocols written against a non-sharded context.

akoshelev:
> It might not be necessary to provide type safety for shard index vs. helper identity in the transport layer, since we have the context layer above it to provide safer APIs to protocols.

I tried to re-use the existing functionality that we implemented for in-memory infrastructure and it seems that the only difference between MPC and shard channels is the identities used to authenticate peers, hence the change at the `Transport` trait. `Gateway` needs to understand that there are two different transport interfaces used and generic type seemed like the best way to provide that, but I am certainly open for suggestions here.

> 
> Do you imagine it will be possible to cast a sharded context into a non sharded context? Because of the potential of shard locality to leak information, I feel like protocols written against a sharded context are more dangerous, and call for more scrutiny, than protocols written against a non-sharded context.

I agree, that's why I spent quite a few hours modifying the `Runner` and parametrizing it with context type. Sharded circuits must be given an instance of `ShardedContext` and that thing cannot be converted to regular semi-honest or malicious context.

ipa-core/src/protocol/context/mod.rs
akoshelev: I was imagining `Gateway` to be the same for sharded vs non-sharded world because that struct is complicated enough with the stall detection wrapper around it. It would just reference both transports and provide means to open shard-to-shard and mpc channels. 

Depending on the context used, shard-to-shard channels will or will not be be accessible by MPC circuits.

I agree that cloning this stuff often may be less ideal - one way to mitigate that is to move this to `Inner` struct - the downside will be that this will have to be done for all context types separately. I am going to add a comment that we may need to do it later (shouldn't be too hard)

martinthomson: I suggested that this also exist on the gateway.

ipa-core/src/protocol/context/mod.rs
```diff
@@ -263,11 +290,36 @@ struct Inner<'a> {
     pub gateway: &'a Gateway,
 }
 
+// #[derive(Clone)]
+// struct WithShardInfo<'a> {
```
martinthomson, line 294: ??

ipa-core/src/helpers/transport/in_memory/handlers.rs
andyleiserson: I was hoping that using dynamic dispatch might make it easier to make pieces of this the same across transport types (helper, and shard, and ideally HTTP as well).

Maybe something like `fn listen` takes `Option<&dyn Handler>` as the callbacks argument? (The HTTP transport doesn't have `fn listen`, but it does use `TransportCallbacks`.) The shard transport can omit the handler entirely (because it only needs to handle records, at least for now), and for all transport types they no longer need to enumerate the management request types ("prepare query", "receive query", etc.), which feels like something that transports don't need to know about, and that shouldn't need to behave differently across the different transport types.

andyleiserson:
> I agree, that's why I spent quite a few hours modifying the `Runner` and parametrizing it with context type. Sharded circuits must be given an instance of `ShardedContext` and that thing cannot be converted to regular semi-honest or malicious context.

I was thinking of something like the `malicious::Upgraded::as_base` method. If the shuffle happens within `oprf_ipa`, then `oprf_ipa` is going to need to take a sharded context. But only the shuffle (and maybe the aggregation) actually needs the sharded context -- everything else that is called withing `oprf_ipa` can use a non-sharded context.

ipa-core/src/test_fixture/world.rs
```diff
@@ -99,43 +156,29 @@ impl Default for TestWorld {
...
+    pub fn with_shards<B: Borrow<TestWorldConfig>>(config: B) -> Self {
+        Self::with_config(config.borrow())
+    }
+
+    fn shards(&self) -> [&ShardWorld<Shard>; SHARDS] {
+        self.shards
+            .iter()
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|_| unreachable!())
```
martinthomson, line 170: I believe that just `unwrap()` will achieve nearly the same thing.  Is this due to a lack of a Debug impl?

ipa-core/src/test_fixture/world.rs
```diff
@@ -217,15 +219,108 @@ impl Drop for TestWorld {
     }
 }
 
+impl<S: ShardingScheme> TestWorld<S> {
```
martinthomson, line 222: Consistent W or S?

ipa-core/src/test_fixture/world.rs
akoshelev: yep, exactly

akoshelev:
> I was thinking of something like the malicious::Upgraded::as_base method. If the shuffle happens within oprf_ipa, then oprf_ipa is going to need to take a sharded context. But only the shuffle (and maybe the aggregation) actually needs the sharded context -- everything else that is called withing oprf_ipa can use a non-sharded context.

I am not sure if we want to make that distinction. ShardedContext should be a fully-functional semi-honest/malicious context that allows to run MPC with the same shards on other helpers.  I am not sure I see a reason why we want to hide the fact that it is sharded from attribution and intra-shard aggregation steps (note that OPRF resharding also needs sharded context)

It feels a bit error prone to change the context types during the execution at least for now when we have the same shard instance running the whole OPRF IPA

ipa-core/src/protocol/context/mod.rs
akoshelev: I see your point now. Yea, I think I can just return it from shard transport and when creating a gateway for non-sharded execution provide a panicking shard transport to get a runtime error. Maybe parametrizing Gateway can turn this into a compile error, but for some reason I seem to prefer not to do it

ipa-core/src/sharding.rs
akoshelev: I don't mind renaming it, We already have `Sharded` struct inside test infrastructure, but it may not be a big deal to have two

akoshelev: Too annoying to have two, renamed `Sharded` struct defined inside `TestWorld` to `WithShards`

ipa-core/src/protocol/context/mod.rs
akoshelev: I started changing Gateway in the next PR, in this change I won't be able to get it, so I'll follow up on it

ipa-core/src/helpers/transport/in_memory/handlers.rs
akoshelev: I removed `IdentityHandlerExt` in favour of `ListenerSetup` - makes it a little less confusing imo.
I am not sure I fully understand your proposal, will sync up with you offline

ipa-core/src/test_fixture/world.rs
andyleiserson: I've been using `.ok().unwrap()` in this situation. (Not worth an edit just to change that, more of a side note.)

ipa-core/src/helpers/transport/in_memory/handlers.rs
andyleiserson: I took a look at `ListenerSetup` and I like it 👍.

akoshelev: We discussed it offline and agreed that `TransportCallbacks` needs to go away. A better interface for handlers would be similar to [this one](https://docs.rs/iron/latest/iron/middleware/trait.Handler.html). 

It is still worth to encapsulate the delivery method inside `Transport` interface, so messages delivered via HTTP or potentially CF workers channel have the same structure when handled by application logic. We may have a good abstraction for it already - `Addr` struct used inside in-memory network. 

So our handler interface may look like this

```rust
trait Handler<I: TransportIdentity> {
   fn handle(&self, _: &Addr<I>) -> Result<Response>
}
```

HTTP transport can parse HTTP requests into `Addr` struct and application will have a unified logic to process requests based on `Addr:RouteId` field

ipa-core/src/test_fixture/world.rs
akoshelev: `ok().unwrap()` is strictly better, worth changing it

ipa-core/src/helpers/transport/in_memory/handlers.rs
akoshelev: This issue is actually blocking now because `prepare_inputs` handler will require all transports to properly initialize the gateway. Naive approach would make MPC transport (that handles callbacks) to depend on Shard transport implementation (just to pass it to Gateway) which would make things way more complicated.

I am thinking that I would implement something like this

```mermaid
flowchart LR
  HelperApp --> Transports
  HelperApp --> QueryProcessor
  HelperApp --> Handler
  Handler --> Transports
  Handler --> QueryProcessor
```

Issue #987, opened 2024-03-20
Refactor `TransportCallbacks` interface


akoshelev:
We discussed it offline and agreed that `TransportCallbacks` needs to go away. A better interface for handlers would be similar to [this one](https://docs.rs/iron/latest/iron/middleware/trait.Handler.html). 

It is still worth to encapsulate the delivery method inside `Transport` interface, so messages delivered via HTTP or potentially CF workers channel have the same structure when handled by application logic. We may have a good abstraction for it already - `Addr` struct used inside in-memory network. 

So our handler interface may look like this

```rust
trait Handler<I: TransportIdentity> {
   fn handle(&self, _: &Addr<I>) -> Result<Response>
}
```

HTTP transport can parse HTTP requests into `Addr` struct and application will have a unified logic to process requests based on `Addr:RouteId` field

_Originally posted by @akoshelev in https://github.com/private-attribution/ipa/pull/982#discussion_r1532586051_

PR #992, opened 2024-03-25
Delete `TransportCallbacks` and use `RequestHandler` trait instead

akoshelev:
See #987 for motivation. I had to decide whether I want to use dynamic dispatch vs clunky HTTP interfaces with another generic parameter propagated through the entire stack. I don't have a conslusive answer which way is better, both have significant downsides.

Problems with DD approach that is proposed in this change:

* Hard to keep `RequestHandler` trait object safe. No generics for `handle` method, use of `async_trait` etc. That removes the opportunity for some optimizations, namely using a trait to pass data down to the handler. It could be better if HTTP layer just passes the same structs it gets from HTTP layer without an extra conversion that must occur if dynamic dispatch is used.
* Non zero-cost abstraction. To get data back from the handler, we have to use the same format, right now it is JSON but I doubt we can do better than binary serialization, which means more work to get the data out.
* `Box<dyn Trait<....` is everywhere now.

Problems with static dispatch (I will link a commit) is more code that requires a change. It is also not clear whether we can make it a zero-cost abstraction.

It is mentioned in #987 but I will reiterate it here that the reason for the intermediate layer data representation (betweeen HTTP and transport) is to support various delivery channels for IPA, that could potentially include something like CF workers. We don't seem to have an opportunity to rely on our network layer being HTTP in the long term.


akoshelev:
https://github.com/private-attribution/ipa/commit/e05f98eb657b87ef74dde79c592a4c69a4baa2fd - version of `RequestHandler` that uses static dispatch instead.

ipa-core/src/helpers/transport/handler.rs
```diff
@@ -0,0 +1,201 @@
...
+
+/// Represents some response sent from MPC helper acting on a given request. It is rudimental now
+/// because we sent everything as HTTP body, but it could evolve.
+///
+/// ## Performance
+/// This implementation is far from being optimal. Between HTTP and transport layer, there exists
+/// one round of serialization and deserialization to properly represent the types. It is not critical
+/// to address, because MPC helpers have to handle a constant number of requests per query. Note
+/// that all requests tagged with [`crate::helpers::transport::RouteId::Records`] are not routed
+/// through [`RequestHandler`], so there is no penalty.
```
andyleiserson, line 31: I feel like this situation might improve somewhat if we stopped using axum. At somewhat increased risk of returning an invalid response, we can just give the transport layer a `Vec<u8>` and a content type, and have it return those without considering whether the bytes are valid for the claimed content type.

ipa-core/src/helpers/transport/in_memory/transport.rs
```diff
@@ -85,15 +85,14 @@ impl<I: TransportIdentity> InMemoryTransport<I> {
     /// out and processes it, the same way as query processor does. That will allow all tasks to be
     /// created in one place (driver). It does not affect the [`Transport`] interface,
     /// so I'll leave it as is for now.
-    fn listen<L: ListenerSetup<Identity = I>>(
+    fn listen(
         self: &Arc<Self>,
-        mut callbacks: L::Handler,
+        handler: Option<Box<dyn RequestHandler<Identity = I>>>,
```
andyleiserson, line 90: Is there a reason this should be optional in the dynamic case, but required in the static case? (Can we not substitute the panicking handler in the cases where this would otherwise be `None`?)

akoshelev: I probably need to be consistent and either use `PanickingHandler` or `Option`. I decided that latter is better suited to indicate that there is no handler

ipa-core/src/helpers/transport/handler.rs
akoshelev: I would also like to have type safety. One of the challenges I faced before was lack of clear contract between API client and server. Here the situation is aggravated by having extra layers that don't support strong type safety. This is somewhat mitigated by e2e tests but it is easy to miss, if API is never hit.

akoshelev:
Sanitize failure was legit - so glad that we have these. I introduced a leak by having two Arc pointers pointing to each other. Resolving it was not trivial and introduced a bunch of new structs that help manage owning and non-owning ends.

ipa-core/src/net/server/handlers/query/create.rs
```diff
@@ -2,7 +2,7 @@ use axum::{routing::post, Extension, Json, Router};
 use hyper::StatusCode;
 
 use crate::{
-    helpers::Transport,
+    helpers::{ApiError::NewQuery, BodyStream, Transport},
```
andyleiserson, line 5: I usually prefer not to import enum variants directly.

ipa-core/src/query/executor.rs
```diff
@@ -36,17 +36,17 @@ use crate::{
 };
 
 pub trait Result: Send + Debug {
-    fn into_bytes(self: Box<Self>) -> Vec<u8>;
+    fn as_bytes(&self) -> Vec<u8>;
```
andyleiserson, line 39: Maybe `to_bytes`, to indicate that it's not a cheap ref-ref conversion?

ipa-core/src/net/transport.rs
```diff
@@ -43,68 +64,84 @@ impl HttpTransport {
...
+    ) -> Result<HelperResponse, ApiError>
+    where
+        Option<QueryId>: From<Q>,
+    {
         /// Cleans up the `records_stream` collection after drop to ensure this transport
         /// can process the next query even in case of a panic.
-        struct ClearOnDrop {
+        ///
+        /// This implementation is a poor man's safety net and only works because we run
+        /// one query at a time and don't use query identifiers.
```
andyleiserson, line 106: Is there an issue to improve this? Besides the issues you note, it seems like it might belong in the app or query processor rather than in (a particular implementation of) the transport layer.

ipa-core/src/helpers/transport/handler.rs
andyleiserson: Definitely we should have type safety (in the sense of checking requests and responses against a well-defined schema) somewhere on the datapath. But also, somewhere along the path the data has to be in a raw form. What I was trying to suggest is that we think about where the transitions are and which communication layers are dealing with raw form vs. type-safe parsed form.

ipa-core/src/helpers/transport/in_memory/transport.rs
andyleiserson: I agree that `Option` probably makes more sense that `PanickingHandler`, although if there's just a few test cases that want to omit a handler or something like that, then maybe better to use a stub in those cases than make everybody else deal with an `Option`.

In any case, I don't have a strong opinion either way, I was just curious why it was different than the static case.

andyleiserson (approved):
Looks good to me. I'm okay with either the static or dynamic version -- I think the cost of dynamic dispatch here is negligible, but the type parameters in the static version also seem within reason. Either one streamlines the transport implementations.

ipa-core/src/helpers/transport/handler.rs
akoshelev: let me create an issue for that

ipa-core/src/net/transport.rs
akoshelev: I am not sure where I want to put this - it feels like transport internal state must be self-managed. We currently share the same transport instance across all queries and therefore having a need to clean up the internal state. As anything can panic at any moment, this state management is repetitive and prone to errors.

Maybe a better model would create a transport per query and then any panic will cause thread to abort, gateway and request handler to be dropped and transport destroyed.

PR #993, opened 2024-03-27
Hashing functionality needed to support Fiat-Shamir

benjaminsavage:
In our zero-knowledge proof generation, we will employ "Fiat-Shamir", which basically entails having the prover generate the random challenge point by hashing the proof, rather than waiting for the verifier to generate a totally random point.

In principle, this is just as secure, since the prover cannot select the random challenge point specifically to cheat. This does make a few assumptions, like the assumption that the hash function is basically a really good one. 

The idea is that the verifier can generate the same hash of the proof, and thus generate the same challenge point.

The overall benefit of this approach is that it turns a multi-round protocol into a single round protocol, since the prover can iterate through all the steps, computing random challenge points along the way (since they are just derived from hashes of each proof), and then just sends a big batch of proofs to the verifiers.


ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use generic_array::GenericArray;
+use sha2::{Digest, Sha256};
+use typenum::U32;
+
+use crate::{
+    ff::{Field, Serializable}, helpers::Message, protocol::prss::FromRandomU128
+};
+
+#[derive(Clone, Copy, Debug, PartialEq)]
```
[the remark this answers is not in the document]
benjaminsavage: I just copied and pasted from @danielmasny 's draft PR =). I will just remove them all for now, and only add these back in when we need them.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Result<Self, Self::DeserializationError> {
+        Ok(Hash(*buf))
+    }
+}
+
+impl Message for Hash {}
+
+/// This function allows to compute a hash of a slice of shared values.
+/// The output is a single `Hash` struct that can be sent over the network channel to other helper parties.
+pub fn compute_hash<'a, I, S>(input: I) -> Hash
```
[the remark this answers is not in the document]
benjaminsavage: Great point. I'll add that.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,142 @@
...
+    for x in input {
+        let mut buf = GenericArray::default();
+        x.serialize(&mut buf);
+        sha.update(buf);
+    }
+    // compute hash
+    Hash(*GenericArray::<u8, U32>::from_slice(&sha.finalize()[0..32]))
+}
+
+/// This function allows to hash a vector of field elements into a single field element
```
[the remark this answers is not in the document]
benjaminsavage: Haha, yeah, you're right. Copy-pasted from @danielmasny 's draft and I didn't update the comment. Fixing.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,142 @@
...
+    // set up hash
+    let mut sha = Sha256::new();
+    // set state
+    for x in input {
+        let mut buf = GenericArray::default();
+        x.serialize(&mut buf);
+        sha.update(buf);
+    }
+    // compute hash
+    Hash(*GenericArray::<u8, U32>::from_slice(&sha.finalize()[0..32]))
```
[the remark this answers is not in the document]
benjaminsavage: Sounds good to me =). I'll use this.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,142 @@
...
+    // however, deserialize might fail for some fields so we use `from_random_128` instead
+    // this results in at most 128 bits of security/collision probability rather than 256 bits as offered by `Sha256`
+    // for field elements of size less than 129 bits, this does not make a difference
+    F::from_random_u128(u128::from_le_bytes(
+        sha.finalize()[0..16].try_into().unwrap(),
+    ))
+}
+
+#[cfg(all(test, unit_test))]
+mod test {
```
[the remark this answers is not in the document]
benjaminsavage: Sure thing. Will do.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,171 @@
...
+    type Size = <Sha256 as OutputSizeUser>::OutputSize;
+
+    type DeserializationError = Infallible;
+
+    fn serialize(&self, buf: &mut GenericArray<u8, Self::Size>) {
+        buf.copy_from_slice(&self.0);
+    }
+
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Result<Self, Self::DeserializationError> {
+        Ok(Hash(*Output::<Sha256>::from_slice(buf)))
```
akoshelev, line 27: for the sake of completeness, we should cover the deserialization phase with a unit test too. Maybe just generate a random hash and run it through serde

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -55,60 +56,60 @@ where
...
+        I: IntoIterator<Item = F>,
+        J: IntoIterator<Item = F>,
+        I::IntoIter: ExactSizeIterator,
+        J::IntoIter: ExactSizeIterator,
     {
-        debug_assert_eq!(self.u.len() % λ::USIZE, 0); // We should pad with zeroes eventually
+        let mut u = u.into_iter();
+        let mut v = v.into_iter();
+
+        debug_assert_eq!(u.len() % λ::USIZE, 0); // We should pad with zeroes eventually
```
danielmasny, line 76: adding the support for non powers of lambda is actually pretty important, but also easy to do

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -55,60 +56,60 @@ where
...
         );
 
-        let mut next_proof_generator = ProofGenerator {
-            u: Vec::<F>::with_capacity(s),
-            v: Vec::<F>::with_capacity(s),
-        };
-
         let denominator = CanonicalLagrangeDenominator::<F, λ>::new();
-        let lagrange_table_r = LagrangeTable::<F, λ, U1>::new(&denominator, &r);
         let lagrange_table = LagrangeTable::<F, λ, <λ as Sub<U1>>::Output>::from(denominator);
```
danielmasny, line 86: the denominator and table only depend on lambda, so you don't need to recompute them for every proof part, you could add a reference to lagrange table to the input

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -55,60 +56,60 @@ where
         Self { u, v }
     }
 
-    pub fn compute_proof<λ: ArrayLength>(
-        &self,
-        r: F,
-    ) -> (ZeroKnowledgeProof<F, TwoNMinusOne<λ>>, ProofGenerator<F>)
+    pub fn compute_proof<λ: ArrayLength, I, J>(
```
danielmasny, line 59: this computes only one share of the proof right? the other part is generated using prss? Maybe rename proof to share of proof?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -55,60 +56,60 @@ where
...
+        }
+        ZeroKnowledgeProof::new(proof)
     }
 
-    pub fn compute_final_proof<λ: ArrayLength>(
-        &self,
+    pub fn compute_final_proof<λ: ArrayLength, I, J>(
+        u: I,
+        v: J,
         p_0: F,
```
danielmasny, line 113: are these the masks?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -117,27 +118,92 @@ where
...
-        assert_eq!(self.u.len(), λ::USIZE); // We should pad with zeroes eventually
+        let mut u = u.into_iter();
+        let mut v = v.into_iter();
+
+        assert_eq!(u.len(), λ::USIZE); // We should pad with zeroes eventually
+        assert_eq!(v.len(), λ::USIZE); // We should pad with zeroes eventually
 
         // We need a table of size `λ + 1` since we add a random point at x=0
         let denominator = CanonicalLagrangeDenominator::<F, Sum<λ, U1>>::new();
         let lagrange_table = LagrangeTable::<F, Sum<λ, U1>, λ>::from(denominator);
```
danielmasny, line 134: you could use compute proof here as well to have less redundant code. You can even use the same lagrange table by only submitting lambda minus 1 real u, v value to leave one entry space for the masks. Then for the final proof you just need to slightly adapt he proof generator/ u, v by appending the masks when calling generate proof the last time

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -117,27 +118,92 @@ where
...
-            .map(|(a, b)| a * b),
-        )
+        for (i, (x, y)) in zip(p_extrapolated.into_iter(), q_extrapolated.into_iter()).enumerate() {
+            proof[λ::USIZE + 1 + i] += x * y;
+        }
+
+        ZeroKnowledgeProof::new(proof)
+    }
+
+    pub fn gen_challenge_and_recurse<λ: ArrayLength, I, J>(
```
danielmasny, line 160: I think this is the only function that actually uses the hashing right? It includes computing the challenge and compress the u and v values right?

danielmasny (commented):
Looks good to me. The hashing looks correct to me and is as we have discussed it. There are some other parts that are not directly related to the hashing. I left some comments anyway. Since I haven't reviewed your previous PRs, I am lacking some contexts. I asked some clarification questions.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: You're right. I was planning to do that but forgot. Thanks for the reminder! I'll add it.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: It computes the actual polynomial G(X), which is what I am calling the "proof". This is then secret shared (using PRSS). But I consider those "shares" of the "proof".

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: Yes. The random weights used for p(0) and q(0) come from PRSS and I'm trying to NOT have any dependency on that module from this code. As such, I'm assuming the caller will invoke PRSS to obtain random field elements and then send them in.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,171 @@
...
+
+pub fn compute_hash<'a, I, S>(input: I) -> Hash
+where
+    I: IntoIterator<Item = &'a S>,
+    S: Serializable + 'a,
+{
+    // set up hash
+    let mut sha = Sha256::new();
+    let mut buf = GenericArray::default();
+    let mut is_empty = true;
```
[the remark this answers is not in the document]
akoshelev: we can require `I: ExactSizeIterator` and check its `len()` but that would make it hard to work with streams (although we have an `ExactSizeStream` trait and we must know the input size)

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    // set state
+    for x in input {
+        is_empty = false;
+        x.serialize(&mut buf);
+        sha.update(&buf);
+    }
+
+    assert!(!is_empty, "must not provide an empty iterator");
```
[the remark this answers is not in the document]
akoshelev: that's exactly the thing you would want to avoid - running hash on inputs empty by accident and yielding a constant value for validation purposes

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,171 @@
...
+    Digest, Sha256,
+};
+
+use crate::{
+    ff::{Field, Serializable},
+    protocol::prss::FromRandomU128,
+};
+
+#[derive(Debug, PartialEq)]
+pub struct Hash(Output<Sha256>);
```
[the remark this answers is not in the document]
benjaminsavage: That was @akoshelev's suggestion =).

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
benjaminsavage: OK, can do.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        let mut rng = thread_rng();
+
+        let mut left = Vec::with_capacity(LIST_LENGTH);
+        let mut right = Vec::with_capacity(LIST_LENGTH);
+        for _ in 0..LIST_LENGTH {
+            left.push(rng.gen::<Fp32BitPrime>());
+            right.push(rng.gen::<Fp32BitPrime>());
+        }
+        let r1: Fp32BitPrime = hash_to_field(&compute_hash(&left), &compute_hash(&right));
```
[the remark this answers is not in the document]
benjaminsavage: Speaking of collisions... 

I generated this test case in a Google spreadsheet and copied all the values in here, including the random values chosen for "r". Well, now that the code actually hashes the shares of the proof in order to select "r", I had to iterate through many different secret-sharings in order to find one that would result in the exact same value of "r" being selected as I had before. It didn't take that long since I'm using Fp31, but yeah...

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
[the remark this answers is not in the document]
benjaminsavage: Yeah... you should be hashing a polynomial... not nothing. There is never a case when we would want to hash an empty polynomial.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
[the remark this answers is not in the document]
benjaminsavage: I'm fine with checking in this way. Let's not require any more traits.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: Yup. I think I may have already actually done it, now that I call `.unwrap_or(F::ZERO)`, but I check to make sure it's correct in a follow up diff when I remove these debug asserts.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: That's right.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: That's a good point. I'll think about this. Maybe I can just remove this function entirely.

ipa-core/src/protocol/ipa_prf/malicious_security/hashing.rs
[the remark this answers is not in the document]
danielmasny: Yes, we definitely need that there is enough entropy in the input, i.e. we cannot use proofs that are too short. For 128 bits of security we need an input length of at least 128 bit. Since the proof length is 2 lambda - 1, we need that the size of field elements is at least 128/(2 lambda -1) bits.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
benjaminsavage: Oh man... I tried, I really did, but the trait bounds are WILD! I just couldn't figure out how to make the compiler understand that it should work.

Issue #588, opened 2023-04-12
Delete "enable-serde" feature


akoshelev:
We've been using serde a lot and the usage of it only increases, so it may make sense to vend it by default.
Came up from discussions over #584

akoshelev:
These are all the places where we still use it: https://github.com/search?q=repo%3Aprivate-attribution%2Fipa%20enable-serde&type=code

PR #996, opened 2024-03-29
Serde is now a non-optional dependency

cberkhoff:
Removed the `enable-serde` flag and made Serde a non-optional dependency. Removed the `cfg` attributes that depended on the feature since now the library is included by default.

### Testing
Both unit tests and `./pre-commit` pass.

closes private-attribution/ipa#588

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -18,7 +18,6 @@ default = [
     "ipa-prf",
 ]
 cli = ["comfy-table", "clap"]
-enable-serde = ["serde", "serde_json"]
 disable-metrics = []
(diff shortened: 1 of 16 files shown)


akoshelev (approved):
lgtm, thanks! The code coverage issues are probably coming from us not using these structs at all. Would you be able to test which ones are needed by simply deleting them and compiling the code? We may be able to clean up even more things

cberkhoff:
I went over the classes that were flagged as part of the coverage check and they were either:
a) Not used anywhere in the code. I removed those classes in the last commits.
b) Tests where not exercising the classes for which I added serialization/deserialization. This issue is orthogonal to this PR and rather tackle that separately.

akoshelev (approved):
lgtm, will merge it tomorrow

PR #997, opened 2024-03-31
Malicious API

danielmasny:
Set up the context to support DZKPs. Would appreciate your feedback!


ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,467 @@
...
+        // check size of store
+        if self.vec.len() < position_vec + (length >> 8) {
+            // recover from wrong size
+            // increase size of store to twice the size + position of end of the segment to be included
+            // expensive, ideally use initialize with correct length
+            self.extend_vec(2 * self.vec.len() - (position_vec + (length >> 8)));
+        }
+
+        if 256 % length == 0 {
+            // segments are small, pack one or more in each entry of `vec`
```
andyleiserson, line 231: This means that the valid tracking of `Option<UnverifiedValues>` only works for large segments, is that okay?

ipa-core/src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -53,6 +53,12 @@ impl<F: PrimeField> ExtendableField for F {
     }
 }
 
+/// Trait for fields compatible with DZKPs
+/// Field needs to support conversion to `SegmentEntry`, i.e. `to_segment_entry` which is required by DZKPs
+pub trait DZKPCompatibleField<'a>: Field {
+    fn to_segment_entry(&self) -> SegmentEntry<'a>;
+}
```
andyleiserson, line 60: How about:

```
pub trait DZKPIntermediate {
    fn left_segment_entry(&self) -> SegmentEntry<'_>;
    fn right_segment_entry(&self) -> SegmentEntry<'_>;
}
```

(The lifetime doesn't need to be on the trait -- I think using the `'_` anonymous lifetime will work, but if not, then `fn left_segment_entry<'a>(&'a self) -> SegmentEntry<'a>`.)

I removed the prefix entirely since that's consistent with other `left_/right_*` functions on additive shares, but if using a prefix, I'd pick `as_` over `to_` since this is a cheap reference conversion.

ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// TODO (alex): Arc is required here because of the `TestWorld` structure. Real world
+    /// may operate with raw references and be more efficient
+    inner: Arc<DZKPUpgradedInner<'a>>,
+    gate: Gate,
+    total_records: TotalRecords,
+}
+
+impl<'a> DZKPUpgraded<'a> {
+    pub(super) fn new<S: Step + ?Sized>(
+        source: &Base<'a>,
```
benjaminsavage, line 36: What is `Base`?

ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -0,0 +1,151 @@
...
+        malicious_step: &S,
```
benjaminsavage, line 37: What is this one `Step`? Is this just a way to open a channel for communicating the ZKPs?

ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -0,0 +1,151 @@
...
+            inner: DZKPUpgradedInner::new(source, batch),
+            gate: source.gate().narrow(malicious_step),
+            total_records: TotalRecords::Unspecified,
+        }
+    }
+}
+
+#[async_trait]
+impl<'a> DZKPContext for DZKPUpgraded<'a> {
+    fn is_safe(&self) -> Result<(), Error> {
```
benjaminsavage, line 53: Can you add a comment explaining what it means to be "safe"?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,467 @@
...
+    },
+};
+
+/// `UnverifiedValues` are intermediate values that occur during a multiplication.
+/// These values need to be verified since there might have been malicious behavior.
+/// For a multiplication, let `(x1, x2), (x2, x3), (x3,x1)`, `(y1, y2), (y2, y3), (y3,y1)` be the shares of `x` and `y`
+/// and let the goal of the multiplication be to compute `x*y`.
+/// `(pi, p(i+1)` is the randomness from `PRSS` used for the multiplication by helper `i`.
+/// `z(i+1)` is the result of the multiplication received from the helper on the right.
+/// We do not need to store `zi` since it can be computed from the other stored values.
```
benjaminsavage, line 31: This is a tradeoff. Just storing it would be simpler, but require more memory.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,467 @@
...
+            y_right: BitArray::try_from(y_right).unwrap(),
+            prss_left: BitArray::try_from(prss_left).unwrap(),
+            prss_right: BitArray::try_from(prss_right).unwrap(),
+            z_right: BitArray::try_from(z_right).unwrap(),
+        }
+    }
+}
+
+/// `Segment` is a variable size set or subset of `UnverifiedValues` using a `BitSlice`
+/// For efficiency reasons we constrain the segments size in bits to be a divisor of 256
```
benjaminsavage, line 81: What is the efficiency reason? It seems like padding with `0`s is wasteful and not more efficient.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,467 @@
...
+                new_offset = ((usize::from(self.offset) - new_offset + 255) >> 8) << 8;
+            }
+            let extension_length = ((usize::from(self.offset) - new_offset) >> 8) * segment_length;
+            // use existing vec and add enough space in front
+            let vec_previous = &mut self.vec;
+            let mut vec: Vec<Option<UnverifiedValues>> =
+                Vec::with_capacity(vec_previous.len() + extension_length);
+            vec.resize(extension_length, None);
+            vec.append(vec_previous);
+            self.offset = RecordId::from(new_offset);
```
benjaminsavage, line 199: Can you tell me a bit more about when you expect to call this function? Why would we need to support pre-pending elements? Why not just support adding to the end?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,467 @@
...
+        vec.resize(new_length, None);
+    }
+
+    /// `insert_segment` allows to include a new segment in `UnverifiedValuesStore`
+    fn insert_segment(&mut self, record_id: RecordId, segment: Segment) {
+        // check offset
+        if record_id < self.offset {
+            // recover from wrong offset, expensive
+            self.reset_offset(record_id, segment.len());
+        }
```
benjaminsavage, line 218: When would this happen?

benjaminsavage (commented):
I mostly am just asking questions to understand the intended use and design.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: I have added a comment as discussed

ipa-core/src/protocol/context/dzkp_malicious.rs
danielmasny: Context is basically a wrapper around Base, I didn't change anything about Base, DZKPUpgraded basically corresponds to Upgraded (MAC-based API) without many changes

ipa-core/src/protocol/context/dzkp_malicious.rs
danielmasny: There is a separate step/gate in DZKPValidator that you can use for the DZKPs and their verification. In the MAC based approach, this was used for the validation as well.

ipa-core/src/protocol/context/dzkp_malicious.rs
danielmasny: I added it to the trait definition. It just checks that there are currently no values that still need to be verified. So you could e.g. reveal shares or drop the validator

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: yes, I think either way, it is simple since z can be easily computed from the other values when needed. I have not implemented the conversion yet. I think storing nor recomputing it adds significant overhead.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: The issue here is that we convert it into 256 bit arrays to have constant size elements. In the long run, we anyway prefer vectorizations so it would be a multiple of 256 bits. I discussed it with Andy and it made sense to us, to support non-vectorized operations, but not necessarily optimize for it.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: I don't think you need it as long as iterate from a low record_id to a high record_id. If you don't follow that order this function makes sure you don't panic. However, it won't be great for efficiency reasons.

danielmasny: More background: We did not want to use gate,record_id as key to the hashmap since there are a lot of records and this would be very expensive. Instead, we arrange the records in a vector where record_id is the index. The vector will have an offset - which is a record_id - that will be stored at index 0. If the offset is set incorrectly and you try to store a record smaller than offset, you would need to redefine the offset and reallocate the vector.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+    /// `update` updates the current store to the next offset which is the previous offset plus `vec.actual_len()`.
+    /// It deallocates `vec`. `chunk_size` remains the same.
+    ///
+    /// ## Panics
+    /// Panics when `segment_size` is `None`
+    fn update(&mut self) {
+        // compute how many records have been added
+        // since `actual_length` is imprecise (i.e. rounded up) for segments smaller than 256,
+        // subtract `1` to make sure that offset is set sufficiently small
+        self.offset += ((self.actual_len() - 1) << 8) / self.segment_size.unwrap();
```
benjaminsavage, line 225: I do not understand this computation. You take the "actual length", subtract 1, then bitshift it up by 8? Why?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+        self.segment_size = Some(segment_size);
+    }
+
+    /// `allocate_vec` allocates `vec`. The length is determined by the amount
+    /// of records, i.e. `chunk_size` and the size of a record, i.e. `segment_size`.
+    /// ## Panics
+    /// Panics when `segment_size` is `None`.
+    fn allocate_vec(&mut self) {
+        // add 255 to round up
+        self.vec = vec![None; (self.chunk_size * self.segment_size.unwrap() + 255) >> 8];
```
benjaminsavage, line 253: Why do we need to add 255? This isn't really "rounding". It might put you just over some kind of nice round number actually.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: This has to do with the translation between record_id and index in the vector. A record can have any bit length, from 1 bit to 256+ bits (the latter when vectorized). We translate this into pieces of 256 bit. To compute the new offset, i.e. setting the offset for the next proof chunk, I need to compute the last record_id of the previous chunk which I can compute by computing the len(), multiply it by 256 (shift by 8) and divide by the segment length. Since in case of records with length less than 256, I pack several records in the same index. Since the last index might not be completely filled, I need to be conservative and set the offset lower to ensure all record ids of the following proof chunk are larger than the offset.

danielmasny: notice that I use an Vector<Option<>> which is always filled with None unless it has an actual value. Actual length just computes the index of the last some element plus 1.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: For the test, I want to make sure that it runs without panics since efficiency is not the main concern. There are several aspects of the DZKPs that we could optimize (Optimizations that I have ignored so far) if the DZKPs turns out to be a bottleneck.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: In what way is it different from rounding? I add 255 and divide by 256 by shifting by 8.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+        }
+        let position_raw = usize::from(record_id) - usize::from(self.offset);
+        let length = segment.len();
+        let position_vec = (length * position_raw) >> 8;
+        // check size of store
+        if self.vec.len() < position_vec + (length >> 8) {
+            // recover from wrong size
+            // increase size of store to twice the size + position of end of the segment to be included
+            // expensive, ideally use initialize with correct length
+            self.extend_vec(2 * self.vec.len() - (position_vec + (length >> 8)));
```
andyleiserson, line 315: I think it would be better to just do `vec.resize(position_vec + (length >> 8))` unless/until we have evidence that the built-in allocation handling is not doing something reasonable.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+                // distance between new_offset and old offset is rounded up to a multiple of 256
+                new_offset = ((usize::from(self.offset) - new_offset + 255) >> 8) << 8;
+            }
+            let extension_length = ((usize::from(self.offset) - new_offset) >> 8) * segment_length;
+            // use existing vec and add enough space in front
+            let vec_previous = &mut self.vec;
+            let mut vec: Vec<Option<UnverifiedValues>> =
+                Vec::with_capacity(vec_previous.len() + extension_length);
+            vec.resize(extension_length, None);
+            vec.append(vec_previous);
```
andyleiserson, line 273: I think you could instead do something like `self.vec.splice(0..0, repeat(None).take(extension_length))`.

(The current code is missing a `self.vec = vec` at the end -- also in `extend_vec`.)

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+        if self.inner.get(gate).is_none() {
+            let value_store = UnverifiedValuesStore::new(self.chunk_size);
+            self.inner.insert(gate.clone(), value_store);
+        }
+
+        // insert segment
+        self.inner
+            .get_mut(gate)
+            .unwrap()
+            .insert_segment(record_id, segment);
```
andyleiserson, line 411: You can use `self.inner.entry(gate).or_insert_with(/* closure */)` to avoid the `is_none` / `unwrap` combo.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+            inner: HashMap::<Gate, UnverifiedValuesStore>::new(),
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+            || self
+                .inner
+                .iter()
+                .fold(true, |acc, (_, value)| acc && value.is_empty())
```
andyleiserson, line 397:
```suggestion
        self.inner.is_empty() || self.inner.values().all(|s| s.is_empty())
```

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+
+        if 256 % length == 0 {
+            // segments are small, pack one or more in each entry of `vec`
+            let position_bit_array = (length * position_raw) % 256;
+            if self.vec[position_vec].is_none() {
+                let entry = UnverifiedValues::new();
+                self.vec[position_vec] = Some(entry);
+            }
+            // get entry
+            let entry = self.vec[position_vec].as_ref().unwrap();
```
andyleiserson, line 326: You can use `Option::get_or_insert_with` to avoid this `is_none` / `unwrap`.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+
+/// `UnverifiedValues` are intermediate values that occur during a multiplication.
+/// These values need to be verified since there might have been malicious behavior.
+/// For a multiplication, let `(x1, x2), (x2, x3), (x3,x1)`, `(y1, y2), (y2, y3), (y3,y1)` be the shares of `x` and `y`
+/// and let the goal of the multiplication be to compute `x*y`.
+/// `(pi, p(i+1)` is the randomness from `PRSS` used for the multiplication by helper `i`.
+/// `z(i+1)` is the result of the multiplication received from the helper on the right.
+/// We do not need to store `zi` since it can be computed from the other stored values.
+#[derive(Clone, Debug)]
+struct UnverifiedValues {
```
andyleiserson, line 36: I wonder if it would be worth adding `begin: u8` and `end: u8` here so that we can track the validity within these arrays. Or else in some wrapper type that would take the place of `Option<UnverifiedValues>`. The space overhead is the same or negligibly more than the `Option`. We don't even necessarily need to do all the accounting right away -- to start with we could just implement `is_some` and `is_none` methods in terms of `begin != end`. This would at least give us a framework in case we start having problems with inconsistent batches.

danielmasny: This only seems to work when the missing element is at the end or beginning of the array. When I am missing one in between, begin, end would not be helpful. Is there another reason to use begin and end other than checking consistency of the batches?

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: That is a good point. It should only do something when it is not empty. I have changed that.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -161,21 +179,78 @@ impl<'a> SegmentEntry<'a> {
...
+        let mut actual_length = self.vec.len();
+        while actual_length > 0 && self.vec[actual_length - 1].is_none() {
+            actual_length -= 1;
         }
+        actual_length
+    }
+
+    /// `is_empty` returns `true` if at least one segment has been added
+    fn is_empty(&self) -> bool {
+        self.vec.is_empty()
```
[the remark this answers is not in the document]
danielmasny: This is technically correct, but not very efficient. Right now, I only allocate vec to all none when I actually add a segment. I will introduce a new function is_unallocated that does what I want.

andyleiserson:
It would let us at least get an exact size for the batch, which might make
some of the other calculations more accurate.

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: It would let us at least get an exact size for the batch, which might make some of the other calculations more accurate.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -269,6 +357,17 @@ impl UnverifiedValuesStore {
...
     }
+
+    /// `get_unverified_field_value` converts a `UnverifiedValuesStore` into an iterator over `UnverifiedFieldValues`
+    /// compatible with DZKPs
+    fn get_unverified_field_values<DF: DZKPBaseField>(
+        &self,
+    ) -> impl Iterator<Item = DF::UnverifiedFieldValues> + '_ {
+        self.vec
+            .iter()
+            .filter_map(|x| x.as_ref())
```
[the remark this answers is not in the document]
danielmasny: I use flatten for now. I think if we want to check the vectors, it would require more consideration (since there are some None values that I don't want to include but they are correct and there might be some None values which indicate that there is an issue)

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -376,81 +477,169 @@ pub(crate) enum Step {
 
 /// Marker Trait `DZKPBaseField` for fields that can be used as base for DZKP proofs and their verification
 /// This is different from trait `DZKPCompatibleField` which is the base for the MPC protocol
-pub trait DZKPBaseField: Field {}
+pub trait DZKPBaseField: Field {
+    type UnverifiedFieldValues;
```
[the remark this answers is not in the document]
danielmasny: The base field will be something like a 61 bit prime field. The type will like consists of 8 field values. The signature of this function probably will need to change. I will implement this in a separate PR.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -376,81 +477,169 @@ pub(crate) enum Step {
...
+    fn is_safe(&self) -> Result<(), Error>;
+
+    /// `get_chunk_size` returns the chunk size of the validator
+    fn get_chunk_size(&self) -> Option<usize>;
 }
 
-pub struct SemiHonestDZKPValidator<'a, DF> {
-    context: SemiHonestDZKPUpgraded<'a>,
-    _f: PhantomData<DF>,
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
danielmasny: the async trait flag actually allowed me to put the function in the trait. Even though Intellij marks it red it still compiles.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+    fn get_chunk_size(&self) -> Option<usize> {
+        None
+    }
+}
+
+/// `MaliciousDZKPValidator` corresponds to pub struct `Malicious` and implements the trait `DZKPValidator`
+/// The implementation of `validate` of the `DZKPValidator` trait depends on generic `DF`
+#[allow(dead_code)]
+pub struct MaliciousDZKPValidator<'a> {
+    batch_list: Arc<Mutex<Batch>>,
```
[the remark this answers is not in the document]
danielmasny: changed it

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -376,81 +477,169 @@ pub(crate) enum Step {
...
+            iter(x).map(move |x| {
+                if valid {
+                    Ok(x)
+                } else {
+                    Err(Error::DZKPValidationFailed)
+                }
+            })
+        })
+        .flatten()
+        .take_while(|x| future::ready(x.is_ok()))
```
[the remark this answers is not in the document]
danielmasny: I like that your code is very concise. I had issues to propagate the error from a chunk to every single element since error does not implement clone. I also did not want to use an Arc or here. How does try_flatten_iters work? Alex suggested to terminate the stream when we hit an error. I think he was concerned that you could ignore the error and just move on with the next element that is not an error which would not be safe against malicious helper parties.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,772 @@
...
+        let position_raw = usize::from(record_id) - usize::from(self.offset);
+        let length = segment.len();
+        let position_vec = (length * position_raw) >> 8;
+        // check size of store
+        if self.vec.len() < position_vec + (length >> 8) {
+            // recover from wrong size
+            // increase size of store to twice the size + position of end of the segment to be included
+            // expensive, ideally use initialize with correct length
+            self.extend_vec(2 * self.vec.len() - (position_vec + (length >> 8)));
+        }
```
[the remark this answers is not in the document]
danielmasny: I also added the metrics here so we can detect when we call this

ipa-core/src/protocol/context/dzkp_validator.rs
benjaminsavage: Oh I see now. Yes, this works.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+            .iter()
+            .map(|x| x.share_with(&mut rng))
+            .collect();
+        let h1_shares: Vec<Replicated<Fp31>> = shared_inputs.iter().map(|x| x[0].clone()).collect();
+        let h2_shares: Vec<Replicated<Fp31>> = shared_inputs.iter().map(|x| x[1].clone()).collect();
+        let h3_shares: Vec<Replicated<Fp31>> = shared_inputs.iter().map(|x| x[2].clone()).collect();
+
+        let futures = context
+            .into_iter()
+            .zip([h1_shares, h2_shares, h3_shares])
```
benjaminsavage, line 723: Usually our tests use this `world.malicious` approach instead of joining all 3 helpers. Can you use that here?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+            .map(|(ctx, input_shares)| async move {
+                let v = ctx.dzkp_validator(chunk_size);
```
benjaminsavage, line 725: Is there some kind of assert statement later that validates that this DZKP validator ran and ran correctly?

danielmasny: Yes, there is a custom implementation of the drop trait so if the validator is dropped and not every multiplication has been verified, it will cause a panic.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: I agree this would make the test much easier to read. Unfortunately, World.malicious still uses the MAC based approach to do malicious security but I want to use the semi honest version. I haven't integrated the dzkp in the multiply function yet and plan to do this in a separate PR and use it then in this test. In the long run, it might make sense to add a world.dzkp_malicious() function for testing.

ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -0,0 +1,151 @@
...
+    sync::Arc,
+};
+
+/// Represents protocol context in malicious setting, i.e. secure against one active adversary
+/// in 3 party MPC ring.
+#[derive(Clone)]
+pub struct DZKPUpgraded<'a> {
+    /// TODO (alex): Arc is required here because of the `TestWorld` structure. Real world
+    /// may operate with raw references and be more efficient
+    inner: Arc<DZKPUpgradedInner<'a>>,
```
akoshelev, line 29: I am 99% certain that you don't need an Arc here

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+/// location out of order. This is important since we do not now the order in which different records
+/// are added to the `UnverifiedValueStore`.
+/// It is not to keep track whether all relevant records have been added. The verification will fail
+/// when a helper party forgets to include one of the records. We could add a check separate from the proof
+/// such that this does not count as malicious behavior.
+#[derive(Clone, Debug)]
+struct UnverifiedValuesStore {
+    offset: RecordId,
+    chunk_size: usize,
+    segment_size: Option<usize>,
```
akoshelev, line 209: what is the reason for having variable at runtime size here? Is it possible to simplify this implementation by

* Assuming the size is known in advance
* Allocating early
* Having a counter that tracks the rightmost position of a non-empty segment. This one may be important because you have quite a few methods that perform linear scan and for large stores which is what we want the cost will be significant

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+    /// Panics when segments have different lengths across records
+    fn insert_segment(&mut self, record_id: RecordId, segment: Segment) {
+        // initialize `segment_size` when necessary
+        if self.segment_size.is_none() {
+            self.initialize_segment_size(segment.len());
+        }
+        // check segment size
+        debug_assert_eq!(segment.len(), self.segment_size.unwrap());
+        // allocate if vec is not allocated yet
+        if self.is_unallocated() {
```
akoshelev, line 308: if you need to do it in more than one method, it is worth having an abstraction over that.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+            self.allocate_vec();
+        }
+        // check offset
+        if record_id < self.offset {
+            // recover from wrong offset, expensive
```
akoshelev, line 313: this comment is not helpful as we won't see it at runtime

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+            self.reset_offset(record_id, segment.len());
+        }
+        let position_raw = usize::from(record_id) - usize::from(self.offset);
+        let length = segment.len();
+        let position_vec = (length * position_raw) >> 8;
```
akoshelev, line 318: compilers are perfectly aware how to perform division by a power of two division I think. Your 32 byte assumption can be encoded as a constant and used everywhere

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+                    &segment.x_right.0[256 * i..256 * (i + 1)],
+                    &segment.y_left.0[256 * i..256 * (i + 1)],
+                    &segment.y_right.0[256 * i..256 * (i + 1)],
+                    &segment.prss_left.0[256 * i..256 * (i + 1)],
+                    &segment.prss_right.0[256 * i..256 * (i + 1)],
+                    &segment.z_right.0[256 * i..256 * (i + 1)],
+                ));
+            }
+        }
+    }
```
akoshelev, line 371: this function screams for refactoring imo

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+
+    fn is_empty(&self) -> bool {
+        self.inner.is_empty() || self.inner.values().all(UnverifiedValuesStore::is_empty)
+    }
+
+    fn push(&mut self, gate: Gate, record_id: RecordId, segment: Segment) {
+        // get value_store & create new when necessary
+        // insert segment
```
akoshelev, line 410:
```suggestion
```

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+            .upgrade()
+            .is_some_and(|x| !x.lock().unwrap().is_empty())
+    }
+}
+
+/// Steps used by the validation component of the DZKP
+#[cfg_attr(feature = "descriptive-gate", derive(ipa_macros::Step))]
+pub(crate) enum Step {
+    /// For the execution of the malicious protocol.
+    DZKPMaliciousProtocol,
```
akoshelev, line 477: lets strip the prefix, we know that everything that happens in this module is related to zkp validation

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+        &'st self,
+        source: S,
+    ) -> impl Stream<Item = Result<O, Error>> + 'st
+    where
+        S: Stream<Item = F> + Send + 'st,
+        F: Future<Output = O> + Send + 'st,
+        O: Send + Sync + Clone + 'static,
+        DF: DZKPBaseField,
+    {
+        // chunk_size is undefined in the semi-honest setting, set it to 10, ideally it would be 1
```
akoshelev, line 535: I don't think this optimization gives us anything useful for semi-honest, but makes this code harder to operate/understand. Vote for making chunk size the same in both settings

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+    }
+}
+
+pub struct SemiHonestDZKPValidator<'a, B: ShardBinding> {
+    context: SemiHonestDZKPUpgraded<'a, B>,
+}
+
+impl<'a, B: ShardBinding> SemiHonestDZKPValidator<'a, B> {
+    pub(super) fn new(inner: Base<'a, B>) -> Self {
+        // This is inconsistent with the malicious new() which calls dzkp_upgrade
```
akoshelev, line 551: It feels like you want to do something about it but I am not sure what exactly

ipa-core/src/protocol/context/mod.rs
```diff
@@ -1,9 +1,13 @@
+#[allow(dead_code)]
+mod dzkp_malicious;
```
akoshelev, line 2: you would make it better if you add a folder zkp, put all implementations there and strip DZKP prefix from everywhere: `context::dzkp_validator::DZKPValidator` really does not need to say it twice

ipa-core/src/secret_sharing/replicated/malicious/additive_share.rs
```diff
@@ -53,6 +53,13 @@ impl<F: PrimeField> ExtendableField for F {
     }
 }
 
+/// Trait for fields compatible with DZKPs
+/// Field needs to support conversion to `SegmentEntry`, i.e. `to_segment_entry` which is required by DZKPs
+#[allow(dead_code)]
+pub trait DZKPCompatibleField: Field {
```
akoshelev, line 59: does it need to be in this module or it can be moved to zkp?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+        // i.e. choice of vectorization dimension and Boolean Array type,
+        // user inputs are not a concern as long as it works for all vectorization options
+        debug_assert_eq!(x_left.len(), x_right.len());
+        debug_assert_eq!(x_left.len(), y_left.len());
+        debug_assert_eq!(x_left.len(), y_right.len());
+        debug_assert_eq!(x_left.len(), prss_left.len());
+        debug_assert_eq!(x_left.len(), prss_right.len());
+        debug_assert_eq!(x_left.len(), z_right.len());
+        // check that length is either multiple of 256 or 256 is multiple of length
+        debug_assert!(x_left.len() % 256 == 0 || 256 % x_left.len() == 0);
```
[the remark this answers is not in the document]
danielmasny: 256 % x_left.len() == 0 checks that 256 is a multiple of x_left.len(). Is there a different way to check this that you would prefer?

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: I remove the comment and just keep it consistent with the previous framework

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: didn't Ben Savage or Andy compare shift and divisions and shifts turned out to be more efficient?

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: we have added a metrics counter in the function that is called. This way we can see how often the function is actually called. When used as intended, it will never be called. I still added it since it is better than a panic.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: right now, I just use it here and I don't plan to add any other methods, if so I will change it as you have suggested.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -0,0 +1,765 @@
...
+        }
+    }
+
+    /// test for testing `validated_seq_join`
+    /// similar to `complex_circuit` in `validator.rs`
+    #[tokio::test]
+    async fn complex_circuit_dzkp() -> Result<(), Error> {
+        const COUNT: usize = 100;
+        let world = TestWorld::default();
+        let context = world.malicious_contexts();
```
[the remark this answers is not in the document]
danielmasny: Andy and I figured that we will need both, so malicious contexts need to be upgradeable to contexts that use MACs as well as to contexts that use dzkps. 

Addressing your comment above, right now the backend is not integrated in the API yet. Once we have done that, I will add a test that will fail.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: Chunk collects the stream into a vector. Therefore it is a synchronization point but I am fine either way since it is anyway just semi honest security for which we probably dont need to optimize.

ipa-core/src/protocol/context/mod.rs
danielmasny: Yes, I think that would be much better. Unfortunately as mentioned above there are parts of the protocol where we want to use MACs and dzkps. I added the prefix since otherwise there are many naming collisions with the MAC based approach. We could strip it and then renaming it when importing the module if you prefer that. I find the renaming (as in case of the additive/replicated shares) more confusing.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: This could cause a collision with MACs since the same context might be upgraded to MACs and DZKPs. It will probably call different subroutines with narrow. I still think it is a bit safer to have a different name.

ipa-core/src/protocol/context/dzkp_validator.rs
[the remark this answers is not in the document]
akoshelev: I missed the fact that both sides compare against `x_left`, I don't think there is a better and more concise way to write it.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: Right now the backend is not integrated in the API yet. Once we have done that, I will add a test that will fail.

PR #1000, opened 2024-04-05
add #[allow(clippy::diverging_sub_expression)]

danielmasny:
temporary fix for clippy, once we have implemented the malicious framework, we can fix it more permanently.


ipa-core/src/test_fixture/world.rs
```diff
@@ -403,6 +403,7 @@ impl<const SHARDS: usize> Runner<WithShards<SHARDS>> for TestWorld<WithShards<SH
             .await
     }
 
+    #[allow(clippy::diverging_sub_expression)]
```
akoshelev, line 406: looks like a false alarm to me

PR #1002, opened 2024-04-05
Enable dual-mode for `RecordsStream`

akoshelev:
`RecordsStream` is a convenient abstraction that takes care of polling the underlying raw byte stream, buffering enough data and deserializing them into a client-provided type `T`. It also handles padding and unaligned errors.

Before this change, stream element could only be a `Vec<T>` which means an extra allocation for clients that want to handle items one by one. This is the case for shard receiving streams.

Shard receiving streams are polled by MPC protocols that get records out of them one-by-one. The key difference between MPC receiving streams and shard receiving streams is that the former wants the ability to receive elements out of order. For shard streams, it is not required.

Shard streams can benefit from using `RecordsStream` abstraction quite a lot. I have it in my private branch and it works well. This change enables `RecordsStream` to be used in mode where they yield only one element

### DIFF
--- a/ipa-core/src/helpers/transport/stream/input.rs
+++ b/ipa-core/src/helpers/transport/stream/input.rs
@@ -107,13 +107,23 @@ impl BufDeque {
         })
     }
 
-    /// Deserialize fixed-length items from the buffer.
+    /// Deserialize a single instance of `T` from the buffer with the guarantee that deserialization
+    /// cannot fail, if there is enough bytes in the buffer.
     ///
-    /// Deserializes a single instance of fixed-length-[`Serializable`] type `T` from the stream.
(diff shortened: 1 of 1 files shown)


andyleiserson (approved):
`RecordsStream` is used for receiving the query input (i.e. reports) from report collectors. I actually can't remember why I made it return batches of reports rather than single reports. It may have been a hypothetical performance concern.

In any case, this looks good.

akoshelev:
> It may have been a hypothetical performance concern.

it is probably a legit concern, allocations are cheap and with batching you let client code to process more data at once. for shard API exposed to circuits, it feels more natural to process data record-by-record similarly to MPC API, but we can change it later

PR #1003, opened 2024-04-05
Update async-trait

andyleiserson:
I had previously determined this could be fixed with a cargo update, but I'm not sure why I didn't just update Cargo.toml at the time. Possibly I thought that the fix was in the `syn` crate or one of the proc macro crates, that are not named directly in Cargo.toml?

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -76,7 +76,7 @@ ipa-prf = []
 ipa-macros = { version = "*", path = "../ipa-macros" }
 
 aes = "0.8.3"
-async-trait = "0.1.68"
+async-trait = "0.1.79"
 async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
 axum = { version = "0.5.17", optional = true, features = ["http2"] }
 axum-server = { version = "0.5.1", optional = true, features = [
--- a/ipa-core/src/test_fixture/world.rs
+++ b/ipa-core/src/test_fixture/world.rs
@@ -403,7 +403,6 @@ impl<const SHARDS: usize> Runner<WithShards<SHARDS>> for TestWorld<WithShards<SH
             .await
     }
 
-    #[allow(clippy::diverging_sub_expression)]
     async fn malicious<'a, I, A, O, H, R>(&'a self, _input: I, _helper_fn: H) -> [O; 3]
     where
         I: IntoShares<A> + Send + 'static,
@@ -415,7 +414,6 @@ impl<const SHARDS: usize> Runner<WithShards<SHARDS>> for TestWorld<WithShards<SH
         unimplemented!()
     }
 
-    #[allow(clippy::diverging_sub_expression)]
     async fn upgraded_malicious<'a, F, I, A, M, O, H, R, P>(
         &'a self,
         _input: I,


ipa-core/Cargo.toml
```diff
@@ -76,7 +76,7 @@ ipa-prf = []
 ipa-macros = { version = "*", path = "../ipa-macros" }
 
 aes = "0.8.3"
-async-trait = "0.1.68"
+async-trait = "0.1.79"
```
akoshelev, line 79: I am fairy certain it gets updated automatically because of semver

```bash
cargo tree -i async-trait
async-trait v0.1.77 (proc-macro)
├── axum v0.5.17
│   └── ipa-core v0.1.0 (/Users/koshelev/workspace/rmstuff/ipa/ipa-core)
├── axum-core v0.2.9
│   └── axum v0.5.17 (*)
├── config v0.14.0
│   └── ipa-core v0.1.0 (/Users/koshelev/workspace/rmstuff/ipa/ipa-core)
└── ipa-core v0.1.0 (/Users/koshelev/workspace/rmstuff/ipa/ipa-core)
```

it obviously does not hurt and we could just specify `0.1` here to make it more clear

andyleiserson: It gets updated if you run `cargo update`, delete `Cargo.lock`, or clone a fresh tree. Bumping the version in Cargo.toml forces existing trees to update to the specified version.

For packages that are relatively stable, rounding down the semver requirement seems reasonable. But for packages like this that have evolved over time under the same semver version (which is fine, as long as the changes are backwards compatible), our tree almost certainly wouldn't build with version 0.1.0, so it seems more appropriate to specify a current version. Similarly, if I'm writing some code looking at version x.y.z of the documentation, it seems safest to specify x.y.z as the semver requirement -- any code that worked against x.y.0 should work against x.y.z, but not the other way around.

akoshelev: I had an impression that specifying 0.1.x does not [stop](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) Cargo from moving to 0.1.x+1: 

> The string "0.1.12" is a version requirement. Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero number in the major, minor, patch grouping. In this case, if we ran cargo update time, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0. If instead we had specified the version string as 1.0, cargo should update to 1.1 if it is the latest 1.y release, but not 2.0. The version 0.0.x is not considered compatible with any other version.


In my head, `dependency = 0.1.x` and `dependency = 0.1` are semantically equivalent, which is obviously not true for stable dependencies

danielmasny (approved):
It fixes the issues for me

ipa-core/Cargo.toml
andyleiserson: > I had an impression that specifying 0.1.x does not [stop](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) Cargo from moving to 0.1.x+1

That's true, it does not stop Cargo from moving to a newer semver-compatible version. But as long as the versions in Cargo.lock are consistent with the version requirements in Cargo.toml, Cargo will not go out of its way to move to a newer version -- you have to run `cargo update`.

Bumping the version requirement in Cargo.toml saves the trouble of everyone needing to individually run into the failure and make the mental connection to try `cargo update`.

(The question of whether to check in Cargo.lock is related -- besides preventing spontaneous CI breakage if a package happens to break semver compatibility, it also ensures that everybody using the tree is building the exact same software.)

PR #1004, opened 2024-04-05
IPA bench: log random seed and runtime at info level

andyleiserson:
This does a few things towards the end goal in the description:
* Increase the log level for the random seed from trace to info.
* Adjust the initialization order so that logging is initialized in time to log the random seed.
* Add configuration of a 2nd log target to TestWorld (specifically, we want the benchmark crate `oneshot_ipa` in addition to `ipa_core`).

### DIFF
--- a/ipa-core/benches/oneshot/ipa.rs
+++ b/ipa-core/benches/oneshot/ipa.rs
@@ -1,4 +1,5 @@
 use std::{
+    env,
     num::{NonZeroU32, NonZeroU64, NonZeroUsize},
     time::Instant,
 };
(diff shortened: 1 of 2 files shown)


ipa-core/benches/oneshot/ipa.rs
```diff
@@ -100,11 +100,14 @@ async fn run(args: Args) -> Result<(), Error> {
     let _prep_time = Instant::now();
     let config = TestWorldConfig {
         gateway_config: GatewayConfig::new(args.active()),
+        extra_log_target: option_env!("CARGO_CRATE_NAME"),
```
akoshelev, line 103: I wonder if it is worth just setting or changing the `RUST_LOG` env variable here, may keep the logging API cleaner and more maintainable if we ever need to add 2nd, 3rd target etc

akoshelev (approved):

PR #1005, opened 2024-04-05
Make flatten_iters terminate after the first error.

andyleiserson:


akoshelev (approved):
In the future we may want to use this along with consistency checks where we don't want to terminate the whole stream, but rather log and reject individual records. 

It should be fairly easy to do and I am in favour of having a strict/safe behaviour by default

danielmasny (approved):
Thanks for adjusting it! 

For the DZKPs, terminating the stream seems the right thing to do. Alex mentioned other applications, where we might filter out elements in the streams that are errors. I think we should keep these two use cases separate from each other.

PR #1006, opened 2024-04-07
Implement shard communication channels for MPC circuits

akoshelev:
This change adds the necessary functionality to `Gateway` to be able to communicate between shards.
`get_shard_sender` and `get_shard_receiver` will allow circuits to send and request data from other shards, similarly to MPC send/recv.

There is a difference in how receives are handled for MPC and shards. The former channels use `UnorderedReceiver` that lets them receive records in any order. Shard receivers return a stream that has a FIFO order. This is a requirement that came from analysing shuffle and other protocols that require cross-shard communication. Each shard does not know in advance how many records it expects to receive from any shard.

This also adds a stub for HTTP shard transport, just to make the code compile. Actual HTTP implementation will come later.

In terms of next steps, there remains a building block for resharding shares and sharded shuffle implementation. After these two are complete, in memory implementation will be fully functional

### DIFF
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -6,23 +6,26 @@ mod transport;
 
 use std::num::NonZeroUsize;
 
-pub(super) use receive::ReceivingEnd;
+pub(super) use receive::{MpcReceivingEnd, ShardReceivingEnd};
 pub(super) use send::SendingEnd;
-#[cfg(all(test, feature = "shuttle"))]
-use shuttle::future as tokio;
 #[cfg(feature = "stall-detection")]
 pub(super) use stall_detection::InstrumentedGateway;
+pub use transport::RoleResolvingTransport;
 
 use crate::{
     helpers::{
         buffers::UnorderedReceiver,
         gateway::{
-            receive::GatewayReceivers, send::GatewaySenders, transport::RoleResolvingTransport,
+            receive::{GatewayReceivers, ShardReceiveStream, UR},
+            send::GatewaySenders,
+            transport::Transports,
         },
-        transport::routing::RouteId,
-        HelperChannelId, LogErrors, Message, Role, RoleAssignment, TotalRecords, Transport,
+        HelperChannelId, LogErrors, Message, MpcMessage, RecordsStream, Role, RoleAssignment,
+        ShardChannelId, TotalRecords, Transport,
     },
     protocol::QueryId,
+    sharding::ShardIndex,
+    sync::{Arc, Mutex},
 };
 
 /// Alias for the currently configured transport.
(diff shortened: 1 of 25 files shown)


ipa-core/src/helpers/gateway/receive.rs
```diff
@@ -1,36 +1,61 @@
...
+pub type UR = UnorderedReceiver<
     LogErrors<<RoleResolvingTransport as Transport>::RecordsStream, Bytes, BoxError>,
     Vec<u8>,
 >;
 
-impl<M: Message> ReceivingEnd<M> {
+/// Stream of records received from a peer shard.
+#[derive(Clone)]
+pub struct ShardReceiveStream(
+    pub(super) Arc<Mutex<<ShardTransportImpl as Transport>::RecordsStream>>,
```
andyleiserson, line 55: Does `Arc<Mutex>` make the most sense here, or does something more like a `RefCell` make more sense? I'm thinking that multiple concurrent readers aren't going to do something sensible, so maybe we should try to prevent that.

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -36,14 +40,30 @@ use crate::{
 
 /// An identity of a peer that can be communicated with using [`Transport`]. There are currently two
 /// types of peers - helpers and shards.
-pub trait Identity: Copy + Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static {}
+pub trait Identity:
+    Copy + Clone + Debug + PartialEq + Eq + PartialOrd + Ord + Hash + Send + Sync + 'static
+{
+    fn as_str<'a>(&self) -> Cow<'a, str>;
```
andyleiserson, line 46: Is the lifetime parameter useful here? Seems like returning `Cow<'static, str>` would be fine. If support for borrowing it from `self` is important, then the lifetime parameter should be attached to `&self`.

ipa-core/src/helpers/gateway/mod.rs
```diff
@@ -6,41 +6,52 @@ mod transport;
...
+#[cfg(feature = "in-memory-infra")]
+pub type ShardTransportImpl = TransportImpl<ShardIndex>;
 
 #[cfg(feature = "real-world-infra")]
-pub type TransportImpl = crate::sync::Arc<crate::net::HttpTransport>;
+type TransportImpl = crate::sync::Arc<crate::net::HttpTransport>;
+#[cfg(feature = "real-world-infra")]
+pub type MpcTransportImpl = TransportImpl;
+#[cfg(feature = "real-world-infra")]
+pub type ShardTransportImpl = crate::net::HttpShardTransport;
```
andyleiserson, line 47:
```suggestion
#[cfg(feature = "real-world-infra")]
pub type MpcTransportImpl = crate::sync::Arc<crate::net::HttpTransport>;
#[cfg(feature = "real-world-infra")]
pub type ShardTransportImpl = crate::net::HttpShardTransport;
```

I think better not to give the impression these two types have a common interface. (I'm not sure the intermediate `TransportImpl` alias has much value in the in-memory case either, vs. importing `InMemoryTransport`, but that's a more minor issue.)

andyleiserson (approved):

ipa-core/src/helpers/gateway/mod.rs
```diff
@@ -6,41 +6,52 @@ mod transport;
...
 
-pub type TransportError = <TransportImpl as Transport>::Error;
+pub type MpcTransportError = <MpcTransportImpl as Transport>::Error;
```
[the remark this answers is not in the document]
akoshelev: let me follow up on that

ipa-core/src/helpers/gateway/receive.rs
akoshelev: I think stall detection and MPC receiver working in parallel may be a valid use case here. It does not seem that `RefCell` can work here even if we don't have concurrent readers - it requires borrowing from behind a Dashmap mutex which is a bad idea and hopefully not possible

andyleiserson: Having the `Arc<Mutex>` within `ShardReceiveStream` means that every poll operation on a shard receive stream (even when built without stall detection) will pay a cost to lock the mutex.

I did spend a little time trying to figure out a concrete solution, but I wasn't successful. The ShardReceiveStream can hold a reference to a record counter in the gateway state (rather than the gateway state holding a reference to the receive stream), but holding dashmap entry references directly doesn't work. Maybe both the ShardReceiveStream and the gateway state could hold an `Arc<AtomicU32>`?

It may not be worth the effort to solve this right now. But I think it would at least be worth a comment that there is a potential optimization here (definitely in the case where stall detection is disabled, and possibly even in the case where it is enabled).

akoshelev: Yea that's not great, although this is not the only place where we lock the mutex this way. My understanding that if it is not contended, it does not require a syscall, so it should be fast. 

I'll add a note because as you said, I am not sure what exactly will be required in stall detection. I also don't know how to get rid of `Mutex` because `RefCell` does not seem to work in this case (where you need to return something under DashMap's mutex). `Arc<Stream>` may work though

ipa-core/src/helpers/gateway/mod.rs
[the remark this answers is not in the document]
akoshelev: #1013

PR #1008, opened 2024-04-11
Transposes for vectorized aggregation

andyleiserson:
Add additional macros in transpose implementation, to reduce repetitive code


andyleiserson:
I removed the special transpose for aggregation because that was the bulk of the missing coverage and enough to get over the threshold. The coverage situation will improve once the vectorized aggregation is added.

akoshelev:
I trust that you do the right thing here, but I also find it very hard to comprehend what is going on there. Having no concrete suggestions how to improve things, my vote is to ship it

akoshelev (approved):

PR #1009, opened 2024-04-11
Vectorized aggregation

andyleiserson:
This changes aggregation to work with shares in $\mathbb{F}_2$ rather than $\mathbb{F}_p$ (required to support DZKP-based malicious security), and vectorizes it.


akoshelev:
there are plenty of changes coming from 1008 that add some noise - can you rebase it on top of main after #1008 is merged?

ipa-core/src/protocol/context/semi_honest.rs
```diff
@@ -272,7 +261,7 @@ impl<'a, B: ShardBinding, F: ExtendableField> UpgradedContext<F> for Upgraded<'a
         T: Send,
         UpgradeContext<'a, Self, F>: UpgradeToMalicious<'a, T, M>,
     {
-        UpgradeContext::new(self.narrow(&UpgradeStep::UpgradeSemiHonest), NoRecord)
+        UpgradeContext::new(self.clone(), NoRecord)
```
andyleiserson, line 264: I don't really understand what's going on here, and I'm not sure this change is appropriate -- but it did resolve issues I had with compact gate after removing modulus conversion from OPRF IPA.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+            })
+            .try_flatten_iters::<BitDecomposed<_>, Vec<_>>(),
+    );
+
+    aggregate_values::<_, _, B>(ctx, aggregation_input, num_chunks * N).await
+}
+
+pub type AggResult<const B: usize> = Result<BitDecomposed<Replicated<Boolean, B>>, Error>;
+
+pub async fn aggregate_values<'fut, C, OV, const B: usize>(
```
bmcase, line 201: maybe nice to add a test for just this `aggregate_values` function.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -119,10 +118,20 @@ where
...
 
             if robust || tree_index + span < breakdown_count {
-                futures.push(row_contribution[tree_index].multiply(bit_of_bdkey, bit_c, record_id));
+                futures.push(async move {
+                    let bit_futures = index_contribution
+                        .zip(repeat((bit_of_bdkey, bucket_c.clone())))
+                        .enumerate()
+                        .map(|(i, (a, (b, bucket_c)))| {
+                            a.multiply(b, bucket_c.narrow(&BitOpStep::Bit(i)), record_id)
+                        });
```
benjaminsavage, line 132: This got a lot more complex looking. 

Does it no longer work to multiply a multi-bit value with a single bit? That's too bad. Could we migrate from BitDecomposed to a BooleanArray and use `select` instead?

Doesn't need to be in this PR - this PR is too long already =).

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -131,9 +140,21 @@ where
...
+                *r -= b;
+            }
             if right_index < breakdown_count {
-                row_contribution[right_index] = bdbit_contribution;
+                for (r, b) in row_contribution[right_index]
+                    .iter_mut()
+                    .zip(bdbit_contribution)
+                {
+                    *r = b;
+                }
```
benjaminsavage, line 157: This part also seems like it could be simplified if we stopped using BitDecomposed and used BooleanArray instead.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+                        Ok(mut chunk_pair) => {
+                            assert_eq!(chunk_pair.len(), 2);
+                            let b = chunk_pair.pop().unwrap();
+                            let a = chunk_pair.pop().unwrap();
+                            let (mut sum, carry) =
+                                integer_add::<_, _, _, _, B>(ctx, RecordId::from(i), &a, &b)
+                                    .await?;
+                            if a.len() < usize::try_from(OV::BITS).unwrap() {
+                                sum.push(carry);
+                            }
```
benjaminsavage, line 238: So this implements **wrapping addition**, basically addition modulo `2^OV::BITS`. This could potentially yield unexpected results (like zero) when the sum wraps around.

An alternative would be to implement a "saturating addition".

We could implement that with:

```
if /* it could potentially overflow on this step */ {
  let sum = integer_sat_add(/* stuff */)
} else {
  let (mut sum, carry) = integer_add(/* stuff */).await?;
  sum.push(carry);
}
```

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -213,90 +212,72 @@ where
...
-    let flattened_stream = seq_join(sh_ctx.active_work(), stream_of_per_user_circuits)
-        .flat_map(|x| stream_iter(x.unwrap()));
-
-    // modulus convert feature vector bits from shares in `Z_2` to shares in `Z_p`
-    let converted_feature_vector_bits = convert_bits(
-        prime_field_ctx
-            .narrow(&Step::ModulusConvertFeatureVectorBits)
-            .set_total_records(num_outputs),
-        flattened_stream,
-        0..<FV as SharedValue>::BITS,
```
benjaminsavage, line 280: It's a thing of beauty =)

benjaminsavage (approved):
This is a very long PR... I really wish it was split into smaller, more digestible chunks...

I mostly focused my review on the protocols. It's pretty awesome how you were able to eliminate the conversions to the prime field. I LOVE the way you are able to just chunk streams now (very nice!), and the highly efficient bitwise addition that only deals with as many digits as actually necessary is great. 

This PR actually does about 1/2 the work I was trying to do in the `feature_label_dot_product` protocol... I should have read this PR first before starting to implement! 

So I'm keen to get this landed so that I can build upon this work. I know @bmcase is as well.

The only concern I have is with the wrapping addition (not saturating). I would suggest a follow-up PR to just make it saturating.

ipa-core/src/protocol/context/semi_honest.rs
akoshelev: iirc, it was implemented because an early version of compact gate had to use an ignorelist for steps that don't trigger communications and tracked active steps through `send`. #711 may shed some light on why it was done this way.

I could be wrong, but I believe compact gate uses `narrow` to track steps now, so it may be the reason why you're not getting panic. If you could share the errors you're getting from modulus conversion, we can dig deeper into this. For now it seems that it should just work

you may want to fix the comment too (lines 254..258)

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -145,34 +166,36 @@ where
...
 
     const MAX_BREAKDOWN_COUNT: usize = 256;
     const VALUE: u32 = 10;
 
-    async fn move_to_bucket(count: usize, breakdown_key: usize, robust: bool) -> Vec<Fp32BitPrime> {
-        let breakdown_key_bits =
-            get_bits::<Fp32BitPrime>(breakdown_key.try_into().unwrap(), Gf8Bit::BITS);
-        let value = Fp32BitPrime::truncate_from(VALUE);
+    async fn move_to_bucket(count: usize, breakdown_key: usize, robust: bool) -> Vec<BA8> {
+        let breakdown_key_bits = BitDecomposed::decompose(Gf8Bit::BITS, |i| {
```
akoshelev, line 186: nit: it may be worth explaining why 8 bit value is enough here. (I believe it is related to `MAX_BREAKDOWN_COUNT` and if that's true, maybe a static assertion is possible too)

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    BitDecomposed<Replicated<Boolean, N>>:
+        for<'a> TransposeFrom<&'a Vec<Replicated<TV>>, Error = LengthError>,
+    Vec<BitDecomposed<Replicated<Boolean, B>>>:
+        for<'a> TransposeFrom<&'a [BitDecomposed<Replicated<Boolean, N>>], Error = Infallible>,
+    Vec<Replicated<HV>>:
+        for<'a> TransposeFrom<&'a BitDecomposed<Replicated<Boolean, B>>, Error = LengthError>,
+{
+    let num_chunks = (contributions_stream_len + N - 1) / N;
+    // Indeterminate TotalRecords is currently required because aggregation does not poll futures in
+    // parallel (thus cannot reach a batch of records).
```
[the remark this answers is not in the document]
akoshelev: Yes, but the size of each record is increased proportionally to the vectorization factor. With `N` == `1024`, network payload/buffer capacity will be the same as unvectorized with `set_total_records(szie).

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
akoshelev: do we want to move this circuit to a separate function? Seeing `wrapping_integer_add` or `sat_integer_add` here makes the intent more obvious imo

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -213,90 +212,72 @@ where
...
+    let flattened_stream = Box::pin(
```
akoshelev, line 267: does this now work w/o multithreading or we still rely on it being on for aggregation?

ipa-core/src/query/runner/oprf_ipa.rs
```diff
@@ -113,11 +113,11 @@ where
 
         let aws = config.attribution_window_seconds;
         match config.per_user_credit_cap {
-            8 => oprf_ipa::<C, BA8, BA3, BA20, BA3, F>(ctx, input, aws).await,
-            16 => oprf_ipa::<C, BA8, BA3, BA20, BA4, F>(ctx, input, aws).await,
-            32 => oprf_ipa::<C, BA8, BA3, BA20, BA5, F>(ctx, input, aws).await,
-            64 => oprf_ipa::<C, BA8, BA3, BA20, BA6, F>(ctx, input, aws).await,
-            128 => oprf_ipa::<C, BA8, BA3, BA20, BA7, F>(ctx, input, aws).await,
+            8 => oprf_ipa::<C, BA8, BA3, HV, BA20, BA3, 256>(ctx, input, aws).await,
```
akoshelev, line 116: is it worth to have a generic histogram value parameter with everything else being hardcoded or I misunderstand what's going on here?

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+        let par_agg_ctx = ctx
+            .narrow(&Step::Aggregate(depth))
+            .set_total_records(TotalRecords::Indeterminate);
+        num_rows = (num_rows + 1) / 2;
+        aggregated_stream = Box::pin(aggregated_stream.try_chunks(2).enumerate().then(
+            move |(i, chunk_res)| {
+                let ctx = par_agg_ctx.clone();
+                async move {
+                    match chunk_res {
+                        Err(e) => Err(e.1),
```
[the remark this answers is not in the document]
akoshelev: [`try_chunks`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_chunks) combinator returns a tuple (buf, error) in case if the underlying stream failed. This code just ignores buffered items and returns the error which is the right thing to do imo

ipa-core/src/secret_sharing/vector/transpose.rs
```diff
@@ -768,6 +789,83 @@ impl_transpose_shares_ba_to_bool_small!(BA8, 256, 8, test_transpose_shares_ba_to
 impl_transpose_shares_ba_to_bool_small!(BA5, 256, 5, test_transpose_shares_ba_to_bool_256x5);
 impl_transpose_shares_ba_to_bool_small!(BA3, 256, 3, test_transpose_shares_ba_to_bool_256x3);
 
+// Special transpose used for "aggregation intermediate". See [`aggregate_contributions`] for details.
+macro_rules! impl_aggregation_transpose {
```
akoshelev, line 793: Maybe we can add a test that just tests the transpose functionality here - I worry about only having coverage for it through the aggregation

akoshelev (approved):

ipa-core/src/helpers/stream/chunks.rs
```diff
@@ -195,52 +193,208 @@ where
...
+            (*this.process_fn)(idx, chunk_data.expect("ensured a full chunk")),
+            chunk_type,
+        ))))
+    }
+}
+
+/// Process stream through a function that operates on chunks.
+///
+/// Processes `stream` by collecting chunks of `N` items into `buffer`, then calling `process_fn`
+/// for each chunk. If there is a partial chunk at the end of the stream, `dummy_fn` is called
```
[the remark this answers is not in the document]
andyleiserson: Yes, I considered that -- it would require implementing `Default` on several layers, down to `AdditiveShare`. There is also a semantic required some places this is used that isn't exactly `Default` -- the dummy records need to contribute nothing to the overall result.

I think I would slightly prefer the version with `Default`, but I didn't want to presume to make that change given the impact.

ipa-core/src/protocol/context/semi_honest.rs
andyleiserson: The errors were "error: ipa_macros::step expects an enum with variants that match the steps in steps.txt" and "error[E0277]: the trait bound `compact::Compact: StepNarrow<UpgradeStep>` is not satisfied" (i.e. the usual errors when a step is not exercised in the steps.txt generation).

Part of why I was unsure what to do here, is that these methods were not being exercised -- so I wasn't sure there would be test coverage for whatever change I made.

But with the benefit of your comment and some more thinking, I have a proposal here that I'm more confident in. I'm going to open a separate PR for it (to be considered in advance of this one).

andyleiserson: #1030. After this PR is merged / rebased to include that, it will need to stub these impls based on `cfg(descriptive-gate)`, because it removes the last place we invoke share upgrades in the protocol (modulus conversion). They'll later need to come back when we add MAC-based malicious security for PRF evaluation and shuffle.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
andyleiserson: The issue here is that we're working with a 2-D array. The type `BitDecomposed<AdditiveShare<Boolean, N>>` actually does contain a boolean array -- the vector type in `AdditiveShare<Boolean, N>` is the boolean array type of length N.

We could flatten everything into a boolean array of size N * HV_BITS, but then we'd have to do a bunch of 2-D array index calculations, which I think would make things even more complicated.

I do like the idea of using select, though. I'll see if I can do something with that.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
andyleiserson: The old version didn't have saturating math either, but I agree that now that we can do it relatively cheaply, it's probably worth doing. Is there any interaction with DP?

ipa-core/src/query/runner/oprf_ipa.rs
andyleiserson: I think that's a fair question -- I was just modeling it after what was already there (`HV` is the bitwise equivalent of `F`).

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
andyleiserson: I believe it is still required, perhaps even more so. We can add the buffered stream adapter if we really want things to work without multi-threading.

ipa-core/src/helpers/stream/chunks.rs
```diff
@@ -195,52 +193,208 @@ where
...
+                None if this.buffer.len() != 0 => {
+                    // Input stream ended, but we still have some items to process.
+                    let remainder_len = this.buffer.len();
+                    this.buffer.resize_with(N, dummy_fn.take());
+                    break (this.buffer.take(), ChunkType::Partial(remainder_len));
+                }
+                None => {
+                    // Input stream ended at a chunk boundary. Unlike the partial chunk case, we return
+                    // None, so we shouldn't be polled again regardless of dummy_fn signalling, but
+                    // might as well make this a fused stream since it's easy.
```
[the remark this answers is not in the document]
andyleiserson: The case where fusing the inner stream doesn't work very well is when we terminate due to an error -- we'd have to drain it.

ipa-core/src/helpers/stream/chunks.rs
```diff
@@ -344,22 +508,54 @@ mod tests {
...
 
-    impl<T: PartialEq, const N: usize> PartialEq<[T]> for Chunk<T, N> {
-        fn eq(&self, other: &[T]) -> bool {
-            if self.chunk_type == ChunkType::Full && other.len() != N {
-                return false;
-            }
-            self.data[..other.len()] == *other
-        }
+    #[tokio::test]
+    async fn process_slice_chunks() {
```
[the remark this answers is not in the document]
andyleiserson: The slice chunk processor was already there; the stream processor is new; neither had a test for empty input, but I'm adding it to both.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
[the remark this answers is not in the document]
andyleiserson: I will add a comment.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
andyleiserson: After looking at it some, I don't think it makes sense to do in this PR... it's a 100 lines elsewhere for 10 lines here tradeoff. Which may be the right tradeoff, if it improves the modularity of the code, but I don't think it's the right tradeoff if those 100 lines are getting added to this PR.

The concrete options I see are:
* Implementing `BooleanArrayMul` for `BitDecomposed<AdditiveShare<Boolean, N>>`. This in turn requires either adding some more `Foo: SecureMul<C>` bounds in various places, or else adding a `C` type parameter to the `BooleanArrayMul` trait.
* Converting `select` into a trait so we can implement it for arbitrary types.

andyleiserson:
Capturing some open items:

- [ ] Test for aggregate_values
- [ ] Saturating addition
- [ ] `dummy_fn` changes (`Default`? rename?)
- [ ] Aggregation transpose test

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
benjaminsavage: I'm totally supportive of landing this PR as is and working on this in a future PR.

andyleiserson: My plan is to defer this and saturating addition to a follow-up. I am working on the (rather gruesome) merge conflicts presently, but I'm hoping this can be ready tonight or tomorrow.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
benjaminsavage: The reason we never had such logic before is that we were moving into a huge field, and assumed we would never encounter a sum that reached 2^32. Now that we wrap around at a configurable value, it's possible to select a number that's fairly low.

I think it might be fine to just document this behavior and warn everyone to please select an output type that's sufficiently large so that there is no overflow. That would keep the code simpler. I don't think there's any impact on DP. If there is wrapping, it will essentially *decrease* the overall contribution of some people to the query output. But it's not predictable. Some people will get their contribution removed and others will not. So if we take a "worst case" scenario type of approach, we probably can't say anything definitive and cannot benefit from this.

andyleiserson:
Double-posting a comment I made in a thread here for better visibility:

My plan is to defer this [using `select` in `move_value_to_bucket`] and saturating addition to a follow-up. I am working on the (rather gruesome) merge conflicts presently, but I'm hoping this can be ready tonight or tomorrow.

ipa-core/src/query/runner/oprf_ipa.rs
benjaminsavage: Yeah, this is an unsolved problem. We never made anything runtime configurable. We just re-built the code for every different combination of query params =).

benjaminsavage:
I think it's fine to defer the saturation addition as well.

andyleiserson:
https://github.com/private-attribution/ipa/pull/1046 has the promised follow-up changes.

PR #1011, opened 2024-04-12
Fix collect_steps

andyleiserson:
Ignore random seed message in captured output.


akoshelev:
so the idea was to print the random seed if test fails. Given that cargo silences the output by default, that seemed like a good way to avoid having noise when tests pass. 

Not sure if there is a better way to deal with it though - I am not sure I am a big fan of having a random seed for unit tests, but it is what it is

akoshelev:
> A more comprehensive solution might be to figure out how to make tracing (at least at/below info level) go to stdout in unit tests. But for now I made collect_steps skip over this message.

yea we probably want to use test writer
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.TestWriter.html

akoshelev (approved):

Issue #967, opened 2024-03-08
[Correctness bug] Running OPRF with 1M rows results in incorrect results


eriktaubeneck:
See [this run](https://draft-mpc.vercel.app/query/view/shaky-flume-2998).

```
Coordinator: 2024-03-08T09:09:16.140228Z INFO ipa_core::cli::playbook::ipa: Running IPA for 1000000 records took 6609.699598126s
Coordinator: 2024-03-08T09:09:16.142187Z INFO report_collector: IpaQueryConfig { per_user_credit_cap: 64, max_breakdown_key: 64, attribution_window_seconds: None, num_multi_bits: 3, plaintext_match_keys: true }
Coordinator: 2024-03-08T09:09:16.142263Z INFO ipa_core::cli::playbook:
Coordinator: +-----+-------------+-------------+-------+
Coordinator: | Row | Expected | Actual | Diff? |
Coordinator: +=========================================+
Coordinator: | 0 | Some(18414) | Some(18417) | X |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 1 | Some(17976) | Some(17976) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 2 | Some(18409) | Some(18409) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 3 | Some(18586) | Some(18586) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 4 | Some(18683) | Some(18683) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 5 | Some(18444) | Some(18444) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 6 | Some(18327) | Some(18327) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 7 | Some(18422) | Some(18422) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 8 | Some(17919) | Some(17919) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 9 | Some(18525) | Some(18525) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 10 | Some(18377) | Some(18377) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 11 | Some(18141) | Some(18141) | |
Coordinator: |-----+-------------+-------------+-------|
Coordinator: | 12 | Some(18752) | Some(18752) | |
[...]

PR #1015, opened 2024-04-16
Duplicate timestamp and input-sorting changes

andyleiserson:
IPAv3 sorts the input records, so add sorting of input records to ipa_in_the_clear, and remove it from input drivers.

IPAv3 also shuffles the input records, so if there are duplicate timestamps among a single user's records, the output may be non-deterministic. To avoid end-to-end test failures, modify the test data generator to avoid generating duplicate timestamps. Add a directed test to exercise duplicate timestamps.

Possibly a fix for #967.

### DIFF
--- a/ipa-core/src/test_fixture/event_gen.rs
+++ b/ipa-core/src/test_fixture/event_gen.rs
@@ -1,3 +1,10 @@
+use std::{
+    collections::HashSet,
+    num::{NonZeroU32, NonZeroU64},
+};
+
+use crate::{rand::Rng, test_fixture::ipa::TestRawDataRecord};
+
 #[derive(Copy, Clone, Hash, Ord, PartialOrd, Eq, PartialEq)]
 struct UserId(u64);
 
@@ -26,6 +33,10 @@ impl UserId {
     pub const FIRST: Self = Self(1);
 }
 
+// 7 days = 604800 seconds fits in 20 bits
+pub type Timestamp = u32;
+pub type NonZeroTimestamp = NonZeroU32;
+
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 pub enum ReportFilter {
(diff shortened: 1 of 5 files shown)


akoshelev:
@eriktaubeneck this may solve #967

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -41,13 +43,31 @@ impl GroupingKey for TestRawDataRecord {
     }
 }
 
+/// Insert `record` into `user_records`, maintaining timestamp order.
+///
+/// If there are existing records with the same timestamp, inserts the new record
+/// randomly in any position that maintains timestamp order.
+fn insert_sorted(user_records: &mut Vec<TestRawDataRecord>, record: TestRawDataRecord) {
```
akoshelev, line 50: have you had a chance to test how long it takes to generate ~20M events with this change? I don't expect it to be terrible, but it would be great to know that for sure

ipa-core/src/test_fixture/event_gen.rs
```diff
@@ -166,15 +170,21 @@ impl<R: Rng> EventGenerator<R> {
...
-        // Generate a new random timestamp between [0..`max_timestamp`).
-        // This means the generated events must be sorted by timestamp before being
-        // fed into the IPA protocols.
-        let current_ts = self.rng.gen_range(0..self.config.max_timestamp.get());
+    fn gen_event(&mut self, idx: usize) -> TestRawDataRecord {
+        let user_id = self.users[idx].user_id;
+
+        // Generate a new random timestamp between [0..`max_timestamp`) and distinct from
+        // already-used timestamps.
+        let current_ts = loop {
```
akoshelev, line 182: this has some potential to stall, so maybe we could check if `used_timestamp` size == max_timestamp?

akoshelev (approved):

ipa-core/src/test_fixture/ipa.rs
andyleiserson: Running the revised event generation and `ipa_in_the_clear` with max_events_per_user up to 1,000 takes 5-10% longer than the previous version, which seems tolerable to me. With max_events_per_user of 10,000 the increase is more like 40%.

Issue #953, opened 2024-02-14
Enable BK, TV, TS to be runtime parameters


danielmasny:
@[andyleiserson](https://github.com/andyleiserson) andyleiserson [on Dec 28, 2023](https://github.com/private-attribution/ipa/pull/904#discussion_r1437914983)
We eventually want at least BK, and maybe also TV and TS to be runtime parameters, right? Are we tracking that work somewhere? (I think it will be a fair amount of work -- certainly not something that makes sense to do in this PR.)

Collaborator
Author
@[danielmasny](https://github.com/danielmasny) danielmasny [on Jan 12](https://github.com/private-attribution/ipa/pull/904#discussion_r1451046654)
Would we compile all possible choices of BA types for them? I am not aware that we are tracking this somewhere.

Collaborator
@[akoshelev](https://github.com/akoshelev) akoshelev [2 days ago](https://github.com/private-attribution/ipa/pull/904#discussion_r1487145690)
yes, likely we would use a macro for it

PR #1019, opened 2024-04-22
61bit field

danielmasny:
straightforward implementation of 61bit fields, i.e. prime fields for prime 2^61-1

### DIFF
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -61,6 +61,9 @@ macro_rules! field_impl {
     ( $field:ident, $store:ty, $store_multiply:ty, $bits:expr, $prime:expr ) => {
         use super::*;
 
+        // check container for multiply is large enough
+        const_assert!((<$store_multiply>::MAX >> $bits) as u128 >= (<$store>::MAX) as u128);
+
         #[derive(Clone, Copy, PartialEq, Eq)]
         pub struct $field(<Self as SharedValue>::Storage);
 
@@ -165,8 +168,6 @@ macro_rules! field_impl {
 
             fn mul(self, rhs: Self) -> Self::Output {
                 debug_assert!(<$store>::try_from(Self::PRIME).is_ok());
-                // check container for multiply is large enough
-                const_assert!((<$store_multiply>::MAX >> $bits) as u128 >= (<$store>::MAX) as u128);
                 let c = <$store_multiply>::from;
                 // TODO(mt) - constant time?
                 // TODO(dm) - optimize arithmetics?
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -4,6 +4,7 @@ use generic_array::GenericArray;
 
 use super::Field;
 use crate::{
+    const_assert,
     ff::{Serializable, U128Conversions},
     impl_shared_value_common,
     protocol::prss::FromRandomU128,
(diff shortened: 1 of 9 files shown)


andyleiserson:
Is it necessary to add the new field to `enum FieldType`? In #1009 I make FieldType a no-op for IPA queries and add a new type parameter `HV` for the output histogram values (which will itself need to be changed, along with `TV`, `BK`, and `TS`, for #953).

ipa-core/src/ff/prime_field.rs
```diff
@@ -428,6 +434,65 @@ mod fp32bit {
     }
 }
 
+mod fp61bit {
+    field_impl! { Fp61BitPrime, u64, u128, 61, 2_305_843_009_213_693_951
+    }
```
andyleiserson, line 439: Did rustfmt insist on this, or would it tolerate putting it back on the previous line?

andyleiserson (approved):

ipa-core/src/ff/prime_field.rs
```diff
@@ -428,6 +431,64 @@ mod fp32bit {
...
+        #[test]
+        fn sixty_one_bit_prime() {
+            let x = Fp61BitPrime::truncate_from(2_305_843_009_213_693_950_u64); // PRIME - 1
+            let y = Fp61BitPrime::truncate_from(2_305_843_009_213_693_949_u64); // PRIME - 2
+
+            assert_eq!(x - y, Fp61BitPrime::ONE);
+            assert_eq!(y - x, Fp61BitPrime::truncate_from(Fp61BitPrime::PRIME - 1));
+            assert_eq!(y + x, Fp61BitPrime::truncate_from(Fp61BitPrime::PRIME - 3));
+
+            assert_eq!(x * y, Fp61BitPrime::truncate_from(2_u32),);
```
akoshelev, line 451:
```suggestion
            assert_eq!(x * y, Fp61BitPrime::truncate_from(2_u32));
```

ipa-core/src/ff/prime_field.rs
```diff
@@ -163,9 +163,15 @@ macro_rules! field_impl {
...
 
             fn mul(self, rhs: Self) -> Self::Output {
-                debug_assert!(u32::try_from(Self::PRIME).is_ok());
-                let c = u64::from;
+                debug_assert!(<$store>::try_from(Self::PRIME).is_ok());
+                // check container for multiply is large enough
+                debug_assert!(
+                    <$store_multiply>::MAX >> $bits
+                        >= <$store_multiply>::try_from(<$store>::MAX).unwrap()
+                );
```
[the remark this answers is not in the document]
akoshelev: Andy recently added `const_assert!` macro that can be useful here.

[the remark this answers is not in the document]
danielmasny: I cannot use try_from since it is non_constant. I used as u128 instead to make const_assert! work. The assert only works as long as we dont use larger (or not unsigned) types than u128.

ipa-core/src/ff/prime_field.rs
```diff
@@ -165,10 +166,7 @@ macro_rules! field_impl {
             fn mul(self, rhs: Self) -> Self::Output {
                 debug_assert!(<$store>::try_from(Self::PRIME).is_ok());
                 // check container for multiply is large enough
-                debug_assert!(
-                    <$store_multiply>::MAX >> $bits
-                        >= <$store_multiply>::try_from(<$store>::MAX).unwrap()
-                );
+                const_assert!((<$store_multiply>::MAX >> $bits) as u128 >= (<$store>::MAX) as u128);
```
akoshelev, line 169: hopefully the compiler will optimize it, but you could also lift this statement outside of this function because it belongs to the macro itself

akoshelev (approved):

PR #1021, opened 2024-04-25
SecureMul Trait Implementation

danielmasny:
This PR changes the SecureMul Trait implementation:

Previously:
The SecureMul Trait was implemented for:
 - any (Context, SemiHonest::AdditiveShare)
 - (Malicious::Upgraded, Malicious::AdditiveShare)

Now: 
SecureMul is implemented for:
 - (SemiHonest::Context, SemiHonest::AdditiveShare)
 - (SemiHonest::Upgraded, SemiHonest::AdditiveShare<Primefield>)
 - (SemiHonest::DZKPUpgraded, SemiHonest::AdditiveShare)
 - (Malicious::Upgraded, Malicious::AdditiveShare)

This is necessary to allow a customized implementation for (which will be part of a follow up PR)
 - (Malicious::DZKPUpgraded, SemiHonest::AdditiveShare<BooleanTypes>)

Now, SemiHonest::AdditiveShare::multiply does not work anymore when the context is Malicious::Context (i.e. a malicious context that is not upgraded). This will add more safety. 
Further, any protocol such as BasicProtocols that require a MPC multiplication now require that SecureMul is implemented (in a trait bound). 
The trait bounds are a bit ugly since sometimes for a context, I need to add to the trait bound that the upgraded version implements SecureMul. I hope this can be improved but it is unclear, especially since Malicious::DZKPUpgraded will only support multiplications for BooleanTypes for now.


ipa-core/src/protocol/basics/mul/mod.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) mod malicious;
 mod semi_honest;
 pub(in crate::protocol) mod sparse;
 
+#[allow(unused_imports)]
```
akoshelev, line 24: why do you need that?

danielmasny: clippy was complaining, some of the github runs failed without it

akoshelev: most of the time it complains about legit issues - if we don't use this multiplication anymore, we shouldn't export it.

this suppression would make more sense to me as temporary measure

danielmasny: I think it is related to the fact that we only use it when running maliciously secure IPA. In semi honest IPA, it is not used.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -192,7 +192,10 @@ where
     Replicated<TS>: BooleanArrayMul,
     Replicated<TV>: BooleanArrayMul,
     F: PrimeField + ExtendableField,
-    Replicated<F>: Serializable,
+    Replicated<F>: Serializable + SecureMul<<C as UpgradableContext>::UpgradedContext<F>>,
+    Replicated<Boolean>: SecureMul<C>,
+    Replicated<Fp25519, 64>: SecureMul<C>,
```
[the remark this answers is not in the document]
danielmasny: I agree that the trait bounds need to be fixed. I would like to just add the trait bound C:Upgradable. I would then implement SecureMul for all Boolean types for <C as Upgradable>::DZKPUpgraded. Further, implement SecureMul for all primefield types for <C as Upgradable>::Upgraded. This would require to change all functions to use the new maliciousAPI and upgrade the context whenever multiplications are used. I believe this would resolve all the trait bounds but I am not quite there yet.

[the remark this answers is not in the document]
danielmasny: https://github.com/private-attribution/ipa/issues/1022

andyleiserson:
I was fighting with similar trait bound issues while working on changes to make reveals more like our previous malicious downgrades. Reveal has a similar issue where it is implemented for `C: Context`, which enables lots of stuff to invoke reveal in ways that are not secure. In any case, informed by that work, here is a version of your change that eliminates some trait bounds: https://github.com/andyleiserson/ipa/commit/bf0855883da9f2db2e32d0f08d1c5da78034866d

A bunch of the problems here are due to our protocols working against `C: Context`. I think it would be more appropriate for most protocols to work against `C: UpgradedContext`. But that is a relatively big change that we probably don't want to tackle now, or all at once. (One complication is that `UpgradedContext` has a type parameter and associated types. I think we can actually simplify that, and I did some work in that direction, but it's based on top of my vectorized sort branch, because I remove modulus conversion from OPRF IPA in the vectorized sort branch and modulus conversion depends heavily on UpgradedContext associated types.)

ipa-core/src/protocol/basics/mul/semi_honest.rs
```diff
@@ -93,24 +98,67 @@ where
 
 /// Implement secure multiplication for semi-honest contexts with replicated secret sharing.
 #[async_trait]
-impl<C, F, const N: usize> super::SecureMul<C> for Replicated<F, N>
+impl<'a, B, F, const N: usize> super::SecureMul<SemiHonestContext<'a, B>> for Replicated<F, N>
```
andyleiserson, line 101: In my branch I added the following comment on this impl:

```
// TODO: This impl should be removed, and the (relatively few) things that truly need
// to invoke multiplies on a base context should call the routines directly. However,
// there are too many places that unnecessarily invoke multiplies on a base context
// to make that change right now.
```

andyleiserson (commented):
Something else to keep an eye on is whether feature_label_dot_product is staying in sync with the main implementation.

danielmasny:
> I was fighting with similar trait bound issues while working on changes to make reveals more like our previous malicious downgrades. Reveal has a similar issue where it is implemented for `C: Context`, which enables lots of stuff to invoke reveal in ways that are not secure. In any case, informed by that work, here is a version of your change that eliminates some trait bounds: [danielmasny/ipa@dzkp_mult...andyleiserson:ipa:dzkp-mult-bounds](https://github.com/danielmasny/ipa/compare/dzkp_mult...andyleiserson:ipa:dzkp-mult-bounds)
> 
> A bunch of the problems here are due to our protocols working against `C: Context`. I think it would be more appropriate for most protocols to work against `C: UpgradedContext`. But that is a relatively big change that we probably don't want to tackle now, or all at once. (One complication is that `UpgradedContext` has a type parameter and associated types. I think we can actually simplify that, and I did some work in that direction, but it's based on top of my vectorized sort branch, because I remove modulus conversion from OPRF IPA in the vectorized sort branch and modulus conversion depends heavily on UpgradedContext associated types.)

Thanks! I have checked the suggested tweaks/changes. Wouldn't it be necessary to change everything back once I want to support both, malicious and semi honest? I agree that the input C should be either upgraded or upgradable in case you upgrade within the function. I think the trait bounds will be easy to resolve once I transition all the protocols to malicious security. At that point, I will need to transition all protocols to upgraded or upgradable which then works for both, semi honest and malicious security.

andyleiserson:
I do think that more changes will be necessary to reach the end state... I'm not sure what in particular gets reverted though. I'm doubtful that `C: Context` is likely to work in the end state. And I'd like to avoid bounds of `C: Context` + `OtherTypes: SecureMul<C>`. Maybe `C: Context` + `OtherTypes: BasicProtocols<C>`, but even that I believe can be avoided in lots of cases where `OtherType` is a concrete type (like `Boolean` or a bool vector).

Maybe it makes sense to change some of the bounds to `DZKPContext` now? As I understand it, `DZKPContext` is basically `UpgradedContext`, but generic over semi-honest and DZKP malicious rather than over semi-honest and MAC malicious. So if the goal is to minimize things that need to change later, `DZKPContext` seems like a good choice. If malicious impls are necessary to get that to compile, we can stub them with `unimplemented!()` for now.

danielmasny:
> I do think that more changes will be necessary to reach the end state... I'm not sure what in particular gets reverted though. I'm doubtful that `C: Context` is likely to work in the end state. And I'd like to avoid bounds of `C: Context` + `OtherTypes: SecureMul<C>`. Maybe `C: Context` + `OtherTypes: BooleanProtocols<C>`, but even that I believe can be avoided in lots of cases where `OtherType` is a concrete type (like `Boolean` or a bool vector).
> 
> Maybe it makes sense to change some of the bounds to `DZKPContext` now? As I understand it, `DZKPContext` is basically `UpgradedContext`, but generic over semi-honest and DZKP malicious rather than over semi-honest and MAC malicious. So if the goal is to minimize things that need to change later, `DZKPContext` seems like a good choice. If malicious impls are necessary to get that to compile, we can stub them with `unimplemented!()` for now.

Yes, C:Context wont work, but we still need a generic C with a trait bound rather than just a concrete struct as input. I was also thinking of DZKPContext (see https://github.com/private-attribution/ipa/issues/1022) which is the trait that both the semi honest and the malicious DZKPUpgraded struct implement. It is similar to the UpgradedContext trait, which is implemented by the semi honest and malicious upgraded struct. I was thinking of adding for any Boolean Type impl<C: DZKPContext> AdditiveShare<Boolean type>::SecureMul<C>, probably in the macro that implements the Boolean Type. 

The Upgradable trait is already tied to Upgraded and DZKPContext, so we won't need trait bounds like AdditiveShare: SecureMul<<C as Upgradable>::Upgraded> in functions that upgrade the context and then run a multiplication with the upgraded context. 

I was hoping to make all these changes in the PR where I upgrade all the protocols to use the malicious API framework and tried to make minimal changes to files like prf_sharding and feature_label_dot_product.

andyleiserson:
I'm not sure that it will be that bad to use `SemiHonestContext` or `DZKPContext` where necessary to satisfy the trait solver. If going with temporary trait bounds, maybe it makes sense to mark them somehow (`TODO(dzkp)` or something like that) so they are easier to identify for cleanup.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -716,6 +716,7 @@ mod tests {
         secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, IntoShares},
         test_fixture::{join3v, Reconstruct, TestWorld},
     };
+    //use crate::test_fixture::Runner;
```
akoshelev, line 719: I'd rather not have trival code commented and committed to main. We can add this import back when we're ready. Same for `.malicious_contexts`

akoshelev:
>  And I'd like to avoid bounds of C: Context + OtherTypes: SecureMul<C>. Maybe C: Context + OtherTypes: BooleanProtocols<C>, but even that I believe can be avoided in lots of cases where OtherType is a concrete type (like Boolean or a bool vector).

I'd consider this basic requirement before we can move forward. I can't imagine us slapping `SecureMul` bound everywhere in OPRF. As long as we are working on it and fix it soon, I am fine pushing this

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -738,7 +738,8 @@ mod tests {
         let h3_shares: Vec<Replicated<Fp31>> = shared_inputs.iter().map(|x| x[2].clone()).collect();
 
         let futures = world
-            .malicious_contexts()
+            .contexts()
+            //.malicious_contexts()
```
andyleiserson, line 742: Here I think it would be better to keep the explanatory TODO comment, and remove the code.

ipa-core/src/protocol/basics/mul/semi_honest.rs
```diff
@@ -92,25 +97,74 @@ where
...
+    }
+}
+
+/// Implement secure multiplication for semi-honest upgraded
+#[async_trait]
+impl<'a, B, F, const N: usize> super::SecureMul<UpgradedSemiHonestContext<'a, B, F>>
+    for Replicated<F, N>
+where
+    B: sharding::ShardBinding,
+    F: Field + FieldSimd<N> + PrimeField,
```
andyleiserson, line 131:
```suggestion
    F: PrimeField + FieldSimd<N>,
```

ipa-core/src/protocol/basics/mul/semi_honest.rs
```diff
@@ -2,18 +2,23 @@ use async_trait::async_trait;
...
 use crate::{
     error::Error,
-    ff::Field,
+    ff::{Field, PrimeField},
     helpers::Direction,
     protocol::{
         basics::{mul::sparse::MultiplyWork, MultiplyZeroPositions},
-        context::Context,
+        context::{
+            dzkp_semi_honest::DZKPUpgraded,
```
andyleiserson, line 10:
```suggestion
            dzkp_semi_honest::DZKPUpgraded as SemiHonestDZKPUpgraded,
```
(and update the references below)

ipa-core/src/protocol/ipa_prf/prf_sharding/bucket.rs
```diff
@@ -76,17 +81,16 @@ impl From<MoveToBucketError> for Error {
 ///
 /// ## Errors
 /// If `breakdown_count` does not fit into `BK` bits or greater than or equal to $2^9$
-pub async fn move_single_value_to_bucket<C, S, F>(
-    ctx: C,
+pub async fn move_single_value_to_bucket<'a, F>(
+    ctx: UpgradedSemiHonestContext<'a, NotSharded, F>,
```
andyleiserson, line 85: I wrote `NotSharded` because it was the easy thing to do, but we'll probably need to change this to `B: ShardBinding` fairly soon. But I don't think it's important to update in this PR.

(I left a comment on one of the sharding PRs about the safety risk of protocols having access to the shard API -- it occurs to me that writing against `B: ShardBinding` is actually reasonably safe, in that if your protocol is generic over sharded or not, then you probably can't access the sharding API.)

andyleiserson (approved):
Looks good to me, but might be good for @akoshelev to take a look as well, since some of this is me reviewing my own changes.

ipa-core/src/ff/boolean.rs
```diff
@@ -30,12 +26,9 @@ impl Boolean {
     }
 }
 
-impl ExtendableField for Boolean {
-    type ExtendedField = Gf32Bit;
-
-    fn to_extended(&self) -> Self::ExtendedField {
-        Gf32Bit::try_from(self.as_u128()).unwrap()
-    }
+impl PrimeField for Boolean {
```
akoshelev, line 29: I wonder if this could hurt us in the future - not in terms of correctness, but performance-wise. Doing inefficient multiplications or any other specialized protocol that assumes prime fields to be large. @andyleiserson?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -458,17 +458,13 @@ where
 /// # Panics
 /// Propagates errors from multiplications
 #[tracing::instrument(name = "attribute_cap_aggregate", skip_all)]
-pub async fn attribute_cap_aggregate<C, BK, TV, TS, SS, S, F>(
-    sh_ctx: C,
+pub async fn attribute_cap_aggregate<'a, BK, TV, TS, SS, F>(
+    sh_ctx: SemiHonestContext<'a>,
```
akoshelev, line 462: I remember some conversations around using the actual type rather than trait bound for contexts, but I can't find it for some reason. Is the idea that once we get to malicious implementation we would only support active security? Are we losing the ability of executing both semi-honest and malicious version with the same protocol code by doing this change or not?

danielmasny: No, this is just a temporary fix

ipa-core/src/ff/boolean.rs
danielmasny: I think this only impacts the attribution logic which we anyway plan to depreciate. Everything else is using vectorization.

andyleiserson: Here is another option: https://github.com/andyleiserson/ipa/commit/0e7307ef8bdc6745c142c65b415313b10a317b82

It is a little weird to implement PrimeField here, since it was mostly used with larger fields in mind. On the other hand, I don't think there's any real impact. It's only used in old protocols that OPRF IPA isn't using. And it's a bit weird to implement ExtendableField too -- what is the context where we want to treat Boolean as a field embedding in Gf32Bit? But we can't implement neither, without more involved changes.

Issue #924, opened 2024-01-18
Upgrade HTTP dependency stack (hyper, rustls, axum, etc)


akoshelev:
Hyper advanced past version 1.0 and we started lagging behind in terms of features/bug fixes for our dependency stack. Axum is past 0.5 version too and upgrading to it requires some heavylifting (I tried and failed spectacularly). 

`rustls` - same story. 0.22 had major [breaking changes](https://github.com/rustls/rustls/releases/tag/v%2F0.22.0) so the longer we want, the harder it gets to upgrade.

Thing that causes me the most pain is `axum-server`. It pins pretty much every HTTP(S) dependency but moves slower. So it is currently blocking `rustls` upgrade.

Ideally I would minimize the number of crates we use for our web stack. Hyper, Tower and rustls should be sufficient to do what we do (unless I am missing something). But at absolute minimum we should move to Hyper 1.0+ and upgrade other dependencies.

cberkhoff:
private-attribution/ipa#1095 Closes this issue

PR #1025, opened 2024-04-26
Using Rustls 0.23

cberkhoff:
This is another step towards updating all of IPA's dependencies (tracked in private-attribution/ipa#924). This change bumps Rustls to the latest version. To make it work I had to fork `hyper-rustls` with some minimal changes so that all versions in the build remain compatible.

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -116,6 +116,8 @@ hyper = { version = "0.14.28", optional = true, features = [
     "stream",
     "runtime",
 ] }
+# The following is a temporary version until we can stabilize the build on a higher version
+# of axum, rustls and the http stack.
 hyper-rustls = { git = "https://github.com/cberkhoff/hyper-rustls/", branch = "rustls-0.23", version = "0.25.1", optional = true, features = ["http2"] }
 iai = { version = "0.1.1", optional = true }
 metrics = "0.21.0"
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -81,7 +81,7 @@ async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
 axum = { version = "0.5.17", optional = true, features = ["http2"] }
 # The following is a temporary version until we can stabilize the build on a higher version
 # of axum, rustls and the http stack.
-axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.22", version = "0.5.2", optional = true, features = [
+axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
     "tls-rustls",
 ] }
 base64 = { version = "0.21.2", optional = true }
@@ -116,7 +116,7 @@ hyper = { version = "0.14.28", optional = true, features = [
     "stream",
     "runtime",
 ] }
-hyper-rustls = { version = "0.25.0", optional = true, features = ["http2"] }
+hyper-rustls = { git = "https://github.com/cberkhoff/hyper-rustls/", branch = "rustls-0.23", version = "0.25.1", optional = true, features = ["http2"] }
 iai = { version = "0.1.1", optional = true }
 metrics = "0.21.0"
 metrics-tracing-context = "0.14.0"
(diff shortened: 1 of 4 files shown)


ipa-core/Cargo.toml
```diff
@@ -116,7 +116,7 @@ hyper = { version = "0.14.28", optional = true, features = [
     "stream",
     "runtime",
 ] }
-hyper-rustls = { version = "0.25.0", optional = true, features = ["http2"] }
+hyper-rustls = { git = "https://github.com/cberkhoff/hyper-rustls/", branch = "rustls-0.23", version = "0.25.1", optional = true, features = ["http2"] }
```
akoshelev, line 119: I am a bit more nervous about forking `hyper-rustls` than `axum-server`. Is it also a temporary measure to help us get to the end state?

ipa-core/Cargo.toml
```diff
@@ -126,11 +126,8 @@ pin-project = "1.0"
 rand = "0.8"
 rand_core = "0.6"
 rcgen = { version = "0.11.3", optional = true }
-rustls = { version = "0.22.3", optional = true }
+rustls = { version = "0.23", optional = true }
 rustls-pemfile = { version = "2.1.2", optional = true }
-# TODO: https://rustsec.org/advisories/RUSTSEC-2023-0053. tokio-rustls and hyper-rustls need to be upgraded first, before
-# we can remove pinning
-rustls-webpki = "0.102.1"
```
akoshelev, line 133: 👍

ipa-core/src/net/client/mod.rs
```diff
@@ -180,6 +180,7 @@ impl MpcHelperClient {
         peer_config: PeerConfig,
         identity: ClientIdentity,
     ) -> Self {
+        setup_crypto_provider();
```
akoshelev, line 183: I don't think this is the right place to setup crypto provider. The defaults must be setup at the application layer (their [doc](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#using-a-specific-cryptoprovider) suggests that too).

Maybe a good alternative would be to call https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#method.builder_with_provider instead.

ipa-core/Cargo.toml
cberkhoff: Correct, this is just a temporary step. It's just easier and faster for me to fully update the TLS stack and be done with that, even though that's not how hyper versions were released. The change I made on hyper-rustls is very minor, just a version bump on the TLS stack. To give some insights, my update plan for IPA is as follows:

1. Update Rustls to 0.23 - This change.
2. Update to Axum 0.6.
3. Update to Axum 0.7 - This PR will update many dependencies, hence I'm trying to do carve smaller changes (like 1 and 2). With this change we should be able to stop using my custom versions.

I'll add a comment in the code explaining this is just a temporary step.

ipa-core/src/net/mod.rs
```diff
@@ -17,6 +21,19 @@ pub use error::Error;
...
+///
+/// # Panics
+/// If there are issues starting the default crypto provider
+pub fn get_crypto_provider() -> &'static Arc<CryptoProvider> {
+    CRYPTO_INIT.call_once(|| {
+        let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+    });
+    CryptoProvider::get_default().expect("Default Crypto provider should be initialized")
+}
+
```
akoshelev, line 36:
```suggestion
/// Provides access to the default Crypto Provider used to setup Rustls for HTTP.
/// Currently, we use the default Rustls offering which is AWS libcrypto. 
pub fn get_crypto_provider() -> Arc<CryptoProvider> {
    Arc::new(rustls::crypto::aws_lc_rs::default_provider())
}
```

`install_default` still overrides the default global provider which is not what we want here I think.

ipa-core/src/net/client/mod.rs
```diff
@@ -193,15 +193,17 @@ impl MpcHelperClient {
             };
             (
                 HttpsConnectorBuilder::new()
-                    .with_native_roots()
+                    .with_provider_and_native_roots(CRYPTO_PROVIDER.as_ref().clone())
```
akoshelev, line 196: that's unfortunate - seems like an inconsistent API from rustls where some methods take `CryptoProvider` and some `Arc<CryptoProvider>` . Former immediately convert them to `Arc`, so really no reason why it shouldn't be `Arc<CryptoProvider>`

ipa-core/src/net/mod.rs
```diff
@@ -17,6 +22,10 @@ pub use error::Error;
 pub use server::{MpcHelperServer, TracingSpanMaker};
 pub use transport::{HttpShardTransport, HttpTransport};
 
+/// Provides access to IPAs Crypto Provider (AWS Libcrypto).
+static CRYPTO_PROVIDER: Lazy<Arc<CryptoProvider>> =
```
akoshelev, line 26: something to consider here - I think the cost of maintaining `Lazy` exceeds the cost of calling `default_provider` more than once. `Lazy` does carry a lot of synchronization required to make it safe to call it from multiple threads while here we are just constructing a simple struct that carries a few fields with it. 

Using Lazy make sense when constructing an object is hard and it overweight the cost of synchronization. Here we could've used a function instead

I don't think it is super important here because we only access this when we construct HTTP client and server and we don't do it often, so I am fine leaving it as is

akoshelev (approved):

PR #1026, opened 2024-04-26
Remove field type parameter from `BooleanProtocols`

andyleiserson:
I don't really know that this is worth taking right now since it will probably conflict with a bunch of other things that are in flight.

This parameter is not completely redundant, in that boolean protocols can be implemented in any prime field. But we only currently use them with `Boolean`.

Our support for other fields is not complete, because we only implement `Not` for `Boolean` (and arrays thereof). `Not` for other fields is `1 - x`, which appears to require `ShareKnownValue`, which requires a context (which doesn't fit with the `Not` trait). But maybe we could define a constant ONE with the value 1/3 (as long as we don't need to support $\mathbb{F}_3$) and implement context-less `Not` in other fields.

This also includes the changes from #1023.


akoshelev:
I think it is fine to remove it - we can add it back if needed

akoshelev (approved):

benjaminsavage:
I'm guessing that this PR is going to conflict massively with the changes made in #1049. Do you want to try to update this PR or just close it?

PR #1028, opened 2024-04-26
Stream multiplexer

akoshelev:
This functionality is required to be able to redistribute the inputs from leader shard to other shards. It works pretty much in the same way as `UnorderedReceiver`, imposing strict order on consumers polling the input stream. The consequence of that is all of them need to be polled in order to make progress. Strict order also gives us the backpressure from slow shards which is good to have in this case

### DIFF
--- a/ipa-core/src/helpers/transport/stream/multiplexer.rs
+++ b/ipa-core/src/helpers/transport/stream/multiplexer.rs
@@ -125,6 +125,8 @@ impl<S: BytesStream> Poller<S> {
             return Poll::Ready(buf.map(Ok));
         }
 
+        // SAFETY: stream is never moved out of this struct. Unsafe code seems to be the only
+        // ergonomic way to get pinned reference from inside a mutex.
         match unsafe { Pin::new_unchecked(&mut inner.stream) }
             .as_mut()
             .poll_next(cx)
--- a/ipa-core/src/helpers/transport/stream/multiplexer.rs
+++ b/ipa-core/src/helpers/transport/stream/multiplexer.rs
@@ -1,0 +1,416 @@
+use std::{
+    collections::VecDeque,
+    io,
+    io::Error,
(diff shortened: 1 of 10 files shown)


ipa-core/src/helpers/transport/stream/multiplexer.rs
```diff
@@ -0,0 +1,417 @@
...
+impl<S: BytesStream> Stream for PollingEnd<S> {
+    type Item = Result<Bytes, MultiplexError>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Short circuit to avoid contention on poller when outer stream is closed.
+        if let Poll::Ready(value) = self.poller.closed_for(self.index) {
+            return Poll::Ready(value);
+        }
+
+        if self.poller.next.load(Ordering::Acquire) == self.index {
```
akoshelev, line 84: I think there is a concurrency bug here (unfortunately shuffle does not seem to be able to catch atomic issues)

* set `next = 0`
* A loads `next` and its turn to poll
* B loads `next` and tries to register waker
* A is done and updates `next` to 1
* B gets its own lock and register wakers
* B exits
* A exits

Now B never gets waken

akoshelev: I was surprised that Shuttle didn't catch that but then I realized that tests don't use any concurrency. I added a few other tests and shuttle immediately found two issues (one that is not mentioned above).

They are now fixed and coverage is also improved

akoshelev:
Synced up with @andyleiserson offline - there may be an easier way to deal with this problem - one that does not require this functionality - I am currently trying to see if it can make `do_query` work for that purpose and leverage `Gateway` API to do the same thing

akoshelev:
Don't need this change

Issue #1031, opened 2024-04-30
Improve DZKP Conversion


danielmasny:
Task: Improve the efficiency of the DZKP Conversion in ipa-core/src/protocol/context/dzkp_field.rs, convert_prover and the corresponding functions of the verifiers. 

The code below does not work, since e.chunks is a slice method, which means it's taking a reference to e, and that reference is only valid until convert_prover returns. Therefore the borrow checker complaints, i.e. error[E0597]: `e` does not live long enough.
```
        // e = ab⊕cd⊕ f = x_left * y_right ⊕ y_left * x_right ⊕ prss_right
        let e = (*x_left & *y_right) ^ (*y_left & *x_right) ^ *prss_right;
        // precompute ab
        let a_times_c = *x_left & *y_left;
        // precompute bd
        let b_times_d = *x_right & *y_right;

        x_left
            .chunks(RECURSION_CHUNK_SIZE)
            .zip(y_right.chunks(RECURSION_CHUNK_SIZE))
            .zip(y_left.chunks(RECURSION_CHUNK_SIZE))
            .zip(x_right.chunks(RECURSION_CHUNK_SIZE))
            .zip(prss_right.chunks(RECURSION_CHUNK_SIZE))
            .zip(e.chunks(RECURSION_CHUNK_SIZE))
            .zip(a_times_c.chunks(RECURSION_CHUNK_SIZE))
            .zip(b_times_d.chunks(RECURSION_CHUNK_SIZE))
            .map(|(((((((a, b), c), d), f), e), ac), bd)| {
                (
                    // g polynomial
                    a.iter()
                        .zip(c.iter())
                        .zip(e.iter())
                        .zip(ac.iter())
                        .map(|(((a, c), e), ac)| {
                            let one_minus_two_e = Fp61BitPrime::ONE
                                + Fp61BitPrime::MINUS_TWO * Fp61BitPrime::truncate_from(*e);
                            [
                                // g1=-2ac(1-2e),
                                Fp61BitPrime::MINUS_TWO
                                    * Fp61BitPrime::truncate_from(*ac)
                                    * one_minus_two_e,
                                // g2=c(1-2e),
                                Fp61BitPrime::truncate_from(*c) * one_minus_two_e,
                                // g3=a(1-2e),
                                Fp61BitPrime::truncate_from(*a) * one_minus_two_e,
                                // g4=-1/2(1-2e),
                                Fp61BitPrime::MINUS_ONE_HALF * one_minus_two_e,
[...]

PR #1032, opened 2024-05-01
Dzkp Conversion

danielmasny:
Conversion from stored multiplication bits into field elements that can be processed by the DZKP


ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+    ) -> impl Iterator<Item = UVPolynomials<Fp61BitPrime>> {
+        x_left
+            .chunks(RECURSION_CHUNK_SIZE)
+            .zip(y_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(y_left.chunks(RECURSION_CHUNK_SIZE))
+            .zip(x_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(prss_right.chunks(RECURSION_CHUNK_SIZE))
+            .map(|((((a, b), c), d), f)| {
+                // e = ab⊕cd⊕ f = x_left * y_right ⊕ y_left * x_right ⊕ prss_right
+                let e: BitArray<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0> = (BitArray::try_from(a)
```
benjaminsavage, line 102: I do not understand why we are taking `RECURSION_CHUNK_SIZE` and shifting it up 3 bits. If all of a, b, c, d, and f are being iterated through in groups of `RECURSION_CHUNK_SIZE`, why is `e` initialised to 8x that size? Or do I just not understand the API for working with BitArray?

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+                    .unwrap()
+                    & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(b).unwrap())
+                    ^ (BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(c).unwrap()
+                        & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(d).unwrap())
+                    ^ BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(f).unwrap();
+                // precompute ac = a & c
+                let ac: BitArray<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0> = BitArray::try_from(a)
+                    .unwrap()
+                    & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(c).unwrap();
+                // precompute bd = b & c
```
benjaminsavage, line 112: comment is wrong.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+    // e = ab⊕cd⊕ f
+    fn convert_prover<'a>(
+        x_left: &'a BitArray32,
+        x_right: &'a BitArray32,
+        y_left: &'a BitArray32,
+        y_right: &'a BitArray32,
+        prss_right: &'a BitArray32,
+    ) -> impl Iterator<Item = UVPolynomials<Fp61BitPrime>> {
+        x_left
+            .chunks(RECURSION_CHUNK_SIZE)
```
benjaminsavage, line 95: I can't seem to import all of these dependencies into the Rust playground to play around with it. When you call:
`x_left.chunks(RECURSION_CHUNK_SIZE)` what is the type of element you receive in the iterator?

Is it a slice of RECURSION_CHUNK_SIZE bit? or is it a slice of RECURSION_CHUNK_SIZE *bytes*?

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+            .zip(y_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(y_left.chunks(RECURSION_CHUNK_SIZE))
+            .zip(x_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(prss_right.chunks(RECURSION_CHUNK_SIZE))
+            .map(|((((a, b), c), d), f)| {
+                // e = ab⊕cd⊕ f = x_left * y_right ⊕ y_left * x_right ⊕ prss_right
+                let e: BitArray<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0> = (BitArray::try_from(a)
+                    .unwrap()
+                    & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(b).unwrap())
+                    ^ (BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(c).unwrap()
```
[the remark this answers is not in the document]
benjaminsavage: Unless I am misunderstanding something, we don't ever need to access them as bits in this part of the code. Couldn't we just cast them to a BitArray after we are done here, and before we need to access the bits?

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+                                // g2=c(1-2e),
+                                Fp61BitPrime::from_bit(*c) * one_minus_two_e,
+                                // g3=a(1-2e),
+                                Fp61BitPrime::from_bit(*a) * one_minus_two_e,
+                                // g4=-1/2(1-2e),
+                                Fp61BitPrime::MINUS_ONE_HALF * one_minus_two_e,
+                            ]
+                        })
+                        .collect::<GenericArray<Fp61BitPrime, InitialRecursionFactor>>(),
+                    // h polynomial
```
benjaminsavage, line 139: I don't understand how `InitialRecursionFactor` came into play here. It seems to me like you're saying the output of this flat_map should have length exactly equal to 32. How do we know that's going to be the case? Each multiplication tuple is generating 8 values (g1, g2, g3, g4, h1, h2, h3, h4). We are Iterating through all of `a` here which is a whatever we get by iterating through `x_left.chunks(RECURSION_CHUNK_SIZE)`. So it has something to do with 8... I'm just lost.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+            .chunks(RECURSION_CHUNK_SIZE)
+            .zip(y_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(prss_right.chunks(RECURSION_CHUNK_SIZE))
+            .zip(z_right.chunks(RECURSION_CHUNK_SIZE))
+            .map(|(((a, c), prss), z)| {
+                // precompute ac = a & c
+                let ac: BitArray<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0> = BitArray::try_from(a)
+                    .unwrap()
+                    & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(c).unwrap();
+                // e = ac ⊕ zright ⊕ prssright
```
benjaminsavage, line 193: This is the first reference to this alternative formulation of `e`. I'd recommend putting the derivation for this up in the comments above with the other one.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+        x_left: &'a BitArray32,
+        y_left: &'a BitArray32,
+        prss_left: &'a BitArray32,
+    ) -> impl Iterator<Item = SingleUVPolynomial<Self>> {
+        y_left
+            .chunks(RECURSION_CHUNK_SIZE)
+            .zip(x_left.chunks(RECURSION_CHUNK_SIZE))
+            .zip(prss_left.chunks(RECURSION_CHUNK_SIZE))
+            .map(|((b, d), f)| {
+                // precompute bd = b & c
```
benjaminsavage, line 244: comment is wrong.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+        let x_right = BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_x_right)).unwrap();
+        let y_left = BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_y_left)).unwrap();
+        let y_right = BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_y_right)).unwrap();
+        let prss_left =
+            BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_prss_left)).unwrap();
+        let prss_right =
+            BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_prss_right)).unwrap();
+        let z_right = BitArray::<[u8; 32]>::try_from(BitSlice::from_slice(&vec_z_right)).unwrap();
+
+        // check correctness of the polynomials
```
benjaminsavage, line 333: This test only tests consistency, not correctness. All we learn from this is that prover and verifier and computing these things in the same way.

It would be good to also have a test for correctness. I'd recommend grabbing some sample values from the excel spreadsheet to make sure they all match up.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -941,22 +944,21 @@ mod tests {
 
             // gen 1024 random values
             for _i in 0..1024 {
+                // we need to set shares left and right identical since it is a test against a parties own shares
```
benjaminsavage, line 947: I'm confused by this. Surely we can just construct 3 batches that correspond to the batches held by each of the 3 parties, and release this constraint that x_left == x_right and so forth.

benjaminsavage (changes requested):

ipa-core/src/ff/prime_field.rs
```diff
@@ -433,6 +433,25 @@ mod fp32bit {
...
+    impl Fp61BitPrime {
+        #[must_use]
+        pub const fn const_truncate(input: u64) -> Self {
+            Self(input % Self::PRIME)
+        }
+
+        #[must_use]
+        pub const fn const_neg(input: Self) -> Self {
+            Self(Self::PRIME - input.0)
+        }
```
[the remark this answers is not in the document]
danielmasny: This is a constant function and needed when declaring constants. Since I am using some field elements like minus 2 or minus one half quite frequently, I don't want to recompute them every time but rather define them as constants

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+        y_left: &'a BitArray32,
+        prss_left: &'a BitArray32,
+    ) -> impl Iterator<Item = SingleUVPolynomial<Self>>;
 }
 
-// TODO(dm) - implement Basefield for Fp61BitPrime in follow up PR
+impl DZKPBaseField for Fp61BitPrime {
+    const INVERSE_OF_TWO: Self = Fp61BitPrime::const_truncate(1_152_921_504_606_846_976u64);
+    const MINUS_ONE_HALF: Self = Fp61BitPrime::const_neg(Self::INVERSE_OF_TWO);
+    const MINUS_TWO: Self = Fp61BitPrime::const_neg(Fp61BitPrime::const_truncate(2));
```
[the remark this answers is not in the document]
danielmasny: I am fine either way, doesn't make a significant difference to me.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: chunks is a function in the bitvec:
https://docs.rs/bitvec/1.0.1/bitvec/slice/struct.BitSlice.html#method.chunks

it iterates over Recursion_chunk_size many bits

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,332 @@ pub trait DZKPCompatibleField: Field {
...
+                let bd: BitArray<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0> = BitArray::try_from(b)
+                    .unwrap()
+                    & BitArray::<[u8; RECURSION_CHUNK_SIZE >> 3], Lsb0>::try_from(d).unwrap();
+                (
+                    // g polynomial
+                    a.iter()
+                        .zip(c.iter())
+                        .zip(e.iter())
+                        .zip(ac.iter())
+                        .flat_map(|(((a, c), e), ac)| {
```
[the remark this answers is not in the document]
danielmasny: flat_map allows me to get an iterator over Fp61BitPrime elements. a, c, e, ac are all bits and I generate 4 field elements from them. I collect 32 field elements (i.e. 8*4 by using flat_map) and collect them in a single polynomial, i.e. a genericarray. 32 is the recursion size of the proof, we can reduce it to 8 or increase it, for now, I just followed the paper.
I want to process them into a format that the prover/verifier easily can use for the lagrange interpolation that they have to do. Flat_map helps me with that. Otherwise I would get interleaved blocks of (g1,g2,g3,g4), (h1,h2,h3,h4) elements that the prover would need to rearrange in memory again. Further, the prover would need to request several elements from the iterator at once. Using flat_map, he can receives whole polynomials p,q with dimension InitialRecursionFactor that he can conveniently process.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: chunks collects 8 bits, for each of the 8 bits I generate a,c,e,ac which define 4 field elements each. So I obtain 32 in total that I collect into a GenericArray.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: I have added a comment on top of the function: 
```
// here e is defined as in the paper (since the the verifier does not have access to b,d,f,
// he cannot use the simplified formula for e)
```

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: which excel spreadsheet are you referring to? How can we be sure that those example values are correct? I will add a test for the prover's bits. Since the provers bit are tested for consistency with the verifiers bits, the correctness of the verifiers bits follows then.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: In the actual protocol, we have 3 different parties. In the actual protocol, HP1 right is the same as HP2 left and so on. Here I basically simulate all 3 parties with a single party that has left and right shares that are identical.

There will be more tests that test the protocol setting, I have a different PR that implements the dzkp_multiplication such that values are collected during multiply (https://github.com/private-attribution/ipa/pull/1034). Once both PRs are shipped, I can add a test that runs a multiplication and then checks the prover and verifiers values for consistency.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: shift by three is just the bit to byte conversion since BitArray wants to have the size in Bytes for the underlying storage container, Recursion_chunk_size is in bits. I will rename it in RECURSION_CHUNK_SIZE_BITS

ipa-core/src/protocol/context/dzkp_field.rs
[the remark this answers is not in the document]
danielmasny: It is a bit annoying but the bitvec crate does not define the Boolean operations ^ & on Bitslices. I need to cast it in a BitArray which is effectively an array of u8s.

I think it makes sense to use BitArrays here since all the Boolean types that we use in the dzkp multiplication are bit array types (including all the vectorized types. Further after computing e, ac, bd, we need to access individual bits, notice that `e.iter()` iterates over the bits of `e`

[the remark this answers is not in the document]
danielmasny: I have already discussed this with Andy last Monday and filed an issue: https://github.com/private-attribution/ipa/issues/1031

I hope we can improve this even further: https://github.com/private-attribution/ipa/issues/1033

I let Andy have a second look on Monday to see whether he can improve this.

ipa-core/src/protocol/context/dzkp_field.rs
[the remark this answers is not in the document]
danielmasny: you actually still need const_truncate_from since truncate_from is not constant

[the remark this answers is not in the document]
danielmasny: I have removed all const functions except const_truncate since it is needed even for hardcoded values. Otherwise I have followed your suggestions.

ipa-core/src/protocol/context/dzkp_validator.rs
benjaminsavage: > Here I basically simulate all 3 parties with a single party that has left and right shares that are identical

I think it's really confusing to the reader and sends confusing messages. I also doesn't test any realistic scenario.

I suggest you do something like this:

```
let mut batch_prover = Batch::new(1);
let mut batch_verifier_left = Batch::new(1);
let mut batch_verifier_right= Batch::new(1);
```

Then generate 3 segments from these values (along with the missing 3rd share of each), then at the end call 

```
batch_prover
            .get_field_values_prover::<Fp61BitPrime>()
            .zip(batch_verifier_left.get_field_values_verifier_left::<Fp61BitPrime>())
            .zip(batch_verifier_right.get_field_values_verifier_right::<Fp61BitPrime>())
            .for_each(|((prover, verifier_left), verifier_right)| {
                assert_eq!(prover.0, verifier_left);
                assert_eq!(prover.1, verifier_right);
            });
```

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +25,466 @@ pub trait DZKPCompatibleField: Field {
...
+                let ac: BitArray<[u8; RECURSION_CHUNK_SIZE_BITS >> 3], Lsb0> =
+                    BitArray::try_from(a).unwrap()
+                        & BitArray::<[u8; RECURSION_CHUNK_SIZE_BITS >> 3], Lsb0>::try_from(c)
+                            .unwrap();
+                // precompute bd = b & d
+                let bd: BitArray<[u8; RECURSION_CHUNK_SIZE_BITS >> 3], Lsb0> =
+                    BitArray::try_from(b).unwrap()
+                        & BitArray::<[u8; RECURSION_CHUNK_SIZE_BITS >> 3], Lsb0>::try_from(d)
+                            .unwrap();
+                (
```
benjaminsavage, line 127: I want to find a better solution that does not involve casing to BitArray. I think this code is just too difficult to read and understand, despite the fact that what's going on is really simple. I'm afraid it's going to make this hard to maintain, and hard for new readers to ramp up on. Let's find a way to just use native types and bitwise operations like '|' and '&'. I think we should consider a different storage type if the current storage type is forcing us to do this.

ipa-core/src/protocol/context/dzkp_field.rs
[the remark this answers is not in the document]
benjaminsavage: I think we need to improve this before we land this PR.

ipa-core/src/protocol/context/dzkp_field.rs
benjaminsavage: Since we do not need to support any arbitrary amount of chunked iteration, and we can easily restrict ourselves to iterating by multiples of 8, or 16, or 32, I don't think we need to use a bitvec. We can just use a vec of u8 or u16 or u32. I don't think we are actually utilizing any of the power of bitvec here. Am I missing something? I just feel like it's making the code messier without any additional upside.

ipa-core/src/protocol/context/dzkp_field.rs
benjaminsavage: I realised there is a problem: The excel spreadsheet cannot work with our 61-bit prime. It doesn't have that level of precision.

benjaminsavage (changes requested):

ipa-core/src/protocol/context/dzkp_field.rs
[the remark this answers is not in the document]
danielmasny: I have changed it. The code is now more readable.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: Yes, I agree that we don't need to support any chunk size. However, for lambda = 8, chunk size would be 2, for lambda = 32, chunk size is 8. We still need to iterate over all the bits, a vec of u8, u16, u32 do not natively support that and I would need to implement my own iterator over bits.

ipa-core/src/protocol/context/dzkp_field.rs
danielmasny: I have changed it so that it is more readable.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -419,16 +442,40 @@ impl UnverifiedValuesStore {
...
+    /// values used by the prover of the DZKPs
+    #[cfg(all(test, unit_test))]
+    fn get_field_values_prover<DF: DZKPBaseField>(
+        &self,
+    ) -> impl Iterator<Item = UVPolynomials<DF>> + '_ {
+        self.vec
+            .iter()
+            .flatten()
+            .flat_map(UnverifiedValues::convert_prover::<DF>)
+    }
```
benjaminsavage, line 455: I think this API is overly complicated. There are nearly identical looking methods on `UnifiedValuesStore` and `Batch`, and I don't know why we need both. Can we eliminate a lot of code by reducing the number of types?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -843,261 +910,150 @@ mod tests {
...
-            }
+    fn generate_batch(check_resizing: bool, segment_size: usize) -> Batch {
+        if check_resizing {
+            let mut batch = Batch::new(1);
+            batch
+                .inner
+                .insert(Gate::default(), UnverifiedValuesStore::new(1));
+            batch.inner.get_mut(&Gate::default()).unwrap().offset = 4;
+            batch
+        } else {
```
benjaminsavage, line 921: `check_resizing` is not defined in any comments. It's also not clear from the code what it means. It's a boolean argument which is the hardest to interpret. This logic is also pretty confusing. This needs to be improved in some fashion.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -843,261 +910,150 @@ mod tests {
...
-                    &vec_prss_left[i * segment_size..(i + 1) * segment_size],
-                );
-                let prss_right = BitVec::<u8>::from_slice(
-                    &vec_prss_right[i * segment_size..(i + 1) * segment_size],
+                // prover
+                // generate segments
+                let segment_prover = Segment::new(
+                    SegmentEntry::new(
+                        &vec_x_left[i * 8 * segment_size..(i + 1) * 8 * segment_size],
+                    ),
```
benjaminsavage, line 978: OK, I think this is another example of the code being too complex with too many different classes. I do not think we need a `Segment` and `SegmentEntry` class. This seems like an area that should be simplified away with fewer types. This instantiation code just points to this very clearly in my mind.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -843,261 +910,150 @@ mod tests {
...
+                ^ vec_prss_right.clone();
+
+            // vector for unchecked elements
+            // i.e. these are used to fill the segments of the verifier and prover that are not part
+            // of this povers proof
+            let vec_z_right = (0..1024).map(|_| rng.gen::<u8>()).collect::<BitVec<u8>>();
+            let vec_x = (0..1024).map(|_| rng.gen::<u8>()).collect::<BitVec<u8>>();
+            let vec_y = (0..1024).map(|_| rng.gen::<u8>()).collect::<BitVec<u8>>();
+            let vec_prss = (0..1024).map(|_| rng.gen::<u8>()).collect::<BitVec<u8>>();
+            let vec_z = (0..1024).map(|_| rng.gen::<u8>()).collect::<BitVec<u8>>();
```
benjaminsavage, line 969: These aren't the best names. These are really confusingly close to the other names. 

Suggestion: either name this something like `x_unused_in_proof` or `x_3rd_share`.

benjaminsavage (changes requested):
@danielmasny - happy to discuss more tomorrow. I'm not happy with this API. I think it's overly complex with too many types. In order to keep this codebase maintainable over the long term I'd like to push for a significant simplification. I think we need a lot fewer types. `Segment`, `SegmentEntry`, `Batch`, `UnverifiedValuesStore`, `UnverifiedValues`, `UVPolynomials`, `SingleUVPolynomial`, `DZKPBaseField`... This is an awful lot of types to support this functionality, and I am confident that we can design a simpler system that doesn't require the code reviewer to learn about and understand the subtle differences between so many classes. Finally, you haven't responded to my comments challenging the use of `BitArray`, `BitArray32` and `BitSlice`. I remain unconvinced of the need for all of this extra machinery. I strongly suspect the code will be easier to read and understand, and actually more efficient as well (through a large reduction in type casting) if we just used raw unsigned integers.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +28,481 @@ pub trait DZKPCompatibleField: Field {
...
+    // notice that the prover is supposed to show that ab⊕cd⊕ f⊕ e =0
+    // therefore e = ab⊕cd⊕ f must hold. (alternatively, you can also see this by substituting z_left,
+    // i.e. z_left = x_left · y_left ⊕ x_left · y_right ⊕ x_right · y_left ⊕ prss_left ⊕ prss_right
+    fn convert_prover<'a>(
+        x_left: &'a BitArray32,
+        x_right: &'a BitArray32,
+        y_left: &'a BitArray32,
+        y_right: &'a BitArray32,
+        prss_right: &'a BitArray32,
+    ) -> impl Iterator<Item = UVPolynomials<Fp61BitPrime>> {
```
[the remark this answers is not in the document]
danielmasny: It does not work that easily, the function returns two polynomials that are interleaved in a single iterator. I already portion them such that the prover can easily process them.

ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -12,17 +28,481 @@ pub trait DZKPCompatibleField: Field {
...
+        x_left
+            .chunks(RECURSION_CHUNK_SIZE_BITS)
+            .map(ToOwned::to_owned)
```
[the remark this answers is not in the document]
danielmasny: We need to iterate over bits, u8 does not support that and we would need to implement our own methods for this. Further, the multiplication operates over bitarrays so it is easy to use bitslices.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: I don't see how to eliminate them, this is also not related to this PR. Batch is a wrapper around a HashMap and plays a similar role as AccumulatorState in the MAC based approach. We use Arc and weak pointer to Batch (and AccumulatorState). The HashMap uses Gates as Keys and under each Gate sits a UnverifierdValuesStore that collects the multiplication bits for that gate. It is important to separate them by gate since we need to have a consistent ordering of the multiplications. Further, different gates might use differently sized bitarray types. So it makes sense to group them by gate.

ipa-core/src/protocol/context/dzkp_validator.rs
danielmasny: Segment and SegmentEntry are simple wrapper around BitSlices. Here we only need to explicitly use them since it is a test. Segment is convenient to use, otherwise I would need to list all 7 values (x_left ,...) as function inputs explicitly.

danielmasny:
> @danielmasny - happy to discuss more tomorrow. I'm not happy with this API. I think it's overly complex with too many types. In order to keep this codebase maintainable over the long term I'd like to push for a significant simplification. I think we need a lot fewer types. `Segment`, `SegmentEntry`, `Batch`, `UnverifiedValuesStore`, `UnverifiedValues`, `UVPolynomials`, `SingleUVPolynomial`, `DZKPBaseField`... This is an awful lot of types to support this functionality, and I am confident that we can design a simpler system that doesn't require the code reviewer to learn about and understand the subtle differences between so many classes. Finally, you haven't responded to my comments challenging the use of `BitArray`, `BitArray32` and `BitSlice`. I remain unconvinced of the need for all of this extra machinery. I strongly suspect the code will be easier to read and understand, and actually more efficient as well (through a large reduction in type casting) if we just used raw unsigned integers.

Actually I did respond to your comments on the BitArray types. You might have not seen it since at least for me, it was under the hidden conversations even though they were not resolved:

> Yes, I agree that we don't need to support any chunk size. However, for lambda = 8, chunk size would be 2, for lambda = 32, chunk size is 8. We still need to iterate over all the bits, a vec of u8, u16, u32 do not natively support that and I would need to implement my own iterator over bits.

benjaminsavage (approved):

PR #1034, opened 2024-05-02
Implement SecureMul for DZKPUpgraded

danielmasny:
SecureMul can now be used for AdditiveShare<F> and an DZKPUpgradedMalicious context when F implements the DZKPCompatible trait. 

The PR also implements the DZKPCompatible trait for Boolean and runs a simple test with a malicious context.

### DIFF
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -1,15 +1,11 @@
 use generic_array::GenericArray;
 use typenum::U1;
 
-use super::Gf32Bit;
 use crate::{
-    ff::{Field, Serializable, U128Conversions},
+    ff::{Field, PrimeField, Serializable, U128Conversions},
     impl_shared_value_common,
     protocol::prss::FromRandomU128,
-    secret_sharing::{
-        replicated::malicious::ExtendableField, Block, FieldVectorizable, SharedValue, StdArray,
-        Vectorizable,
-    },
+    secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
 };
 
 impl Block for bool {
@@ -30,12 +26,9 @@ impl Boolean {
     }
 }
 
-impl ExtendableField for Boolean {
-    type ExtendedField = Gf32Bit;
-
-    fn to_extended(&self) -> Self::ExtendedField {
-        Gf32Bit::try_from(self.as_u128()).unwrap()
-    }
+impl PrimeField for Boolean {
+    type PrimeInteger = u8;
+    const PRIME: Self::PrimeInteger = 2;
 }
 
 impl SharedValue for Boolean {
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -1,24 +1,23 @@
+use bitvec::slice::BitSlice;
 use generic_array::GenericArray;
-use typenum::U1;
 
 use crate::{
     ff::{Field, PrimeField, Serializable, U128Conversions},
     impl_shared_value_common,
-    protocol::prss::FromRandomU128,
+    protocol::{
+        context::{dzkp_field::DZKPCompatibleField, dzkp_validator::SegmentEntry},
+        prss::FromRandomU128,
+    },
     secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
 };
 
-impl Block for bool {
-    type Size = U1;
-}
-
 ///implements shared value framework for bool
 #[derive(Clone, Copy, PartialEq, Debug, Eq)]
-pub struct Boolean(bool);
+pub struct Boolean(u8);
 
 impl Boolean {
-    pub const TRUE: Boolean = Self(true);
-    pub const FALSE: Boolean = Self(false);
+    pub const TRUE: Boolean = Self(1u8);
+    pub const FALSE: Boolean = Self(0u8);
 
     #[must_use]
     pub fn as_u128(&self) -> u128 {
@@ -32,9 +31,9 @@ impl PrimeField for Boolean {
 }
 
 impl SharedValue for Boolean {
-    type Storage = bool;
+    type Storage = u8;
     const BITS: u32 = 1;
-    const ZERO: Self = Self(false);
+    const ZERO: Self = Self(0u8);
 
     impl_shared_value_common!();
 }
@@ -50,6 +49,12 @@ impl FieldVectorizable<1> for Boolean {
 ///conversion to Scalar struct of `curve25519_dalek`
 impl From<Boolean> for bool {
     fn from(s: Boolean) -> Self {
+        s.0 != 0
+    }
+}
+
+impl From<Boolean> for u8 {
+    fn from(s: Boolean) -> Self {
         s.0
     }
 }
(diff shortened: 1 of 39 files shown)


ipa-core/src/ff/boolean.rs
```diff
@@ -190,6 +194,18 @@ impl FromRandomU128 for Boolean {
     }
 }
 
+impl AsRef<u8> for Boolean {
```
akoshelev, line 197: are all the changes for `Boolean` to be able to do this? There may be an easier way to do this

```rust
impl AsRef<u8> for Boolean {
    fn as_ref(&self) -> &u8 {
        match self.0 {
            true => &1,
            false => &0
        }
    }
}
```

ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
```diff
@@ -0,0 +1,186 @@
...
+/// Different to the semi-honest multiply is that it collects intermediate terms in the `dzkp batch`.
+/// These intermediate terms are then verified using a validator that has to be called either explicitly
+/// or implicitly by using `ctx.seq_join`
+///
+/// ## Errors
+/// Lots of things may go wrong here, from timeouts to bad output. They will be signalled
+/// back via the error response
+/// ## Panics
+/// Panics if the mutex is found to be poisoned
+pub async fn multiply<'a, F, const N: usize>(
```
akoshelev, line 33: I don't really like this - it forces us to support two separate code implementations of the multiplication protocol in the long term. They are exactly the same except for collection part and I think it is bad in the long term.

I haven't looked thoroughly into how it can be made generic, but I'd encorage to leverage context customization for this. Can we collect the data we need transparently or with as little augmentation as possible of the generic semi-honest multiplication? Is it possible to ask context to collect prss values for us while doing multiplication?

Things like that would make our codebase much more maintainable

ipa-core/src/ff/boolean.rs
danielmasny: Yes since I need a bitslice and I can use BitSlice::from_element(u8) but BitSlice::from_element(bool) is not supported by the bitvec crate

ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
danielmasny: We already have two different implementations, one for semi honest, one for malicious using MACs. Now, we have a third one that is for malicious with dzkps. 

Collecting the prss values is not enough, we also need to collect in and outputs which are not passed through the context. 

We could implement SecureMul only for upgraded contexts (semi honest and malicious) and then always collect but this would be a major change. We could do this in a separate PR where we depreciate the semi honest multiplication. It would be still not clear how to do multiplications over field elements then since we do not plan to support that when using dzkps. The only way to do that would be to use the MAC based malicious multiply.

akoshelev: > We already have two different implementations, one for semi honest, one for malicious using MACs. Now, we have a third one that is for malicious with dzkps.

My understanding is that malicious MAC-based multiply just calls semi-honest multiply twice, so it is not really an another implementation, but a compose which is fine. 

Collecting results may be tricky but if it is the only thing to do we can think about something. Brainstorming may be a way to go here

ipa-core/src/ff/boolean.rs
akoshelev: I'd just change the AsRef implementation in this case then

ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
danielmasny: Is code redundancy the concern here? Can it be done with a macro? I don't see a way to do it since even though the functionality is similar, it uses different types of contexts as input and depending on the type of context, it would collect the prss, in and output values.

andyleiserson: One strategy might be to create a new semi-honest multiply function that returns the semi-honest result and the intermediates that are used for DZKP (as their native type, not as `Segment`s). Then, turn both the semi-honest multiply and the DZKP malicious multiply into wrappers that call that implementation.

I'm not sure whether that might introduce any code ordering constraints that could impact performance, or have trait bound issues. Even if the DZKP malicious multiply and semi-honest multiply are similar in the abstract, I'd believe that once all the details of supported vectorizations and elemental fields are taken into account, it's hard to actually support them both from the same code.

Given that we also want the MAC malicious multiply to invoke the semi-honest multiplication primitive, trying to make all of these variants supply a context that provides the "store DZKP intermediates" interface seems like it will be quite involved.

akoshelev: I don't want to insist on using the context for this purpose, although I still see it as a convenient entry point for all operations that multiplication requires (mask, send, receive). Vectorization is going to impact the structure of things we store but not the interactions with the context. 

The way I was thinking about it (and please tell me if I am wrong about zkp data required to make it maliciously secure) is to do something along these lines

```rust
async fn proof_based_multiply<C, F>(ctx: C, a: AdditiveShare<F, N>, b: AdditiveShare<F, N>) -> AdditiveShare<F, N> {
  let proof = ctx.new_proof();
  proof.add(a);
  proof.add(b);
  let r = semi_honest_multiply(proof.some_context(), a, b); 
  proof.add(r);
  
  r
}
```

now, the only thing we need to implement is to tie the context to the current proof and make sure send and prss() add things to the proof before returning them. PRSS is very straightforward, send requires an indirection layer, but it does not feel bad. I agree that it is more involved, but performance for actual multiplication part should be at pair with semi-honest counterpart. 

Returning all data required back from multiplication may be a way to start, but I do expect it to perform worse especially at low vectorization factors

danielmasny: I think an easy fix would be to create a sub function for multiply that takes the PRSS shares as input. This function would implement the actual protocol and it is called by the semi honest multiply as well as by the dzkp multiply

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -165,36 +165,86 @@ impl<'a> Segment<'a> {
...
+        self.z_right = z_right;
     }
 }
 
 /// `SegmentEntry` is a simple wrapper to represent one entry of a `Segment`
 /// currently, we only support `BitSlices`
-#[derive(Clone, Debug)]
-pub struct SegmentEntry<'a>(&'a BitSliceType);
+#[derive(Clone, Debug, Default)]
+pub struct SegmentEntry<'a>(Option<&'a BitSliceType>);
```
andyleiserson, line 232: Not sure if the incremental construction will ultimately be necessary (as mentioned in another comment), but if it is, the builder idiom might be useful here. That would work like this: define a `SegmentEntryBuilder` with `Option`s like `SegmentEntry` has here. Implement a method `SegmentEntryBuilder::build(self) -> SegmentEntry` that confirms all of the options are populated (and does the other applicable sanity checks).

ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
```diff
@@ -0,0 +1,154 @@
...
+            .malicious((a, b), |ctx, (a, b)| async move {
+                let validator = ctx.dzkp_validator(10);
+                let mctx = validator.context();
+                let result = a
+                    .multiply(&b, mctx.set_total_records(1), RecordId::from(0))
+                    .await
+                    .unwrap();
+
+                // batch contains elements
+                assert!(mctx.is_verified().is_err());
```
andyleiserson, line 140: Not a big deal as long as there's only one error type that `is_verified` can return, but in general with this sort of check doing something like below can ensure the test is not broken in a way that returns a different error than expected.

```
assert!(matches!(mctx.is_verified(), Err(Error::ContextUnsafe(_))));
```

andyleiserson (changes requested):

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -165,6 +165,10 @@ impl<'a> Segment<'a> {
         }
     }
 
+    /// This function returns the length of the first entry, i.e. `x_left`
+    ///
+    /// The lengths of the entries might be inconsistent, for asserting that the length is identical
+    /// across entries, call function `assert_len`
```
andyleiserson, line 171: I think the comments on lines 170-171, 369, and 404, are no longer applicable in the latest version.

andyleiserson (approved):

PR #1035, opened 2024-05-02
Fix rust 1.78 clippy warnings

andyleiserson:


andyleiserson:
I am still having trouble locally with the sharded shuffle shuttle test, curious if that also shows up in CI.

pre-commit
```diff
@@ -99,7 +99,7 @@ check "Web tests" \
     cargo test -p ipa-core --no-default-features --features "cli web-app real-world-infra test-fixture descriptive-gate"
 
 check "Concurrency tests" \
-    cargo test -p ipa-core --release --features shuttle
+    cargo test -p ipa-core --release --features "shuttle multi-threading"
```
akoshelev, line 102: not sure if this is an accidental change or no - I don't have a preference which test to run as long as we run both on CI

akoshelev (approved):

pre-commit
andyleiserson: This was intentional, trying to make pre-commit closer to what CI is running.

Issue #1036, opened 2024-05-02
Sharded shuffle concurrency test failures with rust 1.78


andyleiserson:
I see this failure consistently on my local machine (aarch64-apple-darwin) after upgrading to rust 1.78, but not in CI (e.g. https://github.com/private-attribution/ipa/actions/runs/8929174792/job/24526364966?pr=1035).

Command: `cargo test -p ipa-core --release --features "shuttle multi-threading" -- shuffle::sharded::tests::empty` (`non_empty_input` also fails)

Output:
```
---- protocol::ipa_prf::shuffle::sharded::tests::empty stdout ----
TestWorld random seed 9634094103382193149
2024-05-02T19:42:39.728593Z  INFO {h=A}: ipa_core::helpers::transport::in_memory::sharding: new
2024-05-02T19:42:39.728612Z  INFO {h=A}:transport_loop{id=ShardIndex(0)}: ipa_core::helpers::transport::in_memory::transport: new
2024-05-02T19:42:39.728625Z  INFO {h=B}: ipa_core::helpers::transport::in_memory::sharding: new
2024-05-02T19:42:39.728627Z  INFO {h=B}:transport_loop{id=ShardIndex(0)}: ipa_core::helpers::transport::in_memory::transport: new
2024-05-02T19:42:39.728633Z  INFO {h=C}: ipa_core::helpers::transport::in_memory::sharding: new
2024-05-02T19:42:39.728636Z  INFO {h=C}:transport_loop{id=ShardIndex(0)}: ipa_core::helpers::transport::in_memory::transport: new
2024-05-02T19:42:39.728646Z  INFO {h=A}:transport_loop{id=ShardIndex(0)}: ipa_core::helpers::transport::in_memory::transport: close time.busy=3.08µs time.idle=31.9µs
2024-05-02T19:42:39.728651Z  INFO {h=A}: ipa_core::helpers::transport::in_memory::sharding: close time.busy=21.5µs time.idle=38.1µs
2024-05-02T19:42:39.728957Z  INFO transport_loop{id=A}: ipa_core::helpers::transport::in_memory::transport: new
2024-05-02T19:42:39.728965Z  INFO {h=B}:transport_loop{id=ShardIndex(0)}: ipa_core::helpers::transport::in_memory::transport: close time.busy=583ns time.idle=337µs
[...]

PR #1037, opened 2024-05-03
Remove will_wake calls

akoshelev:
We used `will_wake` in correctness checks to make sure we don't hang the future/stream forever if a wrong waker is used, but the contract for this function is best-effort and it started generating more and more false positives.

The recent one, found by @aleiserson, is inside rust-lang (https://github.com/rust-lang/rust/pull/119863/), so it is very hard to get around.

We don't have a good replacement for this check unless we implement our own waker, but it is probably too much for now

### DIFF
--- a/ipa-core/src/helpers/buffers/ordering_sender.rs
+++ b/ipa-core/src/helpers/buffers/ordering_sender.rs
@@ -182,7 +182,6 @@ impl WaitingShard {
             match self.wakers[j].i.cmp(&i) {
                 Ordering::Greater => (),
                 Ordering::Equal => {
-                    assert!(item.w.will_wake(&self.wakers[j].w));
                     self.wakers[j] = item;
                     return Ok(());
                 }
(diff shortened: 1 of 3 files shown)


martinthomson (approved):
That's unfortunate.  I guess the real expectation is that we maintain multiple wakers. That seems like a language feature that is really a bug: in that it imposes (unreasonable) costs on those using the API.

akoshelev:
> That's unfortunate. I guess the real expectation is that we maintain multiple wakers. That seems like a language feature that is really a bug: in that it imposes (unreasonable) costs on those using the API.

yea I am not sure I understand the motivation for making it best-effort, probably some hard constraint on the scheduling side. Or it was never designed to avoid false negatives, they were really only interested in true positives.

akoshelev:
related to #1036

andyleiserson (approved):
There is also a comment around line 58 of ordering_sender.rs "here used to be a check that new waker will wake the same task", that can probably be deleted. (This PR and the linked issue seem like sufficient documentation of the demise of `will_wake` checks.)

akoshelev:
> There is also a comment around line 58 of ordering_sender.rs "here used to be a check that new waker will wake the same task", that can probably be deleted. (This PR and the linked issue seem like sufficient documentation of the demise of `will_wake` checks.)

yea I was contemplating about it and was really on the fence about that one. Now the balance is changed so I'll remove it

PR #1038, opened 2024-05-03
Vectorized quicksort

andyleiserson:


danielmasny (approved):
It looks good to me. In the long term, we do not need quicksort since it is not needed in all the on-device attribution proposals.

PR #1039, opened 2024-05-03
Move shard send and recv channel methods to `ShardedContext` trait

akoshelev:
There should be no reason to make them accessible via `Context` - only contexts that are created in the presence of multiple shards should provide the ability to open and close these channels.

This should improve the type safety for sharding functionality as it won't be possible to see those methods on non-sharded executions

### DIFF
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -266,7 +266,10 @@ mod tests {
     use crate::{
         ff::{boolean_array::BA3, Fp31, Fp32BitPrime, Gf2, U128Conversions},
         helpers::{Direction, GatewayConfig, MpcMessage, Role, SendingEnd},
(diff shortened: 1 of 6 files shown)


danielmasny (approved):
Makes sense to me. I also got confused about the shard send channels. Should we do something similar with message? It seems that MPCmessage took the role of the former message while message could now contain sensitive information since it is sent between shards.

akoshelev:
> Makes sense to me. I also got confused about the shard send channels. Should we do something similar with message? It seems that MPCmessage took the role of the former message while message could now contain sensitive information since it is sent between shards.

One is the subset of another. Everything can be a `Message`, i.e. sent across shards, but only individual field values can be sent via MPC channels. `MpcMessage` is much more restrictive and requires an explicit opt-in. At this moment, it is already enforced that MPC channels require `M` to be `MpcMessage` and shard channels are fine sending `M: Message`

PR #1040, opened 2024-05-03
Sum To Zero Check Protocol

danielmasny:
Include Sum To Zero Check Protocol since it is needed for DZKP verification.

It was originally a part of https://github.com/private-attribution/ipa/pull/930

It can also be used to check the consistency between shares, i.e. as a base protocol for https://github.com/private-attribution/ipa/pull/936

I moved Hashing to Prss, since it is similar (basic cryptographic operation). The only difference is that it is not keyed.

### DIFF
--- a/ipa-core/src/protocol/basics/share_validation.rs
+++ b/ipa-core/src/protocol/basics/share_validation.rs
@@ -1,0 +1,137 @@
+use futures_util::future::try_join;
+
+use crate::{
+    error::Error,
+    helpers::{Direction, MpcReceivingEnd, SendingEnd},
+    protocol::{
+        context::Context,
+        ipa_prf::malicious_security::hashing::{compute_hash, Hash},
+        RecordId,
+    },
+    secret_sharing::SharedValue,
+};
+
+/// This function checks that a vector of shares are consistent across helpers
+/// i.e. `H1` holds `(x0,x1)`, `H2` holds `(x1,x2)`, `H3` holds `(x2,x0)`
+/// the function verifies that `H1.x0 == H3.x0`, `H1.x1 == H2.x1`, `H2.x2 == H3.x2`
+///
+/// We use a hash based approach that is secure in the random oracle model
+/// further, only one of left and right helper check that it is zero
+/// this is might not be sufficient in some applications to prevent malicious behavior
+///
+/// The left helper simply hashes the vector and sends it to the right,
+/// the right helper negates his vector, hashes it and compares it to the received hash
+///
+/// # Errors
+/// propagates errors from send and receive
+pub async fn validate_replicated_shares<C, S>(
+    ctx: C,
+    input_left: &[S],
+    input_right: &[S],
+) -> Result<(), Error>
+where
+    C: Context,
+    S: SharedValue,
+{
+    // compute hash of `left.neg` and `right`
+    let hash_left = compute_hash(input_left);
+
+    // set up context
+    let ctx_new = &(ctx.set_total_records(1usize));
+    // set up channels
+    let send_channel: &SendingEnd<_, Hash> =
+        &ctx_new.send_channel(ctx.role().peer(Direction::Right));
+    let receive_channel: &MpcReceivingEnd<Hash> =
+        &ctx_new.recv_channel(ctx.role().peer(Direction::Left));
+
+    let ((), hash_right) = try_join(
+        // send hash
+        send_channel.send(RecordId::from(0usize), compute_hash(input_right)),
+        receive_channel.receive(RecordId::from(0usize)),
+    )
+    .await?;
+
+    debug_assert_eq!(hash_left, hash_right);
+
+    if hash_left == hash_right {
+        Ok(())
+    } else {
+        Err(Error::InconsistentShares)
+    }
+}
+
+/// This function is similar to validate the consistency of shares with the difference
+/// that it validates that the shares sum to zero rather than being identical
+/// i.e. `H1` holds `(x0,x1)`, `H2` holds `(x1,x2)`, `H3` holds `(x2,x0)`
+/// the function verifies that `H1.x0 == -H3.x0`, `H1.x1 == -H2.x1`, `H2.x2 == -H3.x2`
+///
+/// We use a hash based approach that is secure in the random oracle model
+/// further, only one of left and right helper check that it is zero
+/// this is sufficient for Distributed Zero Knowledge Proofs
+/// but might not be sufficient for other applications
+///
+/// The left helper simply hashes the vector and sends it to the right,
+/// the right helper negates his vector, hashes it and compares it to the received hash
+///
+/// # Errors
+/// propagates errors from `validate_replicated_shares`
+pub async fn validate_sum_to_zero<C, S>(
+    ctx: C,
+    input_left: &[S],
+    input_right: &[S],
+) -> Result<(), Error>
+where
+    C: Context,
+    S: SharedValue,
+{
+    validate_replicated_shares(
+        ctx,
+        input_left,
+        &input_right.iter().map(|x| x.neg()).collect::<Vec<_>>(),
+    )
+    .await
+}
+
+#[cfg(all(test, unit_test))]
+mod test {
+    use std::ops::Neg;
+
+    use rand::{thread_rng, Rng};
+
+    use crate::{
+        ff::Fp61BitPrime,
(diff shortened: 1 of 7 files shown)


ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,137 @@
+use futures_util::future::try_join;
```
akoshelev, line 1: this file is not included in the project, hence no coverage. Lets make sure we build it

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,137 @@
...
+/// this is might not be sufficient in some applications to prevent malicious behavior
+///
+/// The left helper simply hashes the vector and sends it to the right,
+/// the right helper negates his vector, hashes it and compares it to the received hash
+///
+/// # Errors
+/// propagates errors from send and receive
+pub async fn validate_replicated_shares<C, S>(
+    ctx: C,
+    input_left: &[S],
```
akoshelev, line 29: is there a reason why you're not taking `Replicated<S>` instead?

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,137 @@
...
+    validate_replicated_shares(
+        ctx,
+        input_left,
+        &input_right.iter().map(|x| x.neg()).collect::<Vec<_>>(),
+    )
+    .await
+}
+
+#[cfg(all(test, unit_test))]
+mod test {
```
akoshelev, line 96: I'd love to see more testing here - we must ensure we can detect inconsistency and reject bad inputs (misaligned left/right)

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,137 @@
...
+    // by generating a vec of zero values
+    // secret share it
+    // and run the sum to zero protocol
+    #[test]
+    fn validate_sum_to_zero_test() {
+        run(|| async move {
+            let world = TestWorld::default();
+            let mut rng = thread_rng();
+
+            let len: usize = rng.gen::<usize>() % 99usize + 1;
```
akoshelev, line 119: gen_range may be helpful here

akoshelev (changes requested):
> It can also be used to check the consistency between shares, i.e. as a base protocol for https://github.com/private-attribution/ipa/pull/936

I can't see how it is compatible with the protocol described in that PR. This proposes having the entire input ready before validation starts, #936 can work with streams and it makes it more attractive to use because it can be just another stream combinator that is easy to plug in.

ipa-core/src/protocol/basics/share_validation.rs
danielmasny: Yes. This check is for the verification of the proofs. I will have 3 provers and 6 verifiers. The two verifiers for each prover need to perform a zero check. You could considers this a 2 party computation where we have 2 out of 2 non-replicated shares and they want to check whether the shares sum up to zero. 

We could interpret the shares as replicated secret shares. One issue might be that the shares won't be consistent so if we add an automated consistency check for replicated shares, it would fail here.

danielmasny: You're right that https://github.com/private-attribution/ipa/pull/936 is quite different. I am not sure whether we should try to make it compatible. In my use-case, we really don't need a stream since I only need to zero check log N many elements, i.e. the dzkp proof size. But the protocol does a very similar thing, i.e. hash the shares and send them over the network for a consistency check.

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+    // compute hash of `left`
+    let hash_left = compute_hash(input_left);
+
+    // set up context
+    let ctx_new = &(ctx.set_total_records(1usize));
+    // set up channels
+    let send_channel = ctx_new.send_channel::<Hash>(ctx.role().peer(Direction::Right));
+    let receive_channel = ctx_new.recv_channel::<Hash>(ctx.role().peer(Direction::Left));
+
+    let ((), hash_right) = try_join(
```
benjaminsavage, line 46: I would recommend naming this something different. This name is super confusing. Can we call it `hash_received` instead?

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+/// further, only one of left and right helper check that it is zero
+/// this is sufficient for Distributed Zero Knowledge Proofs
+/// but might not be sufficient for other applications
+///
+/// The left helper simply hashes the vector and sends it to the right,
+/// the right helper negates his vector, hashes it and compares it to the received hash
+///
+/// # Errors
+/// propagates errors from `validate_replicated_shares`
+pub async fn validate_sum_to_zero<'a, C, I, S>(
```
benjaminsavage, line 75: I think there is a big impedance mismatch between the name of this protocol, this comment, and how it is going to be used. 

It's easy to generate a secret-sharing of zero that would **fail** this test.

Let's generate the 3 shares like this:

x0 = rand()
x1 = rand()
x2 = PRIME - x0 - x1.

And then distribute them exactly as the comment above says:
`/// i.e. `H1` holds `(x0,x1)`, `H2` holds `(x1,x2)`, `H3` holds `(x2,x0)``

This function `validate_sum_to_zero` would error `Error::InconsistentShares`.

That's super confusing because it DOES "sum to zero", and when you look at the comment, you'd say: "But the shares are like that".

What this function is *actually* doing is checking that we have a sharing of the nature:
/// `H1` holds `(x0,-x1)`, `H2` holds `(x1,-x2)`, `H3` holds `(x2,-x0)`

So I'd suggest changing the comment to say *that* about how the shares are distributed, and then rename it to: `validate_three_two_way_sharings_of_zero` or similar.

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+        Misaligned,
+        Changed,
+    }
+
+    // Test sum to zero
+    // by generating a vec of zero values
+    // secret share it
+    // and run the sum to zero protocol
+    #[test]
+    fn validate_sum_to_zero_test() {
```
benjaminsavage, line 119: After updating the protocol name and comments above, please update this test name and comments to be consistent.

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+        run(|| async move {
+            let world = TestWorld::default();
+            let mut rng = thread_rng();
+
+            let len: usize = rng.gen_range(50..100);
+
+            let mut r = vec![Fp61BitPrime::ZERO; len];
```
benjaminsavage, line 126: I'm 90% certain that as written, this code would *still* pass even if `r` was not zero. I think it doesn't matter what value you put here, all the method is checking is that you have a consistent secret-sharing of any value *if* you flip the sign of your right share.

ipa-core/src/protocol/basics/share_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+    secret_sharing::SharedValue,
+};
+
+/// This function checks that a vector of shares are consistent across helpers
+/// i.e. `H1` holds `(x0,x1)`, `H2` holds `(x1,x2)`, `H3` holds `(x2,x0)`
+/// the function verifies that `H1.x0 == H3.x0`, `H1.x1 == H2.x1`, `H2.x2 == H3.x2`
+///
+/// We use a hash based approach that is secure in the random oracle model
+/// further, only one of left and right helper check that it is zero
+/// this is might not be sufficient in some applications to prevent malicious behavior
```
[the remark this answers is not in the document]
danielmasny: I don't understand where the typo is

ipa-core/src/protocol/basics/share_validation.rs
danielmasny: The test was already testing it for random values of r. I have changed the initialization of r to make it more clear.

ipa-core/src/protocol/basics/share_validation.rs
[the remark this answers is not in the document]
benjaminsavage: extra "is"

```
/// this might not be sufficient in some applications to prevent malicious behavior
```

benjaminsavage (approved):
Looks good to me!

PR #1041, opened 2024-05-04
minor simplification to logistic regression

benjaminsavage:
It turns out that `evaluate_per_user_attribution_circuit` was returning a vector of rows, where all rows but one were zero. This seems silly. If we are confident that the logic ensures all rows contribute zero except for one, we can just collapse each user's contribution to a single row by adding all of a user's rows together. No need for any communication, since there can never be any carries.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
+++ b/ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
@@ -7,6 +7,7 @@ use ipa_macros::Step;
 use crate::{
     error::{Error, LengthError},
     ff::{boolean::Boolean, CustomArray, Field, U128Conversions},
+    helpers::stream::TryFlattenItersExt,
     protocol::{
         basics::{select, BooleanArrayMul, BooleanProtocols, SecureMul, ShareKnownValue},
         boolean::or::or,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -277,27 +275,27 @@ async fn evaluate_per_user_attribution_circuit<FV>(
...
     assert!(!rows_for_user.is_empty());
     if rows_for_user.len() == 1 {
-        return Ok(Vec::new());
+        return Ok(Replicated::<FV>::ZERO);
     }
     let first_row = &rows_for_user[0];
     let mut prev_row_inputs = initialize_new_device_attribution_variables(first_row);
 
-    let mut output = Vec::with_capacity(rows_for_user.len() - 1);
+    let mut output = Replicated::<FV>::ZERO;
```
akoshelev, line 290: maybe explain why it works in a comment?

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -277,27 +275,27 @@ async fn evaluate_per_user_attribution_circuit<FV>(
...
-) -> Result<Vec<Replicated<FV>>, Error>
+) -> Result<Replicated<FV>, Error>
 where
     FV: SharedValue + CustomArray<Element = Boolean>,
     Replicated<FV>: BooleanArrayMul,
 {
     assert!(!rows_for_user.is_empty());
     if rows_for_user.len() == 1 {
-        return Ok(Vec::new());
+        return Ok(Replicated::<FV>::ZERO);
```
akoshelev, line 285: it is sad that it is not covered by any test

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -277,30 +281,34 @@ async fn evaluate_per_user_attribution_circuit<FV>(
...
-) -> Result<Vec<Replicated<FV>>, Error>
+) -> Result<Option<Replicated<FV>>, Error>
 where
     FV: SharedValue + CustomArray<Element = Boolean>,
     Replicated<FV>: BooleanArrayMul,
 {
     assert!(!rows_for_user.is_empty());
     if rows_for_user.len() == 1 {
-        return Ok(Vec::new());
+        return Ok(None);
```
benjaminsavage, line 291: @akoshelev - I added a user with just one row to the test in order to get code coverage for this line =).

benjaminsavage:
I'm assuming the CI / miri (pull_request) test failure is spurious. It doesn't seem to have anything to do with these changes. @akoshelev ?

PR #1042, opened 2024-05-06
Walr2

benjaminsavage:
Previously, the implementation of the "feature label dot product" only supported 1-bit features. We need to support more precision than that. This PR is intended to migrate to a multi-bit representation that's intended to be used to represent numbers between 0 and 1. For example, if using 8-bit values, they would represent the values 0/256 through 255/256.


ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -94,19 +98,29 @@ impl InputsRequiredFromPrevRow {
...
+                        c,
+                        record_id,
+                        capped_label_ref,
+                        feature,
+                        &Replicated::<FV>::ZERO,
+                    )
+                    .await
+                },
+            ))
+            .await?;
```
andyleiserson, line 116: You may be able to simplify this with the stuff I added in https://github.com/private-attribution/ipa/pull/1046.

andyleiserson:
I'm not sure I understand your intent for `GenericArray` and `M`. Is there some reason `GenericArray` makes sense here, but not for the main protocol?

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -94,19 +97,23 @@ impl InputsRequiredFromPrevRow {
         )
         .await?;
 
-        let capped_attributed_feature_vector = select(
-            ctx.narrow(&Step::ComputedCappedFeatureVector),
+        let condition: Replicated<Boolean, B> = Replicated::<Boolean, B>::expand(&capped_label);
```
benjaminsavage, line 100: @andyleiserson - please help me. I am stuck on this ONE THING. I cannot figure out how to just take a Replicated<Boolean> and expand it into a Replicated<Boolean, B>... like just duplicate the bit B times...

benjaminsavage: This doesn't work either. There is no `left` or `right` on `capped_label`, there is only `left_arr` and `right_arr`

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
```diff
@@ -94,19 +101,23 @@ impl InputsRequiredFromPrevRow {
...
+            .transpose_from(&input_row.feature_vector)
+            .unwrap_infallible();
+        let capped_attributed_feature_vector = bool_and(
+            ctx,
             record_id,
-            &capped_label,
-            &input_row.feature_vector,
-            &Replicated::<FV>::ZERO,
+            &bit_decomposed_output,
+            repeat_n(&condition, input_row.feature_vector.len()),
```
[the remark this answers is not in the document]
benjaminsavage: Whoops... This used to be right, but then I decided to move the `transpose` call up sooner, so that when I called `bool_and` the internal steps that use `BitOpStep` would *actually* be per bit, and not abuse that name to actually be per-bucket. I didn't update this after I made that change.

ipa-core/src/protocol/ipa_prf/prf_sharding/feature_label_dot_product.rs
andyleiserson: Sorry, `left` and `right` are methods on the `ReplicatedSecretSharing` trait, you need to import it.

PR #1043, opened 2024-05-06
Adding tests to verify that encryption of match key happened correctly for ios side of in app browser

richajaindce:


test_data/fbinfra_integration/ios_encrypted_matchkey.txt
richajaindce, line 1: @akoshelev i am storing this blob file in test_data folder. Is this correct place to store it?

akoshelev: I am tempted to use something like https://github.com/rust-lang/cargo/issues/2841#issuecomment-835353439 for large test files. If it is really small, perhaps you could hex-encode it and store it inside the test file

ipa-core/src/report/matchkey.rs
```diff
@@ -0,0 +1,283 @@
...
+        },
+    };
+
+    // For test purposes only
+    /// A binary report as submitted by a report collector, containing encrypted `MatchKeyReport`
+    /// An `EncryptedOprfReport` consists of:
+    ///     `ct_mk`: Enc(`match_key`)
+    ///     associated data of `ct_mk`: `key_id`, `epoch`, `event_type`, `site_domain`,
+    #[derive(Copy, Clone, Eq, PartialEq)]
+    pub struct EncryptedMatchKeyReport<B>
```
akoshelev, line 35: I feel like this is a lot of code just to test the match key decryption. Is it possible to pad the match key encryption with dummy secret shares of ts, event type, etc and use our generic report struct for this purpose?

ipa-core/src/report/matchkey.rs
```diff
@@ -0,0 +1,283 @@
...
+            site_domain: String::from("www.meta.com"),
+        };
+        let result = report.decrypt(&key_registry).unwrap();
+        assert_eq!(expected, result);
+    }
+
+    #[test]
+    fn test_matchkey_decrypt_encrypt() {
+        let mut rng = thread_rng();
+        let pk = hex::decode("92a6fb666c37c008defd74abf3204ebea685742eab8347b08e2f7c759893947a")
```
akoshelev, line 257: maybe you could avoid copy-pasting the same key material and set it up once as a constant for all tests.

but probably if you could use `GenericReport` you don't need to test the encryption roundtrip

test_data/fbinfra_integration/ios_encrypted_matchkey.txt
richajaindce: I would still prefer for it to sit in a separate file as then we can compare this file with what we generate on client side by simply taking a diff. On the iOS side, I am going to dump a file with the contents and leaving an action item to validate the contents each time a change is made in the implementation

ipa-core/src/report/matchkey.rs
andyleiserson: I have not examined them side by side, but this code looks similar to the existing report parsing code. I agree with @akoshelev that it might be better to fill out dummy data to be able to reuse the existing structs. Or, if there is a new report struct that we need for real-world use cases, then maybe non-test code should be added to handle that report format.

ipa-core/src/report/matchkey.rs
andyleiserson, line 1: Given that this module is only for tests, maybe name it `matchkey_test` or something like that? If the necessary APIs are all public, another option is an integration test in `ipa-core/tests`.

ipa-core/src/report/mod.rs
andyleiserson, line 1: I'm not sure how others feel about this, but I'd be in favor of moving away from `mod.rs` files, i.e. not doing this rename.

richajaindce:
Closing this in favor of #1071

PR #1044, opened 2024-05-06
Use each_ref now that it is stable, and other small tweaks

andyleiserson:


benjaminsavage (approved):
This is pretty awesome. Thank you for simplifying and shortening the codebase!

ipa-core/src/helpers/transport/in_memory/mod.rs
```diff
@@ -57,15 +57,7 @@ impl InMemoryMpcNetwork {
     #[allow(clippy::missing_panics_doc)]
     #[must_use]
     pub fn transports(&self) -> [InMemoryTransport<HelperIdentity>; 3] {
-        let transports: [InMemoryTransport<_>; 3] = self
-            .transports
-            .iter()
-            .map(Arc::downgrade)
-            .collect::<Vec<_>>()
-            .try_into()
-            .map_err(|_| "What is dead may never die")
```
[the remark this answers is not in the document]
akoshelev: yea I was expressing my frustration writing this w/o being able to use `each_ref` :)

ipa-core/src/helpers/mod.rs
```diff
@@ -555,6 +555,18 @@ impl From<usize> for TotalRecords {
     }
 }
 
+pub fn zip3<T, U>(a: [T; 3], b: [U; 3]) -> [(T, U); 3] {
```
akoshelev, line 558: we may want to have a module higher up the tree to put these in

akoshelev (approved):

ipa-core/src/helpers/mod.rs
andyleiserson: Do you have a suggestion? There are a bunch of other utilities stashed within helpers that also aren't exclusive to the infrastructure. If not for the disruption of renaming things, I might rename `helpers` to `infra` and then create something like `utils` for the utilities.

akoshelev: yea I agree that these will be useful outside of infra. Renaming `helpers` to `infra` is long overdue and as much as I don't like `utils` name, I don't have a better suggestion for this module so I think `utils` will be good with internal modules like `array`, `collection` , `stream` etc

andyleiserson: Sounds good. I'm not going to move anything beyond the two new functions right now, but we can move more later.

PR #1045, opened 2024-05-06
Fix build with nightly compiler (affects miri)

andyleiserson:


akoshelev (approved):

akoshelev:
thanks for fixing it!

PR #1046, opened 2024-05-06
Vectorized bit-decomposed boolean AND/OR, and saturating addition

andyleiserson:
Implement saturating addition for aggregation, and generally simplify the aggregation code a bit. Reaching the saturation case required a new config for collect_steps; while updating it, I also pruned the existing set of configs. The set of collected steps was and is fragile with respect to changes in the configuration, hopefully we will have a new strategy for compact gate soon.


ipa-core/src/protocol/boolean/and.rs
```diff
@@ -0,0 +1,42 @@
...
+    <BI as IntoIterator>::IntoIter: ExactSizeIterator<Item = &'a AdditiveShare<Boolean, N>> + Send,
+    Boolean: FieldSimd<N>,
+    AdditiveShare<Boolean, N>: SecureMul<C>,
+{
+    let b = b.into_iter();
+    assert_eq!(a.len(), b.len());
+
+    BitDecomposed::try_from(
+        ctx.parallel_join(zip(a.iter(), b).enumerate().map(|(i, (a, b))| {
+            let ctx = ctx.narrow(&BitOpStep::Bit(i));
```
benjaminsavage, line 37: We might need a new name for this step. This isn't really operating on "bits" anymore, this is now operating on vectorized objects.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -116,26 +115,25 @@ where
...
-        let contributions = ctx.parallel_join(futures).await?;
+
+        let contributions = ctx
+            .parallel_join((0..breakdown_count).step_by(step).enumerate().filter_map(
+                |(i, tree_index)| {
+                    let bucket_c = depth_c.narrow(&BucketStep::from(i));
+
+                    let index_contribution = &row_contribution[tree_index];
+
+                    (robust || tree_index + span < breakdown_count).then(|| {
```
benjaminsavage, line 126: I am surprised to see this usage of `.then()`. I thought this was used for chaining two futures together. But this first expression isn't a future, it's just a boolean expression. And we are ignoring the "return value" of this expression (as evidenced by the `||` as the argument to the closure. That seems strange to me. Is this a common usage of `.then()`?

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -116,26 +115,25 @@ where
...
+                        bool_and(
+                            bucket_c,
+                            record_id,
+                            index_contribution,
+                            repeat_n(bit_of_bdkey, index_contribution.len()),
```
benjaminsavage, line 131: OK, this surprised me. Based on the type signature of `bool_and`, I was expecting this argument to be an iterator of `AdditiveShare<Boolean, N>`, but as far as I can tell, `bit_of_bdkey` is just a single bit. We get it from `bd_key.iter()`, so it seems like that would just be one bit. Can you explain to me what I'm missing?

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -247,13 +248,16 @@ where
...
+                            let record_id = RecordId::from(i);
                             if a.len() < usize::try_from(OV::BITS).unwrap() {
+                                // If we have enough output bits, add and keep the carry.
+                                let (mut sum, carry) =
+                                    integer_add::<_, _, _, _, B>(ctx, record_id, &a, &b).await?;
                                 sum.push(carry);
+                                Ok(sum)
+                            } else {
+                                integer_sat_add::<_, B>(ctx, record_id, &a, &b).await
                             }
```
benjaminsavage, line 260: I think this is probably a good change, but could you add a comment to this function explaining this behavior? Also, saturating add requires more communication than a normal add right? So should we consider adding a toggle so that you can turn this off (to save communication cost) if you're confident that you're not going to experience any overflow?

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -50,38 +43,39 @@ where
...
-    x: &AdditiveShare<S>,
-    y: &AdditiveShare<S>,
-) -> Result<AdditiveShare<S>, Error>
+    x: &BitDecomposed<AdditiveShare<Boolean, N>>,
+    y: &BitDecomposed<AdditiveShare<Boolean, N>>,
+) -> Result<BitDecomposed<AdditiveShare<Boolean, N>>, Error>
 where
-    S: SharedValue + CustomArray<Element = Boolean>,
-    AdditiveShare<S>: BooleanArrayMul + std::ops::Not<Output = AdditiveShare<S>>,
+    SH: ShardBinding,
```
benjaminsavage, line 64: What is this?

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -181,13 +175,13 @@ mod test {
             boolean_array::{BA32, BA64},
             U128Conversions,
         },
-        protocol,
         protocol::{
+            self,
```
benjaminsavage, line 179: I do not understand this import.

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -254,9 +254,16 @@ mod test {
                     .await
                     .unwrap()
                 })
-                .await
+                .await;
+
+            let result_refs: [&BitDecomposed<_>; 3] =
+                result_bits.iter().collect::<Vec<_>>().try_into().unwrap();
+
+            let result = result_refs
```
benjaminsavage, line 262: I had to do the same thing. It's annoying that `Reconstruct` is only implemented for `[&BitDecomposed<_>; 3]` and not for `[BitDecomposed<_>; 3]`. Is there any way we can add support for `Reconstruct` to the output of `world.upgraded_semi_honest()` so that we can just call `.reconstruct()` on it directly?

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -202,8 +202,17 @@ pub async fn test_oprf_ipa<F>(
...
+        world.semi_honest(
+            records.into_iter(),
+            |ctx, input_rows: Vec<OPRFIPAInputRow<BA5, BA8, BA20>>| async move {
+                oprf_ipa::<BA5, BA8, BA16, BA20, BA8, 32>(ctx, input_rows, aws)
+                    .await
+                    .unwrap()
+            },
+        )
+    } else {
+        world.semi_honest(
```
benjaminsavage, line 215: Why is this case, of 256, different from the others? There's already a big match statement below, why couldn't this case just be added there?

benjaminsavage (approved):
Looks good. Most of my comments are just for my own benefit, so that I can learn more about how this code works and be a more effective reviewer / maintainer in the future.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
andyleiserson: It is a [relatively newer method added to `bool`](https://doc.rust-lang.org/std/primitive.bool.html#method.then) as a shorthand for `if /* condition */ { Some(/* value */) } else { None }`. It looks like there are a handful of places we use it, and also a handful where we use the related method `bool::then_some`.

The name collision with the method on futures does seem potentially confusing. I'm not opposed to putting this back the way it was (with a temporary `Vec`), or using the `if / else` longhand rather than `then`, if you think one of those is less confusing.

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
andyleiserson: I will add a comment.

The saturating addition protocol is a significant increase in communication over the regular addition, but the cost of saturating addition only enters the picture once the histogram values have grown from the original TV_BITS to the output size OV_BITS, at which point the record count has been reduced by a significant factor. So I'm not sure it's worth the effort to speculatively do something to optimize for this case.

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
andyleiserson: See https://github.com/private-attribution/ipa/blob/5d5549ab4161f754964f34f37655f6dff24a96c3/ipa-core/src/test_fixture/world.rs#L48 -- this is to support distinguishing sharded contexts from non-sharded contexts. This is one of the things that might go away if we made a "every context is sharded" simplification.

ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
andyleiserson: The only thing that changed here is that my IDE or rustfmt combined a separate import of the protocol module by name (`use crate::protocol; use crate::protocol::{foo, bar}`) with imports of specific symbols inside the module (`use crate::protocol::{self, foo, bar}`).

But also, this import is only used as `protocol::RecordId`, which seems unnecessary and inconsistent with how we generally do it, so I'll get rid of it.

ipa-core/src/test_fixture/ipa.rs
andyleiserson: On the other side of the conditional, the credit cap is the only parameter that changes. This one case has a bunch of parameters that differ. I will add a comment noting this.

The purpose of this change was to support collect_steps coverage. Between two possible upcoming changes (new compact gate and runtime parameter selection), it didn't seem important to put a lot of effort into structuring this optimally right now.

ipa-core/src/protocol/boolean/and.rs
andyleiserson: My hope was that in many cases, vectorized protocols could read like unvectorized protocols -- the code says what to do for each record, and under the hood that magically gets repeated for a vector of records.

Do you have a suggestion how to name it that would be clearer, without making things too verbose?

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
andyleiserson: `bit_of_bdkey` is an `AdditiveShare<Boolean, N>`. Like `BitOpStep`, this is describing how the protocol would operate using language reflecting a single record.

Maybe a way of making it clearer without excessive verbosity would be a `v_` prefix or suffix on variables that are carrying vectorized data?

andyleiserson:
Part of why I vectorized the final aggregation over buckets instead of records (which necessitates the special-case aggregation transpose, and somewhat breaks the model of thinking about vectorized protocols as magically repeating their operation over many records) is to keep the runtimes of the end-to-end tests reasonable for CI. I believe this performance issue arises from switching to boolean shares and not from vectorization. In real-world use cases, we could do aggregation vectorized over records, and then do a final pass that iterates explicitly over both records and buckets in the final chunk to produce the output. This last step might take many seconds, but that doesn't matter in the context of real-world query runtimes.

So if we aggressively want to reduce vectorization complexity, and we're willing to take either an increase in runtime or loss in coverage for CI (or spend some time brainstorming and implementing test strategies to strike a compromise between those), we could get rid of bucket vectorization for the final stage.

ipa-core/src/protocol/boolean/and.rs
benjaminsavage: I see what you're saying. Yeah, I guess it is operating on a "bit" in some sense. Yeah, let's keep this name.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
benjaminsavage: No, this seems like a useful and compact way of expressing things. Thanks for educating me on this! I'm sure I'll wind up using it in the future.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
benjaminsavage: Hmm... Perhaps this approach of using language to describe it as if it was not vectorized, when it actually is, is worth reconsidering. Maybe we just lean into vectorisation and re-name methods to make it obvious when vectorisation is at play. I encountered another one of these situations today and it took me a little while to realise how it worked.

I think I like this naming convention idea of `v_` prefixes.

benjaminsavage (approved):

Issue #1047, opened 2024-05-08
Use new Axum State for Transport passing


cberkhoff:
As part of Axum 0.6 a new way of passing objects was added which we could benefit from to pass the transport around the Axum layer. More details here:

https://tokio.rs/blog/2022-11-25-announcing-axum-0-6-0

PR #1048, opened 2024-05-09
Update to Axum  0.6

cberkhoff:
This is another step towards updating all of IPA's dependencies (tracked in private-attribution/ipa#924).

Note Axum 0.6 adds state to extractors, but we aren't using those for now (created private-attribution/ipa#1047 to track this). This version of Axum still uses hyper 0.14.

Also, as part of this update, Axum extractions separate Body operations from Body Parts (headers, query, etc). A nice improvement is that Body cannot be extracted twice thanks to the new API (BodyAlreadyExtracted exception is gone).

Instead of a having a custom Axum extractor per server handler, with this change we're using separate re-usable extractors for each of the HTTP parts reducing the amount of code in http_serde. Following is an example from the create handler. 

Here's how the code used to be:
```Rust
async fn handler(
    transport: Extension<Arc<HttpTransport>>,
    req: http_serde::query::create::Request, // We got rid of the "Request" custom extractor in http_serde
) -> Result<Json<http_serde::query::create::ResponseBody>, Error> {
    let transport = Transport::clone_ref(&*transport);
...
```

With this change it becomes:
```Rust
async fn handler(
    transport: Extension<Arc<HttpTransport>>,
    // req: http_serde::query::create::Request, gone
    QueryConfigQueryParams(query_config): QueryConfigQueryParams, // We use the existing QueryConfigQueryParams extractor
) -> Result<Json<http_serde::query::create::ResponseBody>, Error> {
    let transport = Transport::clone_ref(&*transport);
...
```

Also, created a few helper functions for the tests and also got rid of trait `IntoFailingReq` and just used Rust's `From`

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -37,7 +37,6 @@ web-app = [
     "toml",
     "tower",
     "tower-http",
-    "mime",
 ]
 test-fixture = ["weak-field"]
 # Include observability instruments that detect lack of progress inside MPC. If there is a bug that leads to helper
@@ -85,7 +84,6 @@ axum = { version = "0.6", optional = true, features = ["http2", "macros"] }
 axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
     "tls-rustls",
 ] }
-mime = { version = "0.3", optional = true}
 base64 = { version = "0.21.2", optional = true }
 bitvec = "1.0"
 bytes = "1.4"
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -37,6 +37,7 @@ web-app = [
     "toml",
     "tower",
     "tower-http",
+    "mime",
 ]
 test-fixture = ["weak-field"]
 # Include observability instruments that detect lack of progress inside MPC. If there is a bug that leads to helper
@@ -78,12 +79,13 @@ ipa-macros = { version = "*", path = "../ipa-macros" }
 aes = "0.8.3"
 async-trait = "0.1.79"
 async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
-axum = { version = "0.5.17", optional = true, features = ["http2"] }
+axum = { version = "0.6", optional = true, features = ["http2", "macros"] }
 # The following is a temporary version until we can stabilize the build on a higher version
 # of axum, rustls and the http stack.
 axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
     "tls-rustls",
 ] }
+mime = { version = "0.3", optional = true}
 base64 = { version = "0.21.2", optional = true }
 bitvec = "1.0"
 bytes = "1.4"
--- a/ipa-core/src/net/http_serde.rs
+++ b/ipa-core/src/net/http_serde.rs
@@ -234,7 +234,7 @@ pub mod query {
             helpers::{query::PrepareQuery, RoleAssignment},
             net::{
                 http_serde::query::{QueryConfigQueryParams, BASE_AXUM_PATH},
-                Error,
+                Error, APPLICATION_JSON,
             },
         };
 
@@ -267,7 +267,7 @@ pub mod query {
                 };
                 let body = hyper::Body::from(serde_json::to_string(&body)?);
                 Ok(hyper::Request::post(uri)
-                    .header(CONTENT_TYPE, "application/json")
+                    .header(CONTENT_TYPE, APPLICATION_JSON)
                     .body(body)?)
             }
         }
@@ -286,7 +286,7 @@ pub mod query {
 
         use crate::{
             helpers::query::QueryInput,
-            net::{http_serde::query::BASE_AXUM_PATH, Error},
+            net::{http_serde::query::BASE_AXUM_PATH, Error, APPLICATION_OCTET_STREAM},
         };
 
         #[derive(Debug)]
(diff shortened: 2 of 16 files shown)


ipa-core/src/net/http_serde.rs
```diff
@@ -1,13 +1,22 @@
-// there isn't an easy way to compose const strings at compile time, so we will hard-code
-// everything
-// I'm not sure what the preceding comment was referring to hard coding, but the way
-// to compose const strings at compile time is concat!()
-
+//! This module provides structs and functions to transform requests into
```
akoshelev, line 1: huge +1 for adding module-level documentation

ipa-core/src/net/http_serde.rs
```diff
@@ -294,40 +272,16 @@ pub mod query {
...
-                        roles,
-                    },
-                })
-            }
-        }
-
         #[derive(Serialize, Deserialize)]
-        struct RequestBody {
-            roles: RoleAssignment,
+        pub struct RequestBody {
```
akoshelev, line 276: curious what is causing it to be `pub` now? is it the new Axum requirement?

ipa-core/src/net/server/handlers/echo.rs
```diff
@@ -1,10 +1,29 @@
...
-    Ok(Json(req))
+async fn handler(
+    Query(query_params): Query<HashMap<String, String>>,
+    hyper_headers: HeaderMap,
+) -> Result<Json<http_serde::echo::Request>, Error> {
+    let headers = hyper_headers
+        .iter()
+        .filter_map(|(name, value)| match value.to_str() {
+            Ok(header_value) => Some((name.to_string(), header_value.to_string())),
+            Err(_) => None,
```
akoshelev, line 20: you can also use `ok` method on `Result` if you don't care about errors

```suggestion
        .filter_map(|(name, value)| value.to_str().ok().map(|name| (name.to_string(), value.to_string())))
```

ipa-core/src/net/server/handlers/query/create.rs
```diff
@@ -12,13 +15,10 @@ use crate::{
 /// to the [`HttpTransport`].
 async fn handler(
     transport: Extension<Arc<HttpTransport>>,
-    req: http_serde::query::create::Request,
+    QueryConfigQueryParams(query_config): QueryConfigQueryParams,
```
akoshelev, line 18: this is pretty nice

ipa-core/src/net/server/handlers/query/create.rs
```diff
@@ -48,50 +47,39 @@ mod tests {
...
             .unwrap();
-        let resp = test_server.server.handle_req(req).await;
-
-        let status = resp.status();
-        let body_bytes = hyper::body::to_bytes(resp.into_body()).await.unwrap();
-        let response_str = String::from_utf8(body_bytes.to_vec()).unwrap();
-
-        assert_eq!(StatusCode::OK, status, "Request failed: {}", &response_str);
+        let handler = make_owned_handler(
+            // Do we need these moves? In particular the first one?
```
akoshelev, line 65: yea, we do because this thing needs to be thread-safe and can't have any references to outside.

ipa-core/src/net/server/handlers/query/mod.rs
```diff
@@ -123,24 +124,34 @@ pub mod test_helpers {
...
+        handler: Arc<dyn RequestHandler<Identity = HelperIdentity>>,
+    ) -> Vec<u8> {
+        let test_server = TestServer::builder()
+            .with_request_handler(handler)
+            .build()
+            .await;
+        let resp = test_server.server.handle_req(req).await;
+        let status = resp.status();
+        let body_bytes = hyper::body::to_bytes(resp.into_body()).await.unwrap();
+        assert_eq!(StatusCode::OK, status,);
```
akoshelev, line 151:
```suggestion
        assert_eq!(StatusCode::OK, status);
```

ipa-core/src/net/server/handlers/query/mod.rs
```diff
@@ -123,24 +124,34 @@ pub mod test_helpers {
...
+        body_bytes.to_vec()
     }
+
+    // TODO: refactor shared stuff
```
akoshelev, line 155: do you want to link an issue here?

akoshelev (approved):
This greatly simplifies the handlers code. Thanks for going an extra mile and fixing the documentation and some other parts of code along the way.

ipa-core/Cargo.toml
```diff
@@ -78,12 +79,13 @@ ipa-macros = { version = "*", path = "../ipa-macros" }
...
 async-trait = "0.1.79"
 async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
-axum = { version = "0.5.17", optional = true, features = ["http2"] }
+axum = { version = "0.6", optional = true, features = ["http2", "macros"] }
 # The following is a temporary version until we can stabilize the build on a higher version
 # of axum, rustls and the http stack.
 axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
     "tls-rustls",
 ] }
+mime = { version = "0.3", optional = true}
```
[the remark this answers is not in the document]
cberkhoff: Sounds good

ipa-core/src/net/http_serde.rs
cberkhoff: So, in the past we used to have a custom request extractor for each API, which would in turn refer these (internal to it) Request objects. That extractor used to be `pub`. Since we got rid of those now the handlers need to see these `Requests`.

ipa-core/src/net/server/handlers/echo.rs
cberkhoff: Will use `ok`

PR #1049, opened 2024-05-09
Change some methods that were generic over vectorized/not to only support vectorization

andyleiserson:
More precisely, these methods still support unvectorized operation (a vectorization dimension of 1), but using them in that mode will now use 8x more memory than the previous version, because bits are not packed within bytes.

Network utilization should not be affected, because these protocols are doing bitwise arithmetic, which never had efficient network utilization in the unvectorized case.

In exchange for the performance loss under particular conditions, a bunch of code can be deleted and some of the remaining code can be simplified. However, in the (unvectorized) attribution code, things get slightly more complicated, to unpack bits into and repack bits from BitDecomposed.

I am offering this for discussion as something that simplifies some of the vectorized code.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
@@ -4,7 +4,7 @@ use ipa_macros::Step;
 
 use crate::{
     error::Error,
-    ff::{boolean::Boolean, ArrayAccessRef, ArrayBuild, ArrayBuilder, Field},
+    ff::{boolean::Boolean, ArrayAccessRef, Field},
     helpers::repeat_n,
     protocol::{
         basics::{BooleanProtocols, SecureMul},
@@ -25,20 +25,18 @@ use crate::{
 ///
 /// # Errors
 /// propagates errors from multiply
-pub async fn integer_add<C, F, XS, YS, const N: usize>(
+pub async fn integer_add<C, const N: usize>(
     ctx: C,
     record_id: RecordId,
-    x: &XS,
-    y: &YS,
-) -> Result<(XS, AdditiveShare<F, N>), Error>
+    x: &BitDecomposed<AdditiveShare<Boolean, N>>,
+    y: &BitDecomposed<AdditiveShare<Boolean, N>>,
+) -> Result<(BitDecomposed<AdditiveShare<Boolean, N>>, AdditiveShare<Boolean, N>), Error>
 where
     C: Context,
-    F: Field + FieldSimd<N>,
-    XS: ArrayAccessRef<Element = AdditiveShare<F, N>> + ArrayBuild<Input = AdditiveShare<F, N>>,
-    YS: ArrayAccessRef<Element = AdditiveShare<F, N>>,
-    AdditiveShare<F, N>: BooleanProtocols<C, F, N>,
+    Boolean: FieldSimd<N>,
+    AdditiveShare<Boolean, N>: BooleanProtocols<C, Boolean, N>,
 {
-    let mut carry = AdditiveShare::<F, N>::ZERO;
+    let mut carry = AdditiveShare::ZERO;
     let sum = addition_circuit(ctx, record_id, x, y, &mut carry).await?;
     Ok((sum, carry))
 }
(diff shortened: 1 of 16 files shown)


ipa-core/src/protocol/ipa_prf/boolean_ops/addition_sequential.rs
```diff
@@ -25,20 +25,24 @@ use crate::{
...
-) -> Result<(XS, AdditiveShare<F, N>), Error>
+    x: &BitDecomposed<AdditiveShare<Boolean, N>>,
+    y: &BitDecomposed<AdditiveShare<Boolean, N>>,
+) -> Result<
+    (
+        BitDecomposed<AdditiveShare<Boolean, N>>,
+        AdditiveShare<Boolean, N>,
+    ),
+    Error,
+>
```
benjaminsavage, line 39: Makes sense to me. This is the only way I can see us using this method. Nice to remove some of these extra generics =). Makes it easier to understand how it's used.

ipa-core/src/protocol/ipa_prf/boolean_ops/comparison_and_subtraction_sequential.rs
```diff
@@ -36,21 +36,18 @@ use crate::{
...
-) -> Result<AdditiveShare<F>, Error>
+    x: &BitDecomposed<AdditiveShare<Boolean>>,
+    y: &BitDecomposed<AdditiveShare<Boolean>>,
+) -> Result<AdditiveShare<Boolean>, Error>
 where
     C: Context,
-    F: Field,
-    XS: ArrayAccessRef<Element = AdditiveShare<F>> + ArrayBuild<Input = AdditiveShare<F>>,
-    YS: ArrayAccessRef<Element = AdditiveShare<F>>,
-    AdditiveShare<F>: BooleanProtocols<C, F>,
```
benjaminsavage, line 50: This is a really nice reduction in complexity =). I love it.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -348,35 +339,23 @@ mod tests {
...
-            + Sync
-            + 'static,
-        YS: ArrayAccessRef<Element = AdditiveShare<Boolean, CHUNK>>
-            + ArrayBuild<Input = AdditiveShare<Boolean, CHUNK>>
-            + FromPrss<usize>
-            + Send
-            + Sync
-            + 'static,
-        for<'a> <XS as ArrayAccessRef>::Ref<'a>: Send,
-        for<'a> <YS as ArrayAccessRef>::Ref<'a>: Send,
```
benjaminsavage, line 376: Nice! I love the red lines =)

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -412,14 +391,9 @@ mod tests {
...
                                 let ctx = ctx.clone();
                                 async move {
-                                    let mut match_keys_builder = XS::builder();
-                                    for _ in 0..MK_BITS {
-                                        match_keys_builder
-                                            .push(AdditiveShare::<Boolean, CHUNK>::ZERO);
-                                    }
-                                    let mut match_keys = match_keys_builder.build();
+                                    let mut match_keys = BitDecomposed::new(iter::empty());
                                     match_keys.transpose_from(&chunk).unwrap_infallible();
```
benjaminsavage, line 395: Unrelated to this PR... but just as an aside, I'm not sure why the `transpose_from` API is shaped like this. Why pass in a mutable `BitDecomposed`? In all of the instances I've seen recently, we just initialize an empty `BitDecomposed`.

ipa-core/src/secret_sharing/decomposed.rs
```diff
@@ -107,18 +111,45 @@ impl<S> BitDecomposed<S> {
...
         self.bits.truncate(len);
     }
+
+    pub fn with_capacity(capacity: usize) -> Self {
+        assert!(capacity <= Self::MAX);
+        Self {
+            bits: Vec::with_capacity(capacity),
+        }
+    }
+}
```
benjaminsavage, line 144: Nice, makes sense. Should avoid a bunch of needless malloc calls.

benjaminsavage (approved):
This is a beautiful PR =). I love all the red lines of code. Thanks @andyleiserson !

PR #1050, opened 2024-05-10
Improvment to the DZKP Batch Storage

danielmasny:
The PR removes the option from vec, and changes some of the more complex and expensive functionalities.

It changes the seq_join function on the validator (now validated_seq_join) and separates the chunk_size parameter in two separate parameters: chunk_size is the amount of futures processed in a single chunk (i.e. DZKP Batch) and multiplication_amount which determines how much memory is allocated within a Batch. 

chunk_size is now an input to validated_seq_join whereas multiplication_amount is specified at the time the validator is generated.

The PR does not make changes to the DZKP conversion which is a separate PR which I will also adjust as discussed: https://github.com/private-attribution/ipa/pull/1032

Upon request by Ben Savage and Alex Koshelev, the PR deletes the reallocation ability of the Batch. If the parameter multiplication_amount is set incorrectly, it will cause a panic and the program crashes (whereas previously it would recover). The benefit is a reduction in code complexity and better maintainability.

### DIFF
--- a/ipa-core/src/protocol/context/dzkp_validator.rs
+++ b/ipa-core/src/protocol/context/dzkp_validator.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 use std::{
     collections::HashMap,
     fmt::Debug,
-    iter::repeat,
     sync::{Mutex, Weak},
 };
 
@@ -29,14 +28,15 @@ use crate::{
     },
     seq_join::{seq_join, SeqJoin},
     sharding::ShardBinding,
-    telemetry::metrics::{DZKP_BATCH_REALLOCATION_BACK, DZKP_BATCH_REALLOCATION_FRONT},
+    telemetry::metrics::DZKP_BATCH_UPDATE,
 };
 
 // constants for metrics::increment_counter!
-const RECORD: &str = "record";
-const OFFSET: &str = "offset";
+const ALLOCATED_AMOUNT: &str = "allocated amount of multiplications";
+const ACTUAL_AMOUNT: &str = "actual amount of multiplications";
+const UNIT_SIZE: &str = "bit size of a multiplication";
 
-pub type BitArray32 = BitArray<[u8; 32], Lsb0>;
+pub type Array256Bit = BitArray<[u8; 32], Lsb0>;
 
 type BitSliceType = BitSlice<u8, Lsb0>;
 
@@ -51,13 +51,13 @@ const BIT_ARRAY_SHIFT: usize = 8;
 /// We do not need to store `zi` since it can be computed from the other stored values.
 #[derive(Clone, Debug)]
 struct UnverifiedValues {
-    x_left: BitArray32,
-    x_right: BitArray32,
-    y_left: BitArray32,
-    y_right: BitArray32,
-    prss_left: BitArray32,
-    prss_right: BitArray32,
-    z_right: BitArray32,
+    x_left: Array256Bit,
+    x_right: Array256Bit,
+    y_left: Array256Bit,
+    y_right: Array256Bit,
+    prss_left: Array256Bit,
+    prss_right: Array256Bit,
+    z_right: Array256Bit,
 }
 
 impl Default for UnverifiedValues {
@@ -165,7 +165,7 @@ impl<'a> Segment<'a> {
         }
     }
 
-    /// This function returns the length of the segment. More specifically it returns the length of
+    /// This function returns the length of the segment in bits. More specifically it returns the length of
     /// the first entry, i.e. `x_left` which is consistent with the length of all other entries.
     #[must_use]
     pub fn len(&self) -> usize {
@@ -189,6 +189,7 @@ impl<'a> SegmentEntry<'a> {
         SegmentEntry(entry)
     }
 
+    /// This function returns the size in bits.
     #[must_use]
     pub fn len(&self) -> usize {
         self.0.len()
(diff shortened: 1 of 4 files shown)


ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -51,13 +51,13 @@ const BIT_ARRAY_SHIFT: usize = 8;
 /// We do not need to store `zi` since it can be computed from the other stored values.
 #[derive(Clone, Debug)]
 struct UnverifiedValues {
```
benjaminsavage, line 53: Could I propose the name `MultiplicationInputsBlock`

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -201,10 +202,13 @@ impl<'a> SegmentEntry<'a> {
 }
 
 /// `UnverifiedValuesStore` stores a vector of `UnverifiedValues` together with an `offset`.
-/// `offset` is the minimum `RecordId` for the current batch.
-/// `chunk_size` is the amount of records within a single batch. It is used to estimate the vector length
-/// during the allocation.
-/// `segment_size` is an option and set to `Some` once we receive the first segment and thus know the size.
+/// `first_record_of_batch` is the first `RecordId` for the current batch.
+/// `last_record_of_batch` keeps track of the highest record that has been added to the batch.
+/// `multiplication_amount` is the maximum amount of multiplications performed within a single batch.
```
benjaminsavage, line 207: I'd propose naming this `max_multiplications_per_batch`

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -201,10 +202,13 @@ impl<'a> SegmentEntry<'a> {
...
+/// It is used to determine the vector length during the allocation.
+/// If there are more multiplications, it will cause a panic!
+/// `multiplication_bit_size` is an option and set to `Some` once we receive the first segment and thus know the size.
```
benjaminsavage, line 210: Why is this? Couldn't the caller provide this value?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -214,146 +218,77 @@ impl<'a> SegmentEntry<'a> {
...
 
-    /// `update` updates the current store to the next offset which is the previous offset plus `vec.actual_len()`.
-    /// It deallocates `vec`. `chunk_size` remains the same.
-    /// `update` does nothing when `self` is empty, i.e. `vec` is empty.
-    ///
-    /// ## Panics
-    /// Panics when `segment_size` is `None`
+    /// `update` updates the current store to the next chunk of records.
+    /// It sets `last_record_of_batch` and `first_record_of_batch` to the record that follows `last_record_of_batch`.
     fn update(&mut self) {
```
benjaminsavage, line 249: I don't think "update" is a very descriptive name for what this method does. This seems more like a "flush" or "reset" function that is only called when a batch is processed. Can you propose a new name?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -214,146 +218,77 @@ impl<'a> SegmentEntry<'a> {
...
-        if i64::from(record_id) < self.offset {
-            // recover from wrong offset, expensive
-            // call to reset_offset is measured using metrics
-            self.reset_offset(i64::from(record_id), segment.len());
+        debug_assert_eq!(segment.len(), self.multiplication_bit_size);
+
+        // update last record
+        if self.last_record_of_batch < record_id {
+            self.last_record_of_batch = record_id;
         }
```
benjaminsavage, line 284:
```
self.last_record_of_batch = cmp::max(self.last_record_of_batch, record_id);
```

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -375,20 +310,19 @@ impl UnverifiedValuesStore {
...
+            (segment.y_left, &mut unverified_values.y_left),
+            (segment.y_right, &mut unverified_values.y_right),
+            (segment.prss_left, &mut unverified_values.prss_left),
+            (segment.prss_right, &mut unverified_values.prss_right),
+            (segment.z_right, &mut unverified_values.z_right),
         ] {
-            // unwrap safe since index are guaranteed to be within bounds
+            // panics when out of bounds
             let values_in_array = array_value
                 .get_mut(position_bit_array..position_bit_array + length)
```
benjaminsavage, line 327: It is unclear from the parameter names what `position_raw` and `length` mean. Please use more descriptive names. Are these positions within the `UnverifiedValues` or within the `Segment` or within the larger vector of `UnverifiedValues`? Length of what? of the segment? If so, why not just call `segment.length()`?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -399,25 +333,20 @@ impl UnverifiedValuesStore {
...
-                    &segment.x_right.0[256 * i..256 * (i + 1)],
-                    &segment.y_left.0[256 * i..256 * (i + 1)],
-                    &segment.y_right.0[256 * i..256 * (i + 1)],
-                    &segment.prss_left.0[256 * i..256 * (i + 1)],
-                    &segment.prss_right.0[256 * i..256 * (i + 1)],
-                    &segment.z_right.0[256 * i..256 * (i + 1)],
-                )
-                .unwrap(),
-            );
+            self.vec[position_vec + i] = UnverifiedValues::new(
```
benjaminsavage, line 340: Wait, why are we calling `UnverifiedValues::new`? Didn't we already pre-allocate all of them and set them to zeroes?

Here's the code above that I have in mind:
```
vec: vec![
    UnverifiedValues::default();
    (amount_of_multiplications * multiplication_bit_size + 255) >> BIT_ARRAY_SHIFT
],
```

It seems like we should not need to do anything but over-write some bits. But I might be missing something or confused.

benjaminsavage (approved):
Thanks for making these changes @danielmasny. I just have a few small naming suggestions, then feel free to land.

PR #1051, opened 2024-05-10
Removing Dead Code

benjaminsavage:
Now that @andyleiserson has improved all of our aggregation protocols to no longer convert bits into prime-fields, we can eliminate all of the old code we used to use to perform this bit conversion.

We also eliminated all references to the code we used to use to decompose prime field values into bits.

There is more code that can be removed related to `multiply_sparse` but that'll require more careful review, so I'll put that in a separate PR so that it's easier to see.


danielmasny (approved):
Thanks for deleting the code. 

The modulus conversion could still be useful for the hybrid proposal and how DP binomial noise is generated in MPC, via sampling a bit, lifting it into a field and adding them but its not clear yet. It is probably easier to reimplement it in case we need it.

akoshelev:
yea we should probably keep modulus conversion

benjaminsavage:
I think @bmcase is planning on using the same boolean addition approach (and Andy's `aggregate` function) that we are using elsewhere for binomial noise. I don't think he is planning on using modulus conversion.

martinthomson:
Do we need to keep the sparse multiplication?

benjaminsavage:
I'm assuming we do *not* need to keep the sparse multiplication, so I have another PR to start removing it: https://github.com/benjaminsavage/ipa-rust/pull/1

PR #1054, opened 2024-05-13
renaming some multiply related functions

benjaminsavage:
Since removing support for "sparse multiply" in #1053, some of these function names do not make sense. Rather than have many functions all named `multiply`, I gave them more descriptive names to help disambiguate.


akoshelev:
It seems to me that the new names expose too much about the internals of the multiplications. Would something like `mac_multiply`, `sh_multiply` and `zkp_multiply` work better here?

ipa-core/src/protocol/basics/mul/mod.rs
```diff
@@ -22,35 +22,18 @@ pub(crate) mod malicious;
...
-        rhs: &Self,
-        ctx: C,
-        record_id: RecordId,
-    ) -> Result<Self, Error>
     where
         C: 'fut;
 }
 
-use semi_honest::multiply as semi_honest_mul;
+use semi_honest::gen_prss_and_multipy as semi_honest_mul;
```
martinthomson, line 36: maybe we can do away with either this line or line 25 (which is conditional, but not really)

martinthomson (commented):
I agree with Alex about names.

danielmasny (approved):
Thanks! The naming makes sense to me and makes the code base cleaner.

PR #1055, opened 2024-05-13
Removing more dead code

benjaminsavage:
None of these protocols are used in our current IPA protocols. I'm a bit on the fence about the "Sum of Products" since I can imagine it potentially being useful again in the future. Thoughts?


martinthomson (approved):
Same as Alex.  Sum of products seems like something we can revive easily enough.

PR #1058, opened 2024-05-13
Limit the number of bits for `bool_and` operation to 9

akoshelev:
Continues the work started in #1056 to reduce the total number of step transitions. Confirmed with @benjaminsavage that in WALR we don't need more than 9 bits and in OPRF IPA it is the same because the number of breakdowns is currently limited to 512


andyleiserson (approved):
I guess for now it's nice to have the 9 bit limit in the name, but as a general practice in naming protocols that doesn't seem like it scales very well. On the other hand, I never liked the `bool_and` name very much, so this offers an excuse to change it if I think of something better.

akoshelev:
> I guess for now it's nice to have the 9 bit limit in the name, but as a general practice in naming protocols that doesn't seem like it scales very well. On the other hand, I never liked the `bool_and` name very much, so this offers an excuse to change it if I think of something better.

yea, for generic MPC I agree. This circuit however is specialized now and I'd love to prevent accidental mistakes, given that 9 bit may be enough to miss the problem in unit tests

PR #1060, opened 2024-05-13
Check num_rows in aggregation

andyleiserson:


benjaminsavage (approved):
Very nice! Thanks for improving this @andyleiserson ! I'm sure this will catch a few cases of misconfigured inputs =).

PR #1062, opened 2024-05-14
Bring back serde(transparent) on RoleAssignment struct

akoshelev:
We used to have this attribute, but it was lost in refactoring sometime in April

This change brings it back and fixes unit tests.


cberkhoff (commented):
Should a test be added for this?

akoshelev:
> Should a test be added for this?

prepare test served a good purpose for that - if this attribute is removed, it will fail.

akoshelev:
I think #1006 removed this attribute but I don't think I know why I did it. I think it was accidental

cberkhoff (approved):

PR #1063, opened 2024-05-14
Narrow the context before calling `integer_add` and `integer_sat_add` inside aggregation

akoshelev:
It is less of an issue for subtraction, but saturated integer addition is used [conditionally](https://github.com/private-attribution/ipa/blob/b56beb5a5a15e320da0c9737fef26b41f4592440/ipa-core/src/protocol/ipa_prf/aggregation/mod.rs#L267) inside PRF IPA and therefore it must use the same set of gates as overflowing addition.


andyleiserson (approved):
I would say that the general contract for protocols is that they can assume the context they are passed is a freshly narrowed context, and that it is a bug in my aggregation protocol that it does not narrow the context separately for invoking saturating and non-saturating addition.

If we changed the aggregation protocol to do that, I don't think it would be necessary to offer this non-obvious and fragile property of the steps in the add/sub implementations.

If we do want to provide the property here, then it would be worth a comment on `AddStep` and `SubStep` explaining it.

Will the step macro support `#[cfg]` attributes on individual steps?

martinthomson (changes requested):
I see that Andy approved this, but the comment suggests that this was the wrong solution.

For me, I agree and would prefer to see unique steps for unique actions, with different narrows invoked by a caller that conditionally invokes code.

martinthomson:
> Will the step macro support #[cfg] attributes on individual steps?

I believe that it does, though I've not tested that.

akoshelev:
I agree that this behaviour is fragile and changing the aggregation protocol is necessary. I am making those changes.

> I would say that the general contract for protocols is that they can assume the context they are passed is a freshly narrowed context

I think the same rule should apply to `addition_circuit` which is a protocol on its own and require a fresh narrow from both `integer_add` and `saturating_add`. I don't see much value making two separate step trees for these - imo it is fine if one uses just `Add` and another one covers both `Add` and `Select`

andyleiserson:
Since integer_add doesn't do anything besides call addition_circuit, it can just pass through the fresh context it got from the caller -- it doesn't need to narrow.

akoshelev:
I see no harm of having the structure proposed originally, but I don't want to push strongly for that. Reverted it

martinthomson (approved):
This is better, yes.

PR #1067, opened 2024-05-14
array access for boolean

bmcase:
small PR to implement ArrayAccess trait for Boolean.  Have run into needing when working on DP noise in MPC but probably not bad to have it generally added.   w/ @akoshelev

### DIFF
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -1,10 +1,11 @@
 use std::iter::Once;
+
 use bitvec::prelude::BitSlice;
 use generic_array::GenericArray;
 use typenum::U1;
 
 use crate::{
-    ff::{Field, PrimeField, Serializable, U128Conversions},
+    ff::{ArrayAccess, Field, PrimeField, Serializable, U128Conversions},
     impl_shared_value_common,
     protocol::{
         context::{dzkp_field::DZKPCompatibleField, dzkp_validator::SegmentEntry},
@@ -12,7 +13,6 @@ use crate::{
     },
     secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
 };
-use crate::ff::ArrayAccess;
 impl Block for bool {
     type Size = U1;
 }
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -234,7 +234,7 @@ mod test {
     use rand::{thread_rng, Rng};
     use typenum::U1;
 
-    use crate::ff::{boolean::Boolean, Serializable};
+    use crate::ff::{ArrayAccess, boolean::Boolean, Serializable};
 
     impl Arbitrary for Boolean {
         type Parameters = <bool as Arbitrary>::Parameters;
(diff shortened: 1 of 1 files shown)


akoshelev:
this change makes sense to me, although I agree with Coverlay - it would be nice to add tests

bmcase:
added tests

ipa-core/src/ff/boolean.rs
```diff
@@ -251,4 +274,26 @@ mod test {
...
+        b.set(0, Boolean::from(false));
+        assert_eq!(b.get(0), Some(Boolean::from(false)));
+
+        // Test iter()
+        let mut count = 0;
+        for elem in b.iter() {
+            count += 1;
+            assert_eq!(elem, Boolean::from(false));
+        }
+        assert_eq!(count, 1);
```
akoshelev, line 297:
```suggestion
        assert_eq(vec![Boolean::from(false)], b.iter().collect::<Vec<_>>());
```

akoshelev (approved):

bmcase:
@akoshelev  finished updating the suggested change.

benjaminsavage (changes requested):
I think @andyleiserson very intentionally did not add support for ArrayAccess. Idea being that people will likely get it wrong and/or do something super intefficient.

Usually, you should be able to use `transpose` to construct these things.

bmcase:
This PR was to part of building DP noise in MPC without vectorization.  But I see that other recent changes have also removed non-vectorized usage for functions like `integer_add` that I was starting to use.  So it may be best to scrap this approach and just start directly on implementing DP noise in MPC with vectorization if that is what all the underlying code is moving towards. 

I'll chat with Andy later and see what he thinks and get some direction to start on the vectorized approach.

ipa-core/src/ff/boolean.rs
```diff
@@ -30,6 +31,28 @@ impl Boolean {
...
+    fn get(&self, index: usize) -> Option<Self::Output> {
+        if index < 1 {
+            Some(*self)
+        } else {
+            None
+        }
+    }
+
+    fn set(&mut self, index: usize, e: Self::Output) {
+        debug_assert!(index < 1);
```
andyleiserson, line 47: In isolation, a debug_assert seems reasonable here. On the other hand, out-of-bounds array indexing will normally panic even in release builds. I'm not sure which way makes more sense here.

andyleiserson (approved):
This seems reasonable to me, as a way of making the boolean array APIs scale down to a width of 1.

> I think @andyleiserson very intentionally did not add support for ArrayAccess.

That was in the context of vectorized shares -- `AdditiveShare<Boolean, N>`. The boolean array types do implement array access, and if you view `Boolean` as `BA1`, then it seems reasonable for it to support the same APIs as the larger arrays.

However, not seeing the broader context where this will be used, I can't say for sure that it seems like the right API for the task.

> I see that other recent changes have also removed non-vectorized usage for functions like integer_add that I was starting to use

It should still be possible to invoke an unvectorized `integer_add`, using `BitDecomposed<AdditiveShare<Boolean>>` rather than `AdditiveShare<BA{n}>`. It will be less efficient, but if efficiency is a dominant concern, then vectorization is they way to go.

benjaminsavage (approved):
If andy is happy then I'm happy :). Shall we merge it?

Issue #1068, opened 2024-05-14
Non Power of 2 BA Types Optimization


danielmasny:
Currently, when using DZKP multiplications, `MultiplicationInputs` are stored in 256 Bit Arrays. 

This does not work well for BA types that have a bitlength that is not a power of two. These types are used in the attribution circuit. The current mitigation is that the length is rounded up when the Multiplications are inserted (protocol/context/dzkp_validator.rs):

```
    fn insert_segment_small(&mut self, record_id: RecordId, segment: Segment) {
        // panics when record_id is less than first_record
        let id_within_batch = usize::from(record_id) - usize::from(self.first_record);
        // round up segment length to a power of two since we want to have divisors of 256
        let length = segment.power_of_two_bit_len();
```

This issue can be closed if we depreciate non power of two type multiplication by e.g. using vectorization or depreciating the on-server attribution. 

Alternatively, we could adjust the Bit Array length at the potential cost of more arithmetics when inserting segments.

Issue #1066, opened 2024-05-14
Bitslice returns size of container


danielmasny:
as_bitslice in boolean_array returns the size os the underlying container, i.e. a multiple of 8 since the container is u8. E.g. in case of BA20 it is 24 bits.
```
                #[inline]
                #[must_use]
                pub fn as_bitslice(&self) -> &BitSlice<u8, Lsb0> {
                    self.0.as_bitslice()
                }
```

similarly in Boolean, from_element returns a bitslice of length 8 rather than 1:
```
impl DZKPCompatibleField for Boolean {
    fn as_segment_entry(array: &<Self as Vectorizable<1>>::Array) -> SegmentEntry<'_> {
        if bool::from(Boolean::from_array(array)) {
            SegmentEntry::from_bitslice(BitSlice::from_element(&1))
        } else {
            SegmentEntry::from_bitslice(BitSlice::from_element(&0))
        }
    }
}
```

PR #1069, opened 2024-05-14
Bitslice Fix

danielmasny:
Fixes Isssue: https://github.com/private-attribution/ipa/issues/1066

The attribution circuit uses e.g. BA3 which have a bitlength that is not power of 2. This PR adds support for that by not optimally filling the `MultiplicationInputBlocks`, i.e. by leaving zeros in between segments. Open Issue: https://github.com/private-attribution/ipa/issues/1068


ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -155,12 +155,9 @@ impl<'a> Segment<'a> {
...
-        // check that length is either multiple of 256 or 256 is multiple of length
+        // check that length is either smaller than 256 or 256 is multiple of length
         debug_assert_eq!(
-            (
-                x_left.len(),
-                x_left.len() % 256 == 0 || 256 % x_left.len() == 0
-            ),
+            (x_left.len(), x_left.len() <= 256 || x_left.len() % 256 == 0),
             (x_left.len(), true)
         );
```
andyleiserson, line 162: I think it makes sense to use the explanatory comment as the assertion message (I'm also correcting "smaller or 256 is a multiple" to "smaller or a multiple of 256" here):
```
    debug_assert!(
        x_left.len() <= 256 || x_left.len() % 256 == 0,
        "length {} should be smaller than 256 or a multiple of 256",
        x_left.len(),
    );
```

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -182,6 +179,15 @@ impl<'a> Segment<'a> {
         self.x_left.len()
     }
 
+    /// This function returns the bit length of the `segment` in bits rounded up to a power of two.
+    ///
+    /// ## Panics
+    /// Panics when conversion from `u32` to `usize` fails.
+    #[must_use]
+    pub fn power_of_two_bit_len(&self) -> usize {
```
andyleiserson, line 187: There is a function [`next_power_of_two`](https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two) on the built-in integer types, I'm not sure the `power_of_two_bit_len` functions are worthwhile over `foo.len().next_power_of_two()`.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -286,25 +309,27 @@ impl MultiplicationInputsBatch {
...
+        if segment.len() <= 256 {
+            self.insert_segment_small(record_id, segment);
         } else {
-            self.insert_segment_large(block_id, &segment);
+            self.insert_segment_large(record_id, &segment);
         }
     }
 
-    /// insert `segments` for `segments` that divide 256
+    /// insert `segments` for `segments` that are smallere than 256
```
andyleiserson, line 320:
```suggestion
    /// Insert a `segment` that is smaller than 256 bits
```

It might be worth a debug assertion to check this precondition (and likewise in `insert_segment_large`).

ipa-core/src/secret_sharing/vector/impls.rs
```diff
@@ -84,37 +84,34 @@ macro_rules! boolean_vector {
...
                     let b = rng.gen::<$vec>();
 
-                    // required by DZKP storage strategy
-                    if (a.as_bitslice().len() % 256 == 0 || 256 % a.as_bitslice().len() == 0) {
-                        let bit_shares = bit.share_with(&mut rng);
-                        let a_shares = a.share_with(&mut rng);
-                        let b_shares = b.share_with(&mut rng);
+                    let bit_shares = bit.share_with(&mut rng);
+                    let a_shares = a.share_with(&mut rng);
+                    let b_shares = b.share_with(&mut rng);
```
andyleiserson, line 89: Is there a reason this test can't use the `TestWorld::semi_honest` helper, which would take care of upgrading the inputs and some of the other boilerplate? (Note that `TestWorld::semi_honest` provides a non-upgraded context, so I think the code to obtain and use a malicious validator should still work.)

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -286,25 +309,27 @@ impl MultiplicationInputsBatch {
...
     ///
     /// ## Panics
     /// Panics when `bit_length` and `block_id` are out of bounds.
-    fn insert_segment_small(&mut self, id_within_batch: usize, block_id: usize, segment: Segment) {
+    fn insert_segment_small(&mut self, record_id: RecordId, segment: Segment) {
+        // panics when record_id is less than first_record
```
andyleiserson, line 325: This should probably go in the panics section of the doc comment (and again, ditto for `insert_segment_large`). It might be worth doing the checked sub explicitly to make the failure cause more apparent and provide some safety in case future code changes move the indexing operation farther away from the subtraction that produces the out of bounds index.

andyleiserson (changes requested):

ipa-core/src/secret_sharing/vector/impls.rs
danielmasny: I have changed it but then decided to change it back since using TestWorld:malcious would require me to implement into shares for the struct (a,b,bit) which creates more boilerplate code. TestWorld:semi_honest does not work since the validator does not do anything for semi honest and upgraded semi honest contexts and I wanted to make sure that the Boolean multiplication adds values to the batch.

andyleiserson: There is an existing `IntoShares` impl for 2-tuples, so you could do `((a, b), bit)`.

> TestWorld:semi_honest does not work since the validator does not do anything for semi honest and upgraded semi honest contexts and I wanted to make sure that the Boolean multiplication adds values to the batch.

I don't understand this. What is different about the contexts that `TestWorld::semi_honest` uses than the ones you get from `world.contexts()`? It seems to me like the malicious test will actually exercise adding values to the batch, and the semi-honest test won't. And I'm not sure how the semi-honest test could add values to the batch, without effectively turning it into the malicious test.

andyleiserson (approved):

PR #1070, opened 2024-05-15
Specializing the BitStep for integer_add and integer_sub

benjaminsavage:
…based on how many are actually needed.

This is my second attempt. @martinthomson says the approach in #1065 will not work in terms of reducing the total number of compact gate steps that will be generated with the new approach. Hopefully this approach works better!

### DIFF
--- a/ipa-core/src/protocol/boolean/or.rs
+++ b/ipa-core/src/protocol/boolean/or.rs
@@ -3,7 +3,7 @@ use std::iter::zip;
 use crate::{
     error::Error,
     ff::{boolean::Boolean, Field},
-    protocol::{basics::SecureMul, context::Context, step::BitOpStep, RecordId},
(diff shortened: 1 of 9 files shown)


akoshelev:
I like this quite a lot, using generic step argument in the addition and subtraction makes it visible to the callee how may bits will be used. One suggestion is to maybe define a new trait for such steps and implement it only for BitStep, so we don't accidentally pass the wrong one that happen to implement `From<usize>`

andyleiserson:
I second @akoshelev's trait suggestion. It also might be useful to define a type alias `DefaultBitOpStep` that can be used for tests where there's no need to be stingy about how many bits we use. (Could also be `TestBitOpStep` or guarded by `#[cfg(test)]` to discourage inappropriate use.)

I am tempted to look for a way to supply the default bit-op step implicitly, but I don't see how to do that without bringing more traits into the picture, and given the issues encountered in #909 and the workarounds that put wrapper functions around traits, I'm not going to advocate for more traits right now.

PR #1071, opened 2024-05-16
Validating the encrypted reports generated on ios can be decrypted in rust

richajaindce:
For now, this process is manual needing copy pasta of the bytes generated as proper integration testing will take more work


ipa-core/src/report.rs
```diff
@@ -631,4 +634,173 @@ mod test {
...
+            &enc_report_bytes1,
+            &enc_report_bytes2,
+            &enc_report_bytes3,
+            &expected,
+        );
+    }
+
+    #[test]
+    fn check_compatibility_conversion_with_ios_encryption() {
+        let enc_report_bytes1 = hex::decode("741cd5012df1cf8f337258066a55c408d1052297af27a35bdef571773215ad7cbd367eab689145a24ad9666a12731a221ff5548cc7591a5ce50da4dcde203cc614175759ef230641adac977187143471b512f1c8fd95eafeb53602d90a69a6411f3af9cb44e02417f6f27b7162f08bff009e82b1c2c2aaaf156f010000007777772e6162632e636f6d").unwrap();
```
andyleiserson, line 781: If you end a line in a string literal with a backslash, rustc removes the backslash and any leading whitespace on the following line. This can be used to embed blobs like this while keeping the line length manageable.

andyleiserson (approved):

PR #1074, opened 2024-05-17
Remove type parameter from `trait UpgradedContext`

andyleiserson:
... and make it an associated type instead.

My intuition for when the type parameter on the trait is needed is when the same trait needs to be implemented multiple times on a particular type, with some aspect of the implementation that varies based on the value of the type parameter. That is not the case for our concrete use case of this trait -- the MAC-based malicious protocol is associated with one particular field and can only work with values of that field.

If many of the places referencing the trait need to specify the value of the type, then it is more compact to specify a type parameter (`Trait<T>`) than an associated type bound (`Trait<Foo = SomeType>`). I applied this rationale for the `SharedValueArray` vectorization-related trait, but I don't think it's relevant here.

I initially made this change before modulus conversion (and a bunch of other IPAv2 protocols) was removed. The modulus conversion stuff had a bunch of places that the `UpgradedContext<Field = ...>` associated type bound was needed.

I expected this change to make other changes I was working on for reveal safety simpler. Given the current (more minimal) state of the code base, I'm not sure this change makes a huge difference. On the other hand, this change makes the MAC malicious `UpgradedContext` trait look more like the DZKP-malicious `DZKPContext` trait (which also doesn't have a type parameter), and that seems like a reasonable justification for it.

### DIFF
--- a/ipa-core/src/protocol/context/malicious.rs
+++ b/ipa-core/src/protocol/context/malicious.rs
@@ -218,18 +218,19 @@ impl<'a, F: ExtendableField> Upgraded<'a, F> {
             NotSharded,
         )
     }
-}
 
-#[async_trait]
-impl<'a, F: ExtendableField> UpgradedContext<F> for Upgraded<'a, F> {
-    type Share = MaliciousReplicated<F>;
-
-    fn share_known_value(&self, value: F) -> MaliciousReplicated<F> {
+    pub fn share_known_value(&self, value: F) -> MaliciousReplicated<F> {
         MaliciousReplicated::new(
             Replicated::share_known_value(&self.clone().base_context(), value),
             &self.inner.r_share * value.to_extended(),
         )
     }
+}
+
+#[async_trait]
+impl<'a, F: ExtendableField> UpgradedContext for Upgraded<'a, F> {
+    type Field = F;
+    type Share = MaliciousReplicated<F>;
 
     async fn upgrade_one(
         &self,
--- a/ipa-core/src/protocol/context/upgrade.rs
+++ b/ipa-core/src/protocol/context/upgrade.rs
@@ -33,12 +33,12 @@ use crate::{
 /// };
 /// // Note: Unbound upgrades only work when testing.
 /// #[cfg(test)]
-/// let _ = <UpgradeContext<C<'_, F>, F, NoRecord> as UpgradeToMalicious<Replicated<F>, _>>::upgrade;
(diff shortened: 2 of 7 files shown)


martinthomson (approved):
There is more to this than moving from `<F>` to `<Field = F>`, but the change to move the `share_known_value` to the upgraded malicious context is also a good one.  I like that we now have fewer lifetime parameters floating around especially.

ipa-core/src/protocol/context/malicious.rs
```diff
@@ -208,18 +208,19 @@ impl<'a, F: ExtendableField> Upgraded<'a, F> {
...
+    pub fn share_known_value(&self, value: F) -> MaliciousReplicated<F> {
         MaliciousReplicated::new(
             Replicated::share_known_value(&self.clone().base_context(), value),
             &self.inner.r_share * value.to_extended(),
         )
     }
+}
+
+#[async_trait]
+impl<'a, F: ExtendableField> UpgradedContext for Upgraded<'a, F> {
```
benjaminsavage, line 221: When I originally introduced this `ExtendableField` thing to the codebase, it was to support malicious security for boolean circuits, by virtue of "extending" them with a GaloisField extension. That concept is now fully deprecated in our codebase, and we are instead relying on DZKPs for malicious security in boolean fields. This makes me wonder if we should just *eliminate* the whole `ExtendableField` concept from our codebase.

It used to denote: "There is some way in which elements of this field can be mapped to elements in a sufficiently large field, that there is enough security". I think for our remaining MAC based approach, there will be no mapping, we will just need to start with a sufficiently large field. If that's the case, consider just killing this thing entirely.

ipa-core/src/protocol/context/upgrade.rs
```diff
@@ -84,36 +76,32 @@ where
...
 {
     async fn upgrade(self, _input: ()) -> Result<(), Error> {
         Ok(())
     }
 }
 
 #[async_trait]
-impl<'a, C, F, T, TM, U, UM> UpgradeToMalicious<'a, (T, U), (TM, UM)>
-    for UpgradeContext<'a, C, F, NoRecord>
+impl<C, T, TM, U, UM> UpgradeToMalicious<(T, U), (TM, UM)> for UpgradeContext<C, NoRecord>
```
benjaminsavage, line 97: Nice! from 7 down to 5 =).

benjaminsavage (approved):
I'm fine with this. I think there might be even more opportunity to simplify this now that we no longer need to "extend" boolean fields.

PR #1078, opened 2024-05-19
Compact gate follow up

akoshelev:
Addresses most of the comments from #961 PR.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/aggregation/step.rs
+++ b/ipa-core/src/protocol/ipa_prf/aggregation/step.rs
@@ -11,7 +11,7 @@ pub(crate) enum AggregationStep {
 #[derive(CompactStep)]
 pub enum BucketStep {
     /// should be equal to `MAX_BREAKDOWNS`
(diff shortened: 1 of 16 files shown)


ipa-core/src/protocol/ipa_prf/aggregation/step.rs
```diff
@@ -9,28 +9,19 @@ pub(crate) enum AggregationStep {
...
-    #[step(count = 512, child = crate::protocol::boolean::step::BoolAndStep)]
-    Bit(usize),
-}
-
-impl From<u32> for BucketStep {
-    fn from(v: u32) -> Self {
-        Self::Bit(usize::try_from(v).unwrap())
-    }
-}
+#[step(count = 512, child = crate::protocol::boolean::step::EightBitStep, name = "b")]
```
benjaminsavage, line 12: I think we should keep the comment about keeping this in sync with MAX_BREAKDOWNS.

benjaminsavage (approved):
LGTM. Maybe just add that one comment back in.

PR #1080, opened 2024-05-20
Small aggregation changes and test speedup

### LINKED ISSUE #1079: Clean up type aliases in `src/protocol/mod.rs`
There are a few type aliases defined in `src/protocol/mod.rs`:

```
pub type MatchKey = Gf40Bit;
pub type BreakdownKey = Gf8Bit;
pub type TriggerValue = Gf3Bit;
pub type Timestamp = Gf20Bit;
```

These are kind of right for OPRF IPA -- except for `MatchKey`, the values at least make sense -- but also kind of wrong -- we use a different breakdown key size in some cases, and they should probably be BA types rather than Gf types.

I think I tried to clean these up and ran into some issues, maybe places where they were used in test code?

andyleiserson:
* Remove some unnecessary `IntoIterator` bounds on `Chunk` methods.
* Add a `Chunk::map` helper to transform chunk data without extracting it from the wrapper. Use this instead of the somewhat confusing `into_raw` method. A side effect of this change is we no longer aggregate extra zero rows in the last block, which improves performance on small inputs.
* To further speed up `protocol::ipa_prf::tests::semi_honest`, switch it to 5-bit breakdown keys and 8-bit histogram values.
* Add a check that the `BK` and `B` parameters are consistent.
* Move type aliases from `protocol` to `cli/playbook/ipa`. Fixes #1079.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
@@ -4,13 +4,13 @@ use std::{
     pin::Pin,
 };
 
-use futures::{Stream, StreamExt, TryStreamExt};
+use futures::{FutureExt, Stream, StreamExt, TryStreamExt};
 
 use crate::{
     error::{Error, LengthError, UnwrapInfallible},
     ff::{boolean::Boolean, CustomArray, U128Conversions},
     helpers::{
-        stream::{process_stream_by_chunks, Chunk, ChunkBuffer, FixedLength, TryFlattenItersExt},
+        stream::{process_stream_by_chunks, ChunkBuffer, FixedLength, TryFlattenItersExt},
         TotalRecords,
     },
     protocol::{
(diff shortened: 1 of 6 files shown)


ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -181,21 +181,25 @@ where
...
+            // transpose within the `Chunk` wrapper using `Chunk::map`, and then invoke
+            // `Chunk::into_iter` via `try_flatten_iters` to produce an unchunked stream of
+            // records, vectorized by buckets.
+            .then(|fut| {
+                fut.map(|res| {
+                    res.map(|chunk| {
+                        chunk.map(|data| Vec::transposed_from(data.as_slice()).unwrap_infallible())
+                    })
+                })
             })
```
benjaminsavage, line 198: Even with the above comment, I'm sure I couldn't tell you the type of `fut`, `res`, `chunk` or `data`. I'm not sure how to fix this. Maybe a comment for each saying what the structure of that data object is?

benjaminsavage (approved):

PR #1083, opened 2024-05-21
Proof protocol

danielmasny:
PR that implements the protocol for generating and distributing the dzkp proofs. 

It does not include verification and only includes some simple tests. I will add extensive tests once the verification is also implemented. All previous tests that check the mathematical components still pass. 

I changed some of the structs in prover.rs to work more smoothly with iterators. Unfortunately, the PR became larger than I initially anticipated.


ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -14,15 +15,64 @@ use crate::{
...
+    <Self as ProofArray>::Length: ArrayLength,
+{
+    pub g: GenericArray<F, <Self as ProofArray>::Length>,
+}
+
+impl<F, R> ProofArray for ZeroKnowledgeProof<F, R>
+where
+    F: PrimeField,
+    R: Add + ArrayLength,
+    <R as Add>::Output: Sub<U1>,
```
akoshelev, line 52: can you explain why you need these bounds in a comment?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
...
+    /// This function allows to set masks to protect sensitive information.
+    /// It is used when the last proof of a batch is generated.
+    ///
+    /// ## Panics
+    /// Panics when `self` contains more than one polynomial tuple
+    /// or when masks cannot be set safely.
+    /// The latter is the case when there are too many points for a polynomial of degree `R minus 1`.
+    /// This function likely panics when used with small fields like `Fp31`
+    /// Use `FP61BitPrime` instead when using this function in test cases.
+    pub fn set_masks(&mut self, p_0: F, q_0: F) {
```
akoshelev, line 156: is there a reason for not using `Result`? That way you could easily test the behaviour and panic in a single place

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
...
+        // and q is the polynomial uv_store.first().1
+        // this is likely to fail for small sized fields,
+        // so dont use Fp31 (but rather FP61BitPrime) in tests for this function.
+        debug_assert_eq!(
+            (self.first().0[R::USIZE - 1], self.first().1[R::USIZE - 1]),
+            (F::ZERO, F::ZERO)
         );
+        // shift 0 positions to the back where no actual point is stored
+        // (F::ZERO are overwhelmingly likely to be just 0 fillers)
+        self.uv[0].0[R::USIZE - 1] = self.uv[0].0[0];
```
akoshelev, line 171: A vector of tuples of arrays is very hard to read.  You are also using `self.first()` and `self.uv[0]` interchangeably which is not helping and they mean the same thing.
Consider adding some API over polynomials to make the code easier to read and maintain.

akoshelev (changes requested):
I would love to polish the API for the prover and verifier before we push this

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
...
+        fn default() -> Self {
+            Self::ZERO
         }
-        ZeroKnowledgeProof::new(proof)
     }
 
-    pub fn compute_final_proof<λ, I, J>(
-        u: I,
-        v: J,
+    // todo: deprecate
```
benjaminsavage, line 268: Why does this need to be deprecated?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
...
         for (i, (x, y)) in zip(p_extrapolated.into_iter(), q_extrapolated.into_iter()).enumerate() {
-            proof[λ::USIZE + 1 + i] += x * y;
+            proof[R::USIZE + 1 + i] += x * y;
         }
 
-        ZeroKnowledgeProof::new(proof)
+        proof
     }
 
-    pub fn gen_challenge_and_recurse<λ, I, J>(
```
benjaminsavage, line 157: It's very hard to review this PR because the corresponding function is not displayed nearby in the diff view.

This would not be a problem if you had cut this PR into a smaller number of more incremental changes, so that one could easily see the side-by-side comparison of how this method looked before and after.

benjaminsavage (changes requested):
@danielmasny - I think this PR needs to be split into a sequence of smaller PRs that make isolated changes. This is too much to review and understand.

Let's start by discussing the API changes in `prover.rs`. There are extensive changes there, and we could make those in isolation. Let's discuss:

1. How prover.rs is currently deficient
2. A new API design for it
3. Then make a PR that just makes those changes.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: final proof is no longer needed. We don't need two different functions for the final and intermediate proofs. I left it there since your tests still use them.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -14,15 +15,64 @@ use crate::{
...
+/// Parameter `R` represents the recursion factor, which is a parameter
+/// used during the proof generation and verification.
+#[derive(Debug)]
+pub struct ZeroKnowledgeProof<F, R>
+where
+    Self: ProofArray,
+    F: PrimeField,
+    <Self as ProofArray>::Length: ArrayLength,
+{
+    pub g: GenericArray<F, <Self as ProofArray>::Length>,
```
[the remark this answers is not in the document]
danielmasny: I think keeping ZeroKnowledgeProof as a struct makes sense. The struct previously existed with a pub g and I mostly simplified the trait bounds by adding a trait ProofArray that uses an associated type ProofArray. I also added the function compute_proof that was previously in a struct ProofGenerator that was mostly a wrapper around a vector of u,v values

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+    ///
+    /// The goal of this function is to generate three proof shares.
+    /// First, it replaces the input proof with the left share of the proof
+    /// for which this helper party is the prover, i.e. `prover_left_proof`.
+    /// Second, it outputs the other share, i.e. `prover_right_proof`.
+    /// Third, the `PRSS` is also used to generate the left share of the proof for which this helper party
+    /// is the verifier, i.e. `verifier_left_proof`. This share is part of the output.
+    ///
+    /// Outputs `(verifier_left_proof, prover_right_proof)`
+    fn share_via_prss<C>(
```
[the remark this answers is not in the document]
danielmasny: Could you elaborate on your comment? Notice that this function generates 3 shares and takes as input a proof. It outputs 2 shares and updates the input proof in place. That is why it uses a mutable input proof.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -14,15 +15,64 @@ use crate::{
...
+/// This struct contains a single `ZeroKnowledgeProof`
+/// The length of the proof is not the generic parameter `R`.
+/// The length can be obtained using the trait `ProofArray`
+/// where the associated type `Length`
+/// represents the length of the proof.
+///
+/// Parameter `R` represents the recursion factor, which is a parameter
+/// used during the proof generation and verification.
+#[derive(Debug)]
+pub struct ZeroKnowledgeProof<F, R>
```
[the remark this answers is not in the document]
danielmasny: It is tied to an internal value, R determines the length of the GenericArray, the length is 2*R-1, the length type is specified in the trait ProofArray.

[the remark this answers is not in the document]
danielmasny: @benjaminsavage If we do that, we have ugly trait bounds everywhere. This was a way to not have to state them everywhere.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
             g: g.into_iter().collect(),
         }
     }
+
+    /// Distributed Zero Knowledge Proofs algorithm drawn from
+    /// `https://eprint.iacr.org/2023/909.pdf`
+    ///
+    /// Uses Andy's borrow trick to work with references and owned values :)
+    pub fn compute_proof<J, B>(
```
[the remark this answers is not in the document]
danielmasny: ProofGenerator was a wrapper around a vec and would generate the proof from the vec. This is not what we need. We need to generate at least the first proof from an iterator since we don't want to store a giant vector of field elements. We want to generate the proofs by lazily generating the field elements.

First I changed ProofGenerator to use an IntoIter. But that also did not make much sense since there were a lot of generics involved that did not make sense when generating the first proof. I think the structure now makes a lot of sense. If you want to have a zero knowledge proof, you call compute_proof on the zero-knowledge proof struct. The default method gives a an all 0 zero knowledge proof which will also verify for all zero field elements, i.e. default field elements.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+    };
+
+    impl Default for Fp61BitPrime {
+        fn default() -> Self {
+            Fp61BitPrime::ZERO
+        }
+    }
+
+    #[test]
+    fn generate_verifier_batch() {
```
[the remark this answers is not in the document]
danielmasny: See PR comment, the verification has not been implemented, only the proof generation.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+                proofs: prover_left_batch,
+            },
+            ProofBatch {
+                proofs: verifier_left_batch,
+            },
+            (left_p_mask, left_q_mask),
+        )
+    }
+
+    /// This function is a helper function that computes the next proof from an iterator
```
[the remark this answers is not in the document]
danielmasny: two functions would be ideally functions of  ZeroKnowledgeProof, i.e. compute_next_proof and share_via_prss. However, they both use Prss and @benjaminsavage intentionally did not want to us PRSS in the malicious security module since it would make that module more involved. 

For the other one, I could convert UVTupleBlock into a struct and define polynomials_from_blocks as an associated function.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+        // This ensures that after the last iteration,
+        // there are at most R::USIZE-1 many real (u,v) field elements
+        // Therefore at least the last field element is a `F:ZERO` filler that can be replaced
+        // Therefore we have space to include the masks during the final proof
+        //
+        // notice that uv_store.len() will always be at least 1
+        // further, polynomial uv_store.uv[0].0 and .1 will always have
+        // R::USIZE many points since they are filled with `F::ZERO`.
+        loop {
+            let stop = uv_store.len() < R::USIZE * (R::USIZE - 1);
```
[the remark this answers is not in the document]
danielmasny: Since they are all constants, isn't that anyway precomputed by the compiler? 

R = 0 does not make sense, R = 1 means that there is 0 compression and by the definition of the zeroknowledge proofs it is supposed to be an infinite loop. I have added an assert that R is larger than 1.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+        // generate masks
+        // The right masks are used during the proof generation.
+        // The left masks are later used during the verification.
+        //
+        // Notice that this helper party's left masks are the helper party on the left's right masks
+        // which it uses for its proof generation
+        // Therefore, when this helper party plays the role of a verifier
+        // (for the proof from the party on the left)
+        // it will need to use the masks.
+        let (left_p_mask, right_p_mask): (F, F) =
```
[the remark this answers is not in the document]
danielmasny: So you want me to create a new context just for generating the two masks via Prss? Thats just two calls to Prss. Further I use the context for the whole proof generation, i.e. secret share all proofs, generate the masks and sending the shares across helpers.

ipa-core/src/protocol/ipa_prf/validation_protocol/mod.rs
```diff
@@ -0,0 +1 @@
+mod proof_generation;
```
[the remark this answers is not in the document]
danielmasny: there is more coming with the verification

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+    /// It generates `prover_left_batch` as its left output
+    /// and `verifier_left_batch` as its right output.
+    /// Further, it outputs the generated masks included in the last proof of the batch.
+    ///
+    /// `PRSS` is invoked since `PRSS` right is used to generate `prover_left_batch`.
+    /// Since `PRSS` cannot be invoked later again on the same `Gate` and `Index`,
+    /// we use `PRSS` left to generate `verifier_left_batch` and output it as well.
+    ///
+    /// Outputs `(prover_left_batch, verifier_left_batch, (p_mask, q_mask))`
+    fn compute_proof_batch<C, I>(ctx: &C, uv_tuple_block: I) -> (Self, Self, (F, F))
```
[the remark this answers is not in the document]
danielmasny: No, because the prover_left_batch needs to be exchanged to get verifier_right_batch. Using VerifierBatch that consists of verifier_left_batch, verifier_right_batch would be confusing here.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,536 @@
...
+///
+/// It uses parameter `R` which is the recursion factor of the zero-knowledge proofs
+/// A zero-knowledge proof with recursion factor `R` has length `2*R::USIZE-1`
+///
+/// The `VerifierBatch` also contains two masks `p_0` and `q_0` in `masks` stored as `(p_0,q_0)`
+/// These masks are used as additional `u,v` values for the final proof
+/// of the `left` batch of `ZeroKnowledgeProofs`.
+/// On the `right` batch of `ZeroKnowledgeProofs` these values are set to `F::ZERO`.
+/// These masks mask sensitive information when the right verifier sends a field value to the left verifier,
+/// i.e. using a channel with `Direction:Right`, who does not have access to these masks.
```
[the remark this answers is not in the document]
danielmasny: Yes, this will happen later during the verification. The **verifier** on the right will send information to the **verifier** on the left using `Direction::Right`. This is why we need the masks. During the proof generation, we use `Direction::Left` to send the proof shares from the **prover** to the verifier on the left.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -31,215 +81,264 @@ where
...
+    /// The function uses `Fiat-Shamir` to generate a challenge `r`
+    /// by hashing the two input proofs, `proof_left` and `proof_right`
+    /// It then evaluates each input polynomial at x coordinate `r`.
+    /// The computed points are used to generate the set of `UVPolynomial<F,M>`.
+    ///
+    /// Uses Andy's borrow trick to work with references and owned values :)
+    pub fn gen_challenge_and_recurse<J, B>(
+        proof_left: &ZeroKnowledgeProof<F, R>,
+        proof_right: &ZeroKnowledgeProof<F, R>,
+        mut uv: J,
```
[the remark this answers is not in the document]
danielmasny: No, because the first proof is not build from an UVStore which is a wrapper around a vector of field elements. The first proof is generated from an iterator over field elements since collecting this iterator in a vector would lead to a gigantic vector. This is also why I use Andy's borrow trick

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: Yes, there is a recursion parameter `R`. From this parameter, you can compute the length of the proof as `2R-1`, `R: Add` means `2R`. Almost any function needs to take recursion parameter `R` so it makes sense to parameterize the proofs also with this parameter rather than the actual length. Using the actual length would lead to a lot of ugly typenum arithemtics in a lot of places. This is why I have changed it.

I use this trait to clean up the trait bounds. Without the ProofArray trait, I would need to include these bounds everywhere I currently include the ProofArray trait bound.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -14,15 +15,64 @@ use crate::{
...
+pub trait ProofArray {
+    type Length: ArrayLength;
+}
+
+/// This struct contains a single `ZeroKnowledgeProof`
+/// The length of the proof is not the generic parameter `R`.
+/// The length can be obtained using the trait `ProofArray`
+/// where the associated type `Length`
+/// represents the length of the proof.
+///
```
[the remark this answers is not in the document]
danielmasny: The issue is that the trait bounds 

```
    R: Add + ArrayLength,
    <R as Add>::Output: Sub<U1>,
    <<R as Add>::Output as Sub<U1>>::Output: ArrayLength,
```

will pop up everywhere, every struct and implementation in the proof_generation.rs as well as in any function in prover.rs and verifier.rs that handles zero knowledge proofs. Further, TwoNMinusOne will pop up everywhere as well.

I fixed this such that you just need to include ZeroKnowledgeProof<F,R>: ProofArray as trait bound which I think is a lot cleaner.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -14,15 +15,64 @@ use crate::{
...
+/// Parameter `R` represents the recursion factor, which is a parameter
+/// used during the proof generation and verification.
```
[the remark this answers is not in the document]
danielmasny: I was planning to change verifier.rs as well as soon as I come to the verification protocol. I changed it to `R` mainly because λ doesn't fit the norm to use capital case letter for generics. I ended up putting 
`#[allow(non_camel_case_types)]` everywhere. λ is also annoying to type.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: I think this is independent of the more complex parts of this PR in proof_generation.rs. 

I think it is a result of that I have made many iterative changes to prover.rs and finally decided to use different structs and associate gen_challenge_and_recurse and compute_proof with different structs.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,539 @@
...
+#[derive(Debug)]
+pub struct VerifierBatch<F, R>
+where
+    F: PrimeField,
+    ZeroKnowledgeProof<F, R>: ProofArray,
+{
+    left: Vec<ZeroKnowledgeProof<F, R>>,
+    right: Vec<ZeroKnowledgeProof<F, R>>,
+    left_masks: (F, F),
```
benjaminsavage, line 46: masks

danielmasny:
Comment from Andy:

> It would be helpful to have a summary of what is changing and why in the PR description.
> 
> If I understand right, `λ` is renamed to `R`, and the `R` type parameter is used directly in places where the parameter N=2λ+1 was used before. How does this simplify the code?
> 
> The `ProofGenerator` type seems to be replaced with `UVStore`. Part of what I understand (based on in-person discussions, as opposed to studying the PR) to be going on here is that the first pass of the proof will work with an iterator over u and v values rather than storing them all in an array, since that is (1) possible for the first pass, and (2) a fully expanded copy of u and v for the first pass would be very large -- a significant multiple larger than the raw multiplication intermediates, at least until we implement packing. Writing the proof generation routines in terms of iterators enables them to work either from an array or iterator of u/v values. Since the name `UVStore` implies a data storage type, I think it would be better not to implement complex prover functionality in methods on the `UVStore` type. Similarly, if we're getting rid of the `ZeroKnowledgeProof` type in favor of using `GenericArray` directly, then we can't implement complex prover functionality as methods on `GenericArray`. Possibly the prover functionality could be implemented as free functions in the `prover` module. I see how the methods are serving as constructors for `UVStore` and `ZeroKnowledgeProof`, but I'm not sure there's a compelling reason for them to be that way. I feel like the logic that structures and terminates the recursion should also be implemented here, rather than left to the caller, but maybe there's a good reason an abstraction around that doesn't work.
> 
> It would be useful to have some additional comments on these routines explaining what they do and how they are to be used. Some amount of the detail is in the paper and when it is, references to the paper are fine, but I don't think it's possible to match the notation in the paper precisely enough that no additional explanation is necessary. It would also be nice if the implementation had enough self-contained documentation that it is possible to use the proof system without referring to the paper. (This concern is not new with the changes in this PR -- I feel the same about the original.)
> 
> Assuming we can align on the changes, is there a reason not to go ahead and update the tests to reflect the revisions (instead of leaving some routines as deprecated but used by tests)?

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        I: Iterator<Item = UVTupleBlock<Fp61BitPrime>> + Clone,
+    {
+        // generate Proof
+        let (prover_left_proof, proof_from_left, (p, q)) =
+            ProofBatch::compute_proof(&ctx, uv_tuple_block);
+
+        // send prover_left_proof and receive proof_from_right
+        let length = prover_left_proof.len();
+        let send = prover_left_proof.send_to_left(&ctx);
+        let receive = ProofBatch::receive_from_right(&ctx, length);
```
benjaminsavage, line 61: Why not put all of these methods on the ProofGenerator / ProofVerifier? I think it would make more sense to call `ProofVerifier.receive_from_right()` than `ProofBatch::receive_from_right()`. A "batch" sounds like a data structure name, not something that should be responsible for sending and receiving data.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+            masks: (p, q),
+        }
+    }
+}
+
+/// This a `ProofBatch` generated by a prover.
+struct ProofBatch {
+    first_proof: [Fp61BitPrime; LargeProofGenerator::PROOF_LENGTH],
+    proofs: Vec<[Fp61BitPrime; SmallProofGenerator::PROOF_LENGTH]>,
+}
```
benjaminsavage, line 79: Are we sure we want to hard-code the assumption that the first iteration is one size while all others are something else? This isn't very flexible.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        Ok(())
+    }
+
+    /// This function receives a `Proof` from the party on the right.
+    async fn receive_from_right<C>(ctx: &C, length: usize) -> Result<Self, Error>
+    where
+        C: Context,
+    {
+        // set up context
+        let ctx_left = ctx.set_total_records(length);
```
benjaminsavage, line 141: I don't think `ctx_left` is a good name for this.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+    ///
+    /// It generates `batch_from_left` as its left output
+    /// and `prover_left_batch` as its right output.
+    /// Further, it outputs the generated masks needed for the verification of proof from the left and right.
+    ///
+    /// Outputs `(batch_from_left, prover_left_batch, (p_mask, q_mask))`
+    ///
+    /// ## Panics
+    /// Panics when the recursion factor is smaller than 2 or masks cannot be set safely (implementation issue).
+    fn compute_proof<C, I>(ctx: &C, uv_tuple_block: I) -> (Self, Self, (Fp61BitPrime, Fp61BitPrime))
```
benjaminsavage, line 169: We can't have `ProofBatch::compute_proof`, `ProofGenerator::compute_proof` and `ProofGenerator::compute_next_proof`.

I think a lot of the logic can move into "prover.rs" and "verifier.rs". If we still feel it is necessary to have another class that does the whole "batch" thing, maybe we need a new name for it (not ProofBatch) and the function should have a more descriptive and unique name.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        // there are at most RECURSION_FACTOR-1 many real (u,v) field elements
+        // Therefore at least the last field element is a `F:ZERO` filler that can be replaced
+        // Therefore we have space to include the masks during the final proof
+        //
+        // we remark that uv_value.len() will always be at least 1
+        // further, polynomial uv_values.uv[0].0 and .1 will always have
+        // RECURSION_FACTOR many points
+        // since they are arrays of fixed length, i.e. RECURSION_FACTOR,
+        // (which are filled with `F::ZERO` when there are not enough elements).
+        let r_minus_one = SmallProofGenerator::RECURSION_FACTOR - 1;
```
benjaminsavage, line 231: `r` is the name of the challenge point. We are not using `r` to indicate the recursion factor. If you want to do it this way, call it `recursion_factor_minus_one` or `lambda_minus_one`

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+            // collect proof
+            proofs_from_left.push(proof_from_left);
+            prover_left_proofs.push(prover_left_proof);
+
+            uv_values = uv_values_new;
+
+            if stop {
+                break uv_values;
+            }
+        };
```
benjaminsavage, line 252: This logic is too subtle and complex. This will make it brittle and hard to maintain, and difficult for readers to understand. We need to break this apart and find a new approach to solve this problem that is implemented in simpler code.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        // precomputation for other proofs
+        // denominator
+        let denominator = CanonicalLagrangeDenominator::<
+            Fp61BitPrime,
+            { SmallProofGenerator::RECURSION_FACTOR },
+        >::new();
+        // lagrange table
+        let lagrange_table = LagrangeTable::<
+            Fp61BitPrime,
+            { SmallProofGenerator::RECURSION_FACTOR },
```
benjaminsavage, line 214: This is basically duplicate code. The const params are also causing the code to wrap and become really unreadable.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+
+            if stop {
+                break uv_values;
+            }
+        };
+
+        // generate masks
+        // verifier on the right has p,
+        // therefore the right share is "implicitly sent" to the right ("communicated" via PRSS)
+        let (p_mask_from_left, p_mask_prover): (Fp61BitPrime, Fp61BitPrime) =
```
benjaminsavage, line 257: suggested names:

`prover_left_mask`, `my_mask_right`

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+            ctx.prss().generate_fields(record_counter);
+        record_counter += 1;
+        // and verifier on the left has q
+        // therefore the left share is "implicitly sent" to the left (communication via PRSS)
+        let (q_mask_prover, q_mask_from_right): (Fp61BitPrime, Fp61BitPrime) =
```
benjaminsavage, line 262: Suggested names:

`my_mask_left`, `prover_right_mask`

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        let (u_values, v_values) = &mut uv_values[0];
+        // check that last elements are really Zero
+        debug_assert_eq!(
+            u_values[SmallProofGenerator::RECURSION_FACTOR - 1],
+            Fp61BitPrime::ZERO
+        );
+        debug_assert_eq!(
+            v_values[SmallProofGenerator::RECURSION_FACTOR - 1],
+            Fp61BitPrime::ZERO
+        );
```
benjaminsavage, line 276: I think if we make a UVStorage struct like I suggested above, it can keep track of its *populated* length (as opposed it it's allocated length), and we can perform these checks more elegantly, and in a more abstracted way.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+        // shift first element to last position
+        u_values[SmallProofGenerator::RECURSION_FACTOR - 1] = u_values[0];
+        v_values[SmallProofGenerator::RECURSION_FACTOR - 1] = v_values[0];
+        // set masks in first position
+        u_values[0] = p_mask_prover;
+        v_values[0] = q_mask_prover;
```
benjaminsavage, line 282: Is this really necessary? Why can't we just append the masks at the end? I think in terms of the interpolated polynomial, either approach should make the value at `r` a randomly distributed number, right?

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
benjaminsavage: Also, this function is way too long. I'm surprised clippy didn't complain about it. We need to shorten this function.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,517 @@
...
+            ctx,
+            &mut record_counter,
+            &lagrange_table,
+            uv_values.iter(),
+        );
+
+        // collect proof
+        proofs_from_left.push(proof_from_left);
+        prover_left_proofs.push(prover_left_proof);
+
```
benjaminsavage, line 299: This is also basically duplicate code. See lines 236 - 245. We need to find a way to avoid this duplication.

benjaminsavage (changes requested):
I think we still need to break this PR into a few smaller pieces. I'd suggest we ignore `proof_generation` for now, and try to start by landing the proposed changes to `prover.rs` - along with the comments there.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: right now, the batch uses two kinds of proofs, generated by different ProofGenerators. I think it is easier to send a batch. I am happy to change the naming from batch to something else. I did not want to call it proof since that would be confusing.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: as mentioned above, the proof batch uses proofs from two different proof generators.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: the duplicate code is resolved. We could shorten the names for the proofgenerator types and associated constants to not cause the code to wrap.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: yes, the 0 coefficient will impact any point on the polynomial, no matter what x is, since its multiplied with x^0, i.e. 1 which is not the case for the other coefficients.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,461 @@
...
+                    (
+                        h,
+                        BatchToVerify::generate_batch_to_verify(ctx, uv_tuple_vec.into_iter())
+                            .await,
+                    )
+                })
+                .await;
+
+            // proof from first party
+            simple_proof_check(result[0].0, &result[2].1, &result[1].1);
```
benjaminsavage, line 367: I think we need a struct here instead of a tuple for this return type. Otherwise it's too hard to remember what .0 and .1 and .2 all refer to.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,461 @@
...
+            // collect proof
+            shares_of_proofs_from_prover_left.push(share_of_proof_from_prover_left);
+            my_proofs_left_shares.push(my_proof_left_share);
+
+            uv_values = uv_values_new;
+        }
+
+        // output proofs
+        // and masks
+        (
```
benjaminsavage, line 248: I think this return type is probably too complex for a tuple. We should use a structure so that we don't have to remember what .0 and .1 and .2 and .3 all refer to.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,461 @@
...
+
+            // proof from second party
+            simple_proof_check(result[1].0, &result[0].1, &result[2].1);
+
+            // proof from third party
+            simple_proof_check(result[2].0, &result[1].1, &result[0].1);
+        });
+    }
+
+    fn simple_proof_check(
```
benjaminsavage, line 377: I don't follow this. This is not how we will eventually verify these proofs, it's not in the paper, and it's not clear to me how this works. I would rather not test the code in this way.

Rather than add all of this extra logic that does a novel approach - which requires reading and understanding, perhaps we should just defer testing this until we have more of the verifier code in place?

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
benjaminsavage: But upon further reflection, I think we should align on the entire approach to testing this.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: Yes I think that makes a lot of sense. Maybe we keep this test for now but once we have the verifier in place, we can delete it?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -73,6 +74,44 @@ where
...
+        // therefore the left share is "implicitly sent" to the left (communication via PRSS)
+        let (my_q_mask, q_mask_from_right_prover): (F, F) =
+            ctx.prss().generate_fields(*record_counter);
+        *record_counter += 1;
+
+        // compute final uv values
+        let (u_values, v_values) = &mut self.uv_chunks[0];
+        // shift first element to last position
+        u_values[SmallProofGenerator::RECURSION_FACTOR - 1] = u_values[0];
+        v_values[SmallProofGenerator::RECURSION_FACTOR - 1] = v_values[0];
```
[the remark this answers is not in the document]
danielmasny: The position of the other values does not matter, it is just the mask that needs to in the 0 position (at least in some cases). It might be ok to put the mask in a different point but the current security analysis does not cover that.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
benjaminsavage: Actually... we should just inline this function. It is only called in a single place. Then we don't have to worry about making any new structures, or thinking about complex return types.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,475 @@
...
+        // denominator
+        let denominator = CanonicalLagrangeDenominator::<
+            Fp61BitPrime,
+            { SmallProofGenerator::RECURSION_FACTOR },
+        >::new();
+        // lagrange table
+        let lagrange_table = LagrangeTable::<
+            Fp61BitPrime,
+            { SmallProofGenerator::RECURSION_FACTOR },
+            { SmallProofGenerator::LAGRANGE_LENGTH },
```
benjaminsavage, line 231: This code takes up a LOT more vertical space and becomes much less readable due to the size of these generic params. Can you assign these to a constant with a short name like `X` or something and avoid all of this wrapping?

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
benjaminsavage: If we plan to just delete it anyway, I would rather not add it in the first place. Let's keep the diff history simpler.

benjaminsavage (changes requested):

benjaminsavage (approved):

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -0,0 +1,475 @@
...
+        let communication_ctx = ctx.set_total_records(self.len());
+
+        // set up channel
+        let send_channel_left =
+            &communication_ctx.send_channel::<Fp61BitPrime>(ctx.role().peer(Direction::Left));
+
+        // send to left
+        // we send the proof batch via sending the individual field elements
+        communication_ctx
+            .parallel_join(
```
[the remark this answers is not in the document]
danielmasny: I have changed this from sequential join to parallel join after Andy suggested it.

PR #1086, opened 2024-05-21
Don't emit rerun-if-changed for the build script

andyleiserson:
`std::file!()` is `ipa-core/build.rs`, which is then interpreted relative to the package directory as the non-existent `ipa-core/ipa-core/build.rs`. This results in rerunning the build script and rebuilding the package on every build.

The default behavior, if the build script does not emit any `rerun-if-changed` directives, is to re-run it on every build. Emitting `rerun-if-changed=build.rs` can be used to suppress this behavior in the case where there are no other applicable rerun-if-changed directives. With or without `rerun-if-changed=build.rs`, cargo is smart enough to rebuild if the build script changes.

I considered whether to emit `rerun-if-changed=build.rs` when no step definition files are configured, but optimizing for that case doesn't seem worth the extra complexity.

### DIFF
--- a/ipa-step-derive/src/track.rs
+++ b/ipa-step-derive/src/track.rs
@@ -196,7 +196,6 @@ impl ToTokens for StepModules {
         tokens.extend(quote! {
             #modules
             fn #fn_name() {
-                println!("cargo:rerun-if-changed={f}", f = ::std::file!());
                 assert!(
                     ::std::env::var(::ipa_step::COMPACT_GATE_INCLUDE_ENV).is_err(),
                     "setting `{e}` in the environment will cause build errors",


andyleiserson:
> With or without rerun-if-changed=build.rs, cargo is smart enough to rebuild if the build script changes.

On the other hand, maybe there's some nuance of our proc-macro-in-build-script use case that I'm not seeing -- please let me know.

akoshelev:
I guess the intent was to rebuild it if any of the step definition module changed. Is there a way to make it work?

andyleiserson:
That was my understanding of the intent as well, and I believe with this change it will work as intended. The code to emit `rerun-if-changed` for the step definition files is at line 207, a few lines after the deleted line.

akoshelev (approved):
lgtm then, but maybe @martinthomson needs to take a look as well

martinthomson (approved):
Yeah, this is something that Cargo has to be doing for itself.  I don't know what I was thinking.

PR #1089, opened 2024-05-22
Modify malicious

danielmasny:
split from https://github.com/private-attribution/ipa/pull/1083

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -153,8 +153,8 @@ where
         B: Borrow<UVPolynomial<F, R>>,
     {
         let r: F = hash_to_field(
-            &compute_hash(proof_left.iter()),
-            &compute_hash(proof_right.iter()),
+            &compute_hash(proof_left.g.iter()),
+            &compute_hash(proof_right.g.iter()),
             R::U128,
         );
 
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -153,8 +153,8 @@ where
         B: Borrow<UVPolynomial<F, R>>,
     {
         let r: F = hash_to_field(
-            &compute_hash(proof_left),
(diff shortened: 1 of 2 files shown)


andyleiserson:
It would be helpful to have a summary of what is changing and why in the PR description.

If I understand right, `λ` is renamed to `R`, and the `R` type parameter is used directly in places where the parameter $N = 2\lambda + 1$ was used before. How does this simplify the code?

The `ProofGenerator` type seems to be replaced with `UVStore`. Part of what I understand (based on in-person discussions, as opposed to studying the PR) to be going on here  is that the first pass of the proof will work with an iterator over u and v values rather than storing them all in an array, since that is (1) possible for the first pass, and (2) a fully expanded copy of u and v for the first pass would be very large -- a significant multiple larger than the raw multiplication intermediates, at least until we implement packing. Writing the proof generation routines in terms of iterators enables them to work either from an array or iterator of u/v values. Since the name `UVStore` implies a data storage type, I think it would be better not to implement complex prover functionality in methods on the `UVStore` type. Similarly, if we're getting rid of the `ZeroKnowledgeProof` type in favor of using `GenericArray` directly, then we can't implement complex prover functionality as methods on `GenericArray`. Possibly the prover functionality could be implemented as free functions in the `prover` module. I see how the methods are serving as constructors for `UVStore` and `ZeroKnowledgeProof`, but I'm not sure there's a compelling reason for them to be that way. I feel like the logic that structures and terminates the recursion should also be implemented here, rather than left to the caller, but maybe there's a good reason an abstraction around that doesn't work.

It would be useful to have some additional comments on these routines explaining what they do and how they are to be used. Some amount of the detail is in the paper and when it is, references to the paper are fine, but I don't think it's possible to match the notation in the paper precisely enough that no additional explanation is necessary. It would also be nice if the implementation had enough self-contained documentation that it is possible to use the proof system without referring to the paper. (This concern is not new with the changes in this PR -- I feel the same about the original.)

Assuming we can align on the changes, is there a reason not to go ahead and update the tests to reflect the revisions (instead of leaving some routines as deprecated but used by tests)?

danielmasny:
close this PR, for discussion see https://github.com/private-attribution/ipa/pull/1083

PR #1090, opened 2024-05-23
change compute_proof in prover.rs

danielmasny:
This PR just makes changes to compute_proof changing the input types from iterators over fields to GenericArrays of fields. It also works now for owned values and references.


benjaminsavage (approved):
This is a nicely isolated change. It's easy to review and understand and the diff view is helpful. This is great.

PR #1091, opened 2024-05-23
changes to compute final proof, proof generator and challenge and recurse

danielmasny:
split from https://github.com/private-attribution/ipa/pull/1083.

follow up from https://github.com/private-attribution/ipa/pull/1090

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -153,11 +153,6 @@ where
                 if let Some(polynomial) = uv_iterator.next() {
                     u[i] = lagrange_table_r.eval(&polynomial.borrow().0)[0];
                     v[i] = lagrange_table_r.eval(&polynomial.borrow().1)[0];
-                } else {
-                    // fill GenericArray with zeros if there are not enough elements
-                    // i.e. length of uv_iterator is not a multiple of λ
-                    u[i] = F::ZERO;
-                    v[i] = F::ZERO;
                 }
             }
             output.push((u, v));
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -147,12 +147,14 @@ where
         while let Some(polynomial) = uv_iterator.next() {
             let mut u = GenericArray::<F, λ>::generate(|_| F::ZERO);
             let mut v = GenericArray::<F, λ>::generate(|_| F::ZERO);
-            u[0] = lagrange_table_r.eval(&polynomial.borrow().0)[0];
-            v[0] = lagrange_table_r.eval(&polynomial.borrow().1)[0];
+            let (u_chunk, v_chunk) = polynomial.borrow();
+            u[0] = lagrange_table_r.eval(u_chunk)[0];
+            v[0] = lagrange_table_r.eval(v_chunk)[0];
             for i in 1..λ::USIZE {
                 if let Some(polynomial) = uv_iterator.next() {
-                    u[i] = lagrange_table_r.eval(&polynomial.borrow().0)[0];
-                    v[i] = lagrange_table_r.eval(&polynomial.borrow().1)[0];
+                    let (u_chunk, v_chunk) = polynomial.borrow();
+                    u[i] = lagrange_table_r.eval(u_chunk)[0];
+                    v[i] = lagrange_table_r.eval(v_chunk)[0];
                 }
             }
             output.push((u, v));
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -56,11 +56,6 @@ where
     F: PrimeField,
     λ: ArrayLength,
 {
-    // pub fn new(u: Vec<F>, v: Vec<F>) -> Self {
-    //     debug_assert_eq!(u.len(), v.len(), "u and v must be of equal length");
-    //     Self { u, v }
-    // }
-
     pub fn compute_proof<J, B>(
         uv_iterator: J,
         lagrange_table: &LagrangeTable<F, λ, <λ as Sub<U1>>::Output>,
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -34,10 +34,13 @@ where
     }
 }
 
+#[allow(non_camel_case_types)]
 #[derive(Debug)]
-pub struct ProofGenerator<F: PrimeField> {
-    u: Vec<F>,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -134,74 +120,66 @@ where
...
+        while let Some(polynomial) = uv_iterator.next() {
+            let mut u = GenericArray::<F, λ>::generate(|_| F::ZERO);
+            let mut v = GenericArray::<F, λ>::generate(|_| F::ZERO);
+            u[0] = lagrange_table_r.eval(&polynomial.borrow().0)[0];
+            v[0] = lagrange_table_r.eval(&polynomial.borrow().1)[0];
+            for i in 1..λ::USIZE {
+                if let Some(polynomial) = uv_iterator.next() {
+                    u[i] = lagrange_table_r.eval(&polynomial.borrow().0)[0];
+                    v[i] = lagrange_table_r.eval(&polynomial.borrow().1)[0];
+                }
```
benjaminsavage, line 156: Why treat index 0 and the others differently?

benjaminsavage: Also, I think it would be more readable to use de-structuring as opposed to `.0` and `.1` which I find particularly unreadable.

```
let (u_chunk, v_chunk) = polynomial.borrow();
...
u[i] = lagrange_table_r.eval(u_chunk)[0];
v[i] = lagrange_table_r.eval(v_chunk)[0];
```

Is this possible or do we encounter issues with "borrow"?

danielmasny: Just changed it and it works!

Why is 0 treated differently: I do a compression here, basically I compress λ many generic arrays of length λ into a single generic array of length λ. That is why I manually iterate by calling next. Whenever I need to start a new array, I generate a 0 array. Since next is `some`, the first element of the new array always has a value that I can place. Now I need to call another λ-1 times next but the iterator might stop, i.e. next() is none. Then I don't set a value and the current index i and all following indexes are remaining 0, I.e. I fill them with 0 such that the total amount of field elements is a multiple of λ and there is no issue with the next recursion.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -134,74 +120,68 @@ where
...
             }
-            let p_r = lagrange_table_r.eval(&p)[0];
-            let q_r = lagrange_table_r.eval(&q)[0];
-            (p_r, q_r)
-        });
-        let (u, v) = pairs.unzip();
-        ProofGenerator::new(u, v)
+            output.push((u, v));
+        }
```
benjaminsavage, line 161:
```
let mut u = GenericArray::<F, λ>::generate(|_| F::ZERO);
let mut v = GenericArray::<F, λ>::generate(|_| F::ZERO);
let mut index = 0;
while let Some(polynomial) = uv_iterator.next() {
    p_r = lagrange_table_r.eval(u_chunk)[0];
    q_r = lagrange_table_r.eval(v_chunk)[0];
    if index >= λ::USIZE {
        u = GenericArray::<F, λ>::generate(|_| F::ZERO);
        v = GenericArray::<F, λ>::generate(|_| F::ZERO);
        output.push((u, v));
        index = 0;
    }
    u[index] = p_r;
    v[index] = q_r;
    index += 1;
}
```

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -134,74 +120,72 @@ where
...
-                p[i] = x;
-                q[i] = y;
+        // iter and interpolate at x coordinate r
+        let mut index = 0;
+        let mut new_u_chunk = GenericArray::<F, λ>::generate(|_| F::ZERO);
+        let mut new_v_chunk = GenericArray::<F, λ>::generate(|_| F::ZERO);
+        for polynomial in uv_iterator {
+            let (u_chunk, v_chunk) = polynomial.borrow();
+            let u = lagrange_table_r.eval(u_chunk)[0];
+            let v = lagrange_table_r.eval(v_chunk)[0];
```
benjaminsavage, line 152: I'd rather not call them "u" and "v", as those are the names of vectors and these are just points. I believe we are evaluating the polynomials p(x) and q(x) at the random challenge point "r" which is why I suggested p_r and q_r.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -134,74 +120,72 @@ where
...
-        });
-        let (u, v) = pairs.unzip();
-        ProofGenerator::new(u, v)
+            new_u_chunk[index] = u;
+            new_v_chunk[index] = v;
+            index += 1;
+        }
+        if index != 0 {
+            output.push((new_u_chunk, new_v_chunk));
+        }
```
benjaminsavage, line 165: This is probably functionally correct, but I think it's rather subtle to rely on the index value to decide if you need to push or not.

I'd suggest instead that we always push each new_u_chunk and new_v_chunk upon creation of these chunks.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -134,74 +120,72 @@ where
...
 
-impl<F> PartialEq<(&[u128], &[u128])> for ProofGenerator<F>
+#[allow(non_camel_case_types)]
+impl<F, λ> PartialEq<(&[u128], &[u128])> for ProofGenerator<F, λ>
 where
     F: PrimeField + std::cmp::PartialEq<u128>,
+    λ: ArrayLength,
 {
     fn eq(&self, other: &(&[u128], &[u128])) -> bool {
         let (cmp_a, cmp_b) = other;
```
benjaminsavage, line 178: I know I chose these names, but I think we can improve them. cmp_a and cmp_b don't tell me what they are.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: Wouldn't the vector take the ownership? Such that I need to insert an individual values within the last element of the vector?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: I think we currently only need this function for tests, so it would make sense to move it into the test section. I am happy to change the names despite this though.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
danielmasny: I could remove the index if statement, the index is only 0 when the iterator is empty and in this case the output vector would contain an array of zeros (without the if statement) rather than being empty.

benjaminsavage (approved):
Merging this to keep progress going. Let's address the comments in follow-up PRs.

PR #1092, opened 2024-05-23
Improve build time for compact gate

akoshelev:
I was investigating why build with compact gate takes that long and everything pointed to large match statements inside `FromStr` and `AsRef` trait implementations for `CompactGate`. The major bottleneck was `AsRef` match statement and overall the size of match statement impacts build time exponentially. Once the number of steps crosses the 70k boundary, release builds do not finish. 

The fix was to replace the match arms with something else for both `FromStr` and `AsRef`. Latter can be done with a simple array indexed by compact gate index, so it is very straightforward. That alone got the compile time down from ~4mins to 50s for 20k steps.

Fixing `FromStr` is more involved because it requires a lookup table. I've experimented with PHF and an array of tuples `(hash(str), gate index)`, sorted by hash value. Both have similar compilation times and I ended up with the array implementation to reduce the runtime cost a bit - PHF may require a few hash evaluations on a given string at runtime and array needs just one + binary search inside a sorted array which should be fast.

## Compile time impact

The following table shows the build times for `cargo bench --bench oneshot_ipa --no-default-features --features "enable-benches compact-gate"` command. For debug builds I used `cargo test --bench oneshot_ipa --no-default-features --features "enable-benches compact-gate"`

I measured the impact of the second build without the fix implemented in #1086 - that allowed me to capture the overhead of compact gate generation and compilation. 

| | debug build (fresh) | debug build (second) | release build (fresh) | release build (second) |
|-|---------------------|----------------------| --- | -- |
| before | 1m 18s  | 14.56s | 4m 02s | 3m 49s |
| after    | 1m 07s  | 10.18s | **1m 30s** | **23s** | 

As it can be seen, compiler optimizations cause the slowdown. Hinting `rustc` that there is nothing to optimize by using arrays improve optimized build times. 

## Runtime impact

I didn't anticipate any impact on `as_ref` implementation because that's just an array lookup and was more concerned about `FromStr` as it now requires a hash + binary search. Fortunately we only use it in HTTP deserialization and only when we receive a new `step` request, so the impact should be proportional to the number of gates we use and limited to one time overhead per gate creation.

We don't have reliable benchmarks to test it at scale, so I used our integration test to validate.

`cargo test --test helper_networks https_semi_honest_ipa --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"`

for release build

`cargo test --release --test helper_networks https_semi_honest_ipa --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"`

| | debug | release |
|-|---------------------|----------------------| 
| before | 6.63s  | 1.54s |
| after    | 4.65s |  1.52s |

### DIFF
--- a/ipa-step/src/gate.rs
+++ b/ipa-step/src/gate.rs
@@ -7,7 +7,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse2, parse_str, Ident, Path};
 
-use crate::{hashing::HashingSteps, name::GateName, CompactGateIndex, CompactStep};
+use crate::{hash::HashingSteps, name::GateName, CompactGateIndex, CompactStep};
 
 fn crate_path(p: &str) -> String {
     let Some((_, p)) = p.split_once("::") else {
@@ -102,9 +102,10 @@ fn compact_gate_impl<S: CompactStep>(gate_name: &str) -> TokenStream {
         impl ::std::str::FromStr for #ident {
             type Err = String;
             fn from_str(s: &str) -> Result<Self, Self::Err> {
-                match s {
-                    "/" => Ok(Self::default()),
-                    v => GATE_LOOKUP.find(v).map(#ident).ok_or_else(|| format!(#from_panic)),
+                if s == "/" {
+                    Ok(Self::default())
+                } else {
+                    GATE_LOOKUP.find(s).map(#ident).ok_or_else(|| format!(#from_panic))
                 }
             }
         }
--- a/ipa-step/src/gate.rs
+++ b/ipa-step/src/gate.rs
@@ -7,7 +7,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse2, parse_str, Ident, Path};
 
-use crate::{name::GateName, CompactGateIndex, CompactStep};
+use crate::{hashing::HashingSteps, name::GateName, CompactGateIndex, CompactStep};
 
 fn crate_path(p: &str) -> String {
     let Some((_, p)) = p.split_once("::") else {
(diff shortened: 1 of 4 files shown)


ipa-step/src/hashing.rs
```diff
@@ -0,0 +1,101 @@
...
+                inner: [(u64, u32); #sz]
+            }
+
+            impl #lookup_type {
+                fn find(&self, input: &str) -> Option<u32> {
+                    let h = Self::hash(input);
+                    self.inner.binary_search_by_key(&h, |(hash, _)| *hash).ok().map(|i| self.inner[i].1)
+                }
+
+                /// This must be kept in sync with proc-macro code that generates the hash.
```
akoshelev, line 92: I couldn't find a way to embed the `hash` function definition inside. Would be great if it was possible somehow

martinthomson: How about:

```rust
pub trait StepHasher {
  fn hash_step(&self) -> u64;
}
impl StepHasher for str {
   fn hash_step(&self) -> u64 { ... }
}
```

Then 

```rust
fn find(&self, input: &str) -> Option<u32> {
  use ::path::to::step::hash::StepHasher;
  self.inner.binary_search_by_key(input.step_hash(), |...| ...).ok().map(...)
}
```

ipa-step/src/hashing.rs
```diff
@@ -0,0 +1,101 @@
...
+///
+/// [`FromStr`]: std::str::FromStr
+pub(crate) struct HashingSteps {
+    inner: BTreeMap<u64, CompactGateIndex>,
+    type_name: Ident,
+}
+
+fn hash(s: &str) -> u64 {
+    let mut hasher = DefaultHasher::default();
+    s.hash(&mut hasher);
```
martinthomson, line 42:
```suggestion
    // If you get a collision, you can avoid that by adding:
    // > hasher.write_u8(0);
    // Tweak the value or remove the tweak if you get another collision.
    hasher.write_str(s);
```

ipa-step/src/hashing.rs
```diff
@@ -0,0 +1,101 @@
...
+/// The complexity of this operation at compile time is O(n) and the cost is hash(str)*n.
+/// Runtime overhead is proportional to hash(str)+log(n).
+///
+/// ## Hash collisions
+/// This currently uses the standard library [`Hasher`] interface to generate hashes that are limited
+/// to 8 bytes in size. This means that there is a small chance of hash collisions that increases
+/// with the overall increase in number of steps.
+///
+/// If that happens, a list of potential mitigations include
+/// * salt the hash,
```
martinthomson, line 25:
```suggestion
/// * salt the hash (see code below),
```

ipa-step/src/gate.rs
```diff
@@ -54,50 +54,57 @@ pub fn build<S: CompactStep>() {
...
+
         impl ::std::convert::AsRef<str> for #ident {
-            #[allow(clippy::too_many_lines)]
             fn as_ref(&self) -> &str {
-                match self.0 {
+                match usize::try_from(self.0).unwrap() {
                     0 => "/",
-                    #as_ref_arms
-                    _ => unreachable!(),
+                    i => STR_LOOKUP[i - 1],
```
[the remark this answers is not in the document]
akoshelev: I did that and reverted that change eventually because the way it is written now shows what is actually stored inside `STR_LOOKUP` and `GATE_LOOKUP`. I don't strongly object it, just feels that it is more explanatory

PR #1093, opened 2024-05-23
use gen array in verifier as well

danielmasny:
doing similar changes to verifier.rs as in https://github.com/private-attribution/ipa/pull/1090

- use GenericArray as inputs and in ProofVerifier vector
- use iterator with Borrow trait
- adjust tests to work with updated inputs

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
@@ -1,6 +1,9 @@
-use std::ops::{Add, Sub};
+use std::{
+    borrow::Borrow,
+    ops::{Add, Sub},
+};
 
-use generic_array::ArrayLength;
+use generic_array::{sequence::GenericSequence, ArrayLength, GenericArray};
 use typenum::{Sum, U1};
 
 use super::prover::{TwoNMinusOne, TwoNPlusOne, ZeroKnowledgeProof};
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -21,81 +28,82 @@ pub struct ProofVerifier<F: PrimeField> {
...
 {
-    pub fn new(u_or_v: Vec<F>, out_share: F) -> Self {
+    pub fn new(u_or_v: Vec<GenericArray<F, λ>>, out_share: F) -> Self {
         Self { u_or_v, out_share }
     }
 
-    pub fn verify_proof<λ>(
-        &self,
+    pub fn verify_proof<J, B>(
+        mut u_or_v_iterator: J,
```
benjaminsavage, line 41: There's something strange about this argument being mutable. I wonder if there is any alternative.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -21,81 +28,82 @@ pub struct ProofVerifier<F: PrimeField> {
...
+        // and interpolate at x coordinate r
+        while let Some(polynomial) = u_or_v_iterator.next() {
+            let mut new_u_or_v = GenericArray::<F, λ>::generate(|_| F::ZERO);
+            new_u_or_v[0] = lagrange_table_p_or_q_r.eval(polynomial.borrow())[0];
+            for i in 1..λ::USIZE {
+                if let Some(polynomial) = u_or_v_iterator.next() {
+                    new_u_or_v[i] = lagrange_table_p_or_q_r.eval(polynomial.borrow())[0];
+                }
+            }
+            new_u_or_v_vec.push(new_u_or_v);
```
benjaminsavage, line 76: Let's employ a similar approach to the one we went with on the last PR, that is creating a mutable generic array outside of the loop, and storing a mutable index, then having only a single while loop that over-writes the correct index of this array until it's full at which point it pushes it and creates a new one.

benjaminsavage (changes requested):

benjaminsavage (approved):

PR #1095, opened 2024-05-25
Upgraded to Axum 0.7

cberkhoff:
Using newest version of Axum (0.7.5) and Hyper which is beyond 1.0 bringing more stability. Since Hyper doesn’t include a client, we are using the one suggested in Hyper-Utils, which is a drop-in replacement from the client that was present in Hyper 0.14.

Hyper now relies on Http-Body 1 which cascaded into a bunch of changes to HTTP bodies used throughout the code. Hyper also gives the option to pick any execution framework, but we’re sticking to Tokio.

During my testing I observed that the tests like `https_semi_honest_ipa` would get stuck. After a lot of debugging, I was able to identify `SETTINGS_MAX_CONCURRENT_STREAMS` option for HTTP2 connections as the culprit. By default, Hyper-Utils sets it to 200 which is insufficient for that test. 1000 was the “smallest” number I was able to identify that passed the tests. It might be worth investigating why the client doesn’t correctly handle the case when the server can’t take more concurrent streams. Right now, it just gets stuck in [when the transport is calling a step](https://github.com/private-attribution/ipa/blob/8c481563314c8f1c5cc9a941ca330fdfce30c2f0/ipa-core/src/net/transport.rs#L203). That await never completes.

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -113,9 +113,9 @@ hpke = { version = "0.11.0", default-features = false, features = [
     "std",
     "x25519",
 ] }
-hyper = { version = "1.3.1", optional = true, features = [ "full" ] }
+hyper = { version = "1.3.1", optional = true, features = [ "http2", "server" ] }
 hyper-rustls = { version = "0.27.1", optional = true, features = ["http2"] }
-hyper-util = { version = "0.1.3", optional = true, features = ["full"] }
+hyper-util = { version = "0.1.3", optional = true, features = ["http2"] }
 http-body-util = { version = "0.1.1", optional = true }
 http-body = { version = "1", optional = true }
 iai = { version = "0.1.1", optional = true }
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -40,6 +40,9 @@ web-app = [
     "toml",
     "tower",
     "tower-http",
+    "hyper-util",
+    "http-body",
+    "http-body-util",
 ]
 test-fixture = ["weak-field"]
 # Include observability instruments that detect lack of progress inside MPC. If there is a bug that leads to helper
@@ -78,10 +81,10 @@ ipa-step-derive = { version = "*", path = "../ipa-step-derive" }
 aes = "0.8.3"
 async-trait = "0.1.79"
 async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
-axum = { version = "0.6", optional = true, features = ["http2", "macros"] }
+axum = { version = "0.7.5", optional = true, features = ["http2", "macros"] }
 # The following is a temporary version until we can stabilize the build on a higher version
 # of axum, rustls and the http stack.
-axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
+axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "0.6.1", version = "0.6.1", optional = true, features = [
     "tls-rustls",
 ] }
 base64 = { version = "0.21.2", optional = true }
@@ -110,15 +113,11 @@ hpke = { version = "0.11.0", default-features = false, features = [
     "std",
     "x25519",
 ] }
-hyper = { version = "0.14.28", optional = true, features = [
-    "client",
-    "h2",
-    "stream",
-    "runtime",
-] }
-# The following is a temporary version until we can stabilize the build on a higher version
-# of axum, rustls and the http stack.
-hyper-rustls = { git = "https://github.com/cberkhoff/hyper-rustls/", branch = "rustls-0.23", version = "0.25.1", optional = true, features = ["http2"] }
+hyper = { version = "1.3.1", optional = true, features = [ "full" ] }
+hyper-rustls = { version = "0.27.1", optional = true, features = ["http2"] }
+hyper-util = { version = "0.1.3", optional = true, features = ["full"] }
+http-body-util = { version = "0.1.1", optional = true }
+http-body = { version = "1", optional = true }
 iai = { version = "0.1.1", optional = true }
 metrics = "0.21.0"
 metrics-tracing-context = "0.14.0"
@@ -143,7 +142,7 @@ tokio-rustls = { version = "0.26", optional = true }
 tokio-stream = "0.1.14"
 toml = { version = "0.8", optional = true }
 tower = { version = "0.4.13", optional = true }
-tower-http = { version = "0.4.0", optional = true, features = ["trace"] }
+tower-http = { version = "0.5", optional = true, features = ["trace"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 typenum = { version = "1.17", features = ["i128"] }
--- a/ipa-core/src/config.rs
+++ b/ipa-core/src/config.rs
@@ -8,7 +8,8 @@ use std::{
     time::Duration,
 };
 
-use hyper::{client::Builder, http::uri::Scheme, Uri};
+use hyper::{http::uri::Scheme, Uri};
+use hyper_util::client::legacy::Builder;
 use rustls_pemfile::Item;
 use rustls_pki_types::{CertificateDer, PrivateKeyDer};
 use serde::{Deserialize, Deserializer, Serialize};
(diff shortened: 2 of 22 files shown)


ipa-core/src/net/server/handlers/query/mod.rs
```diff
@@ -142,8 +150,10 @@ pub mod test_helpers {
             .await;
         let resp = test_server.server.handle_req(req).await;
         let status = resp.status();
-        let body_bytes = hyper::body::to_bytes(resp.into_body()).await.unwrap();
+        let buf = resp.into_body().collect().await.unwrap().aggregate();
```
akoshelev, line 153:
```suggestion
        let buf = resp.into_body().collect().await.unwrap().to_bytes();
```

seems that [to_bytes](https://docs.rs/http-body-util/latest/http_body_util/struct.Collected.html#method.to_bytes) is a better option for your use-case

ipa-core/src/net/error.rs
```diff
@@ -75,7 +77,7 @@ impl Error {
         let status = resp.status();
         assert!(status.is_client_error() || status.is_server_error()); // must be failure
         let (endpoint, body) = resp.into_parts();
-        hyper::body::to_bytes(body)
+        axum::body::to_bytes(body, usize::MAX)
```
akoshelev, line 80: `usize::MAX` is concerning here. Is there a way to put a limit here?

ipa-core/src/net/client/mod.rs
```diff
@@ -369,7 +391,8 @@ impl MpcHelperClient {
         gate: &Gate,
         data: S,
     ) -> Result<ResponseFuture, Error> {
-        let body = hyper::Body::wrap_stream::<_, _, Error>(data.map(Ok));
+        let data = data.map(|v| Ok::<bytes::Bytes, Error>(Bytes::from(v)));
+        let body = axum::body::Body::from_stream(data);
```
akoshelev, line 395: I commented on a similar code down below, but I think we should not mix and match body combinators from Hyper and Axum. It would make it much harder to support both in our code. I'd prefer if we use hyper for everything related to url/request/body management and axum strictly for web server functionality (handlers/middleware)

ipa-core/src/net/client/mod.rs
```diff
@@ -274,22 +287,44 @@ impl MpcHelperClient {
...
+        } else {
+            Err(Error::from_failed_resp(resp).await)
+        }
+    }
+
+    /// Reads the entire response from the server into a String
+    ///
+    /// # Errors
+    /// If there was an error collecting the response stream.
+    async fn response_to_str(resp: ResponseFromEndpoint<'_>) -> Result<String, Error> {
```
akoshelev, line 306: you shouldn't require strings to deal with serialization, serde_json can use byte slices for deserialization. This will make your code easier as you can use `to_bytes` on `Aggregated`

ipa-core/src/net/client/mod.rs
```diff
@@ -247,7 +257,10 @@ impl MpcHelperClient {
         auth_header: Option<(HeaderName, HeaderValue)>,
         conf: &C,
     ) -> Self {
-        let client = conf.configure(&mut Client::builder()).build(connector);
+        let mut builder = Client::builder(TokioExecutor::new());
+        builder.timer(TokioTimer::new()); // this timer is necessary for http2
+                                          //builder.pool_timer(TokioTimer::new());
```
akoshelev, line 262:
```suggestion
```

ipa-core/src/net/client/mod.rs
```diff
@@ -247,7 +257,10 @@ impl MpcHelperClient {
         auth_header: Option<(HeaderName, HeaderValue)>,
         conf: &C,
     ) -> Self {
-        let client = conf.configure(&mut Client::builder()).build(connector);
+        let mut builder = Client::builder(TokioExecutor::new());
```
akoshelev, line 260: if building a client requires more than just `Client::new()` we should have a common entry point for it

Also, consider elaborating more on `timer` use, saying http2 requires it is too vague

ipa-core/src/config.rs
```diff
@@ -275,6 +276,10 @@ pub struct ServerConfig {
 
     /// Configuration needed for encrypting and decrypting match keys
     pub hpke_config: Option<HpkeServerConfig>,
+
+    // Sets the SETTINGS_MAX_CONCURRENT_STREAMS option for HTTP2 connections.
+    // [`hyper_util::server::conn::auto::Http2Builder::max_concurrent_streams`]
+    pub max_concurrent_stream: Option<u32>,
```
akoshelev, line 282: this should go to `Http2Configurator`

ipa-core/Cargo.toml
```diff
@@ -78,10 +81,10 @@ ipa-step-derive = { version = "*", path = "../ipa-step-derive" }
 aes = "0.8.3"
 async-trait = "0.1.79"
 async-scoped = { version = "0.9.0", features = ["use-tokio"], optional = true }
-axum = { version = "0.6", optional = true, features = ["http2", "macros"] }
+axum = { version = "0.7.5", optional = true, features = ["http2", "macros"] }
 # The following is a temporary version until we can stabilize the build on a higher version
 # of axum, rustls and the http stack.
-axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "rustls-0.23", version = "0.5.3", optional = true, features = [
+axum-server = { git = "https://github.com/cberkhoff/axum-server/", branch = "0.6.1", version = "0.6.1", optional = true, features = [
```
akoshelev, line 87: is 0.6.1 the right version for `axum` 0.7.5?

akoshelev (changes requested):

ipa-core/src/net/server/mod.rs
```diff
@@ -619,7 +633,9 @@ mod e2e_tests {
             .await;
 
         // client
-        let client = hyper::Client::new();
+        let client = Client::builder(TokioExecutor::new())
```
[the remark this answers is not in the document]
cberkhoff: Will do

ipa-core/src/net/error.rs
cberkhoff: What number do you think?

ipa-core/src/net/client/mod.rs
cberkhoff: I'll investigate what a switch to Hyper body would entail. BTW, that body is defined on a separate more general http-body crate. The Axum body had a few functions that were useful, that why I was using it. It made the AxumBodyWrapper simpler.

ipa-core/Cargo.toml
cberkhoff: 0.6.0 is the "official" one. I had to patch it for this PR, hence the minor bump

ipa-core/src/net/server/handlers/query/status.rs
```diff
@@ -32,8 +32,11 @@ pub fn router(transport: Arc<HttpTransport>) -> Router {
 
 #[cfg(all(test, unit_test))]
 mod tests {
-    use axum::http::uri::{Authority, Scheme};
-    use hyper::{Body, StatusCode};
+    use axum::{
+        body::Body,
```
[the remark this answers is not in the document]
cberkhoff: Sorry for the delayed response, I wanted to investigate this a bit more. 

The thing is that the Axum Body is a really convenient type since it supports all our use cases for Empty, Full and Streaming types which are disjointed in Http Body. If we don't use it, we'll have to re-implement something very similar. Given that, I would rather not re-implement it now. If we ever want to move away from Axum, we can decide what we want to do at that point.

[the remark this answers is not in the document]
akoshelev: So you're saying that it is not possible to use `Empty`, `Full` and `Streaming` directly in our code without extra code that needs to wrap them to fit our specific use case? I wouldn't re-implement axum functionality in this case, but perhaps my understanding that `axum::body` is just a wrapper around these was wrong

ipa-core/src/net/error.rs
akoshelev: we chatted offline, a few dozen MBs seems reasonable to me

ipa-core/src/config.rs
cberkhoff: I have been investigating how this works in hyper and it seems like SETTINGS_MAX_CONCURRENT_STREAMS is a server only parameter. I was reading the RFC and it also says only the peers (the endpoints) can set this value. Hence I think it makes sense to keep this parameter where it's, in the ServerConfig.

akoshelev: Ok, I am convinced that it should be the server-side setting, but `ServerConfig` is not the right place for it. I just merged https://github.com/private-attribution/ipa/pull/1104 that creates server-side config, this should go there I think

ipa-core/src/net/client/mod.rs
cberkhoff: We only create the client once in this fashion, and I find the code readable as is. Note it has dependencies on the configuration and executor, etc. Instead of a function that would require many params, I find the fluent api to be concise. Just my opinion, don’t feel too strongly about this.

akoshelev (approved):

PR #1101, opened 2024-05-29
Add more tracing to the infrastructure

akoshelev:
I've been using these for quite a while, but forgot to commit them to the main branch. This logs the following operations:

* When Hyper comes in to take the next chunk from ordering sender
* When unordered receiver gets more data from the underlying HTTP stream
* When helper receives a new step request


andyleiserson (approved):
The messages are terse, but that's probably okay as long as they are emitted with the module name.

akoshelev:
> The messages are terse, but that's probably okay as long as they are emitted with the module name.

yea that is the reason - strings are getting long in logs.

akoshelev:
Merging it as the CI run on my local fork passed: https://github.com/akoshelev/raw-ipa/actions/runs/9281219804/job/25536830009

PR #1102, opened 2024-05-29
Remove more GenericArray usage

benjaminsavage:
Moving away from use of GenericArray in favor of just normal vanilla arrays.

### DIFF
--- a/ipa-core/src/protocol/context/dzkp_field.rs
+++ b/ipa-core/src/protocol/context/dzkp_field.rs
@@ -1,19 +1,17 @@
 use bitvec::{order::Lsb0, view::BitView};
-use generic_array::GenericArray;
-use typenum::U32;
 
 use crate::{
     ff::{Field, Fp61BitPrime, PrimeField},
     protocol::context::dzkp_validator::{Array256Bit, SegmentEntry},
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/context/dzkp_field.rs
```diff
@@ -57,6 +55,16 @@ pub trait DZKPBaseField: PrimeField {
     ) -> Vec<UVSingleBlock<Self>>;
 }
 
+impl FromIterator<Fp61BitPrime> for [Fp61BitPrime; B] {
+    fn from_iter<T: IntoIterator<Item = Fp61BitPrime>>(iter: T) -> Self {
+        let mut out = [Fp61BitPrime::ZERO; B];
+        for (i, elem) in iter.into_iter().take(B).enumerate() {
```
danielmasny, line 61: I think it makes sense to panic here instead of taking B many. We don't want to drop elements from the iterator accidentally.

danielmasny (changes requested):

danielmasny (approved):
using assert is fine with me, we could also not use an assert and let the function panic via an index out of bounds panic.

Issue #1085, opened 2024-05-21
IPA stalls when running OPRF protocol over 50,000 records


andyleiserson:
Steps to reproduce:
1. Change the `INPUT_SIZE` definition in `tests/common.mod.rs` to 50000.
2. Run `cargo test --release --no-default-features --features="cli test-fixture real-world-infra web-app disable-metrics multi-threading stall-detection compact-gate" -- https_semi_honest --nocapture`.

Sample stall-detector output (this is from a run in @eriktaubeneck's infrastructure):
```
Helper 3: {
Helper 3: "step=protocol/convert_input_rows_to_prf/eval_prf/reveal_r", from=H2. Waiting to receive records ["[0..765]"].
Helper 3: }
Helper 3: 2024-05-16T01:11:27.337816Z WARN stall_detector{role=H3}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=1130548 state=
Helper 2: {
Helper 2: "step=protocol/sort_by_timestamp/quicksort_pass1/compare/bit0", from=H1. Waiting to receive records ["[0..156]"].
Helper 2: }
Helper 2: 2024-05-16T01:11:27.339636Z WARN stall_detector{role=H2}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=1181394 state=
Helper 1: {
Helper 1: "step=protocol/convert_input_rows_to_prf/eval_prf/revealz", from=H3. Waiting to receive records ["[0..765]"].
Helper 1: }
Helper 1: 2024-05-16T01:11:27.337394Z WARN stall_detector{role=H1}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=1230176 state=
Helper 2: {
Helper 2: "step=protocol/sort_by_timestamp/quicksort_pass1/compare/bit0", from=H1. Waiting to receive records ["[0..156]"].
Helper 2: }
Helper 2: 2024-05-16T01:11:57.342580Z WARN stall_detector{role=H2}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=1181394 state=
Helper 1: {
Helper 1: "step=protocol/convert_input_rows_to_prf/eval_prf/revealz", from=H3. Waiting to receive records ["[0..765]"].
Helper 1: }
[...]

PR #1104, opened 2024-05-29
Disable connection-level flow control for HTTP2 streams

akoshelev:
It was diagnosed as root cause for #1085. Here are the reasons why we think it is safe to turn it off:

* We have flow control built in at the application layer. `OrderingSender` provides backpressure per stream/step and MPC protocol logic breaks the whole execution into chunks where we open large but limited number of streams.
* We have flow control at TCP layer
* Connection level flow control for HTTP2 connections works well for web-servers that don't control clients that can connect to it. In our case, we have a very limited set of clients for each MPC-to-MPC connection.

I had a chat with @aleiserson, and we think ultimately it will be beneficial if we only used one mechanism to provide and given the nature of IPA, it would be better to have full control there. This means that application-level mechanism will be our preferred way to establish congestion control and turning off transport-layer flow control fits into our long-term view for IPA.

This change also reverts the active window back to 1024. I am concerned that the attribution phase may run slower with 32 and if PRF can work with 1024, even at the expense of dropping very large buffers down to the network, we can live with it for now.

Next step I want to implement variable-sized send buffers to maintain the guarantee from our infrastructure to flush fixed-size chunks down to the network layer

## Testing
Local run for 50k left overnight.
draft [run](https://draft-mpc.vercel.app/query/view/sewed-debit-6490) succeeded

### DIFF
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -221,7 +221,7 @@ impl Gateway {
 
 impl Default for GatewayConfig {
     fn default() -> Self {
-        Self::new(32)
+        Self::new(1024)
     }
 }
 
--- a/ipa-core/src/net/mod.rs
+++ b/ipa-core/src/net/mod.rs
@@ -25,6 +25,13 @@ pub use transport::{HttpShardTransport, HttpTransport};
 pub const APPLICATION_JSON: &str = "application/json";
 pub const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
 
+/// This has the same meaning as const defined in h2 crate, but we don't import it directly.
+/// According to the [`spec`] it cannot exceed 2^31 - 1.
+///
+/// Setting up initial window size to this value effectively turns off the flow control mechanism.
+/// [`spec`]: <https://datatracker.ietf.org/doc/html/rfc9113#name-the-flow-control-window>
+pub(crate) const MAX_HTTP2_WINDOW_SIZE: u32 = (1 << 31) - 1;
+
 /// Provides access to IPAs Crypto Provider (AWS Libcrypto).
 static CRYPTO_PROVIDER: Lazy<Arc<CryptoProvider>> =
     Lazy::new(|| Arc::new(rustls::crypto::aws_lc_rs::default_provider()));
--- a/ipa-core/src/net/server/config.rs
+++ b/ipa-core/src/net/server/config.rs
@@ -1,0 +1,33 @@
+use axum_server::HttpConfig;
+
+use crate::net::MAX_HTTP2_WINDOW_SIZE;
+
+pub(super) struct HttpServerConfig(HttpConfig);
+
+impl Default for HttpServerConfig {
+    fn default() -> Self {
+        Self(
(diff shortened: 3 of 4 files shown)


benjaminsavage (approved):

andyleiserson (approved):
> This change also reverts the active window back to 1024. I am concerned that the attribution phase may run slower with 32 and if PRF can work with 1024, even at the expense of dropping very large buffers down to the network, we can live with it for now.

When I decreased the value, I did a quick benchmark run, and it didn't make a big difference. I probably did that with the memory transport though. I have a slight preference for 32, since that's what I used for most of my performance tuning of the vectorized implementation (including end-to-end runs where it also applied to attribution). But it doesn't really matter -- either way, we should keep it in mind for future tuning, and I'm pretty sure that that disabling buffering for vectorized channels will make a bigger difference than 32 vs. 1024.

This config tweak is another reason why we might want to separate the helper-to-helper HTTP server from the client-to helper HTTP server at some point.

akoshelev:
> But it doesn't really matter -- either way, we should keep it in mind for future tuning, and I'm pretty sure that that disabling buffering for vectorized channels will make a bigger difference than 32 vs. 1024.

yea, I am working on the change that fixes the size of chunks that we flush down to the network layer. I did some profiling and `32` is a bit too low for shuffle - I am seeing 500 byte chunks sent over and over, so it does seem to take longer to finish. Maybe it is not important for the protocol, but I think (hope) fixed-size chunks will get vectorized channels to the right place in terms of performance

martinthomson:
I just want to note that the summary here includes a couple of notes that the folks working on HTTP/2 (and HTTP/3) would be very interested to hear.  It's important to understand that connection-level flow control exists to protect endpoints from certain abuse scenarios, as noted.

However, TCP flow control is not an effective substitute because of how HTTP/2 works.  There is also no similar defense for HTTP/3.

What we're saying here is that we believe this to be problematic.  That suggests implementation issues, like not growing credit properly, more than it says that connection-level flow control is fundamentally undesirable.

akoshelev:
Fair point @martinthomson, I don't see any fundamental issues with HTTP/2 connection-level and/or stream-level flow control. Both exists for a good reason. I do agree that the issue of not scaling the window is likely due to generic implementation of HTTP client that we use in IPA that does not take advantage of HTTP/2 specifics and just stalls when the default connection window is exhausted.

https://github.com/hyperium/hyper/issues/1960 seems to be hinting that it may be the case. 

#1107 mentions migrating to reqwest that may be a more sophisticated implementation of HTTP2 client, we can check that. 

Even with the optimal connection-level flow control implemented in network layer, I think the ultimate solution will still require work on the application side. If we are getting close to saturation, we need to prioritize streams that belong to "early"/lower steps in MPC and connection-level flow control may not help us here. 

I am thinking about TCP windowing as last layer of defense where you need to do more damage control than actual work. Our application gets more or less consistent throughput regardless of the input size and can be fine-tuned to avoid that.

martinthomson:
Hah, it's ROC.  I'm a little surprised that remains unfixed, but not hugely surprised.  We only recently added dynamic window scaling to Firefox recently.  It turns out that you can go a long time without really needing that.

PR #1105, opened 2024-05-30
Add constants for proof parameter sets

danielmasny:
add constants & add small parameter set

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
@@ -151,7 +151,7 @@ where
         );
 
         // assertion that table is not too large for the stack
-        assert!(<F as Serializable>::Size::USIZE * N * M < 2024);
+        assert!(<F as Serializable>::Size::USIZE * N * M < 8192);
 
         LagrangeTable {
             table: (N..(N + M))
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -81,30 +81,43 @@ where
     output
 }
 
+const TEST_RECURSION_FACTOR: usize = 4;
+/// `LAGRANGE_LENGTH` is `RECURSION_FACTOR - 1`
+const TEST_LAGRANGE_LENGTH: usize = 3;
+/// `PROOF_LENGTH` is `2 * RECURSION_FACTOR - 1`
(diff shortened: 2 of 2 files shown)


andyleiserson:
To reduce the code duplication, how about this: https://github.com/danielmasny/ipa/compare/aliases-for-constants...andyleiserson:ipa:prover-configs?diff=split&w=

There are two commits on that branch with two different versions. One (the first commit) puts the `F: PrimeField` on the struct, and the other (the second commit / head of the branch) puts it on the functions. The benefit of putting it on the functions is it avoids `PhantomData`, but requires more type annotations where things are invoked. Putting it on the struct (and thus in the type alias) eliminates the most type annotations, but requires `PhantomData`. 

Putting it on the struct also might enable caching the Lagrange table within the prover, but doing that is more involved due to the limitations on generic statics.

Overall I prefer the first version.

danielmasny:
> To reduce the code duplication, how about this: https://github.com/danielmasny/ipa/compare/aliases-for-constants...andyleiserson:ipa:prover-configs?diff=split&w=
> 
> There are two commits on that branch with two different versions. One (the first commit) puts the `F: PrimeField` on the struct, and the other (the second commit / head of the branch) puts it on the functions. The benefit of putting it on the functions is it avoids `PhantomData`, but requires more type annotations where things are invoked. Putting it on the struct (and thus in the type alias) eliminates the most type annotations, but requires `PhantomData`.
> 
> Putting it on the struct also might enable caching the Lagrange table within the prover, but doing that is more involved due to the limitations on generic statics.
> 
> Overall I prefer the first version.



> To reduce the code duplication, how about this: https://github.com/danielmasny/ipa/compare/aliases-for-constants...andyleiserson:ipa:prover-configs?diff=split&w=
> 
> There are two commits on that branch with two different versions. One (the first commit) puts the `F: PrimeField` on the struct, and the other (the second commit / head of the branch) puts it on the functions. The benefit of putting it on the functions is it avoids `PhantomData`, but requires more type annotations where things are invoked. Putting it on the struct (and thus in the type alias) eliminates the most type annotations, but requires `PhantomData`.
> 
> Putting it on the struct also might enable caching the Lagrange table within the prover, but doing that is more involved due to the limitations on generic statics.
> 
> Overall I prefer the first version.

I also like the first version @benjaminsavage what do you think?

benjaminsavage:
I like @andyleiserson's first version best as well.

andyleiserson (approved):

benjaminsavage (approved):

PR #1109, opened 2024-06-03
Adding a few useful helper functions

benjaminsavage:
We will definitely need these two helper functions at the next stage. I've added a few tests for them that illustrate how they will be used and ensure the expected behavior.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -5,8 +5,11 @@ use std::{borrow::Borrow, iter::zip, marker::PhantomData};
 use crate::{
     ff::{Fp31, Fp61BitPrime, PrimeField},
     helpers::hashing::{compute_hash, hash_to_field},
-    protocol::ipa_prf::malicious_security::lagrange::{
-        CanonicalLagrangeDenominator, LagrangeTable,
+    protocol::{
+        context::Context,
+        ipa_prf::malicious_security::lagrange::{CanonicalLagrangeDenominator, LagrangeTable},
+        prss::SharedRandomness,
+        RecordId,
     },
 };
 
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -92,18 +95,52 @@ impl<F: PrimeField, const λ: usize, const P: usize, const M: usize> ProofGenera
 
         output
     }
+
+    fn gen_proof_shares_from_prss<C>(ctx: &C, record_counter: &mut RecordId) -> ([F; P], [F; P])
```
cberkhoff, line 99: I think it would be good to add documentation to ProofGenerator.

danielmasny (approved):
That works for me

andyleiserson:
Is there a way to use `left` and `right` instead of `.0` and `.1`? I understand not wanting to use `Replicated` / `AdditiveShare` since that's not really what these are. Maybe return a simple helper type something like `struct LeftRight<T> { left: T, right: T }` instead of a tuple? Or just deconstruct the return value as `(foo_left, foo_right)` immediately.

It might be possible to use the `StdArray` wrapper type for the proofs (instead of `[F; P]`), which would provide pre-built support for generation with PRSS, subtraction for `gen_other_proof_share`, and sending over channels. Maybe `ShareArray` is the better name for `StdArray` that I wish I had thought of earlier.

benjaminsavage:
These are all great suggestions. @danielmasny and I can follow up with them.

benjaminsavage:
@andyleiserson - I started trying to use StdArray, but I got stuck along the way. It immediately started spreading and before long the whole file needed StdArray throughout, not just in this one function.

Changing `gen_other_proof_share` was easy, but then I needed to change `gen_artefacts_from_recursive_step` which meant changing `gen_challenge_and_recurse` to accept StdArray, and I couldn't make it work there because `compute_hash` wants something it can iterate through and serialize, and I got stuck there.

In general though, are we sure there's a big benefit to this? Vanilla arrays are nice a readable, and the only benefit we seem to get from StdArray is that you *can't* access the elements.

andyleiserson:
I think it would be okay to add some things like `From`, `Into`, and `Deref` to `StdArray`. It wasn't needed for vectorization, but if we're using it for other things, as you found, it's hard to do much with it without some conversions.

Adding them to StdArray doesn't directly make them available to vectorized code in most cases, because the vectorized code is using the `SharedValueArray` trait rather than `StdArray` directly.

andyleiserson:
This is what I had in mind: https://github.com/private-attribution/ipa/commit/a9587f827189ed4aaba1afdb4f47da563230042e

PR #1112, opened 2024-06-03
Reveal wrapper

andyleiserson:
Create a wrapper around revealed values similar to the existing wrapper around downgraded values. Currently just unwrap things where needed, but there is some restructuring of protocols to consolidate reveal operations.

From talking with @danielmasny, it may be better to defer the reveal until after the DZKP has been validated. Possibly we can just do the reveal within the `validate` function, and avoid the wrapper entirely.

I am also changing reveal so that it is only implemented on "upgraded" contexts. This protects against inadvertent use of semi-honest protocols, which is a much greater concern in our new malicious security protocol where the share type is still `semi_honest::AdditiveShare`. In the previous protocol where shares were `malicious::AdditiveShare`, we were protected against inadvertently using semi-honest protocols because the non-upgraded context did not implement protocols for `malicious::AdditiveShare`. Removing protocol support from the base context requires changing a bunch of tests that used to call `TestWorld::semi_honest` to call `TestWorld::upgraded_semi_honest` instead. A similar change should be made for multiplication, but that can be done in a separate PR.

### DIFF
--- a/ipa-core/src/protocol/basics/check_zero.rs
+++ b/ipa-core/src/protocol/basics/check_zero.rs
@@ -2,7 +2,7 @@ use crate::{
     error::Error,
     ff::Field,
     protocol::{
-        basics::{mul::semi_honest_multiply, reveal::Reveal, step::CheckZeroStep as Step},
+        basics::{mul::semi_honest_multiply, semi_honest_reveal, step::CheckZeroStep as Step},
         context::Context,
         prss::{FromRandom, SharedRandomness},
         RecordId,
@@ -31,7 +31,7 @@ use crate::{
 /// 2.) If the randomly chosen value `r` is any other value in the field aside from zero, then the revealed value will NOT be zero.
 ///
 /// Clearly, the accuracy of this protocol is highly dependent on the field that is used.
-/// In a large field, like the integers modulo 2^31 - 1, the odds of `check_zero` returning "true"
+/// In a large field, like the integers modulo 2^31 - 1, the odds of `semi_honest_check_zero` returning "true"
 /// when `v` is NOT actually a sharing of zero is extrmely small; it is less than one two billion odds.
 /// In a silly field, like our test field of the integers modulo 31, the odds are very good. It'll incorrectly return "true"
 /// about 3.2% of the time.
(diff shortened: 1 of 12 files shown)


ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -17,10 +17,117 @@ use crate::{
...
+        });
+    try_join(send_left_fut, send_right_fut).await?;
+
+    if Some(ctx.role()) == excluded {
+        Ok(None)
+    } else {
+        let (share_from_left, share_from_right) = try_join(
+            left_receiver.receive(record_id),
+            right_receiver.receive(record_id),
+        )
```
benjaminsavage, line 118: Is it OK to await the `sends` first and only then await the `receives`? Wouldn't it be better instead to do all 4 in parallel with a try_join4?

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -169,7 +169,11 @@ where
         )
         .await?;
         match (&mut y, y_bit) {
-            (Some(y), Some(y_bit)) => y.push(y_bit),
+            (Some(y), Some(y_bit)) => {
+                // TODO: this should be replaced with real malicious security.
```
benjaminsavage, line 173: @danielmasny - what is appropriate to do here?

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -181,25 +187,34 @@ where
...
             ),
-        )
-        .try_flatten_iters()
-        .map_ok(bool::from)
-        .try_collect()
-        .await?;
+        );
+
+        let unwrapped_compare_results = {
+            // TODO: this should be replaced with a malicious verify.
```
benjaminsavage, line 204: @danielmasny ?

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -31,85 +31,81 @@ use crate::{
...
+pub async fn semi_honest_check_zero<C: Context, F: Field + FromRandom>(
+    ctx: C,
+    record_id: RecordId,
+    v: &Replicated<F>,
+) -> Result<bool, Error> {
     let r_sharing: Replicated<F> = ctx.prss().generate(record_id);
 
     let rv_share =
         semi_honest_multiply(ctx.narrow(&Step::MultiplyWithR), record_id, &r_sharing, v).await?;
+    // TODO: is it really okay to be doing a semi-honest reveal here?
```
benjaminsavage, line 55: Good question. I'm not sure. @danielmasny ? What should we do here?

benjaminsavage (changes requested):
Honestly, this PR is too long. It's hard to review like this. I feel like it could be split into a few smaller PRs.

1.) Introduce VectorizedProtocols and differentiate from Basic Protocols
2.) Upgrade many of the tests to use `world.upgraded_semi_honest(` instead of `world.semi_honest`
3.) Decide how we actually want to approach the use of `reveal` in Quicksort to ensure it is maliciously secure
4.) Decide how we actually want to approach the use of `reveal` in our Share conversion ABY code to ensure it is maliciously secure.
5.)  Decide how we want to approach the use of `reveal` in our "check zero" protocol, which is actually only used in a single place, the validation step of our MAC based validator (so maybe we should just inline the "check_zero" protocol there and not pretend like it's generic?)
6.) Add this wrapper around revealed values.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
danielmasny: Yes we need to change the code here. We need to run the integer_add and then validate. Only afterwards we can reveal. So this is a place where the input context should be an upgradable context such that we can upgrade it to get a validator, then we run integer_add on the upgraded context, we call validate afterwards we can drop the validator and upgraded context. Then we can use the upgradable context for the reveal. We still need to discuss how to handle safety properly. For example we do not want to allow calling reveal when there exists any unvalidated multiplication.

ipa-core/src/protocol/ipa_prf/quicksort.rs
danielmasny: Same as above, we compute the comparison, validate and then reveal. So we would probably call validated_seq_join on the comparison block or do our own customize seq join in chunks that calls validate in between compare and reveal

danielmasny (changes requested):
I assume this is still a draft and there are more changes to come. Have we resolved how we want to make the reveal safe?

PR #1113, opened 2024-06-04
num_bernoulli from epsilon, delta

bmcase:
Small PR to implement Thm 1 of this paper https://arxiv.org/pdf/1805.10559 which allows us to go from an epsilon, delta and other parameters to the minimum number of bernoulli samples we need to achieve that level of DP.
See this doc for more discussion. https://docs.google.com/document/d/116qQIgVoz3uV8axbmVWfif6dTjBVww5THecPqC0kERw/edit
Much of this PR is a translation into Rust of this earlier python script https://github.com/private-attribution/ipa/compare/main...bmcase:raw-ipa:binomial_variable_p

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/dp/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/dp/mod.rs
@@ -204,6 +204,7 @@ fn find_smallest_num_bernoulli(
     ell_2_sensitivity: f64,
     ell_infty_sensitivity: f64,
 ) -> u32 {
+    let mut smallest_num_bernoulli = 0;
     for num_bernoulli in 1..10_000_000 {
         if delta_constraint(
             num_bernoulli,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/dp/mod.rs
```diff
@@ -104,22 +99,214 @@ where
...
+    success_prob: f64,
+    delta: f64,
+    dimensions: f64,
+    quantization_scale: f64,
+    ell_1_sensitivity: f64,
+    ell_2_sensitivity: f64,
+    ell_infty_sensitivity: f64,
+) -> u32 {
+    let mut smallest_num_bernoulli = 0;
+    for num_bernoulli in 1..10_000_000 {
```
benjaminsavage, line 208: Woah... seriously?

Surely we can at least do a binary search?

benjaminsavage (changes requested):
Looping over 10 million is just too inefficient. Please convert to a binary search.

bmcase:
@benjaminsavage  converted to binary search for finding the smallest num_beroulli.

benjaminsavage (approved):

PR #1115, opened 2024-06-04
Simplify some things related to boolean arrays

andyleiserson:
* Add a BooleanArray trait as shorthand for `SharedValue + CustomArray<Element = Boolean>` (and a few other traits that are always implemented for boolean arrays).
* Delete ArrayAccessRef (not needed after #1049 and perhaps `impl DerefMut for BitDecomposed`).
* Remove ArrayAccess for Galois fields (perhaps not needed after fix for #1079).
* Remove some generics from some AdditiveShare-related things that are only used with Boolean arrays.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -9,7 +9,7 @@ use typenum::{U14, U2, U32, U8};
 
 use crate::{
     error::LengthError,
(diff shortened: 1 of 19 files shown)


benjaminsavage (approved):
I love it. Thanks @andyleiserson for simplifying the code base!

PR #1116, opened 2024-06-04
proof recursion in prover.rs

danielmasny:
This is a split from https://github.com/private-attribution/ipa/pull/1083

PR adds recursion function for the prover as well as an UVValues struct that keeps track of the length.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -21,12 +21,14 @@ pub type TestProofGenerator = ProofGenerator<Fp31, 4, 7, 3>;
 pub type SmallProofGenerator = ProofGenerator<Fp61BitPrime, 8, 15, 7>;
 pub type LargeProofGenerator = ProofGenerator<Fp61BitPrime, 32, 63, 31>;
 
+pub type UVValues<F, const N: usize> = Vec<([F; N], [F; N])>;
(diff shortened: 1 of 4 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -296,7 +450,16 @@ mod test {
...
+        let uv_before = U
+            .iter()
+            .zip(V.iter())
+            .map(|(u, v)| {
+                (
+                    Fp61BitPrime::truncate_from(*u),
+                    Fp61BitPrime::truncate_from(*v),
+                )
+            })
+            .collect::<UVValues<Fp61BitPrime, 32>>();
```
benjaminsavage, line 462: See how this code is duplicated with the code above on lines 411-420? When you have that sort of thing going on, it's a good indication that we should wrap this in an abstraction - like I previously did with "zip_chunks()". Maybe make a function to do this called "gen_uv_store()". In fact, this is serving a nearly identical purpose to "zip_chunks()", so maybe you should just eliminate the "zip_chunks()" method entirely. Then on line 378 where you are just calling the current "zip_chunks()" you could replace it with "gen_uv_store().iter()"

benjaminsavage (changes requested):
Looking good. A few small suggestions and then we can land this.

benjaminsavage (approved):

PR #1117, opened 2024-06-05
Moving DP related code

bmcase:
As promised to @martinthomson [earlier](https://github.com/private-attribution/ipa/pull/1082#discussion_r1617625652), here is a separate PR to move around the DP related code.  

I wanted to keep this as a separate PR for just moving code otherwise the diffs will be harder to review if there are also more substantive changes happening. 

Essentially what is happening here are two moves:
1. the code previously in `protocol/dp` is being moved into `protocol/ipa_prf/oprf_padding`.  The main code we will be using from here is related to oprf dummies so makes sense to be in the `ipa_prf` module.  There is some other code for noise generation in the clear that we probably won't be using once the new DP in MPC code is being used, but I'll come back and clean this up later. 
2. the code previously in `protocol/ipa_prf/dp` is being moved into `protocol/dp`.  As Martin noted this is really a general functionality and not necessary to have it in the `ipa_prf` module. 

All other changes are fixing imports.


benjaminsavage (approved):
I trust you

PR #1118, opened 2024-06-05
Add circular buffer implementation to be later used inside `OrderingSender`

akoshelev:
I've been running IPA at scale and one thing #1085 revealed is that network stack (hyper and kernel) does not really like the large buffers we've been sending after vectorization was enabled. @andyleiserson showed that keeping active work at 1024 results in 1.6Mb chunks sent down from the send buffers. Reducing active work to 32 eliminated the stall, but, according to my perf tests, it made our network utilization worse.

Fixing the size of `take`, allowed to increase active work parameter to extreme values (100k) and below are the results of those experiments (1M rows, real hardware, no match key encryption, HTTPS). 

Active work (items) | Max send chunk size (bytes) | Runtime (seconds) | Delta (from baseline) | Notes |
|-- | -- | -- | -- | -- |
1024 | 4096 | 438 | 0 | Baseline
32 | 1024 | 617 | -30% | #1073 change with old vectorization factor of 64 |
1024 | 2048 | 427 | +2% |  
65,536 | 4096 | 422 | +4% |  
65,536 | 8192 | 457 | -5% |  
65,536 | 2048 | 390 | +11% |  
65,536 | 1024 | 421 | +4% |  
32,768 | 2048 | 392 | +10% |  
16,384 | 2048 | 412 | +6% |  
100,000 | 2048 | 414 | +5% |  

This change aims to decouple the vectorization/active work parameters from network send buffers by using circular buffer implementation inside `OrderingSender`. It only brings the implementation of `CircularBuf` but I have the change to make `OrderingSender` use this buf sitting locally in my branch. 

The ultimate goal is obviously to allow more configuration for steps, but having constant size chunks dropped down to hyper send buffers seems useful anyway. 

## Circular buffer
Nothing fancy here, just a vector with read and write pointers. I've been contemplating whether I should make it thread-safe, but I am not convinced we need it in the long term if we pick shard-per-core model. It is not worse than our current implementation of `OrderingSender` because it also operates on a single allocation, but managing two pointers made it too complicated, so I decided to take the buffering business away from `OrderingSender`.

## Ordering sender
With this change we lose two features that `OrderingSender` currently supports

* spare. We always set spare to 0 before
* non-uniform write chunks. `OrderingSender` allowed randomly sized writes as long as they can fit into the buffer. It will no longer be supported if we go with `CircularBuf` - all writes must be of the same size. This is also how we use `OrderingSender` today.

Both of these features were never used, but it is important to note that they won't be available if we migrate to `CircularBuf`.

### DIFF
--- a/ipa-core/src/helpers/buffers/circular.rs
+++ b/ipa-core/src/helpers/buffers/circular.rs
@@ -100,10 +100,6 @@ impl CircularBuf {
             read_size % write_size == 0,
             "\"{write_size}\" write size must divide read_size \"{read_size}\""
         );
-        debug_assert!(
-            read_size <= capacity,
-            "read size \"{read_size}\" must be less than or equal to capacity \"{capacity}\""
-        );
         Self {
             write: 0,
             read: 0,
@@ -537,7 +533,9 @@ mod test {
         output.extend(Six::read_once(&mut buf).into_iter().map(usize::from));
         buf.next().write(&TwoBytes::from(&6));
         buf.next().write(&TwoBytes::from(&7));
+        assert!(!buf.is_closed());
         buf.close();
+        assert!(buf.is_closed());
 
         output.extend(Six::read_once(&mut buf).into_iter().map(usize::from));
 
@@ -776,7 +774,6 @@ mod test {
         }
 
         fn read_write(setup: BufSetup, ops: u32, seed: u64) {
-            println!("{setup:?}, {ops}, {seed}");
             let mut buf = CircularBuf::from(setup);
             let mut cnt = Wrapping::<u8>::default();
             let mut written = Vec::new();
--- a/ipa-core/src/helpers/buffers/circular.rs
+++ b/ipa-core/src/helpers/buffers/circular.rs
@@ -1,0 +1,788 @@
+use std::ops::RangeInclusive;
+
+use generic_array::GenericArray;
+use typenum::Unsigned;
+
+use crate::ff::Serializable;
+
+/// This is a specialized version of circular buffer implementation,
+/// tailored to what [`OrderingSender`] needs.
+///
+/// This construction requires one extra parameter, compared to
+/// traditional ring buffers: `read_size` that specifies the smallest
+/// continuous block of bytes that can be read off this buffer. It
+/// also requires the total `capacity` and the size of one write
+/// `write_size` to be provided at construction time. This allows it
+/// to use a single allocation.
+///
+/// For example: if buffer capacity is set to 32k and `read_size` is
+/// 4k, then buffer can only be read in blocks of 4k. `write_size` can
+/// be any value, aligned with 4k `read_size`.
+///
+/// The capacity of this buffer does not need to be a power of two, although
+/// you may want to have exactly that for performance of modulo operations
+/// widely used internally.
+///
+/// This allows using this buffer in scenarios where the working window
+/// is wide. If readers want smaller chunks of data, but can operate on
+/// them fast enough, then writers experience less interruptions hitting
+/// the capacity limit, because it can be set large enough.
+///
+/// This buffer can also be closed, using [`close`] method. After it is
+/// closed, it allows reads of any size, ubt it guarantees that all of them
+/// will be aligned with `write_size`.
+///
+/// ## Implementation notes
+/// This buffer is built over a [`Vec`] with two extra pointers that indicate
+/// the place to read from, and to write next chunk to. When read happens,
+/// it moves the read pointer until it meets the write pointer.
+/// When read points to the same location as write, this buffer is considered
+/// empty.
+///
+/// Both pointers operate within the range `[0, 2*capacity)` and clamped into
+/// the working range when used as index into the internal buffer. The reason
+/// is to be able to distinguish empty and full buffers.
+///
+/// This implementation does not perform checks in optimized builds,
+/// relying on [`OrderingSender`] to enforce correctness. If taken away,
+/// necessary adjustments need to be made to avoid data corruption.
+///
+/// ## Alternative implementations
+/// If alignment to `read_size` is too much, a [`BipBuffer`] can be used instead.
+///
+/// ## Future improvements
+/// [`OrderingSender`] currently synchronizes reader and writers, but it does not
+/// have to if this implementation is made thread-safe. There exists a well-known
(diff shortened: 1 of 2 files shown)


ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+/// you may want to have exactly that for performance of modulo operations
+/// widely used internally.
+///
+/// This allows using this buffer in scenarios where the working window
+/// is wide. If readers want smaller chunks of data, but can operate on
+/// them fast enough, then writers experience less interruptions hitting
+/// the capacity limit, because it can be set large enough.
+///
+/// This buffer can also be closed, using [`close`] method. After it is
+/// closed, it allows reads of any size, ubt it guarantees that all of them
```
benjaminsavage, line 32: %s/ubt/but

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+///
+/// ## Alternative implementations
+/// If alignment to `read_size` is too much, a [`BipBuffer`] can be used instead.
+///
+/// ## Future improvements
+/// [`OrderingSender`] currently synchronizes reader and writers, but it does not
+/// have to if this implementation is made thread-safe. There exists a well-known
+/// lock-free FIFO implementation for a single producer, single consumer that uses
+/// atomics for read and write pointers. We can't make use of it as is because there
+/// are more than one writer. However, [`OrderingSender`] already knows now to allow
```
benjaminsavage, line 58: Knows *not*

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+/// only one write at a time, so it could be possible to make the entire
```
benjaminsavage, line 59: More than one write at a time?

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+impl CircularBuf {
+    /// Constructs a new instance of [`CircularBuf`] with reserved `capacity` bytes and specified
+    /// `write_size` and `read_size` bytes.
+    ///
+    /// ## Panics
+    /// If any of the following conditions are met:
+    /// * Any provided value is 0
+    /// * `write_size` is not a multiple of `capacity`
+    /// * `read_size` is not a multiple of `write_size`
+    /// * `read_size` is smaller than `write_size`
```
benjaminsavage, line 90: Are you sure? I thought that was the whole idea, that we want small reads and big writes. Why would it panic?

benjaminsavage: Also, I don't see this enforced by any asserts below.

benjaminsavage (approved):
Looks good. A few small suggestions then merge it.

ipa-core/src/helpers/buffers/circular.rs
akoshelev: yea the whole sentence should be:  However, [`OrderingSender`] already knows **how to allow
 only one write** at a time

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+}
+
+impl BufWriteable for [u8] {
+    fn size(&self) -> usize {
+        self.len()
+    }
+
+    fn write(&self, data: &mut [u8]) {
+        data.copy_from_slice(self);
+    }
```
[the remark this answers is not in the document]
akoshelev: that is a good catch by CodeCov - I forgot to add `#[test]` annotation on property test

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+
+    /// Returns the number of bytes in this buffer.
+    pub fn len(&self) -> usize {
+        // Modulo arithmetic and wrapping/overflow rules in Rust
+        // make it difficult to write `(self.write - self.read) % 2*N`.
+        // It works well for power-of-two sizes, but for arbitrary
+        // buffer capacity, it is easier to use N - (a - b) because
+        // write is always ahead of read.
+        if self.write >= self.read {
+            self.wrap(self.write - self.read)
```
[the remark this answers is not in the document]
akoshelev: the difference between write pointer and read pointer can never be more than `C`, but both can grow outside of [0, C] range because we clamp them at 2*C. The reason for mod 2*C is to be able to distinguish between full and empty buffer.

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,788 @@
...
+    /// [`can_read`]: Self::can_read
+    pub fn take(&mut self) -> Vec<u8> {
+        if !self.can_read() {
+            return Vec::new();
+        }
+
+        // Capacity is always a multiple of write_size, so delta is always aligned.
+        let delta = std::cmp::min(self.read_size, self.len());
+
+        let mut ret = Vec::with_capacity(delta);
```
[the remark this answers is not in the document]
akoshelev: yea, HTTP needs an owned buffer to send and we need to free up the space here to allow more writes

ipa-core/src/helpers/buffers/circular.rs
akoshelev: it is enforced by `read_size % write_size == 0` check and validated by `panic_on_bad_read_size` test

`write_size` defines the size of a single write - usually just a few bytes, the size of a single record. In the vectorized case, this may be one kibibyte but still less than or equal to the size of a single read from this buffer. We don't want fragmentation here, so a single record should be entirely in one batch sent over the network

ipa-core/src/helpers/buffers/circular.rs
```diff
@@ -0,0 +1,809 @@
...
+            capacity % write_size == 0,
+            "\"{write_size}\" write size must divide capacity \"{capacity}\""
+        );
+        debug_assert!(
+            read_size % write_size == 0,
+            "\"{write_size}\" write size must divide read_size \"{read_size}\""
+        );
+        Self {
+            write: 0,
+            read: 0,
```
akoshelev, line 105: I don't understand what is going on with coverage, but I double checked that `len` code and other places are covered

PR #1122, opened 2024-06-07
Challenge for DZKP Verification

danielmasny:
compute challenge + test + move BatchToVerify

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
+++ b/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
@@ -1,0 +1,508 @@
+use futures_util::future::try_join;
+
+use crate::{
+    error::Error,
+    ff::Fp61BitPrime,
+    helpers::{
+        hashing::{compute_hash, hash_to_field, Hash},
+        Direction,
+    },
+    protocol::{
+        context::{dzkp_field::UVTupleBlock, Context},
+        ipa_prf::{
+            malicious_security::{
+                lagrange::{CanonicalLagrangeDenominator, LagrangeTable},
+                prover::{LargeProofGenerator, SmallProofGenerator},
+            },
+            validation_protocol::proof_generation::ProofBatch,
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -0,0 +1,508 @@
...
+        (
+            challenges_from_proof_of_left_prover,
+            challenges_from_proof_of_right_prover,
+        )
+    }
+}
+
+struct ProofHash {
+    proof_from_prover_left_hash: Vec<Hash>,
+    proof_from_prover_right_hash: Vec<Hash>,
```
benjaminsavage, line 222: These are vectors of hashes, so it seems odd to call them "proof". I'd suggest a name like: `hashes_for_prover_left` and `hashes_for_prover_right`

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -0,0 +1,508 @@
...
+        // compute other hashes
+        // we don't need to hash last proof
+        for i in 0..(length - 1) {
+            proof_from_prover_left_hash.push(compute_hash(
+                batch_to_verify.proofs_from_left_prover[i].iter(),
+            ));
+            proof_from_prover_right_hash.push(compute_hash(
+                batch_to_verify.proofs_from_right_prover[i].iter(),
+            ));
+        }
```
benjaminsavage, line 255: This `ProofHash` class has a LOT of duplicate code. It does the same thing for the left and then again for the right. It makes me feel like it would be better to change this struct to only contain hashes for one side, and then just instantiate two of them.

benjaminsavage (approved):
Looks pretty good. We still need to improve the tests, but we can follow up with another PR for that.

I have a few suggestions, but I'll just send those in a PR that you can review.

Issue #1120, opened 2024-06-07
Report collector times out talking to MPC ring when it handles large inputs


akoshelev:
When running IPA on inputs with 20M rows or more, RC binary times out executing `query_status` command

```bash
 "Failed to get query status from helper 1: Failed to connect to ec2-44-234-150-14.us-west-2.compute.amazonaws.com:443: http2 error: keep-alive timed out"
```

akoshelev:
Here is the client view of the world

```bash
2024-06-07T02:03:39.519115Z  INFO ipa_core::cli::playbook::ipa: submitting inputs to MPC helpers
...
2024-06-07T02:03:41.412501Z  INFO ipa_core::cli::playbook::ipa: Poll helpers for MPC status
2024-06-07T02:03:41.634659Z  INFO ipa_core::cli::playbook::ipa: Poll helpers for MPC status complete
2024-06-07T02:03:43.635482Z  INFO ipa_core::cli::playbook::ipa: Poll helpers for MPC status
2024-06-07T02:05:31.630683Z ERROR ipa_core::error: ThreadId(1) "main" panicked at /home/ec2-user/workspace/alex-ipa/ipa-core/src/cli/playbook/ipa.rs:138:14:
called `Result::unwrap()` on an `Err` value: "Failed to get query status from helper 0: Failed to connect to ec2-44-234-39-30.us-west-2.compute.amazonaws.com:443: http2 error: keep-alive timed out"
```

this is what server 1 is seeing

```bash
2024-06-07T02:03:39.526867Z  INFO oprf_ipa_query{sz=20000000}: ipa_core::query::runner::oprf_ipa: new
2024-06-07T02:03:39.526893Z  INFO oprf_ipa_query{sz=20000000}: ipa_core::query::runner::oprf_ipa: New query: IpaQueryConfig { per_user_credit_cap: 8, max_breakdown_key: 20, attribution_window_seconds: None, num_multi_bits: 3, plaintext_match_keys: true }
...
2024-06-07T02:03:41.627086Z  INFO ipa_core::query::processor: query_status: waiting for lock
2024-06-07T02:03:41.627104Z  INFO ipa_core::query::processor: query_status: got the lock
2024-06-07T02:03:42.021131Z  INFO oprf_ipa_query{sz=20000000}:shuffle_inputs: ipa_core::protocol::ipa_prf::shuffle: new
[...]

Issue #1100, opened 2024-05-29
IPA fails on large inputs with saturating error


akoshelev:
[here](https://draft-mpc.vercel.app/query/view/dicey-mango-9279) is an example of such run

```bash
Coordinator: 2024-05-29T02:09:55.095547Z  INFO ipa_core::cli::playbook::ipa: Running IPA for 5000000 records took 4177.366415084s
Coordinator: 2024-05-29T02:09:55.104605Z  INFO report_collector: IpaQueryConfig { per_user_credit_cap: 64, max_breakdown_key: 64, attribution_window_seconds: None, num_multi_bits: 3, plaintext_match_keys: true }
Coordinator: 2024-05-29T02:09:55.104682Z  INFO ipa_core::cli::playbook: 
Coordinator: +-----+--------------+-------------+-------+
Coordinator: | Row | Expected     | Actual      | Diff? |
Coordinator: +==========================================+
Coordinator: | 0   | Some(106989) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 1   | Some(105657) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 2   | Some(105677) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 3   | Some(105051) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 4   | Some(106799) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 5   | Some(106926) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 6   | Some(104752) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 7   | Some(104977) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 8   | Some(104764) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 9   | Some(104801) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 10  | Some(104571) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
Coordinator: | 11  | Some(106523) | Some(65535) | X     |
Coordinator: |-----+--------------+-------------+-------|
[...]

PR #1123, opened 2024-06-08
Hand off Tokio bookkeeping tasks before starting a new IPA query

akoshelev:
This is an attempt to mitigate #1120 by hinting Tokio executor that query task maybe blocking the worker thread for longer than expected and it should hand off everything from that worker thread.

Here is how this issue is reproducible in our environment

Tokio executor is a work-stealing scheduler that optimizes for low latency and
        resource utilization. It optimizes for low scheduling overhead and uses the same worker
        threads for polling tasks and to service IO and timer operations, if they are enabled.
        This leads to very low overhead, but also to an unfortunate possibility of tasks that are
        long-running to block the entire runtime, even if it is multithreaded.
        Here is an example in IPA code that led to this discovery:
        - IPA query kicks in and first step it performs a very large reallocation, keeping the
        task running for minutes before yielding.
        - At the time when query started, there was only one active thread in the runtime, and
        it began polling this task
        - because other threads are parked, there is nothing that can indicate to the executor
        that there are more tasks that are ready. Tokio Runtime is not polling any task.
        - Client asks for IPA query status and times out, because IPA application is not
        responding.
        
  Ultimately tokio team decided that it is not a bug in the implementation and made it
  clear that application needs to take care of it. There are a few ways to deal with it,
  one of them is to signal the executor that this task is about to block the worker thread
  for long time, so it has a chance to move the IO driver and pending tasks to other
  threads. This is what this proposed solution does.

### DIFF
--- a/ipa-core/src/query/executor.rs
+++ b/ipa-core/src/query/executor.rs
@@ -135,8 +135,42 @@ where
             .await
             .unwrap();
 
-        tx.send(query_impl(&prss, &gateway, &config, input_stream).await)
-            .unwrap();
+        // Tokio executor is a work-stealing scheduler that optimizes for low latency and
+        // resource utilization. It optimizes for low scheduling overhead and uses the same worker
+        // threads for polling tasks and to service IO and timer operations, if they are enabled.
+        // This leads to very low overhead, but also to an unfortunate possibility of tasks that are
+        // long-running to block the entire runtime, even if it is multithreaded.
+        // Here is an example in IPA code that led to this discovery:
+        // - IPA query kicks in and first step it performs a very large reallocation, keeping the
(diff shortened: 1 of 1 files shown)


akoshelev:
Draft run for 50M records (was timing out before): https://draft-mpc.vercel.app/query/view/paved-stork-5287

akoshelev:
it does not hang anymore, fails because of #1100

akoshelev:
I decided not to create an executor module for now - `block_in_place` is a specialized API and maybe we use some other runtime sooner than we run into another issue. If `block_in_place` needs to be used somewhere else, then a new module will be justified

andyleiserson (approved):

PR #1124, opened 2024-06-09
cleaning up the challenge generation code

benjaminsavage:
Just some refactoring to remove code duplication and mutable objects.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
+++ b/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
@@ -1,3 +1,6 @@
+use std::iter::{once, repeat};
+
+use futures::future::try_join4;
 use futures_util::future::try_join;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -163,178 +166,129 @@ impl BatchToVerify {
...
+            .map(|((hash_left, hash_right), exclude)| {
+                hash_to_field(hash_left, hash_right, exclude)
+            });
+
+        // From the perspective of the *prover_right*, _left_ is this helper and _right_ is the other verifier
+        let challenges_for_prover_right = my_hashes_prover_right
+            .hashes
+            .iter()
+            .zip(other_hashes_prover_right.hashes.iter())
+            .zip(once(exclude_large).chain(repeat(exclude_small)))
```
danielmasny, line 197: I like it! It is much cleaner.

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -163,178 +166,129 @@ impl BatchToVerify {
...
-        ));
-        for (left, right) in other_hashes
-            .proof_from_prover_left_hash
-            .iter()
-            .zip(hashes.proof_from_prover_left_hash.iter())
-        {
-            challenges_from_proof_of_left_prover.push(hash_to_field(left, right, exclude_small));
-        }
+        let my_hashes_prover_left = ProofHashes::generate_hashes(self, Side::Left);
+        let my_hashes_prover_right = ProofHashes::generate_hashes(self, Side::Right);
```
danielmasny, line 170: Splitting it up is a great idea!

danielmasny (approved):
Thanks for cleaning it up! Splitting left and right makes a lot of sense and removes lots of the duplicate code. 

I noticed that the Miri test fails which does not seem related to this PR (it fails for other PRs as well for the same reason). Maybe we should fix that before we merge more PRs.

PR #1125, opened 2024-06-10
fix miri errors

danielmasny:
Miri checks fail due to errors in main

### DIFF
--- a/ipa-core/src/ff/galois_field.rs
+++ b/ipa-core/src/ff/galois_field.rs
@@ -3,10 +3,7 @@ use std::{
     ops::Index,
 };
 
-use bitvec::{
-    prelude::{bitarr, BitArr, Lsb0},
-    slice::Iter,
-};
+use bitvec::prelude::{bitarr, BitArr, Lsb0};
 use generic_array::GenericArray;
 use typenum::{Unsigned, U1, U2, U3, U4, U5};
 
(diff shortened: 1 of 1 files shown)


danielmasny:
BoolIterator seems to be the only issue that caused miri to fail. I have added an issue to remove the dead code flag: https://github.com/private-attribution/ipa/issues/1126.

ipa-core/src/ff/galois_field.rs
```diff
@@ -139,6 +139,7 @@ fn clmul<GF: GaloisField>(a: GF, b: GF) -> u128 {
 /// is that this type is not `Send`.
 ///
 /// [`BitValIter`]: bitvec::slice::BitValIter
+#[allow(dead_code)]
```
akoshelev, line 142: is it used anywhere? if not, would be better to delete this code

danielmasny: I agree. I was wondering whether Andy plans to use it in the future. Otherwise, I would have deleted it. He made some changes end of January. @andyleiserson is it ok to delete or do you still intend to use it?

andyleiserson: I think it should have been removed with #1115. Not sure why the miri check didn't fail on that PR though.

danielmasny: @akoshelev @andyleiserson Ok, I have removed it.

andyleiserson (approved):
Thanks for taking care of this.

akoshelev (approved):

PR #1128, opened 2024-06-11
Revise BasicProtocols

andyleiserson:
The old BasicProtocols was no longer used anywhere. This replaces it with a version that includes only the basic protocols that have been vectorized. It also makes some progress towards not exposing semi-honest protocols on arbitrary contexts (which is unsafe).

Split from #1112.

### DIFF
--- a/ipa-core/src/ff/ec_prime_field.rs
+++ b/ipa-core/src/ff/ec_prime_field.rs
@@ -12,7 +12,9 @@ use crate::{
         prss::{FromPrss, FromRandom, PrssIndex, SharedRandomness},
     },
     secret_sharing::{
-        replicated::{semi_honest::AdditiveShare, ReplicatedSecretSharing},
+        replicated::{
+            malicious::ExtendableField, semi_honest::AdditiveShare, ReplicatedSecretSharing,
+        },
         Block, FieldVectorizable, SharedValue, StdArray, Vectorizable,
     },
 };
@@ -217,6 +219,14 @@ impl Field for Fp25519 {
     const NAME: &'static str = "Fp25519";
 
     const ONE: Fp25519 = Fp25519::ONE;
+}
+
+impl ExtendableField for Fp25519 {
+    type ExtendedField = Self;
+
+    fn to_extended(&self) -> Self::ExtendedField {
+        *self
+    }
 }
 
 impl FromRandom for Fp25519 {
--- a/ipa-core/src/protocol/basics/mod.rs
+++ b/ipa-core/src/protocol/basics/mod.rs
@@ -18,32 +18,44 @@ pub use share_known_value::ShareKnownValue;
 
 use crate::{
     const_assert_eq,
-    ff::{boolean::Boolean, PrimeField},
+    ff::{boolean::Boolean, ec_prime_field::Fp25519, PrimeField},
     protocol::{
-        context::{
-            Context, SemiHonestContext, UpgradedMaliciousContext, UpgradedSemiHonestContext,
-        },
+        context::{Context, SemiHonestContext, UpgradedSemiHonestContext},
         ipa_prf::{AGG_CHUNK, PRF_CHUNK},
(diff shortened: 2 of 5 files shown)


ipa-core/src/ff/ec_prime_field.rs
```diff
@@ -219,6 +221,14 @@ impl Field for Fp25519 {
     const ONE: Fp25519 = Fp25519::ONE;
 }
 
+impl ExtendableField for Fp25519 {
```
andyleiserson, line 224: `ExtendableField` is automatically implemented for `PrimeField`, but `PrimeField` assumes less than 128 bits, which wouldn't work here.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -115,18 +111,20 @@ where
...
-    ctx: SemiHonestContext<'_>,
+pub async fn quicksort_ranges_by_key_insecure<C, K, F, S>(
+    ctx: C,
     list: &mut [S],
     desc: bool,
     get_key: F,
     mut ranges_to_sort: Vec<Range<usize>>,
 ) -> Result<(), Error>
 where
+    C: UpgradableContext,
```
andyleiserson, line 122: We will need this to implement malicious security for quicksort, because it needs to do intermediate validates and reveals.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -181,16 +179,16 @@ where
...
-                        let comparison = compare_gt::<_, ThirtyTwoBitStep, SORT_CHUNK>(
+                        // Compare elements against pivot
+                        let comp = compare_gt::<_, ThirtyTwoBitStep, SORT_CHUNK>(
                             cmp_ctx, record_id, &k, &pivot,
                         )
-                        .await?
-                        .reveal(rvl_ctx, record_id) // reveal outcome of comparison
                         .await?;
+                        // Reveal the comparison result
+                        let revealed_comp = reveal(rvl_ctx, record_id, &comp).await?;
```
andyleiserson, line 188: There is no functional difference here, but calling the free function is necessary to work around rustc issue 100013.

ipa-core/src/ff/ec_prime_field.rs
danielmasny: yes PrimeField stores the prime in a u128

danielmasny (approved):
looks good to me.

ipa-core/src/protocol/basics/mod.rs
```diff
@@ -18,32 +18,44 @@ pub use share_known_value::ShareKnownValue;
...
+}
+
+// For PRF test
+impl<'a, B: ShardBinding> BasicProtocols<UpgradedSemiHonestContext<'a, B, Fp25519>, Fp25519>
+    for AdditiveShare<Fp25519>
+{
+}
+
+impl<'a, B: ShardBinding>
+    BasicProtocols<UpgradedSemiHonestContext<'a, B, Fp25519>, Fp25519, PRF_CHUNK>
```
benjaminsavage, line 51: Why this type of context?

benjaminsavage (approved):
Lgtm

ipa-core/src/protocol/basics/mod.rs
andyleiserson: The idea here is that it should be supported for any context that satisfies `C: UpgradedContext`, although the combinations may not all be implemented if we don't need them anywhere.

andyleiserson:
I will merge this via #1129 since that one passed the coverage check. The missing coverage is on:

```
    fn to_extended(&self) -> Self::ExtendedField {
        *self
    }
```

Not very interesting, and in any case, it should be covered later when we use MAC-based malicious security for Fp25519, for PRF evaluation.

PR #1130, opened 2024-06-12
split u_or_v recursion and verify_proof

danielmasny:
split u_or_v recursion and verify_proof in verifier.rs


benjaminsavage (approved):
As we discussed, this approach makes sense to me. Lots of follow up work to do, but this is a good first step.

PR #1131, opened 2024-06-12
Further simplifying the verifier code

benjaminsavage:
We can further simplify the verifier code by just breaking down the `verify_proof` function into the two separate components.

I'm sorry the diff view is so unhelpful... I removed the ProofVerifier struct as it's no longer needed, and cargo fmt moved everything back one indentation... which is displayed so badly in the diff view.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
@@ -8,38 +8,35 @@ use crate::{
 };
 
 #[allow(non_upper_case_globals)]
-pub struct ProofVerifier<F: PrimeField> {
-    zero_share: F,
-}
+pub struct ProofVerifier {}
 
 ///
 /// Distributed Zero Knowledge Proofs algorithm drawn from
 /// `https://eprint.iacr.org/2023/909.pdf`
 ///
 #[allow(non_upper_case_globals)]
-impl<F> ProofVerifier<F>
-where
-    F: PrimeField,
-{
-    /// This function outputs an `out_share` and a `ProofVerifier` that consists of a `zero_share`.
-    pub fn verify_proof<const λ: usize, const P: usize>(
-        out_share: F,
+impl ProofVerifier {
+    /// This function interprets the zkp as points on a polynomial, and interpolates the
+    /// value of this polynomial at the provided value of `r`
+    pub fn interpolate_at_r<F: PrimeField, const λ: usize, const P: usize>(
         zkp: &[F; P],
         r: F,
-    ) -> (F, Self) {
-        let denominator_g = CanonicalLagrangeDenominator::<F, P>::new();
-        let lagrange_table_g = LagrangeTable::<F, P, 1>::new(&denominator_g, &r);
-        let g_r_share = lagrange_table_g.eval(zkp)[0];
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -8,38 +8,35 @@ use crate::{
 };
 
 #[allow(non_upper_case_globals)]
-pub struct ProofVerifier<F: PrimeField> {
-    zero_share: F,
-}
+pub struct ProofVerifier {}
```
akoshelev, line 11: there may be some further work on this abstraction - I can't say exactly what is going on here, but having a lot of methods that don't take `self` as an argument is suspicious

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -8,38 +8,35 @@ use crate::{
...
-        // Reveal `b_share` to one another to reconstruct `b` and check if `b = 0`. If the check doesn't pass, abort.
-        let zero_share = sum_share - out_share;
+        lagrange_denominator: &CanonicalLagrangeDenominator<F, P>,
+    ) -> F {
+        let lagrange_table_g = LagrangeTable::<F, P, 1>::new(lagrange_denominator, &r);
+        lagrange_table_g.eval(zkp)[0]
+    }
 
-        (g_r_share, Self { zero_share })
+    /// This function outputs an `out_share` and a `ProofVerifier` that consists of a `zero_share`.
```
akoshelev, line 30: this seems to produce only a single field value

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
benjaminsavage: Comment is wrong... will fix.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
benjaminsavage: It's just a collection of static methods. There is no need for any struct, just some place to put them.

akoshelev: I would say it is not idiomatic in Rust to keep stateless structs - OOP went with it because everything has to be a class, but in Rust we can just use standalone functions. structs usually represent some sort of a state, either at compile time or at runtime. 

I was thinking that we are probably missing some abstractions here, but I can't recommend anything at this point.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -7,75 +7,74 @@ use crate::{
...
-#[allow(non_upper_case_globals)]
-pub struct ProofVerifier<F: PrimeField> {
-    zero_share: F,
-}
-
 ///
 /// Distributed Zero Knowledge Proofs algorithm drawn from
 /// `https://eprint.iacr.org/2023/909.pdf`
 ///
 #[allow(non_upper_case_globals)]
```
akoshelev, line 14: you could also define this suppression at the beginning of this module, so you don't have to repeat yourself

```rust
#![allow(non_upper_case_globals)]
```

PR #1132, opened 2024-06-12
Adding dev docs describing query management.

cberkhoff:
MPC Computation piece will be added once I finish improved aggregation to get a better understanding of the pitfalls.


akoshelev (approved):
Would be great to clarify the role of report collector before merging this

PR #1133, opened 2024-06-12
Integrating DP noise for OPRF IPA

bmcase:
This PR integrates adding DP into the `oprf_ipa` function.  The DP parameters are also added to come as query inputs. 

In a future PR we'll add a feature flag and environment variables to make it harder for the no-DP option for testing to be used accidentally in production. 

Currently two network tests are failing, but creating a PR so folks can start to review.


ipa-core/src/protocol/dp/mod.rs
```diff
@@ -98,6 +100,62 @@ where
...
+            let epsilon = query_epsilon;
+            let delta = 1e-6;
+            let success_prob = 0.5;
+            let dimensions = 1.0;
+            let quantization_scale = 1.0;
+            let per_user_credit_cap = 2_f64.powi(i32::try_from(SS_BITS).unwrap());
+            let ell_1_sensitivity = per_user_credit_cap;
+            let ell_2_sensitivity = per_user_credit_cap;
+            let ell_infty_sensitivity = per_user_credit_cap;
+            let num_bernoulli = find_smallest_num_bernoulli(
```
cberkhoff, line 140: I think it would be cleaner if you had a struct that contained all these parameters. This would get rid of `#[allow(clippy::too_many_arguments)]` in `find_smallest_num_bernoulli`.

You could also store the constants for delta, sucess_prob and such as part of that same struct or module so that if you want to provide some default implementations, it can be done as part of a new/default function and such.

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -220,8 +222,95 @@ impl AsRef<str> for QueryType {
...
+            && self.attribution_window_seconds == other.attribution_window_seconds
+            && self.num_multi_bits == other.num_multi_bits
+            && self.dp_params == other.dp_params
+            && self.plaintext_match_keys == other.plaintext_match_keys
+    }
+}
+
+// TODO switch to using an enum
+#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum DPParams {
```
cberkhoff, line 239: Naming nit: Don't like the DP acronym too much. Also "params" doesn't add a lot of substance to the name. I don't know the space well enough to provide a better name.

Why not just have functions pass an `Option<DifferentialPrivacy>` instead of having an explicit no enum? If you use an Option you will get all those helper functions for free.

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -220,8 +222,95 @@ impl AsRef<str> for QueryType {
...
+            DPParams::TestingWithNoDP => write!(f, "TestingWithNoDP"),
+            DPParams::WithDP(eps) => write!(f, "WithDP={eps}"),
+        }
+    }
+}
+// impl ToString for DPParams {
+//     fn to_string(&self) -> String {
+//         self.to_string()
+//     }
+// }
```
cberkhoff, line 294: remove?

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -127,7 +127,8 @@ pub async fn aggregate_contributions<C, St, BK, TV, HV, const B: usize, const N:
     ctx: C,
     contributions_stream: St,
     contributions_stream_len: usize,
-) -> Result<Vec<Replicated<HV>>, Error>
+) -> Result<BitDecomposed<Replicated<Boolean, B>>, Error>
+// -> Result<Vec<Replicated<HV>>, Error>
```
cberkhoff, line 131: remove?

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -393,10 +407,12 @@ pub mod tests {
                 test_input(0, 68362, false, 1, 0),
                 test_input(20, 68362, true, 0, 2),
             ];
-
+            let dp_params = DPParams::TestingWithNoDP;
+            // let testing_with_do_dp = true;
+            // let query_epsilon = -1.0;
```
cberkhoff, line 412: remove?

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -440,12 +456,19 @@ pub mod tests {
             ];
 
             records.shuffle(&mut thread_rng());
-
+            let dp_params = DPParams::TestingWithNoDP;
+            // let testing_with_do_dp = true;
+            // let query_epsilon = -1.0;
```
cberkhoff, line 461: remove?

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -440,12 +456,19 @@ pub mod tests {
...
             let mut result: Vec<_> = world
                 .semi_honest(records.into_iter(), |ctx, input_rows| async move {
-                    oprf_ipa::<BA8, BA3, BA16, BA20, 5, 256>(ctx, input_rows, None)
-                        .await
-                        .unwrap()
+                    oprf_ipa::<BA8, BA3, BA16, BA20, 5, 256>(
+                        ctx, input_rows, None,
+                        dp_params,
+                        // testing_with_do_dp,
+                        // query_epsilon,
```
cberkhoff, line 468: remove?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -404,7 +405,8 @@ pub async fn attribute_cap_aggregate<'ctx, BK, TV, HV, TS, const SS_BITS: usize,
     input_rows: Vec<PrfShardedIpaInputRow<BK, TV, TS>>,
     attribution_window_seconds: Option<NonZeroU32>,
     histogram: &[usize],
-) -> Result<Vec<Replicated<HV>>, Error>
+) -> Result<BitDecomposed<Replicated<Boolean, B>>, Error>
+// -> Result<Vec<Replicated<HV>>, Error>
```
cberkhoff, line 409: remove?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -980,26 +986,29 @@ pub mod tests {
...
+                result_reconstructed
                     .iter()
                     .map(U128Conversions::as_u128)
                     .collect::<Vec<_>>(),
                 &expected
             );
         });
     }
-
+    // attribute_cap_aggregate<'ctx, BK, TV, HV, TS, const SS_BITS: usize, const B: usize>(
```
cberkhoff, line 1011: remove?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -438,7 +440,10 @@ where
     // Chunk the incoming stream of records into stream of vectors of records with the same PRF
     let mut input_stream = stream::iter(input_rows);
     let Some(first_row) = input_stream.next().await else {
-        return Ok(vec![]);
+        // return Ok(vec![]);
```
cberkhoff, line 443: remove?

ipa-core/src/helpers/transport/query/mod.rs
akoshelev: Maybe we should follow MS convention and use `Dp` instead. `DpParams` looks better

martinthomson: Acronyms are words in [the rust style](https://rust-lang.github.io/api-guidelines/naming.html).  So `DpParams` it is.

bmcase: Sounds good. changing to `DpParams` also `WithDp`, `NoDp`

ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
bmcase: yep, thanks will clean these up.

ipa-core/src/bin/report_collector.rs
```diff
@@ -131,7 +131,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
         } => gen_inputs(count, seed, args.output_file, gen_args)?,
         ReportCollectorCommand::ApplyDpNoise(ref dp_args) => apply_dp_noise(&args, dp_args)?,
         ReportCollectorCommand::OprfIpa(config) => {
-            println!("#############################################{}",config.dp_params);
+            println!(
+                "#############################################{}",
+                config.dp_params
+            );
```
martinthomson, line 137:
```suggestion
```

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -234,23 +234,22 @@ impl PartialEq for IpaQueryConfig {
...
 }
 
-// TODO switch to using an enum
 #[derive(Debug, Copy, Clone)]
-pub enum DPParams {
-    TestingWithNoDP,
-    WithDP(f64),
+pub enum DpParams {
+    NoDp,
+    WithDp(f64),
```
martinthomson, line 240:
```suggestion
    WithDp{
      epsilon: f64,
    }
```

Generally `.epsilon` is easier to read than `.0`.  The effect is the same.

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -260,43 +259,45 @@ impl FromStr for DPParams {
...
-        Self::from_str(s).map_err(|e| de::Error::custom(format!("failed to deserialize DpParams object: {e}")))
+        Self::from_str(s)
+            .map_err(|e| de::Error::custom(format!("failed to deserialize DpParams object: {e}")))
     }
 }
 
 #[cfg(test)]
-impl PartialEq<Self> for DPParams {
-    fn eq(&self, other: &DPParams) -> bool {
+impl PartialEq<Self> for DpParams {
```
martinthomson: Isn't this something you can use `#[derive(PartialEq)]` for?

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -260,43 +259,45 @@ impl FromStr for DPParams {
...
 }
 
-impl Serialize for DPParams {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: Serializer {
+impl Serialize for DpParams {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
         serializer.serialize_str(&self.to_string())
```
martinthomson, line 276: Why have this as a string and not a float?  NoDp can be Infinity.

ipa-core/src/helpers/transport/query/mod.rs
bmcase: okay, yeah that will make it more readable.

ipa-core/src/helpers/transport/query/mod.rs
bmcase: we'd been having trouble with serialization and @akoshelev helped come up with this to fix it.  But I think he also thought it was something we should revisit and try some other approaches.

bmcase (commented):
incorporate Martin's feedback

ipa-core/src/helpers/transport/query/mod.rs
bmcase: I can go back and see if that will work again.  Some merge conflict or something breaking things online I need to fix first.

bmcase: yeah, looks like it works with this derived.

ipa-core/src/helpers/transport/query/mod.rs
bmcase: In fact @akoshelev  created an issue related to this. https://github.com/private-attribution/ipa/issues/1149#issue-2360310674  

I'd hope we can get it working with the current approach for this PR and then align on what we want to do longer term.

ipa-core/src/protocol/dp/mod.rs
bmcase: Thanks, that's a good suggestion. I agree there is still more we could do to clean up passing around these parameters.  I think I'd like to clean things up with changes like that in a later PR and do after we'd incorporated the noise for `feature_label_dot_product` as after that things will be more settled and it can be clearer what can be defaults in what cases.

ipa-core/src/helpers/transport/query/mod.rs
martinthomson: Works for me.

akoshelev: yea, this is ugly. Ben is right - it all comes down to our earlier decision to pass query params through the query string. The serialization options are extremely limited there: `serde_urlencoded` supports just plain structs and simple enums, failing at runtime for anything more complicated than that.

Once we fix #1149, we can eliminate this custom serializer

bmcase:
All tests are passing except these tests for compact gate which are failing with this error: 
 `unexpected narrow for ProtocolGate(/ipa_prf/d_p/noise_gen) => AggregationStep(aggregate00)`

having run one of the tests as 
`cargo test --test compact_gate --lib compact_gate_cap_8_no_window_semi_honest -p ipa-core --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"`

I'm not really sure about the latest approach to compact gates, but is there something particular that needs to be done to support it when adding new steps? cc @akoshelev  @benjaminsavage @martinthomson do you know what may be going on?

martinthomson:
Try this:
```rust
#[derive(CompactStep)]
pub(crate) enum DPStep {
  #[step(child = AggregationStep)]
  NoiseGen,
  ApplyNoise,
}
```

bmcase:
Thanks @martinthomson , I think that works with another child of AppleNoise also
```
#[derive(CompactStep)]
pub(crate) enum DPStep {
    #[step(child = crate::protocol::ipa_prf::aggregation::step::AggregationStep)]
    NoiseGen,
    #[step(child = crate::protocol::boolean::step::SixteenBitStep)]
    ApplyNoise,
}
```

ipa-core/src/protocol/dp/mod.rs
bmcase: updated with a `NoiseParams` struct to hold all these internal parameters.

bmcase:
Finished refactoring the DP module to use a struct `NoiseParams` to hold all the internal dp parameters. 

After trying many things, was not able to get clap to parse a complex enum coming directly from the command line, such as 
```
   #[cfg_attr(feature = "clap", arg(long, default_value = "WithDp{3.0}"))]
   pub dp_params: DpParams,
```
so I instead pass two parameters `with_dp` and `epsilon` and once read in construct the `dp_params` enum from those.  I can comment with more to this issue https://github.com/private-attribution/ipa/issues/1149 about what we might like to rather do if we revisit parsing query params in the future.  

Tests/checks are all passing now.  I tried to create draft runs but that is crashing even on main right now.

martinthomson:
This approach seems reasonable.  One option you might try if you want to take another swing: make the parameter `pub epsilon: Option<f64>` and then `impl From<Option<f64>> for DpParams`.

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -253,7 +259,36 @@ pub async fn test_oprf_ipa<F>(
...
+
+            let num_bernoulli = crate::protocol::dp::find_smallest_num_bernoulli(&noise_params);
+            let mean: f64 = f64::from(num_bernoulli) * 0.5; // n * p
+            let standard_deviation: f64 = (f64::from(num_bernoulli) * 0.5 * 0.5).sqrt(); //  sqrt(n * (p) * (1-p))
+            for (index, sample) in result.iter().enumerate() {
+                assert!(
+                    f64::from(*sample) - mean
+                        > f64::from(expected_results[index]) - 5.0 * standard_deviation
+                        && f64::from(*sample) - mean
+                        < f64::from(expected_results[index]) + 5.0 * standard_deviation
```
martinthomson, line 286: This is not obviously correct, because it is a little tricky to read.

I think that you could have a simpler test here:

```rust
assert_eq!(result.len(), expected_results.len());
for (index, (&sample, &expected)) in zip(result.iter(), expected_results.iter()).enumerate() {
  assert!(abs(f64::from(sample) - mean - f64::from(expected)) < 5.0 * stddev,
              "DP result was not within 5 standard deviations from what was expected");
}
```

The next question is whether 5 is the right number.  If we run this 1000 times a day for 3 years, it looks like we'll get one value outside of the range in that time, which seems workable.

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -253,7 +259,36 @@ pub async fn test_oprf_ipa<F>(
...
+                epsilon,
+                delta: 1e-6,
+                ell_1_sensitivity: per_user_credit_cap_f64,
+                ell_2_sensitivity: per_user_credit_cap_f64,
+                ell_infty_sensitivity: per_user_credit_cap_f64,
+                ..Default::default()
+            };
+
+            let num_bernoulli = crate::protocol::dp::find_smallest_num_bernoulli(&noise_params);
+            let mean: f64 = f64::from(num_bernoulli) * 0.5; // n * p
```
martinthomson, line 279:
```suggestion
            // Compute the expected value of the noise from the DP parameters.
            let expected_noise: f64 = f64::from(num_bernoulli) * 0.5; // n * p
```

I wonder whether this is something that should be a function on `DpParams`, so that you don't have to use "internal" functions to compute it.  It is, after all, a consequence of how we implement the noise function.

That also suggests that `DpParams::WithDp` would not be the right name, but instead `DpMechanism::Binomial { epsilon: f64 }` instead.

martinthomson (approved):
This looks pretty good.  I have a few suggestions that I think will tighten this up a little, but it's about time this was landed and put aside.

ipa-core/benches/oneshot/ipa.rs
```diff
@@ -61,6 +61,12 @@ struct Args {
     /// while doing modulus conversion and attribution
     #[arg(long, default_value = "3")]
     num_multi_bits: u32,
+    /// DP parameters. Will run with DP by default. Can only be run without DP if `with_dp` == 0.
+    /// in which case the value of `epsilon` is ignored.
+    #[arg(short = 'd', long, default_value = "1")]
+    with_dp: u32,
```
akoshelev, line 67: have you tried to make it work with `Option`? I remember you mentioning some issues there, wondering if that what prevented you from doing it here

ipa-core/benches/oneshot/ipa.rs
```diff
@@ -155,6 +164,7 @@ async fn run(args: Args) -> Result<(), Error> {
     tracing::trace!("Preparation complete in {:?}", _prep_time.elapsed());
 
     let _protocol_time = Instant::now();
+    println!("about to run test_oprf_ipa");
```
akoshelev, line 167: either a tracing::info, or lets remove this line as it seems to be for debug purposes only

ipa-core/src/cli/playbook/ipa.rs
```diff
@@ -157,6 +157,11 @@ where
                 .unwrap()
         })
         .reconstruct();
+    println!("reconstructed results of length {}", results.len());
+    println!(
+        "query_config.max_breakdown_key is {}",
+        query_config.max_breakdown_key
+    );
```
akoshelev, line 164:
```suggestion
```

ipa-core/src/cli/playbook/mod.rs
```diff
@@ -46,7 +47,75 @@ where
...
+        mismatch.is_empty(),
+        "Expected and actual results don't match: {mismatch:?}",
+    );
+}
+
+/// Validates that the expected result matches the actual.
+///
+/// ## Panics
+/// If results don't match.
+pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_credit_cap: u32) {
```
akoshelev, line 79: this seems to replicate the behavior of `validate` almost entirely except the part where you compare the values. Can you make `validate` generic to work for both cases? It seems that you should be able to by computing the `std` and passing it as optional delta to the generic function

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -226,8 +228,77 @@ impl AsRef<str> for QueryType {
...
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum DpParams {
+    NoDp,
+    WithDp { epsilon: f64 },
+}
+
+impl FromStr for DpParams {
+    type Err = ParseFloatError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
akoshelev, line 252: if we don't really have tests that cover this, I am a little worried here

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -226,8 +228,77 @@ impl AsRef<str> for QueryType {
...
+
+impl<'de> Deserialize<'de> for DpParams {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let s: &str = Deserialize::deserialize(deserializer)?;
+        <DpParams as FromStr>::from_str(s)
+            .map_err(|e| de::Error::custom(format!("failed to deserialize DpParams object: {e}")))
+    }
+}
+
+#[cfg(test)]
```
akoshelev, line 295: you could use `cfg_attr(test, derive(Eq))` on the struct itself too. This will make it more concise because all derives will be grouped in the same place

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -21,6 +21,36 @@ use crate::{
         BitDecomposed, FieldSimd, TransposeFrom, Vectorizable,
     },
 };
+
+/// Struct to hold noise parameters, contains internal values not received from the client
```
akoshelev, line 25: Can you please document these parameters and also give an explanation for default values as well?

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -98,9 +128,59 @@ where
...
+
+// dp_for_aggregation is currently where the DP parameters epsilon, delta
+// are introduced and then from those the parameters of the noise distribution to generate are
+// calculated for use in aggregating histograms.  The DP parameters query_epsilon and
+// per_user_credit_cap come as inputs to the query with per_user_sensitivity_cap = 2^{SS_BITS}
+/// # Errors
+/// will propogate errors from `apply_dp_noise`
+/// # Panics
+/// may panic from asserts down in  `gen_binomial_noise`
+/// may panic if running with DP noise but epsilon is not in the range (0,10].
```
akoshelev, line 144: this does not match what is actually going on. Also I may prefer returning an error rather than panic here - you are reacting on the input you've been given and panic is not the best way to handle that imo

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -21,6 +21,36 @@ use crate::{
         BitDecomposed, FieldSimd, TransposeFrom, Vectorizable,
     },
 };
+
+/// Struct to hold noise parameters, contains internal values not received from the client
+pub struct NoiseParams {
+    pub epsilon: f64,
+    pub delta: f64,
+    pub per_user_credit_cap: f64,
+    pub success_prob: f64,
```
akoshelev, line 30: I would prefer to see compile safety here or at least enforce at construction time that this value is within 0..1 range. You do plenty of math down below and it can underflow easily if this value is outside of that range. We don't want to deal with that in remote servers

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -134,75 +214,61 @@ fn d_p(success_prob: f64) -> f64 {
...
-) -> f64 {
+fn epsilon_constraint(num_bernoulli: u32, noise_params: &NoiseParams) -> f64 {
     let num_bernoulli_f64 = f64::from(num_bernoulli);
-    let first_term_num = ell_2_sensitivity * (2.0 * (1.25 / delta).ln()).sqrt();
-    let first_term_den =
-        quantization_scale * (num_bernoulli_f64 * success_prob * (1.0 - success_prob)).sqrt();
-    let second_term_num = ell_2_sensitivity * c_p(success_prob) * ((10.0 / delta).ln()).sqrt()
-        + ell_1_sensitivity * b_p(success_prob);
-    let second_term_den = quantization_scale
+    let first_term_num =
```
akoshelev, line 219: I think it may be the time to start documenting what is going on here. This code is very dense and there are a lot of constants and math defined. where things like 1.0 with combination of probability can be explained, 10.0, 2.0, 3.0 are not so much I think.

You can improve this by defining constants that will help understand the code better.

```rust
   const MAX_PROBABILITY: f64 = 1.0; 
   const MAX_EPSILON: f64: 20.0
   ...
```

some of them seems to be re-usable in other parts of this module

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -134,75 +214,61 @@ fn d_p(success_prob: f64) -> f64 {
...
-    num_bernoulli: u32,
-    success_prob: f64,
-    delta: f64,
-    quantization_scale: f64,
-    dimensions: f64,
-    ell_1_sensitivity: f64,
-    ell_2_sensitivity: f64,
-    ell_infty_sensitivity: f64,
-) -> f64 {
+fn epsilon_constraint(num_bernoulli: u32, noise_params: &NoiseParams) -> f64 {
```
akoshelev, line 217: there are no validations for noise_params values happening here and plenty of opportunities to shoot ourselves in the foot with bad `noise_params`. I suggest you look at `NoiseParams` object and make sure it can only be constructed with values that won't lead to division by zero, overflow or bad results here

akoshelev (approved):
There are plenty of improvements that I think we should do before we consider this done, but this is a good start. Before we merge it, we should run a performance benchmark with Draft and see how much slower/faster IPA is with this change.

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -226,8 +228,77 @@ impl AsRef<str> for QueryType {
         }
     }
 }
+#[cfg(test)]
+impl PartialEq for IpaQueryConfig {
```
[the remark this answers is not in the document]
bmcase: Didn't seem to work earlier, but now it does which helps clean things up.

ipa-core/src/helpers/transport/query/mod.rs
bmcase: yeah, since we're not using this code right now, I'll just remove it from this PR.

ipa-core/src/test_fixture/ipa.rs
bmcase: Thanks, that's cleaner.

ipa-core/src/helpers/transport/query/mod.rs
```diff
@@ -226,8 +228,77 @@ impl AsRef<str> for QueryType {
...
+}
+
+impl FromStr for DpParams {
+    type Err = ParseFloatError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "NoDp" {
+            Ok(DpParams::NoDp)
+        } else {
+            let parts: Vec<&str> = s.splitn(2, '=').collect();
+            if parts.len() == 2 && parts[0] == "WithDp" {
```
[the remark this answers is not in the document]
bmcase: removed this code for now

ipa-core/src/test_fixture/ipa.rs
bmcase: Renamed as suggested.  I think putting this function on DpMechanism would still be a bit to high up as there will be differences between `feature_label_dot_product` and histograms.  We could put it as a function on NoiseParams.

ipa-core/src/protocol/dp/mod.rs
bmcase: okay, switched it to return an error for epsilon out of bounds.

ipa-core/src/protocol/dp/mod.rs
bmcase: added some more validation checks.

ipa-core/src/cli/playbook/mod.rs
bmcase: Look at it a little and I don't think it could work with the current trait bounds on `validate` as `PartialEq` only supports equal and not equal comparisons. But with DP we would need some <=, >= operations.  

Likely the simplest way to duplicate the code is to use the `validate_dp` function for both.   

Either direction, I'd like to push off to some future PR.

ipa-core/benches/oneshot/ipa.rs
bmcase: I tried a few things. I'm not sure `Option` on only epsilon was one of them or not.  But it is working not so don't want to mess with it anymore for now.

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -253,7 +259,45 @@ pub async fn test_oprf_ipa<F>(
...
+                ell_2_sensitivity: per_user_credit_cap_f64,
+                ell_infty_sensitivity: per_user_credit_cap_f64,
+                dimensions: 256.0, // matches hard coded dimension in oprf_ipa.rs/execute
+                ..Default::default()
+            };
+            let (mean, std) = crate::protocol::dp::noise_mean_std(&noise_params);
+
+            assert_eq!(result.len(), expected_results.len());
+            // would like to use the simpler version below but clippy and compiler disagree on
+            // importing zip and abs
```
akoshelev, line 282: what's the error here? I am seeing `&sample` so that may not work, but `sample` will be a reference. If you want an owned value, you could call `into_iter` instead

ipa-core/tests/common/mod.rs
```diff
@@ -261,14 +267,27 @@ pub fn test_ipa_with_config(mode: IpaSecurityModel, https: bool, config: IpaQuer
...
-        .stdin(Stdio::piped());
+        ]);
+
+    match config.with_dp {
+        0 => {
+            command.args(["--with-dp", &0_u32.to_string()]);
+        }
+        _ => {
+            command
+                .args(["--with-dp", &1_u32.to_string()])
```
akoshelev, line 278: is there a reason for hardcoding `1` here or we could pass `with_dp` value instead?

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -127,82 +294,62 @@ fn c_p(success_prob: f64) -> f64 {
...
     let num_bernoulli_f64 = f64::from(num_bernoulli);
-    let first_term_num = ell_2_sensitivity * (2.0 * (1.25 / delta).ln()).sqrt();
-    let first_term_den =
-        quantization_scale * (num_bernoulli_f64 * success_prob * (1.0 - success_prob)).sqrt();
-    let second_term_num = ell_2_sensitivity * c_p(success_prob) * ((10.0 / delta).ln()).sqrt()
-        + ell_1_sensitivity * b_p(success_prob);
-    let second_term_den = quantization_scale
+    let first_term_num =
+        noise_params.ell_2_sensitivity * (2.0 * (1.25 / noise_params.delta).ln()).sqrt();
+    let first_term_den = noise_params.quantization_scale
```
akoshelev, line 305: I think this code is still very very dense and challenging to follow

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -21,6 +21,103 @@ use crate::{
...
+        per_user_credit_cap: f64,
+        success_prob: f64,
+        dimensions: f64,
+        quantization_scale: f64,
+        ell_1_sensitivity: f64,
+        ell_2_sensitivity: f64,
+        ell_infty_sensitivity: f64,
+    ) -> Result<NoiseParams, String> {
+        if epsilon <= 0.0 {
+            return Err("epsilon must be < 0.0".to_string());
```
akoshelev, line 81:
```suggestion
            return Err("epsilon must be > 0.0".to_string());
```

ipa-core/src/helpers/transport/query/mod.rs
akoshelev: still worth addressing this imo

akoshelev (approved):
we can merge it and follow up on remaining items - just please make sure you squash all commits into one - there are just too many of them to live in `main`

bmcase:
<b style="font-weight:normal;" id="docs-internal-guid-35aff4ed-7fff-26a9-405e-69d2cec4e3ba"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Here is a table of draft runs with the default DP params.</span></p><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Size fixed at 1000.&nbsp;</span></p><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Max_breakdown_key always 256 (draft selection doesn&rsquo;t change this anyway)</span></p><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Max_trigger_value fixed at 7.&nbsp;</span></p><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">You can see that query time does increase as per_user_credit_cap increases, but query time is at most around 30 sec.&nbsp;</span></p><br /><div dir="ltr" style="margin-left:0pt;" align="center">
branch | per_user_credit_cap | Query time | Draft query link
-- | -- | -- | --
main | 32 | 5 sec | https://draft-mpc.vercel.app/query/view/bone-cult2024-08-01T1634
main | 64 | 5 sec | https://draft-mpc.vercel.app/query/view/stout-sneak2024-08-01T1635
main | 128 | 4 sec | https://draft-mpc.vercel.app/query/view/each-intro2024-08-01T1626
main | 256 | Error: Invalid value specified for per-user cap: 256. Must be one of 8, 16, 32, 64, or 128. | https://draft-mpc.vercel.app/query/view/found-favor2024-08-01T1633
#1133 Binomial DP | 32 | 8 sec | https://draft-mpc.vercel.app/query/view/macro-jog2024-08-01T1635
#1133 Binomial DP | 64 | 17 sec | https://draft-mpc.vercel.app/query/view/swish-panic2024-08-01T1622
#1133 Binomial DP | 128 | 32 sec | https://draft-mpc.vercel.app/query/view/tidy-cite2024-08-01T1629
#1133 Binomial DP | 256 | Error: Invalid value specified for per-user cap: 256. Must be one of 8, 16, 32, 64, or 128. | https://draft-mpc.vercel.app/query/view/here-drink2024-08-01T1630

</div><br /></b>

ipa-core/src/helpers/transport/query/mod.rs
bmcase: yeah, I ran into an error when I tried this before.  This is what I still get when I try doing it as

```
#[derive(Debug, Copy, Clone, PartialEq)]
#[cfg_attr(test, derive(Eq))]
pub enum DpMechanism {
    NoDp,
    Binomial { epsilon: f64 },
}
```

image

ipa-core/tests/common/mod.rs
bmcase: yes, we should be able to do that. Will update

ipa-core/src/test_fixture/ipa.rs
bmcase: played with it a bit more and I think it is working now.

ipa-core/src/protocol/dp/mod.rs
bmcase: yeah, you're right.  But Thm1 of the [paper](https://arxiv.org/pdf/1805.10559) that we implement here is very dense too -- I think keeping it as close to the paper is what lets us check most easily that it is computing the correct thing. 

image

akoshelev:
> Here is a table of draft runs with the default DP params.
> 
> Size fixed at 1000. 
> 
> Max_breakdown_key always 256 (draft selection doesn’t change this anyway)
> 
> Max_trigger_value fixed at 7. 
> 
> You can see that query time does increase as per_user_credit_cap increases, but query time is at most around 30 sec. 
> 

How do thee results compare with OPRF IPA w/o noise?

akoshelev:
> This is a merge option, right? I don't think I can do this as don't have merge access. Can you do when you merge?

sure, I'll do it

bmcase:
@akoshelev 
> How do these results compare with OPRF IPA w/o noise?

The branch main rows in the table are these without the noise generation.  Essentially without noise the performance doesn't scale with the per_use_credit_cap as it does when we add noise.

akoshelev:
Got it, thanks! Can you also run it on a large input (50M rows) with 32 per user cap to compare?

bmcase:
@akoshelev  figured out a better way to handle `Eq` for `DpMechanism` -- we can just remove it.  Must have been something else earlier was needing that we don't use in the end. 

Running draft on 50M for main.  2hrs already.  How long do you expect main to take for that?

akoshelev:
yea it normally takes around 5hrs to finish :(

bmcase:
@akoshelev  completed a few draft runs for larger scale queries and it looks like there is negligible performance impact as the query size increases (as would be expected).  We would expect about 8 sec total time added and for these large queries that appears to be much smaller than the normal variance across different runs of the same code.  

For per_user_credit_cap = 32 and the default DP parameters. 

<b style="font-weight:normal;" id="docs-internal-guid-cbb271a3-7fff-9dd0-b104-8c36d247e353"><div dir="ltr" style="margin-left:0pt;" align="center">
Branch and size | per_user_credit_cap | Query time | Draft query link
-- | -- | -- | --
#1133 Binomial DP  5M input rows | 32 | 40:00 | https://draft-mpc.vercel.app/query/view/camp-beat2024-08-05T1841
Main5M input rows | 32 | 40:53 | https://draft-mpc.vercel.app/query/view/sewn-rinse2024-08-03T2137
#1133 Binomial DP  50M input rows | 32 | 6:38 | https://draft-mpc.vercel.app/query/view/elect-goon2024-08-06T0146
main  50M input rows | 32 | 6:35 | https://draft-mpc.vercel.app/query/view/spumy-wing2024-08-02T0042

</div></b>

akoshelev:
agree, lets merge this

PR #1134, opened 2024-06-12
recursive computation of p and q final without allocation

danielmasny:
verifier.rs computes p and q now recursively without allocation in a vector by using streams and the chunks method on streams

Unfortunately Rust treats streams as different types even when they use the same Item type. I used Pin<Box<Dyn<>> to be able to store different types an accomplish evaluating streams recursively. 

I have tried using the chunks method by itertools. Unfortunately, I ran into issues since stream::chunk returns chunks that are vectors, itertools:chunks seems to return a reference to some custom chunk type.

@andyleiserson I think there is probably something that can be improved.

@andyleiserson I did not manage to make your borrow trick work such that I can use streams of owned values or references. The issue is that the  Pin<Box<Dyn<Stream<Item = B>>>> became complex, since B = dyn Borrow<[]>

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
@@ -1,4 +1,7 @@
-use std::borrow::Borrow;
+use std::pin::Pin;
+
+use futures::Stream;
+use futures_util::StreamExt;
 
 use crate::{
     ff::PrimeField,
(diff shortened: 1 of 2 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -252,18 +335,75 @@ mod test {
...
     }
+
+    #[tokio::test]
+    async fn recursive_compute_p() {
+        const U_1: [u128; 32] = [
+            0, 30, 0, 16, 0, 1, 0, 15, 0, 0, 0, 16, 0, 30, 0, 16, 29, 1, 1, 15, 0, 0, 1, 15, 2, 30,
+            30, 16, 0, 0, 30, 16,
+        ];
+
+        const R: [u128; 3] = [22, 17, 30];
```
benjaminsavage, line 354: Suggest `CHALLENGES`

benjaminsavage (approved):
I like it. I'm sure we can continue to improve things, and probably find a way to *not* use streams, but I am happy to continue evolving this part of the codebase.

PR #1135, opened 2024-06-13
Use circular buffer inside the `OrderingSender`

akoshelev:
As proposed in #1118 and #1085, this gets further ahead in having 2 tunable parameters for the IPA infrastructure. The active window size that existed before and the new parameter that defines send network buffer size. Some experiments with setting both to different values are shown in #1118 along with the justification for it.

With this change, `OrderingSender` no longer accepts spare capacity and delegates most of its buffering business to `CircularBuf`, taking care of providing an async API around it and ordering writes.


ipa-core/src/helpers/buffers/ordering_sender.rs
```diff
@@ -79,59 +66,43 @@ impl State {
...
-            self.buf.capacity() - self.written,
-            self.buf.capacity(),
-        );
-        self.written += M::Size::USIZE;
-        m.serialize(GenericArray::from_mut_slice(&mut b[..M::Size::USIZE]));
-
-        if !self.accept_writes() {
+        self.buf.next().write(m);
+
+        // todo: test tht it is not in_full call
```
akoshelev, line 76: remove

ipa-core/src/helpers/buffers/ordering_sender.rs
```diff
@@ -27,28 +25,17 @@ use crate::{
...
+    buf: CircularBuf,
     /// An entity to wake when the buffer is read from.
     write_ready: Option<Waker>,
     /// Another entity to wake when the buffer is read from.
     stream_ready: Option<Waker>,
 }
 
 impl State {
-    fn new(capacity: NonZeroUsize, spare: usize) -> Self {
+    fn new(capacity: usize, write_size: usize, read_threshold: usize) -> Self {
```
akoshelev, line 36: should be write_size and everything else is a multiplier of write_size

ipa-core/src/helpers/buffers/ordering_sender.rs
```diff
@@ -79,59 +66,43 @@ impl State {
...
+        if self.buf.can_read() {
+            if !self.buf.can_write() {
+                // We are ready to unblock writers by taking some data that we know is there off
+                // the buffer
+                Self::wake(&mut self.write_ready);
+            }
 
-            Self::wake(&mut self.write_ready);
-            Poll::Ready(v)
+            Poll::Ready(self.buf.take())
```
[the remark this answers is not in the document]
martinthomson: And include a test that reveals this problem.

martinthomson (approved):

ipa-core/src/helpers/buffers/ordering_sender.rs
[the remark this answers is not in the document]
akoshelev: we discussed it in our meeting - it is not easy to reproduce out of order problem because of the lock around the `State` object that serializes access to internals. Even if we wake another thread before we take the data out, it will still block until the lock is released and that only happens after `.take()` call.

I am going to change the order anyway because if we ever to remove the lock which `CircularBuf` can make possible, then this will cause issues. And I'll try to write a test that unfortunately won't fail today, but it could fail if there is no lock and wrong order

ipa-core/src/helpers/buffers/ordering_sender.rs
akoshelev: I looked at the code and I think it may be easier to leave it as is. The math behind calculating read_size and capacity is quite complex and needs to take into account the total records, active_work and read_size config parameter.

It is much easier to test this math by just validating the `SendChannelConfig`

PR #1136, opened 2024-06-13
Cleaning up test code

benjaminsavage:
I left a few small naming suggestions on #1134. It's easier to just merge that PR and put up a new one with those changes given timezones =).

The tests were also bothering me. I have tried to unify what we are testing with the actual approach we intend to use. Namely:
1. To use a consistent compression factor for every step, including the final step. Not vary the compression factor for the final recursive step.
2. For the final proof, to first move element at index 0 to the end of the array, then overwrite index 0 with the mask.

Now that these are unified, we can just test it two ways, manually calling `recurse_u_or_v` many times, as well as to call `recursively_compute_final_check`. Now we can just assert these give the same answer.

How did I get the expected numbers? I updated the Google Spreadsheet where I have manually implemented this ZKP protocol: https://docs.google.com/spreadsheets/d/10tCSgiZyQbt4bMB3KqhAUOCfgfO8xIni83J2saipFCg/edit?usp=sharing

I like to see that the numbers match up. It gives me confidence that we did things correctly, since mistakes are likely to be independent in these two wildly different implementations.


danielmasny (approved):
Awesome, Thanks!

PR #1139, opened 2024-06-14
Compute zero shares for all proofs at once

danielmasny:
Computes zero shares for all proofs at once.

Also removes async from recurse_u_or_v as @benjaminsavage has pointed out since it was not needed.

renamed iterators that were streams to streams

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
@@ -11,9 +11,12 @@ use crate::{
 };
 
 /// This function computes the shares that sum to zero from the zero-knowledge proofs.
+///
+/// `out_share` is a share of `x` where the proof proves the statement `sum_i u_i * v_i = x`
 pub fn compute_g_differences<F, const P: usize, const λ: usize>(
-    zkp: &Vec<[F; P]>,
+    zkps: &Vec<[F; P]>,
     challenges: &[F],
+    out_share: F,
 ) -> Vec<F>
 where
     F: PrimeField,
@@ -22,28 +25,28 @@ where
     let lagrange_denominator: CanonicalLagrangeDenominator<F, P> =
         CanonicalLagrangeDenominator::<F, P>::new();
 
-    // compute g_r with "empty" first spot
-    let g_r = iter::once(F::ZERO)
+    // compute expected_sum with "out_share" at the first spot
+    let expected_sums = iter::once(out_share)
         .chain(
             challenges
                 .iter()
-                .zip(zkp)
+                .zip(zkps)
                 .map(|(challenge, zkp)| interpolate_at_r(zkp, *challenge, &lagrange_denominator)),
         )
         .collect::<Vec<_>>();
 
     // compute g_sum)
-    let g_sum = zkp
+    let g_sums = zkps
         .iter()
         .map(compute_sum_share::<F, λ, P>)
         // append spot for final sum
         .chain(iter::once(F::ZERO))
         .collect::<Vec<_>>();
 
-    g_sum
+    g_sums
         .iter()
-        .zip(g_r)
-        .map(|(g_sum, g_r)| *g_sum - g_r)
+        .zip(expected_sums)
+        .map(|(g_sum, expected_sum)| *g_sum - expected_sum)
         .collect()
 }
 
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -10,6 +10,43 @@ use crate::{
     },
 };
 
+/// This function computes the shares that sum to zero from the zero-knowledge proofs.
+pub fn compute_g_differences<F, const P: usize, const λ: usize>(
+    zkp: &Vec<[F; P]>,
```
benjaminsavage, line 15: This is a list of multiple proofs right? Can you give it a ural name?

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -10,6 +10,43 @@ use crate::{
...
+    let lagrange_denominator: CanonicalLagrangeDenominator<F, P> =
+        CanonicalLagrangeDenominator::<F, P>::new();
+
+    // compute g_r with "empty" first spot
+    let g_r = iter::once(F::ZERO)
+        .chain(
+            challenges
+                .iter()
+                .zip(zkp)
+                .map(|(challenge, zkp)| interpolate_at_r(zkp, *challenge, &lagrange_denominator)),
```
benjaminsavage, line 31: I don't understand why we are calling interpolate at r and providing r=0 for the first proof. Am I misunderstanding what the code does? Because that's not secure.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -370,4 +401,30 @@ mod test {
...
     }
+
+    #[test]
+    fn differences_are_zero() {
+        const ZKP_1_and_2_LEFT: [u128; 14] = [0, 0, 13, 17, 11, 25, 7, 11, 25, 17, 9, 22, 23, 3];
+        const ZKP_1_and_2_RIGHT: [u128; 14] = [0, 30, 16, 13, 25, 3, 6, 1, 12, 29, 30, 7, 7, 3];
+        const OUT_LEFT: u128 = 27;
+        const OUT_RIGHT: u128 = 0;
+
+        let zkp_left = make_chunks::<Fp31, 7>(&ZKP_1_and_2_LEFT);
```
benjaminsavage, line 412: Interesting. I associate the "make chunks" function with generating the UV iterations. But I suppose it works to generate a list of proofs too...

benjaminsavage: It would be nice to somehow re-use the existing constants, maybe by leveling them up a bit in scope so that multiple methods can all use them. Then we could at least do something more readable like:
[PROOF_1, PROOF_2].map(...).collect()

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -10,6 +10,43 @@ use crate::{
...
+    g_sum
+        .iter()
+        .zip(g_r)
+        .map(|(g_sum, g_r)| *g_sum - g_r)
+        .collect()
+}
+
 ///
 /// Distributed Zero Knowledge Proofs algorithm drawn from
 /// `https://eprint.iacr.org/2023/909.pdf`
```
benjaminsavage, line 52: I would rather not do this here, but instead do it within the function itself.  Basically things the check for the first stage that it equals -m/2.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -10,6 +10,43 @@ use crate::{
...
+) -> Vec<F>
+where
+    F: PrimeField,
+{
+    // compute denominator
+    let lagrange_denominator: CanonicalLagrangeDenominator<F, P> =
+        CanonicalLagrangeDenominator::<F, P>::new();
+
+    // compute g_r with "empty" first spot
+    let g_r = iter::once(F::ZERO)
```
benjaminsavage, line 26: I think we should instead initialize this to -m/2. In the protocol, they suggest P_{i+1} initialize to -m/2 and P_{i-1} initialize to 0.

By doing it here, we avoid the need for an ugly modification after calling this function.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -10,6 +10,43 @@ use crate::{
...
+
+    // compute g_sum)
+    let g_sum = zkp
+        .iter()
+        .map(compute_sum_share::<F, λ, P>)
+        // append spot for final sum
+        .chain(iter::once(F::ZERO))
+        .collect::<Vec<_>>();
+
+    g_sum
```
benjaminsavage, line 43: Instead of doing this in 3 stages, i suggest we just combine them into a single map, that just computes the differences directly.

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -370,4 +401,30 @@ mod test {
...
+        let g_differences_left = compute_g_differences::<_, 7, 4>(&zkp_left, &challenges);
+        let g_differences_right = compute_g_differences::<_, 7, 4>(&zkp_right, &challenges);
+
+        let mut g_differences = g_differences_left
+            .iter()
+            .zip(g_differences_right)
+            .map(|(left, right)| *left + right)
+            .collect::<Vec<_>>();
+
+        g_differences[0] -= Fp31::truncate_from(OUT_LEFT) + Fp31::truncate_from(OUT_RIGHT);
```
benjaminsavage, line 425: I would rather not do this here, but instead do it within the function itself.  Basically things the check for the first stage that it equals -m/2.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -370,4 +401,30 @@ mod test {
...
+    #[test]
+    fn differences_are_zero() {
+        const ZKP_1_and_2_LEFT: [u128; 14] = [0, 0, 13, 17, 11, 25, 7, 11, 25, 17, 9, 22, 23, 3];
+        const ZKP_1_and_2_RIGHT: [u128; 14] = [0, 30, 16, 13, 25, 3, 6, 1, 12, 29, 30, 7, 7, 3];
+        const OUT_LEFT: u128 = 27;
+        const OUT_RIGHT: u128 = 0;
+
+        let zkp_left = make_chunks::<Fp31, 7>(&ZKP_1_and_2_LEFT);
+        let zkp_right = make_chunks::<Fp31, 7>(&ZKP_1_and_2_RIGHT);
+        let challenges = vec![Fp31::truncate_from(22u128), Fp31::truncate_from(17u128)];
```
benjaminsavage, line 414: You can add the third proof and third challenge here as well to better exercise the logic, since the first and last proof are somewhat special logic. Third challenge is 30.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
benjaminsavage, line 31: So after we chain together -m/2 and the random challenge points, let's call the loop variable something like expected_sum.

danielmasny: Here we start with a 0 and append all the g_r = zkp(r) values, i.e. (0, zkp_1(r1) , ..., zkp_n(rn)). For the g_sums we do the opposite, i.e. (gsum_1,...,gsum_n, 0) so we get the differences (gsum_1-0, gsum_2-zkp_1(r1),...,0-zkp_n(rn))

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
danielmasny: I could use a single collect, i.e. not collecting g_r and g_sum in a vector (which would make the code more complex). I am not sure how to use a single map, the challenge is that for the first and last element I need to do something different which is hard to do with a single map (unless we enumerate the iterator and put an if statement in the map).

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
danielmasny: Right now, we use a vector which does not support items of different length. I was thinking of adding the support for having a different length first proof in a follow up PR.

benjaminsavage (approved):
Thanks Daniel!

PR #1140, opened 2024-06-14
Use 32 bit histogram values

akoshelev:
This addresses #1100 by increasing number of bits in histogram values to 32, supporting up to ~500M inputs with 2 bit trigger values (or 200M with 3 bit values)


akoshelev:
draft run (20M): https://draft-mpc.vercel.app/query/view/flip-value2024-06-14T1748

akoshelev:
All right, I think it is working but takes 50% more time to finish, which is expected because we doubled the number of bits used in aggregation. We quickly get to 16 bits I think and after that we start paying the price of adding more bits while previously we just saturated at 16

But it is clearly working as histogram values are greater than $2^16$

```
41 | Some(421047) | Some(421047)
```

ipa-core/src/query/executor.rs
```diff
@@ -105,7 +106,7 @@ pub fn execute(
             move |prss, gateway, config, input| {
                 let ctx = SemiHonestContext::new(prss, gateway);
                 Box::pin(
-                    OprfIpaQuery::<BA16>::new(ipa_config, key_registry)
+                    OprfIpaQuery::<crate::ff::boolean_array::BA16>::new(ipa_config, key_registry)
```
benjaminsavage, line 109: Why fully qualified?

benjaminsavage (approved):

ipa-core/src/query/executor.rs
akoshelev: its behind a feature flag "weak_field"

akoshelev:
We are investigating an issue where 32 bit breakdowns may(?) cause protocols to be unnecessary slow. #1151.
It does not seem to be the case, but I want to be absolutely sure before I can merge this PR

bmcase:
We will also need to have 32 bits in output values for many (particularly large sensitivity) parameter sets we want to generate Binomial DP noise for.

Issue #1141, opened 2024-06-15
IPA fails on 100M input with `unspecified total records` error


akoshelev:
```
Helper 3: unspecified total records for channel[H2,"/ipa_prf/shuffle/transfer_c_hat"]
```

https://draft-mpc.vercel.app/query/view/mined-train2024-06-14T2354

That is very suspicious because we do have `.set_total_records` call there
https://github.com/private-attribution/ipa/blob/103fbe7d1d376309b08d1b46a5e6e8c30bc74887/ipa-core/src/protocol/ipa_prf/shuffle/base.rs#L305

akoshelev:
added more tracing in #1142 - we seem to be interpreting 0 as unspecified (which is bad) but also why batch_size is 0?

```
2024-06-15T00:57:20.794999Z INFO oprf_ipa_query{sz=100000000}:shuffle_inputs:h2_shuffle{batch_size=0}:send_to{sz=0}: ipa_core::protocol::ipa_prf::shuffle::base: close time.busy=2.78ms 
```

akoshelev:
```
oprf_ipa_query{sz=100000000}:shuffle_inputs{sz=0}: ipa_core::protocol::ipa_prf::shuffle: new
```

something is wrong with getting 100M elements out of the input body I think

andyleiserson:
> interpreting 0 as unspecified

Yeah, we do that, I ran into it in #897. In that case it only showed up with a nonsensical configuration (one record per user for every single user), so it didn't seem all that important to fix. Now it seems more important to fix.

akoshelev:
> Helper 3: 2024-06-15T01:56:45.388307Z INFO oprf_ipa_query{sz=100000000}: ipa_core::query::runner::oprf_ipa: starting polling the plaintext stream to get 100000000 items
Helper 3: 2024-06-15T01:56:45.388348Z INFO oprf_ipa_query{sz=100000000}: ipa_core::query::runner::oprf_ipa: ready for OPRF ipa, got 0 items

I mean...

akoshelev:
please let it not be this https://github.com/hyperium/h2/issues/471

akoshelev:
| Input size (records) | Input size (bytes) | Next power of two |
| -------| --------- | ------------- |
| 100,000,000 |  2,800,000,000 | $2^{31}$+ |
| 90,000,000 | 2,520,000,000 | $2^{31}$+ | 
| 80,000,000 | 2,240,000,000 | $2^{31}$+ |
[...]

PR #1142, opened 2024-06-15
[DRAFT] [DO NOT MERGE] reproduce #1141

akoshelev:


akoshelev:
Don't need it as the issue has been reproduced and fixed

PR #1143, opened 2024-06-15
keep track of amount of u,v values included in a proof

danielmasny:
This PR allows to keep track of amount of u,v values when generate proof. This is important since the verification will need to know the amount.


benjaminsavage (changes requested):
I disagree with this design. I do not think compute proof should be responsible for counting the elements added to it. Look at all the complex tuple return types and the frequently ignored values. This is not a good design. 

I think when we perform a multiplication and push it into the list of multiplications that need to be verified, THAT is the place to increment a counter. In fact, it's only that first stage where we need to perform this count.

danielmasny:
> I disagree with this design. I do not think compute proof should be responsible for counting the elements added to it. Look at all the complex tuple return types and the frequently ignored values. This is not a good design.
> 
> I think when we perform a multiplication and push it into the list of multiplications that need to be verified, THAT is the place to increment a counter. In fact, it's only that first stage where we need to perform this count.

@benjaminsavage I would be fine with that too and was thinking about that as well. A reason why I think this PR is more straightforward is that we just count the amount when we actually compute the proof (which will be the same for all three proofs and doesn't need to be sent over the network). Counting the multiplications when they are stored comes with two challenges:
- We collect the multiplication in a hash table so we would need to keep track of the number in each element of the hash table
- There might be zero fillings that we need to account for.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -123,15 +123,22 @@ impl<F: PrimeField, const λ: usize, const P: usize, const M: usize> ProofGenera
     ///
     /// Distributed Zero Knowledge Proofs algorithm drawn from
     /// `https://eprint.iacr.org/2023/909.pdf`
-    fn compute_proof<J, B>(uv_iterator: J, lagrange_table: &LagrangeTable<F, λ, M>) -> [F; P]
+    ///
+    /// returns proof and number of summands `m` which is important for the verification
```
benjaminsavage, line 127: Generally, when you find yourself writing that a function does X *and* Y, it should be two separate functions.

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -123,15 +123,22 @@ impl<F: PrimeField, const λ: usize, const P: usize, const M: usize> ProofGenera
...
+    fn compute_proof<J, B>(
+        uv_iterator: J,
```
benjaminsavage, line 129: If this implemented "ExactSizeIterator" we would be done. There would be no need for this. Can we do that instead?

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -337,7 +347,7 @@ mod test {
         assert_eq!(uv_2, zip_chunks(U_2, V_2));
 
         // next iteration
-        let proof_2 = TestProofGenerator::compute_proof(uv_2.iter(), &lagrange_table);
+        let (proof_2, _) = TestProofGenerator::compute_proof(uv_2.iter(), &lagrange_table);
```
benjaminsavage, line 350: To me, this line makes it so clear that this is the wrong design. We call `compute_proof` multiple times, but we only care about the size of the iterator we pass to it on the first iteration. Surely we should just treat that first iteration differently (e.g. add this counter to either the storage of multiplications to be validated, or possibly to the one-time conversion from binary to large field values).

ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -366,7 +376,7 @@ mod test {
         );
 
         // final iteration
-        let proof_3 = TestProofGenerator::compute_proof(masked_uv_3.iter(), &lagrange_table);
+        let (proof_3, _) = TestProofGenerator::compute_proof(masked_uv_3.iter(), &lagrange_table);
```
benjaminsavage, line 379: And here we ignore it again.

benjaminsavage (changes requested):
I still remain convinced this is the wrong design. I'm waiting for an update to this PR, or perhaps a new PR that approaches this differently.

benjaminsavage (approved):

PR #1145, opened 2024-06-15
enumerate chunks for chunk specific contexts

danielmasny:
add a counter for chunks in validate such that chunk specific contexts for the validation can be generated.

### DIFF
--- a/ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
+++ b/ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
@@ -113,7 +113,7 @@ mod test {
                 assert!(matches!(mctx.is_verified(), Err(Error::ContextUnsafe(_))));
 
                 // validate all elements in the batch
-                validator.validate::<Fp61BitPrime>().await.unwrap();
+                validator.validate::<Fp61BitPrime>(0usize).await.unwrap();
 
                 // batch is empty now
                 assert!(mctx.is_verified().is_ok());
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
```diff
@@ -113,7 +113,7 @@ mod test {
                 assert!(matches!(mctx.is_verified(), Err(Error::ContextUnsafe(_))));
 
                 // validate all elements in the batch
-                validator.validate().await.unwrap();
+                validator.validate(0usize).await.unwrap();
```
benjaminsavage, line 116: This is kind of weird. If you aren't calling the `validated_seq_join` version, passing a `0usize` here is kind of strange.

Elsewhere, the way we approach this is to first narrow the context. Might there be some similar solution here?

benjaminsavage (approved):
I get it, and it works, but I can't help but wonder if there is a better solution.

ipa-core/src/protocol/basics/mul/dzkp_malicious.rs
danielmasny: I agree. I have tried to put the counter as a field in the validator, but then validate needs to take &mut self rather than &self which causes problems with FnMut closures.

andyleiserson: There is a `Mutex` within the validator if you go deep enough, so it seems like it ought to be possible to leverage that somehow to get interior mutability without a major change to the contract / behavior. But it might take a little rearranging of the pieces.

danielmasny: @andyleiserson so you suggest to store the counter on the batch? 

```
pub struct MaliciousDZKPValidator<'a> {
    batch_ref: Arc<Mutex<Batch>>,
    protocol_ctx: MaliciousDZKPUpgraded<'a>,
    validate_ctx: Base<'a>,
}
```

I think that would a bit strange since the counter is used to narrow down `validate_ctx` which is also not stored in the batch.

PR #1146, opened 2024-06-17
Reject zero total records; handle degenerate inputs

### LINKED ISSUE #897: Unspecified record count panic with 1 record per user
As an experiment related to #867 I tried running ORPF IPA with one record per user, and it failed differently. This is at dcb6a39.

```
$ time cargo bench --features enable-benches,descriptive-gate --bench oneshot_ipa -- -o -c 8 -n 100000 -u 1
    Finished bench [optimized] target(s) in 0.13s
     Running benches/oneshot/ipa.rs (target/release/deps/oneshot_ipa-c3dc6b467c2ecadc)
thread 'tokio-runtime-worker' panicked at /Users/Andy/Development/ipa/ipa-core/src/protocol/modulus_conversion/convert_shares.rs:355:5:
unspecified record count for modulus conversion at step=protocol/run-0/prime_field_validator/modulus_convert_breakdown_key_bits_and_trigger_values
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: RuntimeError(JoinError::Panic(Id(4), ...))
error: bench failed, to rerun pass `-p ipa-core --bench oneshot_ipa`

Caused by:
  process didn't exit successfully: `/Users/Andy/Development/ipa/target/release/deps/oneshot_ipa-c3dc6b467c2ecadc -o -c 8 -n 100000 -u 1 --bench` (exit status: 1)

real	4m37.486s
user	4m36.192s
sys	0m0.538s
```

andyleiserson:
This does two things:
* Makes IPA tolerant of empty input or the degenerate case of all users having only one record.
* Rejects attempts to set total records to zero with a clear error message.

~Doing something like `set_total_records(vec.len())` with an empty vec is still a panic at runtime. We could have `set_total_records` take `NonZeroUsize`, but that would require changing a bunch of things that call it to explicitly handle the zero case.~ (Edit: resolved in updated version)

Fixes #897

### DIFF
--- a/ipa-core/src/helpers/mod.rs
+++ b/ipa-core/src/helpers/mod.rs
@@ -6,7 +6,7 @@
 use std::{
     convert::Infallible,
     fmt::{Debug, Display, Formatter},
-    num::{NonZeroUsize, TryFromIntError},
+    num::NonZeroUsize,
 };
 
 use generic_array::GenericArray;
@@ -72,7 +72,7 @@ pub use transport::{
 };
 #[cfg(feature = "in-memory-infra")]
 pub use transport::{InMemoryMpcNetwork, InMemoryShardNetwork, InMemoryTransport};
-use typenum::{Unsigned, U8};
+use typenum::{Const, ToUInt, Unsigned, U8};
 use x25519_dalek::PublicKey;
 
 use crate::{
@@ -509,16 +509,17 @@ impl Display for TotalRecords {
 }
 
 impl TotalRecords {
-    /// Construct a `TotalRecords` specifying a record count of `value`.
-    ///
-    /// ## Panics
-    /// If the supplied record count is zero.
-    #[must_use]
-    pub const fn specified(value: usize) -> Self {
-        // `Option::unwrap` is not yet `const`.
-        match NonZeroUsize::new(value) {
-            Some(value) => Self::Specified(value),
-            None => panic!("TotalRecords cannot be zero"),
+    pub const ONE: Self = match NonZeroUsize::new(1) {
+        Some(value) => TotalRecords::Specified(value),
+        None => unreachable!(),
+    };
+
+    /// ## Errors
+    /// If `value` is zero. Only non-zero counts are supported.
+    pub fn specified(value: usize) -> Result<Self, ZeroRecordsError> {
+        match NonZeroUsize::try_from(value) {
+            Ok(value) => Ok(TotalRecords::Specified(value)),
+            Err(_) => Err(ZeroRecordsError),
         }
     }
 
@@ -556,12 +557,8 @@ impl TotalRecords {
     /// Any new value is OK if the current value is unspecified.
     /// Otherwise the new value can be indeterminate if the old value is specified.
     #[must_use]
-    pub fn overwrite<T: TryInto<TotalRecords>>(&self, value: T) -> TotalRecords {
-        let new = value
-            .try_into()
-            .ok()
-            .expect("zero is not a valid value for TotalRecords");
-        match (self, new) {
+    pub fn overwrite<T: Into<TotalRecords>>(&self, value: T) -> TotalRecords {
+        match (self, value.into()) {
             (Self::Unspecified, v) => v,
             (_, Self::Unspecified) => panic!("TotalRecords needs a specific value for overwriting"),
             (Self::Specified(_), Self::Indeterminate) => Self::Indeterminate,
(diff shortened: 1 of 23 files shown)


ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -219,7 +230,7 @@ where
             list.swap(i - 1, pivot_index);
 
             // mark which ranges need to be sorted in the next pass
-            if i > pivot_index + 1 {
+            if i > pivot_index + 2 {
```
andyleiserson, line 233: `i` is initialized to `pivot_index + 1`. When we get here...

* if `i == pivot_index + 1`, then we didn't move any elements before the pivot.
* if `i == pivot_index + 2`, then we moved one element before the pivot, still no need to sort the before-pivot range n the next pass.

The other clue that something was off here is that one side of the if had +1 / -1, and the other had +1 / 0. +2 / -1 and +1 / 0 makes more sense.

ipa-core/src/protocol/ipa_prf/quicksort.rs
```diff
@@ -134,7 +134,14 @@ where
...
 {
-    assert!(!ranges_to_sort.iter().any(Range::is_empty));
+    // Panic on any empty ranges in the input. Check if all the input ranges have length 1;
+    // if so, there's nothing to do.
+    if ranges_to_sort.iter().fold(true, |trivial, range| {
+        assert!(!range.is_empty());
+        trivial && range.len() == 1
+    }) {
+        return Ok(());
+    }
```
andyleiserson, line 144: I made the changes to quicksort first, before I added the special case in `oprf_ipa` that avoids calling it entirely. So the quicksort changes aren't essential. But given we're already checking for invalid ranges at the start of quicksort, the incremental cost of detecting the trivial case is minimal, and it lets us be more defensive in the rest of the function.

akoshelev:
thanks for addressing this issue quickly Andy!

ipa-core/src/helpers/mod.rs
```diff
@@ -509,6 +509,19 @@ impl Display for TotalRecords {
 }
 
 impl TotalRecords {
+    /// Construct a `TotalRecords` specifying a record count of `value`.
+    ///
+    /// ## Panics
+    /// If the supplied record count is zero.
+    #[must_use]
+    pub const fn specified(value: usize) -> Self {
```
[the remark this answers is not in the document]
andyleiserson: This function is actually only used in one place, which calls it with a constant. I wasn't thinking clearly and mixed up "can be called in const context" with "must be called in const context". I'll just get rid of this one.

To be clear, you're looking for something that makes zero total records an error, rather than a panic? A bunch of the places `set_total_records` is used are in tests, where I'm not sure `?` is an easy way out. On the other hand, while `set_total_records` is used in lots of places, I'm more bothered by the clutter of lots of `unwrap`/`expect` than the work of updating them all.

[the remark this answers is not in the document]
akoshelev: > To be clear, you're looking for something that makes zero total records an error, rather than a panic? 
That's right and only in production code. We can provide a function behind #[cfg(test)] that does panic and I think this is totally fine.

I do agree that `unwrap`/`expect` is just code clutter in this case and I was hoping that providing a function that returns `Result` will avoid it by using `?` - again only inside protocols and only if we need it. Does it make sense or I am missing something?

ipa-core/src/helpers/mod.rs
```diff
@@ -509,6 +509,20 @@ impl Display for TotalRecords {
 }
 
 impl TotalRecords {
+    pub const ONE: Self = match NonZeroUsize::new(1) {
```
akoshelev, line 512: wow this is nice, I didn't think you could do it without `unsafe_unchecked`

ipa-core/src/helpers/mod.rs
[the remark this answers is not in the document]
andyleiserson: Updated it to generate errors in protocol code.

ipa-core/src/protocol/context/upgrade.rs
```diff
@@ -180,7 +183,7 @@ where
         let ctx = if self.ctx.total_records().is_specified() {
             self.ctx
         } else {
-            self.ctx.set_total_records(1)
+            self.ctx.set_total_records(typenum::Const::<1>)
```
akoshelev, line 186: why not `TotalRecords::ONE`?

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -247,7 +247,7 @@ impl ProofHashes {
     /// Sends the one verifier's hashes to the other verifier
     /// `side` indicates the direction of the prover.
     async fn send_hashes<C: Context>(&self, ctx: &C, side: Side) -> Result<(), Error> {
-        let communication_ctx = ctx.set_total_records(self.hashes.len());
+        let communication_ctx = ctx.set_total_records(TotalRecords::specified(self.hashes.len())?);
```
akoshelev, line 250: nice, love it

akoshelev (approved):
thanks for improving type safety

ipa-core/src/protocol/context/upgrade.rs
andyleiserson: No good reason... I think I did a search and replace for places that total records was called with an number.

ipa-core/src/helpers/mod.rs
andyleiserson: This may not have always worked... or maybe it's a case of not thinking to try the obvious thing :). Bonus points to clippy for not complaining that it could be an unwrap.

ipa-core/src/protocol/context/upgrade.rs
akoshelev: I think you may want to use it as people may not want to deal with typenum here

PR #1148, opened 2024-06-18
p, q final multiplication

danielmasny:
consistency check for final p, q and p_times_q computation


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -73,12 +73,12 @@ fn compute_sum_share<F: PrimeField, const λ: usize, const P: usize>(zkp: &[F; P
 /// This function compresses the `u_or_v` values and returns the next `u_or_v` values.
 ///
 /// The function uses streams since stream offers a chunk method.
-fn recurse_u_or_v<'a, F: PrimeField, J, const λ: usize>(
+fn recurse_u_or_v<'a, F: PrimeField, J, const λ_FIRST: usize, const λ: usize>(
```
benjaminsavage, line 76: This feels bad because we only need this on the first call to `recurse_u_or_v`, and yet we call this method many times.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -92,28 +92,40 @@ where
...
-    let mut stream: Pin<Box<dyn Stream<Item = [F; λ]>>> = Box::pin(u_or_v_stream);
-    for lagrange_table in tables.iter().take(recursions - 1) {
-        stream = Box::pin(recurse_u_or_v(stream, lagrange_table));
+    let mut stream: Pin<Box<dyn Stream<Item = [F; λ]> + Send>> =
+        Box::pin(recurse_u_or_v::<_, _, λ_FIRST, λ>(
+            u_or_v_stream,
+            &table_first,
+        ));
+    for lagrange_table in tables.iter().take(recursions - 2) {
+        stream = Box::pin(recurse_u_or_v::<_, _, λ, λ>(stream, lagrange_table));
```
benjaminsavage, line 128: This whole thing just got really, really messy. I'm not happy with where this is going.

Now that we have at least two examples of behavior that is different on the first iteration (needing to compute the count `m`, and a different compression factor), I'm starting to lean towards a design like:

```
let res1 = functionThatDoesIterationOne();
let res_final = functionThatDoesAllTheOtherIterations(res1);
```

Given that there are a lot of other fancy optimizations that also only apply to the first stage, this might be the best way to go. Just fully fork them, rather than trying to affect the generic functions that are used in all the other iterations.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -133,7 +145,7 @@ where
     last_u_or_v_array[0] = p_or_q_0;
 
     // compute and output p_or_q
-    tables[recursions - 1].eval(last_u_or_v_array)[0]
+    tables[recursions - 2].eval(last_u_or_v_array)[0]
```
benjaminsavage, line 148: This is just too subtle.

benjaminsavage (changes requested):
I'm not happy with how messy and subtle this code is getting. I suggest forking the first recursion and all the other steps to keep these separate concerns from leaking through everything.

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -76,74 +80,82 @@ fn compute_sum_share<F: PrimeField, const λ: usize, const P: usize>(zkp: &[F; P
...
-    let recursions = challenges.len();
+    let recursions_after_first = challenges.len() - 1;
 
     // compute Lagrange tables
+    let denominator_p_or_q_first = CanonicalLagrangeDenominator::<F, λ_FIRST>::new();
+    let table_first =
+        LagrangeTable::<F, λ_FIRST, 1>::new(&denominator_p_or_q_first, &challenges[0]);
     let denominator_p_or_q = CanonicalLagrangeDenominator::<F, λ>::new();
-    let tables = challenges
+    let tables = challenges[1..]
```
benjaminsavage, line 117: I like this much better

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -76,74 +80,82 @@ fn compute_sum_share<F: PrimeField, const λ: usize, const P: usize>(zkp: &[F; P
...
     // u and v vectors until it has length strictly less than λ.
-    // For this reason, we can safely assume that in the final proof, the last value
-    // in the ZKP is zero.
-    // A debug assert will help catch any errors while in development.
-    // set mask
-    debug_assert!(
-        last_u_or_v_array[λ - 1] == F::ZERO,
-        "Should not be overwriting non-zero values"
+    assert!(
+        last_u_or_v_values.len() < λ,
```
benjaminsavage, line 136: Nice. This is even better.

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -100,13 +103,13 @@ impl BatchToVerify {
             Vec::<[Fp61BitPrime; SPL]>::with_capacity(expected_len);
 
         // generate masks
-        // verifier on the right has p,
-        // therefore the right share is "implicitly sent" to the right ("communicated" via PRSS)
-        let (p_mask_from_left_prover, my_p_mask) = ctx.prss().generate_fields(record_counter);
+        // verifier on the left has p,
```
benjaminsavage, line 106: This is so confusing. I think maybe "left" and "right" aren't the best terms to use here because it all depends on your frame of reference. 

The verifier that is to the right of the prover has p...

...but each party has "p" for the prover to its left....

Maybe we should use the terminology from the paper. It says:

P_{i} and P_{i+1} both compute 'p(X)'

P_{i} and P_{i-1} both compute 'q(X)'

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -213,6 +219,55 @@ impl BatchToVerify {
...
+    where
+        C: Context,
+        U: Iterator<Item = Fp61BitPrime> + Send,
+        V: Iterator<Item = Fp61BitPrime> + Send,
+    {
+        const LRF: usize = LargeProofGenerator::RECURSION_FACTOR;
+        const SRF: usize = SmallProofGenerator::RECURSION_FACTOR;
+
+        // compute p_r
+        let p_r_right_prover = recursively_compute_final_check::<_, _, LRF, SRF>(
```
benjaminsavage, line 241: I really think this is backwards... if the prover is to the "right" of this party, to me that means that this is party P_{i-1} for the prover P_i. And according to the paper, this verifier should be computing q(r).

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -213,6 +219,55 @@ impl BatchToVerify {
...
+            self.p_mask_from_right_prover,
+        )
+        .await;
+        // compute q_r
+        let q_r_left_prover = recursively_compute_final_check::<_, _, LRF, SRF>(
+            stream::iter(v_from_left_prover),
+            challenges_for_left_prover,
+            self.q_mask_from_left_prover,
+        )
+        .await;
```
benjaminsavage, line 253: You can try_join these two awaits.

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -213,6 +219,55 @@ impl BatchToVerify {
...
+        // send to the left
+        let communication_ctx = ctx.set_total_records(1);
+
+        let send_channel =
+            communication_ctx.send_channel::<Fp61BitPrime>(ctx.role().peer(Direction::Right));
+        let receive_channel =
+            communication_ctx.recv_channel::<Fp61BitPrime>(ctx.role().peer(Direction::Left));
+
+        let ((), q_r_right_prover) = try_join(
+            send_channel.send(RecordId::FIRST, q_r_left_prover),
```
benjaminsavage, line 264: Since party P_{i-1} is the one with q(r), it needs to reveal it to P_{i+1}. That would be *left*, not *right*. That would be sending it back to the prover again.

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -239,14 +294,12 @@ impl ProofHashes {
...
 
-        let length = other_proofs.len();
-
         Self {
             hashes: once(compute_hash(first_proof))
                 .chain(
                     other_proofs
                         .iter()
-                        .take(length - 1) // we don't need to hash last proof
+                        //.take(length - 1) // we don't need to hash last proof
```
benjaminsavage, line 302: Remove

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -469,4 +531,226 @@ pub mod test {
             assert_eq!(helper_1_left, helper_2_right);
         });
     }
+
+    /// This is a helper function to generate `u`, `v` values.
+    ///
+    /// outputs `(my_u_and_v, u_from_right_prover, v_from_left_prover)`
```
benjaminsavage, line 537: So the paper says (on page 21):

"P_{i+1}, P_{i-1} compute vector u and v respectively as P_i does."

So I think this comment is wrong.

"u_from_right_prover" doesn't make sense. If the Prover (that is P_i) is to your right, then you are P_{i-1} in that context, meaning you should be computing *v*.

benjaminsavage (changes requested):
Mostly looks good, but I really think you have swapped u and v or p and q up.

ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
benjaminsavage: I tried and couldn't... because these are kind of not real async functions...

benjaminsavage (approved):

danielmasny:
@benjaminsavage The originial left, right was actually correct since the paper organizes the shares differently from us. E.g. our second helper has shares $(x_2,x_3)$, in the paper he has $(x_2,x_1)$ which basically swaps u, v and p, q.

AND Gates. Suppose the parties wish to compute z := x · y given input sharings $[[x]],[[y]]$. First observe that, holding
sharings $(x_i, x_{i−1})$ and $(y_i, y_{i−1})$, each party Pi can locally compute $z_i:= x_iy_i ⊕x_iy_{i−1} ⊕x_{i−1}y_i$ s.t. $z = z_0 ⊕z_1 ⊕z_2$,

PR #1152, opened 2024-06-19
8-bit sigmoid function through piecewise linear approximation

benjaminsavage:
For 8-bit signed values that are assumed to represent x = -8 to x = 8 (that is, each consecutive 8-bit value is 1/16th larger than the previous one), one can very closely approximate the sigmoid function $1/(1 + exp(-x))$ with just a piecewise linear function. Simply split up the range into 16 segments of width 16 (e.g. [0, 16), [16, 32), ...) and approximate each with a line segment.

Best of all, this function only requires 31 bits of communication (per helper) to compute this, less than a single multiplication in our Fp32BitField. The round-depth is only 3 since mostly it's all parallelised. It's also vectorised, so we can pack many parallel computations together. 

Here's a graph showing the true sigmoid function in blue, and this approximation in red. As you can see, they are quite close.

approximate_sigmoid

The 8-bit output should be interpreted as 256 values that lie on the interval [0, 1). Interpret the 8-bit output as an unsigned integer representing the numerator of a fraction where the denominator is 256. 

Here's a spreadsheet where I developed this algorithm, and it includes the full truth table checking the values generated: https://docs.google.com/spreadsheets/d/1udT5PyEcFQcbBrokwmJj8plTMOGZWg2ae-ei8GIWa7Y/edit?usp=sharing

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -121,7 +121,7 @@ macro_rules! boolean_array_impl_small {
             fn as_i128(&self) -> i128 {
                 let mut out: i128 = i128::try_from(self.as_u128()).unwrap();
                 let msb = (out >> $bits - 1) & 1;
-                out -= msb*(1 << $bits);
+                out -= msb * (1 << $bits);
                 out
             }
         }
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -9,7 +9,10 @@ use typenum::{U14, U2, U32, U8};
 
 use crate::{
     error::LengthError,
-    ff::{boolean::Boolean, ArrayAccess, Expand, Field, Serializable, U128Conversions},
+    ff::{
+        boolean::Boolean, ArrayAccess, Expand, Field, I128Conversions, Serializable,
+        U128Conversions,
+    },
     protocol::prss::{FromRandom, FromRandomU128},
     secret_sharing::{Block, SharedValue, StdArray, Vectorizable},
 };
@@ -111,6 +114,15 @@ macro_rules! boolean_array_impl_small {
 
             fn as_u128(&self) -> u128 {
                 (*self).into()
+            }
+        }
+
+        impl I128Conversions for $name {
+            fn as_i128(&self) -> i128 {
+                let mut out: i128 = i128::try_from(self.as_u128()).unwrap();
+                let msb = (out >> $bits - 1) & 1;
+                out -= msb*(1 << $bits);
+                out
             }
         }
 
@@ -569,6 +581,8 @@ macro_rules! boolean_array_impl {
                 use rand::{thread_rng, Rng};
                 use bitvec::bits;
 
+                use crate::ff::I128Conversions;
+
                 use super::*;
 
                 impl Arbitrary for $name {
@@ -632,6 +646,12 @@ macro_rules! boolean_array_impl {
                         assert_eq!(a * c, if bool::from(c) { a } else { $name::ZERO });
                         assert_eq!(a * &c, if bool::from(c) { a } else { $name::ZERO });
                     }
+                }
+
+                #[test]
+                fn signed_integer_conversions() {
+                    let v = BA4::truncate_from(10_u128);
+                    assert_eq!(v.as_i128(), -6);
                 }
 
                 #[test]
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/sigmoid.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/sigmoid.rs
@@ -1,6 +1,6 @@
 use std::{iter::repeat, ops::Not};
 
-use futures::future::{try_join, try_join4, try_join5, try_join_all};
+use futures::future::{try_join, try_join4, try_join5};
 
 use crate::{
     error::Error,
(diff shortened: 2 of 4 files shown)


ipa-core/src/ff/boolean_array.rs
```diff
@@ -634,6 +648,12 @@ macro_rules! boolean_array_impl {
                     }
                 }
 
+                #[test]
+                fn signed_integer_conversions() {
```
akoshelev, line 652: I suggest you test conversion for 0 and MAX values as well. If you plan to keep it in this module, I think we should have this test for all fields using $name

ipa-core/src/protocol/ipa_prf/boolean_ops/sigmoid.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    fn piecewise_linear_sigmoid_approximation(x: i128) -> u128 {
+        if x < -112 {
+            return 0;
+        }
+        if x < -96 {
+            return 1;
+        }
+        if x < -80 {
+            return 2 + (u128::try_from(x + 96).unwrap() >> 3);
```
akoshelev, line 181: I think there are too many `.unwrap` calls and you could benefit from changing the API of this function to 

```rust
// not sure about the error type, but it should be the same
fn piecewise_linear_sigmoid_approximation(x: i128) -> Result<u128, IntError> {
   match x {
        i128::MIN..-112 => 0,
        -112..-96 => 1,
        -96..-80 => 2 + u128::try_from(x + 96)? >> 3,
        ...
        96..112 => 254,
        _ => 255
   }
}
```

match is optional, you can still use `if` if you don't want to repeat the range boundaries at the expense of using 3 lines per statement instead of one

ipa-core/src/protocol/ipa_prf/boolean_ops/sigmoid.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    use crate::{
+        ff::{boolean::Boolean, boolean_array::BA8, I128Conversions, U128Conversions},
+        protocol::{context::Context, ipa_prf::boolean_ops::sigmoid::sigmoid, RecordId},
+        secret_sharing::{replicated::semi_honest::AdditiveShare, BitDecomposed, TransposeFrom},
+        test_executor::run,
+        test_fixture::{Reconstruct, Runner, TestWorld},
+    };
+
+    fn piecewise_linear_sigmoid_approximation(x: i128) -> u128 {
```
akoshelev, line 173: do you also want to use the real sigmoid and make use the approximation is within some delta from it? May help with correctness errors

danielmasny (commented):
I think we should check academic literature on approximating functions like sigmoids using Boolean operations. That would most likely also get us bounds on the approximation error and we don't have to hardwire everything by hand. It is not clear to me how good this approximation of a sigmoid is, nor whether it is optimal in terms of communication.

benjaminsavage:
My understanding is that it's most critical to closely approximate the sigmoid near x=0. The gradient is maximized there, and this has the biggest impact on the decision boundary / output of the neural network.

It's easy to approximate that region well since it's basically just a line at that point.

I've seen papers where they used a far less precise approximation, basically just three linear segments. A flat one for the left (y=0), a line segment sloping up (y=x+0.5) and a flat one for the right (y=1). And they were able to even do training with this.

I'm totally open to exploring a variety of functions here. For now I would like to start by getting just some proof of concept up and running. Then we can add alternative protocols and compare the cost / output of the neural net across them. How does that sound?

danielmasny:
> My understanding is that it's most critical to closely approximate the sigmoid near x=0. The gradient is maximized there, and this has the biggest impact on the decision boundary / output of the neural network.
> 
> It's easy to approximate that region well since it's basically just a line at that point.
> 
> I've seen papers where they used a far less precise approximation, basically just three linear segments. A flat one for the left (y=0), a line segment sloping up (y=x+0.5) and a flat one for the right (y=1). And they were able to even do training with this.
> 
> I'm totally open to exploring a variety of functions here. For now I would like to start by getting just some proof of concept up and running. Then we can add alternative protocols and compare the cost / output of the neural net across them. How does that sound?

Sounds great! I am very keen to explore the impact of the approximate sigmoid on the model accuracy.

benjaminsavage:
Awesome. Let's land this just to get a baseline (some cost, some precision), then we can pretty easily compare it to other approaches, since this is just a standalone function that we can easily swap out with alternatives. If either @danielmasny or @akoshelev wants to approve this PR, let's merge it.

ipa-core/src/protocol/ipa_prf/boolean_ops/sigmoid.rs
```diff
@@ -0,0 +1,244 @@
...
+        secret_sharing::{
+            replicated::semi_honest::AdditiveShare, BitDecomposed, SharedValue, TransposeFrom,
+        },
+        test_executor::run,
+        test_fixture::{Reconstruct, Runner, TestWorld},
+    };
+
+    fn piecewise_linear_sigmoid_approximation(x: i128) -> Result<u128, TryFromIntError> {
+        Ok(match x {
+            i128::MIN..=-113 => 0,
```
benjaminsavage, line 177: @akoshelev - I couldn't use the exclusive ranges because they're experimental right now. Should be available by 1.80.

akoshelev (approved):

danielmasny (approved):

Issue #1121, opened 2024-06-07
Helper binary requires public key for match key encryption, but it shouldn't


akoshelev:
In order to start a helper binary, one must provide the path to public key for match key encryption

```bash
 target/release/helper --network $CONFIGDIR/h$IDENTITY/pub/network.toml --identity $IDENTITY --tls-cert $CONFIGDIR/h$IDENTITY/pub/h$IDENTITY.pem --tls-key $CONFIGDIR/h$IDENTITY/h$IDENTITY.key --port 443 --mk-public-key $CONFIGDIR/h$IDENTITY/pub/h${IDENTITY}_mk.pub --mk-private-key $CONFIGDIR/h$IDENTITY/h${IDENTITY}_mk.key
```

there should be no reason for it - helpers only decrypt the match keys, never encrypt it

PR #1153, opened 2024-06-19
Helper Binary doesn't require public key [NOT READY TO MERGE]

eriktaubeneck:
This is meant to address #1121. It doesn't yet compile, but pushing to get some feedback on the approach.

### DIFF
--- a/ipa-core/src/hpke/registry.rs
+++ b/ipa-core/src/hpke/registry.rs
@@ -67,8 +67,12 @@ pub struct PrivateKeyOnly {
     sk: IpaPrivateKey,
 }
 
-pub trait PrivateKeyRegistry {
-    fn private_key(&self, key_id: KeyIdentifier) -> Option<&IpaPrivateKey>;
+impl From<IpaPrivateKey> for PrivateKeyOnly {
+    fn from(value: IpaPrivateKey) -> Self {
+        Self {
+            sk: value
+        }
+    }
 }
 
 /// A registry that holds all the keys available for helper/UA to use.
@@ -130,8 +134,8 @@ impl PublicKeyRegistry for KeyRegistry<PublicKeyOnly> {
     }
 }
 
-impl PrivateKeyRegistry for KeyRegistry<PrivateKeyOnly> {
-    fn private_key(&self, key_id: KeyIdentifier) -> Option<&IpaPrivateKey> {
+impl KeyRegistry<PrivateKeyOnly> {
+    pub fn private_key(&self, key_id: KeyIdentifier) -> Option<&IpaPrivateKey> {
         self.key(key_id).map(|v| &v.sk)
     }
 }
--- a/ipa-core/src/hpke/registry.rs
+++ b/ipa-core/src/hpke/registry.rs
@@ -58,6 +58,33 @@ impl Deref for PublicKeyOnly {
     }
 }
 
+// This newtype is necessary because IpaPrivateKey is an associated type from another crate (hpke).
+// The coherence rules prohibit us from implementing `PrivateKeyRegistry` both for our concrete type
+// `KeyPair` and for `IpaPrivateKey`, because the impls would overlap if hpke chose to define
+// `IpaPrivateKey` to be the same as `KeyPair`.
+pub struct PrivateKeyOnly(pub IpaPrivateKey);
+
+impl Deref for PrivateKeyOnly {
+    type Target = IpaPrivateKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<&KeyRegistry<KeyPair>> for KeyRegistry<PrivateKeyOnly> {
+    fn from(key_registry: &KeyRegistry<KeyPair>) -> Self {
+        let keys = key_registry
+            .keys
+            .iter()
+            .map(|k| PrivateKeyOnly(k.sk.clone()))
+            .collect::<Vec<_>>();
+        Self {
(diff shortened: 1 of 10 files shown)


ipa-core/src/hpke/registry.rs
```diff
@@ -62,6 +62,19 @@ pub trait PublicKeyRegistry {
...
+}
+
+impl From<IpaPrivateKey> for PrivateKeyOnly {
+    fn from(value: IpaPrivateKey) -> Self {
+        Self {
+            sk: value
+        }
+    }
+}
+
```
eriktaubeneck, line 77: Without this, I get this compile error:
```
error[E0277]: the trait bound `PrivateKeyOnly: From<hpke::dhkex::x25519::PrivateKey>` is not satisfied
   --> ipa-core/src/config.rs:245:31
    |
245 |     Ok(KeyRegistry::from_keys([IpaPrivateKey::from_bytes(&sk)?]))
    |        ---------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<hpke::dhkex::x25519::PrivateKey>` is not implemented for `PrivateKeyOnly`, which is required by `hpke::dhkex::x25519::PrivateKey: Into<_>`
    |        |
    |        required by a bound introduced by this call
    |
    = note: required for `hpke::dhkex::x25519::PrivateKey` to implement `Into<PrivateKeyOnly>`
note: required by a bound in `KeyRegistry::<K>::from_keys`
   --> ipa-core/src/hpke/registry.rs:91:41
    |
91  |     pub fn from_keys<const N: usize, I: Into<K>>(pairs: [I; N]) -> Self {
    |                                         ^^^^^^^ required by this bound in `KeyRegistry::<K>::from_keys`
```

with it, I get:
```
error[E0119]: conflicting implementations of trait `From<PrivateKeyOnly>` for type `PrivateKeyOnly`
  --> ipa-core/src/hpke/registry.rs:70:1
   |
70 | impl From<IpaPrivateKey> for PrivateKeyOnly {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T> From<T> for T;
```

akoshelev: I think it may be the same issue as @andyleiserson faced when implementing `PublicKeyOnly` (see the comment above) 

I am not sure why you need a wrapper around secret key - can you use `KeyRegistry<IpaPrivateKey>` instead?

eriktaubeneck: That's actually what I tried first, but got some other error, and went down this path (following what Andy had done.) Let me give that a try again.

ipa-core/src/hpke/registry.rs
```diff
@@ -102,13 +107,23 @@ impl KeyRegistry<KeyPair> {
...
 }
 
+impl PrivateKeyRegistry for KeyRegistry<IpaPrivateKey> {
+    #[must_use]
+    fn private_key(&self, key_id: KeyIdentifier) -> Option<&IpaPrivateKey> {
+        self.key(key_id)
+    }
+}
+
+
```
eriktaubeneck, line 126: This is giving me the following compile error (thought it seems to be the same approach as `PublicKeyRegistry` just below):
```
error[E0119]: conflicting implementations of trait `PrivateKeyRegistry` for type `KeyRegistry<hpke::registry::KeyPair>`
   --> ipa-core/src/hpke/registry.rs:119:1
    |
112 | impl PrivateKeyRegistry for KeyRegistry<KeyPair> {
    | ------------------------------------------------ first implementation here
...
119 | impl PrivateKeyRegistry for KeyRegistry<IpaPrivateKey> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `KeyRegistry<hpke::registry::KeyPair>`
    ```

Comment (akoshelev) [reply ipa-core/src/hpke/registry.rs:126]: ah, yea this is the limitation of Rust trait system. You can't implement `Foo` for `Bar<X>` and `Bar<Y>` - it is considered to be the same definition. I think you may want to define a `PrivateKeyContainer` trait and then implement it for `KeyPair` and `IpaPrivateKey`. then you can implement `PrivateKeyRegistry` for any `K: PrivateKeyContainer`

```rust
impl <K: PrivateKeyContainer> PrivateKeyRegistry for KeyRegistry<K> { .. }
```

eriktaubeneck: ok, thanks. (I'm still confused why we can implement `PublicKeyRegistry` for both `KeyRegistry<KeyPair>` and `KeyRegistry<PublicKeyOnly>` just below. That seems to do exactly what this is complaining about.)

akoshelev: ah, maybe because of this

```rust
// This newtype is necessary because IpaPublicKey is an associated type from another crate (hpke).
// The coherence rules prohibit us from implementing `PublicKeyRegistry` both for our concrete type
// `KeyPair` and for `IpaPublicKey`, because the impls would overlap if hpke chose to define
// `IpaPublicKey` to be the same as `KeyPair`.
pub struct PublicKeyOnly(pub IpaPublicKey);
```

My statement about `Bar<X>` and `Bar<Y>` was wrong, I confused generics vs associated types.

eriktaubeneck:
Closing in favor of #1161

PR #1154, opened 2024-06-20
Create test addition circuit

akoshelev:
This change adds a new simple test circuit and integration tests for it. This circuit does not do anything sophisticated and the need for it is dictated by #1141 where I need to test that inputs of size > 2Gb can be handled by our MPC helpers. 

I tried to use `test-multiply` but it just takes too long to process inputs of this size. I estimated ~300M field value pairs need to be submitted to exceed $2^{31}-1$ input size and that makes that circuit to run for 8+ mins. 

This circuit is much faster for two reasons. First, it does not require communication. Second, it generates input and shares it using iterators. That makes a big difference - my experiments showed that it takes roughly 9 mins to get `test-multiply` inputs delivered and ~50 seconds for `test-add`. It can be further optimized because we collect everything. in `.share`, but this implementation is good enough for me.

My goal is to use this circuit in a integration test that we run only in CI and that checks inputs greater than $2^31-1$ can be sent via our `report_collector` client

### DIFF
--- a/ipa-core/src/protocol/step.rs
+++ b/ipa-core/src/protocol/step.rs
@@ -8,7 +8,7 @@ pub enum ProtocolStep {
     #[step(child = crate::protocol::ipa_prf::step::IpaPrfStep)]
     IpaPrf,
     Multiply,
-    Add,
+    PrimeFieldAddition,
     #[cfg(any(test, feature = "test-fixture"))]
     #[step(count = 10, child = crate::test_fixture::step::TestExecutionStep)]
     Test(usize),
--- a/ipa-core/src/protocol/step.rs
+++ b/ipa-core/src/protocol/step.rs
@@ -8,6 +8,7 @@ pub enum ProtocolStep {
     #[step(child = crate::protocol::ipa_prf::step::IpaPrfStep)]
     IpaPrf,
     Multiply,
+    Add,
     #[cfg(any(test, feature = "test-fixture"))]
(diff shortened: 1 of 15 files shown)


ipa-core/src/query/executor.rs
```diff
@@ -81,6 +81,18 @@ pub fn execute(
                 Box::pin(execute_test_multiply::<Fp32BitPrime>(prss, gateway, input))
             })
         }
+        #[cfg(any(test, feature = "weak-field"))]
+        (QueryType::TestAdd, FieldType::Fp31) => {
```
akoshelev, line 85: Rename to `send lots of data to helpers`

ipa-core/src/protocol/step.rs
```diff
@@ -8,6 +8,7 @@ pub enum ProtocolStep {
     #[step(child = crate::protocol::ipa_prf::step::IpaPrfStep)]
     IpaPrf,
     Multiply,
+    Add,
```
benjaminsavage, line 11: Let's call this something else... like "send_random_data_to_helpers"

benjaminsavage (approved):
LGTM

ipa-core/src/protocol/step.rs
akoshelev: branstorming out loud - maybe `AdditionInPrimeField` to indicate what it is actually doing and that it does not require communication?

PR #1156, opened 2024-06-20
Tests for result length

bmcase:
adding an assert that results.len() == max_breakdown_key, breaks 4 compact gate tests (result len = 256 and max breakdown key = 20).  We probably want to add a check for this and figure out why it wouldn't always be true already.  

Current code ignores extra breakdown key values if they are zero, but with DP noise coming those end up non-zero.  Probably the solution is to not return too long a result at any time. 

image

### DIFF
--- a/ipa-core/src/cli/playbook/ipa.rs
+++ b/ipa-core/src/cli/playbook/ipa.rs
@@ -5,6 +5,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use bitvec::macros::internal::funty::Fundamental;
 use futures_util::future::try_join_all;
 use generic_array::GenericArray;
 use rand::rngs::StdRng;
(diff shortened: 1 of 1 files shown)


ipa-core/src/cli/playbook/ipa.rs
```diff
@@ -5,6 +5,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use bitvec::macros::internal::funty::Fundamental;
```
akoshelev, line 8: oh this does not seem right - we shouldn't require this import

bmcase: hmm yeah, not sure how that got added.  I'll try to remove it.

bmcase:
close to clean up PRs.  Related to issues tracked here https://github.com/private-attribution/ipa/issues/1201

PR #1157, opened 2024-06-21
Deprecate SingleUVBlock

danielmasny:
Deprecates SingleUVBlock and replaces streams over arrays of field elements to streams over field elements.


benjaminsavage (approved):
I like this a lot better. Thanks for making this improvement!

PR #1158, opened 2024-06-21
Upgrade OPRF Conversion to Malicious using DZKPs

danielmasny:
upgrades the conversion that uses Boolean integer addition to malicious security by using upgraded contexts.

It still needs to be generalized to use traits, such that it works for DZKP semi honest as well as DZKP malicious. The most important structure is already in the PR.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
@@ -12,7 +12,7 @@ use crate::{
     protocol::{
         basics::{partial_reveal, BooleanProtocols},
         boolean::step::TwoHundredFiftySixBitOpStep,
-        context::Context,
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -153,18 +165,21 @@ where
...
-        ctx.narrow(&Step::IntegerAddMaskToX),
+        m_ctx.narrow(&Step::IntegerAddMaskToX),
         record_id,
         &sh_rs,
         &x,
     )
     .await?;
 
+    // validate before reveal
+    validator.validate(0).await?;
```
benjaminsavage, line 176: I really don't like this passing in 0 as an argument. I'm not a fan of this API design.

benjaminsavage (approved):

PR #1161, opened 2024-06-23
add new PrivateKeyOnly struct to handle KeyRegistry with only a private key

eriktaubeneck:
This adds a new struct `PublicKeyOnly`, similar to the existing `PrivateKeyOnly`, allowing us to only provide a private key for the current helper. A helpers own public key should not need to be known when running.

### DIFF
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -16,7 +16,7 @@ mod info;
 mod registry;
 
 pub use info::Info;
-pub use registry::{KeyPair, KeyRegistry, PublicKeyOnly, PublicKeyRegistry};
+pub use registry::{KeyPair, KeyRegistry, PrivateKeyOnly, PrivateKeyRegistry, PublicKeyRegistry};
 
 use crate::{
     ff::{GaloisField, Serializable as IpaSerializable},
@@ -94,8 +94,8 @@ impl From<io::Error> for CryptError {
 /// If ciphertext cannot be opened for any reason.
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a>(
-    key_registry: &KeyRegistry<KeyPair>,
+pub fn open_in_place<'a, R: PrivateKeyRegistry>(
+    key_registry: &R,
     enc: &[u8],
     ciphertext: &'a mut [u8],
     info: &Info,
@@ -214,7 +214,7 @@ mod tests {
 
         pub fn new(keys: usize, mut rng: R) -> Self {
             Self {
-                registry: KeyRegistry::random(keys, &mut rng),
+                registry: KeyRegistry::<KeyPair>::random(keys, &mut rng),
                 rng,
                 epoch: 0,
             }
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -96,7 +96,7 @@ impl From<io::Error> for CryptError {
 /// If ciphertext cannot be opened for any reason.
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a, R: PrivateKeyRegistry + ?Sized>(
+pub fn open_in_place<'a, R: PrivateKeyRegistry>(
     key_registry: &R,
     enc: &[u8],
     ciphertext: &'a mut [u8],
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -16,7 +16,9 @@ mod info;
 mod registry;
 
 pub use info::Info;
-pub use registry::{KeyPair, KeyRegistry, PrivateKeyOnly, PrivateKeyRegistry, PublicKeyRegistry};
+pub use registry::{
+    KeyPair, KeyRegistry, PrivateKeyOnly, PrivateKeyRegistry, PublicKeyOnly, PublicKeyRegistry,
+};
 
 use crate::{
     ff::{GaloisField, Serializable as IpaSerializable},
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -96,7 +96,7 @@ impl From<io::Error> for CryptError {
 /// If ciphertext cannot be opened for any reason.
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a, R: PrivateKeyRegistry>(
+pub fn open_in_place<'a, R: PrivateKeyRegistry + ?Sized>(
     key_registry: &R,
     enc: &[u8],
     ciphertext: &'a mut [u8],
--- a/ipa-core/src/hpke/registry.rs
+++ b/ipa-core/src/hpke/registry.rs
@@ -58,13 +58,57 @@ impl Deref for PublicKeyOnly {
     }
 }
 
+// This newtype is necessary because IpaPrivateKey is an associated type from another crate (hpke).
+// The coherence rules prohibit us from implementing `PrivateKeyRegistry` both for our concrete type
+// `KeyPair` and for `IpaPrivateKey`, because the impls would overlap if hpke chose to define
+// `IpaPrivateKey` to be the same as `KeyPair`.
+pub struct PrivateKeyOnly(pub IpaPrivateKey);
+
+impl Deref for PrivateKeyOnly {
+    type Target = IpaPrivateKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<&KeyRegistry<KeyPair>> for KeyRegistry<PrivateKeyOnly> {
+    fn from(key_registry: &KeyRegistry<KeyPair>) -> Self {
+        let keys = key_registry
+            .keys
+            .iter()
+            .map(|k| PrivateKeyOnly(k.sk.clone()))
+            .collect::<Vec<_>>();
+        Self {
+            keys: keys.into_boxed_slice(),
+        }
+    }
+}
+
 pub trait PublicKeyRegistry {
     fn public_key(&self, key_id: KeyIdentifier) -> Option<&IpaPublicKey>;
+}
+
+pub trait PrivateKeyRegistry {
+    fn private_key(&self, key_id: KeyIdentifier) -> Option<&IpaPrivateKey>;
 }
 
 /// A registry that holds all the keys available for helper/UA to use.
 pub struct KeyRegistry<K> {
     keys: Box<[K]>,
+}
+
+impl KeyRegistry<PrivateKeyOnly> {
+    // #[cfg(any(test, feature = "test-fixture"))]
+    pub fn random<R: rand::RngCore + rand::CryptoRng>(keys_count: usize, r: &mut R) -> Self {
+        let keys = (0..keys_count)
+            .map(|_| PrivateKeyOnly(KeyPair::gen(r).sk))
+            .collect::<Vec<_>>();
+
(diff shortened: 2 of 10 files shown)


ipa-core/src/hpke/registry.rs
```diff
@@ -102,13 +146,22 @@ impl KeyRegistry<KeyPair> {
             keys: keys.into_boxed_slice(),
         }
     }
+}
 
+impl PrivateKeyRegistry for KeyRegistry<KeyPair> {
```
akoshelev, line 151: with this change, do we have a use case to keep `KeyRegistry<KeyPair>` outside of test module ?

ipa-core/src/query/runner/oprf_ipa.rs
```diff
@@ -31,12 +31,12 @@ use crate::{
...
 pub struct OprfIpaQuery<'a, HV> {
     config: IpaQueryConfig,
-    key_registry: Arc<KeyRegistry<KeyPair>>,
+    key_registry: Arc<KeyRegistry<PrivateKeyOnly>>,
     phantom_data: PhantomData<&'a HV>,
 }
 
 impl<'a, HV> OprfIpaQuery<'a, HV> {
-    pub fn new(config: IpaQueryConfig, key_registry: Arc<KeyRegistry<KeyPair>>) -> Self {
+    pub fn new(config: IpaQueryConfig, key_registry: Arc<KeyRegistry<PrivateKeyOnly>>) -> Self {
```
akoshelev, line 39: can you try `Arc<dyn PrivateKeyRegistry>` here? I think in this case you wouldn't require conversion in unit tests

ipa-core/src/report.rs
```diff
@@ -587,9 +587,9 @@ mod test {
                 .collect(),
         };
 
-        let enc_key_registry = KeyRegistry::random(1, &mut rng);
+        let enc_key_registry = KeyRegistry::<KeyPair>::random(1, &mut rng);
         let enc_key_id = 0;
-        let dec_key_registry = KeyRegistry::random(1, &mut rng);
+        let dec_key_registry = KeyRegistry::<PrivateKeyOnly>::random(1, &mut rng);
```
akoshelev, line 592: I think you should use `KeyPair` here too because `decrypt` takes `PrivateKeyRegistry` trait. That answers my question why you need `random` but also indicates that there is a change that needs to be made and `random` for `PrivateKeyOnly` should go away

ipa-core/src/hpke/mod.rs
```diff
@@ -94,8 +96,8 @@ impl From<io::Error> for CryptError {
 /// If ciphertext cannot be opened for any reason.
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a>(
-    key_registry: &KeyRegistry<KeyPair>,
+pub fn open_in_place<'a, R: PrivateKeyRegistry>(
```
akoshelev, line 99: 👍

ipa-core/src/hpke/registry.rs
```diff
@@ -58,15 +58,59 @@ impl Deref for PublicKeyOnly {
...
+
+impl Deref for PrivateKeyOnly {
+    type Target = IpaPrivateKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<&KeyRegistry<KeyPair>> for KeyRegistry<PrivateKeyOnly> {
```
akoshelev, line 75: I think with traits you shouldn't require this conversion

ipa-core/src/query/runner/oprf_ipa.rs
eriktaubeneck: This gives me the following compilation error. 

```
error[E0277]: the size for values of type `dyn hpke::registry::PrivateKeyRegistry` cannot be known at compilation time
   --> ipa-core/src/query/runner/oprf_ipa.rs:84:38
    |
84  | ...                   .decrypt(key_registry.as_ref())
    |                        ------- ^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |                        |
    |                        required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `dyn hpke::registry::PrivateKeyRegistry`
note: required by a bound in `report::EncryptedOprfReport::<BK, TV, TS, B>::decrypt`
   --> ipa-core/src/report.rs:313:24
    |
311 |     pub fn decrypt(
    |            ------- required by a bound in this associated function
312 |         &self,
313 |         key_registry: &impl PrivateKeyRegistry,
    |                        ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `EncryptedOprfReport::<BK, TV, TS, B>::decrypt`
help: consider removing this method call, as the receiver has type `std::sync::Arc<(dyn hpke::registry::PrivateKeyRegistry + 'static)>` and `std::sync::Arc<(dyn hpke::registry::PrivateKeyRegistry + 'static)>: Sized` trivially holds
    |
84  -                             .decrypt(key_registry.as_ref())
84  +                             .decrypt(key_registry)
    |
help: consider relaxing the implicit `Sized` restriction
   --> ipa-core/src/report.rs:313:47
    |
313 |         key_registry: &impl PrivateKeyRegistry + ?Sized,
    |                                                ++++++++

warning: `ipa-core` (lib test) generated 1 warning (1 duplicate)
error: could not compile `ipa-core` (lib test) due to 1 previous error; 1 warning emitted
```

ipa-core/src/hpke/registry.rs
eriktaubeneck: Yes, it's still used in `ipa-core/src/cli/keygen.rs`.

ipa-core/src/query/runner/oprf_ipa.rs
eriktaubeneck: The suggestion 
```
84  -                             .decrypt(key_registry.as_ref())
84  +                             .decrypt(key_registry)
``` 
produces an incompatible type, as decrypt expects a reference. If we borrow here, we then get:
```
the trait `PrivateKeyRegistry` is not implemented for `Arc<dyn PrivateKeyRegistry>`
```

Adding `?Sized` to in `ipa-core/src/report.rs` simply leads to errors elsewhere about the size being unknown at compile times. 

It seems like keeping the conversion might be the most straight forward path, but happy to be pointed in a different direction.

akoshelev: pushed the change that can make `OprfIpa` accept `Arc<dyn>`. Again, I realized how clunky dynamic dispatch is in Rust and we should probably just parametrize `OprfIpa` with another generic

eriktaubeneck: ty! that works. now let me try to remove the implementation of `From`.

ipa-core/src/hpke/registry.rs
```diff
@@ -58,10 +58,28 @@ impl Deref for PublicKeyOnly {
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 pub trait PublicKeyRegistry {
     fn public_key(&self, key_id: KeyIdentifier) -> Option<&IpaPublicKey>;
 }
 
+pub trait PrivateKeyRegistry: Send + Sync + 'static {
```
akoshelev, line 79: Now that you're using generics, some of these bounds may no longer be required.

akoshelev (approved):
looks good to me - thanks for making this change.

ipa-core/src/hpke/registry.rs
eriktaubeneck: good catch. I removed the function bounds elsewhere, so this should be able to go.

eriktaubeneck: Ok, nevermind. Looks like they are still all required:
Sync and Send:
```
error[E0277]: `R` cannot be sent between threads safely
   --> ipa-core/src/query/executor.rs:107:17
    |
107 | /                 Box::pin(
108 | |                     OprfIpaQuery::<BA16, R>::new(ipa_config, key_registry)
109 | |                         .execute(ctx, config.size, input)
110 | |                         .then(|res| ready(res.map(|out| Box::new(out) as Box<dyn Result>))),
111 | |                 )
    | |_________________^ `R` cannot be sent between threads safely
    |
    = note: required for `Arc<R>` to implement `std::marker::Send`
note: required because it appears within the type `OprfIpaQuery<'_, boolean_array_16::BA16, R>`
   --> ipa-core/src/query/runner/oprf_ipa.rs:32:12
    |
32  | pub struct OprfIpaQuery<'a, HV, R: PrivateKeyRegistry> {
    |            ^^^^^^^^^^^^
note: required because it's used within this `async` fn body
   --> ipa-core/src/query/runner/oprf_ipa.rs:57:5
    |
57  |     #[tracing::instrument("oprf_ipa_query", skip_all, fields(sz=%query_size))]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: required because it appears within the type `Map<impl Future<Output = Result<Vec<AdditiveShare<BA16>>, Error>>, {closure@executor.rs:110:31}>`
   --> /Users/eriktaubeneck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/future/map.rs:15:14
    |
15  |     pub enum Map<Fut, F> {
    |              ^^^
note: required because it appears within the type `Map<impl Future<Output = Result<Vec<AdditiveShare<BA16>>, Error>>, {closure@executor.rs:110:31}>`
   --> /Users/eriktaubeneck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/future/mod.rs:50:5
    |
50  |     Map<Fut, F>(
    |     ^^^
note: required because it appears within the type `Flatten<Map<impl Future<Output = Result<Vec<AdditiveShare<BA16>>, Error>>, {closure@executor.rs:110:31}>, Ready<Result<Box<...>, ...>>>`
   --> /Users/eriktaubeneck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/future/flatten.rs:13:14
    |
13  |     pub enum Flatten<Fut1, Fut2> {
    |              ^^^^^^^
note: required because it appears within the type `Then<impl Future<Output = Result<Vec<AdditiveShare<BA16>>, Error>>, Ready<Result<Box<dyn Result>, Error>>, {closure@executor.rs:110:31}>`
   --> /Users/eriktaubeneck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/future/mod.rs:71:5
    |
71  |     Then<Fut1, Fut2, F>(
    |     ^^^^
    = note: required for the cast from `Pin<Box<Then<impl Future<Output = Result<Vec<AdditiveShare<BA16>>, Error>>, Ready<Result<Box<dyn Result>, Error>>, ...>>>` to `Pin<Box<dyn futures::Future<Output = std::result::Result<Box<(dyn query::executor::Result + 'static)>, error::Error>> + std::marker::Send>>`
help: consider further restricting this bound
    |
63  | pub fn execute<R: PrivateKeyRegistry + std::marker::Send>(
help: consider further restricting this bound
    |
63  | pub fn execute<R: PrivateKeyRegistry + std::marker::Sync>(
```

`static:
```
error[E0310]: the parameter type `R` may not live long enough
   --> ipa-core/src/query/executor.rs:107:17
    |
107 | /                 Box::pin(
108 | |                     OprfIpaQuery::<BA16, R>::new(ipa_config, key_registry)
109 | |                         .execute(ctx, config.size, input)
110 | |                         .then(|res| ready(res.map(|out| Box::new(out) as Box<dyn Result>))),
111 | |                 )
    | |                 ^
    | |                 |
    | |_________________the parameter type `R` must be valid for the static lifetime...
    |                   ...so that the type `R` will meet its required lifetime bounds
    |
help: consider adding an explicit lifetime bound
    |
63  | pub fn execute<R: PrivateKeyRegistry + 'static>(
```

PR #1164, opened 2024-06-24
Integrating DZKP validation in validator 

danielmasny:
We connect the validator with the DZKP generation and validation process. 

Since the previous protocol did not follow the paper and caused issues with the DZKPs, 
we have changed the multiplication protocol and updated the documentation:
/// The multiplication follows the academic papers
/// `https://eprint.iacr.org/2019/1390`,
/// `https://eprint.iacr.org/2023/909.pdf`,
/// which are compatible with distributed zero-knowledge proofs
/// with the only difference that in the paper Helper `i` holds shares `x_{i}, x_{i-1}`
/// whereas in our implementation he holds `x_{i}, x_{i+1}`
///
/// The multiplication protocol works as follows, given the shares `x_1, x_2, x_3`, `y_1, y_2, y_3`
/// such that `x = x_1 + x_2 + x_3`, `y = y_1 + y_2 + y_3`
/// the strategy is to compute `xy` via
/// `x * y = (x_1 + x_2 + x_3) * ( y_1 + y_2 + y_3) =`
/// `  x_1 * y_1 + x_1 * y_2 + x_2 * y_1`
/// `+ x_2 * y_2 + x_2 * y_3 + x_3 * y_2`
/// `+ x_3 * y_3 + x_3 * y_1 + x_1 * y_3`
///
/// Each helper has shares `(x_left, x_right)`, `(y_left, y_right)` and computes
/// `z_left = x_left * y_left + x_left * y_right + x_right * y_left + PRSS_left - PRSS_right`
/// and sends `z_left` to the helper on the left
/// which treats it as `z_right`
/// since it has been received from the right party from his perspective
/// Therefore, each helper has shares `(z_left, z_right)`.

PR #1165, opened 2024-06-25
chunk_array

danielmasny:
adding chunk methods that returns iterator over arrays

this allows us to use iterators instead of streams when chunks is needed during the DZKP validation.

It resolves issues with Mutex and await points in https://github.com/private-attribution/ipa/pull/1164

### DIFF
--- a/ipa-core/src/utils/arraychunks.rs
+++ b/ipa-core/src/utils/arraychunks.rs
@@ -2,6 +2,9 @@ use crate::{ff::Fp61BitPrime, secret_sharing::SharedValue};
 
 #[allow(dead_code)]
 pub trait ArrayChunkIterator: Iterator {
+    /// This function returns an iterator that yields arrays of size `L`.
+    /// When the amount of items in the iterator is not a multiple of `L`
+    /// the iterator fills the last array with zero elements.
     fn chunk_array<const L: usize>(self) -> ArrayChunk<Self, L>
     where
         Self: Sized,
--- a/ipa-core/src/utils/arraychunks.rs
+++ b/ipa-core/src/utils/arraychunks.rs
@@ -1,0 +1,70 @@
+use crate::{ff::Fp61BitPrime, secret_sharing::SharedValue};
+
+#[allow(dead_code)]
+pub trait ArrayChunkIterator: Iterator {
+    fn chunk_array<const L: usize>(self) -> ArrayChunk<Self, L>
+    where
+        Self: Sized,
+    {
+        ArrayChunk { iter: self }
+    }
+}
+
+pub struct ArrayChunk<I, const L: usize> {
+    iter: I,
+}
+
+impl<I: Iterator<Item = Fp61BitPrime>> ArrayChunkIterator for I {}
+
+#[allow(clippy::while_let_on_iterator)]
+impl<I, const L: usize> Iterator for ArrayChunk<I, L>
+where
+    I: Iterator<Item = Fp61BitPrime>,
+{
+    type Item = [I::Item; L];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut array = [Fp61BitPrime::ZERO; L];
+        let mut counter = 0usize;
+        while let Some(element) = self.iter.next() {
+            array[counter] = element;
+            counter += 1;
+            if counter == L {
+                break;
+            }
+        }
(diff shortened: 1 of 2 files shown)


benjaminsavage (approved):
This looks really good

ipa-core/src/utils/arraychunks.rs
```diff
@@ -0,0 +1,73 @@
+use crate::{ff::Fp61BitPrime, secret_sharing::SharedValue};
+
+#[allow(dead_code)]
+pub trait ArrayChunkIterator: Iterator {
```
benjaminsavage, line 4: Is it worth implementing ExactSizeIterator when the input is also an ExactSizeIterator?

ipa-core/src/utils/arraychunks.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut array = [Fp61BitPrime::ZERO; L];
+        let mut counter = 0usize;
+        while let Some(element) = self.iter.next() {
+            array[counter] = element;
+            counter += 1;
+            if counter == L {
+                break;
+            }
+        }
```
martinthomson, line 38: I wonder whether this could be:

```rust
let mut array = [Fp61BitPrime::ZERO; L];
for i in 0..L {
  array[i] = self.iter.next().unwrap_or(Fp61BitPrime::ZERO);
}
array
```

Or even as short as:

```rust
self.iter.chain(repeat(Fp61BitPrime::ZERO)).take(L).collect::<[I::Item; L]>()
```

(I'm not sure about the ability to collect into an array though.)

ipa-core/src/utils/arraychunks.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct ArrayChunk<I, const L: usize> {
+    iter: I,
+}
+
+impl<I: Iterator<Item = Fp61BitPrime>> ArrayChunkIterator for I {}
+
+#[allow(clippy::while_let_on_iterator)]
+impl<I, const L: usize> Iterator for ArrayChunk<I, L>
+where
+    I: Iterator<Item = Fp61BitPrime>,
```
martinthomson, line 25: Does this have to be `Item = Fp61BitPrime` or could it be `Item = T where T: Copy`?

danielmasny: I was considering using a type that implements default or has a zero defined on it since I want to fill the rest of the array with some default or zero elements.

ipa-core/src/utils/arraychunks.rs
danielmasny: That would be nice. I didn't find a better way of doing it. There are two main issues, I need to increment the state of the iterator. The other issue is that if there is no element left, I want to return None rather than an array with all zeros.

ipa-core/src/utils/arraychunks.rs
danielmasny: I think right now, I only have one use-case for this for which I wouldn't need it. In our use case it is also correct to fill the array with zeros. In a general use case it might be a bit weird, so I didn't want to make it more general.

ipa-core/src/utils/arraychunks.rs
martinthomson: Ah, yes.  The code ends up being a little more complex.  The latter fails due to `chain()` consuming the iterator.  The former though...

```rust
if let Some(v) = self.iter.next() {
  let mut array = [Fp61BitPrime::ZERO; L];
  array[0] = v;
  for i in 1..L {
    array[i] = self.iter.next().unwrap_or(Fp61BitPrime::ZERO);
  }
  Some(array)
} else {
  None
}
```

ipa-core/src/utils/arraychunks.rs
martinthomson: Relying on an implementation of `Default` is probably OK, provided that we can do that for the types we care about.  A default of zero is expected, so no problem there.

ipa-core/src/utils/arraychunks.rs
martinthomson: I think that this should implement ExactSizeIterator where possible, if only because it makes some of the code might like to depend on it (for accurate pre-allocation, if nothing else).

PR #1166, opened 2024-06-25
Replace uv stream in verifier and validation

danielmasny:
replaces streams with iterators by using chunk_array instead of chunks

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
@@ -1,14 +1,12 @@
-use std::{
-    iter::{self},
-};
+use std::iter::{self};
 
 use crate::{
     ff::PrimeField,
     protocol::ipa_prf::malicious_security::lagrange::{
         CanonicalLagrangeDenominator, LagrangeTable,
     },
+    utils::arraychunks::ArrayChunkIterator,
 };
-use crate::utils::arraychunks::ArrayChunkIterator;
 
 /// This function computes the shares that sum to zero from the zero-knowledge proofs.
 ///
@@ -107,18 +105,12 @@ fn recurse_u_or_v<'a, F: PrimeField, J, const λ: usize>(
 where
     J: Iterator<Item = F> + 'a,
 {
-    u_or_v_iterator.chunk_array::<λ>()
-        .map(|x| {
-        lagrange_table.eval(&x)[0]
-    })
+    u_or_v_iterator
+        .chunk_array::<λ>()
+        .map(|x| lagrange_table.eval(&x)[0])
 }
 
-pub fn recursively_compute_final_check<
-    F: PrimeField,
-    J,
-    const λ_FIRST: usize,
-    const λ: usize,
->(
+pub fn recursively_compute_final_check<F: PrimeField, J, const λ_FIRST: usize, const λ: usize>(
     u_or_v_iterator: J,
     challenges: &[F],
     p_or_q_0: F,
@@ -140,9 +132,10 @@ where
 
     // generate & evaluate recursive streams
     // to compute last array
-    let mut iterator: Box<dyn Iterator<Item = F>> = Box::new(
-        recurse_u_or_v::<_, _, λ_FIRST>(u_or_v_iterator, &table_first),
-    );
+    let mut iterator: Box<dyn Iterator<Item = F>> = Box::new(recurse_u_or_v::<_, _, λ_FIRST>(
+        u_or_v_iterator,
+        &table_first,
+    ));
     // all following recursion except last one
     for lagrange_table in tables.iter().take(recursions_after_first - 1) {
         iterator = Box::new(recurse_u_or_v::<_, _, λ>(iterator, lagrange_table));
@@ -303,8 +296,8 @@ mod test {
             .collect::<Vec<Fp31>>();
         assert_eq!(u_or_v_2, to_field(&U_2));
 
-        let u_or_v_3 = recurse_u_or_v::<_, _, 4>(u_or_v_2.into_iter(), &tables[1])
-            .collect::<Vec<_>>();
+        let u_or_v_3 =
+            recurse_u_or_v::<_, _, 4>(u_or_v_2.into_iter(), &tables[1]).collect::<Vec<_>>();
 
         assert_eq!(u_or_v_3, to_field(&U_3[..2]));
 
@@ -315,8 +308,8 @@ mod test {
             u_or_v_3[0], // move first element to the end
         ];
 
-        let p_final = recurse_u_or_v::<_, _, 4>(u_or_v_3_masked.into_iter(), &tables[2])
-            .collect::<Vec<_>>();
+        let p_final =
+            recurse_u_or_v::<_, _, 4>(u_or_v_3_masked.into_iter(), &tables[2]).collect::<Vec<_>>();
 
         assert_eq!(p_final[0].as_u128(), EXPECTED_P_FINAL);
 
@@ -387,12 +380,10 @@ mod test {
         // uv values in input format
         let v_1 = to_field(&V_1);
 
-        let u_or_v_2 = recurse_u_or_v(v_1.into_iter(), &tables[0])
-            .collect::<Vec<_>>();
+        let u_or_v_2 = recurse_u_or_v(v_1.into_iter(), &tables[0]).collect::<Vec<_>>();
         assert_eq!(u_or_v_2, to_field(&V_2));
 
-        let u_or_v_3 = recurse_u_or_v(u_or_v_2.into_iter().into_iter(), &tables[1])
-            .collect::<Vec<_>>();
+        let u_or_v_3 = recurse_u_or_v(u_or_v_2.into_iter(), &tables[1]).collect::<Vec<_>>();
 
         assert_eq!(u_or_v_3, to_field(&V_3[..2]));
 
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -104,32 +99,24 @@ pub fn compute_final_sum_share<F: PrimeField, const λ: usize, const P: usize>(z
 ///
 /// The function uses streams since stream offers a chunk method.
 fn recurse_u_or_v<'a, F: PrimeField, J, const λ: usize>(
-    u_or_v_stream: J,
+    u_or_v_iterator: J,
```
akoshelev, line 102: it is better not to encode the type of this parameter inside the name. `u_or_v` is fine imo

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -145,14 +132,15 @@ where
...
-        recurse_u_or_v::<_, _, λ_FIRST>(u_or_v_stream, &table_first),
-    );
+    let mut iterator: Box<dyn Iterator<Item = F>> = Box::new(recurse_u_or_v::<_, _, λ_FIRST>(
+        u_or_v_iterator,
+        &table_first,
+    ));
     // all following recursion except last one
     for lagrange_table in tables.iter().take(recursions_after_first - 1) {
-        stream = Box::pin(recurse_u_or_v::<_, _, λ>(stream, lagrange_table));
+        iterator = Box::new(recurse_u_or_v::<_, _, λ>(iterator, lagrange_table));
```
akoshelev, line 141: why boxing it?

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -104,32 +99,24 @@ pub fn compute_final_sum_share<F: PrimeField, const λ: usize, const P: usize>(z
...
 }
 
-pub async fn recursively_compute_final_check<
-    F: PrimeField,
-    J,
-    const λ_FIRST: usize,
-    const λ: usize,
->(
-    u_or_v_stream: J,
+pub fn recursively_compute_final_check<F: PrimeField, J, const λ_FIRST: usize, const λ: usize>(
```
akoshelev, line 113: Consider adding rustdoc explaining the generics here, especially the difference between λ_FIRST and λ

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -104,32 +99,24 @@ pub fn compute_final_sum_share<F: PrimeField, const λ: usize, const P: usize>(z
...
 /// The function uses streams since stream offers a chunk method.
 fn recurse_u_or_v<'a, F: PrimeField, J, const λ: usize>(
-    u_or_v_stream: J,
+    u_or_v_iterator: J,
     lagrange_table: &'a LagrangeTable<F, λ, 1>,
-) -> impl Stream<Item = F> + 'a
+) -> impl Iterator<Item = F> + 'a
 where
-    J: Stream<Item = F> + 'a,
+    J: Iterator<Item = F> + 'a,
```
akoshelev, line 106: generally speaking, taking `IntoInterator` makes your function more generic as it accepts more arguments. If you take `Iterator`, you have to convert vectors and arrays using `.into_iter`, but if you take `IntoIterator` you can accept all iterators as before but also vectors and other collections that implement it with no extra method calls on the callee side.

ipa-core/src/utils/arraychunks.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{ff::Fp61BitPrime, secret_sharing::SharedValue};
+use crate::ff::PrimeField;
 
 #[allow(dead_code)]
```
akoshelev, line 3: still need it?

ipa-core/src/utils/arraychunks.rs
```diff
@@ -17,17 +17,17 @@ pub struct ArrayChunk<I, const L: usize> {
     iter: I,
 }
 
-impl<I: Iterator<Item = Fp61BitPrime>> ArrayChunkIterator for I {}
+impl<F: PrimeField, I: Iterator<Item = F>> ArrayChunkIterator for I {}
 
 #[allow(clippy::while_let_on_iterator)]
-impl<I, const L: usize> Iterator for ArrayChunk<I, L>
+impl<F: PrimeField, I, const L: usize> Iterator for ArrayChunk<I, L>
```
akoshelev, line 23: Are you sure that `Iterator` is appropriate here? you never return more than one element

ipa-core/src/utils/arraychunks.rs
```diff
@@ -17,17 +17,17 @@ pub struct ArrayChunk<I, const L: usize> {
...
 where
-    I: Iterator<Item = Fp61BitPrime>,
+    I: Iterator<Item = F>,
```
akoshelev, line 25: for consistency, it is better to specify all trait bounds either in `where` clause or next to `impl`. 
In this case, I think you should just move everything to `where`

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -145,14 +132,15 @@ where
 
     // generate & evaluate recursive streams
     // to compute last array
-    let mut stream: Pin<Box<dyn Stream<Item = F> + Send>> = Box::pin(
-        recurse_u_or_v::<_, _, λ_FIRST>(u_or_v_stream, &table_first),
-    );
+    let mut iterator: Box<dyn Iterator<Item = F>> = Box::new(recurse_u_or_v::<_, _, λ_FIRST>(
```
benjaminsavage, line 135: Do you need to include this type? Can you get away with just:
```
let mut iterator = Box::new(recurse_u_or_v::<_, _, λ_FIRST>(
```

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
benjaminsavage: If you can manage to remove the Box, that would be even better.

benjaminsavage (approved):
I'm glad you made this change @danielmasny - I like how much simpler this is making the code. @akoshelev has some great suggestions. Once you make those and tests pass go ahead and merge it!

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
danielmasny: unfortunately not. The dyn trait allows any iterator. When I do not specify the type, the compiler thinks that it is a Box around the return type of recurse_u_or_v after the first recursion

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
danielmasny: the dyn type always needs a Box since the size is not known at compile time. I need a dyn type since I apply lambda times the recurse_u_or_v on the input steam, i.e. recurse_u_or_v(...recurse_u_or_v(u_or_v_iterator)...) and every intermediate output type of recurse is a different type of iterator even though they are all iterators with Items of type F.

ipa-core/src/utils/arraychunks.rs
danielmasny: I don't understand this comment. Is there a better alternative? The reason why I implement this is that I want to convert an iterator over items of type F to an iterator over items [F; L], i.e. I want to have L elements at a time. I don't know how else to do it. With implementing iterator I can have a customized next function that pulls L elements at a time from the underlying iterator of items of type F. @akoshelev please let me know your thoughts, I am happy to change it in a follow up pr

akoshelev: So I looked at the code again and you're right - this does return more than one value. And yes, opaque types require you to do the boxing. 
The recursion implementation is still a bit confusing for me. What is the goal for you here? The state machine that you build in runtime for iterators that point to other iterators on the heap can be quite large and I don't think it is cheaper/faster to do than collecting your input into a vector and then compressing it using your algorithm.


My understanding of the protocol here, which could be completely wrong, is that you have a collection of field values on your input of some size $N$. Your goal is to compress it into a vector of size no more than  λ. You do that by inspecting that collection and taking chunks of size  λ and compressing them into a single field value by evaluating a polynomial. You do it repeatedly until the size of your output is  λ. 

If that's the case, then I don't think you need `ArrayChunk` trait, boxing iterators, recursion that require boxing, etc. I think you can achieve the same result by:

* Collecting your input into a vector of size `M`.  
* Iterating over chunks of size $\lambda$ and compressing them
* repeat `iteration` times


```rust
    let mut container = u_or_v_iterator.collect::<Vec<_>>();
    for i in 0..recursions_after_first {
        let mut write_pointer = 0;
        let step = if i == 0 { λ_FIRST } else { λ };

        for j in (0..container.len()).step_by(step) {
            let end = min(j + step, container.len());
            let chunk = &container[j..end];
            let compressed_val = if i == 0 {
                table_first.eval(chunk)[0]
            } else {
                tables[i - 1].eval(chunk)[0]
            };
            container[write_pointer] = compressed_val;
            write_pointer += 1;
        }

        container.truncate(write_pointer);
    }
```

this achieves the same goal with less complexity imo and I think it is faster. I am sure this can be optimized and a few abstractions may be required to make it look nicer, but the idea stays the same. And you don't need boxing and dynamic dispatch

cc: @danielmasny @benjaminsavage

akoshelev: note that `let mut container = u_or_v_iterator.collect::<Vec<_>>();` can be avoided because you can just pass your container directly into this function to re-use the allocation. The fact that you have an iterator hints that you already have all these field values sitting somewhere in memory, so you can just re-use that

akoshelev: I looked at the code again and seems that `BatchToVerify::verify` can take ownership of `u_and_v`, so the approach above could work

danielmasny: @akoshelev Thanks for your suggestions! I think it is a memory cpu tradeoff. I think it would make sense to implement both versions and then simply benchmark them. This function should be very easy to benchmark since it is not even an async function. For the first iteration, we probably do not want to collect the u or v in a vector before compressing them since that would take around 670 Megabytes in memory for 50 million bit multiplications.

PR #1167, opened 2024-06-25
split BatchToVerify into fn and async fn

danielmasny:
The split is need such that we do not have a mutex across await points. For issue see: https://github.com/private-attribution/ipa/pull/1164


ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -48,6 +52,100 @@ impl ProofBatch {
...
 
+    /// Each helper party generates a set of proofs, which are secret-shared.
+    /// The "left" shares of these proofs are for the helper on the left.
+    /// The "right" shares of these proofs need not be transmitted over the wire, as they
+    /// are generated via `PRSS`, so the helper on the right can independently generate them.
+    /// The final proof must be "masked" with random values drawn from PRSS.
+    /// These values will be needed at verification time.
+    /// The function outputs `my_proofs_left_shares`, `shares_of_proofs_from_prover_left`,
+    /// `p_mask_from_right_prover`, `q_mask_from_left_prover`
+    pub fn generate<C, I>(ctx: &C, uv_tuple_inputs: I) -> (Self, Self, Fp61BitPrime, Fp61BitPrime)
```
akoshelev, line 63: I would recommend a struct that describes the return. I personally wouldn't use tuples greater than the size of 2

akoshelev (approved):
:shipit: but I'd love a follow up to make the return type more self-explanatory. 

Tuples of (T, T, X, X) are not great. It is easy to mismatch which T you're looking for, so it is better to be explicit

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
danielmasny: I will clean it up in a follow up PR, thanks!

PR #1168, opened 2024-06-25
split verify into fn and async fn

danielmasny:
needed since Mutex cannot be used across await points, see issue: https://github.com/private-attribution/ipa/pull/1164

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
+++ b/ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
@@ -216,17 +216,16 @@ impl BatchToVerify {
         )
     }
 
-    /// This function computes and outputs the final `p_r_right_prover * q_r_right_prover` value.
-    async fn p_and_q_r_check<C, U, V>(
+    /// This function computes and outputs `p_r_right_prover`, `q_r_left_prover`.
+    // todo: change from async fn to fn once stream is replaced
+    pub async fn compute_p_and_q_r<U, V>(
         &self,
-        ctx: &C,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/ipa_prf/validation_protocol/validation.rs
```diff
@@ -266,50 +277,44 @@ impl BatchToVerify {
...
-    pub async fn verify<C, U, V>(
+    pub async fn verify<C>(
         &self,
         ctx: &C,
         sum_of_uv_right: Fp61BitPrime,
-        u_from_right_prover: U,
-        v_from_left_prover: V,
+        p_r_right_prover: Fp61BitPrime,
+        q_r_left_prover: Fp61BitPrime,
+        challenges_for_left_prover: &[Fp61BitPrime],
```
akoshelev, line 286: you convinced me in #1173 that you need iterators here. Is it a TODO?

danielmasny: @akoshelev everything is small except the list of all u,v values. There are 50 million u,v values, whereas there are only around 8 challenges in total and 8 "sub" proofs stored in the BatchToVerify for each left and right. So vectors are no issue here.

akoshelev: ah you are changing the contract for this function, so it is no longer taking `u` and `v` but it takes challenges. 

You may consider structuring proofs so `r` values along with challenges come in one struct.

akoshelev (approved):

PR #1170, opened 2024-06-26
Skeleton function for integer multiplication

shinta-liem:
- Skeleton function for integer multiplication - To be filled in the next PR


benjaminsavage (approved):
Yup! This is all correct so far. Now let's fill in that function body!

danielmasny (commented):
@shinta-liem @benjaminsavage I think for the integer multiplication of two $L$ bit integers it might actually make sense to convert the Boolean shares into shares in 
$\mathbb Z_{2^{2\cdot L}}$, i.e. shares over the ring of integers $mod \enspace 2^{2\cdot L}$. The conversion would cost $4\cdot L$ bits of communication and the multiplication another $4\cdot L$ bits (when using MACs). It at least gives a good baseline to compare the efficiency of the integer multiplication.

PR #1172, opened 2024-06-26
[WIP] Breakdown Reveal Aggregation

cberkhoff:
This PR is a work in progress to enable Breakdown Reveal Aggregation (aka Improved Aggregation). Aggregation steps happen after attribution. The input to aggregation is a stream of tuples of (attributed breakdown key, attributed trigger value). The output of aggregation is a Histogram. Breakdown Keys and Trigger Values are assigned by the advertiser and sent in the input of IPA. Breakdown Keys values are expected to be dense. How breakdown keys and trigger values are defined is out-of-scope.

High level explanation of the protocol:
1. Add fake attribution outputs.
2. Shuffle.
3. Reveal Breakdown Keys. By having shuffled and adding fake entries we protected the identities of individuals. Trigger values are not revealed.
4. Aggregation of Trigger Value by Breakdown Key (Think of group by).

DP noise to histogram buckets will be added by the caller of this function. This because adding noise is really unrelated to how things are added up.

Following are the improvements to be done before merging in no particular order:
* TODO: Use sharded shuffle.
* TODO: For aggregation, we pad TV with zeroes until size of HV and add them together. Expanding bits only as needed.
* TODO: seq_join (see quicksort). First reveal, create futures for the remainder.
* TODO: Add vectorization + chunks. Change TotalRecords::Indeterminate. "Merge-sort".
* TODO: Add sharding.


ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+    let contribs = shuffle_attribution_outputs::<_, BK, TV, BA64>(shuffle_ctx, contribs).await?;
+    // at what level do we want to start the paralelization
+    let hv_size: usize = HV::BITS.try_into().unwrap();
+    let mut result = repeat_n(BitDecomposed::new(repeat_n(Replicated::ZERO, hv_size)), B)
+        .collect::<Vec<BitDecomposed<Replicated<Boolean>>>>();
+    // for loop will be replaced with vectorized
+    for (i, attribution_outputs) in contribs.into_iter().enumerate() {
+        let record_id = RecordId::from(i);
+        let ao: SecretSharedAttributionOutputs<BK, TV> = attribution_outputs; // For Rust Analyzer
+        let bk_share = ao.attributed_breakdown_key_bits;
```
danielmasny, line 74: why is ao needed here and why not using attribution_outputs directly?

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+        let reveal_ctx = ctx
+            .narrow(&AggregationStep::RevealStep)
+            .set_total_records(TotalRecords::Indeterminate);
```
danielmasny, line 77: reveal context needs to be defined outside of the loop, otherwise you generate a context for every record.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+        let revealed_bk: BK = BK::from_array(&bk_share.reveal(reveal_ctx, record_id).await?);
+        let pos = usize::try_from(revealed_bk.as_u128())?;
+        //tracing::info!("revealed_bk={pos}, tv={:?}", ao.capped_attributed_trigger_value);
+        let tv: BitDecomposed<Replicated<Boolean>> = ao.capped_attributed_trigger_value.to_bits();
+        let add_ctx = ctx
```
danielmasny, line 82: same, here. Define the context outside of the loop.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+            .narrow(&AggregationStep::Add)
+            .set_total_records(TotalRecords::Indeterminate);
+        let r =
+            integer_sat_add::<_, SixteenBitStep, 1>(add_ctx, record_id, &result[pos], &tv).await?;
+        result[pos] = r;
```
danielmasny, line 87: I think not introducing a new variable r (with an ambiguous naming) would be better. I.e. let result[pos] = integer_sat_add.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+    let resp: Vec<Replicated<HV>> = result
+        .into_iter()
+        .map(|b: BitDecomposed<Replicated<Boolean>>| b.collect_bits())
+        .collect();
```
danielmasny, line 92: Do we need another iteration here? Couldn't we call collect bits before storing the outcome of integer_sat_add in result[pos].

ipa-core/src/error.rs
```diff
@@ -72,6 +72,8 @@ pub enum Error {
     DecompressingInvalidCurvePoint(String),
     #[error(transparent)]
     LengthError(#[from] LengthError),
+    #[error(transparent)]
+    TryFromIntError(#[from] std::num::TryFromIntError),
```
danielmasny, line 76: This error doesn't seemed to be used anywhere.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+            let result = result.iter().map(|&v| v.as_u128()).collect::<Vec<_>>();
+            tracing::info!("result={:?}", result);
+            assert_eq!(32, result.len());
+            assert_eq!(result[0], 0);
+            assert_eq!(result[1], 3);
+            assert_eq!(result[3], 8);
+            assert_eq!(result[4], 3);
+            assert_eq!(result[10], 5);
+            assert_eq!(result[11], 4);
+            assert_eq!(result[22], 5);
```
danielmasny, line 159: Maybe test for all positions, including 2, 5 and rather use less breakdowns in the test?

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
```diff
@@ -51,6 +51,7 @@ impl From<MoveToBucketError> for Error {
 ///
 /// ## Errors
 /// If `breakdown_count` does not fit into `BK` bits or greater than or equal to $2^9$
+#[allow(dead_code)]
```
danielmasny, line 54: Why do we need this in this PR?

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -864,10 +880,10 @@ pub mod tests {
         }
     }
 
-    #[derive(Debug, PartialEq)]
-    struct PreAggregationTestOutputInDecimal {
-        attributed_breakdown_key: u128,
-        capped_attributed_trigger_value: u128,
+    #[derive(Debug, Clone, Ord, PartialEq, PartialOrd, Eq)]
+    pub struct PreAggregationTestOutputInDecimal {
```
danielmasny, line 884: It is still stored in a u128. I am not sure whether I would call it a decimal and whether this is substantially different from the way other numbers are stored.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -111,17 +145,90 @@ where
     }
 }
 
+// This function converts Attribution Outputs to an AdditiveShare needed for shuffle protocol
+pub fn attribution_outputs_to_shuffle_input<BK, TV, YS>(
```
danielmasny, line 149: I think the naming is confusing and I do think it makes more sense to implement shared value (which implies additiveshare<sharedvalue>) for SecretSharedAttributionOutputs rather than merging BK and TV into a single Boolean Array. We might not even implement BooleanArrays that have the correct size needed.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -43,6 +50,33 @@ where
         .collect::<Vec<_>>())
 }
 
+#[tracing::instrument(name = "shuffle_attribution_outputs", skip_all)]
+pub async fn shuffle_attribution_outputs<C, BK, TV, R>(
```
danielmasny, line 54: If we don't try to merge BA just for the sake of shuffling (see comment below on attribution_outputs_to_shuffle) We do not need the generic R and the trait bounds. I don't think we should introduce new BA types just for the sake of shuffling.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -111,17 +145,90 @@ where
...
+    expand_shared_array_in_place(
+        &mut y,
+        &input.capped_attributed_trigger_value,
+        BK::BITS as usize,
+    );
+    y
+}
+
+// This function converts Attribution Outputs  obtained from shuffle protocol to OprfReport
+pub fn shuffled_to_attribution_outputs<YS, BK, TV>(
```
danielmasny, line 168: This could also be removed by implementing shared value for SecretSharedAttributionOutputs

danielmasny (changes requested):
I have left a couple of comments. I think we need to think more about a strategy of how to vectorize the aggregation efficiently.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
cberkhoff: Just a small aid for Rust Analyzer. Rust can't infer the type.

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
cberkhoff: Because I haven't deleted the old aggregation code yet

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+    contributions_stream: St,
+) -> Result<Vec<Replicated<HV>>, Error>
+where
+    St: Stream<Item = Result<SecretSharedAttributionOutputs<BK, TV>, Error>> + Send,
+    BK: BreakdownKey<B>,
+    TV: BooleanArray + U128Conversions,
+    HV: BooleanArray + U128Conversions,
+{
+    // TODO: Maybe move this to the function contract
+    let contribs: Vec<_> = contributions_stream.try_collect().await?;
```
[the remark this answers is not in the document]
cberkhoff: I'll change it on a future revision. Please consider this code just a starting point to better discuss the vectorization part.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
cberkhoff: Will do. The for loop is going away anyway.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,162 @@
...
+// record.
+// The output is a Histogram. BKs are assigned by the advertiser and sent in
+// the input of IPA. BK values are expected to be dense.
+// How breakdown keys are defined is out-of-scope.
+//
+// High level explanation of the protocol:
+// 1. TODO: Add fake attribution outputs.
+// 2. Shuffle.
+// 3. Reveal breakdown. Trigger values are not revelaed.
+// 4. Aggregation if trigger values secret shares.
```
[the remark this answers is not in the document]
cberkhoff: Yeah, I still need to come up with a way to chunk/vectorize things.

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
cberkhoff: I didn't came up with that name. Rather have a separate PR for renaming things if you deem that necessary.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
cberkhoff: This entire part of the shuffling is going away anyway in favor of the sharded version which has a Shuffle-able trait

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
cberkhoff: Will keep this in mind but this whole part of shuffling is going away

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
cberkhoff: Yeah, I don't think these conversions to BA are necessary, in fact the sharded version of shuffle provides better ergonomics. I'm just using this version as a temporary stop-gap.

ipa-core/src/error.rs
cberkhoff: It's used in breakdown_reveal_aggregation

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
cberkhoff: Why?

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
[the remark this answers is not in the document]
akoshelev: don't we know the breakdown in the clear, so we can trivially group them as we reveal it?

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
cberkhoff: Sure. I used that to help me sort out some types while developing but it can go away in the final code.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
cberkhoff: If we do that, we would need to keep decomposing into bits the result on every iteration. This way we only do it once at the end.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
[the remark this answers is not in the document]
danielmasny: yes its all local so you could sort by breakdown and then vectorize within events with the same breakdown. Sorting a large list might still cause a lot of reallocations which might not be necessary.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
danielmasny: I see, makes sense!

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
danielmasny: Since you are allocating space for them in the vector, we also want to check that they are correct, i.e. zero.

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
danielmasny: I see so this is just a temporary change. Maybe mark it with a todo?

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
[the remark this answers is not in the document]
andyleiserson: Yeah, for the breakdown counts that we're contemplating currently, I would say just fan the values out into a bucket per breakdown. Then once we get two chunks worth of records in a bucket, we can do a vectorized reduction.

If we're doing a full resharding here (which a sharded shuffle is), then maybe it makes sense to shard by breakdowns rather than randomly? There is a question whether we get better parallelism across breakdowns or by eagerly pulling more chunks from a particular aggregation stream.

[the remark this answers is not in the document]
akoshelev: > Yes its all local so you could sort by breakdown
I still don't see a need to sort a big list by breakdown key. As Andy pointed out, we fan out events to the corresponding shards by the breakdown key as we drain the input. Grouping comes essentially for free

> If we're doing a full resharding here (which a sharded shuffle is), then maybe it makes sense to shard by breakdowns rather than randomly? 

This makes sense to me - maybe in sharded world in this case we can avoid transpose 

> There is a question whether we get better parallelism across breakdowns or by eagerly pulling more chunks from a particular aggregation stream.

To make a decision here, it would be helpful to understand the distribution of events per breakdown. If it is not close to uniform in presence of fake events (basically follows user-provided distribution), then we can assume the worst case and vectorize across breakdowns per shard

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
akoshelev: yea that's pretty gross, but as @cberkhoff pointed out a necessary change due to shuffle not really supporting any other input except IPA event

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -285,7 +282,32 @@ pub struct AttributionOutputs<BK, TV> {
...
+    TV: BooleanArray + crate::secret_sharing::IntoShares<Replicated<TV>>,
+{
+    fn share_with<R: rand::Rng>(self, rng: &mut R) -> [(Replicated<BK>, Replicated<TV>); 3] {
+        let bk_sh = self.bk.share_with(rng);
+        let tv_sh = self.tv.share_with(rng);
+        [
+            (bk_sh[0].clone(), tv_sh[0].clone()),
+            (bk_sh[1].clone(), tv_sh[1].clone()),
+            (bk_sh[2].clone(), tv_sh[2].clone()),
+        ]
```
akoshelev, line 308:
```suggestion
        let [bk_0, bk_1, bk_2] = self.bk.share_with(rng);
        let [tv_0, tv_1, tv_2] = self.tv.share_with(rng);
        [
            (bk_0, tv_0),
             ...
        ]
```

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
akoshelev: this may be a good suggestion to keep the code concise for the time being, if that works. I much rather prefer traits over standalone functions to keep the nomenclature of things used somewhat under control

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
akoshelev: this made it public which made me questioning whether it is only used in tests and should be marked as such?

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -157,4 +264,43 @@ pub mod tests {
...
+    #[test]
+    fn test_shuffle_attribution_outputs() {
+        const BATCHSIZE: usize = 50;
+        run(|| async {
+            let world = TestWorld::default();
+            let mut rng = rand::thread_rng();
+            let mut inputs = Vec::new();
+            let mut expectation = Vec::new();
+            for _ in 0..BATCHSIZE {
+                let (i, e) =
```
akoshelev, line 277: BA32 can be generated using Rng, so you can simplify this part as well

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -111,17 +145,90 @@ where
...
+                },
+                shuffle::{shuffle_attribution_outputs, shuffle_inputs},
+            },
+        },
+        sharding::NotSharded,
         test_executor::run,
         test_fixture::{ipa::TestRawDataRecord, Reconstruct, Runner, TestWorld},
     };
 
+    fn input_row(
```
akoshelev, line 211: a bit of bikeshedding, but this does generate an input row and expected result, so name can reflect that

ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -157,4 +264,43 @@ pub mod tests {
...
+                let (i, e) =
+                    input_row(rng.gen_range(0u128..1 << 32), rng.gen_range(0u128..1 << 32));
+                inputs.push(i);
+                expectation.push(e);
+            }
+            let mut result: Vec<PreAggregationTestOutputInDecimal> = world
+                .upgraded_semi_honest(
+                    inputs.into_iter(),
+                    |ctx: UpgradedSemiHonestContext<NotSharded, Boolean>, input_rows| async move {
+                        let aos: Vec<_> = input_rows
```
akoshelev, line 286: this variable could have a better name

ipa-core/src/protocol/ipa_prf/aggregation/bucket.rs
cberkhoff: I will delete all the old-aggregation code before I merge but I would like to keep it for reference (easier to navigate code with IDE than Github). I allow the code so that all code checks still run on every push.

ipa-core/src/error.rs
cberkhoff: Since we have length error I though it made sense, but I agree it's too generic.

cberkhoff:
This change is still a WIP. Just want to get some baseline numbers from Draft.

cberkhoff:
I ran a queries and got the following [stas](https://docs.google.com/spreadsheets/d/1JJDNCRiE7fNQ7HcTGRjbQau9vEvznfjbgGOPFjkV9TI/edit?usp=sharing):

  | Old | New |  
-- | -- | -- | --
Runtime [m] | 9.583333333 | 30.0 | 313.48%
Aggregation / Total Runtime | 63.9% | 93.59% |  
Shuffle |   | 0.41% |  
Reveal |   | 4.92% |  
Additions |   | 83.05% |  

Having this in mind I focus on Additions and then Reveal

cberkhoff:
With the `seq_join` in `add_tvs_by_bk` performance improved by ~6x and now the total runtime is the same between old and new:

  | Old | New
-- | -- | --
Runtime [m] | 9.583333333 | 9.5
Aggregation time / Total Runtime | 63.9% | 79.16%
  |   |  
Attribution |   | 23.02%
Shuffle |   | 1.56%
Reveal |   | 18.44%
Additions |   | 56.98%

cberkhoff:
Using `seq_join` in `reveal_breakdowns` made it 14x faster. The new aggregation is 18% faster than the old one.

cberkhoff:
Added a new aggregation logic for Trigger Values into the Histogram. With it, aggregation as a whole is about 3x faster than the current one.

Results running IPA for 1M events |   |   |  
-- | -- | -- | --
  | Old [duration] | New [duration] | Improvement
Total Runtime | 09:35 | 03:57 | 242.62%
Aggregation time | 06:08 | 02:06 | 292.06%
  |   |   |  
  |   | Duration | % of aggregation
Attribution | - | 01:30 | 71.43%
Shuffle | - | 00:07 | 5.56%
Reveal | - | 00:05 | 3.97%
Additions | - | 00:23 | 18.25%

cberkhoff:
Digging trough the logs I think vectorizing can make aggregation take 20seconds off aggregation, which is taking around 126 seconds. There might be more gains from more efficient code.

[Spreadsheet](https://docs.google.com/spreadsheets/d/1JJDNCRiE7fNQ7HcTGRjbQau9vEvznfjbgGOPFjkV9TI/edit?usp=sharing)

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
cberkhoff: It's configured to be a test struct only with `#[cfg]`

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,423 @@
...
+            return Ok(r);
+        }
+
+        let can_expand = grouped_tvs.expand();
+
+        let add_ctx = parent_ctx
+            .narrow(&AggregationStep::Aggregate(d))
+            .set_total_records(TotalRecords::specified(pairs.len()).unwrap());
+
+        let work = pairs.iter().enumerate().map(|(i, p)| {
```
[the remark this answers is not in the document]
akoshelev: also, you could do `zip(pairs, repeat(add_ctx)).map(...)` to get a fresh context for every row

akoshelev (commented):
This is mostly good to go I think. There are some things to improve, but most importantly we shouldn't replace the existing aggregation just yet. We haven't added DP noise in OPRF and we may need a proof or tight bound on noise added per breakdown key.

Once these figured, we can replace the aggregation in a separate PR

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,423 @@
...
+/// BKs is small, there's not a lot of gain by doing operations in place with
+/// references in this structure.
+struct GroupedTriggerValues<HV, const B: usize> {
+    singles: [Option<Operand>; B],
+    pairs: Vec<Pair>,
+    size: usize,
+    phantom: PhantomData<HV>,
+}
+
+impl<HV, const B: usize> GroupedTriggerValues<HV, B>
```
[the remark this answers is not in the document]
cberkhoff: I kind of agree. I'm planning on a different but related struct for the vectorized version (I still need to think a bit more about it though). Even if we have enable vectorization for larger sums, the smaller sums, will still benefit from having this version around.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,423 @@
...
+                if let Some(hv_bits) = s {
+                    r.push(hv_bits.collect_bits());
+                } else {
+                    r.push(Replicated::ZERO);
+                }
+            }
+            return Ok(r);
+        }
+
+        let can_expand = grouped_tvs.expand();
```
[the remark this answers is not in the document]
cberkhoff: We're adding 0's to match the results of the non-saturated additions. The actual sum happens after. 

I could be wrong, but adding these zeroes is "local", there's no communication required. Each helper just adds a Replicated::ZERO to each item in the singles. This should be very fast.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,423 @@
...
+async fn add_tvs_by_bk<TV, HV, const B: usize>(
+    parent_ctx: &UpgradedSemiHonestContext<'_, NotSharded, Boolean>,
+    mut grouped_tvs: GroupedTriggerValues<HV, B>,
+) -> Result<Vec<Replicated<HV>>, Error>
+where
+    TV: BooleanArray,
+    HV: BooleanArray,
+{
+    for d in 0..MAX_DEPTH {
+        let pairs = mem::take(&mut grouped_tvs.pairs);
```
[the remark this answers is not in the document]
cberkhoff: How can we take advantage of this?

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -0,0 +1,423 @@
...
+            let Ok(bk) = usize::try_from(revealed_bk.as_u128()) else {
+                return Err(Error::Internal);
+            };
+            Ok::<(usize, Replicated<TV>), Error>((bk, ao.capped_attributed_trigger_value))
+        }
+    });
+    let tv_size: usize = TV::BITS.try_into().unwrap();
+    let mut grouped_tvs = GroupedTriggerValues::<HV, B>::new(tv_size);
+    let tvs: Vec<(usize, Replicated<TV>)> = seq_join(reveal_ctx.active_work(), reveal_work)
+        .try_collect()
```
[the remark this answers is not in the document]
cberkhoff: the reason I was doing it is because I didn't wanted to modify `grouped_tvs` in an async scope. I wanted all the modifications of `grouped_tvs` to happen in the main thread. If I do something like:
```
seq_join(reveal_ctx.active_work(), reveal_work)
        .try_for_each(move |(bk, tv)| {
            grouped_tvs.push(bk, tv.to_bits())
        })
        .await?;
```
I'm getting an issue with captured variables escaping the closure body

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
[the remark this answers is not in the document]
cberkhoff: you meant `into_iter` ?

with the existing `enumerate` the resulting tuple will look weird: `(i, (p, add_ctx))`.
I find the current code more legible.

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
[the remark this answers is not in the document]
akoshelev: it works if you pin the stream and try_next it

akoshelev (approved):
This is good starting point to keep improving the new aggregation. We should create an issue with all the improvements and action items we need to take before we can make it the default

PR #1173, opened 2024-06-27
Remove iterators and boxing from `recursively_compute_final_check`

akoshelev:
As we discussed in #1166, the way iterators are boxed to achieve the desired result of recursively exhaust them and build the final result made code look more complex that otherwise it should be.

This is a proposal to simplify this implementation by:

* re-using the `u` and `v` allocations for compression algorithm
* using a more straightforward `chunks` method that is available on vectors
* removing custom `arraychunks` implementation
* Simplifying trait bounds for functions in this module


ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -166,17 +173,32 @@ where
...
-                compute_g_differences, compute_sum_share, interpolate_at_r, recurse_u_or_v,
+                compute_g_differences, compute_sum_share, interpolate_at_r,
                 recursively_compute_final_check,
             },
         },
         secret_sharing::SharedValue,
     };
 
+    /// This function compresses the `u_or_v` values and returns the next `u_or_v` values.
+    fn recurse_u_or_v<F: PrimeField, const λ: usize>(
```
akoshelev, line 188: this was moved inside the test module just to avoid changing unit tests, just to avoid making too big of a change

ipa-core/src/protocol/ipa_prf/malicious_security/verifier.rs
```diff
@@ -95,32 +97,38 @@ pub fn compute_final_sum_share<F: PrimeField, const λ: usize, const P: usize>(z
...
     challenges: &[F],
     p_or_q_0: F,
-) -> F
-where
-    J: Iterator<Item = F>,
-{
+) -> F {
     let recursions_after_first = challenges.len() - 1;
 
     // compute Lagrange tables
```
akoshelev, line 134: this probably deserve some changes in separate PR. We don't seem to save anything by collecting all these tables in advance

danielmasny (changes requested):
Thanks for your suggestions! I think it is a memory cpu tradeoff. I think it would make sense to implement both versions and then simply benchmark them. This function should be very easy to benchmark since it is not even an async function.

For the first iteration, we probably do not want to collect the u or v in a vector before compressing them since that would take around 670 Megabytes in memory for 50 million bit multiplications. The current changes would make it necessary to collect the initial iterator over u,v values in a vector.

I would prefer not to delete the existing code but just offer the alternative and then benchmark both for 50 million u,v values.

akoshelev:
> For the first iteration, we probably do not want to collect the u or v in a vector before compressing them since that would take around 670 Megabytes in memory for 50 million bit multiplications. The current changes would make it necessary to collect the initial iterator over u,v values in a vector.

This change does not require collecting iterators into vectors, it simply re-uses the existing allocation for `u` and `v` values

akoshelev:
I synced with @danielmasny and he pointed me to the new PR that actually changes how things are passed into this function. It won't be a vec anymore, but an iterator that yields field values for all bits stored in a batch. So it is definitely suboptimal to collect all 50M field values into a vec.

I don't think this PR makes sense

PR #1174, opened 2024-06-28
[WIP] Integer multiplication implementation

shinta-liem:
Basic grade-school algorithm for integer multiplication for x and y where x is a positive unsigned number, and y is a signed number in two's complement.

This is a WIP:
- Use the correct step and context
- Optimize the algorithm -- calculate the interim products and add it up in parallel
- Explore more efficient algorithm


danielmasny (commented):
Hi Shinta,

I just wrote down how you could do the multiplication using share conversion: https://docs.google.com/document/d/1vCjea2ZZ7dB30YGE2rt-fYxG26du1_lvsUkeJVHHJuk/edit. It seems cheaper, at least for semi honest security though it might require more changes in the infrastructure (e.g. to support shares over the integer or the specific prime field that is large enough to hold the output).

benjaminsavage:
@danielmasny - thanks for sharing that write-up! I think we should definitely try both approaches (and I have a few more in mind as well). The tradeoff here is communication cost vs the simplicity of remaining in a boolean field.

I do not expect the schoolbook multiplication algorithm to be the most efficient. I mainly view this as a simple onboarding task for @shinta-liem - to help familiarise her with the Rust codebase. We can implement this alternative approach as well in a follow-up PR and benchmark them against one another.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
-        },
+        boolean::NBitStep, context::Context,
+        ipa_prf::boolean_ops::addition_sequential::integer_add, BooleanProtocols, RecordId,
     },
     secret_sharing::{replicated::semi_honest::AdditiveShare, BitDecomposed, FieldSimd},
 };
 
-
+/// This function multiplies x by y in these steps:
+/// 1. Double the input precision and repeat the most significant bit in the extra bits (Sign extension)
```
benjaminsavage, line 13: This is only for `y` right?

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
 pub async fn integer_mul<C, S, const N: usize>(
     ctx: C,
     record_id: RecordId,
     x: &BitDecomposed<AdditiveShare<Boolean, N>>,
     y: &BitDecomposed<AdditiveShare<Boolean, N>>,
-) -> Result<
-    BitDecomposed<AdditiveShare<Boolean, N>>,
-    Error,
->
+) -> Result<BitDecomposed<AdditiveShare<Boolean, N>>, Error>
```
benjaminsavage, line 24: run `cargo fmt` at the command line to get the auto-formatter to run. Some of the failing checks will be resolved this way.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+    let new_len = x.len() + y.len();
+    let mut x = x.clone();
+    x.resize(new_len, AdditiveShare::ZERO);
+    let mut y = y.clone();
+    y.resize(new_len, y[y.len() - 1].clone());
+
+    let mut result = BitDecomposed::with_capacity(new_len);
+    for (i, yb) in y.into_iter().enumerate() {
+        let mut t = BitDecomposed::with_capacity(new_len);
+        t.resize(i, AdditiveShare::ZERO);
```
benjaminsavage, line 40: Either in this PR (or if you would prefer, in a follow up), we should avoid doing this, as it's needlessly increasing the amount of work in the addition stage, by adding shares of zero. Instead, we should just make `t` of length `new_len - i` in each step, and make sure to add it up correctly.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+        // TODO fix the context with proper steps
+        let ctx_for_bit_of_y = ctx.narrow(&S::from(i));
+        for (j, xb) in x.iter().take(new_len - i).enumerate() {
+            let ctx_for_x_times_y_combo = ctx_for_bit_of_y.narrow(&S::from(j));
+
+            let m = yb.multiply(xb, ctx_for_x_times_y_combo, record_id).await?;
+            t.push(m);
+        }
```
benjaminsavage, line 48: In this PR, let's at least perform this part in parallel. Go look around in the codebase for examples of `.parallel_join(`.

It should look something like this:
```
let t = ctx_for_bit_of_y.parallel_join(
x.iter().take(new_len - i).enumerate().map(|(j, xb)| {
    let ctx_for_x_times_y_combo = ctx_for_bit_of_y.narrow(&S::from(j));
    async move {
        yb.multiply(xb, ctx_for_x_times_y_combo, record_id).await
    }
})).await?;
```

You'll likely encounter some issues with stuff inside the `async move` closure. There are many restrictions on referencing variables outside of that scope. If you get stuck, feel free to ping me.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+                .await
+                .reconstruct();
+
+            for ((res, x), y) in result
+                .iter()
+                .zip(all_x_values.iter())
+                .zip(random_y_values.iter())
+            {
+                let y_as_signed_number: i128 = as_i128(*y);
+                let x_as_unsigned_number: i128 = x.as_u128() as i128;
```
benjaminsavage, line 153: This variable name seems to not match the implementation.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+                let expected: i128 = y_as_signed_number * x_as_unsigned_number;
+
+                assert_eq!(
+                    (x, y, y_as_signed_number, x_as_unsigned_number, expected),
+                    // (x, y, y_as_unsigned_number, x_as_signed_number, res.as_u128())
```
benjaminsavage, line 158: remove this comment.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+                    (x, y, y_as_signed_number, x_as_unsigned_number, as_i128(*res))
```
benjaminsavage, line 159: Do we really need x, y, y_as_signed_number and x_as_unsigned_number? Was this just for debugging? Maybe we can just use x and y in a "signed" format so that the error message is easy to read.

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
shinta-liem: is there a method for left shifting the bits?
i.e.
```
       1101    13
       1011    -5
----------
       1101
     1101      <-- shift left 1 bit
    0000     <-- shift 2 bits
 1101          <-- shift 3 bits
```

right now t.resize is just appending 0s to shift each intermediate products to the left

martinthomson: You should be able to create a new BitDecomposed by calling into_iter() on it, chaining that with repeat(ZERO).take(n) in whatever order your shift requires.  That might not be the best option, of course.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,158 @@
...
+            .parallel_join(x.iter().take(new_len - i).enumerate().map(|(j, xb)| {
+                let ctx_for_x_times_y_combo = ctx_for_bit_of_y.narrow(&S::from(j));
+                let yb = yb.clone();
+                async move { yb.multiply(xb, ctx_for_x_times_y_combo, record_id).await }
+            }))
+            .await?;
+
+        let mut t = BitDecomposed::with_capacity(new_len);
+        // |x| * y(i) << i
+        t.resize(i, AdditiveShare::ZERO);
```
benjaminsavage, line 51: This will incur additional cost at the aggregation phase, since it will require more carries than necessary.

Instead of this, we can just leave the least significant `i` bits of `result` alone (we know they will be unchanged), and just invoke `integer_add` on the `new_len - i` most significant bits of `result` and `t`.

benjaminsavage (changes requested):

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+            }))
+            .await?;
+
+        let t = BitDecomposed::try_from(product_of_x_and_yb).unwrap();
+        if i == 0 {
+            result = t;
+        } else {
+            // add up bits i.. with the product
+            let mut add_y = BitDecomposed::with_capacity(new_len);
+            result.iter().skip(i).for_each(|b| add_y.push(b.clone()));
```
benjaminsavage, line 57: Can you just create this in one line?

```
let add_y = BitDecomposed::new(result.iter().skip(i));
```

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+                &t,
+                &add_y,
+            )
+            .await?;
+
+            result.truncate(i); // replace the bits on i.. with the one we add up
+            add_result.iter().for_each(|b| result.push(b.clone()));
+            if result.len() < new_len {
+                // if carry bit is more than max length, we let it overflow
+                result.push(carry);
```
benjaminsavage, line 70: I am pretty sure that this line will never be called. Looking at the Google sheet, I think it's the case that result has the maximum length from the very first line.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,164 @@
...
+            let (add_result, carry) = integer_add::<_, S, N>(
+                ctx_for_bit_of_y.narrow::<Step>(&Step::Add),
+                record_id,
+                &t,
+                &add_y,
+            )
+            .await?;
+
+            result.truncate(i); // replace the bits on i.. with the one we add up
+            add_result.iter().for_each(|b| result.push(b.clone()));
```
benjaminsavage, line 67: I'm not sure which is more efficient. This, or something like this:

```
result = BitDecomposed::new(result[..i].chain(add_result));
```

ipa-core/src/protocol/ipa_prf/boolean_ops/step.rs
```diff
@@ -29,3 +29,11 @@ pub(crate) enum Fp25519ConversionStep {
     #[step(count = 256)]
     RevealY(usize),
 }
+
+#[derive(CompactStep)]
+pub(crate) enum MultiplicationStep {
+    #[step(child = crate::protocol::boolean::step::SixteenBitStep)]
+    MultiplyBits,
```
benjaminsavage, line 36: I think this is unused.

benjaminsavage (changes requested):
Very close to ready. Just a few more comments.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
shinta-liem: the intermediate `result` won't have the green part in the Google sheet. I added this because I don't extend x's size anymore (it'll always be padded with 0s). But in that case we have to consider the carry bit from `integer_add` when we add the intermediate products, otherwise there are cases where  it'll drop the carry and err by 1 bit.

ipa-core/src/protocol/ipa_prf/boolean_ops/step.rs
shinta-liem: I think I'm missing something: 

```
the trait `ipa_step::StepNarrow<protocol::ipa_prf::boolean_ops::step::MultiplicationStep>` is not implemented for `protocol::step::ProtocolGate`
```

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
```diff
@@ -1,33 +1,162 @@
...
+        basics::mul::SecureMul, boolean::NBitStep, context::Context,
+        ipa_prf::boolean_ops::addition_sequential::integer_add, BooleanProtocols, Gate, RecordId,
     },
     secret_sharing::{replicated::semi_honest::AdditiveShare, BitDecomposed, FieldSimd},
 };
 
-
+/// This function multiplies x by y in these steps:
+/// 1. Double the input precision for y and repeat the most significant bit in the extra bits (Sign extension)
+///    X is assumed to be a positive number, so we will automatically pad with ZERO.
```
benjaminsavage, line 15: This comment seems out of date now.

ipa-core/src/protocol/ipa_prf/boolean_ops/multiplication.rs
benjaminsavage: Very cool. Makes sense!

benjaminsavage (approved):
Once the checks all pass, I'll merge it!

benjaminsavage (approved):

PR #1177, opened 2024-07-01
Simplify some things by implementing Default for more types

andyleiserson:


danielmasny (approved):
Looks awesome. It actually simplifies a lot of code.

ipa-core/src/ff/prime_field.rs
```diff
@@ -231,6 +231,12 @@ macro_rules! field_impl {
             }
         }
 
+        impl Default for $field {
+            fn default() -> Self {
+                Self::ZERO
+            }
+        }
```
martinthomson, line 238: We might benefit from a 

```rust
for F in [Fp25519, Fp31, ...] {
  #[test] fn default_is_zero() {
    assert_eq!(F::ZERO, F::default());
  }
}
```

Noting of course that you can't iterate over types like that to produce test functions...

ipa-core/src/helpers/stream/chunks.rs
```diff
@@ -371,64 +363,41 @@ impl<const N: usize, T> ChunkBuffer<N> for Vec<T> {
...
     F: Fn(usize, B::Chunk) -> Fut,
     Fut: Future<Output = Result<K, Error>>,
-    D: Fn() -> T,
 {
     #[pin]
     stream: St,
     buffer: B,
     process_fn: F,
-    pad_record_fn: Option<D>,
+    terminated: bool,
```
martinthomson, line 377: If you make `stream: Fuse<St>`, then that will track this state for you and you can call `[is_terminated](https://docs.rs/futures/latest/futures/stream/trait.FusedStream.html#tymethod.is_terminated)` on that.

ipa-core/src/helpers/stream/chunks.rs
```diff
@@ -485,24 +455,23 @@ where
...
 where
     St: Stream<Item = Result<T, Error>> + Send,
+    T: Default,
     B: ChunkBuffer<N, Item = T>,
     F: Fn(usize, B::Chunk) -> Fut,
     Fut: Future<Output = Result<K, Error>>,
-    D: Fn() -> T,
 {
     StreamChunkProcessor {
         stream,
```
martinthomson, line 471: for above:

```suggestion
        stream: stream.fuse(),
```

martinthomson (commented):
I agree with Daniel.  A question about the FusedStream piece though.

ipa-core/src/helpers/stream/chunks.rs
andyleiserson: There is a comment about this on lines 396-397:

> We must implement our own fuse (rather than use `.fuse()` on the inner stream) for the case where we terminate early due to an error.

Using the inner stream as a fuse only works if we poll it until it terminates.

PR #1180, opened 2024-07-02
change convert to fp25519 function

danielmasny:
Upgrades the conversion within the PRF evaluation to compatibility with DZKP malicious security

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
@@ -133,15 +133,20 @@ where
         "conversion chunk should be a multiple of PRF chunk"
     );
 
-    assert!(input_shares.len() <= record_id_range.len());
+    assert!(
+        input_shares.len() <= record_id_range.len(),
+        "received {} inputs but only {} record IDs",
+        input_shares.len(),
+        record_id_range.len()
+    );
 
     let Some(first_input) = input_shares.first() else {
         return Ok(vec![]);
     };
 
-    // Ensure that the probability of leaking information is less than 1/(2^128).
-    // If we make match key size configurable at runtime, then we need to revisit
-    // whether a debug assertion is appropriate here.
+    // Ensure that the probability of leaking information is less than 1/(2^128). If we make match
+    // key size configurable at runtime, then the debug assertion needs to be changed to a runtime
+    // check.
     const_assert!(
         MatchKey::BITS == 64,
         "Match key size should be known at compile time"
--- a/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
+++ b/ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
@@ -97,21 +97,23 @@ use crate::{
 /// # Panics
 /// If values processed by this function is smaller than 256 bits.
 /// If vectorization is too large, i.e. `NC>=100k`.
-pub async fn convert_to_fp25519<C, const NC: usize, const NP: usize>(
-    ctx: C,
-    record_id: RecordId,
-    x: BitDecomposed<AdditiveShare<Boolean, NC>>,
+pub async fn convert_to_fp25519<V, B, const NC: usize, const NP: usize>(
+    dzkp_validator: V,
+    ctx_counter: usize,
+    first_record_id: RecordId,
+    input_shares: Vec<BitDecomposed<AdditiveShare<Boolean, NC>>>,
 ) -> Result<Vec<AdditiveShare<Fp25519, NP>>, Error>
 where
-    C: UpgradableContext,
+    V: DZKPValidator<B>,
+    B: UpgradableContext,
     Fp25519: Vectorizable<NP>,
     Boolean: FieldSimd<NC>,
     BitDecomposed<AdditiveShare<Boolean, NC>>: FromPrss<usize>,
-    AdditiveShare<Boolean, NC>: BooleanProtocols<<C as UpgradableContext>::DZKPUpgradedContext, NC>,
+    AdditiveShare<Boolean, NC>: BooleanProtocols<<B as UpgradableContext>::DZKPUpgradedContext, NC>,
     Vec<AdditiveShare<BA256>>: for<'a> TransposeFrom<&'a BitDecomposed<AdditiveShare<Boolean, NC>>>,
     Vec<BA256>:
         for<'a> TransposeFrom<&'a [<Boolean as Vectorizable<NC>>::Array; 256], Error = Infallible>,
-    <C as UpgradableContext>::DZKPValidator: Send + Sync,
+    <B as UpgradableContext>::DZKPValidator: Send + Sync,
 {
     // `BITS` is the number of bits in the memory representation of Fp25519 field elements. It does
     // not vary with vectorization. Where the type `BA256` appears literally in the source of this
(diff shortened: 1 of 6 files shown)


ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -97,21 +97,23 @@ use crate::{
...
+pub async fn convert_to_fp25519<V, B, const NC: usize, const NP: usize>(
+    dzkp_validator: V,
+    ctx_counter: usize,
+    first_record_id: RecordId,
+    input_shares: Vec<BitDecomposed<AdditiveShare<Boolean, NC>>>,
 ) -> Result<Vec<AdditiveShare<Fp25519, NP>>, Error>
 where
-    C: UpgradableContext,
+    V: DZKPValidator<B>,
+    B: UpgradableContext,
```
andyleiserson, line 108: Why call this type parameter `B` instead of `C`?

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
-    // addition x+rs, where rs=r+s might cause carry
-    // this is not a problem since bit 255 of rs is set to 0
-    let (sh_y, _) = integer_add::<_, TwoHundredFiftySixBitOpStep, NC>(
-        m_ctx.narrow(&Step::IntegerAddMaskToX),
-        record_id,
-        &sh_rs,
-        &x,
-    )
-    .await?;
+    assert!(NC * input_shares.len() < 120_000);
```
andyleiserson, line 140: There is an argument that defensive coding means this property should be checked here (within `convert_to_fp25519`). But I also think it would be reasonable to say that `convert_to_fp25519` isn't a general purpose routine, so doesn't have to be so aggressive about checking its inputs, and instead of checking this property here, just put the comment about the restriction alongside the assignment of `PROOF_CHUNK`.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
+    // vector of intermediate results
+    let mut vec_sh_y =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+    let mut vec_sh_r =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+    let mut vec_sh_s =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+
+    for (i, x) in input_shares.iter().enumerate() {
+        let record_id_x = first_record_id + i;
```
andyleiserson, line 157: I am assuming the name `ctx_masks_x` is referring to the "x" in `AddMaskToX`. But is that also true of `record_id_x`? Or if not, why the name `record_id_x`?

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
-            record_id,
-            Role::H3,
-            sh_y.get(i).unwrap(),
-        )
-        .await?;
-        match (&mut y, y_bit) {
-            (Some(y), Some(y_bit)) => y.push(y_bit),
-            (None, None) => (),
-            _ => unreachable!("inconsistent partial_reveal behavior"),
+    dzkp_validator.validate_chunk(ctx_counter).await?;
```
andyleiserson, line 199: It occurs to me that we should have the implementation of reveal for DZKP contexts do an is_validated check. I made a note and will plan to look at that.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -437,16 +468,19 @@ mod tests {
...
                             let match_keys =
                                 BitDecomposed::transposed_from(&*chunk).unwrap_infallible();
-                            convert_to_fp25519::<_, CONV_CHUNK, PRF_CHUNK>(
-                                ctx.set_total_records((COUNT + CONV_CHUNK - 1) / CONV_CHUNK),
+                            convert_to_fp25519::<_, _, CONV_CHUNK, PRF_CHUNK>(
+                                validator.clone(),
+                                idx,
                                 RecordId::from(idx),
-                                match_keys,
+                                vec![match_keys],
```
andyleiserson, line 483: It's probably worth changing this test to do two chunks, to exercise the proof batching in a context other than end-to-end tests.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -292,18 +298,30 @@ where
 
     let prf_key = gen_prf_key(&eval_ctx);
 
+    // CONV_CHUNK * PROOF_CHUNK
```
andyleiserson, line 301:
```suggestion
```

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -292,18 +298,30 @@ where
...
+                        let input_match_keys: &dyn Fn(usize) -> Replicated<MatchKey> =
+                            &|i| records[CONV_CHUNK * j + i].match_key.clone();
+                        BitDecomposed::<Replicated<Boolean, 256>>::transposed_from(input_match_keys)
+                            .unwrap_infallible()
+                    })
+                    .collect::<Vec<_>>();
+                convert_to_fp25519::<_, C, CONV_CHUNK, PRF_CHUNK>(
+                    validator.clone(),
+                    idx,
+                    RecordId::from(idx * PROOF_CHUNK * CONV_CHUNK),
```
andyleiserson, line 320: I think this should be `idx * PROOF_CHUNK` -- the vectorized operation across CONV_CHUNK input rows only consumes one record ID. (But also see my other note about getting rid of the separate record ID parameter.)

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
+        vec_sh_y.push(
+            integer_add::<_, TwoHundredFiftySixBitOpStep, NC>(
+                ctx_mask_x.clone(),
+                record_id_x,
+                &sh_rs,
+                x,
+            )
+            .await?
+            .0,
+        );
```
andyleiserson, line 193: I think it is worth keeping a separate assignment that grabs the sum, rather than `.0`, for clarity.

```suggestion
        let (sh_y, _) = integer_add::<_, TwoHundredFiftySixBitOpStep, NC>(
            ctx_mask_x.clone(),
            record_id_x,
            &sh_rs,
            x,
        )
        .await?;
        vec_sh_y.push(sh_y);
```

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
-        .expect("sh_s was constructed with the correct number of bits");
+        let sh_r = Vec::<AdditiveShare<BA256>>::transposed_from(&sh_r)
+            .ok()
+            .expect("sh_r was constructed with the correct number of bits");
+        let sh_s = Vec::<AdditiveShare<BA256>>::transposed_from(&sh_s)
+            .ok()
+            .expect("sh_s was constructed with the correct number of bits");
 
-    output_shares::<_, NC, NP>(&m_ctx, &sh_r, &sh_s, y)
+        results.extend(output_shares::<_, NC, NP>(&m_ctx, &sh_r, &sh_s, y)?);
```
andyleiserson, line 235: `output_shares` is already pushing things onto a vector internally, so it would be a pretty simple change (and a likely inconsequential performance improvement) to pass `&mut results` into `output_shares` and save a little work.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +128,113 @@ where
...
-        .await?;
-        match (&mut y, y_bit) {
-            (Some(y), Some(y_bit)) => y.push(y_bit),
-            (None, None) => (),
-            _ => unreachable!("inconsistent partial_reveal behavior"),
+    dzkp_validator.validate_chunk(ctx_counter).await?;
+
+    // output vec
+    // non vectorized Fp25519 shares
+    let mut results = Vec::<AdditiveShare<Fp25519, NP>>::with_capacity(NC * vec_sh_y.len());
```
andyleiserson, line 203:
```suggestion
    // output vec
    // NP-vectorized Fp25519 shares
    let mut results = Vec::<AdditiveShare<Fp25519, NP>>::with_capacity(NC * vec_sh_y.len() / NP);
```

andyleiserson (changes requested):

akoshelev:
before merging this, please do performance runs and make sure there are no regressions. Doing two 50M runs should be reasonable

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +132,103 @@ where
...
-        .await?;
-        match (&mut y, y_bit) {
-            (Some(y), Some(y_bit)) => y.push(y_bit),
-            (None, None) => (),
-            _ => unreachable!("inconsistent partial_reveal behavior"),
+    dzkp_validator.validate_chunk(ctx_counter).await?;
+
+    // output vec
+    // NP-vectorized Fp25519 shares
+    let mut results = Vec::<AdditiveShare<Fp25519, NP>>::with_capacity(NC * vec_sh_y.len());
```
andyleiserson, line 196: There was another piece of the suggestion here besides the comment -- the size should be `NC * vec_sh_y.len() / NP`.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
akoshelev: is there a better way to deal with chunks inside the validator that does not require seeding the context from the callee? Copying the specific step all the way down through all functions in MPC seems very wrong to me

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +132,103 @@ where
...
-
-    // addition r+s might cause carry,
-    // this is no problem since we have set bit 254 of sh_r and sh_s to 0
-    let sh_rs = {
-        let (mut rs_with_higherorderbits, _) = integer_add::<_, TwoHundredFiftySixBitOpStep, NC>(
-            m_ctx.narrow(&Step::IntegerAddBetweenMasks),
-            record_id,
-            &sh_r,
-            &sh_s,
+    debug_assert!(input_shares.first().iter().count() < (BITS - 128));
```
akoshelev, line 135: I don't think this is doing what it may seem to do: https://doc.rust-lang.org/std/option/struct.Iter.html

```rust
    #[test]
    fn test_option_iter() {
        let x = BitDecomposed::new(repeat_n(AdditiveShare::<BA64>::ZERO, 256));
        let bd = vec![x.clone(), x];
        debug_assert!(bd.first().iter().count() < (256 - 128)); // pass
        debug_assert!(bd[0].iter().count() < (256 - 128)); // fail
    }

```

We should add a test that checks it

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +132,103 @@ where
...
+
+    // vector of intermediate results
+    let mut vec_sh_y =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+    let mut vec_sh_r =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+    let mut vec_sh_s =
+        Vec::<BitDecomposed<AdditiveShare<Boolean, NC>>>::with_capacity(input_shares.len());
+
+    for (record_id, x) in record_id_range.clone().zip(input_shares) {
```
akoshelev, line 152: you should make your functions hard to misuse. if range of record ids is smaller than input shares, this will silently not convert some of them because `zip` iterates only until any of the iterators return `None`. 

I suggest to build some defence here, either by not allowing to pass bad parameters at compile time and if it is not possible, enforce checks at runtime

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -126,84 +132,103 @@ where
...
+
+    // get context
+    let m_ctx = dzkp_validator.context();
+    // generate sub contexts
+    let ctx_gen = &m_ctx.narrow(&Step::GenerateSecretSharing);
+    let ctx_masks = &m_ctx.narrow(&Step::IntegerAddBetweenMasks);
+    let ctx_mask_x = &m_ctx.narrow(&Step::IntegerAddMaskToX);
+
+    // vector of intermediate results
+    let mut vec_sh_y =
```
akoshelev, line 145: nit: I would suggest not to encode the type of the variable into its name - it is already known. `y_shares` looks better to me

akoshelev (changes requested):

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
andyleiserson: There was some discussion of this [here](https://github.com/private-attribution/ipa/pull/1145#discussion_r1647375213). I think there probably is a better way, but it requires a little rearranging of the internals of the validator/context. I am hoping to take a look at this at some point.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
```diff
@@ -280,10 +305,11 @@ where
 /// NP records.
 fn output_shares<C, const NC: usize, const NP: usize>(
     ctx: &C,
+    results: &mut Vec<AdditiveShare<Fp25519, NP>>,
```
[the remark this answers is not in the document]
andyleiserson: I suggested it in `https://github.com/private-attribution/ipa/pull/1180#discussion_r1664585805`.

Previously, `output_shares` created a vector, returned it, and it was then passed through as the return value of `convert_to_fp25519`. This PR is changing `convert_to_fp25519` to operate over larger chunks, which meant that it was assembling the vectors returned by `output_shares` into a larger vector. Rather than assemble small vectors and then immediately drain those smaller vectors into a larger vector, it seemed better to just assemble the output in the larger vector directly.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
andyleiserson: An issue here is that we already had variables `sh_y`, `sh_rs`, `sh_r`, `sh_s`. It seems confusing to me to have variables `sh_y` and `y_shares`. We could undertake some more extensive renaming of everything to `foo_share(s)?`, but that seems (1) perhaps undesirably more verbose, (2) not a good idea to do in the same PR as we're making a bunch of functional changes.

So I'm inclined to leave this as-is for now, and come back and do some renaming once this PR is finished if necessary.

ipa-core/src/protocol/ipa_prf/boolean_ops/share_conversion_aby.rs
andyleiserson: This is a good catch, thank you.

andyleiserson:
I pushed a commit that captures the more straightforward revisions, but there is a bigger change coming. The current version increases the minimum work unit PROOF_CHUNK * CONV_CHUNK, which is too much for CI.

danielmasny:
https://github.com/private-attribution/ipa/pull/1208/files

PR #1183, opened 2024-07-05
Matching IETF draft. Using L not lambda

benjaminsavage:
Feedback from @martinthomson was that we should not use lambda as it's hard to type.


akoshelev:
Thank you both!

akoshelev (approved):

Issue #1029, opened 2024-04-26
Allow PRSS module to generate only left or right randomness


akoshelev:
The only protocol that can benefit from it as of 2024 is sharded shuffle, but it generates a large number of masks per record, so savings here can be quite substantial. @andyleiserson suggested that compiler in some cases can derive that part of the masking value is not used and remove the code that generates it, but it would be great to validate that theory.

Alternatively, we can provide a method `generate(&self, record_id: RecordId, direction: Direction)`. It would probably require some rethinking of our current trait family inside PRSS, especially `FromRandom`.

akoshelev:
Another use case that requires supporting it - zero-knowledge proofs require each helper to be a prover and a verifier and as part of that job, they need to generate left and right randomness at different times. Currently, they need to generate both at the same time and then carry unused value until it can be used later

PR #1185, opened 2024-07-05
Enforce index uniqueness inside the Generator struct

akoshelev:
This is the first step to allow PRSS to generate only one part of randomness, shared with one helper. The motivating issue is #1029. To support that, we need to enforce uniqueness checks at the lower layer because each generator now can be called at different times.

This change moves the `used` variable to `Generator` struct. This required some rethinking of `generate` method parameter types - instead of raw `u128`, it now takes `PrssIndexu128`. Mainly for debugging purposes, to be able to print "index:offset" when a duplicate index is used.

### DIFF
--- a/ipa-core/src/protocol/prss/mod.rs
+++ b/ipa-core/src/protocol/prss/mod.rs
@@ -1,17 +1,12 @@
 mod crypto;
-use std::{
-    collections::HashMap,
-    fmt::{Debug, Display, Formatter},
-    marker::PhantomData,
-    num::TryFromIntError,
-    ops::AddAssign,
-};
+use std::{collections::HashMap, fmt::Debug, marker::PhantomData, ops::AddAssign};
 
 pub use crypto::{
     FromPrss, FromRandom, FromRandomU128, Generator, GeneratorFactory, KeyExchange,
     SharedRandomness,
 };
 use generic_array::{sequence::GenericSequence, ArrayLength, GenericArray};
+pub(super) use internal::PrssIndex128;
 use x25519_dalek::PublicKey;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


ipa-core/src/protocol/prss/mod.rs
```diff
@@ -444,28 +415,29 @@ pub mod test {
...
 
-        // Calling g1 twice with the same input produces the same output.
-        assert_eq!(g1.generate(12), g1.generate(12));
+        // Calling generators seeded with the same key produce the same output
+        assert_eq!(g1.generate(12), make(SEED).0.generate(12));
         // Now that g1 has been invoked more times than g2, we can check
         // that it isn't cheating by using an internal counter.
         assert_eq!(g1.generate(100), g2.generate(100));
-        assert_eq!(g2.generate(13), g2.generate(13));
+        assert_eq!(g1.generate(13), g2.generate(13));
```
akoshelev, line 429: @martinthomson I thought that was a typo and you actually wanted to use $G_1$. Let me know if it isn't the case and I can revert this change

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -103,12 +61,25 @@ impl From<PrssIndex128> for u128 {
...
 }
 
+impl TryFrom<u128> for PrssIndex128 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: u128) -> Result<Self, Self::Error> {
+        let value = u64::try_from(value)?;
+
+        Ok(Self {
+            index: u32::try_from(value >> 32).unwrap().into(),
```
benjaminsavage, line 71: I don't understand why we are doing all of this fanciness here. Perhaps a comment could help explain the rationale.

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -232,7 +203,7 @@ impl RngCore for SequentialSharedRandomness {
     // That is OK for the same reason that we use in converting a `u128` to a small `Field`.
     #[allow(clippy::cast_possible_truncation)]
     fn next_u64(&mut self) -> u64 {
-        let v = self.generator.generate(self.counter);
+        let v = self.generator.generate(self.counter.offset(0));
```
benjaminsavage, line 206: I'm not sure what's going on here and how this is connected to the PR.

benjaminsavage (approved):
I am a bit confused by a few sections, but overall this looks pretty straightforward.

ipa-core/src/protocol/prss/mod.rs
akoshelev: This change is required now because `generate` method takes `PrssIndex128` type instead of `u128`. It does feel that for consistency reasons we should use `PrssIndex`/`PrssIndex128` everywhere in PRSS module, but maybe @andyleiserson has different opinion about it

ipa-core/src/protocol/prss/mod.rs
akoshelev: This code was moved from `From<u64>` implementation. The description of `PrssIndex128` is not included in the PR and that's why it probably looks confusing. Basically we need two indices to be able to generate more than 16 bytes of randomness: PRF use-case is one example where we need to get 256 bit values out of PRSS. AES cipher generates 16 byte blocks, so we need to call it more than once with unique IV if we want more entropy. 

Protocols request randomness using `RecordId` which is a 32 bit integer and we don't have a lot of bits there to ask for more than one value for the same index. So to keep the old code working, we just implemented 32 bit `offset` field  and used it to generate large chunks

* A protocol requests BA256 value for row 0, using `prss.generate(RecordId::from(0))`
* PRSS creates an index: `0:0` and generates the lowest 16 bytes of randomness
* PRSS creates index: `0:1` and generates the highest 16 bytes of randomness
* PRSS concatenates them and returns back to the protocol

Offsets are opaque to the MPC code and determined by the number of iterations required from PRSS to generate the desired value.

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -103,12 +61,25 @@ impl From<PrssIndex128> for u128 {
...
 
+impl TryFrom<u128> for PrssIndex128 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: u128) -> Result<Self, Self::Error> {
+        let value = u64::try_from(value)?;
+
+        Ok(Self {
+            index: u32::try_from(value >> 32).unwrap().into(),
+            offset: u32::try_from(value & u64::from(u32::MAX)).unwrap(),
```
martinthomson, line 72: I have a new request...

Can we ensure that either the offset or the index is <= 2^11?  Probably the offset.  The analysis I did for PRSS indicates that we can't use this PRSS algorithm more than 2^43 times and this might be a good way to cap that.

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -133,6 +104,16 @@ impl From<RecordId> for PrssIndex {
     }
 }
 
+impl AddAssign<u32> for PrssIndex {
+    fn add_assign(&mut self, rhs: u32) {
+        if let Some(v) = self.0.checked_add(rhs) {
+            self.0 = v;
```
martinthomson, line 110: This would be an end run around our limits.  If we cap the offset, then we should also cap the value here.  In this case, we would be ensuring that `rhs < (1<<11) && (self.0 + rhs) & 0xfffff_f800 == 0`.

ipa-core/src/protocol/prss/mod.rs
martinthomson: No, this looks fine.

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -209,15 +180,15 @@ impl<'a, Z: ArrayLength> Iterator for ChunksIter<'a, Z> {
 /// in APIs that expect `Rng`.
 pub struct SequentialSharedRandomness {
     generator: Generator,
-    counter: u128,
+    counter: PrssIndex,
```
martinthomson, line 183: If you do as I suggest above, then you should not use `PrssIndex` here.  It is better to start with 0_u128, I think.

martinthomson (commented):
You touched it, now you have a few more problems to think about.

Overall, this is fine, though I think that you are trying to use `PrssIndex` more than necessary.  At some point, it's OK to convert to u128 and carry on from there.  The problem with taking that type too deep is that code that handles the whole business of safety in that type needs to be set aside at some point.

ipa-core/src/protocol/prss/mod.rs
akoshelev: I can cap the `value` to be < 2^43, that seems easier than deciding which side of the index we want to cap. Also gives us the flexibility of giving more bytes to offset or index

ipa-core/src/protocol/prss/mod.rs
akoshelev: If we enforce it at `PrssIndex128`, then it seems we could keep the check in one place?

ipa-core/src/protocol/prss/mod.rs
akoshelev: yea, I can do that. The reason I made this change is that `PrssIndex` is used everywhere in `SharedRandomness` trait and we were going that direction (from u128 to `PrssIndex`). @andyleiserson do you have any thoughts here?

andyleiserson (approved):
Using PrssIndex128 for both types of generators means that the cipher invocation count check only needs to be done in one place. On the other hand, some of the index stuff doesn't apply to sequential randomness -- Ben's question about what `.offset(0)` is for, any indices in error messages will have a useless `:0`, and the block limit for sequential generation is artificially reduced to 2^32. None of these seem like major issues.

I guess if there's a strong desire to maintain the `u128` API to `Generator` for the sequential randomness case, there could be a wrapper around `Generator` that provides the same API and implements the used index tracking and `PrssIndex128` conversion.

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -695,4 +711,41 @@ pub mod test {
...
+        assert_eq!(
+            index128,
+            u128::from(PrssIndex128::try_from(index128).unwrap())
+        );
+    }
+
+    proptest! {
+        #[test]
+        fn prss_index_128_conversions(index in 0..u32::MAX, offset in 0..PrssIndex128::MAX_OFFSET) {
+            assert_8_byte_index_is_valid(index ,usize::try_from(offset).unwrap());
```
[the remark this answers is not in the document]
akoshelev: yea I don't think rustfmt can help here :( spaces in macros may matter

akoshelev:
> the block limit for sequential generation is artificially reduced to 2^32. None of these seem like major issues.

I agree. After generating 2^32 values we are already at 50% chance of collision due to the birthday bound, so I don't see a lot of value trying to get above that limit. 

> I guess if there's a strong desire to maintain the u128 API to Generator for the sequential randomness case, there could be a wrapper around Generator that provides the same API and implements the used index tracking and PrssIndex128 conversion.

We could do that, although because of the `RngCore` interface, we can't fully leverage all 128 bits here and the struct guarantees the uniqueness of index used for PRSS itself

martinthomson:
The birthday bound for AES applies at 2^64, not 2^32 and only if we are feeding in random values.  We are essentially doing that because we don't use the PRP directly, but XOR in the counter.  This is to give us what the paper describes as circular correlation resistance.  Either way, we're a long way from hitting any birthday bounds.

ipa-core/src/protocol/prss/mod.rs
martinthomson: I'm ok with having PrssIndex128 be the place we enforce limits, but the way in which we do that needs to be clear.

akoshelev:
> The birthday bound for AES applies at 2^64,

Sequential randomness reduces the value received from the cipher to 8 bytes by truncation. So I think it still applies here

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -20,86 +15,109 @@ use crate::{
...
 
-#[cfg(debug_assertions)]
-impl Debug for UsedSet {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "IndicesSet(key={})", self.key)
+    #[cfg(debug_assertions)]
+    impl From<u64> for PrssIndex128 {
+        fn from(value: u64) -> Self {
+            Self::try_from(u128::from(value)).unwrap()
+        }
```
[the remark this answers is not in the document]
akoshelev: It is actually being used now in tests, after the change to `Generator` interface

ipa-core/src/protocol/prss/mod.rs
akoshelev: Let me know if it is still not clear in the latest version of this PR.

akoshelev:
I am going to merge this change, if there are more comments here I'll address them in a follow up

PR #1187, opened 2024-07-06
Allow PRSS to generate randomness shared with only one helper

akoshelev:
Builds on top of #1185

This closes https://github.com/private-attribution/ipa/issues/1029 by adding a method to PRSS implementation that allows callers to specify which randomness they want to obtain: shared with the helper on the right or on the left. This saves 50% of computation in cases where only one side of PRSS is required. https://github.com/private-attribution/ipa/issues/1029 lists the cases where it is useful.

Happy to bikeshed on naming, it gets harder and harder to find good names

### DIFF
--- a/ipa-core/src/protocol/prss/crypto.rs
+++ b/ipa-core/src/protocol/prss/crypto.rs
@@ -221,7 +221,7 @@ pub trait SharedRandomness {
         T::from_random(
             self.generate_chunks_one_side(index, direction)
                 .next()
-                .unwrap_or_else(|| panic!("Can generate randomness for index {index:?}")),
+                .unwrap_or_else(|| panic!("Can't generate randomness for index {index:?}")),
         )
     }
 
--- a/ipa-core/src/protocol/prss/crypto.rs
+++ b/ipa-core/src/protocol/prss/crypto.rs
@@ -11,7 +11,7 @@ use x25519_dalek::{EphemeralSecret, PublicKey};
 
 use crate::{
     ff::Field,
-    protocol::prss::PrssIndex,
+    protocol::prss::{PrssIndex, PrssIndex128},
     secret_sharing::{
         replicated::{semi_honest::AdditiveShare, ReplicatedSecretSharing},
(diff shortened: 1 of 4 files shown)


ipa-core/src/protocol/prss/crypto.rs
```diff
@@ -193,6 +204,27 @@ pub trait SharedRandomness {
...
+    fn generate_one_side<T: FromRandom, I: Into<PrssIndex>>(
+        &self,
+        index: I,
+        direction: Direction,
+    ) -> T {
+        let index = index.into();
+        T::from_random(
+            self.generate_chunks_one_side(index, direction)
+                .next()
+                .unwrap_or_else(|| panic!("Can generate randomness for index {index:?}")),
```
benjaminsavage, line 224: Can or can't?

benjaminsavage (approved):

akoshelev:
Here is the link to the benchmark that forced me to implement custom iterator. 


Using `Zip` to combine two iterators. Benchmark against the stable `main`. 15% regression

pdf

Using `ChunksIter` with inline calls. Benchmark against the stable `main`. No perf change 
pdf

ipa-core/src/protocol/prss/mod.rs
```diff
@@ -103,12 +62,25 @@ impl From<PrssIndex128> for u128 {
...
 }
 
+impl TryFrom<u128> for PrssIndex128 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: u128) -> Result<Self, Self::Error> {
+        let value = u64::try_from(value)?;
+
+        Ok(Self {
+            index: u32::try_from(value >> 32).unwrap().into(),
```
martinthomson, line 72:
```suggestion
            index: PrssIndex::from(u32::try_from(value >> 32).unwrap()),
```

martinthomson (approved):
I had a bunch of comments on the previous one, which you should address there.  Here, the changes look fine.

PR #1189, opened 2024-07-10
Add a test that demonstrates a successful additive attack against malicious reshare protocol

akoshelev:
I am happy to be wrong, but there might be a problem with the reshare implementation for active security. The claim we made that it is secure against active adversaries because we are executing semi-honest reshare twice (once for `x` and for `rx`). However, the semi-honest protocol implementation makes one helper's work unchecked by others and they can launch an additive attack because of that. 

```rust
        // `to_helper.left` calculates part1 = (self.0 + self.1) - r1 and sends part1 to `to_helper.right`
        // This is same as (a1 + a2) - r2 in the diagram
        if ctx.role() == to_helper.peer(Direction::Left) {
            ... // some work to send and receive
            Ok(Replicated::new(part1 + part2, r.1)) <- ok, shares are checked by the right helper
        } else if ctx.role() == to_helper.peer(Direction::Right) {
            // `to_helper.right` calculates part2 = (self.left() - r0) and sends it to `to_helper.left`
            // This is same as (a3 - r3) in the diagram
            ... // some work to send and receive
            Ok(Replicated::new(r.0, part1 + part2)) <- ok, shares are checked by the left helper
        } else {
            Ok(Replicated::new(r.0, r.1)) // <- here is the problem. nobody checks these values
        }
```

This PR just adds a test that demonstrates the ability of one helper to successfully launch an additive attack.

### DIFF
--- a/ipa-core/src/protocol/basics/reshare.rs
+++ b/ipa-core/src/protocol/basics/reshare.rs
@@ -212,8 +212,8 @@ mod tests {
                     Reshare,
                 },
                 context::{
-                    Context, SemiHonestContext, UpgradableContext, UpgradedContext,
-                    UpgradedMaliciousContext, Validator,
+                    Context, SemiHonestContext, SpecialAccessToUpgradedContext, UpgradableContext,
+                    UpgradedContext, UpgradedMaliciousContext, Validator,
                 },
                 prss::SharedRandomness,
                 RecordId,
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/basics/reshare.rs
```diff
@@ -212,21 +212,25 @@ mod tests {
...
                 prss::SharedRandomness,
                 RecordId,
             },
             rand::{thread_rng, Rng},
             secret_sharing::{
                 replicated::{
-                    malicious::{AdditiveShare as MaliciousReplicated, ExtendableField},
+                    malicious::{
+                        AdditiveShare as MaliciousReplicated, ExtendableField,
+                        ThisCodeIsAuthorizedToDowngradeFromMalicious,
```
bmcase, line 226: why did you need to add `ThisCodeIsAuthorizedToDowngradeFromMalicious`?

akoshelev:
I spoke with @andyleiserson and he pointed me to the fact that this test simulates a malicious reveal. Yes, the shares are inconsistent, but it is detected during malicious reveal, so bad actor cannot trick other helpers to generate a different permutation. 

So it seems that inconsistent sharing set by helper that does no work shouldn't be a problem for active security. It they could somehow trick everyone in the group to believe they hold $[[v+r]]$ instead of $[[v]]$, then this will be an issue

ipa-core/src/protocol/basics/reshare.rs
akoshelev: That's just to allow calling `downgrade` method manually

PR #1190, opened 2024-07-12
Isolate reveal operation inside PRF evaluation

akoshelev:
In order to get malicious security using MAC, all helpers must run `validate` before revealing any information.

In order to be efficient, validate must check data in batches, so PRF evaluation needs to do all the processing before calling validate.

In terms of efficiency, even the semi-honest implementation will be less efficient. There is an extra step to hold all $z$ and $r$ values in memory before validating them. But that seems necessary to achieve security against active adversaries.

There is no new functionality, just restructuring to prepare for MAC-based validation


akoshelev:
I don't think this is the direction we want to pursue. A much more promising approach based on validating chunks is being implemented for both dzkp and MAC.

PR #1191, opened 2024-07-15
Fix a corner case in stall detection

andyleiserson:
The waker at `start` is actually for `self.next + wakers.len()`, not `self.next`. The `max_polled_idx` case (line 284) can also put `self.next` in the list. Having two zeros in the list (when `self.next` is zero) results in an underflow computing `num - 1` in `to_ranges`.

I am not sure how best to test this (either a regression test for the fix, or to make sure the change doesn't break anything else). But opening a PR anyways so it's less likely the issue just gets lost.


akoshelev:
thanks for fixing it. I've never tested it properly, was waiting for https://github.com/tokio-rs/tracing/tree/master/tracing-mock to be released

akoshelev (approved):

PR #1199, opened 2024-07-22
Add malicious validate to quicksort

andyleiserson:
Without chunking, for now.

The tests for share conversion created the validator in the test code, with a record count tailored for the test. Quicksort creates the validator internally, with a reasonable chunk size for production. However, this resulted in paying the cost to verify a production-sized chunk, even for test cases with minimal input size. There are some changes here to remedy this.

### DIFF
--- a/ipa-core/src/protocol/context/dzkp_validator.rs
+++ b/ipa-core/src/protocol/context/dzkp_validator.rs
@@ -35,7 +35,11 @@ pub type Array256Bit = BitArray<[u8; 32], Lsb0>;
 
 type BitSliceType = BitSlice<u8, Lsb0>;
 
-const BIT_ARRAY_SHIFT: usize = 8;
+const BIT_ARRAY_LEN: usize = 256;
+const BIT_ARRAY_MASK: usize = BIT_ARRAY_LEN - 1;
+const BIT_ARRAY_SHIFT: usize = BIT_ARRAY_LEN.ilog2() as usize;
+
+pub const TARGET_PROOF_SIZE: usize = 50_000_000;
 
 /// `MultiplicationInputsBlock` is a block of fixed size of intermediate values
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -350,6 +363,10 @@ impl MultiplicationInputsBatch {
         let position_within_block_start = (length * id_within_batch) % 256;
         let position_within_block_end = position_within_block_start + segment.len();
 
+        if self.vec.len() <= block_id {
```
akoshelev, line 366: Is there an advantage of using `&mut` pointer to the element inside the vector, or we could just allocate a block, copy the segment into it and do `vec.push`? We can take ourselves off the resize business in this case

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -389,8 +406,14 @@ impl MultiplicationInputsBatch {
 
         let id_within_batch = usize::from(record_id) - usize::from(self.first_record);
         let block_id = (segment.len() * id_within_batch) >> BIT_ARRAY_SHIFT;
-
         let length_in_blocks = segment.len() >> BIT_ARRAY_SHIFT;
+        if self.vec.len() <= block_id + length_in_blocks {
```
akoshelev, line 410: same here?

akoshelev (approved):

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: I don't think we can assume that the records show up in order. I could resize to one less and then push, which doesn't get rid of the resize, but might help if the common case is the records arriving in order.

Issue #1195, opened 2024-07-18
Provide a way for helpers to behave maliciously inside the test infrastructure


akoshelev:
Currently, whenever we want to reproduce malicious behaviour, we reimplement the protocol and run additive attacks manually. There does not seem to be a need to do that, because we can just instruct the test infrastructure to corrupt bits before sending them out to the other helpers.

PR #1202, opened 2024-07-26
Interface for test infrastructure to inspect streams

akoshelev:
This change implements #1195 by providing a means for unit tests to intercept send requests that send circuit-specific data between MPC helpers.

This change also updates the existing unit tests for malicious `reshare` and `reveal` protocols to showcase the API and test it in real scenarios. In both cases, using it, led to significant reduction in code redundancy. The key improvement is that now tests are not required to re-implement the basic protocol to inject additive attacks. It can now be done outside the circuit being tested.

The support for sharding exists at the API level, but no actual implementation allows intercepting traffic between shards. There is currently not a lot of sharded protocols, so this can be postponed until later.

### DIFF
--- a/ipa-core/src/ff/mod.rs
+++ b/ipa-core/src/ff/mod.rs
@@ -90,6 +90,46 @@ pub trait Serializable: Sized {
             .map_err(Into::into)
             .unwrap_infallible()
     }
+
+    /// This method provides the same functionality as [`Self::deserialize`] without
+    /// compile-time guarantees on the size of `buf`. Therefore, it is not appropriate
+    /// to use in production code. It is provided as convenience method
+    /// for tests that are ok to panic.
+    ///
+    /// ## Panics
+    /// If the size of `buf` is not equal to `Self::Size` or if `buf` bytes
+    /// are not a valid representation of this instance. See [`Self::deserialize`] for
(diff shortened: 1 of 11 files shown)


andyleiserson (approved):
This is really cool. Making it easier to write tests of malicious behavior means we'll write more of them.

One thing that we can't do with this API is simulate the case where a helper both sends a manipulated share value to another helper and changes its copy of the share to match. This doesn't seem to me like something we have to do, but it does change the properties of the testing in a way that might be worth thinking about. I think the previous `reshare_with_additive_attack` did simulate this, by corrupting `part1`, which is both sent to the peer and used in the calculation of the output share. Specifically for reshare, we could simulate this with the new infrastructure by corrupting the values sent in both directions between a pair of helpers the same way.

akoshelev:
> One thing that we can't do with this API is simulate the case where a helper both sends a manipulated share value to another helper and changes its copy of the share to match.

yea that probably requires a different mechanism. I didn't think that checking `validate` on malicious helper is useful because they can be assumed to return whatever they want, but if we ever need to do that, we could either write a different protocol, corrupt the data on both ways as you suggested or get a handle into malicious validation, so we can change `u` and/or `v` before calling `validate`

Issue #1204, opened 2024-07-30
MAC-based validation is broken because it uses semi-honest reveal protocol


akoshelev:
As part of validation process, accumulators run the check zero protocol, implemented [here](https://github.com/private-attribution/ipa/blob/05581c1e0f15461f98607d62c4819b3f38fa7c77/ipa-core/src/protocol/basics/check_zero.rs#L42). 

This protocol supports semi-honest security only because it runs the semi-honest reveal, so an active adversary can cheat and send a broken share to its peer and make it think that `rv == ZERO`.

We should close that gap and use malicious reveal instead

andyleiserson:
I'm not sure that even a malicious reveal is sufficient. I pushed a test case (using the new support for simulating malicious behavior) in #1205. With our current implementation, a malicious helper is not limited to choosing share values for a reveal operation based on shares received from other helpers during the reveal -- the malicious helper can choose the share value it sends in the `r * v` multiply that precedes the reveal based on shares received from other helpers as part of the reveal operation (because helpers running ahead of each other is restricted only by data dependencies, i.e., helpers can generally get up to two steps ahead).

The `r * v` multiply is semi-honest, but (1) that's what the paper specifies, (2) it's not clear how to use a malicious multiply here since this is part of the verification step of the malicious protocol (i.e. a "chicken and egg" problem).

I think we need some kind of synchronization within the check zero protocol to prevent this. Although I'm nervous about other places that there could be similar problems.

andyleiserson:
A paper about a similar issue with lack of synchronization: https://eprint.iacr.org/2025/789

PR #1205, opened 2024-07-31
Test of malicious behavior during zero check protocol

andyleiserson:
Test case related to #1204 -- will post more detail to the issue.

### DIFF
--- a/ipa-core/src/helpers/transport/in_memory/config.rs
+++ b/ipa-core/src/helpers/transport/in_memory/config.rs
@@ -1,4 +1,6 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, future::ready, pin::Pin};
+
+use futures::{Future, FutureExt};
 
 use crate::{
     helpers::{HelperIdentity, Role, RoleAssignment},
@@ -42,19 +44,33 @@ pub trait StreamInterceptor: Send + Sync {
     /// from additive attacks without additional measures implemented
     /// at the transport layer, like checksumming, share consistency
     /// checks, etc.
-    fn peek(&self, ctx: &Self::Context, data: &mut Vec<u8>);
+    fn peek<'a>(
+        &'a self,
+        ctx: &'a Self::Context,
+        data: &'a mut Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
 }
 
-impl<F: Fn(&InspectContext, &mut Vec<u8>) + Send + Sync + 'static> StreamInterceptor for F {
+impl<F> StreamInterceptor for F
+where
+    for<'a> F: Fn(&'a InspectContext, &'a mut Vec<u8>) -> Pin<Box<dyn Future<Output = ()> + Send>>
+        + Send
+        + Sync
+        + 'a,
+{
     type Context = InspectContext;
 
-    fn peek(&self, ctx: &Self::Context, data: &mut Vec<u8>) {
-        (self)(ctx, data);
+    fn peek(
+        &self,
+        ctx: &Self::Context,
+        data: &mut Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send>> {
+        (self)(ctx, data)
     }
 }
 
 /// The general context provided to stream inspectors.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct InspectContext {
     /// The shard index of this instance.
     /// This is `None` for non-sharded helpers.
@@ -76,7 +92,7 @@ pub struct InspectContext {
 #[inline]
 #[must_use]
 pub fn passthrough() -> Arc<dyn StreamInterceptor<Context = InspectContext>> {
-    Arc::new(|_ctx: &InspectContext, _data: &mut Vec<u8>| {})
+    Arc::new(|_ctx: &InspectContext, _data: &mut Vec<u8>| ready(()).boxed())
 }
 
 /// This narrows the implementation of stream seeker
@@ -93,7 +109,10 @@ pub struct MaliciousHelper<F> {
     inner: F,
 }
 
-impl<F: Fn(&MaliciousHelperContext, &mut Vec<u8>) + Send + Sync> MaliciousHelper<F> {
+impl<F> MaliciousHelper<F>
+where
+    F: Fn(MaliciousHelperContext, &mut Vec<u8>) -> Pin<Box<dyn Future<Output = ()> + Send>>,
+{
     pub fn new(role: Role, role_assignment: &RoleAssignment, peeker: F) -> Arc<Self> {
         Arc::new(Self {
             identity: role_assignment.identity(role),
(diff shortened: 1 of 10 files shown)


ipa-core/src/helpers/transport/in_memory/config.rs
```diff
@@ -93,7 +109,10 @@ pub struct MaliciousHelper<F> {
     inner: F,
 }
 
-impl<F: Fn(&MaliciousHelperContext, &mut Vec<u8>) + Send + Sync> MaliciousHelper<F> {
+impl<F> MaliciousHelper<F>
+where
+    F: Fn(MaliciousHelperContext, &mut Vec<u8>) -> Pin<Box<dyn Future<Output = ()> + Send>>,
```
akoshelev, line 114: Do we need this given that it is only used in non-async context? it may be better to convert non-async code to async inside this helper - non-async code is much easier to deal with

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -48,26 +55,82 @@ where
...
     let rv = F::from_array(
-        &rv_share
-            .reveal(ctx.narrow(&Step::RevealR), record_id)
-            .await?,
+        &malicious_reveal(ctx.narrow(&Step::RevealR), record_id, None, &rv_share)
+            .await?
+            .expect("full reveal should always return a value"),
     );
 
+    tracing::info!("{:?}", &rv);
```
akoshelev, line 65:
```suggestion
```

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -48,26 +55,82 @@ where
...
     Ok(rv == F::ZERO)
 }
 
+#[cfg(test)]
+pub async fn check_zero_fp32bitprime<C>(
```
akoshelev, line 70: do you have plans to use it outside of this module or we can move it into `tests`?

akoshelev: the same argument applies to `SH1` as well

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -124,4 +187,122 @@ mod tests {
 
         Ok(())
     }
+
+    pub struct NotifyOnceCell<T: Clone + Send + Sync> {
```
akoshelev, line 191: how is it different from https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html? Maybe you could elaborate that in the documentation to this struct

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -124,4 +187,122 @@ mod tests {
...
+            let notify = inner.notify.clone();
+            async move {
+                notify.notified().await;
+                self.inner.lock().unwrap().cell.get().unwrap().clone()
+            }
+            .boxed()
+        }
+    }
+
+    struct MaliciousCheckZeroInterceptor {
```
akoshelev, line 231: lets explain what is the goal for this interceptor because it does seem very specific to check zero use case

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -124,4 +187,122 @@ mod tests {
...
+                    && ctx.identity == HelperIdentity::ONE
+                    && ctx.dest == HelperIdentity::TWO.as_str()
+            {
+                async {
+                    assert_eq!(data.len(), 4);
+                    let (sh1, sh2) = join(super::SH1.get(), self.sh2.get()).await;
+                    tracing::info!("got shares: {sh1:?} {sh2:?}");
+                    let adjusted_share = -sh1 - sh2;
+                    tracing::info!("adjusted share {adjusted_share:?}");
+                    adjusted_share.serialize(
```
akoshelev, line 266: I've added `serialize_to_slice` and `deserialize_from_slice` methods to `Serializable` trait (only for tests). I think you may want to use them here

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -124,4 +187,122 @@ mod tests {
...
+
+        let [res0, res1, res2] = world
+            .semi_honest(v, |ctx, v| async move {
+                check_zero_fp32bitprime(ctx.set_total_records(1), RecordId::FIRST, &v)
+                    .await
+                    .unwrap()
+            })
+            .await;
+
+        assert_eq!(res0, false, "zero check failed on H1");
```
akoshelev, line 304: do you prefer this syntax over `assert!(!res0, "...")`?

ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -124,4 +187,122 @@ mod tests {
...
+        }
+    }
+
+    #[tokio::test]
+    async fn malicious_check_zero() {
+        let mut config = TestWorldConfig::default();
+        config.stream_interceptor = Arc::new(MaliciousCheckZeroInterceptor::new());
+        let world = TestWorld::new_with(&config);
+        let mut rng = thread_rng();
+        let v = rng.gen::<Fp32BitPrime>();
```
akoshelev, line 294: Does this test benefit from using a random value or we can use a constant here to avoid false negatives?

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -80,49 +85,141 @@ pub trait Reveal<C: Context, const N: usize>: Sized {
...
+    if Some(ctx.role()) == excluded {
+        Ok(None)
+    } else {
+        let (share_from_left, share_from_right) = try_join(
+            left_receiver.receive(record_id),
+            right_receiver.receive(record_id),
+        )
+        .await?;
+
+        tracing::info!("reveal ({:?}): left {left:?} right {right:?} from left {share_from_left:?} from right {share_from_right:?}", ctx.role());
```
akoshelev, line 210: that will spam the output I think similarly to another `info!` trace above. Lets change it to `trace!` or remove

akoshelev:
What I haven't realized is that this test demonstrates that a malicious helper can break the security by making other helpers accept the result of check zero protocol even when `u*v` != 0. It is a very neat attack discovered by @andyleiserson:

H1 is malicious, H2, H3 are honest helpers executing `check_zero`

* H1, H2 and H3 execute the first part of check_zero protocol - semi-honest multiply of `r*v` (according to the paper, this is the right thing to do here).
* as part of this multiply, H2 sends data to H3, H3 to H1 and H1 **does not send anything to H2**.
* H1 and H3 move to the reveal phase. H1 because it can do whatever it wants (and it has the shares from multiplication), H3 because from its perspective multiplication is done
* H3 sends its shares to H2 and H1 and waits for others
* H1 now can combine the shares of `r*v` because it has all three of them.

At this point, H1 can craft the response for multiplication to H2 that is still waiting and make the resulting share $\hat{r} = r + a$ equal to 0, completely breaking the security of MAC.

PR #1206, opened 2024-08-01
Implement reveal for upgraded contexts only

andyleiserson:
Another piece of #1112. This modifies the reveal implementations so that reveal is implemented only for upgraded contexts, and not for any `C: Context`. The danger of `impl Reveal<C: Context> for /* semi honest */ AdditiveShare<F>` is that it could accidentally be invoked by a protocol intending to have DZKP malicious security.

This changes the `check_zero` protocol used for MAC-based malicious security to use a malicious reveal, however, there is still a problem with the zero check, see https://github.com/private-attribution/ipa/issues/1204#issuecomment-2261313344.

The reveals in PRF evaluation are changed to call `malicious_reveal` directly because the PRF doesn't yet take a context of an appropriate type. I've added a TODO, which can be cleaned up with the changes for malicious security in the PRF.

### DIFF
--- a/ipa-core/src/protocol/basics/check_zero.rs
+++ b/ipa-core/src/protocol/basics/check_zero.rs
@@ -31,7 +31,7 @@ use crate::{
 /// 2.) If the randomly chosen value `r` is any other value in the field aside from zero, then the revealed value will NOT be zero.
 ///
 /// Clearly, the accuracy of this protocol is highly dependent on the field that is used.
-/// In a large field, like the integers modulo 2^31 - 1, the odds of `check_zero` returning "true"
+/// In a large field, like the integers modulo 2^31 - 1, the odds of `malicious_check_zero` returning "true"
 /// when `v` is NOT actually a sharing of zero is extrmely small; it is less than one two billion odds.
 /// In a silly field, like our test field of the integers modulo 31, the odds are very good. It'll incorrectly return "true"
 /// about 3.2% of the time.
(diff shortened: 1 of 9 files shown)


ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -49,9 +52,9 @@ where
     let rv_share =
         semi_honest_multiply(ctx.narrow(&Step::MultiplyWithR), record_id, &r_sharing, v).await?;
     let rv = F::from_array(
-        &rv_share
-            .reveal(ctx.narrow(&Step::RevealR), record_id)
-            .await?,
+        &malicious_reveal(ctx.narrow(&Step::RevealR), record_id, None, &rv_share)
```
akoshelev, line 55: it is maybe worth renaming this protocol to `check_zero_malicious` because it is no longer "default" implementation (can be made cheaper by using semi-honest reveal). We only use it in MAC-based validation, so it should be just a rename imo

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -383,61 +490,113 @@ mod tests {
...
-    pub fn malicious_validation_fail(partial: bool) {
-        const STEP: &str = "malicious-reveal";
+    #[test]
+    pub fn dzkp_malicious_generic_validation_fail() {
+        run(move || async move {
+            let partial = false;
+            let mut rng = thread_rng();
+            let mut config = TestWorldConfig::default();
+            config.stream_interceptor =
+                MaliciousHelper::new(Role::H3, config.role_assignment(), interceptor::<Boolean>);
```
akoshelev, line 574: nice! it works for DZKP too

akoshelev (approved):

PR #1208, opened 2024-08-01
Malicious Fp25519 share conversion

andyleiserson:
Modify Fp25519 share conversion to support DZKP-based malicious security. By implementing the logic to periodically verify batches in the validator, significantly fewer changes to the protocol are required.

Additional changes are required to make batched proof verification work with seq_join -- right now, there is nothing to restrict seq_join (including validated_seq_join) from scheduling futures that should not be writing into the current batch.


andyleiserson:
I think there should be a test that induces a DZKP validation failure and verifies (1) error from primary record, (2) error from non-primary records, (3) success of unaffected batch. But I didn't want to delay opening this PR to have that, and I may work on rebasing the batching changes first.

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -73,6 +78,52 @@ pub trait Reveal<C: Context> {
...
+        record_id: RecordId,
+        excluded: Option<Role>,
+    ) -> impl Future<Output = Result<Option<Self::Output>, Error>> + Send + 'fut
+    where
+        C: 'fut,
+    {
+        ctx.parallel_join(zip(self, repeat(ctx.clone())).enumerate().map(
+            |(i, (bit, ctx))| async move {
+                generic_reveal(
+                    ctx.narrow(&TwoHundredFiftySixBitOpStep::Bit(i)),
```
akoshelev, line 100: this is probably too specific for a generic reveal on a slice? Do you want to use `BitDecomposed` here or something specific to boolean shares maybe?

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -73,6 +78,52 @@ pub trait Reveal<C: Context> {
...
+            },
+        ))
+        .map(move |res| {
+            res.map(move |vec| {
+                match vec.first() {
+                    None => (excluded != Some(ctx.role())).then(Vec::new),
+                    Some(&None) => None,
+                    Some(&Some(_)) => Some(
+                        // Transform `Vec<Option<V>>` to `Option<Vec<V>>`.
+                        vec.into_iter()
```
akoshelev, line 115: I hope vec specialization works here

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -605,6 +622,18 @@ pub trait DZKPValidator<B: UpgradableContext>: Send + Sync {
...
+    ///
+    /// Validation will not generally occur immediately. The indicated record ID is marked as
+    /// complete and the future blocks pending all other records in the batch also requesting
+    /// validation. Once `validate_record` has been called for all records in the batch, the
+    /// batch is verified, and all of the `validate_record` futures complete.
+    ///
+    /// This API may only be used when the number of records per batch is the same for every
+    /// step submitting intermediates to this validator. It also requires that `set_total_records`
+    /// is set appropriately on the context that is used to create the validator.
+    async fn validate_record(&self, record_id: RecordId) -> Result<(), Error>;
```
akoshelev, line 635: I've been working with MAC-based validation and I have this method on `UpgradedContext`. It works for MAC, so I am curious if you think it should work for ZKP too. Carrying both context and validator was a little annoying and didn't seem to serve the purpose or add extra safety

akoshelev (approved):
Lets merge it ASAP and address any feedback in a follow up. I am working on preparing my MAC-validator ready and batcher interface is really what I am looking to integrate

ipa-core/src/protocol/basics/reveal.rs
andyleiserson: I was thinking that within the scope of a single record, there couldn't be anything too big, but better just to implement it for `BitDecomposed`.

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: That seems reasonable to me. I'm not aware of any reason it's more difficult for DZKP than for MAC. Hopefully it will also let us make validators unclonable again.

PR #1210, opened 2024-08-06
Vectorize malicious shares

akoshelev:
This change gets us closer to running maliciously secure PRF evaluation. As semi-honest protocols are vectorized at this point, malicious functionality needs to be upgraded to match it. This change does that and a few more things

* Vectorize MAC computations
* Vectorize malicious multiply
* Add a few helper methods (induce, expand, transform) to secret sharings

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/prf_eval.rs
+++ b/ipa-core/src/protocol/ipa_prf/prf_eval.rs
@@ -88,7 +88,7 @@ where
         ctx.narrow(&Step::GenRandomMask).prss().generate(record_id);
 
     //compute x+k
-    let mut y = x + AdditiveShare::<Fp25519>::expand(k);
+    let mut y = x + k.expand();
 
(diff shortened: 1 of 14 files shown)


ipa-core/src/protocol/ipa_prf/prf_eval.rs
```diff
@@ -88,7 +88,7 @@ where
         ctx.narrow(&Step::GenRandomMask).prss().generate(record_id);
 
     //compute x+k
-    let mut y = x + AdditiveShare::<Fp25519, N>::expand(k);
+    let mut y = x + AdditiveShare::<Fp25519>::expand(k);
```
andyleiserson, line 91: Maybe change this to `k.expand()`, unless this is in anticipation of something that's needed in the upcoming version that uses additional traits?

andyleiserson (approved):

PR #1211, opened 2024-08-06
Change Validator trait to use UpgradedContext

akoshelev:
This is a minor change that does not affect a lot of code. The reason for making it is to make trait bounds easier for the upcoming PRF eval upgrade where the validator bounds need to be expressed in a form of upgraded context.

It is also easier to reason about because at the time when validator is created, the `UpgradableContext` is gone


andyleiserson:
I did something slightly different in #1208 for the DZKP types -- I removed the `DZKPUpgradedContext` associated type on `trait UpgradableContext` completely and changed the type parameter on `trait DZKPValidator` to an associated type.

I'm not sure if there's a reason to prefer one or the other, but it's probably worth making the two variants the same unless there's a good reason they need to be different.

andyleiserson (approved):

PR #1212, opened 2024-08-06
Speed up dp tests

bmcase:
change parameters in `semi_honest_with_dp` to make faster. 

Move `gen_binomial_noise_16_breakdowns` to be a slow test.


akoshelev (approved):
thanks for making this change @bmcase ! It was taking more than 60 seconds for me to run `semi_honest_with_dp`, so I hope this helps

PR #1213, opened 2024-08-06
Use Upgradable trait in MAC module

akoshelev:
When working on vectorizing malicious shares,
we noticed that the existing upgrade API does not work for vectorized shares. It either required contexts to be vectorization-aware which we tried to avoid up to this point, or a different mechanism to specify how upgrades occur for vectorized shares.

For example: `Replicated<F, N>` when upgraded, gets transformed to `MaliciousReplicated<F, N>` and different `N` values, require different bounds on functions.

That was a blocker for PRF evaluation upgrade and the solution we came up with was to introduce `Upgradable` trait.

The functionality of this trait is very similar to `UpgradeToMalicious` and `Upgrade` but upgrade now is bound explicitly to `RecordId`. This optimizes for protocols readability. Malicious upgrades in tests had to find a way around this limitation, and it is a bit hacky. I don't think it is significantly different from what we've been doing before though.

### DIFF
--- a/ipa-core/src/protocol/basics/reshare.rs
+++ b/ipa-core/src/protocol/basics/reshare.rs
@@ -301,7 +301,7 @@ mod tests {
                         .malicious(a, |ctx, a| async move {
                             let v = ctx.validator();
                             let m_ctx = v.context().set_total_records(1);
-                            let m_a = a.upgrade(RecordId::FIRST, m_ctx.clone()).await.unwrap();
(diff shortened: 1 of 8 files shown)


andyleiserson (approved):
Not sure if it's worth the trouble to change this, but in most cases the argument order is `ctx, record_id` as opposed to `record_id, ctx`. Otherwise this seems fine.

PR #1216, opened 2024-08-07
Test encrypt decrypt

eriktaubeneck:
This gives us two new CLI commands for secret sharing and encrypting a raw input and then decrypting and reconstructing those encrypted files, respectively.

```
test_hpke encrypt --input-file tmp/raw_reports.txt --network tmp/keys/network.toml --output-dir tmp/enc
test_hpke decrypt --input-file1 tmp/enc/helper1.enc --input-file2 tmp/enc/helper2.enc --input-file3 tmp/enc/helper3.enc --mk-private-key1 tmp/keys/h1_mk.key --mk-private-key2 tmp/keys/h2_mk.key --mk-private-key3 tmp/keys/h3_mk.key --output-file tmp/output.dec
```

The `decrypt` command will allow us to test other implementations (e.g., in python or swift.)

I moved the bulk of the code into `ipa-core/src/cli/hpke.rs`, however the test doesn't seem to be run. I'm not sure how to fix that, or if I should move it somewhere else.


ipa-core/src/cli/hpke.rs
```diff
@@ -0,0 +1,337 @@
...
+    output_file: PathBuf,
+}
+
+/// # Panics
+/// if input file or network file are not correctly formatted
+/// # Errors
+/// if it cannot open the files
+pub fn encrypt(args: &EncryptArgs) -> Result<(), BoxError> {
+    let input = InputSource::from_file(&args.input_file);
+    let records = input.iter::<TestRawDataRecord>().collect::<Vec<_>>();
```
akoshelev, line 87:
```suggestion
    let shares: [Vec<OprfReport<BreakdownKey, TriggerValue, Timestamp>>; 3] = input.iter::<TestRawDataRecord>().share();

```

ipa-core/src/cli/hpke.rs
```diff
@@ -0,0 +1,337 @@
...
+                .write(true)
+                .create_new(true)
+                .open(args.output_dir.join(format!("helper{}.enc", index + 1)))?,
+        );
+        for share in shares {
+            let output = share.encrypt(key_id, key_registry, &mut rng).unwrap();
+            let hex_output = hex::encode(&output);
+            writeln!(writer, "{hex_output}")?;
+        }
+    }
```
akoshelev, line 112:
```suggestion
 for (index, (shares, key_registry)) in zip(shares, key_registries).enumerate() {
        let mut writer =
            OpenOptions::new()
                .write(true)
                .create_new(true)
                .open(args.output_dir.join(format!("helper{}.enc", index + 1)))?;

        for share in shares {
            let output = share.encrypt(key_id, key_registry, &mut rng)?;
            let hex_output = hex::encode(&output);
            writeln!(writer, "{hex_output}")?;
        }
    }
```

akoshelev: also consider handling file open error - I believe the default error does not mention file name, so it is also very unhelpful to debug

ipa-core/src/cli/hpke.rs
```diff
@@ -0,0 +1,337 @@
...
+    let key_registry = hpke_registry(mk_encryption.as_ref()).await?;
+    Ok(key_registry)
+}
+
+/// # Panics
+// if input files or private_keys are not correctly formatted
+/// # Errors
+/// if it cannot open the files
+pub async fn decrypt_and_reconstruct(args: DecryptArgs) -> Result<(), BoxError> {
+    let file1 = File::open(args.input_file1)?;
```
akoshelev, line 130: It is probably better to create a routine that opens a file and produces a stream/iterator over encrypted reports. Then, the main routine can just zip them together and reconstruct reports

ipa-core/src/cli/hpke.rs
```diff
@@ -0,0 +1,337 @@
...
+        boolean_array::{BA20, BA3, BA8},
+        U128Conversions,
+    },
+    hpke::{KeyRegistry, PrivateKeyOnly},
+    report::{EncryptedOprfReport, EventType, OprfReport, DEFAULT_KEY_ID},
+    secret_sharing::IntoShares,
+    test_fixture::{ipa::TestRawDataRecord, Reconstruct},
+};
+
+type BreakdownKey = BA8;
```
akoshelev, line 27: I had a feeling that these values are hardcoded somewhere else. It would be better if we use the same hardcoded types across the whole IPA stack, so we can fix/change them in one place

ipa-core/src/cli/mod.rs
```diff
@@ -1,6 +1,8 @@
 #[cfg(feature = "web-app")]
 mod clientconf;
 mod csv;
+#[cfg(all(feature = "test-fixture", feature = "web-app", feature = "cli"))]
+pub mod hpke;
```
akoshelev, line 5: this module name is confusing to me - I assumed that it provides the actual implementation of HPKE spec, which is not the case. Maybe it needs a better name

ipa-core/src/config.rs
```diff
@@ -414,6 +415,40 @@ impl Debug for Http2Configurator {
...
+#[derive(Default)]
+pub struct KeyRegistries(Vec<KeyRegistry<PublicKeyOnly>>);
+
+impl KeyRegistries {
+    /// # Panics
+    /// If network file is improperly formatted
+    pub fn init_from(
+        &mut self,
+        network: &NetworkConfig,
+        key_id: KeyIdentifier,
```
akoshelev, line 427: is there a need to take `key_id` only to return it back sometimes?

ipa-core/src/config.rs
```diff
@@ -414,6 +415,40 @@ impl Debug for Http2Configurator {
...
+            } else {
+                None
+            }
+        })?;
+
+        // Create key registries
+        self.0 = configs
+            .into_iter()
+            .map(|hpke| KeyRegistry::from_keys([PublicKeyOnly(hpke.public_key.clone())]))
+            .collect::<Vec<KeyRegistry<PublicKeyOnly>>>();
```
akoshelev, line 443: try this

```rust
 let configs = network.peers.each_ref()
            .map(|peer_config| Some(PublicKeyOnly(peer_config.hpke_config.clone()?.public_key)))
            .map(|pk| Some(KeyRegistry::from_keys([pk?])))
            .into_iter()
            .collect::<Option<Vec<_>>>()?
            .try_into()
            .ok()?;
```

ipa-core/src/config.rs
```diff
@@ -414,6 +415,40 @@ impl Debug for Http2Configurator {
...
 }
 
+#[derive(Default)]
+pub struct KeyRegistries(Vec<KeyRegistry<PublicKeyOnly>>);
+
+impl KeyRegistries {
+    /// # Panics
+    /// If network file is improperly formatted
+    pub fn init_from(
+        &mut self,
```
akoshelev, line 425: you don't mutate the state of self, this seems to be a static initializer instead

ipa-core/src/cli/hpke.rs
eriktaubeneck: yea, that makes sense. they weren't public where I found them, but I can make them public.

ipa-core/src/cli/mod.rs
eriktaubeneck: that's fair. `hpke_util`?

akoshelev: `crypto` maybe? HPKE may give false suspicions to readers that it is somehow related to the existing `hpke` module, but there is no HPKE-specific functionality here, just encrypt/decrypt

ipa-core/src/config.rs
eriktaubeneck: this is copied almost exactly from [up here](https://github.com/private-attribution/ipa/pull/1216/files#diff-b2bf3717000008c1caa689282081c4c56fea43a9e4da1d68dd01c4825a0b2d98L178-L205). I added the `key_id` (because it was previously returning a hardcoded DEFAULT_KEY_ID), but it can probably go.

ipa-core/Cargo.toml
```diff
@@ -193,6 +193,11 @@ name = "test_mpc"
 required-features = ["cli", "test-fixture", "web-app", "weak-field"]
 bench = false
 
+[[bin]]
+name = "crypto_util"
+required-features = ["cli", "test-fixture", "web-app"]
```
akoshelev, line 198:
```suggestion
required-features = ["cli", "test-fixture"]
```

akoshelev: I don't think you need `web-app` feature

akoshelev: probably `test-fixture` is redundant too

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+/// if it cannot open the files
+pub async fn decrypt_and_reconstruct(args: DecryptArgs) -> Result<(), BoxError> {
+    let key_registry1 = build_hpke_registry(args.mk_private_key1).await?;
+    let key_registry2 = build_hpke_registry(args.mk_private_key2).await?;
+    let key_registry3 = build_hpke_registry(args.mk_private_key3).await?;
+    let decrypted_reports1 = DecryptedReports::new(&args.input_file1, key_registry1);
+    let decrypted_reports2 = DecryptedReports::new(&args.input_file2, key_registry2);
+    let decrypted_reports3 = DecryptedReports::new(&args.input_file3, key_registry3);
+
+    let mut writer: Box<dyn Write> = Box::new(
```
akoshelev, line 178:
```suggestion
    let mut writer = 
```

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+        let match_key = [
+            dec_report1.match_key,
+            dec_report2.match_key,
+            dec_report3.match_key,
+        ]
+        .reconstruct()
+        .as_u128();
+
+        assert_eq!(dec_report1.event_type, dec_report2.event_type);
+        assert_eq!(dec_report2.event_type, dec_report3.event_type);
```
akoshelev, line 205:
```suggestion
```

you don't need to validate, `reconstruct` will [panic](https://github.com/private-attribution/ipa/blob/7e894c36a453d7918829c28da2ccddfab6553f19/ipa-core/src/test_fixture/sharing.rs#L55) if shares don't match.

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+        .as_u128();
+
+        writeln!(
+            writer,
+            "{},{},{},{},{}",
+            timestamp.try_into().unwrap_or(u64::MAX),
+            match_key.try_into().unwrap_or(u64::MAX),
+            u8::from(is_trigger_report),
+            breakdown_key.try_into().unwrap_or(u32::MAX),
+            trigger_value.try_into().unwrap_or(u32::MAX)
```
akoshelev, line 231:
```suggestion
            timestamp,
            match_key,
            u8::from(is_trigger_report),
            breakdown_key,
            trigger_value,
```

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+        for line1 in reader1 {
+            let line2 = reader2.next().expect("Files have different lengths");
+            assert_eq!(line1.unwrap(), line2.unwrap());
+        }
+        assert!(reader2.next().is_none(), "Files have different lengths");
+    }
+
+    #[tokio::test]
+    async fn encrypt_and_decrypt() {
+        let count: u32 = 10;
```
akoshelev, line 272:
```suggestion
        let count = 10;
```

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+        let rng = thread_rng();
+        let event_gen_args = EventGeneratorConfig::new(10, 5, 20, 1, 10, 604_800);
+
+        let event_gen: Vec<TestRawDataRecord> = EventGenerator::with_config(rng, event_gen_args)
+            .take(count as usize)
```
akoshelev, line 277:
```suggestion
            .take(count)
```

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+        assert!(reader2.next().is_none(), "Files have different lengths");
+    }
+
+    #[tokio::test]
+    async fn encrypt_and_decrypt() {
+        let count: u32 = 10;
+        let rng = thread_rng();
+        let event_gen_args = EventGeneratorConfig::new(10, 5, 20, 1, 10, 604_800);
+
+        let event_gen: Vec<TestRawDataRecord> = EventGenerator::with_config(rng, event_gen_args)
```
akoshelev, line 276:
```suggestion
        let event_gen = EventGenerator::with_config(rng, event_gen_args)
```

ipa-core/src/cli/crypto.rs
```diff
@@ -0,0 +1,359 @@
...
+            .take(count as usize)
+            .collect::<Vec<_>>();
+        let mut raw_input = NamedTempFile::new().unwrap();
+
+        for event in event_gen {
+            let _ = event.to_csv(raw_input.as_file_mut());
+            writeln!(raw_input.as_file_mut()).unwrap();
```
akoshelev, line 283:
```suggestion
            writeln!(raw_input.as_file()).unwrap();
```

akoshelev (approved):
minor things, generally good to go

ipa-core/Cargo.toml
eriktaubeneck: just tested and removing `web-app` is fine. `test-fixture` is used here:

```
error[E0433]: failed to resolve: could not find `test_fixture` in the crate root
  --> ipa-core/src/cli/crypto.rs:23:5
   |
23 |     test_fixture::{ipa::TestRawDataRecord, Reconstruct},
   |     ^^^^^^^^^^^^ could not find `test_fixture` in the crate root
   ```

eriktaubeneck: oh, looks like I'm wrong. I use `cli::playbook` , and that relies on `web-app`.

ipa-core/src/cli/crypto.rs
eriktaubeneck: these don't come from `reconstruct` though

ipa-core/Cargo.toml
akoshelev: heh, yea - these dependencies are not really `web-app` specific, but it is not the purpose of this PR to fix them.

ipa-core/src/cli/crypto.rs
eriktaubeneck: adding a comment in the code to make it more clear why this is here.

PR #1217, opened 2024-08-08
In market test runbook

eriktaubeneck:
This is the run book we'll use for the in-market test. There's a readme with all the steps, and a handful of ansible scripts that get it up and running.

I spun up 3 AWS servers, ran these scripts as described in the README, and then started a query from my machine to test that it works (and it does.)

However, I also added 1 Google Cloud server to the mix, and it seems to stall at the very end. The AWS run takes about 1m for 1000 rows (I'm not using big machines). After 5-6 minutes, the gcloud HPN gives the following error:

```
ERROR ipa_core::net::server: connection error: peer is incompatible: SignatureAlgorithmsExtensionRequired
```
or on another run, I received
```
ERROR ipa_core::net::server: connection error: tls handshake eof\n
```


in-market-test/v2/README.md
```diff
@@ -0,0 +1,110 @@
...
+Host ipa-helper
+  HostName <ipa-helper-ip-or-public-dns-name>
+  User <associated-user>
+  IdentityFile ~/.ssh/helper_connect.pem
+```
+
+Now, copy the inventory template
+
+```
+cp in-market-test/v2/ansible/inventory-template.ini in-market-test/v2/ansible/inventory.ini
```
akoshelev, line 48: it is better to keep the templates and actual files used to deploy in different folders I think. Makes it easier to clean up and avoid committing local state. `ansible/templates/...` and `ansible/...` may be one way to do it

in-market-test/v2/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+[myhosts:vars]
+ansible_python_interpreter=/usr/bin/python3
+```
+
+### Install dependencies and compile IPA
+
+This command installs dependencies and compiles IPA. Note: it uses `yum`. If you need to use a different package manager, you can update the file.
+```
+ansible-playbook -i in-market-test/v2/ansible/inventory.ini in-market-test/v2/ansible/provision.yaml
```
akoshelev, line 64: is there a way to instruct ansible to use the default value for `-i`? if no, maybe setting an environment variable that points to the runbook makes this instruction easier to maintain

in-market-test/v2/README.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+### Generate Keys
+
+```
+ansible-playbook -i in-market-test/v2/ansible/inventory.ini in-market-test/v2/ansible/gen_keys.yaml
+```
+This should download `cert.pem` and `mk.pub` locally.
+
+Warning: After sharing the public keys with other helpers (in the next step), you'll want to avoid rerunning this command. However, the are generated in the form of `<timestamp>-cert.pem` and `<timestamp>-mk.pub`, to prevent overwrite if you happen to run this command again. On the machine, it makes a copy of these timestamps into a fixed location (`in-market-test/v2/deployed_keys/cert.pem`, ...), which are used below to start the helper. If they are accidentally overwritten, you'll need to manually copy the correct ones into the expected location.
```
akoshelev, line 74:
```suggestion
Warning: After sharing the public keys with other helpers (in the next step), you'll want to avoid rerunning this command. However, they are generated in the form of `<timestamp>-cert.pem` and `<timestamp>-mk.pub`, to prevent overwrite if you happen to run this command again. On the machine, it makes a copy of these timestamps into a fixed location (`in-market-test/v2/deployed_keys/cert.pem`, ...), which are used below to start the helper. If they are accidentally overwritten, you'll need to manually copy the correct ones into the expected location.
```

in-market-test/v2/ansible/network-template.toml
```diff
@@ -0,0 +1,60 @@
+[[peers]]
+certificate = """
+-----BEGIN CERTIFICATE-----
+MIIBgDCCASagAwIBAgIIJIo0r8LAFEQwCgYIKoZIzj0EAwIwHTEbMBkGA1UEAwwS
```
akoshelev, line 4: can we make those broken so we don't accidentally run these in prod?

akoshelev (approved):

in-market-test/v2/README.md
eriktaubeneck: [default value](https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html) is `/etc/ansible/hosts`, and I'd prefer to keep it project specific. given most of these are just going to be copy and pasted, an environmental variable just seems like more room for something to get messed up. with this as is, you can copy and paste it without the need for anything else.

in-market-test/v2/ansible/network-template.toml
eriktaubeneck: they should be broken because they are tied to local `draft.test` domains, but I can make that even more explicit but putting in a line at the top that says "NOT A REAL CERT" and mixing up some of the data further.

in-market-test/v2/README.md
eriktaubeneck: i'm going to keep this as is, but if you feel strongly let's sync offline

PR #1221, opened 2024-08-12
Upgrade PRF evaluation circuit to active security

akoshelev:
Use validate_record before the reveal and write a test that demonstrates that malicious errors can be caught


ipa-core/src/protocol/basics/mod.rs
```diff
@@ -41,7 +42,7 @@ use crate::{
 /// difficulty of resolving `V` vs. `[V; 1]` issues for the known value type. `Reshare` hasn't been
 /// attempted.)
 pub trait BasicProtocols<C: Context, V: SharedValue + Vectorizable<N>, const N: usize = 1>:
-    SecretSharing<V> + Reveal<C, Output = <V as Vectorizable<N>>::Array> + SecureMul<C> + FromPrss
+    SecretSharing<V> + Reveal<C, Output = <V as Vectorizable<N>>::Array> + SecureMul<C>
```
akoshelev, line 45: @andyleiserson `FromPrss` bound cannot be fulfilled for MAC-malicious shares, so I had to remove it. Does not seem that it breaks anything, but let me know if it may stay in your way of adding some other functionality

ipa-core/src/protocol/ipa_prf/prf_eval.rs
```diff
@@ -74,39 +97,45 @@ where
...
+    // validate everything before reveal
+    ctx.validate_record(record_id).await?;
+    let (gr, z): (
+        <RP25519 as Vectorizable<N>>::Array,
+        <Fp25519 as Vectorizable<N>>::Array,
+    ) = try_join(
+        // TODO: these should invoke reveal via the trait when this function
+        // takes a context of an appropriate type.
+        malicious_reveal(ctx.narrow(&Step::RevealR), record_id, None, &sh_gr)
+            .map(|v| v.map(|arr| arr.unwrap())),
```
andyleiserson, line 135: You will need the changes in #1229 for this to work (yet another piece of #1112), but I think it would be cleaner and more correct to do:
```suggestion
        reveal(ctx.narrow(&Step::RevealR), record_id, &sh_gr),
```
The behavior will be different when running against a semi-honest context. It's not harmful to do the malicious reveal when the context is semi-honest, but I think it's better to follow the principle of not invoking primitives (`semi_honest_reveal` / `malicious_reveal`) directly from high-level protocols that use a generic context. (The opposite case, invoking `semi_honest_reveal`, would be harmful, since it may forgo a malicious reveal when one is called for.)

ipa-core/src/protocol/ipa_prf/prf_eval.rs
```diff
@@ -118,16 +147,52 @@ where
...
+    /// Propagates errors from multiplications
+    pub async fn compute_match_key_pseudonym<C>(
+        ctx: C,
+        prf_key: AdditiveShare<Fp25519>,
+        input_match_keys: Vec<AdditiveShare<Fp25519>>,
+    ) -> Result<Vec<u64>, Error>
+    where
+        C: UpgradableContext,
+        AdditiveShare<Fp25519>:
+            PrfSharing<<C::Validator<Fp25519> as Validator<Fp25519>>::Context, 1, Field = Fp25519>,
```
andyleiserson, line 183:
```suggestion
        AdditiveShare<Fp25519>: PrfSharing<MacUpgraded<C, Fp25519>, 1, Field = Fp25519>,
```

andyleiserson (approved):

ipa-core/src/protocol/basics/mod.rs
```diff
@@ -57,6 +58,14 @@ impl<'a, B: ShardBinding>
 {
 }
 
+impl<'a, const N: usize> BasicProtocols<UpgradedMaliciousContext<'a, Fp25519>, Fp25519, N>
+    for malicious::AdditiveShare<Fp25519, N>
+where
+    Fp25519: FieldSimd<N>,
+    AdditiveShare<Fp25519, N>: FromPrss,
```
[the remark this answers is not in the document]
andyleiserson: I see, it's needed so that the malicious share is `SecureMul`.

PR #1225, opened 2024-08-15
DP Padding

bmcase:
PR to add DP padding for both OPRF and the new aggregation. 

Right now the code sets parameters internally to `PaddingParameters::default()` which are moderate DP parameters.  If we would like to set the parameters at run time we can add to the query config in a future PR. 

Additional parameter sets are used for some tests: 

- `PaddingParameters::relaxed()` which is needed to make sure the circuit stays very small so concurrency tests and explore many execution paths.  
-  `PaddingParameters::no_padding()` which is used in some tests where no DP for outputs is being applied.


bmcase:
With draft: 
- [A Draft run](https://draft-mpc.vercel.app/query/view/rust-salad2024-08-15T1717) with `default` padding parameters and 1k input rows, had just under 10k fake rows added.  Ran in 14sec. 
- This compares similarly to a [draft run](https://draft-mpc.vercel.app/query/view/tepid-hour2024-08-15T1732) on main with 11k input rows also taking 14sec.

However, the added rows by default seem to be majorly slowing down the CI tests, so will need to scale that back so they don't take so long. 

Also created a table of parameters (in the test code) and printed the [results here](https://docs.google.com/spreadsheets/d/1N0WEUkarP_6nd-7W8O9r-Xurh9OImESgAC1Jd_6OfWw/edit?gid=0#gid=0) for expected number of dummies for different parameter sets.

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -410,14 +426,21 @@ pub mod tests {
                 test_input(10, 12345, true, 0, 5),
                 test_input(0, 68362, false, 1, 0),
                 test_input(20, 68362, true, 0, 2),
-            ];
+            ]; // trigger value of 2 attributes to earlier source row with breakdown 1.
```
benjaminsavage, line 429: Do we really need to add this comment? If so, why not add another one above for the other one, which attributes to a source row with breakdown 2?

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -426,12 +449,22 @@ pub mod tests {
...
                 EXPECTED,
             );
+
+            for (index, actual) in result.iter().enumerate() {
+                println!(
+                    "actual = {}, expected = {}",
+                    u32::try_from(actual.as_u128()).unwrap(),
+                    EXPECTED[index]
+                );
+            }
```
benjaminsavage, line 459: Is this leftover debugging code?

ipa-core/src/protocol/ipa_prf/oprf_padding/README.md
```diff
@@ -56,7 +58,27 @@ $Y=\theta + X_1 - X_2$
...
 
-<!-- The variance of a double geometric is the sum of the variances of the two independent geometrics, $X_1$ and $X_2$, so is $2 (\frac{1-p}{p^2})$ -->
+The variance of a double geometric is the sum of the variances of the two independent geometrics, $X_1$ and $X_2$, so is $2 * (\frac{1-p}{p^2})$
 
 ### Samples from the Truncated Double Geometric Distribution
 Once we can draw samples from a double geometric, we can sample from our desired truncated double geometric by sampling the double geometric with rejection if the sample lies outside the support set $\{0,...,2n\}$.
+
+The variance of a truncated double geometric distribution is (TODO), but the variance is always less than the variance of the underlying (non-truncated) double geometric distribution.
+
+# Padding Breakdowns Keys for New Aggregation
```
benjaminsavage, line 68: instead of "New" aggregation (a name which will rapidly go out of date), maybe we call it "reveal based aggregation"

ipa-core/tests/common/mod.rs
```diff
@@ -93,12 +93,18 @@ pub trait CommandExt {
...
+    // the binaries. e.g. when running: `cargo test --test compact_gate --lib
+    // compact_gate_cap_8_no_window_semi_honest -p ipa-core --no-default-features
+    // --features "cli web-app real-world-infra test-fixture compact-gate"`
     fn silent(&mut self) -> &mut Self {
         if std::env::var("VERBOSE").ok().is_none() {
             self.arg("--quiet")
         } else {
             self.arg("-vv")
         }
+        // return self;
```
benjaminsavage, line 107: Do you want to leave this in, or was this leftover from debugging?

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+            &mut padding_input_rows,
+            h_i,
+            h_i_plus_one,
+            padding_params,
+        )?;
+    }
+
+    // Step 2: h_i and h_i_plus_one will send the send total_number_of_fake_rows to h_out
+    let send_ctx = ctx
+        .narrow(&SendTotalRows::SendFakeNumRecords)
```
benjaminsavage, line 220: Can we call is "NumFakeRecords" instead of "FakeNumRecords"?

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+        total_number_of_fake_rows = two_parties_add_dummies::<C, BK, TV, TS, B>(
+            &ctx,
+            &mut padding_input_rows,
+            h_i,
+            h_i_plus_one,
+            padding_params,
+        )?;
+    }
+
+    // Step 2: h_i and h_i_plus_one will send the send total_number_of_fake_rows to h_out
```
benjaminsavage, line 218: This constitutes *some* kind of leakage, even if it's not concerning. It would be cool to have some kind of a comment either explaining why this is OK, or pointing to some resource explaining why this is OK.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+    }
+    if ctx.role() == h_i_plus_one {
+        let send_channel = send_ctx.send_channel::<BA32>(send_ctx.role().peer(Direction::Right));
+        let _ = send_channel
+            .send(
+                RecordId::FIRST,
+                BA32::truncate_from(u128::from(total_number_of_fake_rows)),
+            )
+            .await;
+    }
```
benjaminsavage, line 239: Can you combine these into a single code block? This is very repetitive. And maybe add a comment about the rationale for this (redundancy in case of a malicious helper)

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+            Ok(v) => u32::try_from(v.as_u128()).unwrap(),
+            Err(e) => return Err(e.into()),
+        };
+
+        let recv_channel_left =
+            send_ctx.recv_channel::<BA32>(send_ctx.role().peer(Direction::Left));
+        let from_left = match recv_channel_left.receive(RecordId::FIRST).await {
+            Ok(v) => u32::try_from(v.as_u128()).unwrap(),
+            Err(e) => return Err(e.into()),
+        };
```
benjaminsavage, line 254: I think you can do these in parallel using try_join.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+        assert_eq!(from_right, from_left);
+        total_number_of_fake_rows = from_right;
+    }
+
+    // Step 3: `h_out` will generate secret shares of zero for as many rows as the `total_number_of_fake_rows`
```
benjaminsavage, line 259: I don't like calling these "secret shares of zero", since they actually are not. These are just 2 out of 3 shares. Can we update the comment to say something like "h_out will set its shares to zero for the fake rows".

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+    let mut total_number_of_fake_rows = 0;
+    let (mut left, mut right) = ctx.prss_rng();
+    // The first is shared with the helper to the "left", the second is shared with the helper to the "right".
+    let mut rng = &mut right;
+    if ctx.role() == h_i {
+        rng = &mut right;
+    }
+    if ctx.role() == h_i_plus_one {
+        rng = &mut left;
+    }
```
benjaminsavage, line 303: Can you use a match statement for this, and panic if ctx.role() is h_out?

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+                for _ in 0..sample {
+                    let dummy_mk: BA64 = rng.gen();
+                    for _ in 0..cardinality {
+                        let mut match_key_shares: Replicated<BA64> = Replicated::default();
+                        if ctx.role() == h_i {
+                            match_key_shares = Replicated::new(BA64::ZERO, dummy_mk);
+                        }
+                        if ctx.role() == h_i_plus_one {
+                            match_key_shares = Replicated::new(dummy_mk, BA64::ZERO);
+                        }
```
benjaminsavage, line 331: Again, I think you can potentially avoid mutability with a match expression.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+            aggregation_delta,
+            aggregation_padding_sensitivity,
+        } => {
+            let aggregation_padding = OPRFPaddingDp::new(
+                aggregation_epsilon,
+                aggregation_delta,
+                aggregation_padding_sensitivity,
+            )?;
+            let num_breakdowns: u32 = u32::try_from(B).unwrap();
+            // // for every breakdown, sample how many dummies will be added
```
benjaminsavage, line 360: extra //

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+                total_number_of_fake_rows += sample;
+
+                // now add `sample` many fake rows with this `breakdownkey`
+                // These fake rows need to have random matchkeys, but we don't want to have any with
+                // cardinality 1 or they may be dropped after matching.  So instead we add with the
+                // following cardinalities:
+                //  - in the case `sample` is even, all matchkeys have cardinality 2
+                //  - in the case `sample` is odd, all matchkeys have cardinality 2 except one set
+                //    of cardinality 3.
+                //  - in the case `sample` = 1, we add a matchkey of cardinality 1.
```
benjaminsavage, line 372: I don't think we should add these dummy records before attribution. I think we should add these at the last possible minute, so that we don't even have to think about this sort of stuff.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,814 @@
...
+
+    let mut breakdownkey_shares: Replicated<BK> = Replicated::default();
+    if ctx.role() == h_i {
+        breakdownkey_shares =
+            Replicated::new(BK::ZERO, BK::truncate_from(u128::from(breakdownkey)));
+    }
+    if ctx.role() == h_i_plus_one {
+        breakdownkey_shares =
+            Replicated::new(BK::truncate_from(u128::from(breakdownkey)), BK::ZERO);
+    }
```
benjaminsavage, line 436: I really think match statements and avoiding mutability it better.

benjaminsavage (approved):
Looks good to be aside from the tests failing.

ipa-core/src/protocol/ipa_prf/mod.rs
bmcase: I think it could be helpful, especially for folks still getting familiar with how IPA works overall.   I'll extend the comment to cover the other attributed pair.

ipa-core/src/protocol/ipa_prf/mod.rs
bmcase: I think so; I'll delete it.

ipa-core/src/protocol/ipa_prf/oprf_padding/README.md
bmcase: good point, changing to "reveal based aggregation"

ipa-core/tests/common/mod.rs
bmcase: Yes, it is left from debugging, but I think it is worth leaving it together with the added documentation for this function.  This was a trick I learned from Alex for debugging tests that run the compiled binaries, but I think we should somewhere makes how to do this better known.  Probably not just here but in some developer documentation.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: Okay, will point to the `oprf_padding/README.md` and make sure it is more fully explained there.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: Okay, refactored a bit to remove some duplication.  Also added comment about why we check the value received from both other parties. 
```
    match ctx.role() {
        role if role == h_i || role == h_i_plus_one => {
            let direction = if role == h_i {
                Direction::Left
            } else {
                Direction::Right
            };
            let send_channel = send_ctx.send_channel::<BA32>(send_ctx.role().peer(direction));
            let _ = send_channel
                .send(
                    RecordId::FIRST,
                    BA32::truncate_from(u128::from(total_number_of_fake_rows)),
                )
                .await;
        }
        _h_out => {
            let recv_channel_right =
                send_ctx.recv_channel::<BA32>(send_ctx.role().peer(Direction::Right));
            let from_right = match recv_channel_right.receive(RecordId::FIRST).await {
                Ok(v) => u32::try_from(v.as_u128()).unwrap(),
                Err(e) => return Err(e.into()),
            };
            let recv_channel_left =
                send_ctx.recv_channel::<BA32>(send_ctx.role().peer(Direction::Left));
            let from_left = match recv_channel_left.receive(RecordId::FIRST).await {
                Ok(v) => u32::try_from(v.as_u128()).unwrap(),
                Err(e) => return Err(e.into()),
            };
            assert_eq!(from_right, from_left);
            total_number_of_fake_rows = from_right;
        }
    }
    ```

File: ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs (hunk 8 above)
Comment (bmcase) [reply ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs:254]: Okay, refactored it to use `try_join` as follows: 
```
            let recv_channel_right =
                send_ctx.recv_channel::<BA32>(send_ctx.role().peer(Direction::Right));
            let recv_channel_left =
                send_ctx.recv_channel::<BA32>(send_ctx.role().peer(Direction::Left));
            let (from_right, from_left) = try_join!(
                async {
                    match recv_channel_right.receive(RecordId::FIRST).await {
                        Ok(v) => Ok::<u32, Error>(u32::try_from(v.as_u128()).unwrap()),
                        Err(e) => Err(e.into()),
                    }
                },
                async {
                    match recv_channel_left.receive(RecordId::FIRST).await {
                        Ok(v) => Ok::<u32, Error>(u32::try_from(v.as_u128()).unwrap()),
                        Err(e) => Err(e.into()),
                    }
                }
            )?;

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: yes, that's more clear

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: Not sure we can. h_out isn't in scope in this function.  I can add an extra assert that `assert!(h_i != h_i_plus_one);`

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: with h_i and h_i_plus_one variables, it doesn't seem to want to match as follows, saying the first one might "match any value"

```
                        let match_key_shares: Replicated<BA64> = match ctx.role() {
                            _h_i => Replicated::new(BA64::ZERO, dummy_mk),
                            _h_i_plus_one => Replicated::new(dummy_mk, BA64::ZERO),
                            _ => Replicated::default(),
                        };

                        let row = OPRFIPAInputRow {
                            match_key: match_key_shares,
                            is_trigger: Replicated::new(Boolean::FALSE, Boolean::FALSE),
                            breakdown_key: Replicated::new(BK::ZERO, BK::ZERO),
                            trigger_value: Replicated::new(TV::ZERO, TV::ZERO),
                            timestamp: Replicated::new(TS::ZERO, TS::ZERO),
                        };
                        
error: unreachable pattern
   --> ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs:349:29
    |
348 | ...                   _h_i => Replicated::new(BA64::ZERO, dummy_mk),
    |                       ---- matches any value
349 | ...                   _h_i_plus_one => Replicated::new(dummy_mk, BA64::ZERO),
    |                       ^^^^^^^^^^^^^ unreachable pattern

**Comment by bmcase** [comment]
Working on updating the approach for DP padding to add separate batches of dummies for OPRF and then right before Breakdown Reveal Agg.  To do so working on introducing a Paddable trait so we can reuse as much of the code in either case.  Have this doc to summarize the approach, but having issues with trait bounds.   @benjaminsavage  can you take a look at this new direction? Latest code on the PR but this doc summarizes.   https://docs.google.com/document/d/1heIeBE1lMSnmRGAUoopmDuOnKLkR1-_CA9EpqcbPdug/edit#heading=h.f83tb89xflfw

File: ipa-core/src/protocol/ipa_prf/oprf_padding/insecure.rs
Comment (andyleiserson) [inline ipa-core/src/protocol/ipa_prf/oprf_padding/insecure.rs:1]: Maybe `OPRFPaddingDp` and `Error` should move out of this file, if they're used for the secure mode of operation? (It may make sense to do that as a follow-up, though.)

File: ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs [hunk 15]
```diff
@@ -1,5 +1,866 @@
...
+    let send_ctx = ctx
+        .narrow(&SendTotalRows::SendNumFakeRecords)
+        .set_total_records(TotalRecords::ONE);
+    match ctx.role() {
+        role if role == h_i || role == h_i_plus_one => {
+            let direction = if role == h_i {
+                Direction::Left
+            } else {
+                Direction::Right
+            };
```
andyleiserson, line 409: I'd suggest adding the following helper to `impl Role`:

```rust
    /// Returns the direction to the peer with the specified role.
    ///
    /// If `self == role`, returns `None`.
    #[must_use]
    pub const fn direction_to(&self, role: Role) -> Option<Direction> {
        match (self, role) {
            (H1, H2) | (H2, H3) | (H3, H1) => Some(Direction::Right),
            (H1, H3) | (H2, H1) | (H3, H2) => Some(Direction::Left),
            (H1, H1) | (H2, H2) | (H3, H3) => None,
        }
    }
```

Then, you can:
1. Replace lines 401-407 with:
```rust
    if let Some(direction) = ctx.role().direction_to(h_out)
```
2. Change the `_h_out` branch to an `else`, like we talked about this morning.
3. Change `two_parties_add_dummies` and `add_padding_items` to take a `Direction` rather than `Role`s.

It may also make sense to move steps 1 and 3 into the if/else cases.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: Updated with this approach.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,866 @@
...
+                },
+                async {
+                    match recv_channel_left.receive(RecordId::FIRST).await {
+                        Ok(v) => Ok::<u32, Error>(u32::try_from(v.as_u128()).unwrap()),
+                        Err(e) => Err(e.into()),
+                    }
+                }
+            )?;
+
+            assert_eq!(from_right, from_left);
```
andyleiserson, line 438: An assert will crash the server, so the rule of thumb for asserts is that they shouldn't be reachable based on external input. It would be better to propagate this as an error, maybe `Error::InconsistentShares` (which should probably have a member to provide some context about what shares where inconsistent, maybe the name of the function throwing it.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,866 @@
...
+                                match_key_shares = AdditiveShare::new(dummy_mk, BA64::ZERO);
+                            }
+                            let row = OPRFIPAInputRow {
+                                match_key: match_key_shares,
+                                is_trigger: AdditiveShare::new(Boolean::FALSE, Boolean::FALSE),
+                                breakdown_key: AdditiveShare::new(BK::ZERO, BK::ZERO),
+                                trigger_value: AdditiveShare::new(TV::ZERO, TV::ZERO),
+                                timestamp: AdditiveShare::new(TS::ZERO, TS::ZERO),
+                            };
+                            padding_input_rows.extend(std::iter::once(row));
```
andyleiserson, line 187: You can use `Extend::extend_one` for adding a single element.

ipa-core/src/test_fixture/ipa.rs
```diff
@@ -225,19 +228,19 @@ pub async fn test_oprf_ipa<F>(
...
-                    8 => oprf_ipa::<BA8, BA3, BA32, BA20, 3, 256>(ctx, input_rows, aws, dp_params)
+                    8 => oprf_ipa::<BA8, BA3, BA32, BA20, 3, 256>(ctx, input_rows, aws, dp_params, padding_params)
                     .await
                     .unwrap(),
-                    16 => oprf_ipa::<BA8, BA3, BA32, BA20, 4, 256>(ctx, input_rows, aws, dp_params)
+                    16 => oprf_ipa::<BA8, BA3, BA32, BA20, 4, 256>(ctx, input_rows, aws, dp_params, padding_params)
                     .await
                     .unwrap(),
-                    32 => oprf_ipa::<BA8, BA3, BA32, BA20, 5, 256>(ctx, input_rows, aws, dp_params)
+                    32 => oprf_ipa::<BA8, BA3, BA32, BA20, 5, 256>(ctx, input_rows, aws, dp_params,padding_params)
```
andyleiserson, line 237: Is this in a macro? I'm surprised rustfmt didn't clean this up, but there is a missing space before `padding_params`.

ipa-core/src/protocol/ipa_prf/oprf_padding/distributions.rs
```diff
@@ -139,7 +139,7 @@ impl Distribution<i32> for DoubleGeometric {
 /// Truncated Double Geometric distribution.
 #[derive(Debug, PartialEq)]
 pub struct TruncatedDoubleGeometric {
-    shift_doubled: u32, // move 2 * shift to constructor instead of sample
+    pub shift_doubled: u32, // move 2 * shift to constructor instead of sample
```
andyleiserson, line 142: Would it be worth providing a method like `TruncatedDoubleGeometric::mean` so this doesn't need to be public? Or perhaps even if the parameters are public, to encapsulate code related to the distribution? It does seem reasonable for the distribution parameters to be public, but if they are it might make more sense to do so consistently for all of them.

bmcase:
Re @andyleiserson got email of comment but can't find to reply inline.  
>On [ipa-core/src/protocol/ipa_prf/oprf_padding/insecure.rs](https://github.com/private-attribution/ipa/pull/1225#discussion_r1727763810):

>Maybe OPRFPaddingDp and Error should move out of this file, if they're used for the secure mode of operation? (It may make sense to do that as a follow-up, though.)

Yes, I agree it may make sense to move some code around and do some renaming.  Would probably put moving into a new PR to make it easier to review and would do a few things
1. rename the folder `oprf_padding` to something like `padding_dp` and move out of `ipa_prf` since also used for aggregation.  
2. I'm not surer why `insecure.rs` is called that, but would move the Gaussian code out of it and into `dp` and move the `OPRFPaddingDP` code into `distributions.rs`.  
3. Will also rename `OPRFPaddingDP` to `PaddingDP` since used for both OPRF and Breakdown Reveal Agg.

ipa-core/src/protocol/ipa_prf/oprf_padding/distributions.rs
bmcase: Yes, I made the function `mean_and_std_bound` in `insecure.rs` to do this.  I think that was when I needed to make this parameter public.   I think moving some code around is a good follow up step and there would move lots much of this code from `insecure.rs` into this `distributions.rs` file and then also not need to make this param public.

andyleiserson:
> Re @andyleiserson got email of comment but can't find to reply inline.

Sorry about that. I accidentally clicked the button to post a single comment rather than start a review. Then I deleted the comment and [re-posted it as part of the review](https://github.com/private-attribution/ipa/pull/1225#discussion_r1727768338).

The code rearrangements you list make sense to me, and I agree it would be better to do in a separate PR.

ipa-core/src/protocol/ipa_prf/oprf_padding/insecure.rs
```diff
@@ -42,6 +45,31 @@ impl From<BernoulliError> for Error {
     }
 }
 
+impl From<Error> for error::Error {
```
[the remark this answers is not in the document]
bmcase: No, I was hoping there was a better way to do this. This is great; let me try this.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: okay, that's nice to simplify some of the role checking in `apply_dp_padding_pass`.   I'm not sure I follow how to only use a `Direction` in `add_padding_item` as will need to compare with `ctx.role()`

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: okay, created an `Error::InconsistentPadding` to return here.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
bmcase: I get an error about `extend_one` being an unstable library feature.

andyleiserson: Yup, I missed that, so disregard this suggestion.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
andyleiserson: What I was thinking is replace the tests against `ctx.role()` with:

In `two_parties_add_dummies`:
```rust
// If the excluded helper is on our left, we want to use the PRSS shared with the
// helper on our right, and vice-versa.
let (mut left, mut right) = ctx.prss_rng();
let rng = match direction_to_excluded_helper {
    Direction::Left => &mut right,
    Direction::Right => &mut left,
};
```

In `add_padding_items` for `OPRFIPAInputRow`:
```rust
let match_key_shares = match direction_to_excluded_helper {
    Direction::Left => AdditiveShare::new(BA64::ZERO, dummy_mk),
    Direction::Right => AdditiveShare::new(dummy_mk, BA64::ZERO),
};
```

... and similarly in the other `add_padding_items` implementation. This initializes these variables without requiring mutability.

bmcase: Okay, got it.   Following in that direction was able to simplify elsewhere as well to only pass in the `excluded_helper: Role` to `apply_dp_padding_pass`.

ipa-core/src/protocol/ipa_prf/oprf_padding/README.md
```diff
@@ -56,7 +58,27 @@ $Y=\theta + X_1 - X_2$
...
+We need to add fake rows for matchkeys and fake rows for breakdown keys.  It makes sense to try and add the fake breakdown
+keys to the fake rows already being generated for fake matchkeys. But this approach has a couple challenges:
+1. We shouldn't add any fake breakdown keys to fake matchkey rows when the matchkey is being added with cardinality equal to one.
+Because these rows can be dropped after matching and never have the fake breakdowns revealed.
+2. There may need to be some adjustment made to the DP parameters achieved. TODO
+3. We should not be adding fake breakdown keys to matchkeys that have a cardinality larger than the cap we have established for
+the number of breakdowns per user. Otherwise, those breakdown keys would never be revealed as they will be dropped.
+
+Instead of this approach we will the fake rows for matchkey padding first and then the fake rows for breakdown key padding. When
+we generate the fake rows for breakdown key padding, the fake matchkeys generated will all have cardinality two or three (and with small probability one).
```
andyleiserson, line 84: I don't understand this sentence, is it referring to the older version that added the fake rows for breakdown key padding earlier in the protocol?

ipa-core/src/protocol/ipa_prf/oprf_padding/insecure.rs
```diff
@@ -8,6 +8,7 @@ use rand_core::{CryptoRng, RngCore};
 use crate::protocol::ipa_prf::oprf_padding::distributions::{
     BoxMuller, RoundedBoxMuller, TruncatedDoubleGeometric,
 };
+pub use crate::protocol::ipa_prf::oprf_padding::insecure::Error as DpError;
```
andyleiserson, line 11: I think `pub type DpError = Error` would be more appropriate here. Or, if you're referencing it by fully qualified path in `src/error.rs`, you could do away with this entirely. Having `DpError` as an alias for this `Error` makes more sense in other modules where plain `Error` could conflict.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,798 @@
...
+                for cardinality in 0..matchkey_cardinality_cap {
+                    expected_oprf_total_rows += mean * f64::from(cardinality);
+                }
+            }
+        }
+        (expected_oprf_total_rows, expected_agg_total_rows)
+    }
+
+    #[test]
+    pub fn table_of_padding_parameters() {
```
andyleiserson, line 679: Since this test does not check anything, you could add the `#[ignore]` attribute to skip running it in CI, unless you think that the test running to completion without an assertion failure or panic is providing meaningful coverage.

andyleiserson (approved):

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -1,5 +1,865 @@
...
+                match recv_channel_left.receive(RecordId::FIRST).await {
+                    Ok(v) => Ok::<u32, Error>(u32::try_from(v.as_u128()).unwrap()),
+                    Err(e) => Err(e.into()),
+                }
+            }
+        )?;
+        if from_right != from_left {
+            return Err::<Vec<T>, error::Error>(Error::InconsistentPadding);
+        }
+        total_number_of_fake_rows = from_right;
```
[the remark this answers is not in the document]
bmcase: nice simplification. Thanks!

ipa-core/src/protocol/ipa_prf/oprf_padding/README.md
bmcase: yes, this was for an earlier version that put all the fake rows in at the same place at the start of the protocol.  I'll update the README.md.

bmcase:
Ran draft runs with the `default` DP parameters 


```
        OPRFPadding::Parameters {
            oprf_epsilon: 5.0,
            oprf_delta: 1e-6,
            matchkey_cardinality_cap: 10,
            oprf_padding_sensitivity: 2, // should remain 2
        }
          AggregationPadding::Parameters {
            aggregation_epsilon: 5.0,
            aggregation_delta: 1e-6,
            aggregation_padding_sensitivity: 10, // for IPA is most natural to set
                                                 // equal to the matchkey_cardinality_cap
        }
```
  
  No really impact on draft run performance.   

<b style="font-weight:normal;" id="docs-internal-guid-1387af5a-7fff-23c1-6e21-78b778250808"><div dir="ltr" style="margin-left:0pt;" align="center">
Branch | Query input size | # rows for oprf | # rows for aggregation | Query time | Draft link
-- | -- | -- | -- | -- | --
DP Padding with default parameters | 1000 | 360 | 9220 | 7 sec | https://draft-mpc.vercel.app/query/view/lone-putty2024-08-27T1549
Main | 1000 | NA | NA | 7 sec | https://draft-mpc.vercel.app/query/view/inept-black2024-08-27T1557
Main | 100,000 | NA | NA | 27 sec | https://draft-mpc.vercel.app/query/view/sudsy-cords2024-08-27T1748
DP Padding with default parameters | 100,000 | 654 | 9214 | 27 sec | https://draft-mpc.vercel.app/query/view/elect-leach2024-08-27T1752

</div></b>

 Even with default parameters there are some tests in the CI that run very slowly, so I am also adjusting back to `relaxed` parameters for the OPRF so the tests complete quickly. I will follow up with a PR that lets us pass in a selection for the parameters.

bmcase:
@andyleiserson  or @benjaminsavage  I think this PR is ready to merge now.  Ran some draft runs above and have set the parameters to make CI tests run fast enough.  Will follow up with a PR that lets us pass in a choice of more conservative parameters.  If one of you can merge, that'd be great as I don't have merge access.  If you feel this is too many commits, Alex sometimes asks to use the merge option which squashes into one commit on main.  Thanks

PR #1227, opened 2024-08-15
Cross cloud playbook

eriktaubeneck:
This adds a couple more helper scripts, and updates ansible commands to accommodate:

1. `build_network_file.py`. Builds the network.toml file from the keys stored in `in-market-test/v2/deployed_keys`. Works automatically if testing with 3 servers, or will work if copying those files in with 3 different parties.
2. `parse_logs.py`. Generates a report from the logs copies from the server with breakdown of run time and errors observed.


in-market-test/v2/ansible/build_network_file.py
```diff
@@ -0,0 +1,126 @@
+import argparse
+import configparser
+import os
+import re
+import sys
+from pathlib import Path
+
+
+def get_urls_from_config():
```
cberkhoff, line 9: please specify the type hints at least for the functions

in-market-test/v2/ansible/build_network_file.py
```diff
@@ -0,0 +1,126 @@
...
+        Path("in-market-test/v2/deployed_keys/2-cert.pem").read_text(),
+        Path("in-market-test/v2/deployed_keys/3-cert.pem").read_text(),
+    )
+    public_keys = (
+        Path("in-market-test/v2/deployed_keys/1-mk.pub").read_text(),
+        Path("in-market-test/v2/deployed_keys/2-mk.pub").read_text(),
+        Path("in-market-test/v2/deployed_keys/3-mk.pub").read_text(),
+    )
+
+    network_template = f"""
```
cberkhoff, line 90: can we put the template on a separate file? like ```network.toml.template```

in-market-test/v2/ansible/parse_logs.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @property
+    def size(self):
+        if not self.steps:
+            return None
+        return self.steps[0].size
+
+    @property
+    def report(self):
+        if len(self.steps) == 0:
```
cberkhoff, line 106: can we use a library to create the table? or at least separate the "table printing" logic from Steps?

cberkhoff (changes requested):
Why don't we have a separate repository with a python project? Like ipa-tools or something.

in-market-test/v2/ansible/build_network_file.py
eriktaubeneck: unfortunately not with an f-string

in-market-test/v2/ansible/parse_logs.py
eriktaubeneck: I'm hoping to avoid an libraries - this should be able to run with just python3. 

as for separating it, this is just a quick and dirty script to use when testing. not trying to get too fancy.

eriktaubeneck:
> Why don't we have a separate repository with a python project? Like ipa-tools or something.

I wouldn't call this a "project", it's just some scripts used for deploying/testing for the in market test.

in-market-test/v2/ansible/build_network_file.py
cberkhoff: Can we use at the very least something like what's described here with strings? https://stackoverflow.com/questions/6385686/is-there-a-native-templating-system-for-plain-text-files-in-python

in-market-test/v2/ansible/build_network_file.py
eriktaubeneck: added.

in-market-test/v2/ansible/build_network_file.py
eriktaubeneck: what's the advantage? the f-string makes it very clear what's going where, and also will allow for syntax highlighting if you try to access a variable that doesn't exist.

in-market-test/v2/ansible/parse_logs.py
cberkhoff: Can we move this to a separate module function?  (same with errors): `steps_to_report(steps: Steps)` and `errors_to_report(...)`

There's so much going on in these methods I don't think they should be marked as @property.

in-market-test/v2/ansible/build_network_file.py
cberkhoff: It separates the template from where the values are coming from. The same principle as to why web frameworks separate "views" from controllers/models.

cberkhoff (changes requested):

in-market-test/v2/ansible/build_network_file.py
eriktaubeneck: plenty of modern web frameworks don't separate those, e.g., React is all about mixing html and JS. 

I might be inclined to take a template route if we had lots of templates, and wanted to be super generic, but here we have a single file to build. it seems like a net-loss to separate the template and the variables, because we no longer get automatic linting that can tell you if you have a typo in one of the variables inserted into the template.

cberkhoff (approved):
Had an offline discussion.

PR #1229, opened 2024-08-19
Make malicious reveal more flexible

andyleiserson:
Allow malicious reveal of shared values not in the same field as the MAC. This will be useful for #1221.


andyleiserson:
Changes were included in #1221.

PR #1230, opened 2024-08-20
Python HPKE sample code add csv file format and output 3 reports

shinta-liem:
Added parsing comma delimited format:
File format: <timestamp>,<match_key>,<event_type>,<breakdown_key>,<trigger_value>

Output 3 reports, 1 for each helper.

### DIFF
--- a/in-market-test/v2/hpke/encrypt_report.py
+++ b/in-market-test/v2/hpke/encrypt_report.py
@@ -1,4 +1,10 @@
 import argparse
+from enum import Enum
+import math
+import os
+import secrets
+from typing import Any
+import numpy as np
 
 import pyhpke
 from cryptography.hazmat.primitives.asymmetric import x25519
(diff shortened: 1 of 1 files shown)


in-market-test/v2/hpke/encrypt_report.py
```diff
@@ -107,31 +146,202 @@ def encrypt_share(
...
+        self.right: bytes = right
+
+    def to_bytes(self) -> bytes:
+        return self.left + self.right
+
+    def xor(a: bytes, b: bytes) -> bytes:
+        a_array = np.frombuffer(a, dtype=np.uint8)
+        b_array = np.frombuffer(b, dtype=np.uint8)
+        result_array = a_array ^ b_array
+        return result_array.tobytes()
```
eriktaubeneck, line 163: No need to use numpy here, and best to avoid that dependency (it's a pretty heavy package.)

```suggestion
    def xor(a: bytes, b:bytes) -> bytes:
        _len = max((len(a), len(b)))
        a = int.from_bytes(a, "little")
        b = int.from_bytes(b, "little")
        x = a ^ b
        return x.to_bytes(_len, "little")
```

in-market-test/v2/hpke/encrypt_report.py
```diff
@@ -1,8 +1,47 @@
 import argparse
+from enum import Enum
+import math
+import os
+import secrets
+from typing import Any
+import numpy as np
```
eriktaubeneck, line 7: See the suggestion below for removing this dependency.

```suggestion
```

eriktaubeneck:
thank you for adding this! one small suggestion to remove the `numpy` dependency, otherwise looks good!

eriktaubeneck (approved):

in-market-test/v2/hpke/encrypt_report.py
shinta-liem: Removed

PR #1233, opened 2024-08-20
update README with how to pass DP params on command line

bmcase:
Add documentation to README.md for how to pass a custom epsilon command line or disable the default DP for outputs.


eriktaubeneck (approved):

eriktaubeneck:
out of scope for this PR, but it would be useful to also have this worked into the `report_collector` binary for running test queries (used in `draft` and the in-market-test.)

PR #1234, opened 2024-08-20
Updates to malicious validation batcher

andyleiserson:
Some of these changes are general improvements, and some will be needed to use the batcher for DZKPs (currently the DZKP validator inlines an early version of the batcher from before it was refactored into a separate file -- this will be addressed in a later change).

* Make total_records optional in the batcher. This will be used for the unbatched `DZKPValidator::validate` API (which still uses the batcher to store intermediates, because that is easier than conditionally using different storage).
* Handle total_records correctly in the multiple batch case.
* Remove an unnecessary atomic for the pending record counter.
* Also track records waiting for validation in a bitvec, as a sanity-check.
* Invoke the batch validation function inside `validate_record`, to reduce duplication between the MAC and DZKP implementations.
* Report validation success/failure to parallel validate_record waiters.
* Add tests for the batcher.


ipa-core/src/protocol/context/batcher.rs
```diff
@@ -1,50 +1,43 @@
...
-    }
-
-    fn right(value: R) -> Self {
-        Self::Right(value)
-    }
-}
-
 #[derive(Debug)]
-pub struct BatchState<B> {
+pub(super) struct BatchState<B> {
```
cberkhoff, line 13: It would be good to add some docs on these structs. What they do and how they coordinate together.

ipa-core/src/protocol/context/batcher.rs
```diff
@@ -1,50 +1,43 @@
...
     batches: VecDeque<BatchState<B>>,
     first_batch: usize,
     records_per_batch: usize,
-    total_records: usize,
+    total_records: Option<usize>,
     batch_constructor: Box<dyn Fn(usize) -> B + Send + 'a>,
 }
 
+// Helper for `Batcher::validate_record`.
+enum Validate<B> {
```
cberkhoff, line 29: Should this be named something like `ValidationTime` ?

ipa-core/src/protocol/context/batcher.rs
```diff
@@ -89,40 +83,404 @@ impl<'a, B> Batcher<'a, B> {
         self.get_batch_by_offset(self.batch_offset(record_id))
     }
 
-    pub fn validate_record(
-        &mut self,
-        record_id: RecordId,
-    ) -> Either<(usize, BatchState<B>), Arc<Notify>> {
-        tracing::trace!("validate record {record_id}");
+    fn validate_record_action(&mut self, record_id: RecordId) -> Result<Validate<B>, Error> {
```
cberkhoff, line 86: Naming ideas: `is_ready_to_validate` ?

cberkhoff (approved):
Just had a few comments on names and documentation

ipa-core/src/protocol/context/batcher.rs
andyleiserson: I went with `Ready::No` and `Ready::Yes` since this is the return value from `is_ready_to_validate`.

cberkhoff (approved):

PR #1236, opened 2024-08-21
Fix bug in DP README.md

eriktaubeneck:
fix math formatting

### DIFF
--- a/ipa-core/src/protocol/dp/README.md
+++ b/ipa-core/src/protocol/dp/README.md
@@ -8,10 +8,11 @@ introduces Binomials for DP.
 considers their use for d-dimension queries (such as we will need for WALR).
 
 
-To achieve a desired $(\varepsilon, \delta)$-DP guarantee, we generate $num_bernoulli$ secret shared samples of a 
+To achieve a desired $(\varepsilon, \delta)$-DP guarantee, we generate $num\\_bernoulli$ secret shared samples of a 
 Bernoulli having probability $0.5$ using PRSS.  Next we aggregate them to get a Binomial sample. The result of the 2018 
-paper above is that for small epsilon (TODO, how small required?), we require the following number of samples
-$$ num_bernoulli \geq 8 \log(2/\delta) /\varepsilon^2$$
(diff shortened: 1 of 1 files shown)


ipa-core/src/protocol/dp/README.md
```diff
@@ -8,10 +8,11 @@ introduces Binomials for DP.
...
 
 
-To achieve a desired $(\varepsilon, \delta)$-DP guarantee, we generate $num_bernoulli$ secret shared samples of a 
+To achieve a desired $(\varepsilon, \delta)$-DP guarantee, we generate $num\\_bernoulli$ secret shared samples of a 
 Bernoulli having probability $0.5$ using PRSS.  Next we aggregate them to get a Binomial sample. The result of the 2018 
-paper above is that for small epsilon (TODO, how small required?), we require the following number of samples
-$$ num_bernoulli \geq 8 \log(2/\delta) /\varepsilon^2$$
+paper above is that for small epsilon (TODO, how small required?), we require the following number of samples:
+
+$$ num\\_bernoulli \geq \frac{8 \log \left( \frac{2}{\delta}\right)}{\varepsilon^2}$$
```
bmcase, line 15: This formula is a high level summary of the paper's result.  The complete calculation for determining num_bernoulli is more involved in what the code actually does.  We using this to get some early estimates but now that we have the full analysis implemented we should probably just point to that from the readme and drop this approximation.

eriktaubeneck: that works too. just noticed that it wasn't formatting correctly so I figured I'd fix it 🙂

eriktaubeneck: it would be useful to provide the instructions for computing the expectation and arriving at the unbiased estimate (even if it's not yet implemented.)

eriktaubeneck:
@bmcase I removed the formula. is this good to close?

bmcase:
yep, looks good. thanks for the edits

bmcase (approved):

PR #1240, opened 2024-08-27
fix precommit bug when using zsh

eriktaubeneck:
`echo " ${@@Q}"` causes a `Bad Substitution error` in zsh (now the default on macOS), which very unfortunately gives a 0 exit code.  `echo " ${*}"` still prints the command that it failed on, and it should work with bash and zsh.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -73,7 +73,7 @@ check() {
         res=$?
         echo "-----------------------"
         echo "$msg: FAILED ($res)"
-        echo "  ${@@Q}"
+        echo "  ${*}"
         exit $res
     fi
 }
--- a/pre-commit
+++ b/pre-commit
@@ -80,9 +80,6 @@ check() {
     fi
 }
 
-check "Fail fast" \
-    python fail_fast.py --foo "bar"
-
 check "Benchmark compilation" \
     cargo build --benches --no-default-features --features "enable-benches compact-gate"
 
--- a/pre-commit
+++ b/pre-commit
@@ -75,6 +75,8 @@ check() {
         echo "$msg: FAILED ($res)"
         if [[ "$BASH_VERSINFO" -ge 4 ]]; then
             echo " ${@@Q}"
+        else
+            echo "$*"
         fi
         exit $res
     fi
--- a/pre-commit
+++ b/pre-commit
@@ -73,10 +73,15 @@ check() {
         res=$?
         echo "-----------------------"
         echo "$msg: FAILED ($res)"
-        echo "  ${*}"
+        if [[ "$BASH_VERSINFO" -ge 4 ]]; then
+            echo " ${@@Q}"
+        fi
         exit $res
     fi
 }
+
+check "Fail fast" \
+    python fail_fast.py --foo "bar"
 
 check "Benchmark compilation" \
     cargo build --benches --no-default-features --features "enable-benches compact-gate"


pre-commit
```diff
@@ -73,7 +73,7 @@ check() {
         res=$?
         echo "-----------------------"
         echo "$msg: FAILED ($res)"
-        echo "  ${@@Q}"
```
andyleiserson, line 76: I am sorry to be pedantic on a one line PR, but I'm a little confused. The shebang line calls for bash, so why is the the script running in zsh? I also can't reproduce the exit code of zero -- I get an exit code of 1. (I also got a glob failure on `.git/MERGE_*` until I made a file to match it, so that might be worth fixing as well if we really want to run this under bash.

All that aside, it might better to do `" ""$@"`, as I think that's closer to `" ${@@Q}"` than `" ${*}"`.

eriktaubeneck: Maybe this issue isn't bash vs zsh... I don't get much other than the "Bad Substitution" error, and a little bit of Googling suggested that `{@@Q}` may not be available in zsh (and my terminal runs zsh, but you raise a good point about the shebang.)

Let me give your suggestion a try.

martinthomson: The point behind using `${@@Q}` was to ensure that any quoting needed to have the command-line work was included.  That way, you could have a command with spaces, like `cargo test -- 'something with spaces'` print in a way that could be copied and pasted back.  `"$@"` or `"$*"` don't achieve that.

Of course, if this is failing for someone, that's probably because Apple are shipping an ancient version of bash that predates the move to GPLv3.  That is why new macs tend to use zsh: that old bash is now quite bad.

eriktaubeneck: that's probably what's going on. suggestion on the right way to fix it?

martinthomson: https://formulae.brew.sh/formula/bash#default perhaps.

eriktaubeneck: sure, but it would be nice to have this work out of the box on macOS, no? the fact that it fails but the commit still goes forward seems like a failure case you'd want to prevent as universally as possible.

martinthomson: Hmm, it shouldn't.  We have `set -e` on.  That should cause the error to trigger an `exit 1`.

I can't test this easily.  Maybe you can work out how to test for something that isn't broken so that you don't get this failing open like that.

eriktaubeneck: ok let me look into it further. it's certainly the worst failure mode and it's important to at least fail correctly

eriktaubeneck: this is what I get with `${@@Q}` 😦 
image

but, as you note, with `${*}` or with `" ""$@"`, it drops the quotes, e.g. the command I have in there is `python fail_fast.py --foo "bar"` and it drops those quotes:
image

martinthomson: Sounds like a bug in that version of bash.

If I add a bad substitution, I get:

$ ./pre-commit
-----------------------
fail: FAILED (1)
  'false' '1 2 3'
./pre-commit: line 77:   ${@(}: bad substitution
$ echo $?
1

pre-commit
```diff
@@ -73,11 +73,16 @@ check() {
...
+        if [[ "$BASH_VERSINFO" -ge 4 ]]; then
+            echo " ${@@Q}"
+        fi
```
martinthomson, line 78:
```suggestion
        else
            echo "$*"
        fi
```

No point in doing the "$@" thing.

pre-commit
```diff
@@ -73,11 +73,16 @@ check() {
...
         exit $res
     fi
 }
 
+check "Fail fast" \
+    python fail_fast.py --foo "bar"
+
```
martinthomson, line 85:
```suggestion
```

martinthomson (approved):

pre-commit
eriktaubeneck: that makes more sense. just pushed a check for the version, and it correctly exits with the stock version of bash.

however, I also updated bash, and what I get out is 
```
$ ./pre-commit
-----------------------
Fail fast: FAILED (1)
 'python' 'fail_fast.py' '--foo' 'bar'
```
where I'd expect
```
python fail_fast.py --foo "bar"
```

martinthomson: The quoting is not preserved, but re-added.  That means that you get more quotes than is necessary.  But at least `"foo bar"` doesn't turn into two arguments (`foo bar`) but `foo bar`.

PR #1241, opened 2024-08-28
Truncated Discrete Laplace mechanism

bmcase:
adding a Laplace DP mechanism for output privacy.   Every pair of helpers adds Truncated Discrete Laplace noise to the output.


ipa-core/src/cli/playbook/mod.rs
```diff
@@ -106,18 +113,41 @@ pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_
...
+            DpMechanism::DiscreteLaplace { epsilon: _ } => {
+                let truncated_discrete_laplace = OPRFPaddingDp::new(
+                    noise_params.epsilon,
+                    noise_params.delta,
+                    noise_params.per_user_credit_cap,
+                )
+                .unwrap();
+
+                let (mean, std) = truncated_discrete_laplace.mean_and_std();
+                let tolerance_factor = 12.0;
```
eriktaubeneck, line 141: should this be an argument with a default of 12? can we also add a comment explaining what this is and why it's set to 12?

ipa-core/src/cli/playbook/mod.rs
```diff
@@ -106,18 +113,41 @@ pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_
...
+                let truncated_discrete_laplace = OPRFPaddingDp::new(
+                    noise_params.epsilon,
+                    noise_params.delta,
+                    noise_params.per_user_credit_cap,
+                )
+                .unwrap();
+
+                let (_, std) = truncated_discrete_laplace.mean_and_std();
+                let tolerance_factor = 12.0;
+                // println!("mean = {mean}, std = {std}, tolerance_factor * std = {}",tolerance_factor * std);
```
eriktaubeneck, line 142:
```suggestion
```

ipa-core/src/cli/playbook/mod.rs
```diff
@@ -106,18 +113,41 @@ pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_
...
+                (actual_expect_f64 - next_expected_f64).abs() < tolerance_factor * std
+            }
+            DpMechanism::NoDp => {
+                let tolerance = 0.001;
```
eriktaubeneck, line 146: why isn't this tolerance 0?

ipa-core/src/cli/playbook/mod.rs
bmcase: sure, will add some comment about why we set it so large. What we want is for this check to fail randomly with very small probability.

ipa-core/src/cli/playbook/mod.rs
bmcase: we are doing a float comparison here, so we have some tolerance to account for small round-off errors (I think it was a clippy suggestion to add in case of float comparisons).  But let me see if we can use the earlier u32 representations here instead.

bmcase: okay updated to change this.

ipa-core/src/cli/playbook/mod.rs
```diff
@@ -106,18 +113,53 @@ pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_
...
+
+                let (_, std) = truncated_discrete_laplace.mean_and_std();
+                let tolerance_factor = 20.0; // set so this fails randomly with small probability
+                                             // println!("mean = {mean}, std = {std}, tolerance_factor * std = {}",tolerance_factor * std);
+                (actual_expect_f64_shifted - next_expected_f64).abs() < tolerance_factor * 3.0 * std
+            }
+            DpMechanism::NoDp => {
+                // let tolerance = 0.00001;
+                // actual_expect_f64 > next_expected_f64 - tolerance
+                //     && actual_expect_f64 < next_expected_f64 + tolerance
```
eriktaubeneck, line 159:
```suggestion
```

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -236,15 +242,16 @@ where
                 return Err(EpsilonOutOfBounds);
             }
 
-            let per_user_credit_cap = 2_f64.powi(i32::try_from(SS_BITS).unwrap());
+            let per_user_credit_cap = 2_u32.pow(u32::try_from(SS_BITS).unwrap());
```
eriktaubeneck, line 245: suppose we are setting `per_user_credit_cap` to, say, 5, but the `SS_BITS` are larger (for 5, you'd need at least 3 bits). doesn't this result in more noise than needed? more generally, why infer this from `SS_BITS` when we have a more precise value as an input to the query?

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -268,9 +275,149 @@ where
...
+where
+    C: Context,
+    OV: BooleanArray + U128Conversions,
+    Boolean: Vectorizable<B> + FieldSimd<B>,
+    Replicated<Boolean, B>: BooleanProtocols<C, B>,
+    BitDecomposed<AdditiveShare<Boolean, B>>:
+        for<'a> TransposeFrom<&'a [AdditiveShare<OV>; B], Error = Infallible>,
+    AdditiveShare<OV>: ReplicatedSecretSharing<OV>,
+{
+    let mut noise_values = vec![];
```
eriktaubeneck, line 354: probably cleaner to use `Vec::with_capacity(B);`, since we know the size. @akoshelev can you double check me here?
```suggestion
    let mut noise_values = Vec::with_capacity(B);
```

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -388,19 +535,91 @@ mod test {
...
+    pub async fn test_laplace_noise() {
+        type OV = BA8;
+        const NUM_BREAKDOWNS: u32 = 16;
+        const SS_BITS: usize = 3;
+        let epsilon = 2.0;
+        let dp_params = DpMechanism::DiscreteLaplace { epsilon };
+        let world = TestWorld::default();
+        // let input_values = [10000, 8, 6, 41, 0, 0, 0, 0, 10, 8, 6, 41, 0, 0, 0, 0];
+        // let input_values = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        // let input_values = [1, 1, 1, 1,1, 1, 1, 1,1, 1, 1, 1,1, 1, 1, 1];
```
eriktaubeneck, line 582:
```suggestion
```

ipa-core/src/cli/playbook/mod.rs
```diff
@@ -81,7 +82,13 @@ where
...
 /// ## Panics
 /// If results don't match.
-pub fn validate_dp(expected: Vec<u32>, actual: Vec<u32>, epsilon: f64, per_user_credit_cap: u32) {
+pub fn validate_dp(
+    expected: Vec<u32>,
+    actual: Vec<u32>,
+    epsilon: f64,
+    per_user_credit_cap: u32,
+    dp_mechanism: DpMechanism,
+) {
```
eriktaubeneck, line 91: per our offline conversation, the `table` below (line 96) is what get's printed out. one idea would be to add two columns, e.g. `Expected Lower Bound (p95)` and `Expected Upper Bound (p95)`. otherwise I think this is now covered all the way up the stack.

eriktaubeneck (approved):
a couple small comments but looks good

ipa-core/src/protocol/dp/mod.rs
akoshelev: array approach makes sense to me, given that we ultimately need an array anyway

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -268,9 +275,149 @@ where
...
+            noise_params.epsilon,
+            noise_params.delta,
+            noise_params.per_user_credit_cap,
+        )?;
+        let shift = truncated_discrete_laplace.get_shift();
+        for _i in 0..B {
+            let sample = truncated_discrete_laplace.sample(rng);
+            assert!(OV::BITS <= 32);
+            let symmetric_sample = if OV::BITS < 32 {
+                let modulus = 2_u32.pow(OV::BITS);
```
akoshelev, line 377: we shouldn't be doing this in a loop I think

ipa-core/src/protocol/dp/mod.rs
eriktaubeneck: we probably should split this into smaller functions as well

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -268,9 +275,149 @@ where
...
+                sample.wrapping_sub(shift) % modulus
+            } else {
+                sample.wrapping_sub(shift)
+            };
+            // println!("sample = {sample}, symmetric sample = {symmetric_sample}");
```
eriktaubeneck, line 382:
```suggestion
```

ipa-core/src/protocol/dp/mod.rs
```diff
@@ -268,9 +275,149 @@ where
...
+
+        let truncated_discrete_laplace = OPRFPaddingDp::new(
+            noise_params.epsilon,
+            noise_params.delta,
+            noise_params.per_user_credit_cap,
+        )?;
+        let shift = truncated_discrete_laplace.get_shift();
+        for _i in 0..B {
+            let sample = truncated_discrete_laplace.sample(rng);
+            assert!(OV::BITS <= 32);
```
eriktaubeneck, line 375: does this change through out the loop?

ipa-core/src/protocol/dp/mod.rs
bmcase: merged Erik's refactor to build array directly

ipa-core/src/protocol/dp/mod.rs
bmcase: Right now the code only supports a `per_user_credit_cap` that is a power of two and when calling `oprf_ipa` the conversions happens to pass in a generic parameters of the power of 2 `SS_BITS`  https://github.com/private-attribution/ipa/blob/77ac0b009afcdef2152ce5b9ed0499f4f7cdebb5/ipa-core/src/query/runner/oprf_ipa.rs#L123 

By guess is that it is capping that most heavily uses this power of 2 assumption.  When we shift to the hybrid without a capping stage is probably a good time to refactor to support a more flexible value for the sensitivity in the query. 

It will add more noise than needed if the true cap is less than a power of two, but the caller of the API get around this by scaling up their inputs (by (`next_power_of_two/true_cap`)) and then rescale the outputs.

ipa-core/src/cli/playbook/mod.rs
bmcase: may take me a little longer to get a good confidence interval as need to analyze sum of truncated discrete laplace.   i'd say we should go ahead and merge these two PRs so we can start observing the performance changes 
1.https://github.com/private-attribution/ipa/pull/1247#pullrequestreview-2283787281
2. https://github.com/private-attribution/ipa/pull/1241

eriktaubeneck: makes sense!

PR #1245, opened 2024-08-29
Add helper command line arg to override active_work

andyleiserson:


pre-commit
```diff
@@ -102,6 +102,9 @@ check "Tests" \
 check "Web tests" \
     cargo test -p ipa-core --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"
 
+check "Web tests (descriptive gate)" \
+    cargo test -p ipa-core --no-default-features --features "cli web-app real-world-infra test-fixture"
+
```
eriktaubeneck, line 107: maybe a dumb question, but how is descriptive gate related to the active work?

eriktaubeneck (approved):
one question, but otherwise looks good. thanks for adding this.

the missing coverage seems to be small changes in areas already missing coverage.

ipa-core/src/helpers/gateway/mod.rs
```diff
@@ -98,6 +102,7 @@ impl Gateway {
         mpc_transport: MpcTransportImpl,
         shard_transport: ShardTransportImpl,
     ) -> Self {
+        tracing::info!("active_work = {}", config.active);
```
akoshelev, line 105: maybe `debug`?

akoshelev (approved):

pre-commit
andyleiserson: It's not, but some of the plumbing I had to change to add the option broke a test case that only runs with descriptive gate, that is run by the coverage job, but not by `pre-commit`, so I added this.

ipa-core/src/helpers/gateway/mod.rs
andyleiserson: The reason I made it `info` is that there is logic in `helpers::transport::query` that configures the active work dynamically to be no more than the query input size. At the time I didn't see the note in the commit description "Input size constraint comes from our CI that is so slow, it cannot allow very large active work sizes.", and the logic didn't make sense to me -- but I figured that if we really do need to be tuning it adaptively per query, then we probably also want to log it per query, at a level that we don't need to restart the server in debug mode to see.

But now that I have seen the commit message, it makes more sense. I will add the comment about CI in the code though.

PR #1246, opened 2024-08-30
Conversion between Boolean Array and Gf32Bit

danielmasny:
conversion between BA types and Vec<Gf32Bit>. Also adds pub function that allows to get Gf32Bit as raw slice.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -266,7 +266,7 @@ macro_rules! boolean_array_impl {
         mod $modname {
             use super::*;
             use crate::{
-                ff::{boolean::Boolean, ArrayAccess, Expand, Serializable},
+                ff::{boolean::Boolean, ArrayAccess, Expand, Serializable, Gf32Bit},
                 impl_shared_value_common,
                 secret_sharing::{
                     replicated::semi_honest::{BAASIterator, AdditiveShare},
@@ -497,6 +497,28 @@ macro_rules! boolean_array_impl {
                 }
             }
 
+            /// This function converts an `BA` type into a vector of `32 bit` Galois field elements.
+            ///
+            /// ##Errors
+            /// Outputs an error when conversion from raw slice to Galois field element fails.
+            impl TryFrom<$name> for Vec<Gf32Bit> {
+                type Error = crate::error::Error;
+
+                fn try_from(value: $name) -> Result<Self, Self::Error> {
+                    // len() returns bits, so divide by 8
+                    // further divide by 4 since Gf32Bit has 4 byte
+                    // add 31 to round up
+                    let length = (value.0.len() + 31) / 32;
+                    let mut chunks = Vec::<Gf32Bit>::with_capacity(length);
+                    let last_chunk_start = length-1;
+                    for i in 0..last_chunk_start {
+                        chunks.push(Gf32Bit::try_from(&value.0.as_raw_slice()[4 * i..4 * (i + 1)])?);
+                    }
+                    chunks.push(Gf32Bit::try_from(&value.0.as_raw_slice()[4 * last_chunk_start..])?);
+                    Ok(chunks)
+                }
+            }
+
             impl SharedValueArray<Boolean> for $name {
                 const ZERO_ARRAY: Self = <$name as SharedValue>::ZERO;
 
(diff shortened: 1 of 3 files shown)


ipa-core/src/ff/galois_field.rs
```diff
@@ -133,7 +136,7 @@ fn clmul<GF: GaloisField>(a: GF, b: GF) -> u128 {
 }
 
 macro_rules! bit_array_impl {
-    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $one:expr, $polynomial:expr, $deser_type: tt, $({$($extra:item)*})? ) => {
+    ( $modname:ident, $name:ident, $store:ty, $bits:expr, $bytes:expr, $one:expr, $polynomial:expr, $deser_type: tt, $({$($extra:item)*})? ) => {
```
akoshelev, line 139: does `BITS/8` work to avoid an extra argument here?

ipa-core/src/ff/boolean_array.rs
```diff
@@ -497,6 +497,28 @@ macro_rules! boolean_array_impl {
                 }
             }
 
+            /// This function converts an `BA` type into a vector of `32 bit` Galois field elements.
+            ///
+            /// ##Errors
+            /// Outputs an error when conversion from raw slice to Galois field element fails.
+            impl TryFrom<$name> for Vec<Gf32Bit> {
```
akoshelev, line 504: is it better to define conversion to an array for each BA type? we know the size at compile time, so we can save some cost here

ipa-core/src/ff/boolean_array.rs
```diff
@@ -497,6 +497,28 @@ macro_rules! boolean_array_impl {
...
+                type Error = crate::error::Error;
+
+                fn try_from(value: $name) -> Result<Self, Self::Error> {
+                    // len() returns bits, so divide by 8
+                    // further divide by 4 since Gf32Bit has 4 byte
+                    // add 31 to round up
+                    let length = (value.0.len() + 31) / 32;
+                    let mut chunks = Vec::<Gf32Bit>::with_capacity(length);
+                    let last_chunk_start = length-1;
```
akoshelev, line 513: the formatting here is messed up quite a bit, can you take that function out and run `cargo fmt` before taking it back inside the macro?

ipa-core/src/ff/boolean_array.rs
danielmasny: Wouldn't each Array/GenericArray be its own type? When generating the vector/array of Gf32Bit from different BA types it would be nice to receive the same type. A function like 
```
fn<B:BooleanArray>(input: B)
{
    let b = Vec<Gf32Bit>::from(input);
}
```
works. When using an array, I think I would need to use a constant generic or arraylength something like this
```
fn<B:BooleanArray, U:ArrayLength>(input: B)
{
    let b = <GenericArray<Gf32Bit,U>>::from(input);
}
```

danielmasny: For performance reason, I agree with you, we should definitely replace Vec. I first want to ship something that is functioning and will take it on later: https://github.com/private-attribution/ipa/issues/1254.

ipa-core/src/ff/galois_field.rs
```diff
@@ -153,6 +156,15 @@ macro_rules! bit_array_impl {
                 }
             }
 
+            impl $name {
+
+                #[inline]
+                #[must_use]
+                pub fn as_raw_slice(&self) -> &[u8] {
```
akoshelev, line 163: @andyleiserson do you have any concerns about implicit endianness here?

akoshelev (approved):
lgtm but I'd love @andyleiserson eyes here too

andyleiserson:
I think it's better to squash the changes into this PR (it will reduce the diff significantly). One thing that's worth doing before merging is adding a comment on `trait SerializeAs`. Maybe ``Serves a role similar to `Borrow<T>` for the type parameter to `compute_hash`. The problem with `Borrow<T>` is that it is implemented for both `T` and `&T`, which then requires explicitly specifying `S` when `compute_hash` is called to resolve the ambiguity.``

ipa-core/src/ff/galois_field.rs
danielmasny: I actually do not need this, so I will remove it.

PR #1247, opened 2024-08-30
Relaxed Padding DP behind feature flag

bmcase:
Make is such that we run with `default` Padding DP parameters unless certain feature flags are turned on for testing.  Following Andy's suggested approach to add a new option to the `QueryType` enum, `QueryType::OprfIpaRelaxedDpPadding`


ipa-core/src/bin/report_collector.rs
```diff
@@ -183,7 +183,7 @@ async fn ipa(
     let query_type: QueryType;
     match (security_model, &query_style) {
         (IpaSecurityModel::SemiHonest, IpaQueryStyle::Oprf) => {
-            query_type = QueryType::OprfIpa(ipa_query_config);
+            query_type = QueryType::OprfIpaRelaxedDpPadding(ipa_query_config);
```
cberkhoff, line 186: why not add an optional attribute in IpaQueryConfig called `relaxed_dp_padding`?

bmcase: I'm sure there are other ways of doing it, but this was the suggested approach from @andyleiserson.    I feel this way could be easy to extend in the future to supporting a RelaxedDP mode for DP noise generation in MPC, not just the relaxed for PaddingDP.

eriktaubeneck: I suppose the key question here is if semi-honest implies that relaxed DP is appropriate. the inverse is certainly true, i.e., malicious implies not relaxed DP, but there could be cases where we want semi-honest, but regular DP padding. 

if we're comfortable with tying these two together (and that probably makes sense for now), this works. otherwise, we may want this to be a distinct `IpaQueryStyle`. (which is incompatible with the malicious security model.)

ipa-core/src/query/executor.rs
```diff
@@ -101,6 +101,22 @@ pub fn execute<R: PrivateKeyRegistry>(
...
+            input,
+            move |prss, gateway, config, input| {
+                let ctx = SemiHonestContext::new(prss, gateway);
+                Box::pin(
+                    OprfIpaQuery::<BA32, R>::new(ipa_config, key_registry)
+                        .execute(ctx, config.size, input)
+                        .then(|res| ready(res.map(|out| Box::new(out) as Box<dyn Result>))),
+                )
+            },
+        ),
```
eriktaubeneck, line 117: I'm confused by this warning. it seems like the test coverage should be (intentionally) missing for lines 120-131. but otherwise this is fine, we are doing this specifically to not add full padding in tests.

ipa-core/src/query/executor.rs
```diff
@@ -133,6 +154,22 @@ pub fn execute<R: PrivateKeyRegistry>(
...
```
eriktaubeneck, line 172: ok, this codecov makes sense and is fine.

ipa-core/src/query/executor.rs
eriktaubeneck: it seems like we test this elsewhere? i'm not immediately familiar with exactly how this is tested (or if we'd expect codecov to catch it)

eriktaubeneck (approved):
a couple comments to double check that the codecov warnings are expected (e.g., we're intentionally not testing those lines), but otherwise looks good.

ipa-core/src/query/executor.rs
bmcase: I think it was shown as missing before for other QueryTypes so copying one of those to make a new one just increased the ratio past 20%.  So I don't think we're really missing some tests that other QueryTypes have somewhere.

PR #1249, opened 2024-09-03
changes to shuffle function

danielmasny:
change shuffle signature, store intermediate messages and tests

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/shuffle/base.rs
+++ b/ipa-core/src/protocol/ipa_prf/shuffle/base.rs
@@ -97,6 +97,8 @@ where
     x_1.shuffle(&mut rng_perm_r);
 
     // need to output x_1
+    // call to clone causes allocation
+    // ideally in the semi honest setting, we would not clone
     let mut x_2 = x_1.clone();
     add_single_shares_in_place(&mut x_2, z_31);
     x_2.shuffle(&mut rng_perm_l);
@@ -156,6 +158,8 @@ where
     .await?;
 
     // we need to output x_2
+    // call to clone causes allocation
+    // ideally in the semi honest setting, we would not clone
     let mut x_3 = x_2.clone();
     add_single_shares_in_place(&mut x_3, z_23);
     x_3.shuffle(&mut rng_perm_r);
@@ -223,6 +227,8 @@ where
     .await?;
 
     // need to output y_1
+    // call to clone causes allocation
+    // ideally in the semi honest setting, we would not clone
     let mut y_2 = y_1.clone();
     add_single_shares_in_place(&mut y_2, z_31);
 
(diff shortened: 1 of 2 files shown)


ipa-core/src/protocol/ipa_prf/shuffle/base.rs
```diff
@@ -18,7 +18,10 @@ use crate::{
 
 /// # Errors
 /// Will propagate errors from transport and a few typecasts
-pub async fn shuffle<C, I, S>(ctx: C, shares: I) -> Result<Vec<AdditiveShare<S>>, Error>
+pub async fn shuffle<C, I, S>(
```
akoshelev, line 21: is this function going to be the entry point for shuffle protocol or you plan to create another wrapper for it?

danielmasny: For semi honest shuffle, you can call it directly (and ignore the IntermediateMessages output). For malicious shuffle, there will be a malicious shuffle function that will call this function.

akoshelev: it would be better imo to avoid returning data that is not used in semi-honest setting. Maybe we could define a trait that specifies the output type for semi-honest contexts to be `Vec<AdditiveShare<S>>` and Vec<> + Intermediates for malicious

Alternatively we could explore an option to validate intermediates inside `shuffle` - for semi-honest, validation will be a no-op and for malicious it should check MACs. Then callers would just specify which protocol they want to run by passing the right context version (as with other protocols)

ipa-core/src/protocol/ipa_prf/shuffle/base.rs
```diff
@@ -44,12 +53,23 @@ where
     }
 }
 
+#[allow(dead_code)]
+/// This struct stores some intermediate messages during the shuffle.
+/// In a maliciously secure shuffle,
+/// these messages need to be checked for consistency across helpers.
+/// `H1` stores `x1`, `H2` stores `x2` and `H3` stores `y1` and `y2`.
+#[derive(Debug, Clone)]
```
akoshelev, line 61:
```suggestion
#[derive(Debug, Clone, Default)]
```

ipa-core/src/protocol/ipa_prf/shuffle/base.rs
```diff
@@ -76,21 +96,29 @@ where
...
 
     let res = combine_single_shares(a_hat, b_hat).collect::<Vec<_>>();
-    Ok(res)
+    // we only need to store x_1 in IntermediateShuffleMessage
+    Ok((
+        res,
+        IntermediateShuffleMessages {
+            x1_or_y1: Some(x_1),
+            x2_or_y2: None,
+        },
```
akoshelev, line 112:
```suggestion
        IntermediateShuffleMessages::default(),
```

akoshelev (approved):

PR #1250, opened 2024-09-03
Use batcher for DZKPs. Enable tests of DZKP batching.

andyleiserson:

PR #1251, opened 2024-09-03
Malicious aggregation and DP noising

andyleiserson:
Adds test coverage for malicious attribution and aggregation

### DIFF
--- a/ipa-core/src/protocol/context/dzkp_validator.rs
+++ b/ipa-core/src/protocol/context/dzkp_validator.rs
@@ -7,6 +7,7 @@ use futures::{stream, Future, FutureExt, Stream, StreamExt};
 use crate::{
     error::{BoxError, Error},
     ff::{Fp61BitPrime, U128Conversions},
+    helpers::TotalRecords,
     protocol::{
         context::{
             batcher::Batcher,
@@ -600,6 +601,11 @@ pub trait DZKPValidator: Send + Sync {
 
     fn context(&self) -> Self::Context;
 
+    /// Sets the validator's total number of records field. This is required when using
+    /// the validate_record API, if it wasn't already set on the context used to create
+    /// the validator.
+    fn set_total_records<T: Into<TotalRecords>>(&mut self, total_records: T);
+
     /// Validates all of the multiplies associated with this validator.
     ///
     /// Only one of the `DZKPValidator::validate` or the `DZKPContext::validate_record`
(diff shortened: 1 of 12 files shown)


andyleiserson:
The latest CI failure is real (and potentially non-deterministic):

```
thread 'protocol::ipa_prf::boolean_ops::share_conversion_aby::tests::test_malicious_convert_to_fp25519' panicked at library/core/src/panicking.rs:229:5:
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
2024-09-05T19:21:38.867473Z  INFO stall_detector{role=H3}: ipa_core::helpers::gateway::stall_detection::gateway: new
2024-09-05T19:21:42.730417Z  INFO ipa_core::protocol::context::batcher: batch 0 is ready for validation
2024-09-05T19:21:42.732472Z  INFO ipa_core::protocol::context::batcher: batch 1 is ready for validation
thread 'tokio-runtime-worker' panicked at ipa-core/src/protocol/context/batcher.rs:129:13:
assertion `left == right` failed: Batches should be processed in order. Batch 1 is ready for validation, but the first batch is 0.
  left: 1
 right: 0
```

The failure is relevant to the following comment I had left in the code:

```
// I am not sure if this is okay, or if we need to tolerate batch validation requests
// arriving out of order. (If we do, I think we would still want to actually fulfill
// the validations in order.)
```

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -600,6 +601,11 @@ pub trait DZKPValidator: Send + Sync {
 
     fn context(&self) -> Self::Context;
 
+    /// Sets the validator's total number of records field. This is required when using
+    /// the validate_record API, if it wasn't already set on the context used to create
```
akoshelev, line 605: I was thinking that we could always set the total records for validator from the context and panic if context is not prepared correctly - do you remember why we need to support this use case (I assume we still don't expose batcher API to the protocol code, otherwise the point of this comment is moot)

andyleiserson: It is needed for attribution, which has a variable number of records per step. The rate of progression through records is uniform, which is required by the `validate_record` API. But the number of records per step, which for attribution is "number of users having that row number", varies.

Since it is not allowed to overwrite `total_records` with a new value once it has been set, it can't be set when creating the validator.

akoshelev: ah, right - for hybrid we may get rid of it

akoshelev (approved):

andyleiserson:
The problem with out-of-order validation (the CI failure) applies to #1250 as well. I added the fix there.

PR #1253, opened 2024-09-03
Hash Owned Values

danielmasny:
allow hashing iterators on owned values + test

Previously, we could only hash iterators over references. This would make it necessary to collect an iterator in a vector and return the vector to the compute_hash function. By allowing hashing owned values, we do not need to collect the iterator into a vector

### DIFF
--- a/ipa-core/src/helpers/hashing.rs
+++ b/ipa-core/src/helpers/hashing.rs
@@ -1,4 +1,4 @@
-use std::convert::Infallible;
+use std::{borrow::Borrow, convert::Infallible};
 
 use generic_array::GenericArray;
 use sha2::{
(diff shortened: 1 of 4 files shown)


akoshelev:
I don't like the callsites of `compute_hash` but there seem to be no other way to provide this functionality. Trait coherence rules sometimes can be very limiting, I'd love to make `compute_hash` implementation more complicated and avoid `<_, _, _>` , but my experiments yielded no results

akoshelev (approved):

andyleiserson:
How about this: https://github.com/danielmasny/ipa/compare/owned-hashing...andyleiserson:ipa:dm-owned-hashing?expand=1

danielmasny:
I am happy with your solution Andy. Thanks! Should we merge it in this pull request or do we create a new one?

PR #1255, opened 2024-09-04
Shuffle verification

danielmasny:
builds on top of https://github.com/private-attribution/ipa/pull/1253, https://github.com/private-attribution/ipa/pull/1249 and https://github.com/private-attribution/ipa/pull/1246.

We should first merge those before reviewing this PR.


ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,319 @@
...
+    let mut keys = ctx
+        .parallel_join(key_shares.iter().enumerate().map(|(i, key)| async move {
+            malicious_reveal(ctx.clone(), RecordId::from(i), None, key).await
+        }))
+        .await?
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>();
+    // add a one, since last row element is tag which is not multiplied with a key
+    keys.push(iter::once(Gf32Bit::ONE).collect());
```
akoshelev, line 269: this may cause reallocation, it should be better to collect everything at once like this

```rust
    // reveal MAC keys
    let keys = ctx
        .parallel_join(key_shares.iter().enumerate().map(|(i, key)| async move {
            malicious_reveal(ctx.clone(), RecordId::from(i), None, key).await
        }))
        .await?
        .into_iter()
        .flatten()
        // add a one, since last row element is tag which is not multiplied with a key
        .chain(iter::once(StdArray::from_fn(|_| Gf32Bit::ONE)))
        .collect::<Vec<_>>();

```

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    Ok(())
+}
+
+/// This function computes for each item in the iterator the inner product with `keys`.
+/// It concatenates all inner products and hashes them.
+///
+/// ## Panics
+/// Panics when conversion from `BooleanArray` to `Vec<Gf32Bit` fails.
+fn compute_row_hash<S, B, I>(keys: &[StdArray<Gf32Bit, 1>], row_iterator: I) -> Hash
```
akoshelev, line 229: you probably want similar optimizations as for `compute_hash` to avoid turbofish syntax. Maybe in this case it could be easier because you really want `S` and not `&S` (you clone it inside), so you can ask for it

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,319 @@
...
+        .narrow(&OPRFShuffleStep::HashH2toH1)
+        .set_total_records(TotalRecords::specified(1)?);
+    let channel_h3 = &h3_ctx.recv_channel::<Hash>(ctx.role().peer(Direction::Left));
+    let channel_h2 = &h2_ctx.recv_channel::<Hash>(ctx.role().peer(Direction::Right));
+
+    // receive hashes
+    let (hashes_h3, hash_h2) = try_join(
+        h3_ctx.parallel_join(
+            (0usize..=1).map(|i| async move { channel_h3.receive(RecordId::from(i)).await }),
+        ),
```
akoshelev, line 98: this is better written as try_join3 because you are receiving two records. Parallel join will be more expensive to use in this setting

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,319 @@
...
+            .map(|share| share.left() + share.right()),
+    );
+
+    // setup channels
+    let h3_ctx = ctx
+        .narrow(&OPRFShuffleStep::HashesH3toH1)
+        .set_total_records(TotalRecords::specified(2)?);
+    let h2_ctx = ctx
+        .narrow(&OPRFShuffleStep::HashH2toH1)
+        .set_total_records(TotalRecords::specified(1)?);
```
akoshelev, line 90: nit - just to save one extra question mark

```suggestion
        .set_total_records(TotalRecords::ONE);
```

akoshelev (approved):
good stuff, I think we should definitely write a test to validate malicious behavior (hopefully the existing framework makes it easy)

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,319 @@
...
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>();
+    // add a one, since last row element is tag which is not multiplied with a key
+    keys.push(iter::once(Gf32Bit::ONE).collect());
+    Ok(keys)
+}
+
+#[cfg(all(test, unit_test))]
+mod tests {
```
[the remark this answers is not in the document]
danielmasny: https://github.com/private-attribution/ipa/issues/1262

PR #1257, opened 2024-09-05
Compute tags for malicious shuffle

danielmasny:
compute the MAC tags for the malicious shuffle function, builds on top of https://github.com/private-attribution/ipa/pull/1246


ipa-core/src/ff/boolean_array.rs
```diff
@@ -41,9 +48,28 @@ impl<A> BooleanArray for A where
...
 
+impl<A> TryInto<Vec<AdditiveShare<Gf32Bit>>> for &AdditiveShare<A>
+where
+    A: BooleanArray,
+    AdditiveShare<A>: Sized,
+{
+    type Error = Error;
+
+    fn try_into(self) -> Result<Vec<AdditiveShare<Gf32Bit>>, Self::Error> {
+        let left_shares: Vec<Gf32Bit> = self.left().try_into()?;
```
akoshelev, line 63: it is sad that we need 3 allocations here - I missed that. 

Is there a chance we could define a function on `BA` type that returns an iterator over `Gf32Bit` values instead of defining `TryFrom` implementation? This won't require trait bounds and callers can decide when to allocate. In this case, we would just zip two iterators and allocate at the end

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,237 @@
...
+                        p_ctx.clone(),
+                        RecordId::from(i * row_length + j),
+                        row_entry,
+                        key,
+                    )
+                    .await
+                },
+            ))
+            .await
+            .unwrap()
```
akoshelev, line 54:
```suggestion
?
```

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,237 @@
...
+    keys: &[AdditiveShare<Gf32Bit>],
+    rows: &[AdditiveShare<S>],
+) -> Result<Vec<AdditiveShare<B>>, Error> {
+    let length = rows.len();
+    let row_length = keys.len();
+    let tag_ctx = ctx.set_total_records(TotalRecords::specified(length * row_length)?);
+    let p_ctx = &tag_ctx;
+
+    let futures = rows.iter().enumerate().map(|(i, row)| async move {
+        let row_entries: Vec<AdditiveShare<Gf32Bit>> = row.try_into().unwrap();
```
akoshelev, line 39:
```suggestion
        let row_entries: Vec<AdditiveShare<Gf32Bit>> = row.try_into()?;
```

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -0,0 +1,237 @@
...
+/// ## Panics
+/// When conversion fails or when `S::Bits + 32 != B::Bits`.
+async fn compute_and_add_tags<C: Context, S: BooleanArray, B: BooleanArray>(
+    ctx: C,
+    keys: &[AdditiveShare<Gf32Bit>],
+    rows: &[AdditiveShare<S>],
+) -> Result<Vec<AdditiveShare<B>>, Error> {
+    let length = rows.len();
+    let row_length = keys.len();
+    let tag_ctx = ctx.set_total_records(TotalRecords::specified(length * row_length)?);
```
akoshelev, line 35: this will blow up if arrays are of size 0. It is fine to check these outside of this function, just make sure the check is there somewhere - we had already a case where shuffle received 0 rows and failed in a random place.

I would generally just slap `debug_assert` if I expect that check to be somewhere up the stack

ipa-core/src/ff/boolean_array.rs
danielmasny: I see. That makes sense and should work.

ipa-core/src/ff/boolean_array.rs
```diff
@@ -799,6 +820,41 @@ macro_rules! boolean_array_impl {
     };
 }
 
+macro_rules! boolean_array_impl_large {
+    ($modname:ident, $name:ident, $bits:tt, $deser_type:tt, $arraylength:ty) => {
+        boolean_array_impl!($modname, $name, $bits, $deser_type);
+
+        // used to convert into Fp25519
+        impl From<(u128, u128)> for $name {
```
akoshelev, line 828: I don't think you can define this conversion for any boolean array. While it works for <32 byte arrays, it will panic at runtime.

Andy added a bunch of `const_assert` helpers and I think you can use them here to reject invocations for arrays > 32 bytes in size

akoshelev (approved):

PR #1258, opened 2024-09-05
Removed keygen from Dockerfile, running docker build on every push

cberkhoff:


.github/workflows/docker.yml
```diff
@@ -3,13 +3,14 @@ name: Publish Helper Image CI
...
   workflow_dispatch:
   push:
+    branches:
+      - master
     paths:
       - "docker/ci/**"
       - ".github/workflows/docker.yml"
-      - "Cargo.toml"
-  schedule:
-    # Once monthly at a randomly selected time.
```
akoshelev, line 11: is it bad to keep running it this way?

eriktaubeneck (approved):
lgtm

.github/workflows/docker.yml
```diff
@@ -3,13 +3,14 @@ name: Publish Helper Image CI
 on:
   workflow_dispatch:
   push:
+    branches:
+      - master
```
akoshelev, line 7: do we have `master` branch?

docker/helper.Dockerfile
```diff
@@ -14,20 +13,7 @@ RUN set -eux; \
...
-# make sure these names are consistent with the ones defined in CliPaths trait: src\cli\paths.rs
-RUN set -eux; \
-    mkdir -p $CONF_DIR/pub; \
-    $HELPER_BIN_PATH keygen \
-    --name $HOSTNAME \
-    --tls-cert $CONF_DIR/pub/h$IDENTITY.pem \
-    --tls-key $CONF_DIR/h$IDENTITY.key \
-    --mk-public-key $CONF_DIR/pub/h${IDENTITY}_mk.pub \
-    --mk-private-key $CONF_DIR/h${IDENTITY}_mk.key
+COPY --from=builder ${SOURCES_DIR}/target/release/helper $HELPER_BIN_PATH
```
akoshelev, line 19: nit: newline

docker/helper.Dockerfile
```diff
@@ -14,20 +13,7 @@ RUN set -eux; \
...
-ARG IDENTITY
-ARG HOSTNAME
 ARG SOURCES_DIR
 
 RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
-COPY --from=builder ${SOURCES_DIR}/target/release/helper $HELPER_BIN_PATH
-
-# generate certificate/private key for TLS
-# make sure these names are consistent with the ones defined in CliPaths trait: src\cli\paths.rs
-RUN set -eux; \
```
akoshelev, line 26: what's the motivation behind these changes?

.github/workflows/docker.yml
cberkhoff: Will change to main

docker/helper.Dockerfile
cberkhoff: The idea is to use this docker inside the Kubernetes cluster. We want the key generation to happen before the Docker is run.

.github/workflows/docker.yml
cberkhoff: Bad how? The idea is to create a fresh new Docker on each merge, not once a month randomly. I envision this Docker going to a beta stage.

akoshelev: My understanding is that this is complementary to each merge. If for some reason merges are not happening, it is good to keep generating images to catch regressions in the infrastructure

cberkhoff: So, this is just a start. I think in the future it might be better to tag each build with the commit_id AND with latest so that we can pin prod to a specific build and beta can just get the latest.

eriktaubeneck: when I start working on CI, we'll actually want to build every commit for PRs too. but we can sort that out down the road

.github/workflows/docker.yml
```diff
@@ -34,5 +49,6 @@ jobs:
         context: .
         file: ./docker/ci/helper.Dockerfile
         push: true
-        tags: ghcr.io/${{ github.repository }}/ipa-helper:latest
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
```
akoshelev, line 53: Does it work when `labels` section is not defined (also outputs seems to be implicit)

akoshelev (approved):

.github/workflows/docker.yml
cberkhoff: It ran fine for me here: https://github.com/cberkhoff/ipa/actions/runs/10744939741/job/29802940811

PR #1259, opened 2024-09-05
Malicious Shuffle

danielmasny:
Malicious Shuffle function


ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -223,16 +323,23 @@ async fn h3_verify<C: Context, S: BooleanArray>(
 ///
 /// ## Panics
 /// Panics when conversion from `BooleanArray` to `Vec<Gf32Bit` fails.
-fn compute_row_hash<S, I>(keys: &[StdArray<Gf32Bit, 1>], row_iterator: I) -> Hash
+fn compute_row_hash<S, B, I>(keys: &[StdArray<Gf32Bit, 1>], row_iterator: I) -> Hash
```
andyleiserson, line 326: Getting a `StdArray` is an artifact of using the low-level reveal directly. I think it would be better to do `Gf32Bit::from_array` immediately upon reveal and work with regular values, rather than pass the array type around to a bunch of other code.

I also think it would be worth adding a comment at the call to `malicious_reveal` explaining why that is done here (because shuffle is a bespoke malicious protocol thus can't rely on the context to select appropriately between semi-honest and malicious reveal).

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -27,16 +29,110 @@ use crate::{
...
+/// ## Panics
+/// Panics when `S::Bits + 32 != B::Bits` or type conversions fail.
+pub async fn malicious_shuffle<C, S, B, I>(
+    ctx: C,
+    shares: I,
+) -> Result<Vec<AdditiveShare<S>>, Error>
+where
+    C: Context,
+    S: BooleanArray,
+    B: BooleanArray,
```
andyleiserson, line 46: Not for this PR, but a `MaliciousShufflable` trait could simplify the trait bounds related to `S` and `B`.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -27,16 +29,110 @@ use crate::{
     seq_join::seq_join,
 };
 
+/// This function executes the maliciously secure shuffle protocol on the input: `shares`.
+///
+/// ## Errors
+/// Propagates network, multiplication and conversion errors from sub functions.
+///
+/// ## Panics
+/// Panics when `S::Bits + 32 != B::Bits` or type conversions fail.
```
andyleiserson, line 38: It would be clearer to check this property on entry to `malicious_shuffle`, rather than relying on the eventual call to `concatenate_row_and_tags` to check this property -- this captures the API contract in one place even if the implementation gets refactored in the future.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -27,16 +29,110 @@ use crate::{
...
+    for<'a> &'a B: Add<&'a B, Output = B>,
+    Standard: Distribution<B>,
+{
+    // compute amount of MAC keys
+    let amount_of_keys: usize = (usize::try_from(S::BITS).unwrap() + 31) / 32;
+    // // generate MAC keys
+    let keys = (0..amount_of_keys)
+        .map(|i| ctx.prss().generate_fields(RecordId::from(i)))
+        .map(|(left, right)| AdditiveShare::new(left, right))
+        .collect::<Vec<AdditiveShare<Gf32Bit>>>();
```
andyleiserson, line 60: I believe you can directly generate `AdditiveShare<Gf32Bit>`.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
andyleiserson, line 326: The name `row_iterator` is used for two different things in this function -- an iterator over rows, and an iterator within a row. Maybe `rows_iterator` and `row_iterator`? Or `row_iterator` and `col_iterator`?

There might also be a better name for the  function. Maybe `compute_and_hash_tags` (contrast with `compute_and_add_tags`) or `hash_rows_and_tags`?

Another difference with `compute_and_add_tags` is whether or not the tags are secret shared. I don't think that has to be reflected in the name (although it could be), but maybe is worth calling out in a comment.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -282,19 +389,28 @@ async fn reveal_keys<C: Context>(
...
-    let length = rows.len();
+    rows: I,
+) -> Result<Vec<AdditiveShare<B>>, Error>
+where
+    C: Context,
+    S: BooleanArray,
+    B: BooleanArray,
+    I: IntoIterator<Item = AdditiveShare<S>>,
+    I::IntoIter: ExactSizeIterator,
+    <I as IntoIterator>::IntoIter: Send,
```
andyleiserson, line 403: These could be combined into one bound.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -354,12 +470,138 @@ mod tests {
...
+
+            let result_ba = BA112::deserialize_from_slice(&result[0].as_raw_slice()[0..14]);
+
+            assert_eq!(record, result_ba);
+
+            let tag = <BA112 as TryInto<Vec<Gf32Bit>>>::try_into(record)
+                .unwrap()
+                .iter()
+                .zip(keys)
+                .fold(Gf32Bit::ZERO, |acc, (entry, key)| acc + *entry * key);
```
andyleiserson, line 521: Not necessarily worth changing, since this seems pretty clearly correct, but this is effectively the same logic used by the implementation. As a general principle, I would favor a test that avoids replicating the implementation, e.g. by doing an explicit calculation of the inner product.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -354,12 +470,138 @@ mod tests {
...
+
+            let tag_mpc = <BA32 as TryInto<Vec<Gf32Bit>>>::try_into(BA32::deserialize_from_slice(
```
andyleiserson, line 523:
```suggestion
            let tag_mpc = Vec::<Gf32Bit>::try_from(BA32::deserialize_from_slice(
```
and similarly above.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -354,12 +470,138 @@ mod tests {
...
+
+    /// This test checks the correctness of the malicious shuffle.
+    /// It does not check the security against malicious behavior.
+    #[test]
+    fn check_shuffle_correctness() {
+        const RECORD_AMOUNT: usize = 10;
+        run(|| async {
+            let world = TestWorld::default();
+            let mut rng = thread_rng();
+            // using Gf32Bit here since it implements cmp such that vec can later be sorted
```
andyleiserson, line 539: You can sort any BA up to 128 bits with `records.sort_by_key(BA32::as_u128)`. This saves the effort of converting back and forth with `Gf32Bit`, and makes it pretty easy to add a correctness check to the `BA112` test. (Or perhaps to combine the tests into a generic test using `S: BooleanArray + U128Conversions`.)

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -531,4 +782,116 @@ mod tests {
...
+                .semi_honest(records.into_iter(), |ctx, shares| async move {
+                    malicious_shuffle::<_, BA32, BA64, _>(ctx, shares).await
+                })
+                .await;
+
+            // x2 should be consistent with y2
+            let _ = h2.unwrap();
+
+            // but this should fail
+            let _ = h1.unwrap();
```
andyleiserson, line 894: I believe the `let _` are unnecessary here?

The way the test is written is fine, but FYI, you could also use `assert!(matches!(...))` to verify the error without `should_panic`.

ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -27,16 +29,110 @@ use crate::{
...
+/// `B` in bytes needs to be equal to `tag_offset + 4`.
+fn split_row_and_tag<S: BooleanArray, B: BooleanArray>(
+    row_with_tag: B,
+    tag_offset: usize,
+) -> (S, Gf32Bit) {
+    let mut buf = GenericArray::default();
+    row_with_tag.serialize(&mut buf);
+    (
+        S::deserialize(GenericArray::from_slice(&buf.as_slice()[0..tag_offset])).unwrap(),
+        Gf32Bit::deserialize(GenericArray::from_slice(&buf.as_slice()[tag_offset..])).unwrap(),
```
andyleiserson, line 122: I'm not sure the unwrap of `S::deserialize` is okay. It will fail if `S` is not a multiple of bytes and the padding bits are non-zero. (Which means a server panic reachable from external input.)

Possibly it can be changed to `unwrap_or_default()`, and in the case that a row is replaced with the default (zero), the verification fails whp. If that's sound, please note the justification in a comment.

andyleiserson (approved):
Looks good overall, just a handful of suggestions.

PR #1261, opened 2024-09-06
Encrypted Input Buffer

eriktaubeneck:
Here's an initial draft for how we can allow for encrypted inputs in 3 distinct files.

It moves the old way of running it into a new command "oprf-ipa-test". One open question here is if we should remove the idea of plaintext vs encrypted match keys, since we now expect the full event to be encrypted for the encrypted code path.

This still needs tests, and a little clean up, but wanted to open this to get feedback on the overall approach.

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -25,7 +25,7 @@ use ipa_core::{
     ff::{boolean_array::BA32, FieldType},
     helpers::query::{DpMechanism, IpaQueryConfig, QueryConfig, QuerySize, QueryType},
     net::MpcHelperClient,
-    report::{EncryptedOprfReportFiles, DEFAULT_KEY_ID},
+    report::{EncryptedOprfReportStreams, DEFAULT_KEY_ID},
     test_fixture::{
         ipa::{ipa_in_the_clear, CappingOrder, IpaQueryStyle, IpaSecurityModel, TestRawDataRecord},
         EventGenerator, EventGeneratorConfig,
@@ -322,7 +322,7 @@ async fn ipa(
         &encrypted_inputs.enc_input_file3,
     ];
 
-    let encrypted_oprf_report_files = EncryptedOprfReportFiles::from(files);
+    let encrypted_oprf_report_files = EncryptedOprfReportStreams::from(files);
 
     let query_config = QueryConfig {
         size: QuerySize::try_from(encrypted_oprf_report_files.query_size).unwrap(),
@@ -342,7 +342,7 @@ async fn ipa(
             // implementation, otherwise a runtime reconstruct error will be generated.
             // see ipa-core/src/query/executor.rs
             run_query_and_validate::<BA32>(
-                encrypted_oprf_report_files.stream,
+                encrypted_oprf_report_files.streams,
                 encrypted_oprf_report_files.query_size,
                 helper_clients,
                 query_id,
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -4,23 +4,30 @@ use std::{
     fmt::Debug,
     fs::{File, OpenOptions},
     io,
-    io::{stdout, Write},
+    io::{stdout, BufRead, BufReader, Write},
     ops::Deref,
     path::{Path, PathBuf},
 };
 
+use bytes::BufMut;
 use clap::{Parser, Subcommand};
 use comfy_table::{Cell, Table};
 use hyper::http::uri::Scheme;
 use ipa_core::{
     cli::{
         noise::{apply, ApplyDpArgs},
-        playbook::{make_clients, playbook_oprf_ipa, validate, validate_dp, InputSource},
+        playbook::{
+            make_clients, playbook_oprf_ipa, run_query_and_validate, validate, validate_dp,
+            InputSource,
+        },
         CsvSerializer, IpaQueryResult, Verbosity,
     },
     config::{KeyRegistries, NetworkConfig},
     ff::{boolean_array::BA32, FieldType},
-    helpers::query::{DpMechanism, IpaQueryConfig, QueryConfig, QuerySize, QueryType},
+    helpers::{
+        query::{DpMechanism, IpaQueryConfig, QueryConfig, QuerySize, QueryType},
+        BodyStream,
+    },
     net::MpcHelperClient,
     report::DEFAULT_KEY_ID,
     test_fixture::{
@@ -54,7 +61,7 @@ struct Args {
     input: CommandInput,
 
     /// The destination file for output.
-    #[arg(long, value_name = "FILE")]
+    #[arg(long, value_name = "OUTPUT_FILE")]
     output_file: Option<PathBuf>,
 
     #[command(subcommand)]
@@ -97,11 +104,27 @@ enum ReportCollectorCommand {
     },
     /// Apply differential privacy noise to IPA inputs
     ApplyDpNoise(ApplyDpArgs),
-    /// Execute OPRF IPA in a semi-honest setting
+    /// Execute OPRF IPA in a semi-honest majority setting with known test data
+    /// and compare results against expectation
+    OprfIpaTest(IpaQueryConfig),
+    /// Execute OPRF IPA in a semi-honest majority setting with unknown encrypted data
     #[command(visible_alias = "oprf-ipa")]
-    SemiHonestOprfIpa(IpaQueryConfig),
+    SemiHonestOprfIpa {
+        #[clap(flatten)]
+        encrypted_inputs: EncryptedInputs,
+
+        #[clap(flatten)]
+        ipa_query_config: IpaQueryConfig,
+    },
     /// Execute OPRF IPA in an honest majority (one malicious helper) setting
-    MaliciousOprfIpa(IpaQueryConfig),
+    /// with unknown encrypted data
+    MaliciousOprfIpa {
+        #[clap(flatten)]
+        encrypted_inputs: EncryptedInputs,
+
+        #[clap(flatten)]
+        ipa_query_config: IpaQueryConfig,
+    },
 }
 
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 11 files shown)


eriktaubeneck:
I updated the test to take this path, however it's not passing. I'm a little confused as to why. It complains about:

```
called `Result::unwrap()` on an `Err` value: FailedHttpRequest { dest: "localhost:64323", status: 500, reason: "query execution failed: failed to parse: failed to parse: The provided byte slice contains non-zero value(s) [117, 234, 118] in padding bits [20..]" }
```

We should be replicating [this test](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/cli/crypto.rs#L608) that works, however that calls the OprfIpaQuery directly:
```
OprfIpaQuery::<BA16, KeyRegistry<PrivateKeyOnly>>::new(
                    query_config,
                    private_registry,
                )
                .execute(ctx, query_size, input)
```

Is there potentially some additional padding that happens in the http layer which needs to also be added here? The `playbook_oprf_ipa` seems to essentially just add the result of `delimited_encrypt_to` to the buffer, which is what we do in `encrypt` as well.

ipa-core/tests/common/mod.rs
```diff
@@ -250,11 +251,27 @@ pub fn test_ipa_with_config(mode: IpaSecurityModel, https: bool, config: IpaQuer
...
+    let enc1 = dir.path().join("helper1.enc");
+    let enc2 = dir.path().join("helper2.enc");
+    let enc3 = dir.path().join("helper3.enc");
+
     // Run IPA
     let mut command = Command::new(TEST_RC_BIN);
     command
         .args(["--network".into(), dir.path().join("network.toml")])
-        .args(["--input-file".as_ref(), inputs_file.as_os_str()])
+        .args(["--enc-input-file1".as_ref(), enc1.as_os_str()])
```
eriktaubeneck, line 272: for now, I've just replaced the old code path with plaintext input data with the new encrypted input code path. do we want to test both paths in these integration tests?

ipa-core/src/bin/report_collector.rs
```diff
@@ -171,25 +203,150 @@ fn gen_inputs(
...
-    let input = InputSource::from(&args.input);
-    let query_type: QueryType;
-    match (security_model, &query_style) {
-        (IpaSecurityModel::SemiHonest, IpaQueryStyle::Oprf) => {
-            query_type = QueryType::OprfIpaRelaxedDpPadding(ipa_query_config);
+    let query_type = get_query_type(security_model, &query_style, ipa_query_config);
+
+    let enc1 = &args
+        .enc_input_file1
+        .as_ref()
```
eriktaubeneck, line 273: is there a better way to do this without `.as_ref()`? the compiler keeps saying that it's required for `File::open`, but it smells like there should be a cleaner way to do this.

ipa-core/src/bin/report_collector.rs
```diff
@@ -53,8 +60,20 @@ struct Args {
     #[clap(flatten)]
     input: CommandInput,
 
+    /// The encrypted input for H1
+    #[arg(long, value_name = "H1_ENCRYPTED_INPUT_FILE")]
```
akoshelev, line 64: I suggest a different approach - these parameters seem to be used only for `OprfIpa`, so it may make sense to move them insode `ReportCollectorCommand` enum. That way you can make them non-optional and get rid of `as_ref` calls when you actually want to get the value out

ipa-core/tests/common/mod.rs
akoshelev: I'd say yes - it is good to have most of our code executed in integration tests. If we support it, we should run it

ipa-core/src/bin/report_collector.rs
```diff
@@ -171,25 +203,150 @@ fn gen_inputs(
...
+                helper_clients,
+                query_id,
+                ipa_query_config,
+            )
+            .await
         }
     };
 
+    if let Some(ref path) = args.output_file {
+        let _ = write_ipa_output_file(path, &actual);
```
akoshelev, line 332: this just swallows the errors, you should use ?` instead to propagate them

ipa-core/src/bin/report_collector.rs
akoshelev: you don't need `&` but I don't think you can get away w/o using `as_ref` to borrow the content of an option. See the suggestion above that should help with getting rid of options alltogether

ipa-core/src/bin/report_collector.rs
```diff
@@ -171,25 +203,150 @@ fn gen_inputs(
...
+    } else {
+        println!("{}", serde_json::to_string_pretty(&actual)?);
+    }
+    Ok(())
```
akoshelev, line 336: this code is all red - sometimes code coverage gives us false negatives. Can you please check that we indeed are executing it (I usually just insert a panic and check that test fails)

ipa-core/tests/common/mod.rs
```diff
@@ -24,6 +24,7 @@ pub mod tempdir;
 pub const HELPER_BIN: &str = env!("CARGO_BIN_EXE_helper");
 pub const TEST_MPC_BIN: &str = env!("CARGO_BIN_EXE_test_mpc");
 pub const TEST_RC_BIN: &str = env!("CARGO_BIN_EXE_report_collector");
+pub const CRYPTO_UTIL_BIN: &str = env!("CARGO_BIN_EXE_crypto_util");
```
eriktaubeneck, line 27: it seems this works locally, but is not finding it in the github tests. is there something else I need to do to make sure this compiles for the tests?

ipa-core/src/bin/report_collector.rs
```diff
@@ -171,25 +203,150 @@ fn gen_inputs(
...
+        .expect("--enc_input_file3 is required");
+
+    let mut buffers: [_; 3] = std::array::from_fn(|_| Vec::new());
+    for (i, path) in [enc1, enc2, enc3].iter().enumerate() {
+        let file = File::open(path).unwrap_or_else(|e| panic!("unable to open file {path:?}. {e}"));
+        let reader = BufReader::new(file);
+        for line in reader.lines() {
+            let line = line.unwrap();
+            let encrypted_report_bytes = hex::decode(line.trim()).unwrap();
+            buffers[i].put_u16_le(encrypted_report_bytes.len().try_into().unwrap());
```
akoshelev, line 291: I think we should leverage`EncryptedOprfReport` struct here. It has several benefits:

* We keep the serialization logic in one module, including the knowledge about length size
* We do some basic validation checks at construction time (size of the report, etc)
* We can be reasonably confident that we will continue using it on the server side

It may not support serialization yet, but it should be straightforward to add it

akoshelev, line 291:
```suggestion
            let encrypted_report_bytes = hex::decode(line?.trim())?;
            buffers[i].put_u16_le(encrypted_report_bytes.len().try_into()?);
```

ipa-core/src/bin/report_collector.rs
```diff
@@ -171,25 +203,150 @@ fn gen_inputs(
...
+            buffers[i].put_slice(encrypted_report_bytes.as_slice());
         }
-        (IpaSecurityModel::Malicious, IpaQueryStyle::Oprf) => {
-            panic!("OPRF for malicious is not implemented as yet")
+    }
+    // Panic if input sizes are not the same
+    assert!(buffers[0].len() == buffers[1].len());
+    assert!(buffers[1].len() == buffers[2].len());
```
akoshelev, line 297:
```suggestion
    assert_eq!(buffers[0].len(), buffers[1].len());
    assert_eq!(buffers[1].len(), buffers[2].len());
```

akoshelev: or you can return an error too - usually it is better to handle user input through result and error propagation rather than panic. Not that it matters here that much, but just for consistency reasons it may be better to do it this way

akoshelev (changes requested):

ipa-core/src/bin/report_collector.rs
eriktaubeneck: sure, that make sense. it would make sense to do this in the crypto cli too then

ipa-core/src/bin/report_collector.rs
eriktaubeneck: does code coverage catch stuff covered by the integration tests?

ipa-core/tests/common/mod.rs
eriktaubeneck: Agreed that we should test what we support. Is it useful to continue to support this mode though? In the longer run, maybe we want to only support the encrypted input?

ipa-core/tests/common/mod.rs
eriktaubeneck: @andyleiserson do you know if I need to update something to get this binary to also compile?

ipa-core/src/bin/report_collector.rs
akoshelev: it should, `coverage-ci` should include integration tests

ipa-core/src/bin/report_collector.rs
eriktaubeneck: look at that, and generally errors are reserved for something that might be recoverable. this seems unrecoverable. I added a comment to that effect, but lmk if you disagree.

ipa-core/src/bin/report_collector.rs
eriktaubeneck: now that the tests work, it shows as covered!

akoshelev (approved):
This is looking pretty good. One remaining piece of feedback is to leverage rustdoc more to document new structs and functions - usually there is always something to say about them. Comments like this are pretty useless

```rust
   /// Encrypted OPRF reports
   struct EncryptedOprfReports { .. }
```

but typically you can provide some context that otherwise is lost/hidden in code

```rust
     /// Holds the references to encrypted reports written in 3 files (one per helper).
     /// Unifies the serialization logic between report collector commands that
     /// encrypt reports on the fly and ones that consume them already encrypted
     /// outside.  
   struct EncryptedOprfReports { .. }
```

ipa-core/tests/common/mod.rs
```diff
@@ -263,12 +288,23 @@ pub fn test_ipa_with_config(mode: IpaSecurityModel, https: bool, config: IpaQuer
...
-        IpaSecurityModel::SemiHonest => "semi-honest-oprf-ipa",
-        IpaSecurityModel::Malicious => "malicious-oprf-ipa",
+    let protocol = match (mode, encrypted_inputs) {
+        (IpaSecurityModel::SemiHonest, true) => "semi-honest-oprf-ipa",
+        (IpaSecurityModel::SemiHonest, false) => "semi-honest-oprf-ipa-test",
+        (IpaSecurityModel::Malicious, true) => "malicious-oprf-ipa",
+        (IpaSecurityModel::Malicious, false) => "malicious-oprf-ipa-test",
     };
+    command.arg(protocol);
+    if encrypted_inputs {
```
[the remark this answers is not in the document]
eriktaubeneck: yes agreed it's getting gnarly. unfortunately, the order of these matter. the standard args have to come before the subcommand name, and the subcommand specific ones come after, making this particularly complex. 

the report collector bin could be cleaned up, though. we shouldn't need the DP noise posthoc subcommand anymore. without that, we can remove `--input-file` and replace it with a new arg `--plaintext-input-file` that's specific to the plaintext subcommands, and then cleaning up the order here. that's better off in a separate PR, though.

PR #1263, opened 2024-09-07
 update feature flags for relaxed DP

bmcase:
simpler way to have a feature flag for relaxed padding DP parameters.  should fix draft crashing.

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -183,7 +183,7 @@ async fn ipa(
     let query_type: QueryType;
     match (security_model, &query_style) {
         (IpaSecurityModel::SemiHonest, IpaQueryStyle::Oprf) => {
-            query_type = QueryType::OprfIpaRelaxedDpPadding(ipa_query_config);
+            query_type = QueryType::OprfIpa(ipa_query_config);
         }
         (IpaSecurityModel::Malicious, IpaQueryStyle::Oprf) => {
             panic!("OPRF for malicious is not implemented as yet")
--- a/ipa-core/src/helpers/transport/query/mod.rs
+++ b/ipa-core/src/helpers/transport/query/mod.rs
@@ -198,8 +198,6 @@ pub enum QueryType {
     TestMultiply,
     #[cfg(any(test, feature = "test-fixture", feature = "cli"))]
     TestAddInPrimeField,
-    #[cfg(any(test, feature = "test-fixture", feature = "cli"))]
-    OprfIpaRelaxedDpPadding(IpaQueryConfig),
     OprfIpa(IpaQueryConfig),
 }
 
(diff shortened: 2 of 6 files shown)


bmcase:
runs on draft https://draft-mpc.vercel.app/query/view/sandy-graze2024-09-07T1228

andyleiserson:
Isn't this removing the flag? I don't see what it is replaced with.

The reason I suggested making it a conditionally-compiled `QueryType` variant was so that we could exclude relaxed DP from production builds -- that's not essential, but it was part of the discussion when we talked about this.

andyleiserson (approved):

bmcase:
It is removing the flag in all places except the following lines in `oprf_ipa.rs` where we set the padding parameters.
```
        #[cfg(any(test, feature = "cli", feature = "test-fixture"))]
        let padding_params = PaddingParameters::relaxed();
        #[cfg(not(any(test, feature = "cli", feature = "test-fixture")))]
        let padding_params = PaddingParameters::default();
```
Had to revisit this as `main` is crashing on draft runs after merging the earlier PR https://github.com/private-attribution/ipa/pull/1247. I must have missed something in adding the new `QueryType`; I'm sure there is a way to make that way work but in revisiting it it really seemed to me that we can do it more simply by just having the flags for where we set the parameters.  I think that still achieves the goal of not having relaxed be in production builds.

bmcase:
If that seems fine, then be great if anyone who has merge access can go ahead and merge so we can unbreak main.

akoshelev (approved):

eriktaubeneck:
@akoshelev it works on draft: https://draft-mpc.vercel.app/query/view/bushy-setup2024-09-09T1601

We can merge this to unbreak main, but it would be useful to better articulate the goals of this. Right now, (if I'm reading this correctly):
1. The RelaxedDp is used if `helper` is compiled with `cli` or `test-fixture` feature flags.
2. It's independent of how `report_collector` is compiled, or the query config it sends to the helpers.

What I think we actually want is:
1. `report_collector` specifies if it wants RelaxedDp or not in the query config.
2. `helper` rejects queries with RelaxedDp if it's not compiled with `cli` or `test-fixture`

I'm also unclear what the `cli` feature flag enables, and how it's related here.

akoshelev:
@eriktaubeneck I probably misread this change, it seemed to me that we just deleted relaxed padding implementation

eriktaubeneck:
@akoshelev take a look at Ben's [comment above](https://github.com/private-attribution/ipa/pull/1263#issuecomment-2336530134). It retains the switch that flips which version is used based on the feature flags (which was in the older PR).

bmcase:
> We can merge this to unbreak main, but it would be useful to better articulate the goals of this. Right now, (if I'm reading this correctly):
> 1. The RelaxedDp is used if helper is compiled with cli or test-fixture feature flags.
> 2. It's independent of how report_collector is compiled, or the query config it sends to the helpers.
>
> What I think we actually want is: 
> 1. report_collector specifies if it wants RelaxedDp or not in the query config.
> 2. helper rejects queries with RelaxedDp if it's not compiled with cli or test-fixture

I think the two goals here are 

1.  we never run with `RelaxedDp` in production (so for this reason I don't think the report collector should be able to specific it wants it) and 
2.  we want our tests to run fast enough which is why we run with RelaxedDp with `cli` or `test-fixture` tests.

PR #1271, opened 2024-09-11
remove unneeded input-data-100.txt

eriktaubeneck:
@cberkhoff I think this was added by accident


danielmasny (approved):
lgtm, Alex as confirmed that it is indeed not needed via the wire chat.

PR #1273, opened 2024-09-11
remove duplicated crypto_util test, update CI to run CLI tests

eriktaubeneck:
the deleted test in this PR is fully duplicated [here](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/cli/crypto.rs#L540), however the one in `ipa-core/src/cli/crypto.rs` wasn't running. I updated the `pre-commit` and Github config to run those tests (and only those tests), to minimize test overlap in precommit.


.github/workflows/check.yml
```diff
@@ -76,8 +76,8 @@ jobs:
       - name: Run Web Tests
         run: cargo test -p ipa-core --no-default-features --features "cli web-app real-world-infra test-fixture compact-gate"
 
-      - name: Run Integration Tests
-        run: cargo test --test encrypted_input --features "cli test-fixture web-app in-memory-infra"
+      - name: Run CLI tests
+        run: cargo test cli:: -p ipa-core --no-default-features --features "cli test-fixture web-app in-memory-infra"
```
akoshelev, line 80: we probably need to open an issue to investigate - this shouldn't be required and all tests from inside `cli` module should be exercised as part of our in-memory suite (`cargo test`)

akoshelev (approved):

.github/workflows/check.yml
eriktaubeneck: I think it's because of some combination of the feature flags here. I called these out specificly to avoid running the other ~770 tests.

akoshelev (approved):
👍

PR #1276, opened 2024-09-12
Fixing Docker builds and few improvements

cberkhoff:
Right now, if you try to run a Docker image you will get this error:

```
/bin/ipa-helper: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /bin/ipa-helper)
/bin/ipa-helper: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /bin/ipa-helper)
/bin/ipa-helper: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /bin/ipa-helper)
```
We used to have one version to build and a different one to run which led to the aforementioned error.
Before this change we had 2 Docker files, which was very confusing. Now we only have one. 

With this change I'm updating the Rust images to use a newer Debian as well as newer Docker Github actions.


eriktaubeneck (approved):
thanks for cleaning this up. makes sense to move from bullseye to bookworm, since bullseye is now in LTS.

akoshelev:
what's the size of the image after this change? Just want to make sure rust repo does not pull anything extra that leads to increase

cberkhoff:
It's 43.5MB. I think it was 48 so it came down.

PR #1279, opened 2024-09-17
Detect out-of-bounds step indices

andyleiserson:
This does a few things:
* Adds (panicking) `From<int>` impls for tuple struct steps, converts a few single-variant-enum steps to tuple struct steps, and adds (also panicking) `Step::variant(int)` constructors for enum variant indexed steps.
* Adds a check for out-of-range index in the low-level `base_index` function in the generated compact gate code. The error message suggests using the bounds-checked constructors of the previous bullet. The tuple struct member is private, so can't be constructed directly, but in the enum variant case, we can't prevent constructing it directly with an out-of-range index.
* Adds tests for the preceding two items, and a test for the panic message when invoking attribution with more than 64 records per user.

### DIFF
--- a/ipa-step-derive/src/variant.rs
+++ b/ipa-step-derive/src/variant.rs
@@ -214,6 +214,8 @@ pub struct Generator {
     arm_count: ExtendedSum,
     // This tracks the index of each item.
     index_arms: TokenStream,
+    // This tracks integer variant constructors.
+    int_variant_constructors: TokenStream,
     // This tracks the arrays of names that are used for integer variants.
     name_arrays: TokenStream,
     // This tracks the arms of the `AsRef<str>` match implementation.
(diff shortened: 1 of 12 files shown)


akoshelev (approved):
awesome! thanks for fixing it quickly!

ipa-step-derive/src/variant.rs
```diff
@@ -339,6 +341,16 @@ impl Generator {
             quote!(Self)
         };
 
+        if is_variant {
+            let constructor = format_ident!("{}", step_ident.to_string().to_snake_case());
+            self.int_variant_constructors.extend(quote! {
+                pub fn #constructor(v: #step_integer) -> Self {
+                    assert!(v < #step_integer::try_from(#step_count).unwrap());
```
akoshelev, line 348: maybe panic with a more specific message would be better here

PR #1280, opened 2024-09-18
Using new public version of Axum Server

cberkhoff:


.gitignore
```diff
@@ -11,3 +11,4 @@
 /in-market-test/hpke/bin
 /in-market-test/hpke/lib
 /in-market-test/hpke/pyvenv.cfg
+input-data-*.txt
```
akoshelev, line 14: is it related to this change?

akoshelev (approved):

.gitignore
cberkhoff: No. Just a minor measure to avoid pushing those input files to the repo by mistake.

PR #1282, opened 2024-09-18
New in the clear implementation for Hybrid

benjaminsavage:
To write tests for the "Hybrid" protocol, we need to compare the results generated to some kind of "ground truth" correct answer. This provides that.

### DIFF
--- a/ipa-core/src/test_fixture/hybrid.rs
+++ b/ipa-core/src/test_fixture/hybrid.rs
@@ -1,0 +1,127 @@
+use std::collections::{HashMap, HashSet};
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq)]
+pub enum TestHybridRecord {
+    TestImpression { match_key: u64, breakdown_key: u32 },
+    TestConversion { match_key: u64, value: u32 },
+}
+
+pub fn hybrid_in_the_clear(input_rows: &[TestHybridRecord], max_breakdown: usize) -> Vec<u32> {
+    let mut conversion_match_keys = HashSet::<u64>::new();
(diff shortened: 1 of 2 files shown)


ipa-core/src/test_fixture/hybrid.rs
```diff
@@ -0,0 +1,127 @@
...
+            TestHybridRecord::TestImpression { match_key, .. } => {
+                impression_match_keys.insert(*match_key);
+            }
+            TestHybridRecord::TestConversion { match_key, .. } => {
+                conversion_match_keys.insert(*match_key);
+            }
+        }
+    }
+
+    let mut attributed_conversions = HashMap::<u64, (u32, u32), _>::new();
```
eriktaubeneck, line 24:
```suggestion
        // this HashMap uses the `match_key` as it's key, and `(breakdown_key, value)` as it's value
        // below, we aggregate values for each entry in this HashMap
        let mut attributed_conversions = HashMap::<u64, (u32, u32), _>::new();
```

ipa-core/src/test_fixture/hybrid.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+    }
+
+    let mut output = vec![0; max_breakdown];
+    for (_, (breakdown_key, value)) in attributed_conversions {
+        output[usize::try_from(breakdown_key).unwrap()] += value;
+    }
+
+    return output;
```
eriktaubeneck, line 55:
```suggestion
    output
```

eriktaubeneck (approved):

PR #1283, opened 2024-09-18
Hybrid and sharded suffle query types

akoshelev:
This is the first step in getting our query infrastructure ready to handle hybrid protocol requests. It adds `Hybrid` query type that is not implemented.

At the same time, we are about to start testing our sharding infrastructure and we need a query type for that as well. @cberkhoff asked for this change so it is easier to put them together


cberkhoff (approved):

akoshelev:
code coverage misses are expected because it is just a skeleton

PR #1284, opened 2024-09-18
Creating an event gen script for Hybrid

benjaminsavage:


ipa-core/src/test_fixture/hybrid_event_gen.rs
```diff
@@ -0,0 +1,241 @@
...
+        Self {
+            config,
+            rng,
+            in_flight: Vec::with_capacity(max_capacity),
+        }
+    }
+
+    fn gen_batch(&mut self) {
+        match self.config.conversion_distribution {
+            ConversionDistribution::OnlyImpressions => {
```
akoshelev, line 86: I think we should test these branches too if we plan to offer them. Alternatively we could remove it and add it with tests when we need them

akoshelev (approved):
apart from coverage, this looks good to me

akoshelev (approved):

PR #1293, opened 2024-09-20
Check for large increases in step count

andyleiserson:


akoshelev:
nice! It would be cool if we could also catch long running tests in our CI - another source of pain

akoshelev (approved):

PR #1294, opened 2024-09-20
Move decrypt/encrypt functions to their own modules

akoshelev:
We will add secret sharing, so crypto module will grow in size. It makes sense to keep things in their own modules now.

It also does a bit of a refactoring to prepare to encrypt something that is not `IpaPrivateData`, but no new functionality is added by this change.


ipa-core/src/cli/crypto/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+    use crate::{
+        cli::CsvSerializer,
+        hpke::{IpaPrivateKey, IpaPublicKey},
+        test_fixture::{ipa::TestRawDataRecord, EventGenerator, EventGeneratorConfig},
+    };
+
+    /// Keys that are used in crypto tests
+    #[derive(Clone)]
+    pub(super) struct TestKeys {
+        inner: [(IpaPublicKey, IpaPrivateKey); 3],
```
eriktaubeneck, line 24: wouldn't `keys` be more descriptive than `inner`? maybe `TestKeys.keys` is too repetitive...

ipa-core/src/cli/crypto/encrypt.rs
```diff
@@ -0,0 +1,269 @@
...
+            .unwrap();
+    }
+
+    #[test]
+    #[should_panic = "invalid length 2, expected an array of length 3"]
+    fn encrypt_incomplete_network_file() {
+        let input_file = sample_data::write_csv(sample_data::test_ipa_data().take(10)).unwrap();
+
+        let output_dir = tempdir().unwrap();
+        let network_data = r#"
```
eriktaubeneck, line 252: should this use the `format!` macro used below? I didn't know that existed when I wrote this

eriktaubeneck (approved):

ipa-core/src/cli/crypto/mod.rs
akoshelev: I am probably abusing `inner` too much, I agree that `keys` is better if we add anything else inside after. I'll stick with `key_pairs`

ipa-core/src/cli/crypto/encrypt.rs
akoshelev: we could use it, but I thought for this test we don't need to inject the correct keys, so the benefit of using `format!` is not clear

PR #1299, opened 2024-09-20
Making this test less flaky, and good error reports

benjaminsavage:


ipa-core/src/test_fixture/hybrid_event_gen.rs
```diff
@@ -175,10 +175,25 @@ mod tests {
 
     #[test]
     fn default_config() {
+        const EXPECTED_HISTOGRAM_WITH_TOLERANCE: [(i32, f64); 12] = [
```
eriktaubeneck, line 178: I think this tolerance is a %, but it would be nice to have a comment.

ipa-core/src/test_fixture/hybrid_event_gen.rs
```diff
@@ -200,44 +215,25 @@ mod tests {
...
-        assert!(
-            (0 - histogram[11]).abs() < 10,
-            "expected {:?} unmatched events, got {:?}",
-            0,
-            histogram[11]
-        );
+        for (actual, (expected, tolerance)) in
+            zip(histogram, EXPECTED_HISTOGRAM_WITH_TOLERANCE.iter())
+        {
+            let max_tolerance = (*expected as f64) * tolerance + 10.0;
```
eriktaubeneck, line 221: why + 10?

ipa-core/src/test_fixture/hybrid_event_gen.rs
```diff
@@ -200,44 +215,25 @@ mod tests {
...
+            assert!(
+                (expected - actual).abs() as f64 <= max_tolerance,
+                "expected {:?} unmatched events, got {:?}",
```
eriktaubeneck, line 224: it might be worth reframing this around "expected {:?} +/- {tolerance}"

ipa-core/src/test_fixture/hybrid_event_gen.rs
```diff
@@ -279,40 +275,14 @@ mod tests {
...
-
-        assert!(
-            (1_718 - histogram[11]).abs() < 100,
-            "expected {:?} unmatched events, got {:?}",
-            1_718,
-            histogram[11]
-        );
+        for (expected, actual) in zip(EXPECTED_HISTOGRAM.iter(), histogram) {
+            assert!(
+                (expected - actual).abs() <= expected / 20 + 10,
```
eriktaubeneck, line 280: something explaining this math would be helpful

eriktaubeneck (approved):

ipa-core/src/test_fixture/hybrid_event_gen.rs
benjaminsavage: It's another way to prevent flaky failures for the smallish expected outputs. Basically, the variance is WAY higher for these rare events. Adding a constant factor is effectively another way of increasing the tolerance for the small stuff.

PR #1302, opened 2024-09-21
Moving to reveal based aggregation

benjaminsavage:


benjaminsavage:
Ran a draft successfully with 100,000 rows: https://draft-mpc.vercel.app/query/view/slow-room2024-09-21T0455

benjaminsavage:
Still seems to stall out on 1 million: https://draft-mpc.vercel.app/query/view/sole-scar2024-09-21T0501

akoshelev:
Works for 100k: https://draft-mpc.vercel.app/query/view/holy-joist2024-09-21T1853

akoshelev:
stalls for 500k: https://draft-mpc.vercel.app/query/view/iron-hop2024-09-21T1917

akoshelev:
I am testing a fix in #1306 - was able to get past 500k records, running 1M now: https://draft-mpc.vercel.app/query/view/frank-snow2024-09-23T0241

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -505,13 +505,14 @@ where
...
-    let num_outputs = input_rows.len() - histogram[0];
     let mut dzkp_validator = sh_ctx.clone().dzkp_validator(
         MaliciousProtocolSteps {
             protocol: &Step::Attribute,
             validate: &Step::AttributeValidate,
         },
-        chunk_size,
+        // The size of a single batch should not exceed the active work limit,
+        // otherwise it will stall
+        std::cmp::min(sh_ctx.active_work().get(), chunk_size),
```
akoshelev, line 515: @andyleiserson does this change make sense to you? I had to change it because otherwise I was getting stalls even on very small inputs, when `active_work` is larger than `chunk_size`.

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
andyleiserson: Is this expected to interact with #1307? I would expect active work (even at 32k) to be quite a bit less than the target proof size. I believe there are only ~50 multiplications per record in attribution? Which would mean a chunk size of 1M for a proof size of 50M.

akoshelev: Sort of. If the chunk size is larger than active work in this case(I think it may only apply to `validated_seq_join` though), then we are going to stall, because we won't schedule enough futures to fill up the batch and `validate_record` will prevent `seq_join` from scheduling more futures.

so it is crucial that `validated_seq_join` populates the whole batch and validates it inside one active work window

andyleiserson: Oof yeah. I feel like we talked about this in July and then I clearly forgot about it.

I just looked at the original malicious share conversion code (before all the `validate_record` stuff) and it has two seq_joins, one before the reveal, and one after the reveal.

PR #1303, opened 2024-09-21
more aggressively turn on relaxed-dp feature

eriktaubeneck:
the GitHub checks are very slow. fingers crossed this may help speed them up. with this, we use the relaxed-dp feature everywhere except integration tests.


eriktaubeneck:
I'm not sure this made it any faster :(

PR #1304, opened 2024-09-22
Add a handler to kill a query

akoshelev:
In our testing with external parties, it became quite evident that bouncing helpers if query fails is far from having good experience for people who run the test. We need to have an ability to kill queries that failed on one or more helper and restart the runs.

This change does exactly that - adding a `/query_id/kill` handler that will respond with immediately aborting the task that runs an IPA query.

### How to terminate a query?

if helper address is `helper.com`, one can terminate a query by sending a POST request to 

```
http(s)://helper.com/query/0/kill
```

For local testing it can be done easily with `curl`

```
curl -X POST http://localhost:3002/query/0/kill
```

if query is killed successfully, the following response is provided by the helper

```json
{"query_id":"0","status":"killed"}
```

### DIFF
--- a/ipa-core/src/net/http_serde.rs
+++ b/ipa-core/src/net/http_serde.rs
@@ -569,12 +569,12 @@ pub mod query {
 
         impl Request {
             /// see above for the reason why this needs to be behind a feature flag
-            #[cfg(all(test, not(feature = "shuttle")))]
+            #[cfg(all(test, unit_test))]
             pub fn new(query_id: QueryId) -> Self {
                 Self { query_id }
             }
 
-            #[cfg(all(test, not(feature = "shuttle")))]
+            #[cfg(all(test, unit_test))]
             pub fn try_into_http_request(
                 self,
                 scheme: axum::http::uri::Scheme,
--- a/ipa-core/src/net/http_serde.rs
+++ b/ipa-core/src/net/http_serde.rs
@@ -542,7 +542,6 @@ pub mod query {
             protocol::QueryId,
         };
 
-        #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
         pub struct Request {
             pub query_id: QueryId,
         }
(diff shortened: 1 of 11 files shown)


richajaindce:
Thanks for adding this!! It's going to be really helpful. Can you also add an example of how to run this.

Also, (n00b qn) I am supposing we would get a query id whenever a query is run to call this API?

akoshelev:
> Thanks for adding this!! It's going to be really helpful. Can you also add an example of how to run this.
> 
> Also, (n00b qn) I am supposing we would get a query id whenever a query is run to call this API?

good point! I've updated the overview section to show how to terminate a query. Please let me know if something is not clear

ipa-core/src/net/http_serde.rs
```diff
@@ -533,4 +533,78 @@ pub mod query {
...
+                NoStep
+            }
+
+            fn extra(&self) -> Self::Params {
+                String::new()
+            }
+        }
+
+        impl Request {
+            /// see above for the reason why this needs to be behind a feature flag
```
eriktaubeneck, line 571: why are these different feature flags? does `cli` need to be included, so that this is available when run from the cli as a helper?

eriktaubeneck (approved):
thank you! this is awesome. 

it would be worthwhile to also add a subcommand to the report collector binary that calls this on all helpers in a network file, but I can take that as a work item in follow up.

akoshelev:
> it would be worthwhile to also add a subcommand to the report collector binary that calls this on all helpers in a network file, but I can take that as a work item in follow up.

Great idea, this can totally be a separate PR to keep things manageable

ipa-core/src/net/http_serde.rs
akoshelev: yea, I changed the comment because it didn't really explain why those specific features were used to gate

Issue #1300, opened 2024-09-20
Malicious IPA with DP stalls on 10 record input


danielmasny:
We've identified a few issues here. Correctness is the first one:

When we compute `read_size` inside Infra, we set it this way

```rust
     std::cmp::min(
                    total_capacity,
                    // closest multiple of record_size to read_size
                    gateway_config.read_size.get() / record_size * record_size,
                )
```

the value used in config is `2048`. For PRF evaluations, record size is set to 512 (Vectorization factor of 16 for 32 byte records). This means, `read_size` is 4 records. For inputs that are not a multiple of 4, PRF will stall because `active_work` is set to minimum of input size or 32k. 

For example, input size 10 will trigger sends for first two batches and then everything will hang because third batch will be half full. `validate_record` blocks the futures inside the seq_join until we resolve 10, but we could only resolve 8.  

The proper fix would be to account for `total_records` when computing `read_size`. 

Another issue is performance. When adding dummies to small inputs, we significantly increase the total number of records processed by the protocol, but we never adjust the active work. If we add 600 dummies to 5 row input, our active work would be still 5.

andyleiserson:
Any objections to closing this?

akoshelev:
Let me capture the active work problem separately and this one can be closed

PR #1307, opened 2024-09-23
Fix send buffer misalignment issue

akoshelev:
The side-effect of this issue is described in #1300. The issue was how `Gateway` set up the read size compared to total capacity of a single buffer.

As a reminder, there are three configurable values inside each send buffer:

* *Total capacity*, bytes. This describes how wide the send buffer is and has a direct relationship to active work. Total capacity must be able to accommodate all in-flight items inside the buffer. 
* *Record size*, bytes. How many bytes a single record takes on this buffer. Every send buffer is homogeneous, i.e. records must be of the same size.
* `Read size`, bytes. How many bytes are taken off this buffer by the network layer at once. It waits until there are at least that many bytes filled up from the head of the buffer, before taking them off and releasing this capacity.

What was missing before this change is that **read size must be aligned with total capacity**.  If it is not, then buffer can be left half-full, which was happening in #1300.

The failure scenario was setting up the record size to 512 (record * vectorization $$32 \times 16$$) and total capacity of 10 items (5120 bytes) with default read size of 2048 bytes.

Now, total capacity in this case is $$2.5\times 2048$$ and it leads to a deadlock. First 8 records get sent, but the remaining two get stuck.

## Why was it working before?

Before DP, active work for small inputs was set to the input size and it remains unchanged for the duration of the query. So record 10 closes the channel and triggers a flush. With DP, we add more records, so the total number of items to process is greater than 10.

## The fix

We enforce read size to be aligned with total capacity. In some cases it leads to smaller batches sent down to HTTP, but it should impact small runs only

### DIFF
--- a/ipa-core/src/helpers/gateway/send.rs
+++ b/ipa-core/src/helpers/gateway/send.rs
@@ -275,12 +275,12 @@ impl SendChannelConfig {
                     gateway_config.read_size.get() / record_size * record_size,
                     min_capacity,
                 );
-                // if min capacity is not a multiple of read size.
+                // if min capacity is not a multiple of read size,
                 // we must adjust read size. Adjusting total capacity is not possible due to
                 // possible deadlocks - it must be strictly aligned with active work.
                 // read size goes in `record_size` increments to keep it aligned.
-                // rem is aligned with both capacity and read_size, so subtracting
-                // it will keep read_size and capacity aligned
+                // `record_size` is aligned with both capacity and read_size, so subtracting
+                // it will keep both of these values aligned with it.
                 // Here is an example how it may work:
                 // lets say the active work is set to 10, record size is 512 bytes
                 // and read size in gateway config is set to 2048 bytes (default value).
@@ -313,6 +313,7 @@ impl SendChannelConfig {
         debug_assert!(this.total_capacity.get() >= this.read_size.get());
         debug_assert!(this.total_capacity.get() >= this.record_size.get());
         debug_assert!(this.read_size.get() >= this.record_size.get());
+        debug_assert_eq!(0, this.read_size.get() % this.record_size.get());
 
         this
     }
--- a/ipa-core/src/helpers/gateway/send.rs
+++ b/ipa-core/src/helpers/gateway/send.rs
@@ -1,5 +1,6 @@
 use std::{
     borrow::Borrow,
+    cmp::min,
     fmt::Debug,
     marker::PhantomData,
     num::NonZeroUsize,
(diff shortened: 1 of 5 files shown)


ipa-core/src/helpers/gateway/send.rs
```diff
@@ -249,35 +250,79 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
-            }
-            .try_into()
-            .unwrap(),
+                    min_capacity,
+                );
+                // if min capacity is not a multiple of read size.
+                // we must adjust read size. Adjusting total capacity is not possible due to
+                // possible deadlocks - it must be strictly aligned with active work.
+                // read size goes in `record_size` increments to keep it aligned.
+                // rem is aligned with both capacity and read_size, so subtracting
```
benjaminsavage, line 282: What is "rem"?

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -249,35 +250,79 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+                // 1024.
+                proposed_read_size - min_capacity % proposed_read_size
+            };
+
+        // total capacity must be a multiple of both read size and record size.
+        // Record size is easy to justify: misalignment here leads to either waste of memory
+        // or deadlock on the last write. Aligning read size and total capacity
+        // has the same reasoning behind it: reading less than total capacity
+        // can leave the last chunk half-written and backpressure from active work
+        // preventing the protocols to make further progress.
```
benjaminsavage, line 300: This is too complicated. I worry that this code will become unmaintainable as this is so impenetrable that people simply will not understand it. I certainly don't understand it and we discussed it in person today and I've read this twice.

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -249,35 +250,79 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
             total_records,
-        }
+        };
+
+        // make sure we've set these values correctly.
+        debug_assert_eq!(0, this.total_capacity.get() % this.read_size.get());
+        debug_assert_eq!(0, this.total_capacity.get() % this.record_size.get());
+        debug_assert!(this.total_capacity.get() >= this.read_size.get());
+        debug_assert!(this.total_capacity.get() >= this.record_size.get());
+        debug_assert!(this.read_size.get() >= this.record_size.get());
```
benjaminsavage, line 315: Based on the comments above, I think you also want to check that:
```
debug_assert_eq!(0, this.read_size.get() % this.record_size.get());
```

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -249,35 +250,79 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+        // set to the size of one record, to trigger buffer flush on every write
+        let read_size =
+            if total_records.is_indeterminate() || gateway_config.read_size.get() <= record_size {
                 record_size
             } else {
-                std::cmp::min(
-                    total_capacity,
-                    // closest multiple of record_size to read_size
+                // closest multiple of record_size to read_size
+                let proposed_read_size = min(
```
benjaminsavage, line 274: Are you sure this isn't supposed to be `max`?

ipa-core/src/helpers/gateway/send.rs
benjaminsavage: With the addition of these two checks, I think this set of checks are equivalent to:

1. read_size = a * record_size
2. total_capacity = b * read_size
3. a * b >= active_work

Where `a` and `b` are positive integers >= 1.

Can we just find `a` and `b` like this:
```
if total_records.is_indeterminate() || gateway_config.read_size.get() <= record_size {
    a = 1;
    b = active_work;
} else {
    a = gateway_config.read_size.get() / record_size;
    b = ceil(active_work / a);
}
```

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -391,4 +436,62 @@ mod test {
...
+    /// chunks through a channel with total capacity 5120 (active work = 10 records) and read size
+    /// of 2048 bytes.
+    /// The problem was that read size of 2048 does not divide 5120, so the last chunk was not sent.
+    #[test]
+    fn total_capacity_is_a_multiple_of_read_size() {
+        let config =
+            send_config::<StdArray<BA256, 16>, 10, 2048>(TotalRecords::specified(43).unwrap());
+
+        assert_eq!(0, config.total_capacity.get() % config.read_size.get());
+        assert_eq!(config.total_capacity.get(), 10 * config.record_size.get());
```
benjaminsavage, line 450: If the number `10` on line 447 and line 450 are meant to be kept in sync, can you use a constant for them?

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -391,4 +436,62 @@ mod test {
...
+        assert!(config.total_capacity.get() >= record_size);
+        assert!(config.total_capacity.get() <= record_size * active);
+        assert!(config.total_capacity.get() >= config.read_size.get());
+        assert_eq!(0, config.total_capacity.get() % config.record_size.get());
+
+        // read size checks
+        assert!(config.read_size.get() > 0);
+        assert!(config.read_size.get() >= config.record_size.get());
+        assert_eq!(0, config.total_capacity.get() % config.read_size.get());
+        assert_eq!(0, config.read_size.get() % config.record_size.get());
```
benjaminsavage, line 483: These are duplicated between the test and the actual code. I don't think there is a need for this duplication. Just leaving them in the function seems sufficient.

andyleiserson:
Why does `read_size` need to be a multiple of the record size? Is it possible that `read_size` dividing the `active_work`-determined capacity was always the more important property? It does seem like having `read_size` be a multiple of the record size is nice, because then we don't leave small pieces of data in the outbound buffer. But as long as we don't leave a small piece of data in the buffer adjacent to an `active_work` boundary, are the small pieces of data are a problem?

In any case, I'm not sure we can get what we need without factoring `active_work`, or doing something like restricting `active_work` to a power of two so we can readily pick factors.

ipa-core/src/helpers/gateway/send.rs
andyleiserson:
```
record_size = 128
active_work = 33
a = 2048 / 128 = 16
b = ceil(33 / 16) = 3
read_size = a * record_size = 2048
total_capacity = b * read_size = 6144
```

The read size (2048) does not divide the `active_work`-determined capacity (4224), so this will hang.

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -249,35 +250,80 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+                proposed_read_size - min_capacity % proposed_read_size
+            };
+
+        // total capacity must be a multiple of both read size and record size.
+        // Record size is easy to justify: misalignment here leads to either waste of memory
+        // or deadlock on the last write. Aligning read size and total capacity
+        // has the same reasoning behind it: reading less than total capacity
+        // can leave the last chunk half-written and backpressure from active work
+        // preventing the protocols to make further progress.
+        let total_capacity = min_capacity / read_size * read_size;
```
andyleiserson, line 301:
```
record_size = 128
active_work = 33
min_capacity = 4224
proposed_read_size = 2048
read_size = 2048 - 128 = 1920
total_capacity = 3840
```

akoshelev:
I am going to abandon this PR as the change we want to make is very different. @andyleiserson and I debated several options and here are our conclusions:

* It is very hard to reason about tuning 3 different numbers (read size, active work and batch size) and keep them all aligned it not easy either. tests from #1316 showed that if at least one of them is not aligned with 2 others, we run into a deadlock
* We prefer read size to be internal to network implementation. Protocol writers should not concern themselves with this number and irrespective of our choice for optimal buffer size for HTTP, it should just work.

Based on these, we think the optimal way forward would be to **set active work to be strictly equal to ZKP batch size** for malicious contexts. Also, we **align read size with active work** inside send buffer, allowing partial sends in rare cases. These are handled entirely at the network layer, so it shouldn't be a problem for protocol code. 

This approach has a number of benefits. Firstly, it exposes only one knob to tune the active window for ZKP contexts, making it harder to run into a deadlock by writing wrong protocol code. Secondly, it keeps network stuff unexposed to the context layer, providing encapsulation and making it easy to tune/refactor later. 

Based n

PR #1315, opened 2024-09-26
Parametrize MAC-malicious contexts with ShardBinding parameter

akoshelev:
This is just a first step to enable sharded circuits to be written with malicious security. It is far from being complete, but it gradually moves us towards that goal.

This change just enables malicious contexts to support sharding. Because of type aliases, everything else from the outside of this module still uses non-sharded versions


cberkhoff (approved):

ipa-core/src/protocol/context/malicious.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
             validator::{self, BatchValidator},
             Base, Context as ContextTrait, InstrumentedSequentialSharedRandomness,
             SpecialAccessToUpgradedContext, UpgradableContext, UpgradedContext,
+            UpgradedMaliciousContext,
```
andyleiserson, line 24: It seems weird to be using both `UpgradedMaliciousContext` and `Upgraded` in this module. Is there a reason (besides avoiding writing the `NotSharded` type parameter value) to do it this way?

akoshelev: hmm, I think this is a mistake, let me try to fix it

andyleiserson (approved):

Issue #1308, opened 2024-09-23
IPA runs with 3M input fails with ZKP batcher assertion error


akoshelev:
Link to the run: https://draft-mpc.vercel.app/query/view/adult-sum2024-09-23T1639

```
2024-09-23T17:36:39.945Z | Helper3:
assertion failed: self.batches.len() <= 1
2024-09-23T17:36:39.932Z | Helper1:
2024-09-23T17:36:39.929625Z ERROR oprf_ipa_query{sz=3000000}:attribute_cap_aggregate: ipa_core::error: ThreadId(31) "tokio-runtime-worker" panicked at ipa-core/src/protocol/context/batcher.rs:251:9:
2024-09-23T17:36:39.933Z | Helper2:
2024-09-23T17:36:39.929600Z ERROR oprf_ipa_query{sz=3000000}:attribute_cap_aggregate: ipa_core::error: ThreadId(31) "tokio-runtime-worker" panicked at ipa-core/src/protocol/context/batcher.rs:251:9:
2024-09-23T17:36:39.945Z | Helper3:
assertion failed: self.batches.len() <= 1
2024-09-23T17:36:39.932Z | Helper1:
2024-09-23T17:36:39.929625Z ERROR oprf_ipa_query{sz=3000000}:attribute_cap_aggregate: ipa_core::error: ThreadId(31) "tokio-runtime-worker" panicked at ipa-core/src/protocol/context/batcher.rs:251:9:
2024-09-23T17:36:39.945Z | Helper1:
assertion failed: self.batches.len() <= 1
```

akoshelev:
@andyleiserson pointed out that we could set the batch size to an artificially large number (1B) to get past this error. 

```rust
    let validator = sh_ctx.dzkp_validator(
        MaliciousProtocolSteps {
            protocol: &Step::Aggregate,
            validate: &Step::AggregateValidate,
        },
        // 1B batch size, suboptimal. But only to test that it works for 3M
        1 << 30,
    );
```

I did that and got past 3M, but the same error (maybe from a different callsite) occurs at 4M
https://draft-mpc.vercel.app/query/view/rust-hack2024-09-24T0615

akoshelev:
```
2024-09-24T06:55:58.312Z | Helper1:
[...]

PR #1316, opened 2024-09-26
Make active_work match records_per_batch

andyleiserson:
Adjust active_work to match records_per_batch. If it is less, we will stall, since every record in the batch remains incomplete until the batch is validated. More might be okay, but making it the same seems simpler for now.

This change also modifies an existing DZKP proptest to be more extensive. I believe the test is capable of exposing both the hang when active_work is less than batch size (when run without the fix in this PR), and the hang due to read_size not dividing active work (not yet fixed).

I moved a couple tests that are doing DZKPs with various BA types from `secret_sharing::vector::impls` to the DZKP module. My intent was to adapt them to cover the interaction with record size, but I haven't done that yet. (The move is in a separate commit, although in the current state I don't think there's interesting diff on top of the move anyways.)

- [x] Record size tests
- [x] Tune proptest runtime for CI


ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -29,16 +30,24 @@ use crate::{
...
     base_ctx: MaliciousContext<'a>,
+    active_work: NonZeroUsize,
 }
 
 impl<'a> DZKPUpgraded<'a> {
     pub(super) fn new(
         validator_inner: &Arc<MaliciousDZKPValidatorInner<'a>>,
         base_ctx: MaliciousContext<'a>,
     ) -> Self {
+        // Adjust active_work to be at least records_per_batch. If it is less, we will
```
akoshelev, line 41: we should probably log the fact that we've adjusted it - maybe at the debug level

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -827,33 +827,150 @@ mod tests {
...
+        let [ab0, ab1, ab2] = join3v(futures).await;
+
+        let ab = [ab0, ab1, ab2].reconstruct();
+
+        assert_eq!(ab, if bit.into() { a } else { b });
+    }
+
+    #[tokio::test]
+    async fn select_semi_honest() {
+        test_select_semi_honest::<BA8>().await;
```
akoshelev, line 909: is it worth testing it for for weird types like BA3 and BA7 as well?

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -1021,19 +1138,24 @@ mod tests {
         Ok(())
     }
 
-    prop_compose! {
-        fn arb_count_and_chunk()((log_count, log_multiplication_amount) in select(&[(5,5),(7,5),(5,8)])) -> (usize, usize) {
-            (1usize<<log_count, 1usize<<log_multiplication_amount)
-        }
+    fn record_count_strategy() -> impl Strategy<Value = usize> {
+        prop_oneof![1usize..=512, (1usize..=9).prop_map(|i| 1usize << i)]
```
akoshelev, line 1142: I am being stupid and it is 12pm now - is it really testing $$2^{512}$$ records?

ipa-core/src/protocol/context/malicious.rs
```diff
@@ -174,13 +175,21 @@ pub(super) type MacBatcher<'a, F> = Mutex<Batcher<'a, validator::Malicious<'a, F
...
+    active_work: NonZeroUsize,
 }
 
 impl<'a, F: ExtendableField> Upgraded<'a, F> {
-    pub(super) fn new(batch: &Arc<MacBatcher<'a, F>>, ctx: Context<'a>) -> Self {
+    pub(super) fn new(batch: &Arc<MacBatcher<'a, F>>, base_ctx: Context<'a>) -> Self {
+        // Adjust active_work to be at least records_per_batch. The MAC validator
+        // currently configures the batcher with records_per_batch = active_work, which
+        // makes this adjustment a no-op, but we do it this way to match the DZKP validator.
+        let records_per_batch = batch.lock().unwrap().records_per_batch();
```
akoshelev, line 186: would it be better to assert this to make sure we don't miss the misalignment between MAC and ZKP?

akoshelev (approved):
lgtm, but it seems some tests are broken.

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: Yes, I think so. I had BA4 but took it out because it's not `boolean_vector!` (not for any good reason other than we haven't needed it). But I think it is worth having a less-than-one-byte case, and maybe even adding a new BA type so we can cover the between-one-and-two-bytes case.

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: What it does is:
* 50% of the time it picks a random power of two between 2 and 512 (9 possible values)
* 50% of the time it picks a random integer between 1 and 512 (512 possible values)

Two reasons to focus on powers of two, although I didn't put a huge amount of thought into this (e.g. it just occurred to me to add $2^0$):
* Because we plan to restrict batch sizes to powers of two.
* As the count increases, it makes sense to me to sample the space more sparsely, there's relatively less interesting about testing batch sizes 399, 400, and 401 than about testing batch sizes 7, 8, and 9.

I think it is clearer to have the two `prop_oneof` cases on different lines, but rustfmt insisted on a single line.

ipa-core/src/protocol/context/malicious.rs
andyleiserson: Sure, I'm fine with that.

ipa-core/src/protocol/context/validator.rs
```diff
@@ -217,7 +217,7 @@ impl<'a, F: ExtendableField> BatchValidator<'a, F> {
 
         // TODO: Right now we set the batch work to be equal to active_work,
         // but it does not need to be. We can make this configurable if needed.
-        let records_per_batch = ctx.active_work().get().min(total_records.get());
+        let records_per_batch = ctx.active_work().get();
```
andyleiserson, line 220: Reducing if larger than `total_records` may have been necessary with an earlier version of the batcher, but the current version should take care of that internally, so I removed it here. No relation to the rest of these changes though.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -825,35 +825,158 @@ mod tests {
...
         test_fixture::{join3v, Reconstruct, Runner, TestWorld},
     };
 
+    async fn test_select_semi_honest<V>()
+    where
+        V: BooleanArray,
+        for<'a> Replicated<V>: BooleanArrayMul<DZKPUpgradedSemiHonestContext<'a, NotSharded>>,
+        Standard: Distribution<V>,
+    {
+        let world = TestWorld::default();
```
akoshelev, line 872: I just realized that we probably want to use the same seed for rng and test world (it is supported via `TestWorldConfig` struct). That way we can make it reproducible if it ever fails

TeONy46: Support

andyleiserson:
The newest commit is to fix the failure in https://github.com/andyleiserson/ipa/actions/runs/11061292252/job/30733595263.

ipa-core/src/protocol/context/dzkp_validator.rs
andyleiserson: I made an issue for this. It's definitely worth doing in general, but it doesn't seem all that important for this particular test case, where the input values should be unrelated to the behavior of the test.

andyleiserson:
Draft runs:

100k: https://draft-mpc.vercel.app/query/view/agile-book2024-09-27T1606, successful
2M: https://draft-mpc.vercel.app/query/view/aided-icon2024-09-27T1612, stalled

I suspect the issue is that, without the changes to make batch sizes power of two / align read size with batch sizes, we can't remove the change that uses active work as the batch size for attribution.

andyleiserson:
Draft result at d2512f1: https://draft-mpc.vercel.app/query/view/drunk-widow2024-09-27T1647

It failed due to #1308. But when I added the workaround (that was 2333492), CI failed due to excessive memory allocation.

akoshelev:
> CI failed due to excessive memory allocation.
Can you please post a link? It's first time we are failing on OOM and I want to quickly check whether we're missing something

andyleiserson:
Here's the failure with large batch size for breakdown reveal aggregation: https://github.com/private-attribution/ipa/actions/runs/11075152193/job/30775478393

PR #1317, opened 2024-09-26
Print send config in tracing

akoshelev:
It was proven useful in several investigations, so we should just enable it


akoshelev:
test failures are unrelated and being investigated currently

PR #1323, opened 2024-09-27
Proof batching for breakdown reveal aggregation

andyleiserson:


andyleiserson:
Draft run, 2M: https://draft-mpc.vercel.app/query/view/upset-poll2024-09-27T1918

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
```diff
@@ -541,7 +540,9 @@ where
             protocol: &Step::Aggregate,
             validate: &Step::AggregateValidate,
         },
-        aggregate_values_proof_chunk(B, usize::try_from(TV::BITS).unwrap()),
+        // TODO: add batching for breakdown reveal aggregation
```
akoshelev, line 543: Lets file an issue for that!

ipa-core/src/protocol/context/batcher.rs
```diff
@@ -152,10 +158,16 @@ impl<'a, B> Batcher<'a, B> {
         let total_count = min(self.records_per_batch, remaining_records);
         let record_offset_in_batch = usize::from(record_id) - first_record_in_batch;
         let batch = self.get_batch_by_offset(batch_offset);
-        assert!(
-            !batch.pending_records[record_offset_in_batch],
-            "validate_record called twice for record {record_id}",
-        );
+        if batch.pending_records.len() <= record_offset_in_batch {
+            batch
```
akoshelev, line 162: I think it is worth writing a test for this - it is important that we can resize this vec if we get over 50M (we definitely do in PRF)

ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs
andyleiserson: #1324. Thanks for filing.

ipa-core/src/protocol/context/batcher.rs
```diff
@@ -111,13 +116,14 @@ impl<'a, B> Batcher<'a, B> {
     fn get_batch_by_offset(&mut self, batch_offset: usize) -> &mut BatchState<B> {
         if self.batches.len() <= batch_offset {
             self.batches.reserve(batch_offset - self.batches.len() + 1);
+            let pending_records_capacity = self.records_per_batch.min(TARGET_PROOF_SIZE);
```
andyleiserson, line 119: Note that `TARGET_PROOF_SIZE` is not really valid here -- what we really want is `TARGET_PROOF_SIZE / steps_in_this_protocol`. But it doesn't seem worth going down that road, because what we really really want is for `records_per_batch` to have a useful value.

ipa-core/src/protocol/context/batcher.rs
andyleiserson: After thinking about it, maybe we don't want to take this change at all? I don't think it's much better than `1 << 30`, even with a test. What we really want is for the protocols to specify an appropriate batch size.

akoshelev: Maybe. Protocols should be able to correctly estimate the number of records, compared to number of multiplications. It could be also true that we are ok giving them some slack.

We could start without resizing it and bring it back if needed

andyleiserson:
The first commit in this PR is the same as before, but I've also now included the changes to do proofs in batches for breakdown reveal aggregation. I think that I could split the commits into two different PRs if needed, but it's not too bad.

ipa-core/src/protocol/context/batcher.rs
andyleiserson: I added a test. I added a smaller value of `TARGET_PROOF_SIZE` for `cfg(test)`. This test case would be fine with the default, but I think it's valuable to cover the multi-batch case for batched aggregation and for other protocols.

andyleiserson:
Rough accounting of added steps:
* Aggregation
  * Depth 4 for outer iteration (only need 2 for production, but 4 for tests with small TARGET_PROOF_SIZE)
  * Depth 24 for inner iteration (aggregation chunk size must be <= 16M
  * 101 steps: 32 each for add, sat_add/add, sat_add/select
  * (4 - 1) × 24 × 101 = 7,272
* Validation: 600 × 7 = 4,200

andyleiserson:
* [ ] `vec_chunks` is inefficient and ought to be improved
* [ ] fix issue in draft

ipa-core/src/utils/vec_chunks.rs
```diff
@@ -0,0 +1,58 @@
+use std::cmp::min;
+
+pub struct VecChunks<T: Clone> {
+    vec: Vec<T>,
+    pos: usize,
+    chunk_size: usize,
+}
+
+impl<T: Clone> Iterator for VecChunks<T> {
+    type Item = Vec<T>;
```
akoshelev, line 10: I checked the usage and it seems that you could yield `&[u8]`. Maybe there are plans to use owned chunks later?

ipa-core/src/utils/vec_chunks.rs
```diff
@@ -0,0 +1,58 @@
...
+    assert!(chunk_size != 0);
+    VecChunks {
+        vec,
+        pos: 0,
+        chunk_size,
+    }
+}
+
+#[cfg(all(test, unit_test))]
+mod tests {
```
akoshelev, line 32: probably also want to test chunk_size = 0 to cause panic

ipa-core/src/protocol/context/dzkp_malicious.rs
```diff
@@ -38,7 +38,7 @@ impl<'a, B: ShardBinding> DZKPUpgraded<'a, B> {
         base_ctx: MaliciousContext<'a, B>,
     ) -> Self {
         let records_per_batch = validator_inner.batcher.lock().unwrap().records_per_batch();
-        let active_work = if records_per_batch == 1 {
+        let active_work = if records_per_batch == 1 || records_per_batch == usize::MAX {
```
akoshelev, line 41: we should explain `usize::MAX` usage too here

akoshelev (approved):
I am going to merge this because it unblocks large inputs. We can address feedback in a follow up

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -34,6 +34,22 @@ const BIT_ARRAY_LEN: usize = 256;
...
+//
+// To enable computing a read size for `OrdereringSender` that achieves good network
+// utilization, the number of records in a proof must be a power of two. Protocols
+// typically compute the size of a proof batch by dividing TARGET_PROOF_SIZE by
+// an approximate number of multiplies per record, and then rounding up to a power
+// of two. Thus, it is not necessary for TARGET_PROOF_SIZE to be a power of two.
+//
+// A smaller value is used for tests, to enable covering some corner cases with a
+// reasonable runtime. Some of these tests use TARGET_PROOF_SIZE directly, so for tests
+// it does need to be a power of two.
```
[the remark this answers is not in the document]
andyleiserson: I don't understand -- I think the comment is explaining that tests user a power of two, protocols don't.

ipa-core/src/protocol/basics/reveal.rs
```diff
@@ -1,3 +1,27 @@
+// Several of the reveal impls use distinct type parameters for the value being revealed
```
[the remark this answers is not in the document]
andyleiserson: It's a note that applies to several of the implementations here, but it doesn't seem sufficient to document the module. That said, if you prefer to make it a module doc comment even though it's incomplete, I'm fine doing that.

ipa-core/src/utils/vec_chunks.rs
andyleiserson: I started from "aggregate_values takes a stream of owned values" and ended up with this, but the reasoning wasn't sound, getting `&[T]` from `slice::chunks` is sufficient.

andyleiserson (commented):
Opened #1347 with the follow-up.

ipa-core/src/protocol/context/dzkp_validator.rs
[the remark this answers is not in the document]
akoshelev: yea, you're right. I read that again and it does say tests need it to be a power of two

PR #1325, opened 2024-09-30
Pinning shuffle futures that are too large for Shuttle

cberkhoff:
This code was written to address a failure observed in `protocol::ipa_prf::aggregation::breakdown_reveal::tests::semi_honest_happy_path` tests. After some investigating I realized the issue only manifested when using the Shuttle crate (not the `multithreading` feature). Shuttle sets a [limit on the stack of 32kb](https://github.com/awslabs/shuttle/blob/main/src/lib.rs#L248) instead of the usual 2MBs.

Instead of just increasing that threshold, I used a few tools (Like the crate [TopTypeSizes](https://docs.rs/top-type-sizes/latest/top_type_sizes/) and [size_of_val](https://doc.rust-lang.org/std/mem/fn.size_of_val.html)) to analize what types could be causing the stack to grow too large, focusing on Futures which tend to be the culprits in these cases. The root stems from `shuffle_protocol` and to address the problem I moved the future to a `Box`. Moreover, we expect to move to Sharded Shuffle which follows a different code path.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -7,3 +7,5 @@ disallowed-methods = [
   { path = "std::mem::ManuallyDrop::new", reason = "Not running the destructors on futures created inside seq_join module will cause UB in IPA. Make sure you don't leak any of those." },
   { path = "std::vec::Vec::leak", reason = "Not running the destructors on futures created inside seq_join module will cause UB in IPA. Make sure you don't leak any of those." },
 ]
+
+future-size-threshold = 8192
--- a/ipa-core/src/protocol/context/dzkp_validator.rs
+++ b/ipa-core/src/protocol/context/dzkp_validator.rs
@@ -972,7 +972,7 @@ mod tests {
         test_select_malicious::<BA20>().await;
         test_select_malicious::<BA32>().await;
         test_select_malicious::<BA64>().await;
-        test_select_malicious::<BA256>().await;
+        Box::pin(test_select_malicious::<BA256>()).await;
     }
 
     #[tokio::test]
--- a/ipa-core/src/protocol/ipa_prf/shuffle/base.rs
+++ b/ipa-core/src/protocol/ipa_prf/shuffle/base.rs
@@ -62,9 +62,9 @@ where
     let zs = generate_random_tables_with_peers(shares_len, &ctx_z);
 
     match ctx.role() {
-        Role::H1 => run_h1(&ctx, shares_len, shares, zs).await,
-        Role::H2 => run_h2(&ctx, shares_len, shares, zs).await,
-        Role::H3 => run_h3(&ctx, shares_len, zs).await,
+        Role::H1 => Box::pin(run_h1(&ctx, shares_len, shares, zs)).await,
+        Role::H2 => Box::pin(run_h2(&ctx, shares_len, shares, zs)).await,
+        Role::H3 => Box::pin(run_h3(&ctx, shares_len, zs)).await,
     }
 }
 


akoshelev (commented):
This makes sense and often happens unfortunately. My hope was that [`large_futures`](https://rust-lang.github.io/rust-clippy/master/index.html#/large_futures) lint should flag those and I don't understand why it doesn't. 

Boxing those futures is not harmful for performance because shuffle is a top level protocol. 

I am worried about not having a test here - is it possible to use `size_of_val` and make sure it stays well below a few kilos for the top level future?

cberkhoff:
> This makes sense and often happens unfortunately. My hope was that [`large_futures`](https://rust-lang.github.io/rust-clippy/master/index.html#/large_futures) lint should flag those and I don't understand why it doesn't.
> 
> Boxing those futures is not harmful for performance because shuffle is a top level protocol.
> 
> I am worried about not having a test here - is it possible to use `size_of_val` and make sure it stays well below a few kilos for the top level future?

`large_futures` does capture large futures. The issue is that this future isn't considered large by the default setting: 16kb.

If I bring the limit to 8kb I can see these futures popping up. They are about 9kb.

I will set the threshold to 8kb so that we can get a proper warning by the linter and not have to do this investigation again.

(I also had to pin another future while adding that threshold)

andyleiserson:
Good job tracking this down. 32 kB seems quite small, I'm surprised we haven't had more problems. I would be in favor of either increasing the shuttle stack to something like 128 kB, or reducing the `large_futures` limit. I expect the latter will require a bit more fixing.

Not that it matters all that much (given we want to move to the sharded shuffle), but I can't spot what in the shuffle is using so much stack.

akoshelev:
> reducing the large_futures limit. I expect the latter will require a bit more fixing.
I am in favor of that approach as well. There can be actually perf issues copying 4kb+ futures through the stack, so I would try to get down to 4kb if that's possible.

> I can't spot what in the shuffle is using so much stack.
yea that is what I realized as well. You can't really just inspect input and output, you need to look at the state captured by any future and it is very hard to trace across function calls. From that perspective, lowering `large_futures` matters even more I think.

akoshelev:
> I will set the threshold to 8kb so that we can get a proper warning by the linter and not have to do this investigation again.

> (I also had to pin another future while adding that threshold)

thanks for doing this! I love this approach

.clippy.toml
```diff
@@ -7,3 +7,5 @@ disallowed-methods = [
   { path = "std::mem::ManuallyDrop::new", reason = "Not running the destructors on futures created inside seq_join module will cause UB in IPA. Make sure you don't leak any of those." },
   { path = "std::vec::Vec::leak", reason = "Not running the destructors on futures created inside seq_join module will cause UB in IPA. Make sure you don't leak any of those." },
 ]
+
+future-size-threshold = 8192
```
akoshelev, line 11: nit: newline

akoshelev (approved):
If you plan to follow up on this, it would be nice to add a comment explaining why we went down to 8k in `clippy.toml`.

andyleiserson:
I think that the reason the shuffle futures are so large is the PRSS. Each PRSS generator contains an `Aes256` instance. The software AES implementation expands keys to 960 B (the natural size of an expanded AES-256 key is 240 B). Depending on the helper, the shuffle implementation has ~5-8 PRSS instances.

If we cared to improve this, we could restructure the shuffle implementation to have fewer PRSS instances outstanding. (It looks like the sharded shuffle naturally ended up structured in such a way.) Or, we could compile with software AES disabled entirely, since we don't expect to run anywhere there isn't hardware AES support.

PR #1327, opened 2024-10-02
Fix large-future clippy warning

akoshelev:
I don't know why it wasn't showing before as #1325 caused it. To my surprise CI was green on it

Here is an example of this failure: https://github.com/private-attribution/ipa/actions/runs/11135685423/job/30946164436?pr=1315


akoshelev:
#1329  takes better care of these

PR #1329, opened 2024-10-02
Reduce size of some futures

andyleiserson:
My understanding is that the future for an `async fn` needs space for both a copy of the arguments, and any sub-futures. Since `join3v` calls `join3`, rewriting these without `async fn` reduces the future size by 2/3.

A related blog post: https://swatinem.de/blog/future-size/


akoshelev:
yea now I remember I read this some time ago. I like this more than boxing futures - clippy should yell at us if we revert this, so we're good here

akoshelev (approved):

martinthomson (approved):
LGTM.

How would we measure the effect of this?

andyleiserson:
> How would we measure the effect of this?

See #1325 for more context. There is a clippy warning if the futures get too big. The immediate tangible impacts in this case were: (1) the `Box::pin` calls that this change removes, (2) stack overflows in shuttle tests, because shuttle defaults to a 32 kB stack. At some point large futures could also be a performance issue.

PR #1332, opened 2024-10-02
Fix send buffer misalignment [No 2]

akoshelev:
## Overview

This is the second attempt to mitigate send buffer misalignment. Previous one (https://github.com/private-attribution/ipa/pull/1307) didn't handle all the edge cases and was abandoned in favour of this PR.

What I believe makes this change work is the new requirement for active work to be a power of two. With this constraint, it is much easier to align the read size with it. Given that `total_capacity = active * record_size`, we can represent `read_size` as a multiple of `record_size` too:
`read_size = X * record_size`. If X is a power of two and active_work is a power of two, then they will always be aligned with each other.

For example, if active work is 16, read size is 10 bytes and record size is 3 bytes, then:

```
total_capacity = 16*3
read_size = X*3 (close to 10)
X = 2 (power of two that satisfies the requirement)
```

when picking up the read size, we are rounding down to avoid buffer overflows. In the example above, setting X=3 would make it closer to the desired read size, but it is greater than 10, so we pick 2 instead.

## What is included in this change?

* Send buffer now enforces the active work to be a power of two
* Subsequently, Gateway interface now puts the same restriction on `GatewayConfig`
* Fix for unit tests that use ZKP - batch sizes now need to be a power of two as well (after #1316 active work and batch size must be aligned now)

### DIFF
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -318,6 +318,7 @@ mod tests {
         stream::StreamExt,
     };
     use proptest::proptest;
+    use tokio::sync::Barrier;
 
     use crate::{
         ff::{
@@ -751,7 +752,11 @@ mod tests {
         ) where
             M: MpcMessage + Clone + PartialEq,
         {
-            let send_notify = Arc::new(tokio::sync::Notify::new());
+            let last_batch_size = total_records % active_work;
+            let last_batch = total_records / active_work;
+
+            let barrier = Arc::new(Barrier::new(active_work));
+            let last_batch_barrier = Arc::new(Barrier::new(last_batch_size));
 
             // perform "multiplication-like" operation (send + subsequent receive)
             // and "validate": block the future until we have at least `active_work`
@@ -762,7 +767,8 @@ mod tests {
                     |(record_id, msg)| {
                         let send_channel = &send_channel;
                         let recv_channel = &recv_channel;
-                        let send_notify = Arc::clone(&send_notify);
+                        let barrier = Arc::clone(&barrier);
+                        let last_batch_barrier = Arc::clone(&last_batch_barrier);
                         async move {
                             send_channel
                                 .send(record_id.into(), msg.clone())
@@ -771,13 +777,12 @@ mod tests {
                             let r = recv_channel.receive(record_id.into()).await.unwrap();
                             // this simulates validate_record API by forcing futures to wait
                             // until the entire batch is validated by the last future in that batch
-                            if record_id % active_work == active_work - 1
-                                || record_id == total_records - 1
-                            {
-                                send_notify.notify_waiters();
+                            if record_id >= last_batch * active_work {
+                                last_batch_barrier.wait().await;
                             } else {
-                                send_notify.notified().await;
+                                barrier.wait().await;
                             }
+
                             assert_eq!(msg, r);
                         }
                     },
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -30,6 +30,7 @@ use crate::{
     protocol::QueryId,
     sharding::ShardIndex,
     sync::{Arc, Mutex},
+    utils::NonZeroU32PowerOfTwo,
 };
 
 /// Alias for the currently configured transport.
@@ -73,8 +74,7 @@ pub struct State {
 pub struct GatewayConfig {
     /// The number of items that can be active at the one time.
     /// This is used to determine the size of sending and receiving buffers.
-    /// Any value that is not a power of two will be rejected
-    pub active: NonZeroUsize,
+    pub active: NonZeroU32PowerOfTwo,
 
     /// Number of bytes packed and sent together in one batch down to the network layer. This
     /// shouldn't be too small to keep the network throughput, but setting it large enough may
@@ -155,7 +155,7 @@ impl Gateway {
         &self,
         channel_id: &HelperChannelId,
         total_records: TotalRecords,
-        active_work: NonZeroUsize,
+        active_work: NonZeroU32PowerOfTwo,
     ) -> send::SendingEnd<Role, M> {
         let transport = &self.transports.mpc;
         let channel = self.inner.mpc_senders.get::<M, _>(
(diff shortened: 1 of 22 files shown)


akoshelev:
I want to run some Draft tests on this

akoshelev:
| Size | Link | Status | 
| ---- | ---- | ------- |
| 1000 | https://draft-mpc.vercel.app/query/view/lax-intro2024-10-02T2329 | Ok |
| 10000 |  | |
| 98721 |  | |
| 100000 | | |

ipa-core/src/helpers/gateway/mod.rs
```diff
@@ -84,6 +85,10 @@ pub struct GatewayConfig {
     /// payload may not be exactly this, but it will be the closest multiple of record size to this
     /// number. For instance, having 14 bytes records and batch size of 4096 will result in
     /// 4088 bytes being sent in a batch.
+    ///
+    /// The actual size for read chunks may be bigger or smaller, depending on the record size
+    /// sent through each channel. Read size will be aligned with [`Self::active_work`] value to
```
akoshelev, line 90: Add a comment explaining what to do if we want to align read size perfectly with the target read size

benjaminsavage: I suggest rewriting this entire comment. It feels out of date.

- line 85 has a typo "side" instead of "size".
- line 86 refers to "batch size" but this comment is above "read_size".
- I think the example is wrong, because 4088 bytes is not a power of two multiple of 14. If I'm not wrong, the best you can do is now 14 * 2^8 < 4096 (it's 3,584).

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,59 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
 
-        let record_size = M::Size::USIZE;
         let total_capacity = gateway_config.active.get() * record_size;
-        Self {
+        // define read size as a multiplier of record size. The multiplier must be
+        // a power of two to align perfectly with total capacity.
+        let read_size_multiplier = {
+            // next_power_of_two returns a value that is greater than or equal to.
+            // That is not what we want here: if read_size / record_size is a power
+            // of two, then subsequent division will get us further away from desired target.
```
benjaminsavage, line 266: I am so confused by this comment. Didn't you just say you *do* want the multiplier to be a power of two? But here you're saying it shouldn't be?

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,59 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+            // For example: if read_size / record_size = 4, then prev_power_of_two = 2.
+            // In such cases, we want to stay where we are, so we add +1 for that.
```
benjaminsavage, line 268: What do you mean "stay where we are"?

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,59 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+            let prev_power_of_two =
+                (gateway_config.read_size.get() / record_size + 1).next_power_of_two() / 2;
+            std::cmp::max(1, prev_power_of_two)
+        };
```
benjaminsavage, line 272: This feels like a really complicated way to get the previous power of two. Isn't it easier to just zero-out all but the most significant bit?

benjaminsavage (approved):
Left a few comments.

ipa-core/src/helpers/gateway/send.rs
akoshelev: yea that would be easier. Thanks for the feedback!

ipa-core/src/helpers/gateway/send.rs
akoshelev: it is all gone now as logic to compute the previous power of two is simplified

ipa-core/src/helpers/gateway/send.rs
akoshelev: yea that was my poor attempt to explain the algorithm, but I think bitshift approach makes it very straightforward  to understand what is going on , so I removed this comment entirely

ipa-core/src/helpers/gateway/mod.rs
akoshelev: You are absolutely right. Just did that

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,54 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+        record_size: usize,
+    ) -> Self {
+        assert!(record_size > 0, "Message size cannot be 0");
 
-        let record_size = M::Size::USIZE;
         let total_capacity = gateway_config.active.get() * record_size;
-        Self {
+        // define read size as a multiplier of record size. The multiplier must be
+        // a power of two to align perfectly with total capacity. We don't want to exceed
+        // the target read size, so we pick a power of two <= read_size.
```
benjaminsavage, line 263: Something still doesn't make sense here. The power of two is not the same thing as read_size. We might pick 2^8 or something, but 8 isn't comparable with "read_size".

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,54 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+        let read_size_multiplier =
+            // this computes the highest power of 2 less than or equal to read_size / record_size.
+            // Note, that if record_size is greater than read_size, we round it to 1
+            1 << (std::cmp::max(1, usize::BITS - (gateway_config.read_size.get() / record_size).leading_zeros()) - 1);
```
benjaminsavage, line 267: I believe this statement is correct, but it's really hard to read. Can you split it up across a few lines for clarity?

```
let target_multiplier = gateway_config.read_size.get() / record_size;
let num_bits = usize::BITS - target_multiplier.leading_zeros();
1 << (std::cmp::max(1, num_bits) - 1);
```

ipa-core/src/helpers/gateway/send.rs
```diff
@@ -248,40 +248,54 @@ impl<I: Debug> Stream for GatewaySendStream<I> {
...
+                std::cmp::min(total_capacity, read_size_multiplier * record_size)
             }
             .try_into()
             .unwrap(),
             total_records,
-        }
+        };
+
+        assert!(this.total_capacity.get() >= record_size * gateway_config.active.get());
+        assert_eq!(0, this.total_capacity.get() % this.read_size.get());
```
benjaminsavage, line 283: Little comments here (with link to issues) would be helpful for readers to understand the necessity of these conditions.

benjaminsavage (approved):

PR #1334, opened 2024-10-03
Skip the unverified multiplies check if we're already unwinding

andyleiserson:
The extra panic is noisy and potentially confusing.


akoshelev (approved):
Thanks for making this change! I also noticed that the stack is wild if we panic somewhere. This is going to make our life much better in debugging

PR #1335, opened 2024-10-03
Support sharded malicious protocols inside the test infrastructure

akoshelev:
This is the final PR in the series (#1333, #1315) that enables writing sharded malicious protocols.

The support is added for `Runner::malicious` function, and it can be used to execute malicious circuits. Other methods like `upgraded_semi_honest` and `upgraded_malicious` haven't been touched yet - I am not convinced that these are necessary and we can update them as we go.

The core goal here is to unblock efforts to implement Hybrid protocol

### DIFF
--- a/ipa-core/src/test_fixture/world.rs
+++ b/ipa-core/src/test_fixture/world.rs
@@ -851,8 +851,7 @@ impl<const SEED: u64> Distribute for Random<SEED> {
     }
 }
 
-// #[cfg(all(test, unit_test))]
-#[cfg(test)]
+#[cfg(all(test, unit_test))]
 mod tests {
     use std::{
         collections::{HashMap, HashSet},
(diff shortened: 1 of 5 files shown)


ipa-core/src/test_fixture/world.rs
```diff
@@ -961,4 +1004,65 @@ mod tests {
...
+            let r = world
+                .malicious(input.clone().into_iter(), |ctx, input| async move {
+                    assert_eq!(1, input.iter().len());
+                    let validator = ctx.set_total_records(1).validator();
+                    let ctx = validator.context();
+                    let (a_upgraded, b_upgraded) = (input[0].clone(), input[0].clone())
+                        .upgrade(ctx.clone(), RecordId::FIRST)
+                        .await
+                        .unwrap();
+                    let _ = a_upgraded
```
benjaminsavage, line 1052: Wait a second, why aren't you using the result of the multiplication? Surely you meant to return this, not `input`, because then it's trivially true that input == input.

benjaminsavage (approved):
Looks good aside from the one comment on the test

ipa-core/src/test_fixture/world.rs
akoshelev: probably not in this case because these shares are upgraded and you can't reconstruct from them. It is possible to do downgrade and then return, but the purpose of this test is to check that `validate_record` and `upgrade` API are available inside sharded infrastructure, so the output does not really matter here.

PR #1337, opened 2024-10-07
Follow up on feedback from #1332

akoshelev:


benjaminsavage (approved):
Thanks for the follow up @akoshelev!

PR #1338, opened 2024-10-08
Set up separate runtimes for HTTP and MPC queries

akoshelev:
This attempts to mitigate panics we're seeing in PRF where computing proofs + PRF values stalls all available CPUs for long time and that prevents HTTP heartbeats and causes receive timeouts on all helpers. We create two dedicated runtimes to avoid scheduling conflicts.

This shouldn't impact our existing backpressure mechanisms as application receivers are still responsible for not polling the H2 streams if they are full

### DIFF
--- a/ipa-core/src/bin/helper.rs
+++ b/ipa-core/src/bin/helper.rs
@@ -158,9 +158,12 @@ async fn server(args: ServerArgs) -> Result<(), BoxError> {
     let network_config_path = args.network.as_deref().unwrap();
     let network_config = NetworkConfig::from_toml_str(&fs::read_to_string(network_config_path)?)?
         .override_scheme(&scheme);
-    let clients = MpcHelperClient::from_conf(&network_config, &identity);
-
     let http_runtime = new_http_runtime();
+    let clients = MpcHelperClient::from_conf(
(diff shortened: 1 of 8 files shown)


akoshelev:
I tried to investigate why codecov is not seeing coverage inside `helper.rs` binary. I spent some time here but I don't know the root cause. We definitely execute that binary code - I confirmed that, but for some reason it shows up in the report only partially.

It could be that spawning a separate process breaks it as coverage tool cannot follow its execution path, but `crypto_util` binary shows coverage correctly. So I don't know what is going on here

cberkhoff (approved):
Looks good to me.

Having a single place to create runtimes seems nice. Maybe we need to do something similar for http clients?

https://github.com/private-attribution/ipa/blob/main/ipa-core/src/net/client/mod.rs#L260

akoshelev:
> Having a single place to create runtimes seems nice. Maybe we need to do something similar for http clients?
Yep, that is a good idea. For consistency it is better to use HTTP runtime throughout the stack. Unfortunately I wasn't able to get rid of runtime reference from inside `HttpTransport` - we still need to poll the send future. But it should help with scheduling issues anyway

PR #1342, opened 2024-10-09
Validate uniqueness

eriktaubeneck:
This validates uniqueness across all the encrypted inputs (currently `EncryptedOprfReports`) and then decrypts and converts them into `HybridReports`. 

This currently is not sharded, that will be the next step. I plan to break that up into 1+ more PRs, as we need to figure out how to reshard as we receive reports from the RC (and not keep it all in memory for the leader), which will likely necessitate a reshard function that can operate on streams instead of iterators.


ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,285 @@
...
+        let Self {
+            config,
+            key_registry,
+            phantom_data: _,
+        } = self;
+        tracing::info!("New hybrid query: {config:?}");
+        let _ctx = ctx.narrow(&Hybrid);
+        let sz = usize::from(query_size);
+
+        let seen_encrypted_reports: Arc<Mutex<HashSet<Vec<u8>>>> =
```
akoshelev, line 56: I suggest to build a separate struct to check for duplicates. It may provide better encapsulation and it could work with reports, rather than just bytes. There is also a benefit of testing it more easily, in isolation with `execute` method

```rust
struct Duplicates(HashSet<...>);

impl Duplicates {

   fn check_duplicate(&mut self, report: EncryptedReport<>) -> Result<(), Duplicate>> {}

   // this should tell you the index of duplicate, if it is there
   fn check_duplicates(&mut self, reports: &[EncryptedReport]) -> Result<(), DuplicateReportIndex> {}
}
```

Allowing it to operate on a vec, would help `execute` function too, as it no longer needs to dance around `map_ok`.

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,285 @@
...
+                .map_ok(move |enc_reports| {
+                    iter(enc_reports.into_iter().map({
+                        let seen_encrypted_reports =
+                            Arc::<Mutex<HashSet<Vec<u8>>>>::clone(&seen_encrypted_reports);
+
+                        let key_registry = Arc::<R>::clone(&key_registry);
+                        move |enc_report| {
+                            let ciphertext = enc_report.mk_ciphertext().to_owned();
+                            let mut guard = seen_encrypted_reports.lock().unwrap();
+                            assert!(guard.insert(ciphertext), "Duplicate report found.");
```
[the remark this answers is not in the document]
akoshelev: and also, we shouldn't panic on bad inputs, we should return an error back to the client in this case.

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,285 @@
...
+                .map_err(Into::<Error>::into)
+                .map_ok(move |enc_reports| {
+                    iter(enc_reports.into_iter().map({
+                        let seen_encrypted_reports =
+                            Arc::<Mutex<HashSet<Vec<u8>>>>::clone(&seen_encrypted_reports);
+
+                        let key_registry = Arc::<R>::clone(&key_registry);
+                        move |enc_report| {
+                            let ciphertext = enc_report.mk_ciphertext().to_owned();
+                            let mut guard = seen_encrypted_reports.lock().unwrap();
```
akoshelev, line 72: I think `Arc<Mutex` is unnecessary here. Stream is polled by one thread only, so we don't need synchronization here. Rust probably steered you towards that path because of double closure usage here, but it is not really needed. Something like this should work (and if you create a struct, this would be even better)

```rust
        let mut seen_encrypted_reports: HashSet<Vec<u8>> = HashSet::with_capacity(sz);
        ...
        LengthDelimitedStream::<EncryptedOprfReport<BA8, BA3, BA20, _>, _>::new(input_stream)
                .map_err(Into::<Error>::into)
                .map_ok(|enc_reports| {
                    // check uniqueness
                    let before = seen_encrypted_reports.len();
                    // extend, or iterate over `enc_reports` - both should work
                    seen_encrypted_reports.extend(enc_reports.iter().map(|report| report.mk_ciphertext().to_vec()));
                    assert_eq!(before + enc_reports.len(), seen_encrypted_reports.len());
                    
                    iter(enc_reports.into_iter().map(|enc_report|
                        HybridReport::<BA8, BA3>::from_encrypted_oprf_report::<R, _, BA20>(
                            &enc_report,
                            &key_registry,
                        ).map_err(Into::<Error>::into)))
                })
```

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,285 @@
...
+        query::runner::HybridQuery,
+        report::{OprfReport, DEFAULT_KEY_ID},
+        secret_sharing::IntoShares,
+        test_fixture::{ipa::TestRawDataRecord, join3v, Reconstruct, TestWorld},
+    };
+
+    #[tokio::test]
+    // placeholder until the protocol is complete. can be updated to make sure we
+    // get to the unimplemented() call
+    #[should_panic(
```
akoshelev, line 119: 👍

ipa-core/src/protocol/hybrid/step.rs
```diff
@@ -0,0 +1,6 @@
+use ipa_step_derive::CompactStep;
+
+#[derive(CompactStep)]
+pub(crate) enum HybridStep {
+    VerifyCiphertextUniqueness,
```
akoshelev, line 5: doesn't seem to be used anywhere?

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,285 @@
...
+        } = self;
+        tracing::info!("New hybrid query: {config:?}");
+        let _ctx = ctx.narrow(&Hybrid);
+        let sz = usize::from(query_size);
+
+        let seen_encrypted_reports: Arc<Mutex<HashSet<Vec<u8>>>> =
+            Arc::new(Mutex::new(HashSet::with_capacity(sz)));
+
+        let _input = if config.plaintext_match_keys {
+            unimplemented!()
```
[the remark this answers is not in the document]
eriktaubeneck: Sure. My thought was that we plan to implement this eventually, but we can do that.

ipa-core/src/protocol/hybrid/step.rs
eriktaubeneck: you're right. originally I was going to pass it into a separate function and call `ctx.narrow`, but I dropped that when I changed the approach. is it useful to set the step here, or should I just drop that?

akoshelev: If you plan to use it soon, we can keep it, otherwise it is probably better to remove it

ipa-core/src/query/runner/hybrid.rs
eriktaubeneck: is this doing some sort of automatic chunking? (just trying to reason about what's different here)

Rust steered me towards `Arc<Mutux` because it's async, I believe. but let me see if it complains about this (and also the struct approach.)

ipa-core/src/query/runner/hybrid.rs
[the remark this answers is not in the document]
eriktaubeneck: are you sure this includes the tag? my understanding of [this byte assignment](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/report/ipa.rs#L258-L259) was that the `encap_key_1` was the tag. If that's not the case, what bytes should I get?

ipa-core/src/query/runner/hybrid.rs
eriktaubeneck: this makes sense, though it doesn't seem that HashSets are ordered, so there isn't an index. I think it's sufficient to just return an Error.

akoshelev: yea, I meant the index in the `reports` slice  - you want to know which exact report is a duplicate

ipa-core/src/query/runner/hybrid.rs
akoshelev: it is all about returning an iterator that grabs a closure I think. async can and should work with references only here - there is only one thread polling this stream.

I tested the code above and it worked for me, although I am not entirely happy using `extend` here

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,346 @@
...
+                "Hybrid queries do not currently support plaintext match keys".to_string(),
+            ));
+        }
+
+        let _input = LengthDelimitedStream::<EncryptedHybridReport, _>::new(input_stream)
+            .map_err(Into::<Error>::into)
+            .map_ok(|enc_reports| {
+                unique_encrypted_hybrid_reports
+                    .check_duplicates(&enc_reports)
+                    .unwrap();
```
eriktaubeneck, line 70: todo: this needs to propagate, not panic

ipa-core/src/report/hybrid.rs
```diff
@@ -90,28 +115,100 @@ where
...
+        UniqueBytesValidator {
+            hash_set: HashSet::<Vec<u8>>::with_capacity(size),
+            incrementer: 0,
+        }
+    }
+
+    fn insert(&mut self, value: Vec<u8>) -> bool {
+        let r = self.hash_set.insert(value);
+        if r {
+            self.incrementer += 1;
```
akoshelev, line 174: I am not sure if this serves the purpose here. If you want to point to the exact location of the duplicate report, this does not work because it does not take into account previous duplicates. Lets say you have [foo, bar, baz, foo, baz]. This correctly points at second `foo`, but baz will have an index 3 instead of 4. 

If you want to detect only first duplicate and stop, then `incrementer` is always `hash_set.len()` I think. 

I would probably just keep a global counter, or make the caller to be responsible to that. In the latter case, you only need to signal back the index inside `check_duplicates` function and point to the item that failed duplicate check. Then caller can find it in its own array and compute the index and/or just point at the ciphertext

ipa-core/src/report/hybrid.rs
```diff
@@ -179,12 +285,70 @@ mod test {
...
+
+        assert!(
+            expected_err.is_err(),
+            "Expected an error, but got Ok instead"
+        );
+
+        expected_err.unwrap_or_else(|err| match err {
+            Error::DuplicateBytes(value) => assert_eq!(value, 3),
+            _ => panic!("Expected DuplicateBytes error, got: {err}"),
+        });
```
akoshelev, line 342: you could use `assert!(matches!(expected_err, Err(Error::DuplicateBytes())` here

akoshelev (approved):
I feel strongly about using the constant for AEAD tag instead of hardcoded  value of 16. You also assume the size of the ciphertext to be 16 bytes which techinically can change if we use more than 8 bytes per match key. So I would compute (ct.len() - TAG) instead

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,350 @@
...
+
+        if config.plaintext_match_keys {
+            return Err(Error::Unsupported(
+                "Hybrid queries do not currently support plaintext match keys".to_string(),
+            ));
+        }
+
+        let _input = LengthDelimitedStream::<EncryptedHybridReport, _>::new(input_stream)
+            .map_err(Into::<Error>::into)
+            .and_then(|enc_reports| {
```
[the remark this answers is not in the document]
eriktaubeneck: it took a lot of attempts to figure out the right way to do this...

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,350 @@
...
+        .await;
+
+        assert_eq!(
+            results.reconstruct()[0..3]
+                .iter()
+                .map(U128Conversions::as_u128)
+                .collect::<Vec<u128>>(),
+            EXPECTED
+        );
+    }
```
eriktaubeneck, line 216: lol why is codedev saying a test isn't covered by tests?

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,350 @@
...
+    // get to the unimplemented() call
+    #[should_panic(
+        expected = "not implemented: query::runnner::HybridQuery.execute is not fully implemented"
+    )]
+    async fn encrypted_hybrid_reports() {
+        const EXPECTED: &[u128] = &[0, 8, 5];
+
+        // TODO: When Encryption/Decryption exists for HybridReports
+        // update these to use that, rather than generating OprfReports
+        let records: Vec<TestRawDataRecord> = vec![
```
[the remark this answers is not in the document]
eriktaubeneck: this one is intended to evolve into the actual test, once we work our way thought everything being implemented (it's pretty close to a copy of the test we had for oprf_ipa)

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -1,37 +1,350 @@
...
+        let mut buffers: [_; 3] = std::array::from_fn(|_| Vec::new());
+
+        let shares: [Vec<OprfReport<BA8, BA3, BA20>>; 3] = records.into_iter().share();
+        for (buf, shares) in zip(&mut buffers, shares) {
+            for share in shares {
+                share
+                    .delimited_encrypt_to(key_id, key_registry.as_ref(), &mut rng, buf)
+                    .unwrap();
+            }
+        }
```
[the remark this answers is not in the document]
eriktaubeneck: i tried, fought with the compiler for a while, and gave up. i can try again...

ipa-core/src/report/hybrid.rs
```diff
@@ -90,28 +115,100 @@ where
...
+}
+
+pub trait UniqueBytes {
+    fn unique_bytes(&self) -> Vec<u8>;
+}
+
+impl UniqueBytes for EncryptedHybridReport {
+    fn unique_bytes(&self) -> Vec<u8> {
+        let start: usize = EncapsulationSize::USIZE;
+        let end: usize = start + 16;
```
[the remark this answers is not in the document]
eriktaubeneck: yes, good call out for IpaAead::SIZE. 

as far as the actual ciphertext length, it's just hardcoded on the implementation of `EncryptedOprfReport`. actually instantiating that struct requires all the crazy trait bounds, so for now, it'll be much nicer to just copy the right details from that implementation, and then update this when we have a proper implementation for `EncryptedHybridReport`. I'll change this a bit, and at TODOs.

ipa-core/src/report/hybrid.rs
eriktaubeneck: ahh yea, good point. what we want is an `check_counter` that increments for every check, regardless of success.

ipa-core/src/report/hybrid.rs
eriktaubeneck: ooo that's much nicer

ipa-core/src/report/hybrid.rs
```diff
@@ -90,28 +115,100 @@ where
...
+pub struct UniqueBytesValidator {
+    hash_set: HashSet<Vec<u8>>,
+    incrementer: usize,
+}
+
+impl UniqueBytesValidator {
+    #[must_use]
+    pub fn new(size: usize) -> Self {
+        UniqueBytesValidator {
+            hash_set: HashSet::<Vec<u8>>::with_capacity(size),
```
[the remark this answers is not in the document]
akoshelev: compiler can derive the type here because it knows the type of variable

ipa-core/src/query/runner/hybrid.rs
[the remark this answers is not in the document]
eriktaubeneck: I added a comment here. the test actually has the assertion for when the implementation is complete, so I'd like to keep this and work towards actually making this test pass.

ipa-core/src/query/runner/hybrid.rs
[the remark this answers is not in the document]
eriktaubeneck: made it work!

eriktaubeneck:
> I feel strongly about using the constant for AEAD tag instead of hardcoded value of 16. You also assume the size of the ciphertext to be 16 bytes which techinically can change if we use more than 8 bytes per match key. So I would compute (ct.len() - TAG) instead

updated this to use the TAG. Small nit: ct.len = TagSize + Replicated<BA64>::Size, so computing that subtraction is redundant. I did add a method to get the ciphertext, and then from that take the TagSize off of it.

akoshelev:
@eriktaubeneck don't forget to squash commits before merging - this PR had a lot of iterations and some of them don't need to be in main

PR #1343, opened 2024-10-09
Make reshard work with streams too

akoshelev:
Internally, reshard used streams already, so it is only a matter of changing the API and connecting things together

### DIFF
--- a/ipa-core/src/protocol/context/mod.rs
+++ b/ipa-core/src/protocol/context/mod.rs
@@ -35,8 +35,8 @@ pub(crate) use malicious::TEST_DZKP_STEPS;
 use crate::{
     error::Error,
     helpers::{
-        ChannelId, Direction, Gateway, Message, MpcMessage, MpcReceivingEnd, Role, SendingEnd,
-        ShardReceivingEnd, TotalRecords,
+        stream::ExactSizeStream, ChannelId, Direction, Gateway, Message, MpcMessage,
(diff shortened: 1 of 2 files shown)


akoshelev:
@eriktaubeneck This may be what you want. This PR is not complete (I just quickly draw an API change), so we can't merge it, but the only thing that is missing is test coverage.

eriktaubeneck:
test added @akoshelev

ipa-core/src/protocol/context/mod.rs
```diff
@@ -822,15 +867,42 @@ mod tests {
 
     /// Ensure global record order across shards is consistent.
```
akoshelev, line 868: the comment needs to be changed, otherwise lgtm

akoshelev: heh, I can't approve my own PRs

eriktaubeneck (approved):
great work, thanks for unblocking!

ipa-core/src/protocol/context/mod.rs
akoshelev: I thought the purpose of this test is to ensure we can use stream API, ensuring global record order in two tests seems redundant. But that's a nit, lets merge it and follow up if needed

eriktaubeneck: ahh I see you're point. I can open a small PR to clarify that. we want global ordering, and we want that to work with both APIs

PR #1345, opened 2024-10-10
Make disable-metrics feature turn off global recorder

akoshelev:
It turns out we've been running with metrics overhead the whole time. I tested a run locally with `disable-metrics` feature and to my surprise I still saw the telemetry emitted from helper binary.

Upon investigating, I came across our `Verbosity` struct that was installing the collector unconditionally. I changed that and tested that metrics are no longer emitted

### DIFF
--- a/ipa-core/src/cli/verbosity.rs
+++ b/ipa-core/src/cli/verbosity.rs
@@ -32,23 +32,28 @@ impl Verbosity {
     #[must_use]
     pub fn setup_logging(&self) -> LoggingHandle {
         let filter_layer = self.log_filter();
+        info!("Logging setup at level {}", filter_layer);
+
         let fmt_layer = fmt::layer()
             .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
             .with_ansi(std::io::stderr().is_terminal())
             .with_writer(stderr);
 
-        tracing_subscriber::registry()
-            .with(self.log_filter())
-            .with(fmt_layer)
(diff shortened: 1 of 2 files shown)


akoshelev:
I can see the difference now

with metrics
Running IPA for 1000000 records took 633.11591412s
https://draft-mpc.vercel.app/query/view/lyric-days2024-10-10T0425

without metrics
Running IPA for 1000000 records took 558.098657199s
https://draft-mpc.vercel.app/query/view/gutsy-flyer2024-10-10T0414

eriktaubeneck (approved):
ship it!

ipa-core/src/cli/verbosity.rs
```diff
@@ -32,24 +32,29 @@ impl Verbosity {
...
-        tracing_subscriber::registry()
-            .with(self.log_filter())
-            .with(fmt_layer)
-            .with(MetricsLayer::new())
-            .init();
+        let registry = tracing_subscriber::registry()
+            .with(filter_layer)
+            .with(fmt_layer);
+
+        if cfg!(feature = "disable-metrics") {
```
cberkhoff, line 46: nit: I feel it would be better to have the `init()` outside of the if-else. Mostly for future features that might modify the struct.

```
if cfg!(not(feature = "disable-metrics")) {
registry.with(MetricsLayer::new())
}
registry.init();
```

cberkhoff (approved):

PR #1347, opened 2024-10-14
Review follow-up from #1323

andyleiserson:


ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -100,7 +99,8 @@ where
 
     while intermediate_results.len() > 1 {
         let mut record_ids = [RecordId::FIRST; AGGREGATE_DEPTH];
-        for chunk in vec_chunks(mem::take(&mut intermediate_results), agg_proof_chunk) {
+        let mut next_intermediate_results = Vec::new();
```
akoshelev, line 102: is it possible to set the capacity here or it does not matter?

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
andyleiserson: It would be possible to calculate a capacity as the input length divided by agg_proof_chunk, but given that we're pushing one element onto this vector per tens of thousands of input records, it didn't really seem important to do so.

PR #1348, opened 2024-10-14
[Sharding 1] TransportRestriction and Client

cberkhoff:
This is the first of a series of pull requests (PR) to enable sharding on IPA.

The key introduction of this PR is the `TransportRestriction` trait. Please take a read at the included docs for more details. To make more sense of the intended usage of this new trait, I'm modifying `MpcHelperClient` to make use of it, but without altering it's API, yet. I understand the scope of this example is limited, hence I can answer questions about its future use if needed.

Renamed `ClientIdentity::Helper` to `ClientIdentity::Header` since I'm planning to use the same mechanism for Shard to Shard identity resolution.

### DIFF
--- a/ipa-core/src/helpers/transport/mod.rs
+++ b/ipa-core/src/helpers/transport/mod.rs
@@ -44,21 +44,46 @@ pub trait Identity:
     Copy + Clone + Debug + PartialEq + Eq + PartialOrd + Ord + Hash + Send + Sync + 'static
 {
     fn as_str(&self) -> Cow<'static, str>;
+
+    /// Parses a ref to a string representation of this identity
+    ///
+    /// # Errors
+    /// If there where any problems parsing the identity.
+    fn from_str(s: &str) -> Result<Self, crate::error::Error>;
 }
 
 impl Identity for ShardIndex {
     fn as_str(&self) -> Cow<'static, str> {
         Cow::Owned(self.to_string())
     }
+
+    fn from_str(s: &str) -> Result<Self, crate::error::Error> {
+        s.parse::<u32>()
+            .map_err(|_e| {
+                crate::error::Error::InvalidId(format!("The string {s} is an invalid Shard Index"))
+            })
+            .map(ShardIndex::from)
+    }
 }
 impl Identity for HelperIdentity {
     fn as_str(&self) -> Cow<'static, str> {
         Cow::Borrowed(match *self {
-            Self::ONE => "A",
-            Self::TWO => "B",
-            Self::THREE => "C",
+            Self::ONE => Self::ONE_STR,
+            Self::TWO => Self::TWO_STR,
+            Self::THREE => Self::THREE_STR,
             _ => unreachable!(),
         })
+    }
+
+    fn from_str(s: &str) -> Result<Self, crate::error::Error> {
+        match s {
+            Self::ONE_STR => Ok(Self::ONE),
+            Self::TWO_STR => Ok(Self::TWO),
+            Self::THREE_STR => Ok(Self::THREE),
+            _ => Err(crate::error::Error::InvalidId(format!(
+                "The string {s} is an invalid Helper Identity"
+            ))),
+        }
     }
 }
 
@@ -67,6 +92,17 @@ impl Identity for HelperIdentity {
 impl Identity for Role {
     fn as_str(&self) -> Cow<'static, str> {
         Cow::Borrowed(Role::as_static_str(self))
+    }
+
+    fn from_str(s: &str) -> Result<Self, crate::error::Error> {
+        match s {
+            Self::H1_STR => Ok(Self::H1),
+            Self::H2_STR => Ok(Self::H2),
+            Self::H3_STR => Ok(Self::H3),
+            _ => Err(crate::error::Error::InvalidId(format!(
+                "The string {s} is an invalid Role"
+            ))),
+        }
     }
 }
 
@@ -227,5 +263,55 @@ pub trait Transport: Clone + Send + Sync + 'static {
     #[must_use]
     fn clone_ref(&self) -> Self {
         <Self as Clone>::clone(self)
+    }
+}
+
+#[cfg(all(test, unit_test))]
+mod tests {
+    use crate::{
+        helpers::{HelperIdentity, Role, TransportIdentity},
+        sharding::ShardIndex,
+    };
+
+    #[test]
+    fn helper_from_str() {
+        assert_eq!(HelperIdentity::from_str("A").unwrap(), HelperIdentity::ONE);
+        assert_eq!(HelperIdentity::from_str("B").unwrap(), HelperIdentity::TWO);
+        assert_eq!(
+            HelperIdentity::from_str("C").unwrap(),
+            HelperIdentity::THREE
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "The string H1 is an invalid Helper Identity")]
+    fn invalid_helper_from_str() {
+        assert_eq!(HelperIdentity::from_str("H1").unwrap(), HelperIdentity::ONE);
+    }
+
+    #[test]
+    fn shard_from_str() {
+        assert_eq!(ShardIndex::from_str("42").unwrap(), ShardIndex::from(42));
+        assert_eq!(ShardIndex::from_str("9").unwrap(), ShardIndex::from(9));
+        assert_eq!(ShardIndex::from_str("0").unwrap(), ShardIndex::from(0));
+    }
+
+    #[test]
+    #[should_panic(expected = "The string -1 is an invalid Shard Index")]
+    fn invalid_shard_from_str() {
+        assert_eq!(ShardIndex::from_str("-1").unwrap(), ShardIndex::from(0));
+    }
+
+    #[test]
+    fn role_from_str() {
+        assert_eq!(Role::from_str("H1").unwrap(), Role::H1);
+        assert_eq!(Role::from_str("H2").unwrap(), Role::H2);
+        assert_eq!(Role::from_str("H3").unwrap(), Role::H3);
+    }
+
+    #[test]
+    #[should_panic(expected = "The string A is an invalid Role")]
+    fn invalid_role_from_str() {
+        assert_eq!(Role::from_str("A").unwrap(), Role::H1);
     }
 }
--- a/ipa-core/src/net/client/mod.rs
+++ b/ipa-core/src/net/client/mod.rs
@@ -41,11 +41,11 @@ use crate::{
 };
 
 #[derive(Default)]
-pub enum ClientIdentity<R: ConnectionFlavor = Helper> {
+pub enum ClientIdentity<F: ConnectionFlavor = Helper> {
     /// Claim the specified helper identity without any additional authentication.
     ///
     /// This is only supported for HTTP clients.
-    Header(R::Identity),
+    Header(F::Identity),
 
     /// Authenticate with an X.509 certificate or a certificate chain.
     ///
@@ -57,7 +57,7 @@ pub enum ClientIdentity<R: ConnectionFlavor = Helper> {
     None,
 }
 
-impl<R: ConnectionFlavor> ClientIdentity<R> {
+impl<F: ConnectionFlavor> ClientIdentity<F> {
     /// Authenticates clients with an X.509 certificate using the provided certificate and private
     /// key. Certificate must be in PEM format, private key encoding must be [`PKCS8`].
     ///
@@ -82,7 +82,7 @@ impl<R: ConnectionFlavor> ClientIdentity<R> {
     /// to own a private key, and we need to create 3 with the same config, we provide Clone
     /// capabilities via this method to `ClientIdentity`.
     #[must_use]
-    pub fn clone_with_key(&self) -> ClientIdentity<R> {
+    pub fn clone_with_key(&self) -> ClientIdentity<F> {
         match self {
             Self::Certificate((c, pk)) => Self::Certificate((c.clone(), pk.clone_key())),
             Self::Header(h) => Self::Header(*h),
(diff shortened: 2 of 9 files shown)


ipa-core/src/net/server/mod.rs
```diff
@@ -722,6 +730,7 @@ mod e2e_tests {
         let expected = expected_req(addr.to_string());
         let req = http_req(&expected, uri::Scheme::HTTP, addr.to_string());
         let response = client.request(req).await.unwrap();
+        println!("{}", response.status());
```
akoshelev, line 733: nit: you can print the status inside `assert_eq!(a, b, "{a} != {b}")

ipa-core/src/net/server/mod.rs
```diff
@@ -443,6 +444,13 @@ impl<S> SetClientIdentityFromHeader<S> {
     fn new(inner: S) -> Self {
         Self { inner }
     }
+
+    fn parse_client_id(header_value: &HeaderValue) -> Result<ClientIdentity, Error> {
```
akoshelev, line 448:
```suggestion
    fn try_parse_client_id(header_value: &HeaderValue) -> Result<ClientIdentity, Error> {
```

akoshelev: also why it is in this impl - there is nothing that associated this function with the struct it is defined on. `ClientIdentity` is probably the better place for it

ipa-core/src/net/mod.rs
```diff
@@ -24,6 +25,10 @@ pub use transport::{HttpShardTransport, HttpTransport};
 
 pub const APPLICATION_JSON: &str = "application/json";
 pub const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
+pub static HTTP_CLIENT_ID_HEADER: HeaderName =
```
akoshelev, line 28: this name is really confusing, we should clearly indicate the difference between MPC-to-MPC and shard-to-shard authentication headers.

ipa-core/src/net/mod.rs
```diff
@@ -3,6 +3,7 @@ use std::{
     sync::Arc,
 };
 
+use axum::http::HeaderName;
```
akoshelev, line 6:
```suggestion
use hyper::header::HeaderName;
```

ipa-core/src/net/mod.rs
```diff
@@ -24,6 +25,10 @@ pub use transport::{HttpShardTransport, HttpTransport};
 
 pub const APPLICATION_JSON: &str = "application/json";
 pub const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
+pub static HTTP_CLIENT_ID_HEADER: HeaderName =
+    HeaderName::from_static("x-unverified-client-identity");
+pub static HTTP_SHARD_INDEX_HEADER: HeaderName =
```
akoshelev, line 30: this should be internal to this module and its children imo

ipa-core/src/net/client/mod.rs
```diff
@@ -153,39 +155,15 @@ impl Future for ResponseFuture {
 /// TODO: It probably isn't necessary to always use `[MpcHelperClient; 3]`. Instead, a single
 ///       client can be configured to talk to all three helpers.
 #[derive(Debug, Clone)]
-pub struct MpcHelperClient {
+pub struct MpcHelperClient<R: TransportRestriction = Ring> {
```
akoshelev, line 158: Maybe `ConnectionFlavor` or `Connectivity` are better names for it. Also we may need a follow up to remove the default value - these tend to produce cryptic compile errors when generics are omitted

ipa-core/src/net/client/mod.rs
```diff
@@ -680,7 +692,7 @@ pub(crate) mod tests {
             .await
             .unwrap();
 
-        MpcHelperClient::resp_ok(resp).await.unwrap();
+        MpcHelperClient::<Ring>::resp_ok(resp).await.unwrap();
```
akoshelev, line 695: I would move `resp_ok` out of this struct, so we don't need to spell out full name everywhere. I assume resp_ok just generates an 200 HTTP response which shouldn't depend on whether we use MPC or shard client

ipa-core/src/sharding.rs
```diff
@@ -3,9 +3,53 @@ use std::{
...
+/// trait can be used to limit the functions exposed by a struct impl depending on the context that
+/// it's being used. Continuing the previous example, the functions a
+/// [`crate::net::client::MpcHelperClient<R>`] provides are dependent on whether it's communicating
+/// with another Shard or another Helper.
+///
+/// This trait is a safety restriction so that structs or traits only expose an API that's
+/// meaningful for their specific context. When used as a generic bound, it also spreads through
+/// the types making it harder to be misused or combining incompatible types, e.g. Using a
+/// [`ShardIndex`] with a [`Ring`].
+pub trait TransportRestriction: Debug + Send + Sync + Clone + 'static {
```
akoshelev, line 22: this place does not look like a good home for this trait, unless you plan to use it outside of HTTP module?

ipa-core/src/sharding.rs
```diff
@@ -3,9 +3,53 @@ use std::{
...
+}
+
+/// This marker is used to restrict communication inside a single Helper, with other shards.
+#[derive(Debug, Copy, Clone)]
+pub struct Sharding;
+
+/// This marker is used to restrict communication inter Helpers. This communication usually has
+/// more restrictions. 3 Hosts with the same sharding index are conencted in a Ring.
+#[derive(Debug, Copy, Clone)]
+pub struct Ring;
```
akoshelev, line 34: I like how `Client<Helper>` and `Client<Shard>` look and I think ring/sharding leaves some room for interpretation and I'd love to be crystal clear here.

ipa-core/src/net/client/mod.rs
```diff
@@ -153,39 +155,15 @@ impl Future for ResponseFuture {
...
-            Self::new(
-                runtime.clone(),
-                &conf.client,
-                peer_conf.clone(),
-                identity.clone_with_key(),
-            )
-        })
-    }
-
+impl<R: TransportRestriction> MpcHelperClient<R> {
```
akoshelev, line 166: It is no longer an MPC client I think. It is fine to follow up on renaming, but lets leave a trace for this action item somewhere

ipa-core/src/net/client/mod.rs
```diff
@@ -198,7 +176,8 @@ impl MpcHelperClient {
         runtime: IpaRuntime,
         client_config: &ClientConfig,
         peer_config: PeerConfig,
-        identity: ClientIdentity,
+        identity: ClientIdentity<R>,
+        header_name: &'static HeaderName,
```
akoshelev, line 180: This is less readable and harder to understand in terms of allocations (because you clone it down below) than just passing it by value and cloning (HeaderName is just a wrapper around a smart pointer, so cloning is fine)

akoshelev (commented):
Good direction, some polishing of this code is required imo and we may be good to go

ipa-core/src/net/client/mod.rs
cberkhoff: Why do you think it's not an "MPC" client? It's still used to talk to multiple parties in a protocol.

Or is it the "Helper" part that bothers you? I can just omit it and call it `MpcClient`?

ipa-core/src/net/client/mod.rs
cberkhoff: To be sure...

You want me to give ownership (no `&`) in new `MpcHelperClient::<R>::new` and have the callers of this method e.g. `MpcHelperClient<Ring>::from_conf` clone it?

What's the readability issue? You rather see clones happening "higher-up" instead inside `new`? Let me know if you rather talk online about that part

ipa-core/src/sharding.rs
cberkhoff: I'll add more docs

ipa-core/src/net/client/mod.rs
cberkhoff: I spent some time thinking about this name and it's a tricky one.

I don't think `Connectivity` conveys enough information. I can go with `ConnectionFlavor` if you think its better. In [the Sharding article in Wikipedia](https://en.wikipedia.org/wiki/Shard_(database_architecture)) they refer to vertical and horizontal as "Partitions". So maybe `ConnectionPartition` ?

ipa-core/src/net/mod.rs
cberkhoff: I was following the names of the respective structs: `ClientIdentity` and `ShardIndex`.  Do you think we really need to change this?

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -44,22 +44,47 @@ pub trait Identity:
...
+    /// If there where any problems parsing the identity.
+    fn from_str(s: &str) -> Result<Self, crate::error::Error>;
 }
 
 impl Identity for ShardIndex {
     fn as_str(&self) -> Cow<'static, str> {
         Cow::Owned(self.to_string())
     }
+
+    fn from_str(s: &str) -> Result<Self, crate::error::Error> {
```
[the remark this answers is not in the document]
cberkhoff: Yes, I'll do that instead

ipa-core/src/net/client/mod.rs
```diff
@@ -32,18 +33,19 @@ use crate::{
...
     },
-    net::{http_serde, server::HTTP_CLIENT_ID_HEADER, Error, CRYPTO_PROVIDER},
+    net::{http_serde, Error, CRYPTO_PROVIDER, HTTP_CLIENT_ID_HEADER},
     protocol::{Gate, QueryId},
+    sharding::{Ring, TransportRestriction},
 };
 
-#[derive(Default)]
-pub enum ClientIdentity {
+#[derive(Default, Debug)]
```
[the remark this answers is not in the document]
cberkhoff: Ok, I'll remove `Debug`

ipa-core/src/sharding.rs
```diff
@@ -3,9 +3,53 @@ use std::{
...
+}
+
 /// A unique zero-based index of the helper shard.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ShardIndex(u32);
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize)]
+#[serde(from = "u32")]
+pub struct ShardIndex(pub u32);
+
+impl From<ShardIndex> for u32 {
```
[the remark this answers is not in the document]
cberkhoff: You mean the Serde stuff? I'll remove that, not gonna be needed anymore with the String functions

ipa-core/src/net/client/mod.rs
andyleiserson: Possibly you can get the header name from a constant or method on `TransportRestriction` rather than storing it explicitly.

andyleiserson:
The authentication requirements for shard communication are different than for helper-to-helper communication. Between helpers, it is important to prevent e.g. impersonation of H1 by H2. Between shards, that is not important. I would still want a secure connection with some kind of authentication (i.e., don't presume that shard-to-shard communication is over a trusted network). But the different shards are under the control of a single helper operator, so it's not important to prevent shard 2 from claiming to be shard 3.

Where I'm going with this is, we don't need to provision a unique certificate for each shard, if it saves work to not do so.

cberkhoff:
> Where I'm going with this is, we don't need to provision a unique certificate for each shard, if it saves work to not do so.

I'm planning to use the same certificate for both helper-to-helper and shard-to-shard. The certificate identifying the host regardless of the communication. Let me know what you think.

ipa-core/src/net/client/mod.rs
cberkhoff: Originally I didn't wanted to tie HTTP (net module) specific things to `TransportRestriction`. If we make it a net only parameter, I think that makes sense.

ipa-core/src/helpers/transport/mod.rs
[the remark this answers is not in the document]
akoshelev: I see, this is a trait method, so we should be with this contract here

ipa-core/src/net/client/mod.rs
akoshelev: Maybe @andyleiserson can chime in too - naming is hard. I don't think MPC connection can be classified as partitions, so `ConnectionFlavor` looks reasonable to me

ipa-core/src/net/client/mod.rs
akoshelev: It is actually `Mpc` that bothers me. Shard connections shouldn't be confused with Mpc and having a struct that  has `Mpc` in its name but serving shard-to-shard connections looks confusing enough to me

ipa-core/src/net/client/mod.rs
akoshelev: Andy's suggestion is even better I think.

ipa-core/src/net/mod.rs
akoshelev: This PR seems to  change `ClientIdentity` meaning to be able to represent both shard and MPC origins, so I don't think it is tied up to MPC connections only.

ipa-core/src/sharding.rs
akoshelev: This is more about naming rather than documentation

cberkhoff: How about: `HelperToHelper` and `ShardToShard` ?
Or `Ring` and `ShardToShard` ?

Let me know what you think is clearer.

ipa-core/src/sharding.rs
akoshelev: Probably worth consolidating this in `net` module in this case

ipa-core/src/net/mod.rs
cberkhoff: sorry, I meant to say `HelperIdentity` instead of `ClientIdentity`. Should I rename to `HTTP_HELPER_ID_HEADER` instead?

ipa-core/src/sharding.rs
akoshelev: this is def a bikeshedding territory, but this struct names don't matter in isolation because they are never used directly and only used to disambiguate "parent" struct. So `Client<HelperToHelper>` looks less concise than `Client<Helper>` to me

ipa-core/src/net/client/mod.rs
cberkhoff: Mmm... What about `IpaClient<Helper>` and `IpaClient<Shard>` ?

akoshelev: sure that looks great

ipa-core/src/net/mod.rs
```diff
@@ -22,8 +28,11 @@ pub use error::Error;
 pub use server::{MpcHelperServer, TracingSpanMaker};
 pub use transport::{HttpShardTransport, HttpTransport};
 
-pub const APPLICATION_JSON: &str = "application/json";
-pub const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
+const APPLICATION_JSON: &str = "application/json";
+const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
+static HTTP_HELPER_ID_HEADER: HeaderName = HeaderName::from_static("x-unverified-helper-identity");
+pub static HTTP_SHARD_INDEX_HEADER: HeaderName =
```
akoshelev, line 34: Do you need it to be `pub`?

akoshelev (approved):

ipa-core/src/net/mod.rs
cberkhoff: No, I'll change this in PR 2

PR #1349, opened 2024-10-15
[Sharding 2] TransportRestriction-aware Config and Server

cberkhoff:
This is the second of a series of pull requests (PR) to enable sharding on IPA. See [1](https://github.com/private-attribution/ipa/pull/1348). Until 1 is merged, this PR will show both 1 and 2 changes.

**Configs** now contain a Vec of PeerConfigs and Identities so that we can re-use Config for Ring and Shard communication. In spite of its internal representation, `Config<Ring>` provides a familiar interface with arrays (except in cases where Vec was simpler). Didn't felt the need to add more Config tests since it's exercised by `TestConfigBuilder` in many places. In an upcoming PR (Sharding 3) the new API is going to be further utilized.
Also, moved the `identify_cert` function from the Server into the Config. This allows the Server to just depend on the respective Sharded or Ring Config to find the appropriate certificate depending on the case.

**Server** was generalized to be able to work on Sharded and Ring cases. For the most part this only meant separating the client identification layer. Also, Server, unlike Client, doesn't change its API depending on the TransportRestriction. Instead we will initialize a different Axum router when creating.

### DIFF
--- a/ipa-core/src/bin/helper.rs
+++ b/ipa-core/src/bin/helper.rs
@@ -127,7 +127,7 @@ async fn server(args: ServerArgs) -> Result<(), BoxError> {
                 }),
             )
         }
-        (None, None) => (ClientIdentity::Helper(my_identity), None),
+        (None, None) => (ClientIdentity::Header(my_identity), None),
         _ => panic!("should have been rejected by clap"),
     };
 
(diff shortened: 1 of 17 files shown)


cberkhoff:
Deleted in favor of #1360

PR #1350, opened 2024-10-15
Use record IDs to index proof batches

### LINKED ISSUE #1269: DZKPs should use record IDs, not steps, to iterate over proof batches
The DZKP malicious security implementation uses steps to iterate over proof batches, meaning it does something like `ctx.narrow(&EnumeratedStep(batch_num))`. This implies a fixed upper limit on the query size we can handle, determined by the proof size, and the dimension of `EnumeratedStep`. It's possible it will reach the input size we need to support, but it's also wasteful of the somewhat precious space of compact gate IDs to do things this way. When adding compact gate support for malicious security, I reduced the DZKP batch limit from 256 -> 192 to keep the gate count under 100k. (At the time, that was where clippy started complaining about not having underscores in the gate ID literals. The clippy warning has since been waived, and the limit is now higher.)

The MAC implementation is already using record IDs. Doing so requires some annoying record ID arithmetic. For MACs it's not too bad, for DZKPs it will be worse, but still seems manageable, and it's more correct for things that scale with the input size to use record IDs for enumeration.

andyleiserson:
Since we are not guaranteed that all proofs have the same size, I serialize the proof values into fixed-size structures and use one record ID per proof, rather than doing record ID arithmetic, except for PRSS.

Fixes #1269


andyleiserson:
Draft, 5M: https://draft-mpc.vercel.app/query/view/sharp-mover2024-10-15T2018

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -52,6 +55,17 @@ pub const TARGET_PROOF_SIZE: usize = 8192;
...
+/// Maximum proof recursion depth.
+//
+// This is a hard limit. Each GF(2) multiply generates four G values and four H values,
+// so the restriction is:
+// $$ large_recursion_factor * small_recursion_factor ^ (depth - 1) >= 4 * target_proof_size $$
+// With large_recursion_factor = 32 and small_recursion_factor = 8, this means:
+// $$ depth >= log_8 target_proof_size $$
+// Because multiplication intermediate storage gets rounded up to blocks of 256, leaving
+// some margin is advised.
+pub const MAX_PROOF_RECURSION: usize = 9;
```
andyleiserson, line 67: The math written here is not quite right. With the correct math, recursion depth of 9 corresponds to TARGET_PROOF_SIZE of about 117M. Without quicksort batching, this is not quite enough. So either we can wait for quicksort batching, or we can temporarily increase MAX_PROOF_RECURSION slightly. I lean towards the former unless we identify an urgency to take this change.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -88,32 +103,41 @@ impl ProofBatch {
...
-        // approximate length of proof vector (rounded up)
-        let uv_len_bits: u32 = usize::BITS - uv_values.len().leading_zeros();
-        let small_recursion_factor_bits: u32 = usize::BITS - SRF.leading_zeros();
-        let expected_len = 1 << (uv_len_bits - small_recursion_factor_bits);
+        // `MAX_PROOF_RECURSION - 2` because:
+        //  * The first level of recursion has already happened.
+        //  * We need (SRF - 1) at the last level to have room for the masks.
+        let max_uv_values: usize =
+            (SRF - 1) * SRF.pow(u32::try_from(MAX_PROOF_RECURSION - 2).unwrap());
+        tracing::info!("uv_values.len() = {}", uv_values.len());
```
akoshelev, line 116: trace?

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -165,52 +189,40 @@ impl ProofBatch {
...
-            &communication_ctx.recv_channel::<Fp61BitPrime>(ctx.role().peer(Direction::Right));
-
-        // receive from the right
+        assert!(length <= ARRAY_LEN);
         Ok(ctx
-            .parallel_join(
-                (0..length)
-                    .map(|i| async move { receive_channel_right.receive(RecordId::from(i)).await }),
-            )
+            .set_total_records(TotalRecords::Indeterminate)
```
akoshelev, line 220: oh, this is much better than parallel join, can't wait until I see the impact of this change

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -165,52 +189,40 @@ impl ProofBatch {
...
     /// ## Errors
     /// Propagates errors from receiving values over the network channel.
-    pub async fn receive_from_right<C>(ctx: &C, length: usize) -> Result<Self, Error>
+    ///
+    /// ## Panics
+    /// If the recursion depth implied by `length` exceeds `MAX_PROOF_RECURSION`.
+    pub async fn receive_from_right<C>(
+        ctx: &C,
+        record_id: RecordId,
+        length: usize,
```
akoshelev, line 213: How often do we need to shrink?

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -88,32 +103,41 @@ impl ProofBatch {
...
-        let (my_p_mask, p_mask_from_right_prover) = ctx.prss().generate_fields(record_counter);
-        record_counter += 1;
+        let (my_p_mask, p_mask_from_right_prover) =
+            ctx.prss().generate_fields(prss_record_ids.next().unwrap());
         // Prover `P_i` and verifier `P_{i+1}` both compute q(x)
         // therefore the "left" share computed by this verifier corresponds to that which
         // was used by the prover to the left.
-        let (q_mask_from_left_prover, my_q_mask) = ctx.prss().generate_fields(record_counter);
-        record_counter += 1;
+        let (q_mask_from_left_prover, my_q_mask) =
```
akoshelev, line 139: `record_id_range.next().unwrap()` is dominantly used everywhere it seems. I think we should either implement a helper function to make it look less ugly at callsite and potentially not implement the iterator trait for it as we don't have (unless I missed it) `for x in range` use case

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
andyleiserson: I just wanted to see the values in the draft run, I am going to remove this.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
andyleiserson: I don't expect that this is big enough contributor that the difference will show up in overall runtime, but I do think that getting rid of the parallel join more than offsets any cost of space overhead in the fixed-size structures.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
andyleiserson: Since many of our batches end up being smaller than the target, reasonably often... but since we do this only once per proof, it doesn't seem worth a lot of effort to optimize.

ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
andyleiserson: Thoughts on what to call the helper? Maybe `expect_next()`?

andyleiserson:
Draft run, 25M: https://draft-mpc.vercel.app/query/view/flaky-stern2024-10-17T2136. 3 hr 16 min, slightly slower than the run in #1355, which was 3 hr 4 min. Not sure if the difference is real or just variation.

akoshelev (approved):

akoshelev:
> Not sure if the difference is real or just variation.
seems more than just variation to me, but we could try a few runs.

andyleiserson:
> > Not sure if the difference is real or just variation.
> 
> seems more than just variation to me, but we could try a few runs.

Yeah, I think it is real. From some smaller runs, the difference appears to be in the share conversion step. I suspect it has to do with the way that OrderingSender sequences communications (record n+1 is not accepted into the channel until after record n) somehow reducing the amount of validation work that can be done in parallel, but I haven't completely worked out how.

PR #1351, opened 2024-10-15
Hybrid Impression Report Encryption

tyurek:
Some initial code to serialize and encrypt Hybrid Impression Reports. If this all looks good, I'll follow up with Conversion Reports.

### DIFF
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -96,19 +96,15 @@ impl From<io::Error> for CryptError {
 /// If ciphertext cannot be opened for any reason.
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a, R: PrivateKeyRegistry>(
-    key_registry: &R,
+pub fn open_in_place<'a>(
+    sk: &IpaPrivateKey,
     enc: &[u8],
     ciphertext: &'a mut [u8],
-    key_id: u8,
     info: &[u8],
 ) -> Result<&'a [u8], CryptError> {
     let encap_key = <IpaKem as hpke::Kem>::EncappedKey::from_bytes(enc)?;
     let (ct, tag) = ciphertext.split_at_mut(ciphertext.len() - AeadTag::<IpaAead>::size());
     let tag = AeadTag::<IpaAead>::from_bytes(tag)?;
-    let sk = key_registry
-        .private_key(key_id)
-        .ok_or(CryptError::NoSuchKey(key_id))?;
 
     single_shot_open_in_place_detached::<_, IpaKdf, IpaKem>(
         &OpModeR::Base,
@@ -135,20 +131,15 @@ pub(crate) type Ciphertext<'a> = (
 
 /// ## Errors
 /// If the match key cannot be sealed for any reason.
-pub(crate) fn seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
-    key_registry: &K,
+pub(crate) fn seal_in_place<'a, R: CryptoRng + RngCore>(
+    pk: &IpaPublicKey,
     plaintext: &'a mut [u8],
-    key_id: u8,
     info: &[u8],
     rng: &mut R,
 ) -> Result<Ciphertext<'a>, CryptError> {
-    let pk_r = key_registry
-        .public_key(key_id)
-        .ok_or(CryptError::NoSuchKey(key_id))?;
-
     let (encap_key, tag) = single_shot_seal_in_place_detached::<IpaAead, IpaKdf, IpaKem, _>(
         &OpModeS::Base,
-        pk_r,
+        pk,
         info,
         plaintext,
         &[],
@@ -167,6 +158,7 @@ mod tests {
     use rand_core::{CryptoRng, RngCore, SeedableRng};
     use typenum::Unsigned;
 
+    use super::{PrivateKeyRegistry, PublicKeyRegistry};
     use crate::{
         ff::{Gf40Bit, Serializable as IpaSerializable},
         hpke::{open_in_place, seal_in_place, CryptError, Info, IpaAead, KeyPair, KeyRegistry},
@@ -229,9 +221,11 @@ mod tests {
             match_key.serialize(&mut plaintext);
 
             let (encap_key, ciphertext, tag) = seal_in_place(
-                &self.registry,
+                self.registry
+                    .public_key(info.key_id)
+                    .ok_or(CryptError::NoSuchKey(info.key_id))
+                    .unwrap(),
                 plaintext.as_mut_slice(),
-                info.key_id,
                 &info.to_bytes(),
                 &mut self.rng,
             )
@@ -282,10 +276,11 @@ mod tests {
             )
             .unwrap();
             open_in_place(
-                &self.registry,
+                self.registry
+                    .private_key(info.key_id)
+                    .ok_or(CryptError::NoSuchKey(info.key_id))?,
                 &enc.enc,
                 enc.ct.as_mut(),
-                info.key_id,
                 &info.to_bytes(),
             )?;
 
@@ -472,7 +467,8 @@ mod tests {
                     _ => panic!("bad test setup: only 5 fields can be corrupted, asked to corrupt: {corrupted_info_field}")
                 };
 
-                open_in_place(&suite.registry, &encryption.enc, &mut encryption.ct, info.key_id, &info.to_bytes()).unwrap_err();
+                open_in_place(suite.registry.private_key(info.key_id)
+                .ok_or(CryptError::NoSuchKey(info.key_id))?, &encryption.enc, &mut encryption.ct, &info.to_bytes()).unwrap_err();
             }
         }
     }
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -127,6 +127,8 @@ pub fn open_in_place<'a, R: PrivateKeyRegistry>(
 }
 
 /// Version of `open_in_place` that doesn't require Info struct.
+/// ## Errors
+/// If ciphertext cannot be opened for any reason.
 pub fn hybrid_open_in_place<'a, R: PrivateKeyRegistry>(
     key_registry: &R,
     enc: &[u8],
@@ -145,7 +147,7 @@ pub fn hybrid_open_in_place<'a, R: PrivateKeyRegistry>(
         &OpModeR::Base,
         sk,
         &encap_key,
-        &info,
+        info,
         ct,
         &[],
         &tag,
@@ -192,6 +194,8 @@ pub(crate) fn seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
 }
 
 /// Version of `seal_in_place` that doesn't require Info struct.
+/// ## Errors
+/// If the match key cannot be sealed for any reason.
 pub(crate) fn hybrid_seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
     key_registry: &K,
     plaintext: &'a mut [u8],
@@ -206,7 +210,7 @@ pub(crate) fn hybrid_seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegis
     let (encap_key, tag) = single_shot_seal_in_place_detached::<IpaAead, IpaKdf, IpaKem, _>(
         &OpModeS::Base,
         pk_r,
-        &info,
+        info,
         plaintext,
         &[],
         rng,
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -126,6 +126,36 @@ pub fn open_in_place<'a, R: PrivateKeyRegistry>(
     Ok(pt)
 }
 
+/// Version of `open_in_place` that doesn't require Info struct.
+pub fn hybrid_open_in_place<'a, R: PrivateKeyRegistry>(
+    key_registry: &R,
+    enc: &[u8],
+    ciphertext: &'a mut [u8],
+    key_id: u8,
+    info: &[u8],
+) -> Result<&'a [u8], CryptError> {
+    let encap_key = <IpaKem as hpke::Kem>::EncappedKey::from_bytes(enc)?;
+    let (ct, tag) = ciphertext.split_at_mut(ciphertext.len() - AeadTag::<IpaAead>::size());
+    let tag = AeadTag::<IpaAead>::from_bytes(tag)?;
+    let sk = key_registry
+        .private_key(key_id)
+        .ok_or(CryptError::NoSuchKey(key_id))?;
+
+    single_shot_open_in_place_detached::<_, IpaKdf, IpaKem>(
+        &OpModeR::Base,
+        sk,
+        &encap_key,
+        &info,
+        ct,
+        &[],
+        &tag,
+    )?;
+
+    // at this point ct is no longer a pointer to the ciphertext.
+    let pt = ct;
+    Ok(pt)
+}
+
 // Avoids a clippy "complex type" warning on the return type from `seal_in_place`.
 // Not intended to be widely used.
 pub(crate) type Ciphertext<'a> = (
@@ -144,6 +174,31 @@ pub(crate) fn seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
 ) -> Result<Ciphertext<'a>, CryptError> {
     let key_id = info.key_id;
     let info = info.to_bytes();
+    let pk_r = key_registry
+        .public_key(key_id)
+        .ok_or(CryptError::NoSuchKey(key_id))?;
+
+    let (encap_key, tag) = single_shot_seal_in_place_detached::<IpaAead, IpaKdf, IpaKem, _>(
+        &OpModeS::Base,
+        pk_r,
+        &info,
+        plaintext,
+        &[],
+        rng,
+    )?;
+
+    // at this point `plaintext` is no longer a pointer to the plaintext.
+    Ok((encap_key, plaintext, tag))
+}
+
+/// Version of `seal_in_place` that doesn't require Info struct.
+pub(crate) fn hybrid_seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
+    key_registry: &K,
+    plaintext: &'a mut [u8],
+    key_id: u8,
+    info: &[u8],
+    rng: &mut R,
+) -> Result<Ciphertext<'a>, CryptError> {
     let pk_r = key_registry
         .public_key(key_id)
         .ok_or(CryptError::NoSuchKey(key_id))?;
--- a/ipa-core/src/hpke/mod.rs
+++ b/ipa-core/src/hpke/mod.rs
@@ -100,39 +100,6 @@ pub fn open_in_place<'a, R: PrivateKeyRegistry>(
     key_registry: &R,
     enc: &[u8],
     ciphertext: &'a mut [u8],
-    info: &Info,
-) -> Result<&'a [u8], CryptError> {
-    let key_id = info.key_id;
-    let info = info.to_bytes();
-    let encap_key = <IpaKem as hpke::Kem>::EncappedKey::from_bytes(enc)?;
-    let (ct, tag) = ciphertext.split_at_mut(ciphertext.len() - AeadTag::<IpaAead>::size());
-    let tag = AeadTag::<IpaAead>::from_bytes(tag)?;
-    let sk = key_registry
-        .private_key(key_id)
-        .ok_or(CryptError::NoSuchKey(key_id))?;
-
-    single_shot_open_in_place_detached::<_, IpaKdf, IpaKem>(
-        &OpModeR::Base,
-        sk,
-        &encap_key,
-        &info,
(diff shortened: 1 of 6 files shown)


ipa-core/src/hpke/mod.rs
```diff
@@ -126,6 +126,38 @@ pub fn open_in_place<'a, R: PrivateKeyRegistry>(
     Ok(pt)
 }
 
+/// Version of `open_in_place` that doesn't require Info struct.
+/// ## Errors
+/// If ciphertext cannot be opened for any reason.
+pub fn hybrid_open_in_place<'a, R: PrivateKeyRegistry>(
```
cberkhoff, line 132: Is it worth avoid duplication between this and `open_in_place` ?

ipa-core/src/hpke/mod.rs
```diff
@@ -161,6 +193,33 @@ pub(crate) fn seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
     Ok((encap_key, plaintext, tag))
 }
 
+/// Version of `seal_in_place` that doesn't require Info struct.
+/// ## Errors
+/// If the match key cannot be sealed for any reason.
+pub(crate) fn hybrid_seal_in_place<'a, R: CryptoRng + RngCore, K: PublicKeyRegistry>(
```
cberkhoff, line 199: Same as above. Should we avoid duplicating?

ipa-core/src/report/hybrid.rs
```diff
@@ -1,18 +1,79 @@
...
+        KeyIdentifier,
+    },
     secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
 };
 
+// TODO(679): This needs to come from configuration.
+static HELPER_ORIGIN: &str = "github.com/private-attribution";
+
+#[derive(Debug)]
+pub struct NonAsciiStringError {
```
cberkhoff, line 32: Can you add a few tests for this error? Specially for the cases dealing with non ascii and how does it handle them?

What happens if the input isn't even utf8?

ipa-core/src/report/hybrid.rs
```diff
@@ -22,6 +83,114 @@ where
...
 
+impl<BK: SharedValue> Serializable for HybridImpressionReport<BK>
+where
+    BK: SharedValue,
+    Replicated<BK>: Serializable,
+    <Replicated<BK> as Serializable>::Size: Add<U16>,
+    <<Replicated<BK> as Serializable>::Size as Add<<Replicated<BA64> as Serializable>::Size>>:: Output: ArrayLength,
+{
+    type Size = <<Replicated<BK> as Serializable>::Size as Add<<Replicated<BA64> as Serializable>::Size>>:: Output;
+    type DeserializationError = InvalidHybridReportError; //<Replicated<BK> as Serializable>::DeserializationError;
```
cberkhoff, line 94: nit: remove commented code

ipa-core/src/report/hybrid.rs
```diff
@@ -22,6 +83,114 @@ where
...
+
+        self.breakdown_key
+            .serialize(GenericArray::from_mut_slice(&mut buf[mk_sz..mk_sz + bk_sz]));
+    }
+    fn deserialize(buf: &GenericArray<u8, Self::Size>) -> Result<Self, Self::DeserializationError> {
+        let mk_sz = <Replicated<BA64> as Serializable>::Size::USIZE;
+        let bk_sz = <Replicated<BK> as Serializable>::Size::USIZE;
+        let match_key =
+            Replicated::<BA64>::deserialize(GenericArray::from_slice(&buf[..mk_sz]))
+            .map_err(|e| InvalidHybridReportError::DeserializationError("match_key", e.into()))?;
```
cberkhoff, line 111: is there a way to show more context in this error? Like where did it happen?

I'm imagining we get this error and we need to find the issue

ipa-core/src/report/hybrid.rs
```diff
@@ -22,6 +83,114 @@ where
...
+    }
+    /// # Errors
+    /// If there is a problem encrypting the report.
+    pub fn encrypt<R: CryptoRng + RngCore>(
+        &self,
+        key_id: KeyIdentifier,
+        key_registry: &impl PublicKeyRegistry,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, InvalidHybridReportError> {
+        let mut out = Vec::with_capacity(usize::from(self.encrypted_len()));
```
cberkhoff, line 142: how large are these reports? Does `out` fit in memory?

ipa-core/src/report/hybrid.rs
```diff
@@ -341,4 +606,52 @@ mod test {
         let expected_err = unique_bytes.check_duplicates(&[bytes4, bytes3]);
         assert!(matches!(expected_err, Err(Error::DuplicateBytes(6))));
     }
+
+    #[test]
+    fn serialization_hybrid_impression() {
```
cberkhoff, line 611: Can a test be added that tests serialization against a pre-existing value? 

This test is making sure that serialize and deserialize are mirrored but if both methods make the same mistake then you wouldn't know.

cberkhoff (commented):
I just have a few questions and suggestions.

ipa-core/src/hpke/mod.rs
tyurek: Probably. I didn't want to step on anyone's toes but it would make sense to have just one function with the new interface that doesn't require an Info struct

ipa-core/src/report/hybrid.rs
tyurek: I think these should be pretty small. In the test it's something like 116 bytes. I'd check with @eriktaubeneck though

ipa-core/src/report/hybrid.rs
tyurek: Hmm, good point. I see that ipa.rs has some tests that use constants. Do you happen to know how those were generated?

ipa-core/src/report/hybrid.rs
tyurek: I added a non-ascii test, but it looks like rust doesn't even support strings that aren't utf-8

ipa-core/src/report/hybrid.rs
tyurek: Screenshot 2024-10-16 at 2 41 40 PM
Rust-analyzer indicates that this operation is infallible. I would guess that's because it looks like this can take arbitrary values, but then why isn't `Replicated::<BK>::deserialize` infallible?

ipa-core/src/report/hybrid.rs
cberkhoff: Right, Strings in rust they are utf-8. If the error can happen on arbitrary `[u8]`... do you need to make`[u8]` your `input` instead of `String` in order to capture what was the reason of the error?

In other words, this error is just for non ascii or for any non-conforming string?

ipa-core/src/report/hybrid.rs
cberkhoff: I thought `out` would contain all the reports. Not worried about the test use case, but more concerned with the real use case, when we have 1 billion events.

ipa-core/src/hpke/mod.rs
tyurek: I'll just go ahead and do it, looks like these functions are only used one other place and it's easy enough to change

ipa-core/src/report/hybrid.rs
tyurek: Oh, I see. The implementation for OprfReport uses a Vec. Are the hybrid reports expected to be larger?

ipa-core/src/report/hybrid.rs
tyurek: Ohhh, yeah I see the problem now. Good catch. I just removed from [u8] capability. Requiring a string should guarantee that it's utf-8

ipa-core/src/hpke/mod.rs
```diff
@@ -100,10 +100,9 @@ pub fn open_in_place<'a, R: PrivateKeyRegistry>(
     key_registry: &R,
     enc: &[u8],
     ciphertext: &'a mut [u8],
-    info: &Info,
+    key_id: u8,
```
akoshelev, line 103: Previously we used key id to authenticate ct and to pick the appropriate private key from `key_registry`. Based on this change, I think we should no longer take `key_registry` and instead just take `&PrivateKey` parameter without `key_id`. I don't remember why we decided to authenticate the cipher with key id, but I agree this doesn't seem necessary

ipa-core/src/report/hybrid.rs
akoshelev: It doesn't seem enough to me to require UTF-8. Any encoding that requires 2 bytes or more would not be a valid ASCII, so we still need to check `is_ascii`

ipa-core/src/report/hybrid.rs
akoshelev: >  why isn't Replicated::< BK >::deserialize infallible
two reasons.
* We use prime fields where values > PRIME are rejected at deserialization layer to avoid silent truncation
* We use bitstrings like BA3, BA5 where we also reject values greater than MAX

Our macro implementation guarantees that Replicated<BA8> is infallible, but generic `Replicated<S>` is not (unless a trait bound is added here)

For Replicated::<BA64> I think we could use `deserialize_infallible` method that should be available

ipa-core/src/report/hybrid.rs
```diff
@@ -22,6 +70,114 @@ where
...
+            Replicated::<BK>::deserialize(GenericArray::from_slice(&buf[mk_sz..mk_sz + bk_sz]))
+            .map_err(|e| InvalidHybridReportError::DeserializationError("breakdown_key", e.into()))?;
+        Ok(Self { match_key, breakdown_key })
+    }
+}
+
+impl<BK> HybridImpressionReport<BK>
+where
+    BK: SharedValue,
+    Replicated<BK>: Serializable,
```
akoshelev, line 109: my hope was to drop these trait bounds and do runtime checks at serialization time. Would that be possible with this approach?

ipa-core/src/report/hybrid.rs
```diff
@@ -80,6 +214,119 @@ where
...
+
+impl<BK, B> EncryptedHybridImpressionReport<BK, B>
+where
+    B: Deref<Target = [u8]>,
+    BK: SharedValue,
+    Replicated<BK>: Serializable,
+    <Replicated<BK> as Serializable>::Size: Add<U16>,
+    <<Replicated<BK> as Serializable>::Size as Add<U16>>::Output: ArrayLength,
+{
+    const ENCAP_KEY_MK_OFFSET: usize = 0;
```
akoshelev, line 235: same here. If the only reason for these bounds is to be able to write `const E: X + Y + ... + Replicated<BA64>::Size::USIZE` instead of `const E = X + Y + Replicated<BA>::size()`, then it is not worth it imo

ipa-core/src/report/hybrid.rs
```diff
@@ -162,6 +409,21 @@ impl UniqueBytes for EncryptedHybridReport {
...
+where
+    B: Deref<Target = [u8]>,
+    BK: SharedValue,
+    Replicated<BK>: Serializable,
+    <Replicated<BK> as Serializable>::Size: Add<U16>,
+    <<Replicated<BK> as Serializable>::Size as Add<U16>>::Output: ArrayLength,
+{
+    /// We use the `TagSize` (the first 16 bytes of the ciphertext) for collision-detection
+    /// See [analysis here for uniqueness](https://eprint.iacr.org/2019/624)
+    fn unique_bytes(&self) -> Vec<u8> {
```
akoshelev, line 422: I think it would be more flexible to return `&[u8]` as it would work for both cases: when you need to peek into those bytes and when you need to clone them. This API always clones which may not be desirable

ipa-core/src/report/hybrid.rs
akoshelev: Note that you may want https://github.com/private-attribution/ipa/pull/1357

ipa-core/src/report/hybrid.rs
tyurek: I was thinking we could have a different PR where we redo a bunch of serialization, so that we can take care of it all at once. Does that work?

ipa-core/src/report/hybrid.rs
tyurek: Makes sense. I'll change it

ipa-core/src/report/hybrid.rs
tyurek: I didn't see an easy way to use `deserialize_infallible` so I went with `unwrap_infallible`

ipa-core/src/hpke/mod.rs
tyurek: this function is shared with the OprfReport implementation. Is it okay to alter that?

ipa-core/src/report/hybrid.rs
tyurek: I'm confused, where do you want to add the `is_ascii` check? The function that can throw the NonAsciiStringError (the new() function for HybridImpressionInfo) calls `is_ascii` and throws this error if not. But the error has the option to print the non-ascii string, so I don't think adding a `is_ascii` check there makes sense.

ipa-core/src/report/hybrid.rs
tyurek: That would be cool. I'll look into that when it lands :)

ipa-core/src/report/hybrid.rs
akoshelev: yea, I just wanted to clarify that we can't simply check for UTF8 and be done with it. The `is_ascii` check still needs to be in place in order for this to be correct

ipa-core/src/hpke/mod.rs
akoshelev: I think it should be ok to change this, given that we need to update `OprfReport` to call the new version of it anyway

ipa-core/src/report/hybrid.rs
eriktaubeneck: HybridReports are likely a little smaller than OprfReports. This method only encrypts an individual report, so somewhere between 100-150 bytes. 

When we are generating a lot of them (for the 1B case), we will have to worry about memory (probably streaming them into a file, for example), but that happens elsewhere. Using a Vec for individual reports seems to be the right approach as we cannot know the actual length at compile time, due to some of the associated data being arbitrary length strings.

ipa-core/src/report/hybrid.rs
eriktaubeneck: > same here. If the only reason for these bounds is to be able to write const E: X + Y + ... + Replicated<BA64>::Size::USIZE instead of const E = X + Y + Replicated<BA>::size(), then it is not worth it imo

agreed. in fact, I don't think there is any advantage at all to having these be known at compile time. even with all these bounds, we can still submit bad input which will cause an error.

ipa-core/src/report/hybrid.rs
eriktaubeneck: I'm not sure, but I would venture to guess they we're generated by the current implementation. @cberkhoff do you have any ideas for how to accomplish what you're suggesting?

ipa-core/src/report/hybrid.rs
akoshelev: Should be relatively straightforward

```rust
        let match_key =
            Replicated::<BA64>::deserialize_infallible(GenericArray::from_slice(&buf[..mk_sz]));
```

it seems to compile for me

tyurek: oh hmm, I thought I'd have to implement it or something because rust-analyzer said it didn't recognize `deserialize_infallible`, but I guess it was just slow and I should have tried compiling lol

ipa-core/src/report/hybrid.rs
akoshelev: I think this is a good design because we also have `encrypt_to` that allows to use reusable buffer for encryptions if this ever becomes a bottleneck

ipa-core/src/report/hybrid.rs
cberkhoff: Can't we do something simple like?

```
let input = "<hybrid-report>";
assert_eq!(serialize(input), expectation);
```

tyurek: If expectation is generated using the serialize function, wouldn't you have the same potential issue of both methods making the same mistake?

cberkhoff: My idea was that the input string would be "set in stone" and easy to see and review. Since you would be writing it down you can make sure it's exactly what you want it to be. I imagine you would use serialize to do this once and then copy paste it.

tyurek: I just added a few lines to the end of the test to show deserialzation from a bytestring. Let me know if this is what you had in mind

cberkhoff: I think this is ok. Maybe testing serialization would have been more legible (my bad, I did mention deserialization before).

Also having this as a separate/independent test would make it easier to triage if issues happen.

ipa-core/src/report/hybrid_info.rs
```diff
@@ -0,0 +1,65 @@
+use crate::report::{hybrid::NonAsciiStringError, KeyIdentifier};
+
+const DOMAIN: &str = "private-attribution";
+
+#[derive(Debug)]
```
cberkhoff, line 5: Why only this one has `Debug`?

ipa-core/src/report/ipa.rs
```diff
@@ -407,11 +407,25 @@ where
...
             *GenericArray::from_slice(self.mk_ciphertext());
-        let plaintext_mk = open_in_place(key_registry, self.encap_key_mk(), &mut ct_mk, &info)?;
+        let plaintext_mk = open_in_place(
+            key_registry
+                .private_key(self.key_id())
+                .ok_or(CryptError::NoSuchKey(self.key_id()))?,
+            self.encap_key_mk(),
+            &mut ct_mk,
+            &info.to_bytes(),
+        )?;
```
cberkhoff, line 417: Maybe it's more readable if this is separate:
```
let key = key_registry
                .private_key(self.key_id())
                .ok_or(CryptError::NoSuchKey(self.key_id()))?;

let plaintext_mk = open_in_place(
            key,
            self.encap_key_mk(),
            &mut ct_mk,
            &info.to_bytes(),
        )?;
```

cberkhoff (approved):
I'm ok as it is. I just mentioned a few small improvements.

ipa-core/src/hpke/mod.rs
```diff
@@ -231,9 +221,12 @@ mod tests {
             match_key.serialize(&mut plaintext);
 
             let (encap_key, ciphertext, tag) = seal_in_place(
-                &self.registry,
+                self.registry
+                    .public_key(info.key_id)
+                    .ok_or(CryptError::NoSuchKey(info.key_id))
+                    .unwrap(),
```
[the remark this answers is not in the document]
akoshelev: this requires changing the contract of this function and updating callers. Given that encryption is not used inside MPC, I am less worried about this going wrong. Although it may be nice to do this work eventually

ipa-core/src/hpke/mod.rs
```diff
@@ -96,26 +96,21 @@ impl From<io::Error> for CryptError {
...
 ///
 /// [`HPKE decryption`]: https://datatracker.ietf.org/doc/html/rfc9180#name-encryption-and-decryption
-pub fn open_in_place<'a, R: PrivateKeyRegistry>(
-    key_registry: &R,
+pub fn open_in_place<'a>(
+    sk: &IpaPrivateKey,
     enc: &[u8],
     ciphertext: &'a mut [u8],
-    info: &Info,
+    info: &[u8],
```
akoshelev, line 103: maybe you can take `R: AsRef<[u8]>` here to make it work smoothly with box and other smart pointers

ipa-core/src/report/hybrid.rs
```diff
@@ -1,20 +1,69 @@
...
 
+// TODO(679): This needs to come from configuration.
+static HELPER_ORIGIN: &str = "github.com/private-attribution";
+
+#[derive(Debug)]
+pub struct NonAsciiStringError {
+    input: String,
+}
+
+impl Display for NonAsciiStringError {
```
akoshelev, line 39: nit: you can derive `thiserror::error` and save a few keystrokes here

```rust
#[derive(Debug, thiserror::Error)]
#[error("string contains ...")]
pub struct NonAsciiStringError
```

ipa-core/src/report/hybrid.rs
akoshelev: That PR is merged, so we can follow up on it

akoshelev (approved):

ipa-core/src/report/hybrid_info.rs
tyurek: because that's the only stuct here that is used elsewhere. It didn't make sense to implement Debug for dead code. The followup PR will change this when it uses these types.

ipa-core/src/report/hybrid.rs
tyurek: After playing around with this a little, it looks like I can't use ::size() when defining a type. This means that I need the bounds to implement Serializable (to define Size type), as well as to implement decryption (to define CTMKLength and CTBTTLength<BK> types). Then for encryption, when I try and remove the Add bounds, it has a problem with SITE_DOMAIN_OFFSET, even when I use size() to define the constants leading up to that.

PR #1355, opened 2024-10-16
Fixes for aggregation batching

andyleiserson:
* Don't start all aggregation batches from record 0.
* Fix a bug in insert_segment_large when segments are added out-of-order.
* Remove unused `last_record` and tweak `is_empty`.
* Add a few tests.

### DIFF
--- a/ipa-core/src/protocol/context/dzkp_validator.rs
+++ b/ipa-core/src/protocol/context/dzkp_validator.rs
@@ -494,6 +494,15 @@ impl Batch {
     /// Creates a new `Batch` for multiplication intermediates from multiple gates. The
     /// first record is specified by `first_record`, or if that is `None`, is set
     /// automatically for each gate the first time a segment from that gate is added.
+    ///
+    /// Once the first record is set, attempting to add a segment before the first
+    /// record will panic. It is likely, but not guaranteed, that protocol execution
+    /// proceeds in order, so a problem here can easily escape testing.
+    ///  * When using the `Batcher` in multi-batch mode, `first_record` is calculated
+    ///    from the batch index and the number of records in a batch, so there is no
+    ///    possibility of attempting to add a record before the start of the batch.
+    ///  * The only protocol that manages batches explicitly is the aggregation protocol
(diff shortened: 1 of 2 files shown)


akoshelev:
Can you elaborate a bit what was causing such long runtimes for aggregation? It is not clear to me how this code fixes it, although I am sure it does

akoshelev (approved):

andyleiserson:
> Can you elaborate a bit what was causing such long runtimes for aggregation?

First some context. [There are two ways of using the `Batcher`](https://github.com/private-attribution/ipa/blob/57fa8a5fc4bccc3f58e33a78fc889bdb9dbeb37c/ipa-core/src/protocol/context/batcher.rs#L13). The first is to call `validate_record` individually for each record. (Sometimes this is done explicitly by the protocol, and sometimes it is done via `validated_seq_join`, which is implemented using `validate_record`.) Most things work this way. The other way of using the batcher is to call `into_single_batch` and validate everything at once. This is used when there is a single `validator.validate()` call.

The problem was in setting `first_record`. There's only one code path for creating batches, [here](https://github.com/private-attribution/ipa/pull/1355/files#diff-3a91a8e8da33a607b15c2499881f7b1ce3dd3f4f26ca6b3cef971ce105239677L806). Even unbatched (or manually batched) protocols use this code path, and then later do `into_single_batch` as described above.

Aggregation repeatedly creates validators. Each time a validator is created, a new batcher is created, with a `batch_index` of zero. Prior to this fix, this in turn created a batch with a `first_record` of zero. When multiplication intermediates were added to the batch, the code pushed as many zeros as necessary onto the store to reach the record ID being added. So every batch contained values for all record IDs from zero to the current record. (Most of the values were zero, but they were included in the proof anyways.)

With the fix, in the case where max_multiplications_per_gate (records_per_batch) is unlimited (`usize::MAX`), `first_record` is left unset when the batch is created and is set from the first multiplication intermediates added to the batch.

andyleiserson:
Draft run, 25M: https://draft-mpc.vercel.app/query/view/lank-lock2024-10-17T1656. 3 hours 4 minutes.

PR #1356, opened 2024-10-17
Add ipa-metrics crate

akoshelev:
Follows the design described here
https://docs.google.com/document/d/1cRnE024bi7KJYeqMT8yMYluIplA8992bNIaeKBmrXp4/edit#heading=h.d8g3vju6yaqd

I have a working solution where this crate is integrated with IPA. There will be 3 separate PRs to make the transition

* [This PR] ipa-metrics crate with core functionality
* ipa-metrics-tracing that adds support for metric partitioning in tests
* Migrate IPA to new metrics engine

### DIFF
--- a/ipa-metrics/src/context.rs
+++ b/ipa-metrics/src/context.rs
@@ -32,10 +32,6 @@ impl CurrentThreadContext {
         METRICS_CTX.with_borrow_mut(MetricsContext::flush);
     }
 
-    pub fn is_connected() -> bool {
-        METRICS_CTX.with_borrow(MetricsContext::is_connected)
-    }
-
     pub fn store<F: FnOnce(&MetricsStore) -> T, T>(f: F) -> T {
         METRICS_CTX.with_borrow(|ctx| f(ctx.store()))
     }
--- a/ipa-metrics/src/context.rs
+++ b/ipa-metrics/src/context.rs
@@ -91,6 +91,10 @@ impl MetricsContext {
     }
 
     fn flush(&mut self) {
+        if self.store.is_empty() {
+            return;
+        }
+
         if self.is_connected() {
             let store = mem::take(&mut self.store);
             match self.tx.as_ref().unwrap().send(store) {
@@ -118,7 +122,9 @@ impl Drop for MetricsContext {
 
 #[cfg(test)]
 mod tests {
-    use crate::MetricsContext;
+    use std::thread;
+
+    use crate::{CurrentThreadPartitionContext, MetricsContext};
 
     /// Each thread has its local store by default, and it is exclusive to it
     #[test]
@@ -126,7 +132,7 @@ mod tests {
     fn local_store() {
         use crate::context::CurrentThreadContext;
 
-        crate::set_partition(0xdeadbeef);
+        CurrentThreadPartitionContext::set(0xdeadbeef);
         counter!("foo", 7);
 
         std::thread::spawn(|| {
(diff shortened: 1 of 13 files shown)


ipa-metrics/src/context.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    fn flush(&mut self) {
+        if self.store.is_empty() {
+            return;
+        }
+
+        if self.is_connected() {
+            let store = mem::take(&mut self.store);
+            match self.tx.as_ref().unwrap().send(store) {
```
andyleiserson, line 98: nit: I would probably do a match on `self.tx.as_ref()` rather than the `is_some` / `unwrap`.

andyleiserson: ... and once again I click "add single comment" when I mean to start a review.

ipa-metrics/src/store.rs
```diff
@@ -0,0 +1,224 @@
...
+            RawEntryMut::Vacant(slot) => {
+                let (_, val) = slot.insert_hashed_nocheck(hash, key.to_owned(), Default::default());
+                CounterHandle { val }
+            }
+        }
+    }
+
+    /// Returns the value for the specified metric taking into account
+    /// its dimensionality. That is (foo, dim1 = 1, dim2 = 2) will be
+    /// different from (foo, dim1 = 1).
```
andyleiserson, line 68: This description is confusing to me. Isn't the behavior of the partial match logic that a query for `(foo, dim1 = 1)` will include data for metric names `(foo, dim1 = 1)`, `(foo, dim1 = 1, dim2 = 2)`.

Taking a stab at rewording it:
> Returns the value for the specified metric, limited by any specified dimensions, but not by any unspecified dimensions. If metric `foo` has dimensions `dim1` and `dim2`, a query for `(foo, dim1 = 1)` will sum the counter values having `dim1 = 1` and any value of `dim2`.

ipa-metrics/src/label.rs
```diff
@@ -0,0 +1,170 @@
...
+        state.write_u64(self.val.hash());
+    }
+}
+
+impl PartialEq for Label<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        // name check should be fast - just pointer comparison.
+        // val check is more involved with dynamic dispatch, so we can consider
+        // making label immutable and storing a hash of the value in place
+        self.name == other.name && self.val.hash() == other.val.hash()
```
andyleiserson, line 71: Implementing `eq` by comparing hashes seems dangerous. Maybe add an `eq` function to the `(Label)Value` trait so that value types can implement `eq` in terms of the hash only when it is safe to do so?

andyleiserson (commented):
I don't see the periodic flushing to the collector in here, is that something that needs to be done as part of the integration?

On the whole this looks good.

ipa-metrics/src/label.rs
akoshelev: Object safety is on our way to provide it :( I added a note to `hash` function that it will be used to check uniqueness. Happy to do more if there is a reasonable way to deal with it.

ipa-metrics/src/key.rs
```diff
@@ -0,0 +1,299 @@
...
+
+pub use Name as MetricName;
+pub(super) use OwnedName as OwnedMetricName;
+
+use crate::label::{Label, OwnedLabel, MAX_LABELS};
+
+#[macro_export]
+macro_rules! metric_name {
+    // Match when two key-value pairs are provided
+    // TODO: enforce uniqueness at compile time
```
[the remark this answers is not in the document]
akoshelev: I thought there may be a way to enforce uniqueness at compile time. `enforce_unique` only works at runtime and it was my quick way get out of the macro writing exercise. 

It feels that there should be (at least const X = 1, const Y = 2, const_assert_ne!(X, Y) must work), I just couldn't find an ergonomic way to write this

ipa-metrics/src/context.rs
```diff
@@ -0,0 +1,168 @@
...
+                Ok(()) => {}
+                Err(e) => {
+                    tracing::warn!("MetricsContext is not connected: {e}");
```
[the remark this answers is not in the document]
akoshelev: yea you're right, we should be able to distinguish between those cases

akoshelev:
> I don't see the periodic flushing to the collector in here, is that something that needs to be done as part of the integration?

Yea I decided not to put it in this iteration. It both simplifies the reviews and I don't see a strong need for that just yet. Test runs on draft shown that metrics get updated fairly frequently. It is possible to add it later without changing the API I think

ipa-metrics/src/label.rs
```diff
@@ -0,0 +1,172 @@
...
+        f.debug_struct("Label")
+            .field("name", &self.name)
+            .field("val", &format!("{}", self.val))
+            .finish()
+    }
+}
+
+impl Hash for Label<'_> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write(self.name.as_bytes());
```
andyleiserson, line 61: Does it make sense to change this to `Hash::hash`? I think that makes this prefix free, which in turn makes the name hashing prefix free.

andyleiserson (approved):

PR #1357, opened 2024-10-17
Add compile time function to compute the size of AdditiveShare

akoshelev:
The benefit of using it is it does not require trait bounds math as `Serializable` trait and it can be used in const context


akoshelev:
It seems that coverage is codecov issue: https://github.com/private-attribution/ipa/actions/runs/11379712512/job/31657642117?pr=1357#step:6:1295. This code is covered

PR #1358, opened 2024-10-17
Sharded decrypt and uniqueness verification

eriktaubeneck:
This makes decryption and uniqueness verification work in a sharded context.

Because we cannot reshard dynamically sized structs (e.g., an EncryptedHybridReport), this takes the approach of:
1. EncryptedHybridReports -> (UniqueTag, HybridReport) (e.g., decryption)
2. Reshard UniqueTags (but not HybridReports)
3. Verify uniqueness over UniqueTags
4. Continue with protocol (TODO)

Now, this is not idea because we have to complete all decryption before we can do the uniqueness check. I attempted to try and get this working with streaming, but ran into a bunch of issues. Still happy to keep trying that approach, but wanted to get this up to get some feedback.


ipa-core/src/report/hybrid.rs
```diff
@@ -150,37 +151,94 @@ impl TryFrom<Bytes> for EncryptedHybridReport {
...
+impl UniqueTag {
+    // Function to attempt to create a UniqueTag from a UniqueBytes implementor
+    pub fn from_unique_bytes<T: UniqueBytes>(item: &T) -> Self {
+        UniqueTag {
+            bytes: item.unique_bytes(),
+        }
+    }
+
+    /// Maps the tag into a consistent shard.
+    /// Note that `ShardIndex` is limited to u32, so we only use the first 4 bytes.
```
akoshelev, line 191: Doesn't it introduce bias? Can we lift shard index to u128, do the modulo operation and then downcast to u32?

ipa-core/src/report/hybrid.rs
```diff
@@ -150,37 +151,94 @@ impl TryFrom<Bytes> for EncryptedHybridReport {
...
+                .try_into()
+                .expect("This is larger than 4 bytes"),
+        );
+        let shard_count = u32::from(shard_count);
+        ShardIndex::from(num % shard_count)
+    }
+}
+
+// This will vaild at compile time if TAG_SIZE doesn't match U16
+const _: [(); 16] = [(); TAG_SIZE];
```
akoshelev, line 209: nice, you could use `const_assert!` macro for that to have even better error message

akoshelev: or even `[(); U16::Size::USIZE]` could work

ipa-core/src/report/hybrid.rs
```diff
@@ -239,9 +296,9 @@ mod test {
         }
     }
 
-    fn generate_random_bytes(size: usize) -> Vec<u8> {
+    fn generate_random_bytes() -> [u8; 16] {
```
akoshelev, line 299:
```suggestion
    fn generate_random_bytes() -> [u8; TAG_SIZE] {
```

ipa-core/src/test_fixture/mod.rs
```diff
@@ -147,6 +147,37 @@ where
     join3(fut0, fut1, fut2)
 }
 
+/// Wrapper for joining the first item from three iterator in an iterator into an array.
+/// # Panics
+/// If the tasks return `Err` or if `a` is the wrong length.
+pub fn join3v0<T, I, V>(a: V) -> JoinAll<T>
```
akoshelev, line 153:
```suggestion
pub fn join3v0<T, I, V>(a: V) -> impl Future<Output = ?> 
```

ipa-core/src/report/hybrid.rs
akoshelev: Based on the usage, it seems that you simply want `fn generate_random_tag() -> UniqueTag`

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -50,41 +49,51 @@ where
...
+                .try_flatten()
+                .take(sz)
+                .try_fold((Vec::new(), Vec::new()), |mut acc, result| async move {
+                    acc.0.push(result.0);
+                    acc.1.push(result.1);
+                    Ok(acc)
+                })
+                .await?;
+
+        let resharded_tags = reshard_iter(ctx.clone(), tags, |ctx, _, tag| {
```
akoshelev, line 86: I think we ned to narrow to a `ReshardTag` step or something along those lines

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -50,41 +49,51 @@ where
...
+                                .decrypt::<R, BA8, BA3, BA20>(key_registry.as_ref())
+                                .map_err(Into::<Error>::into);
+                            let unique_tag = UniqueTag::from_unique_bytes(&enc_report);
+                            dec_report.map(|dec_report1| (dec_report1, unique_tag))
+                        }
+                    }))
+                })
+                .try_flatten()
+                .take(sz)
+                .try_fold((Vec::new(), Vec::new()), |mut acc, result| async move {
```
akoshelev, line 79:
```suggestion
                .try_fold((Vec::with_capacity(sz), Vec::with_capacity(sz)), |mut acc, result| async move {
```

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -50,41 +49,51 @@ where
...
+                    acc.0.push(result.0);
+                    acc.1.push(result.1);
+                    Ok(acc)
+                })
+                .await?;
+
+        let resharded_tags = reshard_iter(ctx.clone(), tags, |ctx, _, tag| {
+            tag.shard_picker(ctx.shard_count())
+        })
+        .await?;
```
akoshelev, line 89: I want to look into this - I think it may be possible to unzip the stream into two and feed tag stream to resharding and keep collecting reports from a separate stream

ipa-core/src/test_fixture/mod.rs
akoshelev: also a better name would be helpful here - I suspect that this may get used quite a bit, so `join3flatten` maybe?

akoshelev (commented):
overall lgtm

ipa-core/src/query/runner/hybrid.rs
eriktaubeneck: agreed. my thought was to land this that works, then give that another attempt (which will make that diff easier to review and also much easier to share what's not working if I get stuck.)

ipa-core/src/report/hybrid.rs
eriktaubeneck: we certainly can, but does it introduce bias? I suppose I'm assuming that all the bytes in the tag are uniformly distributed, so taking 4 random bytes will still be uniformly distributed.

ipa-core/src/report/hybrid.rs
eriktaubeneck: oh cool, let me see if `const_assert!` works. basically I want something to fail if `TAG_SIZE` is no longer 16 bytes so we know to fix this.

ipa-core/src/report/hybrid.rs
eriktaubeneck: oh good call out. I made a simple change here (since it used to be bytes), but your suggestion makes sense!

ipa-core/src/test_fixture/mod.rs
eriktaubeneck: good call out. the comment needs updating too. the original intention was to have it be like `join3v` but only return the result from shard 0, but then the functionality shifted a bit as I was working out how to get everything to actually poll.

ipa-core/src/report/hybrid.rs
akoshelev: I think we can say that about AES tag, but there is no such guarantee on a subset of AAD bytes afaik

eriktaubeneck: I'm pretty sure a uniformly distributed tag implies any subsets of tags are also uniformly distributed. (look at it from the other direction, a non-uniformly random bit must result in the full tag being non-uniformly random.)

doesn't matter though, I'll push a change that does it on u128s.

PR #1359, opened 2024-10-17
fix typo in shuffle/sharded.rs

eriktaubeneck:
pretty sure this is a typo


akoshelev (approved):
indeed, thanks for catching it!

PR #1360, opened 2024-10-18
[Sharding 2] ConnectionFlavor-aware Server and Config

cberkhoff:
This is the second of a series of pull requests (PR) to enable sharding on IPA. See [1](https://github.com/private-attribution/ipa/pull/1348).

**Configs** now contain a Vec of PeerConfigs and Identities so that we can re-use Config for Ring and Shard communication. In spite of its internal representation, `Config<Helper>` provides a familiar interface with arrays (except in cases where Vec was simpler). Didn't felt the need to add more Config tests since it's exercised by `TestConfigBuilder` in many places. In an upcoming PR (Sharding 3) the new API is going to be further utilized.

Also, just moved the `identify_cert` function from the Server into the Config. This allows the Server to just depend on the respective Sharded or Ring Config to find the appropriate certificate depending on the case.

I made these changes before we moved `ConnectionFlavor` (CF) deeper into the `net` module (as part of PR #1348). I'm leaving `NetworkConfig` using CF here to show how it "looks like". If we want to avoid depending on CF, I will switch to use `TransportIdentity`. But maybe we want to pull CF back out into the `sharding` module.

**Server** was generalized to be able to work on Sharded and Ring cases. For the most part this only meant separating the client identification layer. Also, Server, unlike Client, doesn't change its API depending on the TransportRestriction. Instead we will initialize a different Axum router when creating. In here I'm creating the `ring_router`.

### DIFF
--- a/ipa-core/src/config.rs
+++ b/ipa-core/src/config.rs
@@ -1,13 +1,13 @@
 use std::{
-    array,
     borrow::{Borrow, Cow},
+    collections::HashMap,
     fmt::{Debug, Formatter},
-    iter::Zip,
+    iter::zip,
     path::PathBuf,
-    slice,
     time::Duration,
 };
 
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
 use hyper::{http::uri::Scheme, Uri};
 use hyper_util::client::legacy::Builder;
 use rustls_pemfile::Item;
@@ -22,6 +22,8 @@ use crate::{
         Deserializable as _, IpaPrivateKey, IpaPublicKey, KeyRegistry, PrivateKeyOnly,
         PublicKeyOnly, Serializable as _,
     },
+    net::{ConnectionFlavor, Helper, Shard},
+    sharding::ShardIndex,
 };
 
 pub type OwnedCertificate = CertificateDer<'static>;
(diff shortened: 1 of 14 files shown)


ipa-core/src/net/server/mod.rs
```diff
@@ -467,9 +460,8 @@ impl<B, S: Service<Request<B>, Response = Response>> Service<Request<B>>
     }
 
     fn call(&mut self, mut req: Request<B>) -> Self::Future {
-        if let Some(header_value) = req.headers().get(&HTTP_HELPER_ID_HEADER) {
-            let id_result = ClientIdentity::try_from(header_value.clone())
-                .map_err(|e| Error::InvalidHeader(format!("{HTTP_HELPER_ID_HEADER}: {e}").into()));
+        if let Some(header_value) = req.headers().get(F::identity_header()) {
+            let id_result = ClientIdentity::<F::Identity>::try_from(header_value.clone());
```
akoshelev, line 464: why do you need to clone it?

ipa-core/src/net/server/mod.rs
```diff
@@ -76,34 +76,38 @@ impl TracingSpanMaker for () {
...
         config: ServerConfig,
-        network_config: NetworkConfig,
+        network_config: NetworkConfig<Helper>,
     ) -> Self {
+        let router = handlers::ring_router(transport);
         MpcHelperServer {
-            transport,
             config,
             network_config,
+            router,
```
akoshelev, line 101: It does not seem obvious to me why we had to to change the approach and call `self.router().clone().op()` instead of `self.router()`? can you please elaborate on that?

ipa-core/src/config.rs
```diff
@@ -37,23 +39,123 @@ pub enum Error {
...
+        }
+        Self {
+            peers,
+            client,
+            identities,
+        }
+    }
+
+    #[must_use]
+    pub fn peers_map(&self) -> HashMap<ShardIndex, &PeerConfig> {
```
akoshelev, line 136: Can't it be a vec indexed by shard index?

ipa-core/src/config.rs
```diff
@@ -37,23 +39,123 @@ pub enum Error {
...
+pub struct NetworkConfig<R: ConnectionFlavor = Helper> {
+    peers: Vec<PeerConfig>,
 
     /// HTTP client configuration.
     #[serde(default)]
     pub client: ClientConfig,
+
+    /// The identities of the index-matching peers. Separating this from [`Self::peers`](field) so
+    /// that parsing is easy to implement.
+    #[serde(skip)]
```
akoshelev, line 58: how does it get populated, if not from deserializing from toml?

ipa-core/src/config.rs
```diff
@@ -37,23 +39,123 @@ pub enum Error {
...
+    }
+}
+
+impl NetworkConfig<Shard> {
+    #[must_use]
+    pub fn new_shards(peers: Vec<PeerConfig>, client: ClientConfig) -> Self {
+        let mut identities = Vec::with_capacity(peers.len());
+        for (i, _p) in zip(0u32.., peers.iter()) {
+            identities.push(ShardIndex(i));
+        }
```
akoshelev, line 127:
```suggestion
let identities = (0..peers.len()).map(ShardIndex::from).collect::<Vec<_>>()
```

ipa-core/src/config.rs
```diff
@@ -37,23 +39,123 @@ pub enum Error {
...
+        for (ix, p) in zip(self.identities.iter(), self.peers.iter()) {
+            indexed_peers.insert(*ix, p);
+        }
+        indexed_peers
+    }
 }
 
-impl NetworkConfig {
+impl NetworkConfig<Helper> {
+    /// Creates a new ring configuration.
```
akoshelev, line 146: I am not sure whether it is worth introducing the term ring that we haven't used before and from previous PR conversations we seem to be settling on word MPC

ipa-core/src/net/server/mod.rs
```diff
@@ -315,73 +315,54 @@ async fn rustls_config(
 // at some inconvenience (e.g. `MaybeExtensionExt`), we avoid using `Option` within the extension,
 // to avoid possible confusion about how many times the return from `req.extensions().get()` must be
 // unwrapped to ensure valid authentication.
-#[derive(Clone, Copy, Debug)]
-struct ClientIdentity(pub HelperIdentity);
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct ClientIdentity<I: TransportIdentity>(pub I);
```
akoshelev, line 319: The documentation can be improved to mention that it also works for shards

akoshelev (approved):
lgtm, merge it after addressing feedback

ipa-core/src/net/server/mod.rs
cberkhoff: Ok, I can use references as well (`TryFrom<&HeaderValue>`)

ipa-core/src/net/server/mod.rs
cberkhoff: If we set the router during the init we can re-use the same logic for '<Shard>' and '<Helper>' variants. 

The rest of the code, in particular `start_on`, can just assume that some Router has been set and work the same way for Shard to Shard or Helper to Helper.

ipa-core/src/config.rs
cberkhoff: I didn't wanted to change the file "format".
We had this convention where the position in the file gives the identity. With that field I'm following through with that idea an extending it to Shards in a way that allowed me to re-use as much code as possible.

If you check `from_toml_str` you will see I'm setting that field.

ipa-core/src/config.rs
cberkhoff: To me, MPC implies all the shards together. Do you think that the term "MPC" conveys this config is for 3 hosts ?

PR #1361, opened 2024-10-19
DRAFT: streaming uniqueness and decrypt

eriktaubeneck:
@akoshelev this is roughly the shape of how this could work with streaming, though this still doesn't compile. the issue currently is getting the right type coming out of the new `verify_uniqueness` function (both conforming to it in the function, and then handling it after.)


eriktaubeneck:
addressed in #1362

PR #1362, opened 2024-10-20
Streaming tag resharding and match key collection in Hybrid 

akoshelev:
In https://github.com/private-attribution/ipa/pull/1358 it was mentioned that waiting until all AAD tags and match keys have been collected before starting resharding process adds latency that is unnecessary. We can start resharding process right when we received the first tag and do everything in parallel.

This change does that by leveraging newly added `reshard_try_stream` and a few helper structs and functions to make it ergonomic to use

### DIFF
--- a/ipa-core/src/protocol/context/mod.rs
+++ b/ipa-core/src/protocol/context/mod.rs
@@ -11,12 +11,12 @@ pub mod upgrade;
 mod batcher;
 pub mod validator;
 
-use std::{collections::HashMap, iter, num::NonZeroUsize, pin::pin};
+use std::{collections::HashMap, num::NonZeroUsize, pin::pin};
 
 use async_trait::async_trait;
 pub use dzkp_malicious::DZKPUpgraded as DZKPUpgradedMaliciousContext;
 pub use dzkp_semi_honest::DZKPUpgraded as DZKPUpgradedSemiHonestContext;
-use futures::{stream, Stream, StreamExt};
+use futures::{stream, Stream, StreamExt, TryStreamExt};
 use ipa_step::{Step, StepNarrow};
 pub use malicious::MaliciousProtocolSteps;
 use prss::{InstrumentedIndexedSharedRandomness, InstrumentedSequentialSharedRandomness};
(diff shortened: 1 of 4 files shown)


ipa-core/src/protocol/context/mod.rs
```diff
@@ -519,21 +509,65 @@ where
...
+    input: L,
+    shard_picker: S,
+) -> Result<Vec<K>, crate::error::Error>
+where
+    L: ExactSizeStream<Item = K>,
+    S: Fn(C, RecordId, &K) -> ShardIndex,
+    K: Message + Clone,
+    C: ShardedContext,
+{
+    reshard_try_stream(ctx, input.map(Ok), shard_picker).await
```
eriktaubeneck, line 560: is awaiting here going to cause this to block until the whole stream is processed? wouldn't we want to await higher up the stack?

ipa-core/src/protocol/context/mod.rs
```diff
@@ -917,6 +956,145 @@ mod tests {
...
+                .into_iter()
+                .flat_map(|v| v.reconstruct())
+                .collect::<Vec<_>>();
+
+            assert_eq!(input, r);
+        });
+    }
+
+    #[test]
+    fn reshard_try_stream_less_items_than_expected() {
```
eriktaubeneck, line 984: what happens in the inverse? e.g., stream has more items than expected?

ipa-core/src/query/runner/reshard_tag.rs
```diff
@@ -0,0 +1,149 @@
...
+        context::{reshard_try_stream, ShardedContext},
+        RecordId,
+    },
+    sharding::ShardIndex,
+};
+
+type DataWithTag<D, A> = Result<(D, A), Error>;
+
+/// Helper function to work with inputs to hybrid queries. Each encryption needs
+/// to be checked for uniqueness and we use AAD tag for that. While match keys are
```
eriktaubeneck, line 22: nit: it's not just match keys, but the full decrypted report

eriktaubeneck (commented):
looks like there are still some compiler errors, but overall this approach looks promising! a few small nits

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -61,35 +58,24 @@ where
...
-                        acc.1.push(result.1);
-                        Ok(acc)
-                    },
-                )
-                .await?;
-
-        let resharded_tags = reshard_iter(
+        let stream = LengthDelimitedStream::<EncryptedHybridReport, _>::new(input_stream)
+            .map_err(Into::<Error>::into)
+            .map_ok(|enc_reports| {
```
eriktaubeneck, line 63: one of the big challenges I've had thus far is how to reason about `enc_reports` and how it could interact with reshard. but collection all of these into a stream solves that and makes it much easier to deal with.

I wonder if we should consider hiding some of that complexity "under the hood" in `LengthDelimitedStream` and have it operate like a "normal" stream that where `.next()` gives an individual item, and not a Vec<item>. not in scope for this PR, but curious of others opinions here.

ipa-core/src/protocol/context/mod.rs
akoshelev: nothing great, but definitely deserves a test to demonstrate that

eriktaubeneck (approved):

PR #1364, opened 2024-10-21
[Sharding 3] ConnectionFlavor-aware HttpTransport

cberkhoff:
This is the third of a series of pull requests (PR) to enable sharding on IPA. See #1360 

Biggest changes are concentrated in `net/transport.rs`. 

`HttpTransport` has now 2 "sub-classes"; Mpc and Shard, each implemented as a composition. E.g.
```
#[derive(Clone)]
pub struct MpcHttpTransport {
    inner_transport: Arc<HttpTransport<Helper>>,
}
```
Having `MpcHttpTransport` own the inner Arc simplified the interactions with these HttpTransports since they know how to clone themselves. Shard and MPC HTTP Transports are very similar, hence most of the functionality has been factored inside `HttpTransport`.

`bin/helper.rs` and `net/test.rs` have a few temporary changes required for the code to compile. These will go away as I continue pushing the changes required for sharding.

### DIFF
--- a/ipa-core/src/net/server/mod.rs
+++ b/ipa-core/src/net/server/mod.rs
@@ -39,6 +39,10 @@ use tower::{layer::layer_fn, Service};
 use tower_http::trace::TraceLayer;
 use tracing::{error, Span};
 
+use super::{
+    transport::{MpcHttpTransport, ShardHttpTransport},
+    Shard,
+};
 use crate::{
     config::{
         NetworkConfig, OwnedCertificate, OwnedPrivateKey, PeerConfig, ServerConfig, TlsConfig,
@@ -48,7 +52,7 @@ use crate::{
     helpers::TransportIdentity,
     net::{
         parse_certificate_and_private_key_bytes, server::config::HttpServerConfig,
-        ConnectionFlavor, Error, Helper, HttpTransport, CRYPTO_PROVIDER,
+        ConnectionFlavor, Error, Helper, CRYPTO_PROVIDER,
     },
     sync::Arc,
     telemetry::metrics::{web::RequestProtocolVersion, REQUESTS_RECEIVED},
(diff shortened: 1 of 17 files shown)


ipa-core/src/net/transport.rs
```diff
@@ -152,30 +187,88 @@ impl HttpTransport {
...
+        network_config: NetworkConfig<Helper>,
+        clients: [MpcHelperClient; 3],
+        handler: Option<HandlerRef<HelperIdentity>>,
+    ) -> (Self, MpcHelperServer<Helper>) {
+        let transport = Self::new_internal(http_runtime, identity, clients, handler);
+        let server = MpcHelperServer::new_mpc(&transport, server_config, network_config);
+        (transport, server)
+    }
+
+    fn new_internal(
```
akoshelev, line 207: shouldn't we inline it?

ipa-core/src/net/transport.rs
```diff
@@ -152,30 +187,88 @@ impl HttpTransport {
...
+        http_runtime: IpaRuntime,
+        identity: HelperIdentity,
+        clients: [MpcHelperClient; 3],
+        handler: Option<HandlerRef<HelperIdentity>>,
+    ) -> Self {
+        let mut id_clients = HashMap::new();
```
akoshelev, line 213: nit: `HashMap::from_iter` is nicer for this use case. Also you may consider using arrays here - default hasher in Rust is pretty heavyweight. It should work for both shard index and helper identity as these are just numbers and they are not sparse

ipa-core/src/net/server/mod.rs
```diff
@@ -103,6 +108,21 @@ impl MpcHelperServer<Helper> {
...
 }
 
+impl MpcHelperServer<Shard> {
+    #[must_use]
+    pub fn new_shards(
+        _transport: &ShardHttpTransport,
+        config: ServerConfig,
+        network_config: NetworkConfig<Shard>,
+    ) -> Self {
+        MpcHelperServer {
```
akoshelev, line 118: This seems incomplete, is it temporary similar to the TODO above? 

```
// TODO: Following is just temporary until Shard Transport is actually used.
```

akoshelev (approved):

ipa-core/src/net/server/mod.rs
cberkhoff: Correct. This will be filled once I push the changes for the Shard routes.

ipa-core/src/net/transport.rs
cberkhoff: I would just leave that to the compiler...  This code-path is hardly on the hot path.

akoshelev: it is not a performance concern, but a question about readability. It seems that `new_internal` is not used anywhere except `new` and it is not doing anything sophisticated

cberkhoff: Ahh ok. This is the way this was organized before but I agree it makes little sense to keep it separate.

PR #1369, opened 2024-10-24
[Sharding 4] Base testing support for sharding

cberkhoff:
This is the fourth of a series of pull requests (PR) to enable sharding on IPA. See #1364 

Biggest change here is in `net/test.rs` module, which was enhanced to create Sharded infrastructure. This will become more useful for the sharded shuffle test, but releasing early allows reviewers to check for the intended architecture and its usage.


ipa-core/src/net/test.rs
```diff
@@ -10,33 +10,127 @@
...
+/// Simple struct to keep default port configuration organized.
+pub struct Ports<C> {
+    ring: C,
+    shards: C,
+}
+
+/// A **single** ring with 3 hosts, each with a ring and sharding port.
+pub const DEFAULT_TEST_PORTS: Ports<[u16; 3]> = Ports {
+    ring: [3000, 3001, 3002],
+    shards: [6000, 6001, 6002],
```
akoshelev, line 44: how many shard does this setup support? it is not clear to me why we need sharding ports open on a single MPC ring and any other configuration with more than one shard per helper cannot be supported by this configuration it seems

ipa-core/src/net/test.rs
```diff
@@ -10,33 +10,127 @@
...
+}
+
+/// Uber container for test configuration. Provides access to a vec of MPC rings and 3 sharding
+/// networks (one for each Helper)
 pub struct TestConfig {
     pub disable_https: bool,
-    pub network: NetworkConfig<Helper>,
-    pub servers: [ServerConfig; 3],
-    pub sockets: Option<[TcpListener; 3]>,
+    pub rings: Vec<TestNetwork<Helper>>,
```
akoshelev, line 112: nit: I think we need consistent naming for variables here. `rings`/`shards` or something along these lines

ipa-core/src/net/test.rs
```diff
@@ -91,30 +185,40 @@ pub fn server_config_https(
     }
 }
 
-#[derive(Default)]
 pub struct TestConfigBuilder {
-    ports: Option<[u16; 3]>,
+    /// Can be empty meaning that free ports should be obtained from the operating system.
```
akoshelev, line 189: empty vector is a poor fit for this purpose. `Option<Vec<...>>` while more verbose, makes it much more clear that it is a different mode

ipa-core/src/net/test.rs
```diff
@@ -140,31 +251,75 @@ impl TestConfigBuilder {
...
-                url: format!("{scheme}://localhost:{}", ports[i])
-                    .parse()
-                    .unwrap(),
-                certificate: cert,
-                hpke_config: if self.disable_matchkey_encryption {
+        TestServers::new(configs, sockets)
+    }
+
+    /// Defines a test network with a single MPC ring (3 helpers).
+    fn build_ring_network(
```
akoshelev, line 286: After this change `TestConfig` looks a lot more complicated that it was before. I think it will benefit from some refactoring to split MPC and shard network initializations

ipa-core/src/net/test.rs
```diff
@@ -173,30 +328,76 @@ impl TestConfigBuilder {
...
+        // The following will accumulate the shards as we create the rings.
+        let mut sharding_networks_shards = vec![
+            TestServers::default(),
+            TestServers::default(),
+            TestServers::default(),
+        ];
+        let rings: Vec<TestNetwork<Helper>> = (0..self.sharding_factor)
+            .map(|s| {
+                let ring_ports = self.ports_by_ring.get(s).map(|p| p.ring.clone());
+                let shards_ports = self.ports_by_ring.get(s).map(|p| p.shards.clone());
```
akoshelev, line 360:
```suggestion
        let r = zip(ring_ports, shard_ports).map(|(r_port, s_port)| { ... })
```

ipa-core/src/net/test.rs
```diff
@@ -318,17 +508,40 @@ impl TestServerBuilder {
...
 }
 
-fn get_test_certificate_and_key(id: HelperIdentity) -> (&'static [u8], &'static [u8]) {
+pub struct ClientIdentities {
+    pub helper: ClientIdentity<Helper>,
+    pub shard: ClientIdentity<Shard>,
+}
+
+#[must_use]
+pub fn create_ids(disable_https: bool, id: HelperIdentity, ix: ShardIndex) -> ClientIdentities {
```
akoshelev, line 517: should it be a constructor on `ClientIdentities`?

ipa-core/src/net/test.rs
```diff
@@ -318,17 +508,40 @@ impl TestServerBuilder {
...
     (TEST_CERTS[id], TEST_KEYS[id])
 }
 
+/// Creating a cert client identity. Using the same certificate for both shard and mpc.
 #[must_use]
-pub fn get_test_identity(id: HelperIdentity) -> ClientIdentity {
-    let (mut certificate, mut private_key) = get_test_certificate_and_key(id);
-    ClientIdentity::from_pkcs8(&mut certificate, &mut private_key).unwrap()
+pub fn get_client_test_identity(id: HelperIdentity, ix: ShardIndex) -> ClientIdentities {
+    let pos = ix.as_index() * 3 + id.as_index();
```
akoshelev, line 535: that deserves a comment and a common function to compute the absolute offset as I believe it is used in more than one place

ipa-core/src/net/test.rs
```diff
@@ -404,3 +686,48 @@ const TEST_HPKE_PUBLIC_KEY: &str = "\
...
+#[cfg(all(test, unit_test))]
+mod tests {
+    use super::TestConfigBuilder;
+    use crate::{helpers::HelperIdentity, net::test::Ports};
+
+    /// This simple test makes sure that testing networks are created properly.
+    /// The network itself won't be excersized as that's tested elsewhere.
+    #[test]
+    fn create_4_shard_http_network() {
+        let ports: Vec<Ports<Vec<u16>>> = vec![
```
akoshelev, line 699: do you want to test that overlapping ports are rejected?

akoshelev (changes requested):

ipa-core/src/net/test.rs
```diff
@@ -91,30 +185,40 @@ pub fn server_config_https(
...
             use_http1: false,
             disable_matchkey_encryption: false,
         }
     }
+}
 
+impl TestConfigBuilder {
     #[must_use]
-    pub fn with_open_ports() -> Self {
+    pub fn with_http_and_default_test_ports() -> Self {
```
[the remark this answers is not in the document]
cberkhoff: Requesting ports from the kernel is part of the default method.

In the past we used to have both, with_open_ports and default, which did the same thing. I thought less code is better.

ipa-core/src/net/test.rs
```diff
@@ -91,30 +185,40 @@ pub fn server_config_https(
...
 
-#[derive(Default)]
 pub struct TestConfigBuilder {
-    ports: Option<[u16; 3]>,
+    /// Can be empty meaning that free ports should be obtained from the operating system.
+    /// One ring per shard in a helper (sharding factor). For each ring we need 3 shard and
+    /// 3 mpc ports.
+    ports_by_ring: Vec<Ports<Vec<u16>>>,
+    /// Describes the number of shards per helper
+    sharding_factor: usize,
```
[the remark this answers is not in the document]
cberkhoff: That was intended because you're getting 3x hosts for the factor.

If you still think shard_count is easier I can do that, but this factor affects all 3 helpers.

ipa-core/src/net/test.rs
cberkhoff: 1) how many shard does this setup support? 1 shard. It's an "unsharded" scenario. At this point I'm only adding facilities to the code. Exercising more shards will come later. This to make PRs simpler.

2) "it is not clear to me why we need sharding ports open on a single MPC ring"
We spoke about this sometime back. The reason is simplicity. Having shards open ports all the time allows for the code to not have to bother for the special case of one shard.

3) "any other configuration with more than one shard per helper cannot be supported by this configuration it seems"
`DEFAULT_TEST_PORTS` is just a default for a single ring MPC. I was planning on adding one for sharded at some point later.

ipa-core/src/net/test.rs
cberkhoff: Sure, I can add that

ipa-core/src/net/test.rs
cberkhoff: Ok

ipa-core/src/net/test.rs
cberkhoff: Sure. The conditional on disable_https had me on the fence

ipa-core/src/net/test.rs
cberkhoff: Yeah, Test config holds now the entire sharded network. But on the other hand, having everything described on this single object makes it easy to operate on the network as a whole which is the use case (check the transport.rs test). Separating this into 2 seems like it will only add complexity on the caller side.

Can you be more explicit on how you want to split the network init?

ipa-core/src/net/test.rs
cberkhoff: Makes sense

ipa-core/src/net/test.rs
```diff
@@ -140,31 +251,75 @@ impl TestConfigBuilder {
...
+                .map(|_| Some(TcpListener::bind("localhost:0").unwrap()))
+                .collect();
+            sockets
+                .iter()
+                .map(|sock| sock.as_ref().unwrap().local_addr().unwrap().port())
+                .collect()
         });
-        let (scheme, certs) = if self.disable_https {
-            ("http", [None, None, None])
+        // Ports will always be defined after this. Socks only if there were no ports set.
```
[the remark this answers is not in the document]
cberkhoff: Not sure to what part you're referring...

But this function creates 3 servers.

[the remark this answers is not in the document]
cberkhoff: After line 270 all the port situation will be set. Either you gave it as param or we asked the os.

Not sure to what part you're referring when you say "Why we use all ports available to build servers"...  But this function creates 3 servers.

ipa-core/src/net/test.rs
akoshelev: you don't have to make callers suffer. Providing a single entry point for them to construct the config is reasonable. Doing everything inside one struct after that is not necessary - power through composition

ipa-core/src/net/test.rs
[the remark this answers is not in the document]
akoshelev: then `/// Describes the number of shards per helper`  confuses me even more

ipa-core/src/net/test.rs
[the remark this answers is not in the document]
akoshelev: The confusion comes from either using `optional_ports` or synthesizing `ports` inside 

`optional_ports: Option<Vec<u16>>` which can be arbitrary size according to this function (it does not enforce size to be 3, if it should, then it would be better to take an array). So `ports` may have variable size

ipa-core/src/net/test.rs
akoshelev: I am not sure whether simplicity savings here are justified, especially for in-memory tests, as it invites us to setup listeners and other heavy stuff for these ports that otherwise won't be used and complexity is contained to this module.

If you think it will take too long to fix, I am ok proceeding with this but at least we need a comment explaining what is going on here

ipa-core/src/net/test.rs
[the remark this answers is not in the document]
cberkhoff: Why? it says per helper aka 3x

ipa-core/src/net/test.rs
cberkhoff: there are cases where ports_by_ring won't be set though.

ipa-core/src/net/test.rs
```diff
@@ -9,46 +9,171 @@
...
+    shards: [6000, 6001, 6002],
+};
+
+/// Configuration of a server that can be reached via socket or port.
+pub struct AddressableTestServer {
+    pub id: ShardedHelperIdentity,
+    /// Contains the ports
+    pub config: ServerConfig,
+    /// Sockets are created if no port was specified.
+    pub socket: Option<TcpListener>,
```
akoshelev, line 54: as much as I like the idea of holding onto TcpListener until the moment when server is spawned - minimizes the chances of Kernel re-using that port in between - I don't think the approach of storing `TcpListener` in the configuration struct is appropriate here. `TcpListener` does not describe the configuration, it actively uses a system resource (socket)

Maybe we can get away with using file descriptors, but honestly I wouldn't bother as I've never seen an issue with bind/release.

ipa-core/src/net/test.rs
```diff
@@ -140,63 +353,42 @@ impl TestConfigBuilder {
...
+        }
+    }
+
+    /// Get all the shard ports in a helper.
+    fn get_ports_for_helper_identity(&self, id: HelperIdentity) -> Vec<Option<u16>> {
+        if let Some(ports_by_ring) = &self.ports_by_ring {
+            ports_by_ring
+                .iter()
+                .map(|r| r.shards[id.as_index()])
+                .map(Some)
```
akoshelev, line 379:
```suggestion
                .map(|r| Some(r.shards[id.as_index()]))
```

ipa-core/src/net/test.rs
```diff
@@ -140,63 +353,42 @@ impl TestConfigBuilder {
...
-            ("http", [None, None, None])
+    /// Creates a HTTP1 or HTTP2 client config.
+    pub fn create_client_config(&self) -> ClientConfig {
+        self.use_http1
+            .then(ClientConfig::use_http1)
+            .unwrap_or_default()
+    }
+
+    /// Get all the MPC ports in a ring specified by the shard index.
+    fn get_ports_for_shard_index(&self, ix: ShardIndex) -> Vec<Option<u16>> {
```
akoshelev, line 364: from the ergonomic perspective, a better type to return would be `Option<Vec<u16>>` here

ipa-core/src/net/test.rs
```diff
@@ -9,46 +9,171 @@
 
 #![allow(clippy::missing_panics_doc)]
 use std::{
-    array,
+    collections::HashSet,
```
akoshelev, line 12: Consider splitting this file into multiple - it grew too big and Github now hides all the changes for it by default

akoshelev (approved):
I think we shouldn't store a `TcpListener` inside a configuration struct, so I'd like this to be fixed before pushing it, but other than that lgtm

ipa-core/src/net/test.rs
cberkhoff: In general, do you prefer larger blocks of code inside a map function instead of more map calls?

ipa-core/src/net/test.rs
cberkhoff: I'm thinking on separating the TestServer stuff

ipa-core/src/net/test.rs
cberkhoff: Ahh, I think that might be problematic with many tests running at the same time.

akoshelev: Kernel is good not releasing the port right away

ipa-core/src/net/test.rs
akoshelev: it is always a balance, cluttered map calls are bad for readability, more map calls result in more complex types used and more work for compiler to unwind them

PR #1372, opened 2024-10-24
Save intermediates during sharded shuffle

andyleiserson:
First step towards malicious sharded shuffle. This saves the intermediates from the sharded shuffle, like #1249 did for the unsharded shuffle.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
@@ -14,6 +14,7 @@ use crate::{
         boolean_array::{BooleanArray, BA112, BA144, BA64, BA96},
         ArrayAccess,
     },
+    helpers::Role,
     protocol::{
         context::{Context, MaliciousContext, SemiHonestContext},
         ipa_prf::{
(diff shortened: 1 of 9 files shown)


ipa-core/src/protocol/ipa_prf/shuffle/mod.rs
```diff
@@ -34,6 +34,16 @@ pub mod malicious;
 mod sharded;
 pub(crate) mod step;
 
+/// This struct stores some intermediate messages during the shuffle.
+/// In a maliciously secure shuffle,
+/// these messages need to be checked for consistency across helpers.
+/// `H1` stores `x1`, `H2` stores `x2` and `H3` stores `y1` and `y2`.
+#[derive(Debug, Clone)]
+struct IntermediateShuffleMessages<S> {
```
akoshelev, line 42: any thoughts on making it an enum instead? seems that it can prevent some dumb errors where H1 can actually assign both

ipa-core/src/protocol/ipa_prf/shuffle/sharded.rs
```diff
@@ -263,30 +265,33 @@ impl<V: SharedValue + FromRandomU128> Shuffleable for AdditiveShare<V> {
...
+) -> Result<(Vec<S>, IntermediateShuffleMessages<S::Share>), crate::error::Error>
 where
     I: IntoIterator<Item = S>,
     I::IntoIter: Send + ExactSizeIterator,
     C: ShardedContext,
     S: Shuffleable,
 {
     // Generate X_1 = perm_12(left ⊕ right ⊕ z_12).
-    let x1 = ctx
+    let x1: Vec<S::Share> = ctx
```
akoshelev, line 279: 👍

ipa-core/src/protocol/ipa_prf/shuffle/sharded.rs
```diff
@@ -441,14 +480,25 @@ where
...
+        },
+    ))
 }
 
 /// Entry point to execute sharded shuffle.
 /// ## Errors
 /// Failure to communicate over the network, either to other MPC helpers, and/or to other shards
 /// will generate a shuffle error.
-pub async fn shuffle<I, S, C>(ctx: C, shares: I) -> Result<Vec<S>, crate::error::Error>
+pub async fn shuffle<I, S, C>(
```
akoshelev, line 498: nit: I would probably note here that this function must not be called directly (maybe we should even hide it)

ipa-core/src/secret_sharing/mod.rs
```diff
@@ -62,7 +62,7 @@ pub trait Block: Sized + Copy + Debug {
     type Size: ArrayLength;
 }
 
-pub trait Sendable: Send + Debug + Serializable + 'static {}
+pub trait Sendable: Send + Sync + Debug + Serializable + 'static {}
```
akoshelev, line 65: curious what drives this to be `Sync` as well?

akoshelev (approved):

ipa-core/src/secret_sharing/mod.rs
andyleiserson: It is a consequence of the change to `mask_and_shuffle` to accept borrowed input. For the input `&T` to `mask_and_shuffle` to be `Send`, `T` needs to be `Sync`.

PR #1374, opened 2024-10-25
Metrics ergonomic and unit test improvements

akoshelev:
These changes are required to make the metric engine work in IPA. See commit messages for details

### DIFF
--- a/ipa-metrics/src/key.rs
+++ b/ipa-metrics/src/key.rs
@@ -203,7 +203,7 @@ impl Hash for OwnedName {
 
 #[cfg(test)]
 pub fn compute_hash<V: Hash>(value: V) -> u64 {
-    let mut hasher = std::hash::DefaultHasher::default();
+    let mut hasher = crate::label_hasher();
     value.hash(&mut hasher);
 
     hasher.finish()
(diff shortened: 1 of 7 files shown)


ipa-metrics/src/label.rs
```diff
@@ -3,8 +3,19 @@ use std::{
     hash::{Hash, Hasher},
 };
 
+use rustc_hash::FxHasher;
+
 pub const MAX_LABELS: usize = 5;
 
+/// Provides a fast, non-collision resistant implementation of [`Hasher`]
+/// for label values. T
```
andyleiserson, line 11:
```suggestion
/// for label values.
```

ipa-metrics/src/key.rs
```diff
@@ -114,8 +114,8 @@ impl<'lv, const LABELS: usize> Name<'lv, LABELS> {
 /// This is the key inside metric stores which are simple hashmaps.
 #[derive(Debug, Clone, Eq)]
 pub struct OwnedName {
-    key: &'static str,
-    labels: [Option<OwnedLabel>; 5],
+    pub key: &'static str,
+    pub labels: [Option<OwnedLabel>; 5],
```
andyleiserson, line 118: I haven't dug in to where this gets used, but exposing the padded array as a public interface seems like it has some potential for confusion.

andyleiserson (approved):

ipa-metrics/src/key.rs
akoshelev: you're right and I don't think it is even necessary given that `labels()` method exists

PR #1375, opened 2024-10-25
add hybrid_protocol function, unimplemented

eriktaubeneck:
This creates a new function that will take in the indistinguishable reports and compute the hybrid protocol. This PR is just plumming.

It includes a bunch of trait bounds that we also have for the OPRF protocol. Given that we will be using most of the same sub-protocols, I think we'll need them all, but certainly open to feedback here if we think we should add them at a different point.

### DIFF
--- a/ipa-core/src/protocol/hybrid/mod.rs
+++ b/ipa-core/src/protocol/hybrid/mod.rs
@@ -39,20 +39,19 @@ impl BreakdownKey<256> for BA8 {}
 /// impressions or conversion. The protocol joins these based on their matchkeys,
 /// sums the values from conversions grouped by the breakdown key on impressions.
 /// To accomplish this, hte protocol performs the follwoing steps
-///  1. Converts secret-sharings of boolean arrays to secret-sharings of elliptic curve points
-///  2. Generates a random number of "dummy records" (needed to mask the information that will
-///     be revealed in step 4, and thereby provide a differential privacy guarantee on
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/hybrid/mod.rs
```diff
@@ -1 +1,76 @@
...
+    helpers::query::DpMechanism,
+    protocol::{
+        context::{ShardedContext, UpgradableContext},
+        ipa_prf::{oprf_padding::PaddingParameters, shuffle::Shuffle},
+    },
+    report::hybrid::IndistinguishableHybridReport,
+    secret_sharing::replicated::semi_honest::AdditiveShare as Replicated,
+};
+
+// In theory, we could support (runtime-configured breakdown count) ≤ (compile-time breakdown count)
```
akoshelev, line 18: do you need to copy-paste this or OPRF definitions can work here too?

eriktaubeneck: yea, I could certainly just import these. my intention was to make it easier to delete the other file completely down the road, but that's potentially a premature consideration.

akoshelev: no that's fine. Can you just mark these as duplicates from OPRF so we won't forget about this decision?

akoshelev (approved):

Issue #1370, opened 2024-10-24
Optimize performance of sequential DZKPs


andyleiserson:
The change in #1350 to use record IDs to index proof batches caused a modest performance regression.

Comparing https://draft-mpc.vercel.app/query/view/yummy-full2024-10-18T2044 and https://draft-mpc.vercel.app/query/view/ugly-fuel2024-10-18T2108, the difference appears to be in the share conversion step. I suspect it has to do with the way that OrderingSender sequences communications (record n+1 is not accepted into the channel until after record n) somehow reducing the amount of validation work that can be done in parallel, but I haven't completely worked out how.

_Originally posted by @andyleiserson in https://github.com/private-attribution/ipa/issues/1350#issuecomment-2427135447_

PR #1376, opened 2024-10-25
Revise comment about chunk override for attribution

andyleiserson:
This PR originally removed the chunk override for attribution, but that was observed to impact performance, so now is just revising the comment to accurately describe the situation.


andyleiserson:
2M, 19:16: https://draft-mpc.vercel.app/query/view/plain-witch2024-10-25T2258
25M, 3:44:12: https://draft-mpc.vercel.app/query/view/elder-guide2024-10-26T0013

This is slower than the baseline in #1370, so I don't think we want to take the change, but I will revise the comment.

akoshelev (approved):

PR #1377, opened 2024-10-25
Notes about updating rust version

andyleiserson:


eriktaubeneck (approved):
great idea, thanks for adding this

Issue #1353, opened 2024-10-16
Integrate IPA metrics with open-source telemetry implementation


akoshelev:
## Overview

IPA currently does not emit a lot of telemetry but it will soon because we need to look closely at protocol performance at scale. This includes adding metrics and having a means to consume and aggregate them from running instances. This task needs to cover the following

* Pick an open source standard/engine to emit telemetry from IPA web server. `statsd`, open telemetry, Prometheus can be viable choices. We should strive for k8s compatibility and ease of integration with existing UIs to present them. A bonus point is integration with common clouds.
* Build integration by exposing an endpoint through which metrics can be consumed from a shard/helper. 
* Integrate with IPA metrics by building a means to ingest metrics from running queries and send them to the monitoring instance

PR #1278, opened 2024-09-17
[DRAFT] [DO NOT MERGE] [IGNORE] IPA metrics

akoshelev:
Using this PR to make some draft runs

PR #1379, opened 2024-10-26
Migrate IPA code to new metrics engine

akoshelev:
This is the final PR to upgrade IPA to use new metric engine. This change removes the old metric dependency and updates the code to use new metrics.

This approach was tested in #1278 and the only difference here is that we don't consume metrics yet. #1353 will take care of it

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -52,7 +52,7 @@ web-app = [
     "http-body",
     "http-body-util",
 ]
-test-fixture = ["weak-field"]
+test-fixture = ["weak-field", "ipa-metrics-tracing", "ipa-metrics/partitions"]
 # Include observability instruments that detect lack of progress inside MPC. If there is a bug that leads to helper
 # miscommunication, this feature helps to detect it. Turning it on has some cost.
 # If "shuttle" feature is enabled, turning this on has no effect.
@@ -86,6 +86,8 @@ ipa-prf = []
 relaxed-dp = []
 
 [dependencies]
+ipa-metrics = { path = "../ipa-metrics", features = [] }
+ipa-metrics-tracing = { optional = true, path = "../ipa-metrics-tracing" }
 ipa-step = { version = "*", path = "../ipa-step" }
 ipa-step-derive = { version = "*", path = "../ipa-step-derive" }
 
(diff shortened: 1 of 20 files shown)


ipa-core/Cargo.toml
```diff
@@ -86,6 +86,8 @@ ipa-prf = []
 relaxed-dp = []
 
 [dependencies]
+ipa-metrics = { path = "../ipa-metrics", features = [] }
```
andyleiserson, line 89:
```suggestion
ipa-metrics = { path = "../ipa-metrics" }
```
Unless this is somehow needed to remove the partitions feature.

ipa-core/src/telemetry/mod.rs
```diff
@@ -33,84 +31,20 @@ pub mod metrics {
             }
         }
 
-        impl From<RequestProtocolVersion> for KeyName {
+        impl From<RequestProtocolVersion> for &'static str {
```
andyleiserson, line 34: Maybe this should be an `as_str` method?

ipa-core/src/sharding.rs
```diff
@@ -84,6 +86,16 @@ impl<T> IndexMut<ShardIndex> for Vec<T> {
     }
 }
 
+impl LabelValue for ShardIndex {
```
andyleiserson, line 89: Does it makes sense to implement  `LabelValue` for `T: Into<u64>`? Given the restrictions on overlapping impls, that does seem like it could lead to problems down the line. On the other hand, since this isn't a published API, we can always switch back to the set of impls in this PR if the blanket impl becomes a problem.

andyleiserson (approved):

ipa-core/src/telemetry/mod.rs
akoshelev: its a little less ergonomic at the callsite, but not too bad

ipa-core/src/sharding.rs
akoshelev: yea, overlapping impls was the reason not to. I ran into this problem in an early implementation and it gets very annoying if you need to implement `LabelValue` a bit differently. Scraping out blanket implementation when you would need that change would probably be less efficient. 

Maybe I can provide a macro to implement `LabelValue` on those types....

andyleiserson: I'm not opposed to a macro, but I also don't think there's a clear justification for one here vs. just live with a little boilerplate.

akoshelev: yea, that is my thinking too - `LabelValue` does not add too much clutter, but provides explicitness

PR #1380, opened 2024-10-28
Seed management to improve reproducibility

### LINKED ISSUE #1321: Ensure TestWorld seed is propagated to all RNGs
I just realized that we probably want to use the same seed for rng and test world (it is supported via `TestWorldConfig` struct). That way we can make it reproducible if it ever fails.

This was originally raised in the context of tests that use `rng.gen()` to generate random input values, but we should consider all kinds of randomness, e.g., proptests.

_Originally posted by @akoshelev in https://github.com/private-attribution/ipa/pull/1316#discussion_r1777874860_

andyleiserson:
Changes to tests and test infrastructure to be able to reproduce test runs from a seed.

This just updates one example of each kind of test I could think of to demonstrate how to achieve reproducibility. There are still many other tests that would need to be updated to match.

Fixes #1321

### DIFF
--- a/ipa-core/src/lib.rs
+++ b/ipa-core/src/lib.rs
@@ -262,6 +262,8 @@ pub(crate) mod executor {
 pub(crate) mod test_executor {
     use std::future::Future;
 
+    use shuttle::rand::{rngs::ThreadRng, thread_rng};
+
     pub fn run<F, Fut>(f: F)
     where
         F: Fn() -> Fut + Send + Sync + 'static,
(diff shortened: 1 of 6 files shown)


ipa-core/src/protocol/ipa_prf/aggregation/mod.rs
```diff
@@ -503,37 +501,38 @@ pub mod tests {
...
-                let row: [u32; PROP_BUCKETS] = array::from_fn(|_| rng.gen_range(0..1 << tv_bits));
+            for row in &inputs {
                 for (exp, val) in expected.iter_mut().zip(row) {
                     *exp = min(*exp + val, (1 << PropHistogramValue::BITS) - 1);
                 }
-                row
-            })
-            .take(len)
-            .collect();
+            }
```
andyleiserson, line 516: These changes are not essential to this PR, but it seemed clearer not to have two different seeds, and using proptest for input generation enables using its other capabilities like shrinking.

ipa-core/src/lib.rs
```diff
@@ -277,15 +279,31 @@ pub(crate) mod test_executor {
...
     use std::future::Future;
 
+    use rand::{rngs::StdRng, thread_rng, Rng, SeedableRng};
+
+    // These routines use `FnOnce` because it is easier than dealing with lifetimes of
+    // `&mut rng` borrows in futures. If there were a need to support multiple
+    // iterations (or to make the API use `Fn` to match the shuttle version), the
+    // simplest strategy might be to seed per-iteration RNGs from a primary RNG, like
+    // `TestWorld::rng`.
+
```
akoshelev, line 303:
```suggestion
```

ipa-core/src/test_fixture/world.rs
```diff
@@ -190,6 +200,13 @@ impl Default for TestWorld {
     }
 }
 
+impl TestWorld {
+    #[must_use]
+    pub fn with_seed(seed: u64) -> Self {
```
akoshelev, line 205: Default values for generic parameters is a blessing and a curse.

`impl TestWorld<NotSharded>` that exists on the line 266 is probably a better place for this fn.

akoshelev (approved):

akoshelev:
Thank you for making this change @andyleiserson, we are getting closer to fully reproducible tests with it

ipa-core/src/test_fixture/world.rs
andyleiserson: Is there a reason to implement `Default` only for `TestWorld<NotSharded>`?

akoshelev: backward compatibility was the only one - I didn't want to change a bunch of non-sharded tests when introducing the in-memory sharding infrastructure

PR #1381, opened 2024-10-28
Hybrid padding

eriktaubeneck:
This implements the `Paddable` trait for `IndistinguishableHybridReport` so that it can work with `apply_dp_padding`. It also implements `TestIndistinguishableHybridReport` and `Reconstruct` to support the tests.

Note that this is fairly repetitive to the OPRF implementation. If we wanted to keep that around, we'd want to make this more generic, but given our intention is to delete that, it's copied and modified to work directly with  `IndistinguishableHybridReport`.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
@@ -166,11 +166,11 @@ where
                                 Direction::Left => AdditiveShare::new(BA64::ZERO, dummy_mk),
                                 Direction::Right => AdditiveShare::new(dummy_mk, BA64::ZERO),
                             };
-                            let row = IndistinguishableHybridReport::new(
-                                match_key_shares,
-                                AdditiveShare::ZERO,
-                                AdditiveShare::ZERO,
-                            );
+                            let row = IndistinguishableHybridReport {
+                                match_key: match_key_shares,
+                                value: AdditiveShare::ZERO,
+                                breakdown_key: AdditiveShare::ZERO,
+                            };
                             padding_input_rows.extend(std::iter::once(row));
                         }
                     }
@@ -185,11 +185,11 @@ where
         total_number_of_fake_rows: u32,
     ) {
         for _ in 0..total_number_of_fake_rows as usize {
-            let row = IndistinguishableHybridReport::new(
-                AdditiveShare::ZERO,
-                AdditiveShare::ZERO,
-                AdditiveShare::ZERO,
-            );
+            let row = IndistinguishableHybridReport {
+                match_key: AdditiveShare::ZERO,
+                value: AdditiveShare::ZERO,
+                breakdown_key: AdditiveShare::ZERO,
+            };
 
             padding_input_rows.extend(std::iter::once(row));
         }
--- a/ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
@@ -492,6 +492,7 @@ mod tests {
             },
             RecordId,
         },
+        report::hybrid::IndistinguishableHybridReport,
         secret_sharing::replicated::semi_honest::AdditiveShare,
         test_fixture::{Reconstruct, Runner, TestWorld},
     };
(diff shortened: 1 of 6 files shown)


ipa-core/src/test_fixture/hybrid.rs
```diff
@@ -1,11 +1,52 @@
...
+}
+
+impl<BK, V> Reconstruct<TestIndistinguishableHybridReport>
+    for [&IndistinguishableHybridReport<BK, V>; 3]
+where
+    BK: BooleanArray + U128Conversions + IntoShares<Replicated<BK>>,
+    V: BooleanArray + U128Conversions + IntoShares<Replicated<V>>,
+{
+    fn reconstruct(&self) -> TestIndistinguishableHybridReport {
+        let [s0, s1, s2] = self;
```
akoshelev, line 30: you should be able to do something like this

```rust
self.each_ref().map(|v| v.match_key).reconstruct().as_u128()
```

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -130,6 +131,71 @@ pub trait Paddable {
...
+        let mut total_number_of_fake_rows = 0;
+        match padding_params.oprf_padding {
+            OPRFPadding::NoOPRFPadding => {}
+            OPRFPadding::Parameters {
+                oprf_epsilon,
+                oprf_delta,
+                matchkey_cardinality_cap,
+                oprf_padding_sensitivity,
+            } => {
+                let oprf_padding =
```
akoshelev, line 154: do we still plan to use `OPRFPaddingDp` or it will be replicated as well for hybrid?

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -130,6 +131,71 @@ pub trait Paddable {
         Self: Sized;
 }
 
+impl<BK, V> Paddable for IndistinguishableHybridReport<BK, V>
```
akoshelev, line 134: This trait could be made better...

akoshelev (approved):

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
eriktaubeneck: This gets somewhat confusing, because we use the term "OPRF" to name both the entire ipa v2 protocol as well as the pseudo random value that the match key is converted into.

In this instance, we are padding values so that the process of converting match keys into OPRF values is differentially private (same as previously), as opposed to padding values for aggregation or breakdown key reveal. In that sense, I think this naming still makes sense, and actually "HybridPadding" would be less clear (at least once the old protocol is purged.)

Happy for input on this - but that was my thinking. I can add a comment here to make that more clear as well.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -130,6 +131,71 @@ pub trait Paddable {
...
+where
+    BK: BooleanArray + U128Conversions,
+    V: BooleanArray,
+{
+    fn add_padding_items<VC: Extend<Self>, const B: usize>(
```
cberkhoff, line 139: Please add documentation

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -130,6 +131,71 @@ pub trait Paddable {
...
+        }
+        Ok(total_number_of_fake_rows)
+    }
+
+    fn add_zero_shares<VC: Extend<Self>>(
+        padding_input_rows: &mut VC,
+        total_number_of_fake_rows: u32,
+    ) {
+        for _ in 0..total_number_of_fake_rows as usize {
+            let row = IndistinguishableHybridReport {
```
cberkhoff, line 188: nit: use Default

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
eriktaubeneck: can you take a look and see if this impl is cleaner?

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -451,6 +518,31 @@ mod tests {
         Ok(input)
     }
 
+    pub async fn set_up_apply_dp_padding_pass_for_indistinguishable_reports<
```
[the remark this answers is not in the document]
eriktaubeneck: just the standard formatter. I don't think it makes sense to move to a separate module, all the padding functions live in this module.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
```diff
@@ -130,6 +133,64 @@ pub trait Paddable {
...
+            OPRFPadding::NoOPRFPadding => {}
+            OPRFPadding::Parameters {
+                oprf_epsilon,
+                oprf_delta,
+                matchkey_cardinality_cap,
+                oprf_padding_sensitivity,
+            } => {
+                let oprf_padding =
+                    OPRFPaddingDp::new(oprf_epsilon, oprf_delta, oprf_padding_sensitivity)?;
+                for cardinality in 1..=matchkey_cardinality_cap {
```
eriktaubeneck, line 158: I attempted to get this all into a single `padding_input_rows.extend` but I ran into a conflict between needing to use `move` so that cardinality didn't outlive the closure and a mutable `rng` not working with the `move`.

ipa-core/src/protocol/ipa_prf/oprf_padding/mod.rs
eriktaubeneck: Looking at this more closely, this is actually just poor naming of the existing `OPRFPaddingDp` struct, which itself isn't tied specifically to the OPRF, and is used for aggregation padding as well. I'll open a new PR to make that naming more generic. It still makes sense to me to use the `OPRFPadding` struct here since we're still using it to provide DP for the match_key to OPRF conversion.

PR #1382, opened 2024-10-28
Hybrid conversion

tyurek:
This d̶i̶f̶f̶ PR finishes the implementation of Hybrid Conversion reports and wraps them up into general Hybrid reports.
The main major change compared to OprfReports is that I wrote the encryption/decryption functions to take a HybridInfo struct as an argument, since there's a lot more info used for conversion reports now.

I'll say this is ready for review now and add some more coverage tomorrow.

Open question: In query/runner/hybrid.rs, I need to pass around HybridInfo structs to do decryption. I’m not totally sure the best way to structure that, since HybridInfo is currently written as an enum, but you would need both the Impression and Conversion types if you wanted to be able to decrypt both types of reports. I could write it to require both, but I’m not sure what makes the most sense given the overall system design. Currently I just comment out the Impression cases to test basic correctness, but that's just temporary until we figure this out.


ipa-core/src/report/hybrid.rs
```diff
@@ -308,6 +441,282 @@ where
...
+    pub fn key_id(&self) -> u8 {
+        match self {
+            EncryptedHybridGeneralReport::Impression(impression_report) => {
+                impression_report.key_id()
+            }
+            EncryptedHybridGeneralReport::Conversion(conversion_report) => {
+                conversion_report.key_id()
+            }
+        }
+    }
```
eriktaubeneck, line 617: do we actually need access to these on the `EncryptedHybridGeneralReport`? I think we should only need `decrypt` and `from_bytes` (which call into the enum items).

ipa-core/src/report/hybrid.rs
```diff
@@ -308,6 +441,282 @@ where
...
+            )),
+            value: Replicated::<V>::deserialize(GenericArray::from_slice(plaintext_btt)).map_err(
+                |e| InvalidHybridReportError::DeserializationError("trigger_value", e.into()),
+            )?,
+        })
+    }
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub enum EncryptedHybridGeneralReport<BK, V>
```
eriktaubeneck, line 544: to stay consistent with the decrypted `HybridReport` enum, this can probably just be called `EncryptedHybridReport`.

ipa-core/src/report/hybrid.rs
```diff
@@ -308,6 +441,282 @@ where
...
+                let conversion_report = EncryptedHybridConversionReport::<V>::from_bytes(bytes)?;
+                Ok(EncryptedHybridGeneralReport::Conversion(conversion_report))
+            }
+            _ => Err(InvalidHybridReportError::UnknownEventType(bytes[0])),
+        }
+    }
+
+    /// ## Errors
+    /// If decryption of the provided oprf report fails.
+    pub fn decrypt_from_oprf_report_bytes<P, TS>(
```
eriktaubeneck, line 643: we don't need to keep this long term. it would be great if we could get everything to work without this, but if we want to split that into another PR, that's fine.

eriktaubeneck:
> 1. This spec (especially for conversions) has a lot more information contained in Info structs than the old OPRF implementation. The report struct does not have all of this information so I changed the encrypt/decrypt interface for hybrid reports to take the appropriate Info struct as an argument. Should I do it this way or add the extra fields (conversion_site_domain, timestamp, epsilon, sensitivity) to the report struct?

Most of this should live in the associated data (`Info`), and we'll need to be able to access it on the `EncryptedConversionReport` in order to do some filtering and preprocessing. It should be fine to access it through the `info` field, thought we could also implement getters on the `EncryptedConversionReport` struct if we wanted.

All these fields do not need to persist into the `HybridImpressionReport` or `HybridConversionReport` after decryption.

> 2. I left some of the old EncryptedHybridReport struct code and implemented my changes in a new struct EncryptedHybridGeneralReport. All the tests in hybrid.rs work with the new code, but there is other code here: https://github.com/tyurek/ipa/blob/hybrid_conversion/ipa-core/src/query/runner/hybrid.rs#L65 that isn't compatible: It assumes that EncryptedHybridReport can be initialized from the encrypted bytes of an OPRF report. I did write a function to create a (decrypted) HybridReport from encrypted OPRF report bytes, but EncryptedHybridReport doesn't store bytes: it's an enum. So I'm not really sure what do do with this part of the code, but I figured I would get some feedback on the rest before I tear this up too badly

Our overall goal is to remove the reliance on the OPRF report. The high level flow should be 
```
BodyStream → EncryptedHybridReport → HybridReport → IndistinguishableHybridReport
```
and this diff is primarily concerned with the first two conversions. It's fine to rewrite the existing tests that us OPRF reports to use the new structs that you are creating.

> but EncryptedHybridReport doesn't store bytes: it's an enum. So I'm not really sure what do do with this part of the code, 

The two underlying structs in that enum do store bytes, so it should be possible to implement `TryFrom<Bytes>` on `EncryptedHybridReport` and build the correct sub-struct (either the `EncryptedHybridConversionReport` or the `EncryptedHyrbidImpressionReport`.)

Hope that all helps / makes sense.

ipa-core/src/report/hybrid.rs
```diff
@@ -50,8 +46,37 @@ pub enum InvalidHybridReportError {
     DeserializationError(&'static str, #[source] BoxError),
     #[error("report is too short: {0}, expected length at least: {1}")]
     Length(usize, usize),
+    #[error("unknown event type: {0}. Only 0 and 1 are allowed")]
+    UnknownEventType(u8),
+    #[error("Incorrect hybrid info type: Expected {0}")]
+    WrongInfoType(&'static str),
 }
 
+/*#[derive(thiserror::Error, Debug)]
```
eriktaubeneck, line 55: there are a few commented out lines like this scattered though out that should be cleaned up

ipa-core/src/report/hybrid.rs
tyurek: I saw that mk_ciphertext was needed for UniqueBytes. Not sure about the other ones though (or if we might need other ones later?)

ipa-core/src/report/hybrid.rs
tyurek: Done, it was just taking me a while to get it working with the usage in query/runner/hybrid.rs

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -20,31 +20,40 @@ use crate::{
...
     },
     secret_sharing::replicated::semi_honest::AdditiveShare as Replicated,
 };
 
 #[allow(dead_code)]
-pub struct Query<C, HV, R: PrivateKeyRegistry> {
+pub struct Query<'a, C, HV, R: PrivateKeyRegistry> {
     config: HybridQueryParams,
     key_registry: Arc<R>,
+    hybrid_info: HybridInfo<'a>,
```
akoshelev, line 36: I probably missed that on the previous review, but helper origin is assumed to be known [statically](https://github.com/private-attribution/ipa/blob/9477db8a38f4bf40cfb946ab59960cee04a9f69c/ipa-core/src/report/hybrid.rs#L60) to helpers. This should help to avoid the lifetime proliferation and the need to keep it inside query struct

If there is a need for having it, we can assume helper origin is `&'static str`

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -165,7 +179,7 @@ mod tests {
         // TODO: When Encryption/Decryption exists for HybridReports
         // update these to use that, rather than generating OprfReports
         vec![
-            TestRawDataRecord {
+            /*TestRawDataRecord {
```
akoshelev, line 182: delete?

ipa-core/src/report/hybrid.rs
```diff
@@ -224,35 +363,132 @@ impl<BK, V> HybridReport<BK, V>
...
+    Replicated<BK>: Serializable,
+    Replicated<V>: Serializable,
+    <Replicated<BK> as Serializable>::Size: Add<U16>,
+    <Replicated<V> as Serializable>::Size: Add<U16>,
+    <<Replicated<BK> as Serializable>::Size as Add<<Replicated<BA64> as Serializable>::Size>>:: Output: ArrayLength,
+    <<Replicated<V> as Serializable>::Size as Add<<Replicated<BA64> as Serializable>::Size>>:: Output: ArrayLength,
 {
+    /// # Panics
+    /// If report length does not fit in `u16`.
+    pub fn encrypted_len(&self) -> u16 {
```
akoshelev, line 375: It seems Codecov wants some coverage for the impression path as well

ipa-core/src/report/hybrid.rs
```diff
@@ -343,6 +580,105 @@ where
...
+    /// contents properly, which would be a bug.
+    pub fn decrypt<P: PrivateKeyRegistry>(
+        &self,
+        key_registry: &P,
+        info: &HybridConversionInfo,
+    ) -> Result<HybridConversionReport<V>, InvalidHybridReportError> {
+        type CTMKLength = Sum<<Replicated<BA64> as Serializable>::Size, TagSize>;
+        type CTBTTLength<V> = <<Replicated<V> as Serializable>::Size as Add<TagSize>>::Output;
+
+        println!("data: {:?}", self.data);
```
akoshelev, line 658:
```suggestion
```

ipa-core/src/report/hybrid.rs
```diff
@@ -262,12 +498,14 @@ where
...
         + TagSize::USIZE
-        + <Replicated<BA64> as Serializable>::Size::USIZE);
+        //+ <Replicated<BA64> as Serializable>::Size::USIZE);
+        + Replicated::<BA64>::size());
     const CIPHERTEXT_BTT_OFFSET: usize = Self::ENCAP_KEY_BTT_OFFSET + EncapsulationSize::USIZE;
 
     const KEY_IDENTIFIER_OFFSET: usize = (Self::CIPHERTEXT_BTT_OFFSET
         + TagSize::USIZE
-        + <Replicated<BK> as Serializable>::Size::USIZE);
+        //+ <Replicated<BK> as Serializable>::Size::USIZE);
```
akoshelev, line 507: Lets clean up commented code before merging it

ipa-core/src/report/hybrid.rs
```diff
@@ -224,35 +363,132 @@ impl<BK, V> HybridReport<BK, V>
...
-        _key_registry: &impl PublicKeyRegistry,
-        _rng: &mut R,
-    ) -> Result<Vec<u8>, InvalidReportError> {
-        unimplemented!()
+        key_id: KeyIdentifier,
+        key_registry: &impl PublicKeyRegistry,
+        info: &HybridInfo,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, InvalidHybridReportError> {
+        match self {
```
akoshelev, line 427: maybe you could do something like this

```rust
match (self, info) {
   (Impression, Impression) => { .. }
   (Conversion, Conversion) => { .. }
   _ => Err()
}
```

ipa-core/src/report/hybrid.rs
```diff
@@ -224,35 +363,132 @@ impl<BK, V> HybridReport<BK, V>
...
+        key_registry: &impl PublicKeyRegistry,
+        info: &HybridInfo,
+        rng: &mut R,
+        out: &mut B,
+    ) -> Result<(), InvalidHybridReportError> {
+        match self {
+            HybridReport::Impression(impression_report) => match info {
+                HybridInfo::Impression(impression_info) =>{
+                    out.put_u16_le(self.encrypted_len());
+                    out.put_u8(1u8);
```
akoshelev, line 400: this is a nit, but I think we are used to impressions encoded as 0 and conversions as 1. Maybe it is worth maintaining that

ipa-core/src/report/hybrid.rs
```diff
@@ -224,35 +363,132 @@ impl<BK, V> HybridReport<BK, V>
...
-        unimplemented!()
+        key_id: KeyIdentifier,
+        key_registry: &impl PublicKeyRegistry,
+        info: &HybridInfo,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, InvalidHybridReportError> {
+        match self {
+            HybridReport::Impression(impression_report) => match info {
+                HybridInfo::Impression(impression_info) =>
+                    // Prepend a 1u8 byte to indicate this is an impression report
```
akoshelev, line 430: I'd prefer if we use a type here as encoding 0 and 1 in more than one place can get out of sync quickly.

ipa-core/src/report/hybrid.rs
```diff
@@ -224,35 +363,132 @@ impl<BK, V> HybridReport<BK, V>
...
+                    impression_report.encrypt(key_id, key_registry, impression_info, rng).map(|v| vec![1u8].into_iter().chain(v).collect()),
```
akoshelev, line 431:
```suggestion
                    impression_report.encrypt(key_id, key_registry, impression_info, rng).map(|v| once(1_u8).chain(v).collect()),
```

ipa-core/src/query/runner/hybrid.rs
tyurek: Currently the test only uses raw records for conversions (impressions are commented). Ideally we'd have both, but I'm not sure how exactly we want to structure HybridInfo to allow that. I mention this a bit in the initial PR message

ipa-core/src/report/hybrid.rs
tyurek: Yeah, fixing the query runner tests to use both impressions and conversions should help here

ipa-core/src/report/hybrid.rs
tyurek: Good catch, I'll change it

ipa-core/src/report/hybrid.rs
tyurek: I'm curious, does this perform differently, or is it just more succinct?

ipa-core/src/report/hybrid.rs
tyurek: Ah sorry, missed that one

ipa-core/src/query/runner/hybrid.rs
tyurek: The design doc for this project has the helper origin as part of both HybridConversionInfo and HybridImpressionInfo, so I assume it's needed? I can try static though

akoshelev: yes, it is definitely needed, but it is assumed to be per MPC ring, i.e. known in advance. We hardcoded it in our Rust implementation, but generally I think `&static str` would be more appropriate

ipa-core/src/report/hybrid.rs
akoshelev: I am not sure if rustc is smart enough to avoid an allocation here, so just prefer to be explicit

ipa-core/src/query/runner/hybrid.rs
tyurek: Is conversion_site_domain also static? If so, we can remove the lifetime

ipa-core/src/report/hybrid.rs
tyurek: oh cool, didn't know you could do that :)

ipa-core/src/report/hybrid.rs
tyurek: yeah, I think I'll just remove this once all the oprf is completely disentangled

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -162,10 +176,8 @@ mod tests {
     const EXPECTED: &[u128] = &[0, 8, 5];
 
     fn build_records() -> Vec<TestRawDataRecord> {
```
eriktaubeneck, line 178: `TestRawDataRecord` is for the IpaOprfRerports, and we kept that around as we we're still decrypting `EncryptedOprfReport`s. We should be updating this to [`TestHybridRecords`](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/test_fixture/hybrid.rs#L11). This will require a little bit of plumbing (I think you'll need to implement `IntoShares` to produce `HybridReport`s and `Reconstruct`to convert 3 `HybridReport`s back into a `TestHybridReport`.)

ipa-core/src/report/hybrid.rs
```diff
@@ -224,50 +385,135 @@ impl<BK, V> HybridReport<BK, V>
...
-#[derive(Copy, Clone, Eq, PartialEq)]
-pub struct EncryptedHybridImpressionReport<BK, B>
+#[derive(Clone, Eq, PartialEq)]
+pub struct EncryptedHybridImpressionReport<BK>
 where
-    B: Deref<Target = [u8]>,
     BK: SharedValue,
 {
-    data: B,
+    data: Bytes,
```
eriktaubeneck, line 498: this feels much cleaner, but I just want to make sure @akoshelev sees this, since he may know the history of why we used `B: Deref<Target = [u8]>` previously, and can confirm we don't need it anymore.

ipa-core/src/test_fixture/input/sharing.rs
```diff
@@ -69,6 +72,47 @@ where
     }
 }
 
+impl<BK, V> IntoShares<HybridReport<BK, V>> for TestRawDataRecord
```
eriktaubeneck, line 75: with respect to the comment above about using `TestHybridRecord`, you should be able to essentially reuse this with some small tweaks!

eriktaubeneck (commented):
Overall this is looking good. I left a couple comments about using `TestHybridRecord` instead of `TestRawDataRecord`.

Is the plan to detangle `OprfReport` out entirely in this PR, or do that in the next PR? I'm OK with either.

ipa-core/src/report/hybrid.rs
akoshelev: yea, we chatted about it with @tyurek. We need to inspect the first byte and cut it off before sending it down for further processing. It is not possible with the `B` bounds. `Bytes` would provide similar ergonomics

tyurek:
I'm trying to detangle exactly as much as is needed to get tests to pass, with the goal of doing the rest in the next PR

ipa-core/src/query/runner/hybrid.rs
akoshelev: Conversion site is dynamic and provided by the submitter when they want to execute this query

ipa-core/src/report/hybrid_info.rs
```diff
@@ -2,43 +2,95 @@ use crate::report::{hybrid::NonAsciiStringError, KeyIdentifier};
...
+    }
+}
+
+#[derive(Clone, Debug)]
 pub struct HybridConversionInfo<'a> {
     pub key_id: KeyIdentifier,
-    pub helper_origin: &'a str,
-    pub converion_site_domain: &'a str,
+    pub helper_origin: &'static str,
+    pub conversion_site_domain: &'a str,
```
akoshelev, line 58: There can be many conversion sites in publisher queries - we never properly implemented it and I believe just hardcoded both helper origin and conversion site. I think this is fine for now, just wanted to explain it

akoshelev (approved):

PR #1383, opened 2024-10-29
Tolerate empty vectors of shuffle intermediates

andyleiserson:
This is more important for the sharded shuffle, which for small inputs is reasonably likely to produce an empty output on some shard. It adds a new `compute_non_empty_hash` function with the existing behavior of rejecting empty input, and changes the `compute_hash` function to accept an empty input. Then, it changes all of the existing calls to `compute_hash`, except the one in `shuffle::malicious`, to call `compute_non_empty_hash` instead. (I did not analyze the existing uses for others that may be able to tolerate an empty input.)

### DIFF
--- a/ipa-core/src/helpers/hashing.rs
+++ b/ipa-core/src/helpers/hashing.rs
@@ -76,12 +76,10 @@ where
 
 /// Computes Hash of serializable values from an iterator
 ///
-/// This version panics if an empty input is provided. This can offer defense-in-depth
-/// by helping to prevent fail-open bugs when the input should never be empty.
-///
 /// ## Panics
-/// Panics when Iterator is empty.
-pub fn compute_non_empty_hash<I, T, S>(input: I) -> Hash
+/// Panics if an empty input is provided. This can offer defense-in-depth by helping to
+/// prevent fail-open bugs when the input should never be empty.
+pub fn compute_hash<I, T, S>(input: I) -> Hash
 where
     I: IntoIterator<Item = T>,
     T: SerializeAs<S>,
@@ -93,13 +91,15 @@ where
 }
 
 /// Computes Hash of serializable values from an iterator
-pub fn compute_hash<I, T, S>(input: I) -> Hash
+///
+/// Unlike `compute_hash`, this version accepts empty inputs.
+pub fn compute_possibly_empty_hash<I, T, S>(iter: I) -> Hash
 where
     I: IntoIterator<Item = T>,
     T: SerializeAs<S>,
     S: Serializable,
 {
-    let (hash, _) = compute_hash_internal(input);
+    let (hash, _) = compute_hash_internal(iter);
     hash
 }
 
@@ -136,7 +136,7 @@ where
     );
 
     // set state
-    let combine = compute_non_empty_hash([left, right]);
+    let combine = compute_hash([left, right]);
     let mut buf = GenericArray::default();
     combine.serialize(&mut buf);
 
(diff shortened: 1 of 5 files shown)


ipa-core/src/helpers/hashing.rs
```diff
@@ -74,9 +70,37 @@ where
...
+}
+
+/// Computes Hash of serializable values from an iterator
+///
+/// This version panics if an empty input is provided. This can offer defense-in-depth
+/// by helping to prevent fail-open bugs when the input should never be empty.
+///
+/// ## Panics
+/// Panics when Iterator is empty.
+pub fn compute_non_empty_hash<I, T, S>(input: I) -> Hash
```
akoshelev, line 84: Is it better to return `Option<Hash>` here and let callers explicitly decide what to do when the input is empty? I feel like treating 0 as a valid hash is dangerous if you accidentally call `compute_hash` because after that all bets are off

ipa-core/src/helpers/hashing.rs
```diff
@@ -74,9 +70,37 @@ where
         sha.update(&buf);
```
martinthomson, line 70: Is there a risk that this is a bit churn-y.  Is there value in writing multiple items before updating the hash?  Or are we OK with our vectorization code.

ipa-core/src/helpers/hashing.rs
martinthomson: How often do you need an empty hash?  Or could those potentially-empty cases be amended with once(foo).chain(input) at the caller?

ipa-core/src/helpers/hashing.rs
```diff
@@ -74,9 +70,37 @@ where
...
+    T: SerializeAs<S>,
+    S: Serializable,
+{
+    let (hash, empty) = compute_hash_internal(input);
+    assert!(!empty, "must not provide an empty iterator");
+    hash
+}
+
+/// Computes Hash of serializable values from an iterator
+pub fn compute_hash<I, T, S>(input: I) -> Hash
```
martinthomson, line 96: I'd flip the naming, so that the (maybe) empty hash requires the special name.

ipa-core/src/helpers/hashing.rs
akoshelev: I think we just saw our first case where some shards after shuffle ended up with 0 shares. To me, the fact that hash was never computed, feels like something callers need to be aware of.

ipa-core/src/helpers/hashing.rs
andyleiserson: My guess is that this can optimize reasonably well. I'm inclined to leave it unless/until it shows up in profiling.

ipa-core/src/helpers/hashing.rs
andyleiserson: `Option<Hash>` would mean adding an `expect` to most calls, and would also mean doing something like `unwrap_or(SHA256_HASH_OF_EMPTY_MESSAGE)` for the empty-ok case, since we need to send something to the other helper to signify "I have no data".

I'd rather expose an extra function than prepend a dummy value to force a message to be non-empty, which seems like a kludge.

akoshelev (approved):

andyleiserson:
Another question is whether empty vs. not is even the right property to be looking at, or if we should be looking at the amount of entropy in the input (i.e. at least 128 bits) instead. On the other hand, I don't know if I necessarily think these properties should be enforced in this routine at all. The callers know more about the input and can check a more precise set of properties. Looking at the places `compute_hash` gets used, I see three categories: (1) malicious shuffle verification (this PR); (2) input is an array; (3) in `validate_three_two_way_sharing_of_zero`, which is unused.

PR #1384, opened 2024-10-29
Remove use of MAC-upgraded contexts with Boolean

andyleiserson:
This changes a bunch of tests that were upgrading Boolean contexts to MAC-semi-honest to instead use DZKP-semi-honest. An upgrade of some kind became necessary after removing support for MPC primitives from the base context. When that first happened, DZKP upgrades were not ready. Now that DZKP upgrades work, it makes more sense to use them here -- it's unlikely we would want to use MAC security with a Boolean field, because it is significantly more expensive than the DZKPs.

Making this change avoids the need to maintain MAC support for `Boolean`, which was never properly implemented. (When we actually used it, it was with `Gf2`, which is distinct from `Boolean`, for better or worse.)

### DIFF
--- a/ipa-core/src/protocol/basics/mod.rs
+++ b/ipa-core/src/protocol/basics/mod.rs
@@ -79,11 +79,6 @@ where
 {
 }
 
-impl<'a, B: ShardBinding> BooleanProtocols<UpgradedSemiHonestContext<'a, B, Boolean>>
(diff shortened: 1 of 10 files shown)


akoshelev (commented):
👍

akoshelev (approved):

PR #1385, opened 2024-10-29
Renaming MpcHelperClient to IpaHttpClient<Helper>

cberkhoff:
Two things happened in here:
1. Removed the default `Helper` for `ConnectionFlavor` in the Http Client. Now you have to be explicit.
2. Renamed `MpcHelperClient` to `IpaHttpClient` since it can handle both Mpc and Shard  communication.


akoshelev (approved):
👍

PR #1387, opened 2024-10-29
Hybrid shuffle

eriktaubeneck:
This implements the plumbing required to shuffle `IndistinguishableHybridReports` and adds the shuffle step to the protocol. 

As we discussed earlier, we may need to add the right Sharded + Malicious trait which will then make the `ctx` used here have the sharded + malicious shuffle.


ipa-core/src/protocol/hybrid/shuffle.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    Ok(shuffled
+        .into_iter()
+        .map(|item| shuffled_to_hybrid_report(&item))
+        .collect::<Vec<_>>())
+}
+
+/// Converts `IndistinguishableHybridReport` into
+/// an `AdditiveShare` needed for shuffle protocol.
+pub fn hybrid_report_to_shuffle_input<YS, BK, V>(
```
eriktaubeneck, line 46: would it make more sense to implement these as `From` traits. 

Right now, `ctx.shuffle` has the trait bound 
```
I: IntoIterator<Item = AdditiveShare<S>> + Send,
```
but you could also consider updating that so that `Item` implements conversion into `AdditiveShare<S>`, rather than being an `AdditiveShare` directly. 

Updating this now will either mean updating existing IPA to work the same way (yuck!), so maybe we wait for now.

ipa-core/src/protocol/hybrid/shuffle.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    offset += BK::BITS as usize;
+    expand_shared_array_in_place(&mut y, &input.value, offset);
+
+    y
+}
+
+/// Converts an `AdditiveShare` obtained from shuffle protocol
+/// into an `IndistinguishableHybridReport`.
+pub fn shuffled_to_hybrid_report<YS, BK, V>(
```
cberkhoff, line 69: these go away with the new sharded shuffle and its `Shufflable` trait?

ipa-core/src/protocol/hybrid/shuffle.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    #[test]
+    fn test_shuffle_hybrid_inputs() {
+        const BATCHSIZE: usize = 50;
+        run(|| async {
+            let world = TestWorld::default();
+
+            let mut rng = rand::thread_rng();
+            let mut records = Vec::new();
```
cberkhoff, line 133: nit: Can you do something like?
```
let records = (0..BATCHSIZE).map(|_| TestIndistinguishableHybridReport{ ... })
```

akoshelev:
I don't think we should use the old shuffle as it is not sharded and it won't be possible to make it sharded. Instead, we should take the sharded implementation for Hybrid (protocol/prf/shuffle/sharded.rs)

akoshelev:
mm, I didn't mean to close it, it was a misclick

ipa-core/src/protocol/hybrid/shuffle.rs
```diff
@@ -0,0 +1,248 @@
...
+                .map(|_| TestIndistinguishableHybridReport {
+                    match_key: rng.gen::<u64>(),
+                    breakdown_key: rng.gen_range(0u32..1 << 8),
+                    value: rng.gen_range(0u32..1 << 3),
+                })
+                .collect::<Vec<_>>();
+
+            // let mut result: Vec<TestIndistinguishableHybridReport> = world
+            //     .semi_honest(records.clone().into_iter(), |ctx, input_rows| async move {
+            //         ctx.shuffle(input_rows).await.unwrap()
```
eriktaubeneck, line 238: This gives a compile error:
```
ctx.shuffle(input_rows).await.unwrap()
    |                         ^^^^^^^ expected `AdditiveShare<_>`, found `IndistinguishableHybridReport<_, _>`
 ```
 and I can't seem to find the right trait that accepts `I: IntoIterator<Item = Shuffleable> + Send` (instead of `I: IntoIterator<Item = AdditiveShare<S>> + Send`).

File: ipa-core/src/protocol/hybrid/shuffle.rs [hunk 5]
```diff
@@ -0,0 +1,248 @@
...
+    match_key: BA64,
+    value: V,
+    breakdown_key: BK,
+}
+
+impl<BK, V> Serializable for IndistinguishableHybridShare<BK, V>
+where
+    BK: SharedValue,
+    V: SharedValue,
+    <BK as Serializable>::Size: Add<U8>,
```
eriktaubeneck, line 28: It was really unclear why this needs to be a `U8`, and this doesn't work with `Add<<BA64 as Serializable>::Size>` which seems to resolve to `U8`. Very open to any suggestions to how we may clean this up.

ipa-core/src/protocol/hybrid/shuffle.rs
eriktaubeneck: That doesn't seem to be what's happening, as `ctx.shuffle` seems to be coming from [here](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/protocol/ipa_prf/shuffle/mod.rs#L115). 

That said, it looks like this is going to change with #1402 (and even before that, this needs to bring in changes from main.) I'll sync with @andyleiserson tomorrow, but it looks like I may need to take a different approach with this to work with that PR.

PR #1391, opened 2024-10-30
Optimize Ristretto points handling in IPA

akoshelev:
I noticed that we do quite a bit of compress/decompress operations inside IPA to compute PRF values. For each row we do

Compress:
* 2 operations to convert from `Fp25519` (secret share)
* 2 operations to reveal (reveal requires two additions, each addition is one compress)
* 1 multiplication (on revealed value)

Decompress:
* 4 operations to reveal (two additions, each requires decompressing both RHS and LHS)
* 1 operation to multiply

In total we have 10 and each of those requires an inverse operation which shows up in the profiler as quite expensive. I ran some DRAFT queries on preliminary implementation and it showed ~14% latency drop for storing uncompressed Ristretto points for **semi-honest** implementation. I want to check malicious as well, although there is a bigger issue there, so it may hinder performance gains. Regardless of that, 14% speed up seems worth pursuing

### DIFF
--- a/ipa-core/src/ff/curve_points.rs
+++ b/ipa-core/src/ff/curve_points.rs
@@ -211,7 +211,9 @@ impl rand::distributions::Distribution<RP25519> for rand::distributions::Standar
 #[derive(Clone, Copy, Eq, Debug)]
 enum RistrettoRepr {
     /// In PRF code this path is never used as
-    /// we always construct Ristretto points from scalars
+    /// we always construct Ristretto points from scalars.
+    /// Constructing this value and attempting to use it
+    /// as Ristretto point will panic
     Zero,
     Point(RistrettoPoint),
 }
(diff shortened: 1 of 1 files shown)


akoshelev:
I want to do some draft runs before marking it ready for review

akoshelev:
| Mode | Before this change | After this change |
| -----  | ------------------- | ----------------- |
| Semi-honest/1M | [1h 42m](https://draft-mpc.vercel.app/query/view/tidy-slate2024-10-30T0602) | [1h 27m](https://draft-mpc.vercel.app/query/view/lazy-duty2024-10-30T2146)

akoshelev:
I won't merge it until I get the data for malicious, but it would be great to get some feedback as I am not planning to change the code significantly

akoshelev:
No penalty on malicious, but I'd take the perf win on semi-honest, it will benefit us in the long term when malicious implementation is fixed

tyurek:
This looks good to me. Keeping around an uncompressed point for math operations makes more sense than constantly compressing and decompressing. I don't see anything that looks suspect here with the usage

ipa-core/src/ff/curve_points.rs
```diff
@@ -196,7 +196,55 @@ impl rand::distributions::Distribution<RP25519> for rand::distributions::Standar
...
+            let zero = CompressedRistretto([0_u8; 32]);
+            // we could also cache the compressed Ristretto, if we end up
+            // sending a lot of zeroes
+            zero.decompress().unwrap()
+        })
+    }
+
+    pub fn as_point(&self) -> &RistrettoPoint {
+        match self {
+            Self::Zero => Self::zero(),
```
andyleiserson, line 239: I'm inclined to panic here. The current implementation is not going to have constant-time operations when the repr is `Zero`. As you note, we don't create zero points now, so we are fine to use this implementation, but it would be safer to prevent someone trying to use it in a different context where it wouldn't be okay.

andyleiserson (approved):

ipa-core/src/ff/curve_points.rs
akoshelev: Panicking seems like a safer option for sure. Your point about constant time operation is valid, but I think it also applies to our previous implementation. the cost of `Rp25519::ZERO + Rp25519::from(Scalar)` is not the same as `Rp25519::from(Scalar) + Rp25519::from(Scalar)`.

PR #1392, opened 2024-10-31
Adding testing infrastructure to support sharded tests

cberkhoff:
Now we can start a sharded infrastructure for tests. I'm using the following changes to test the upcoming Query Workflow. I'm front-loading the tests that I will later use to make sure everything keeps working on every PR.

Created the `TestApp` struct which is used to create `HelperApp` by HTTP tests. HTTP Transport tests uses this new struct to spin up the hosts.  

One of the certificates I had uploaded before was wrong so I'm updating it now.

I plan to breakdown `tests.rs` in a later PR, once the changes I have in the backlog are merged.

### DIFF
--- a/ipa-core/src/net/test.rs
+++ b/ipa-core/src/net/test.rs
@@ -88,18 +88,17 @@ impl AddressableTestServer {
 /// Either a single Ring on MPC connection or all of the shards in a Helper.
 pub struct TestNetwork<F: ConnectionFlavor> {
     pub network: NetworkConfig<F>, // Contains Clients config
-    pub servers: Vec<Option<AddressableTestServer>>,
+    pub servers: Vec<AddressableTestServer>,
 }
 
 impl<F: ConnectionFlavor> TestNetwork<F> {
     /// Helper function that creates [`PeerConfig`]
     fn create_peers(
-        servers: &[Option<AddressableTestServer>],
+        servers: &[AddressableTestServer],
         conf: &TestConfigBuilder,
     ) -> Vec<PeerConfig> {
         servers
             .iter()
-            .flatten()
             .map(|addr_server| {
                 let port = addr_server
                     .config
@@ -139,7 +138,7 @@ impl TestNetwork<Shard> {
             .zip(ports)
             .map(|(ix, p)| {
                 let sid = ShardedHelperIdentity::new(id, ix);
-                Some(AddressableTestServer::new(sid, p, conf))
+                AddressableTestServer::new(sid, p, conf)
             })
             .collect();
         let peers = Self::create_peers(servers.as_slice(), conf);
@@ -158,7 +157,7 @@ impl TestNetwork<Helper> {
             .zip(ports)
             .map(|(id, p)| {
                 let sid = ShardedHelperIdentity::new(id, ix);
-                Some(AddressableTestServer::new(sid, p, conf))
+                AddressableTestServer::new(sid, p, conf)
             })
             .collect();
         let peers = Self::create_peers(servers.as_slice(), conf);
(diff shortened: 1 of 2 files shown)


ipa-core/src/net/test.rs
```diff
@@ -694,48 +816,52 @@ a0778c3e9960576cbef4312a3b7ca34137880fd588c11047bd8b6a8b70b5a151
...
-        let leader_ring_configs = &conf.leaders_ring().servers;
-        assert_eq!(leader_ring_configs[2].config.port, Some(10002));
+
+        let apps = conf.into_apps();
+        for (i, ports) in FOUR_SHARDS.iter().enumerate() {
+            for (j, port) in ports.ring.into_iter().enumerate() {
+                assert_eq!(apps[i * 3 + j].mpc_server.config.port, Some(port));
+                assert_eq_configs(
+                    &apps[i * 3].mpc_network_config,
+                    &apps[i * 3 + j].mpc_network_config,
```
akoshelev, line 854: I remember you added a function to compute that index, would it work here?

ipa-core/src/net/test.rs
```diff
@@ -619,14 +741,14 @@ JMepVZwIWJrVhnxdcmzOuONoeLZPZraFpw==
 ",
     b"
 -----BEGIN CERTIFICATE-----
-MIIBZTCCAQugAwIBAgIIXTgB/bkN/aUwCgYIKoZIzj0EAwIwFDESMBAGA1UEAwwJ
-bG9jYWxob3N0MB4XDTI0MTAwNjIyMTIwM1oXDTI1MDEwNTIyMTIwM1owFDESMBAG
-A1UEAwwJbG9jYWxob3N0MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEyzSofZIX
-XgLUKGumrN3SEXOMOAKXcl1VshTBzvyVwxxnD01WVLgS80/TELEltT8SMj1Cgu7I
-tkDx3EVPjq4pOKNHMEUwFAYDVR0RBA0wC4IJbG9jYWxob3N0MA4GA1UdDwEB/wQE
+MIIBZTCCAQugAwIBAgIITIDzw5k9qXIwCgYIKoZIzj0EAwIwFDESMBAGA1UEAwwJ
```
akoshelev, line 744: what drove this change?

ipa-core/src/net/test.rs
```diff
@@ -45,6 +46,38 @@ pub const DEFAULT_TEST_PORTS: Ports = Ports {
     shards: [6000, 6001, 6002],
 };
 
+/// A network with two shards per helper.
+pub const TWO_SHARDS: [Ports; 2] = [
```
akoshelev, line 50: are we using this to spin up an actual server in unit tests? if so, I don't think we should do it - we have to pick ports randomly

ipa-core/src/net/test.rs
```diff
@@ -87,17 +120,18 @@ impl AddressableTestServer {
 /// Either a single Ring on MPC connection or all of the shards in a Helper.
 pub struct TestNetwork<F: ConnectionFlavor> {
     pub network: NetworkConfig<F>, // Contains Clients config
-    pub servers: Vec<AddressableTestServer>,
+    pub servers: Vec<Option<AddressableTestServer>>,
```
akoshelev, line 123: is it what you want, or `Option<Vec<..>>` represents the state where you don't have ports assigned better?

ipa-core/src/net/test.rs
cberkhoff: We have both tests. I can remove the one using this set.

ipa-core/src/net/test.rs
cberkhoff: Not really, in here is more of a rust ownership constraint.
I need to be able to take the ports.

ipa-core/src/net/test.rs
cberkhoff: Since I'm enumerating over the ports it won't be straight forward. I don't want to iterate over the computed results since they could have a problem (perhaps a missing ring), I rather iterate over the "known" result.

ipa-core/src/net/test.rs
akoshelev: this to me indicates that there is something off here. Consider consuming `self` and using `servers.into_iter()` to get `AddressableTestServer` instead of `&mut Option<AddressableTestServer>` that you can later replace

ipa-core/src/net/test.rs
akoshelev: I looked at the change and I don't understand whether we ever create a server with hardcoded ports or no. My line of thinking is that we should never need that - hardcoded ports are only good for config validation, everything else should use ports assigned by Kernel

ipa-core/src/net/test.rs
cberkhoff: Sounds good. This was a vestigial change before I used destructuring and zip in the `to_app` function. It was simple to remove. Good catch!

ipa-core/src/net/test.rs
cberkhoff: Yes, I modified the code so that we only use specified ports for config unit tests. All the HTTP tests now use OS given ports.

akoshelev (approved):
Thanks for addressing comments @cberkhoff, I like the conciseness of this change now. Great job simplifying the code

PR #1393, opened 2024-10-31
[DRAFT} Batch convert RP25519 to u64

akoshelev:
Builds on top of #1391 and local testing shows some significant improvements in latency. Will do some Draft runs before making it ready for review.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -558,6 +558,10 @@ macro_rules! boolean_array_impl {
 
                     res
                 }
+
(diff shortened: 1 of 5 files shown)


akoshelev:
in-memory runs showed better performance, but with real io, I am only seeing 1.5 min lower latency with this change (~1.7% improvement.

The comparison is done against #1391 

| Mode | Before this change | After this change |
| -----  | ------------------- | ----------------- |
| Semi-honest/50M | 1h 28m](https://draft-mpc.vercel.app/query/view/lazy-duty2024-10-30T2146) | [1h 27 mins](https://draft-mpc.vercel.app/query/view/dingy-toxin2024-11-01T0233) |

PR #1395, opened 2024-10-31
Include `Add<&Self>` in `SharedValue`

andyleiserson:
This will be required to eliminate `Add` bounds from the shuffle traits.

This is on top of #1394.

### DIFF
--- a/ipa-core/src/ff/boolean.rs
+++ b/ipa-core/src/ff/boolean.rs
@@ -108,12 +108,20 @@ impl rand::distributions::Distribution<Boolean> for rand::distributions::Standar
     }
 }
 
-impl std::ops::Add for Boolean {
+impl std::ops::Add<&Boolean> for Boolean {
     type Output = Self;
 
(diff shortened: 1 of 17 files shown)


andyleiserson:
We actually don't have to do this, if we believe that `Copy` is always cheap. Shuffle operates on `BooleanArray`s, which are `SharedValue`s, which are `Copy`. (Originally, `ShuffleShare` only required `Send + Clone + FromRandom + Add<Output = Self>`. That's not sufficient for the malicious shuffle MACs. But we could come up with something less than `BooleanArray`.) 

For the things that we currently shuffle, I think it's reasonable to assume `Copy` is cheap. But I'm not sure we want to bake that assumption in to the implementation, which otherwise could easily be used to shuffle larger things.

We may also want to remove the implication that `BooleanArray`s are `Copy` if we add larger boolean arrays to increase vectorization dimensions for performance.

andyleiserson:
I don't think we need this. I am going to use a more precise set of bounds for `ShuffleShare`.

PR #1396, opened 2024-10-31
Prepare HTTP Endoint

cberkhoff:
Adding the HTTP endpoint for the prepare query API. Modified it so that the same code can be used for either shard/helper communication (both just forward the input to their respective transports).

### DIFF
--- a/ipa-core/src/net/transport.rs
+++ b/ipa-core/src/net/transport.rs
@@ -38,13 +38,14 @@ pub struct HttpTransport<F: ConnectionFlavor> {
 /// HTTP transport for helper to helper traffic.
 #[derive(Clone)]
 pub struct MpcHttpTransport {
-    inner_transport: Arc<HttpTransport<Helper>>,
+    pub(super) inner_transport: Arc<HttpTransport<Helper>>,
 }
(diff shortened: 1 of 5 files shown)


ipa-core/src/net/transport.rs
```diff
@@ -38,13 +38,13 @@ pub struct HttpTransport<F: ConnectionFlavor> {
 /// HTTP transport for helper to helper traffic.
 #[derive(Clone)]
 pub struct MpcHttpTransport {
-    inner_transport: Arc<HttpTransport<Helper>>,
+    pub(super) inner_transport: Arc<HttpTransport<Helper>>,
```
akoshelev, line 41: this should be fine as long as we plan to get rid of the wrapper struct (which I believe is what we want) - seems that `HttpTransport<Helper>` gives us what we want from it

akoshelev (approved):

ipa-core/src/net/transport.rs
cberkhoff: In a follow up PR I add shard count to the shared http transport. Just FYI

PR #1397, opened 2024-11-01
Use Shuffleable trait for shuffle inputs 

andyleiserson:
This implements the `Shuffleable` trait for `AttributionOutputs` and `OPRFIPAInputRow` and removes the glue functions around shuffle. The amount of code is about the same, but this avoids additional copies of the input/output that were allocated in the glue functions.


ipa-core/src/protocol/ipa_prf/shuffle/sharded.rs
```diff
@@ -97,14 +109,10 @@ pub trait ShuffleContext: ShardedContext {
             let mut resharded = assert_send(reshard_iter(
                 self.clone(),
                 data.enumerate().map(|(i, item)| {
-                    // FIXME(1029): update PRSS trait to compute only left or right part
```
eriktaubeneck, line 100: hurray!

eriktaubeneck (approved):

ipa-core/src/protocol/ipa_prf/aggregation/breakdown_reveal.rs
```diff
@@ -24,18 +25,74 @@ use crate::{
...
+            .write(&breakdown_key)
+            .write(&trigger_value);
+
+        share
+    }
+
+    fn split_fields(share: &<Self as Shuffleable>::Share) -> (BK, TV) {
+        let bits = BooleanArrayReader::new(share);
+        let (breakdown_key, bits) = bits.read();
+        let (trigger_value, _bits) = bits.read();
```
eriktaubeneck, line 58: possibly dumb question, but how does this know how many bits to read? (more practically, how would I do the same thing for the new type I'm working with in the hybrid protocol?)

andyleiserson: It is determined by the type parameter (i.e. `T` is `BK` and `BK::BITS` is 5 or 8). The type parameter in turn is determined by type inference working backwards from the return value.

eriktaubeneck: 👨‍🍳 😽

PR #1398, opened 2024-11-02
Prepare Query API Plumbing

cberkhoff:
A few changes required to make Prepare Query API work for the sharded scenario:

1. HandlerRef was generalized to support both `<ShardIndex>` and `<HelperIdentity>`
2. `app::Setup` returns also a `HandlerRef<ShardIndex>` to give to the ShardHttpTransport.
3. Added `ShardTransports` trait. This one provides an additional function to retrieve my shard configuration (`Sharded`). I need shard transports to know how many shards are in total. I will use this functionality in the Prepare Query API (query processor) so that the leader shard can send messages to all other shards. Note there's no Context at this stage, these are the preliminary steps of the query.
4. Created a simple `net::errors::ShardError`. This one only wraps the general net Error with shard information. This is needed by the processor to be able to distinguish errors that happen in the MPC transport vs the ones in the shard transport. Also provides more context information for the error itself.
5. Related to 3; So that TestWorlds can create In-Memory Shard Transport I had to expand `ShardContext = Option<ShardIndex>` to also contain the total count of shards. For simplicity I ended up using `Option<ShardIndex>`. The name ShardContext was overly generic so I removed it.

### DIFF
--- a/ipa-core/src/app.rs
+++ b/ipa-core/src/app.rs
@@ -13,6 +13,7 @@ use crate::{
     hpke::{KeyRegistry, PrivateKeyOnly},
     protocol::QueryId,
     query::{NewQueryError, QueryProcessor, QueryStatus},
+    sharding::ShardIndex,
     sync::Arc,
     utils::NonZeroU32PowerOfTwo,
 };
@@ -46,7 +47,8 @@ impl AppConfig {
 
 pub struct Setup {
     query_processor: QueryProcessor,
-    handler: HandlerRef,
+    mpc_handler: HandlerRef<HelperIdentity>,
+    shard_handler: HandlerRef<ShardIndex>,
 }
 
 /// The API layer to interact with a helper.
@@ -67,21 +69,25 @@ struct Inner {
 
 impl Setup {
     #[must_use]
-    pub fn new(config: AppConfig) -> (Self, HandlerRef) {
+    pub fn new(config: AppConfig) -> (Self, HandlerRef<HelperIdentity>, HandlerRef<ShardIndex>) {
         let key_registry = config.key_registry.unwrap_or_else(KeyRegistry::empty);
         let query_processor = QueryProcessor::new(key_registry, config.active_work, config.runtime);
-        let handler = HandlerBox::empty();
+        let mpc_handler = HandlerBox::empty();
+        let shard_handler = HandlerBox::empty();
         let this = Self {
             query_processor,
-            handler: handler.clone(),
+            mpc_handler: mpc_handler.clone(),
+            shard_handler: shard_handler.clone(),
         };
 
         // TODO: weak reference to query processor to prevent mem leak
-        (this, handler)
+        (this, mpc_handler, shard_handler)
     }
 
     #[must_use]
-    pub fn with_key_registry(key_registry: KeyRegistry<PrivateKeyOnly>) -> (Self, HandlerRef) {
+    pub fn with_key_registry(
+        key_registry: KeyRegistry<PrivateKeyOnly>,
+    ) -> (Self, HandlerRef<HelperIdentity>, HandlerRef<ShardIndex>) {
         Self::new(AppConfig::default().with_key_registry(key_registry))
     }
 
@@ -96,9 +102,10 @@ impl Setup {
             mpc_transport,
             shard_transport,
         });
-        self.handler.set_handler(
-            Arc::downgrade(&app) as Weak<dyn RequestHandler<Identity = HelperIdentity>>
-        );
+        self.mpc_handler
+            .set_handler(Arc::downgrade(&app) as Weak<dyn RequestHandler<HelperIdentity>>);
+        self.shard_handler
+            .set_handler(Arc::downgrade(&app) as Weak<dyn RequestHandler<ShardIndex>>);
 
         // Handler must be kept inside the app instance. When app is dropped, handler, transport and
         // query processor are destroyed.
@@ -160,12 +167,34 @@ impl HelperApp {
 }
 
 #[async_trait]
-impl RequestHandler for Inner {
-    type Identity = HelperIdentity;
-
+impl RequestHandler<ShardIndex> for Inner {
     async fn handle(
         &self,
-        req: Addr<Self::Identity>,
+        req: Addr<ShardIndex>,
+        _data: BodyStream,
+    ) -> Result<HelperResponse, ApiError> {
+        let qp = &self.query_processor;
+
+        Ok(match req.route {
+            RouteId::PrepareQuery => {
+                let req = req.into::<PrepareQuery>()?;
+                HelperResponse::from(qp.prepare_shard(req)?)
+            }
+            r => {
+                return Err(ApiError::BadRequest(
+                    format!("{r:?} request must not be handled by shard query processing flow")
+                        .into(),
+                ))
+            }
+        })
+    }
+}
+
+#[async_trait]
+impl RequestHandler<HelperIdentity> for Inner {
+    async fn handle(
+        &self,
+        req: Addr<HelperIdentity>,
         data: BodyStream,
     ) -> Result<HelperResponse, ApiError> {
         fn ext_query_id(req: &Addr<HelperIdentity>) -> Result<QueryId, ApiError> {
@@ -179,7 +208,8 @@ impl RequestHandler for Inner {
         Ok(match req.route {
             r @ RouteId::Records => {
                 return Err(ApiError::BadRequest(
-                    format!("{r:?} request must not be handled by query processing flow").into(),
+                    format!("{r:?} request must not be handled by MPC query processing flow")
+                        .into(),
                 ))
             }
             RouteId::ReceiveQuery => {
--- a/ipa-core/src/app.rs
+++ b/ipa-core/src/app.rs
@@ -170,23 +170,10 @@ impl HelperApp {
 impl RequestHandler<ShardIndex> for Inner {
     async fn handle(
         &self,
-        req: Addr<ShardIndex>,
+        _req: Addr<ShardIndex>,
         _data: BodyStream,
     ) -> Result<HelperResponse, ApiError> {
-        let qp = &self.query_processor;
-
-        Ok(match req.route {
-            RouteId::PrepareQuery => {
-                let req = req.into::<PrepareQuery>()?;
-                HelperResponse::from(qp.prepare_shard(req)?)
-            }
-            r => {
-                return Err(ApiError::BadRequest(
-                    format!("{r:?} request must not be handled by shard query processing flow")
-                        .into(),
-                ))
-            }
-        })
+        Ok(HelperResponse::ok())
     }
 }
 
--- a/ipa-core/src/bin/helper.rs
+++ b/ipa-core/src/bin/helper.rs
@@ -19,7 +19,7 @@ use ipa_core::{
     executor::IpaRuntime,
     helpers::HelperIdentity,
     net::{ClientIdentity, IpaHttpClient, MpcHttpTransport, ShardHttpTransport},
-    sharding::ShardIndex,
+    sharding::{ShardIndex, Sharded},
     AppConfig, AppSetup, NonZeroU32PowerOfTwo,
 };
 use tokio::runtime::Runtime;
@@ -143,7 +143,7 @@ async fn server(args: ServerArgs, logging_handle: LoggingHandle) -> Result<(), B
         .with_active_work(args.active_work)
         .with_runtime(IpaRuntime::from_tokio_runtime(&query_runtime));
 
-    let (setup, handler) = AppSetup::new(app_config);
+    let (setup, handler, shard_handler) = AppSetup::new(app_config);
 
     let server_config = ServerConfig {
         port: args.port,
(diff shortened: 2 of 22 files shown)


ipa-core/src/bin/helper.rs
```diff
@@ -185,11 +185,14 @@ async fn server(args: ServerArgs, logging_handle: LoggingHandle) -> Result<(), B
     let shard_network_config = NetworkConfig::new_shards(vec![], shard_clients_config);
     let (shard_transport, _shard_server) = ShardHttpTransport::new(
         IpaRuntime::from_tokio_runtime(&http_runtime),
-        ShardIndex::FIRST,
+        Sharded {
```
akoshelev, line 188: something like `Sharded::with_shards(1)` could be more concise to read

ipa-core/src/config.rs
```diff
@@ -131,6 +131,13 @@ impl NetworkConfig<Shard> {
             identities,
         }
     }
+
+    /// # Panics
+    /// In the unlikely event a usize cannot be turned into a u32
+    #[must_use]
+    pub fn shard_count(&self) -> ShardIndex {
+        ShardIndex::from(u32::try_from(self.peers.len()).unwrap())
```
akoshelev, line 139:
```suggestion
        ShardIndex::try_from(self.peers.len()).unwrap()
```

ipa-core/src/config.rs
```diff
@@ -131,6 +131,13 @@ impl NetworkConfig<Shard> {
             identities,
         }
     }
+
+    /// # Panics
+    /// In the unlikely event a usize cannot be turned into a u32
```
akoshelev, line 136: nit: I wouldn't say it is unlikely for `usize`, but it is surely unexpected to have 4B shards

ipa-core/src/sharding.rs
```diff
@@ -167,26 +167,22 @@ pub trait ShardConfiguration {
     }
 }
 
-/// This is a runtime version of `ShardBinding`. It is used by the stream interceptor to
-/// avoid type parameter proliferation. It should not be used by protocols.
-pub type ShardContext = Option<ShardIndex>;
-
 pub trait ShardBinding: Debug + Send + Sync + Clone + 'static {
-    fn context(&self) -> ShardContext;
+    fn shard_config(&self) -> Option<Sharded>;
```
akoshelev, line 171: this deserves documenting the purpose of it like we had before for `ShardContext`

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -333,6 +333,10 @@ pub trait Transport: Clone + Send + Sync + 'static {
     }
 }
 
+pub trait ShardedTransport: Transport {
```
akoshelev, line 336: I think we should extend Transport trait instead with these 

```rust
    fn peer_count(&self) -> u32;

    /// Broadcasts a message to all peers, excluding this instance, collecting all failures and
    /// successes. This method waits for all responses and returns only when all peers responded.
    async fn broadcast<Q, S, R>(
        &self,
        route: R,
    ) -> Result<(), BroadcastError<Self::Identity, Self::Error>>
    where
        Option<QueryId>: From<Q>,
        Option<Gate>: From<S>,
        Q: QueryIdBinding,
        S: StepBinding,
        R: RouteParams<RouteId, Q, S> + Clone;
```

It is easy to implement, works for both MPC and Shard and avoids transport specialization which this code invites us to do

ipa-core/src/helpers/transport/handler.rs
```diff
@@ -173,18 +173,13 @@ pub enum Error {
 /// Trait for custom-handling different request types made against MPC helper parties.
 /// There is a limitation for RPITIT that traits can't be made object-safe, hence the use of async_trait
 #[async_trait]
-pub trait RequestHandler: Send + Sync {
-    type Identity: TransportIdentity;
+pub trait RequestHandler<I: TransportIdentity>: Send + Sync {
```
akoshelev, line 176: It was intentional to have `Identity` as an associated type - that ensures for each peer type we have exactly one handler. Unless we have to change it, I would prefer to have it as AT

akoshelev (changes requested):

ipa-core/src/bin/helper.rs
cberkhoff: This initialization to be the single shard is temporary. I created a `new` function to help with the conciseness.

ipa-core/src/helpers/transport/in_memory/transport.rs
```diff
@@ -268,7 +274,7 @@ impl Debug for InMemoryStream {
 }
 
 pub struct Setup<I> {
-    identity: I,
+    pub identity: I,
```
andyleiserson, line 277:
```suggestion
    identity: I,
```
Seems like `pub` is not needed here?

ipa-core/src/test_fixture/world.rs
```diff
@@ -1097,7 +1098,9 @@ mod tests {
                 config.role_assignment(),
                 |ctx: &MaliciousHelperContext, data: &mut Vec<u8>| {
                     assert!(ctx.shard.is_some());
```
andyleiserson, line 1100:
```suggestion
```
No need to have the assert and the unwrap.

ipa-core/src/helpers/transport/handler.rs
cberkhoff: Can you please clarify "ensures for each peer type we have exactly one handler"? The following code is perfectly fine to add:
```
struct AnotherHandler;

#[async_trait]
impl RequestHandler for AnotherHandler {
    type Identity = HelperIdentity;

    async fn handle(
        &self,
        req: Addr<HelperIdentity>,
        data: BodyStream,
    ) -> Result<HelperResponse, ApiError> {
        Ok(HelperResponse::ok())
    }
}
```
In other words, an associated type CAN have multiple handlers.


An associated type prevents having a single implementor support multiple associated types. I don't think that's a desirable property here. More concretely, It's simpler to have `Inner` implement both Shard and Mpc Handlers. The alternative is to have 2 structs which are going to be interconnected anyway because they want to use the same Query Processor and transports. I would advocate on keeping things simple and avoid proliferation of wrappers. That's what I'm proposing on this PR.


Let me know what you think, but I will need some more clarification on what you're overall trying to do to find the right abstraction.

andyleiserson: The concrete case where it's needed / desirable to implement both kinds of handlers is the `HelperApp` (it wasn't needed previously, but it is needed now, to handle prepare-query messages from the leader shard).

A lot of the uses of `RequestHandler` are via `make_owned_handler`, where there's an indirect limitation that the handler can only handle one identity type, because the handler wraps `Fn(Addr<I>)`.

I'm a bit surprised, but I can't even find other concrete types implementing `RequestHandler` at all, so I can't assess risk of incorrectly implementing `RequestHandler` for multiple identity types. Are there places I'm missing?

akoshelev: @andyleiserson We wanted to keep request handlers in one place, so not having too many implementations of `RequestHandlers` was the goal of #992. 

Given that we need a single type to implement both, generic is the way to go

ipa-core/src/helpers/transport/mod.rs
cberkhoff: I understand you would still allow Transport users to check the number of shards via `peer_count`?

akoshelev: that's right. Depending on prepare query orchestration logic, you may just use `broadcast`

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -333,6 +335,27 @@ pub trait Transport: Clone + Send + Sync + 'static {
...
+    /// These errors are usually related to an underlying [`Transport::Error`]. This adds more
+    /// information or allows to describe sharding specific issues.
+    type ShardError: std::fmt::Debug;
+
+    fn peer_count(&self) -> ShardIndex;
+
+    /// Broadcasts a message to all peers, excluding this instance, collecting all failures and
+    /// successes. This method waits for all responses and returns only when all peers responded.
+    /// The routes and data will be cloned.
+    async fn broadcast<Q, S, R, D>(&self, route: R, data: D) -> Result<(), Self::ShardError>
```
akoshelev, line 349: can we make it part of the `Transport` trait? My goal here was to avoid having `ShardedTransport` trait altogether. Specializing shard transport based on the existing transport interface seems like a wrong way to go to me.

cberkhoff: While it's technically possible, I don't see advantage in having one Transport. In fact, I think making both have the same API can lead to misuse. I rather have the compiler help the developer identify the transport it's using and its capabilities.

As an example; The query processor needs to know if it's dealing with a sharded or mpc transport for the Prepare API. Say you're the leader of helper 2 and you need to let your shards know about the incoming query. If both transports are interchangeable, then you could by mistake (a bug) call the broadcast on the mpc but not on the shard transport. With 2 different traits the compiler with protect you.

akoshelev: Imo, the distinction is already made in `Transport::Identity`, so it is not possible to swap them. Currently both are implemented using the same backend and making them conform to the same API is a better strategy as long as we plan to use the same implementation for both.

ipa-core/src/helpers/transport/in_memory/transport.rs
```diff
@@ -73,6 +73,7 @@ pub enum Error<I> {
 /// incoming messages.
 pub struct InMemoryTransport<I> {
     identity: I,
+    all_identities: Vec<I>,
```
akoshelev, line 69: you don't need to pass that, it can be derived

```rust
    pub fn peer_count(&self) -> u32 {
        u32::try_from(self.connections.len()).expect("Less than 4B peers")
    }
```

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -289,15 +289,33 @@ impl RouteParams<RouteId, QueryId, NoStep> for (RouteId, QueryId) {
...
 pub trait Transport: Clone + Send + Sync + 'static {
     type Identity: TransportIdentity;
     type RecordsStream: BytesStream;
-    type Error: std::fmt::Debug;
+    type Error: std::fmt::Debug + Send;
 
+    /// Return my identity in the network (MPC or Sharded)
     fn identity(&self) -> Self::Identity;
 
+    /// Returns all the identities in this network (MPC or Sharded), myself included.
```
akoshelev, line 310: curious what is the use case to see all identities, and not just peers? because former can be easily derived on every side

ipa-core/src/helpers/transport/mod.rs
```diff
@@ -323,6 +341,31 @@ pub trait Transport: Clone + Send + Sync + 'static {
...
+    ) -> Result<(), BroadcastError<Self::Identity, Self::Error>>
+    where
+        Option<QueryId>: From<Q>,
+        Option<Gate>: From<S>,
+        Q: QueryIdBinding,
+        S: StepBinding,
+        R: RouteParams<RouteId, Q, S> + Clone,
+        D: Stream<Item = Vec<u8>> + Clone + Send + 'static,
+    {
+        try_join_all(self.peers().map(|peer| {
```
akoshelev, line 361: I love that it is a default implementation on this trait, `try_join_all` can be problematic though as it aggressively polls every future and is not optimized for large iterators. For MPC it is not a big deal, for shard connections it can be problematic. Given that you don't care about ordering here, `FuturesUnordered` is a better default imo

self.peers().map(|peer| {
            self.send(peer, route.clone(), data.clone())
                .map_err(move |source| BroadcastError { peer, source })
        })).collect::<FuturesUnordered>().await?;

ipa-core/src/helpers/transport/in_memory/transport.rs
cberkhoff: I wanted the API to return an iterator over identities `I` (instead of just integers) so that later I could use it directly in the `broadcast` function. In your example `peer_count` works fine for the generic `InMemoryTransport<I>` because it's just an `u32`, but since I was working with `I` I needed a generic way that works for any `I`. I could't derive a `ShardIndex` or a `HelperIdentity` on the same function.

It looks kind of redundant to pass a parameter that's derivable, but the problem is that I can only set it at a point where I know if it's a Helper or Shard `InMemoryTransport`.

One thing to consider is that this issue only happens in the `InMemoryTransport` which is shared for both Helper and Shard.

I could alternatively have `peer_count` and try to derive the peers using integers. Let me know.

ipa-core/src/helpers/transport/mod.rs
cberkhoff: It's just a helper function I can use to derive the peers regardless of the generic `Identity`.

ipa-core/src/helpers/transport/in_memory/transport.rs
andyleiserson: You may also be able to get it from the keys of the `connections` `HashMap` (plus `identity` for the local one).

ipa-core/src/helpers/transport/in_memory/transport.rs
```diff
@@ -162,15 +162,13 @@ impl<I: TransportIdentity> Transport for Weak<InMemoryTransport<I>> {
...
-        let this = self.identity();
-        let all: Vec<I> = self
-            .upgrade()
+        self.upgrade()
             .unwrap()
             .connections
             .keys()
             .copied()
-            .collect();
-        all.into_iter().filter(move |&id| id != this)
```
cberkhoff: why are you removing this filter? We want to broadcast to all the peers (not including ourselves).

cberkhoff: Discussed offline: in the in-memory we don't create a connection to ourselves.

ipa-core/src/app.rs
```diff
@@ -160,12 +167,21 @@ impl HelperApp {
...
 #[async_trait]
-impl RequestHandler for Inner {
-    type Identity = HelperIdentity;
+impl RequestHandler<ShardIndex> for Inner {
+    async fn handle(
+        &self,
+        _req: Addr<ShardIndex>,
+        _data: BodyStream,
+    ) -> Result<HelperResponse, ApiError> {
+        Ok(HelperResponse::ok())
```
akoshelev, line 176: It is more future proof to panic by default - returning an ACK for all request is a bad response

akoshelev (approved):
glad to see progress here!

PR #1400, opened 2024-11-04
Suppress metrics table in passing test runs

andyleiserson:
`std::io::stdout()` is not captured in test runs -- see https://github.com/rust-lang/rust/issues/90785.

The `Display` implementation for `comfy_table::Table` assembles everything into a string anyways (actually worse -- it assembles into a `Vec<String>` and then calls `join`). So, even if we were to have other places where we want the `std::io::Write` interface, there is unlikely to be a performance benefit of the previous implementation.


akoshelev (approved):
oh, that was my bad, thanks for fixing it!

PR #1401, opened 2024-11-04
Optimized modulo operator for `Fp61BitPrime`

akoshelev:
Zero-knowledge proofs use multiplication on this field extensively. For every bit of multiplication output, we perform [~10 multiplications](https://github.com/private-attribution/ipa/blob/cf25c69f7c641ed70de62c6536c0293e1e6b2db5/ipa-core/src/protocol/context/dzkp_field.rs#L122) in `Fp61BitPrime`. That's around $86 \times 10^9$ modulo reduction operations per 35 million MPC multiplications.

My preliminary testing shows ~30% improvement in local runs, will do some Draft runs to confirm

### DIFF
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -57,18 +57,44 @@ pub trait PrimeField: Field + U128Conversions {
 #[error("Field value {0} provided is greater than prime: {1}")]
 pub struct GreaterThanPrimeError<V: Display>(V, u128);
 
+/// Default modulo reduction implementation using the remainder operator.
+/// Works for any prime field, but can be inefficient for large primes
+/// and if it needs to be done often. Fields operating over Mersenne
+/// primes provide an optimized implementation.
+macro_rules! rem_modulo_impl {
+    ( $field:ident, $op_store:ty ) => {
+        impl $field {
+            #[must_use]
+            fn modulo_prime_base(input: $op_store) -> Self {
+                #[allow(clippy::cast_possible_truncation)]
+                Self((input % <$op_store>::from(Self::PRIME)) as <Self as SharedValue>::Storage)
+            }
+
+            #[must_use]
+            fn modulo_prime_u128(input: u128) -> Self
+            where
+                Self: U128Conversions,
+            {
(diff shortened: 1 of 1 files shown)


ipa-core/src/ff/prime_field.rs
```diff
@@ -506,6 +560,27 @@ mod fp61bit {
             let y = Fp61BitPrime::truncate_from((u64::MAX >> 3) - 1); // PRIME - 1
             assert_eq!(x + y, Fp61BitPrime::truncate_from((u64::MAX >> 3) - 2));
         }
+
+        proptest! {
```
andyleiserson, line 564: Test subtraction here too?

ipa-core/src/ff/prime_field.rs
```diff
@@ -454,10 +478,40 @@ mod fp61bit {
...
+            Self: U128Conversions,
+        {
+            const PRIME: u128 = Fp61BitPrime::PRIME as u128;
+            debug_assert_eq!(0, PRIME & (PRIME + 1), "{PRIME} is not a Mersenne prime");
+
+            let val = (val & PRIME) + (val >> Self::BITS);
+            // another round if val ended up being greater than PRIME
+            let val = (val & PRIME) + (val >> Self::BITS);
+            if val == PRIME {
+                Self::ZERO
```
andyleiserson, line 504: It may be possible to save a little more if we can reduce some of the operations here from `u128` to `u64`. I guess to do that we may need to figure out how to get rustc/llvm to infer an add-with-carry.

The other win will be if we can get `LagrangeTable::eval` (and possibly other places) to do several multiplies before each modulo reduction.

andyleiserson (approved):

akoshelev:
Draft shows 25% improvement: https://draft-mpc.vercel.app/query/view/astir-fang2024-11-04T2255

8 hours before (from #1391)
6 hours with this change

ipa-core/src/ff/prime_field.rs
akoshelev: lets talk about it in our next 1:1 - sounds very promising

PR #1404, opened 2024-11-05
Vectorize the computation of MAC tags for malicious shuffle 

andyleiserson:
On 2M records, good for about 10%, 17:30 → 15:45. https://draft-mpc.vercel.app/query/view/bushy-due2024-11-05T2234

This is on top of the `Shuffleable` PRs.

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/mod.rs
@@ -374,7 +374,7 @@ where
 
     let shuffled = ctx
         .narrow(&Step::Shuffle)
-        .shuffle(padded_input_rows)
+        .shuffle(&padded_input_rows)
         .instrument(info_span!("shuffle_inputs"))
         .await?;
     let mut prfd_inputs = compute_prf_for_inputs(ctx.clone(), &shuffled).await?;
--- a/ipa-core/src/protocol/ipa_prf/mod.rs
+++ b/ipa-core/src/protocol/ipa_prf/mod.rs
@@ -374,7 +374,7 @@ where
 
     let shuffled = ctx
         .narrow(&Step::Shuffle)
-        .shuffle(&padded_input_rows)
+        .shuffle(padded_input_rows)
         .instrument(info_span!("shuffle_inputs"))
         .await?;
     let mut prfd_inputs = compute_prf_for_inputs(ctx.clone(), &shuffled).await?;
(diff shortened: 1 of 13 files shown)


andyleiserson:
Currently, this changes the shuffle traits to take `&[S]` instead of `I: IntoIterator`. I am not sure what would be best here, since the non-malicious shuffles still want an owning iterator and passing a slice may keep the inputs in memory longer than needed. Possibly just passing the `Vec` into the shuffle?

ipa-core/src/protocol/ipa_prf/mod.rs
```diff
@@ -202,6 +205,92 @@ where
...
+    }
+}
+
+impl<BK, TV, TS> shuffle::Shuffleable for OPRFIPAInputRow<BK, TV, TS>
+where
+    BK: BooleanArray,
+    TV: BooleanArray,
+    TS: BooleanArray,
+{
+    type Share = BA112;
```
akoshelev, line 258: this only works for specific TV and TS, is there a place for at least a runtime check? Or implementing it only for `OPRFIPAInputRow<BA8, BA3, BA20>`?

ipa-core/src/protocol/ipa_prf/shuffle/sharded.rs
```diff
@@ -233,15 +265,79 @@ impl<V: BooleanArray + FromRandom> Shuffleable for AdditiveShare<V> {
...
 
-impl MaliciousShuffleable for AdditiveShare<BA32> {
-    type ShareAndTag = BA64;
+/// Trait for a share of an input to malicious shuffle.
+///
+/// This trait should be implemented using the `impl_malicious_shuffleable!` macro,
+/// which will check the size of the `ShareAndTag` type and compute `TAG_OFFSET`
+/// automatically.
+pub trait MaliciousShuffleShare: TryInto<Vec<Gf32Bit>, Error = LengthError> {
+    type ShareAndTag: ShuffleShare + SharedValue;
```
akoshelev, line 312: At some point it may be worth dropping `SharedValue` bound. It seems to me that we need it to be able to hold `AdditiveShare<Share+Tag>` to pass it down to semi-honest shuffle, but it is not really needed. We could just have our own container that is not an additive share and implement Shuffleable for it.

AdditiveShare carries luggage we don't need to bring into shuffle imo - we don't need complex arithmetic and many other bounds that it puts on us

akoshelev (approved):
The size of this PR indicates a herculean effort to refactor both shuffles. The code is much more structured now - I think we can merge it and improve if needed.

akoshelev:
> Possibly just passing the Vec into the shuffle?
Yea I think that is definitely preferred

ipa-core/src/protocol/ipa_prf/mod.rs
andyleiserson: It would have failed at runtime within the packing routines, but since that may not give a clear message, I'll add a debug assertion that the share type is big enough. (I pushed this to #1397 since it goes with those changes.)

Since we support two `BK` types, we'd need `Shuffleable` implementations for two sets of type parameter values for each of the shuffle input types. Which is doable with macros, and I think we'd need to go that route if we wanted a compile-time check.

PR #1408, opened 2024-11-06
add oprf for hybrid; tests not yet compiling

eriktaubeneck:
This adds the conversion of the `match_key` shares to an `oprf` value for the Hybrid protocol.


ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,255 @@
...
+                .try_into()
+                .expect("Expected exactly 3 elements");
+
+        #[allow(clippy::large_futures)]
+        let results = flatten3v(chunked_reports.into_iter().zip(contexts).map(
+            |(reports_by_helper, helper_ctxs)| {
+                reports_by_helper
+                    .into_iter()
+                    .zip(helper_ctxs)
+                    .map(|(reports, ctx)| compute_prf_for_inputs(ctx, &reports))
```
eriktaubeneck, line 247: this gives me a compile error:
```
error[E0515]: cannot return value referencing local variable `reports`
   --> ipa-core/src/protocol/hybrid/oprf.rs:247:43
    |
247 |                     .map(|(reports, ctx)| compute_prf_for_inputs(ctx, &reports))
    |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^--------^
    |                                           |                           |
    |                                           |                           `reports` is borrowed here
    |                                           returns a value referencing data owned by the current function
```
I believe this has to do with `reports` being created in the `map` closure, but I'm not sure how to fix it. Cloning doesn't seem to help, and I'm not sure how else to potentially restructure this.

eriktaubeneck: rewriting this as a loop may work here

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,262 @@
...
+                .try_into()
+                .expect("Expected exactly 3 elements");
+
+        // #[allow(clippy::large_futures)]
+        // let results = flatten3v(chunked_reports.into_iter().zip(contexts).map(
+        //     |(reports_by_helper, helper_ctxs)| {
+        //         reports_by_helper
+        //             .into_iter()
+        //             .zip(helper_ctxs)
+        //             .map(|(reports, ctx)| compute_prf_for_inputs(ctx, &reports))
```
eriktaubeneck, line 247: this gives me a compile error:
```
error[E0515]: cannot return value referencing local variable `reports`
   --> ipa-core/src/protocol/hybrid/oprf.rs:247:43
    |
247 |                     .map(|(reports, ctx)| compute_prf_for_inputs(ctx, &reports))
    |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^--------^
    |                                           |                           |
    |                                           |                           `reports` is borrowed here
    |                                           returns a value referencing data owned by the current function
 ```

Comment (akoshelev) [reply ipa-core/src/protocol/hybrid/oprf.rs:247]: you may need `async move { compute_prf_for_inputs(ctx, &reports).await }`  here - not 100% sure if it is going to help though

Comment (eriktaubeneck) [reply ipa-core/src/protocol/hybrid/oprf.rs:247]: i think that was it. fingers crossed

File: ipa-core/src/protocol/hybrid/oprf.rs [hunk 3]
```diff
@@ -0,0 +1,294 @@
...
+        //     .map(|(reports_by_helper, helper_ctxs)| {
+        //         reports_by_helper
+        //             .into_iter()
+        //             .zip(helper_ctxs)
+        //             .map(|(reports, ctx)| async move {
+        //                 compute_prf_for_inputs(ctx, &reports).await.unwrap()
+        //             })
+        //             .collect::<Vec<_>>()
+        //     })
+        //     .collect::<Vec<_>>();
```
eriktaubeneck, line 255: I can't seem to get this to work. the loop approach below seems to work. it's a bit less ergonomic, but hopefully it's acceptable for a test.

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,294 @@
...
+        // remove before merging
+        println!("{:?}", prfs);
+
+        // check to make sure each helper has the same PRF values
+        assert!(prfs.iter().all(|x| *x == prfs[0]));
+
+        // check to make sure the reports with the same match keys have the same prf values
+        for (i, j) in indices_to_compare {
+            assert_eq!(prfs[0][i], prfs[0][j]);
+        }
```
eriktaubeneck, line 292: this is currently failing for the case where match keys are the same, but the OPRF is computed on different shards.

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,313 @@
...
+                .await?;
+        }
+        Ok(leader_prf_key)
+    } else {
+        let mut receiver = prf_key_ctx
+            .shard_recv_channel::<Replicated<Fp25519>>(ShardIndex::FIRST)
+            .take(1);
+        receiver.next().await.unwrap()
+    }
+    .unwrap();
```
eriktaubeneck, line 154: I couldn't seem to get this to work with `?` instead of the `unwrap()`s here, which is not great. any suggestions on this would be helpful.

ipa-core/src/protocol/hybrid/oprf.rs
eriktaubeneck: I'll delete this if we're ok with the approach below.

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,317 @@
...
+            //     .map(|(reports_by_helper, helper_ctxs)| {
+            //         reports_by_helper
+            //             .into_iter()
+            //             .zip(helper_ctxs)
+            //             .map(|(reports, ctx)| async move {
+            //                 compute_prf_for_inputs(ctx, &reports).await.unwrap()
+            //             })
+            //             .collect::<Vec<_>>()
+            //     })
+            //     .collect::<Vec<_>>();
```
eriktaubeneck, line 280: I can't seem to get this to work. the loop approach below seems to work. it's a bit less ergonomic, but hopefully the loop acceptable for a test. happy to try any suggestions, otherwise I'll delete this.

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -0,0 +1,317 @@
...
+            PRFIndistinguishableHybridReport {
+                prf_of_match_key,
+                value: value.clone(),
+                breakdown_key: breakdown_key.clone(),
+            }
+        })
+        .collect())
+}
+
+#[cfg(all(test, not(feature = "shuttle")))]
```
eriktaubeneck, line 192: in order to get the tests to work with `multi-threading`, I needed to use `test_executor::run`, and it seems that's [incompatible with the `shuttle` feature](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/lib.rs#L292). 

can someone confirm this is ok? I don't want to ignore a real issue.

cberkhoff: Don't think run is incompatible. If you scroll up on that link you will see there's another implementation for run with shuttle.

Here are some tests you can check that also use run
https://github.com/private-attribution/ipa/blob/main/ipa-core/src/query/processor.rs#L623

eriktaubeneck: thanks for the code pointer. let me see if I can get this running. it may be failing for a different reason

cberkhoff (commented):
By the way. Is all this code brand new? If not, it would be more efficient for reviewers to have 2 PRs. One with the important changes and a separate one where you just move things around.

If it is. Are we duplicating too many things?

eriktaubeneck:
> By the way. Is all this code brand new? If not, it would be more efficient for reviewers to have 2 PRs. One with the important changes and a separate one where you just move things around.
> 
> If it is. Are we duplicating too many things?

Quite a bit of this is duplicated from the OPRF from the old protocol. We've been taking the approach of not touching the old protocol code, but minimizing our reliance on it. If we wanted to support it long term, we could update the OPRF protocol to take a more generic approach (e.g. have a trait like `OPRFable`), but we've explicitly decided not to try and support both in the long run.

We will do clean up of the old protocol, but until we're complete with the in-market test, we want to keep it around just in case we need to fix a bug.

cberkhoff:
Can we get this at least as 2 separate PRs? One with the copy and another with the modification?

eriktaubeneck:
Closing this in favor of #1412 and #1413

PR #1409, opened 2024-11-07
Lagrange evaluation performance improvements

akoshelev:
rustc didn't optimize the `eval` function and ended up doing
131072*32*992 loop iteration per Lagrange compute.

For some unknown reason to me, optimizer works only if we operate
on integers, not on fields. Maybe modulo reduction is to blame

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Borrow, fmt::Debug};
+use std::fmt::Debug;
 
 use typenum::Unsigned;
 
@@ -79,8 +79,7 @@ pub struct LagrangeTable<F: Field, const N: usize, const M: usize> {
 
 impl<F, const N: usize> LagrangeTable<F, N, 1>
 where
-    F: Field + TryFrom<u128>,
-    <F as TryFrom<u128>>::Error: Debug,
+    F: PrimeField,
 {
     /// generates a `CanonicalLagrangeTable` from `CanoncialLagrangeDenominators` for a single output point
     /// The "x coordinate" of the output point is `x_output`.
@@ -95,25 +94,16 @@ where
 
 impl<F, const N: usize, const M: usize> LagrangeTable<F, N, M>
 where
-    F: Field,
+    F: PrimeField,
 {
     /// This function uses the `LagrangeTable` to evaluate `polynomial` on the _output_ "x coordinates"
     /// that were used to generate this table.
     /// It is assumed that the `y_coordinates` provided to this function correspond the values of the _input_ "x coordinates"
     /// that were used to generate this table.
-    pub fn eval(&self, y_coordinates: &[F; N]) -> [F; M]
-    {
+    pub fn eval(&self, y_coordinates: &[F; N]) -> [F; M] {
         self.table
-            .iter()
-            .map(|table_row| {
-                table_row
-                    .iter()
-                    .zip(y_coordinates)
-                    .fold(F::ZERO, |acc, (&base, y)| acc + base * (*y.borrow()))
-            })
-            .collect::<Vec<F>>()
-            .try_into()
-            .unwrap()
+            .each_ref()
+            .map(|row| dot_product(row, y_coordinates))
     }
 
     /// helper function to compute a single row of `LagrangeTable`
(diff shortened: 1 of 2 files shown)


akoshelev:
Maybe we can take this PR as is and improve later. It gives another 50% boost: 

[4hrs](https://draft-mpc.vercel.app/query/view/meaty-pasta2024-11-07T0154) vs [6hrs](https://draft-mpc.vercel.app/query/view/astir-fang2024-11-04T2255)

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -176,6 +160,29 @@ where
...
+
+    let mut sum = 0;
+
+    // I am cautious about using zip in hot code
+    // https://github.com/rust-lang/rust/issues/103555
+    for i in 0..N {
+        sum += a[i].as_u128() * b[i].as_u128();
+    }
+
+    F::truncate_from(sum)
```
tyurek, line 183: How does this work? It looks like truncate_from() uses the argument as a seed to map to a pseudorandom field element. Are there correctness tests for this?

akoshelev: that would be very bad - this code just defers modulo reduction operation till the very end: https://github.com/private-attribution/ipa/blob/bef33119b463c3cbd34f82a962f00cc3063055d0/ipa-core/src/ff/prime_field.rs#L130

akoshelev: hmm, can you please point me to `Fp25519::truncate_from`? I couldn't find it and it would be also bad if we had `truncate_from` implementation for it like you described.

>is it possible to use try_from instead

`TryFrom` wouldn't work here because it will reject values > PRIME: https://github.com/private-attribution/ipa/blob/bef33119b463c3cbd34f82a962f00cc3063055d0/ipa-core/src/ff/prime_field.rs#L226

tyurek: If this line is always doing a modulo reduction to get back to the field then that's fine. Truncate to me implies dropping bits, but is this just how the modulo-reduce-into-field function is named?

tyurek: The function I was looking at was this one: https://github.com/akoshelev/raw-ipa/blob/8bb84e5ae9acfd529e482d8663c44ea96e64ea27/src/ff/ec_prime_field.rs#L176

akoshelev: I see - this looks like old code, we don't have it in main: https://github.com/private-attribution/ipa/blob/main/ipa-core/src/ff/ec_prime_field.rs

akoshelev: > Truncate to me implies dropping bits, but is this just how the modulo-reduce-into-field function is named?
Agree, this got very confusing over time. Originally it was dropping bits but we changed it many times and here we are.

Let me open an issue on this one, I agree that naming is important

akoshelev: https://github.com/private-attribution/ipa/issues/1421

tyurek (approved):
Looks good!

ipa-core/src/protocol/ipa_prf/malicious_security/lagrange.rs
```diff
@@ -176,6 +160,29 @@ where
...
+/// Computes the dot product of two arrays of the same size.
+/// It is isolated from Lagrange because there could be potential SIMD optimizations used
+fn dot_product<F: PrimeField, const N: usize>(a: &[F; N], b: &[F; N]) -> F {
+    // Staying in integers allows rustc to optimize this code properly
+    // with any reasonable N, we won't run into overflow with dot product.
+    // (N can be as large as 2^32 and still no chance of overflow for 61 bit prime fields)
+    debug_assert!(
+        F::PRIME.into() < (1 << 64),
+        "The prime {} is too large for this dot product implementation",
+        F::PRIME.into()
```
andyleiserson, line 172: We are multiplying two field values and then summing. I'd say that the conservative requirement is `2 * F::BITS + N.next_power_of_two().ilog2() <= u128::BITS`.

akoshelev: that's right, I think this assertion just assumes `N` being reasonable and the second argument of this expression negligible compared to `F::BITS`. I don't mind changing it, the reasoning behind this was just to optimize for readers

andyleiserson (approved):

PR #1410, opened 2024-11-07
Support cross-shard PRSS in sharded MPC circuits (in memory only for now)

akoshelev:
We ran into an issue where shards need coordinated randomness in more than one protocol:

* Malicious shuffle needs it to generate MAC keys
* PRF needs it to generate PRF key to mask the values

At this point I strongly suspect we will need it in more places and the current mitigation is not sustainable. We do simulate key exchange in shuffle and about to in PRF, so I decided to stop the bleeding and implemented coordinated randomness for PRSS.

The only API difference between those two is to use cross-shard PRSS, one would call `ShardedContext::cross_shard_prss` and shard-scoped PRSS is still available by calling `ShardedContext::prss`. Both return the same structure, so anything that can be generated with shard-scoped PRSS, can also be generated with cross-shard randomness. It is basically a drop-in replacement, only available for sharded MPC circuits.

Note that this only brings in-memory implementation, but I have implemented 50% of distribution circuit already and it looks pretty straightforward - just generate a key using PRSS and distribute it across shards - it will be used to bootstrap generators on every shard. This will be a separate PR and will take some time to implement - but folks shouldn't be blocked as this change unblocks using the in-memory infrastructure.


ipa-core/src/test_fixture/shard_configurator.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::{
+    helpers::Role,
+    protocol::prss::Endpoint as PrssEndpoint,
+    sharding::{NotSharded, ShardBinding, ShardIndex, Sharded},
+    test_fixture::{make_participants, world::ShardingScheme, Distribute, WithShards},
+};
+
+/// This trait serves the purpose of setting up shard contexts. Each shard shares some
+/// global state with others (cross-shard PRSS) and also owns its own state (per-shard PRSS).
+/// This construction allows [`ShardWorld`] to
```
eriktaubeneck, line 15: did the end of this get cut off?

ipa-core/src/test_fixture/world.rs
```diff
@@ -1329,4 +1317,45 @@ mod tests {
             assert_eq!(first_result, second_result);
         });
     }
+
+    #[test]
+    fn cross_shard_prss() {
```
eriktaubeneck, line 1322: is it worth also adding a test to ensure that regular PRSS is *not* consistent across shards?

ipa-core/src/protocol/context/mod.rs
```diff
@@ -211,6 +211,14 @@ impl ShardedContext for Base<'_, Sharded> {
             .gateway
             .get_shard_receiver(&ChannelId::new(origin, self.gate.clone()))
     }
+
+    fn cross_shard_prss(&self) -> InstrumentedIndexedSharedRandomness<'_> {
```
eriktaubeneck, line 215: would it make sense to implement this on non-sharded contexts, so that a protocol can use this regardless of sharding (in the non-sharded case, it just acts like normal prss)?

eriktaubeneck (approved):
a couple questions on functionality and testing, but overall this looks great.

ipa-core/src/protocol/context/mod.rs
akoshelev: I think this is where Rust type system plays nicely for us - in non-sharded world you should never require cross-shard PRSS and use regular PRSS instead. All sharded circuits we write operate over `ShardedContext` that will make this API available. Non-sharded MPC won't compile if it tries to use it, which is what we would want imo

ipa-core/src/test_fixture/shard_configurator.rs
akoshelev: ah, missed that one!

ipa-core/src/test_fixture/world.rs
akoshelev: Great suggestion, we do already have it. It was implemented with in-memory sharding. Check `unique_prss_per_shard` test in this file

ipa-core/src/protocol/context/mod.rs
eriktaubeneck: got it. so for example with [`gen_prf_key`](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/protocol/ipa_prf/prf_eval.rs#L76) we'd just have two different implementations, one with `Upgradable` constraint on `C`, and another with `Upgradable + Sharded`? (just to make sure I'm understanding correctly.)

akoshelev: yea, we will delete non-sharded version anyway, so the only one that we will use is `gen_prf_key` with `C: UpgradableContext + ShardedContext` bound. That will provide access to `cross_shard_prss`

akoshelev:
I've discussed this PR with @andyleiserson  - our conclusion is to merge it to unblock work on hybrid, but there are some things we want to investigate in #1418

PR #1412, opened 2024-11-07
copy ipa oprf code over for use with hybrid

eriktaubeneck:
As requested [here](https://github.com/private-attribution/ipa/pull/1408#issuecomment-2462937702), this is the first of 2 PRs, splitting up #1408, adding the OPRF step to the hybrid protocol.

This only copies over the [IPA OPRF code ](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/protocol/ipa_prf/mod.rs), without modifying it yet. Originally, it was in the `protocol/ipa_prf/mod.rs` file, which contained other stuff as well, so this isn't a complete copy of that file.

cc @cberkhoff

### DIFF
--- a/ipa-core/src/protocol/hybrid/mod.rs
+++ b/ipa-core/src/protocol/hybrid/mod.rs
@@ -1,3 +1,4 @@
+pub(crate) mod oprf;
 pub(crate) mod step;
 
 use step::HybridStep as Step;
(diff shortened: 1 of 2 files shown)


cberkhoff (approved):
Thanks for separating the copy

PR #1413, opened 2024-11-07
Hybrid oprf pt2

eriktaubeneck:
This is stacked on top of #1412, so to look at the just the diff, make sure to ignore the `d1199d1` commit in the diff viewer.

This updates the OPRF function to work with hybrid reports, as well as negotiating a key exchange across shards. Note that that will be replaced by the functionality in #1410 once merged.

It also adds a direct test for this function, which was lacking before.

### DIFF
--- a/ipa-core/src/protocol/hybrid/oprf.rs
+++ b/ipa-core/src/protocol/hybrid/oprf.rs
@@ -258,21 +258,6 @@ mod test {
                     .try_into()
                     .expect("Expected exactly 3 elements");
 
-            // #[allow(clippy::large_futures)]
-            // let results = chunked_reports
-            //     .into_iter()
-            //     .zip(contexts)
-            //     .map(|(reports_by_helper, helper_ctxs)| {
-            //         reports_by_helper
-            //             .into_iter()
-            //             .zip(helper_ctxs)
-            //             .map(|(reports, ctx)| async move {
-            //                 compute_prf_for_inputs(ctx, &reports).await.unwrap()
-            //             })
-            //             .collect::<Vec<_>>()
-            //     })
-            //     .collect::<Vec<_>>();
-
             let mut results = Vec::new();
             for (reports_by_helper, helper_ctxs) in chunked_reports.into_iter().zip(contexts) {
                 for (reports, ctx) in reports_by_helper.into_iter().zip(helper_ctxs) {
--- a/ipa-core/src/protocol/hybrid/oprf.rs
+++ b/ipa-core/src/protocol/hybrid/oprf.rs
@@ -25,9 +25,9 @@ use crate::{
         hybrid::step::HybridStep,
         ipa_prf::{
             boolean_ops::convert_to_fp25519,
-            prf_eval::{eval_dy_prf, gen_prf_key, PrfSharing},
+            prf_eval::{eval_dy_prf, PrfSharing},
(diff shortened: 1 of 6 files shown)


ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -122,9 +131,31 @@ where
...
+                .await?;
+        }
+        Ok(leader_prf_key)
+    } else {
+        let mut receiver = prf_key_ctx
+            .shard_recv_channel::<Replicated<Fp25519>>(ShardIndex::FIRST)
+            .take(1);
+        receiver.next().await.unwrap()
+    }
+    .unwrap();
```
eriktaubeneck, line 155: I couldn't seem to get this to work with ? instead of the unwrap()s here, which is not great. any suggestions on this would be helpful.

that said, this will get replaced with #1410, so maybe work either just waiting on that, or accepting this until it can be replaced.

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -144,22 +175,148 @@ where
...
+            //     .map(|(reports_by_helper, helper_ctxs)| {
+            //         reports_by_helper
+            //             .into_iter()
+            //             .zip(helper_ctxs)
+            //             .map(|(reports, ctx)| async move {
+            //                 compute_prf_for_inputs(ctx, &reports).await.unwrap()
+            //             })
+            //             .collect::<Vec<_>>()
+            //     })
+            //     .collect::<Vec<_>>();
```
eriktaubeneck, line 285: I can't seem to get this to work. the loop approach below seems to work. it's a bit less ergonomic, but hopefully the loop acceptable for a test. happy to try any suggestions, otherwise I'll delete this.

ipa-core/src/protocol/hybrid/oprf.rs
akoshelev: #1410 is merged so we can replace this I think

ipa-core/src/protocol/hybrid/oprf.rs
akoshelev: this test is really doing 70% of what `world.semi_honest()` already does. It is recommended to use it as it takes care of sharing and reconstructing the outputs. https://github.com/private-attribution/ipa/blob/277fbeb8332128aa52c58d16896fd377757c286a/ipa-core/src/protocol/ipa_prf/prf_eval.rs#L304 maybe a good place to start

akoshelev (approved):

ipa-core/src/protocol/hybrid/oprf.rs
eriktaubeneck: If I remember correctly, there we're two issues with that approach:
1. it wasn't clear if/how to work with sharded inputs
2. the prf isn't reconstructed

akoshelev: it does support sharding https://github.com/private-attribution/ipa/blob/f587a37e6e9bd9b9eb8580cea9d6ea4914bda2f9/ipa-core/src/protocol/ipa_prf/shuffle/sharded.rs#L590

If you don't need to reconstruct, that makes it easier - just return it and you'll get 3 vectors of PRF values back

PR #1414, opened 2024-11-07
Sharded Prepare Query API

cberkhoff:
Added `processor::prepare_helper` and `processor::prepare_shard` to Processor and corresponding tests to validate the correct behavior. Had to add functionality in `InMemoryShardNetwork` to be able to set custom Handlers. Since the amount of testing code in `processor` was getting out of hand I created a few structs to help streamlining them.

Following is a description of the overall approach we want to take for Sharded APIs. 

## Sharding Query Workflow

Prepare Query Diagram

Query APIs (Create Query, Query Status, etc), the Report Collector (RC) will reach out to the Leader Shards **only**. The diagram above shows how Shard 1-0 receives a Create Query from RC (`processor::new_query`). He validates its own state and then forwards the calls to the other leaders; Shard 2-0 and Shard 3-0, using the `processor::prepare_helper` (blue arrows in the diagram). So far this is the same operating mode as now, but in a sharded environment, each leader will subsequently call all his shard's `processor::prepare_shard` (red arrows in diagram). If all shards in the network are Ok, then an Ok is sent back to RC by the Leader. To the RC, this is a synchronous HTTP call.

The following doesn't involve this PR but is helpful to understand how PrepareQuery fits with the rest of the APIs. 

Similarly, while the query is running, RC will call Query Status on all leaders; Shard 1-0, Shard 2-0 and Shard 3-0, which each returns a unified response for its Helper (same as Prepare Query). Query Results works in the same way, aggregating all responses for a Helper.

The only odd case is Query Input, in which the RC will individually push the input to each Shard in all helpers. This unwanted exposure of Shards can be mitigated by implementing a pull model for inputs, but will be left as future work.

### DIFF
--- a/ipa-core/src/query/processor.rs
+++ b/ipa-core/src/query/processor.rs
@@ -3,7 +3,7 @@ use std::{
     fmt::{Debug, Formatter},
 };
 
-use futures::{future::try_join, stream, TryFutureExt};
+use futures::{future::try_join, stream};
 use serde::Serialize;
 
 use crate::{
@@ -11,8 +11,8 @@ use crate::{
     executor::IpaRuntime,
     helpers::{
         query::{PrepareQuery, QueryConfig, QueryInput},
-        Gateway, GatewayConfig, MpcTransportError, MpcTransportImpl, Role, RoleAssignment,
-        ShardTransportError, ShardTransportImpl, Transport,
+        BroadcastError, Gateway, GatewayConfig, MpcTransportError, MpcTransportImpl, Role,
+        RoleAssignment, ShardTransportError, ShardTransportImpl, Transport,
     },
     hpke::{KeyRegistry, PrivateKeyOnly},
     protocol::QueryId,
@@ -68,7 +68,7 @@ pub enum NewQueryError {
     #[error(transparent)]
     MpcTransport(#[from] MpcTransportError),
     #[error(transparent)]
-    ShardTransport(#[from] ShardTransportError),
+    ShardBroadcastError(#[from] BroadcastError<ShardIndex, ShardTransportError>),
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -83,7 +83,7 @@ pub enum PrepareQueryError {
         source: StateError,
     },
     #[error(transparent)]
-    ShardTransport(#[from] ShardTransportError),
+    ShardBroadcastError(#[from] BroadcastError<ShardIndex, ShardTransportError>),
 }
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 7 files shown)


ipa-core/src/query/processor.rs
```diff
@@ -169,44 +175,51 @@ impl Processor {
...
-        // Inform other parties about new query. If any of them rejects it, this join will fail
+        // Inform other helpers about new query. If any of them rejects it, this join will fail
         try_join(
             transport.send(left, prepare_request.clone(), stream::empty()),
             transport.send(right, prepare_request.clone(), stream::empty()),
         )
         .await
         .map_err(NewQueryError::MpcTransport)?;
 
+        shard_transport.broadcast(prepare_request.clone()).await?;
```
akoshelev, line 186: I think we should first align our own shards and confirm that they are ready to execute before talking to other helpers. This code does not have a path to recover or a path to inform other helpers that query cannot move forward because one or more shards failed to confirm the query request

ipa-core/src/query/processor.rs
```diff
@@ -169,44 +175,51 @@ impl Processor {
...
-        let my_role = req.roles.role(transport.identity());
+        let my_role = req.roles.role(mpc_transport.identity());
+        let shard_index = shard_transport.identity();
 
         if my_role == Role::H1 {
             return Err(PrepareQueryError::WrongTarget);
         }
+        if shard_index != ShardIndex::FIRST {
+            return Err(PrepareQueryError::WrongTarget);
+        }
```
akoshelev, line 215:
```suggestion
        if my_role == Role::H1 || shard_index != ShardIndex::FIRST {
            return Err(PrepareQueryError::WrongTarget);
        }        
```

although it may be better to use a different error message when someone submits prepare request to wrong shard

ipa-core/src/query/processor.rs
```diff
@@ -216,7 +229,29 @@ impl Processor {
         Ok(())
     }
 
-    /// Receive inputs for the specified query. That triggers query processing
+    /// On prepare, each shard:
+    /// * ensures that it is not the leader on this query
```
akoshelev, line 233: how does this requirement get enforced in the code?

ipa-core/src/query/processor.rs
```diff
@@ -530,57 +698,83 @@ mod tests {
...
 
         #[tokio::test]
         async fn rejects_if_query_exists() {
-            let network = InMemoryMpcNetwork::default();
-            let identities = HelperIdentity::make_three();
-            let req = prepare_query(identities);
-            let transport = network.transport(identities[1]);
-            let processor = Processor::default();
-            processor.prepare(&transport, req.clone()).unwrap();
+            let req = prepare_query();
```
akoshelev, line 758: can you write a test that ensures if any follower shard on H1 rejects the request, we end up in a clean state on all helpers and all shards?

akoshelev (changes requested):
this looks promising and API looks exactly how I would expect it to work. I just think we need to work out more on error cases before pushing this

ipa-core/src/query/processor.rs
cberkhoff: I'll add a separate type. I rather contribute with the error granularity.

PR #1416, opened 2024-11-07
[Bugfix] Add cfg(debug_assertions) to dependent CircularBuf tests

tyurek:
Multiple tests for CircularBuf rely on matching a panic initiated by `debug_assert!`. However, these tests will fail when we're not in debug mode (such as when running `cargo test -r`), so I added specifiers to the tests that rely on `debug_assert!`.

Alternatively, we could replace the `debug_assert!`s with `assert!`s. I don't have a preference, but we should do one of the two


akoshelev:
annotating tests with debug_assertions is fine by me, just note that this may be an uphill battle because we always run them in debug mode. They are written to test correctness 98% of the time. For release/optimizations we use benchmarks

akoshelev (approved):

PR #1419, opened 2024-11-08
[IGNORE] Performance test of Lagrange and ConvertValues PRs

akoshelev:
#1409 and #1411 together


akoshelev:
3.5hrs vs 4hrs before that: https://draft-mpc.vercel.app/query/view/plane-arbor2024-11-08T1740

PR #1420, opened 2024-11-08
Reshard by oprf

eriktaubeneck:
This is currently a draft, as it's stacked on top of #1413. For review, you'll want to make sure to just look at the commits after `4db9c544`.

This implements `Serializable` for `PRFIndistinguishableHybridReport<BA8, BA3>`. Offline @akoshelev and I discussed this, and instead of dealing with the complicated trait bounds to implement `Serializable` generically for and `BK, V`, we'd only implement it for the types we actually use. If at some point we decide to use this for a large set of replicated types for breakdown keys and values, we can look into other approaches to implement this more efficiently, like a macro.

### DIFF
--- a/ipa-core/src/protocol/hybrid/mod.rs
+++ b/ipa-core/src/protocol/hybrid/mod.rs
@@ -1,16 +1,18 @@
 pub(crate) mod oprf;
 pub(crate) mod step;
 
+use oprf::PRFIndistinguishableHybridReport;
+
 use crate::{
     error::Error,
     ff::{
         boolean::Boolean, boolean_array::BooleanArray, curve_points::RP25519,
-        ec_prime_field::Fp25519, U128Conversions,
+        ec_prime_field::Fp25519, Serializable, U128Conversions,
     },
     helpers::query::DpMechanism,
     protocol::{
         basics::{BooleanProtocols, Reveal},
-        context::{DZKPUpgraded, MacUpgraded, ShardedContext, UpgradableContext},
+        context::{reshard_iter, DZKPUpgraded, MacUpgraded, ShardedContext, UpgradableContext},
         hybrid::{
             oprf::{compute_prf_for_inputs, BreakdownKey, CONV_CHUNK, PRF_CHUNK},
             step::HybridStep as Step,
@@ -67,6 +69,7 @@ where
         PrfSharing<MacUpgraded<C, Fp25519>, PRF_CHUNK, Field = Fp25519> + FromPrss,
     Replicated<RP25519, PRF_CHUNK>:
         Reveal<MacUpgraded<C, Fp25519>, Output = <RP25519 as Vectorizable<PRF_CHUNK>>::Array>,
+    PRFIndistinguishableHybridReport<BK, V>: Serializable,
 {
     if input_rows.is_empty() {
         return Ok(vec![Replicated::ZERO; B]);
@@ -83,7 +86,14 @@ where
     // TODO shuffle input rows
     let shuffled_input_rows = padded_input_rows;
 
-    let _prf_input_rows = compute_prf_for_inputs(ctx.clone(), &shuffled_input_rows).await?;
+    let prf_input_rows_stream = compute_prf_for_inputs(ctx.clone(), &shuffled_input_rows).await?;
+
+    let _sharded_prf_rows = reshard_iter(
+        ctx.narrow(&Step::ReshardByPrf),
+        prf_input_rows_stream,
+        |ctx, _, report| report.shard_picker(ctx.shard_count()),
+    )
+    .await?;
 
     unimplemented!("protocol::hybrid::hybrid_protocol is not fully implemented")
 }
--- a/ipa-core/src/protocol/hybrid/oprf.rs
+++ b/ipa-core/src/protocol/hybrid/oprf.rs
@@ -1,16 +1,17 @@
 use std::iter::zip;
 
 use futures::{stream, StreamExt, TryStreamExt};
-use typenum::Const;
+use generic_array::GenericArray;
+use typenum::{Const, Unsigned, U12};
 
 use crate::{
     error::{Error, UnwrapInfallible},
     ff::{
         boolean::Boolean,
-        boolean_array::{BooleanArray, BA5, BA64, BA8},
+        boolean_array::{BooleanArray, BA3, BA5, BA64, BA8},
         curve_points::RP25519,
         ec_prime_field::Fp25519,
-        U128Conversions,
+        Serializable, U128Conversions,
     },
     helpers::{
         stream::{div_round_up, process_slice_by_chunks, Chunk, ChunkData, TryFlattenItersExt},
@@ -30,12 +31,13 @@ use crate::{
         prss::{FromPrss, SharedRandomness},
         RecordId,
     },
-    report::hybrid::IndistinguishableHybridReport,
+    report::hybrid::{IndistinguishableHybridReport, InvalidHybridReportError},
     secret_sharing::{
         replicated::semi_honest::AdditiveShare as Replicated, BitDecomposed, SharedValue,
         TransposeFrom, Vectorizable,
     },
     seq_join::seq_join,
+    sharding::ShardIndex,
 };
 
 // In theory, we could support (runtime-configured breakdown count) ≤ (compile-time breakdown count)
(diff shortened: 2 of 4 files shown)


ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -74,17 +78,82 @@ pub const PRF_CHUNK: usize = 16;
...
+    #[must_use]
+    pub fn shard_picker(&self, shard_count: ShardIndex) -> ShardIndex {
+        let shard_count = u64::from(shard_count);
+        ShardIndex::try_from(self.prf_of_match_key % shard_count)
+            .expect("Modulo a u32 will fit in a u32")
+    }
+}
+
+impl Serializable for PRFIndistinguishableHybridReport<BA8, BA3> {
+    type Size = U12;
```
akoshelev, line 105: 12 bytes to store 11 bits seems a bit excessive

ipa-core/src/report/hybrid.rs
```diff
@@ -673,7 +673,7 @@ where
 /// and `HybridImpressionReport`s so that they can be made indistingushable.
 /// Note: these need to be shuffled (and secret shares need to be rerandomized)
 /// to provide any formal indistinguishability.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Default)]
```
akoshelev, line 676: do you need it? I don't think providing `Default` implementation makes sense anywhere except tests, so `[cfg_attr(test, derive(Default))]` is a better option imo

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -74,17 +78,82 @@ pub const PRF_CHUNK: usize = 16;
...
+    type DeserializationError = InvalidHybridReportError;
+
+    fn serialize(&self, buf: &mut GenericArray<u8, Self::Size>) {
+        const PRF_MK_SZ: usize = 8;
```
akoshelev, line 109: nit: `std::mem:size_of_val(self.prf_of_match_key)` - not ideal, but at least won't break badly if it is no longer a u64. Note that it won't be a constant anymore, but it does not seem like a big deal to me

ipa-core/src/protocol/hybrid/oprf.rs
eriktaubeneck: BA3 and BA8 are both 2 bytes, it seems

ipa-core/src/report/hybrid.rs
eriktaubeneck: pretty sure it is. this is the compiler error I get without it
```
error[E0277]: the trait bound `IndistinguishableHybridReport<BK, V>: Default` is not satisfied
   --> ipa-core/src/protocol/hybrid/oprf.rs:177:21
    |
177 |       let curve_pts = seq_join(
    |  _____________________^
178 | |         ctx.active_work(),
179 | |         process_slice_by_chunks(input_rows, move |idx, records: ChunkData<_, CONV_CHUNK>| {
180 | |             let record_id = RecordId::from(idx);
...   |
189 | |     .map_ok(Chunk::unpack::<PRF_CHUNK>)
190 | |     .try_flatten_iters()
    | |________________________^ the trait `Default` is not implemented for `IndistinguishableHybridReport<BK, V>`
    ```

File: ipa-core/src/protocol/hybrid/oprf.rs (hunk 3 above)
Comment (eriktaubeneck) [reply ipa-core/src/protocol/hybrid/oprf.rs:109]: that works for `serialize`, but `self` isn't part of the function signature for `deserialize` :/

Comment (eriktaubeneck) [reply ipa-core/src/protocol/hybrid/oprf.rs:109]: also, if will still break if it's not a u64, or at least if it's bigger, as it won't fit into the `U12` anymore.

Comment (akoshelev) [reply ipa-core/src/protocol/hybrid/oprf.rs:109]: https://github.com/private-attribution/ipa/blob/f5d9ec786b0750fe73640a45ebd7249462ff1d5f/ipa-core/src/ff/mod.rs#L61

File: ipa-core/src/protocol/hybrid/oprf.rs (hunk 1 above)
Comment (eriktaubeneck) [reply ipa-core/src/protocol/hybrid/oprf.rs:105]: maybe I'm being dense here, but if the widths are
```
    const PRF_MK_SZ: usize = 8;
    const V_SZ: usize = <Replicated<BA3> as Serializable>::Size::USIZE;
    const BK_SZ: usize = <Replicated<BA8> as Serializable>::Size::USIZE;
```

then
```
        self.breakdown_key.serialize(GenericArray::from_mut_slice(
            &mut buf[Self::PRF_MK_SZ + Self::V_SZ..Self::PRF_MK_SZ + Self::V_SZ + Self::BK_SZ],
        ));
        ```

ipa-core/src/protocol/hybrid/oprf.rs
eriktaubeneck: I'm not sure I understand the relevance. The complication here is that `prf_of_match_key` is a `u64` in the clear, not any sort of secret sharing that already implements Serializable. 

I refactored this slightly so that it's less repetitive, and left a comment that the implementation relies on it being a `u64`.

akoshelev: I am just saying that the size is specified in bytes, not bits

ipa-core/src/protocol/hybrid/oprf.rs
akoshelev: Erik was right, I was wrong here

eriktaubeneck: there's a first for everything!

akoshelev: ah, not the first time for sure

ipa-core/src/report/hybrid.rs
akoshelev: hmm, something is off here - we shouldn't require to process those reports by chunks

akoshelev: nope, we really need it to generate curve points from match keys

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -293,4 +361,22 @@ mod test {
             }
         });
     }
+
+    #[test]
+    fn prf_indistinguishable_serialize_deserialize() {
+        run(|| async {
+            let world = TestWorld::default();
+            let mut rng = world.rng();
+            let report = PRFIndistinguishableHybridReport::<BA8, BA3> {
```
akoshelev, line 370: given how often you use it, it is probably worth implementing `impl rand::distributions::Distribution<PRFIndistinguishableHybridReport> for rand::distributions::Standard {
}` for tests

ipa-core/src/protocol/hybrid/mod.rs
```diff
@@ -83,7 +86,14 @@ where
     // TODO shuffle input rows
     let shuffled_input_rows = padded_input_rows;
 
-    let _prf_input_rows = compute_prf_for_inputs(ctx.clone(), &shuffled_input_rows).await?;
+    let prf_input_rows_stream = compute_prf_for_inputs(ctx.clone(), &shuffled_input_rows).await?;
+
+    let _sharded_prf_rows = reshard_iter(
+        ctx.narrow(&Step::ReshardByPrf),
+        prf_input_rows_stream,
+        |ctx, _, report| report.shard_picker(ctx.shard_count()),
```
akoshelev, line 94: is this what you want me to work on?

akoshelev (approved):

akoshelev (approved):

ipa-core/src/protocol/hybrid/mod.rs
eriktaubeneck: yes, though most the work is changing the function above `compute_prf_for_inputs` to return a TryStream instead of a Vec. at that point, we should be able to replace `reshard_iter` with `reshard_try_stream` here.

akoshelev: I looked into that and decided that changing `compute_prf_for_inputs` is better. Returning a stream that borrows from local variables is hard in Rust, so doing in-place resharding is better. Besides, this function is already async and returning a stream from already async function is a bit weird

akoshelev: #1425 should address that

PR #1423, opened 2024-11-08
Use constant-time comparisons

andyleiserson:
I suspect it would be quite difficult to construct a meaningful attack on one of these, but it still seems wise to use constant-time comparisons where best practice would dictate doing so.

### DIFF
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -1,6 +1,7 @@
 use std::{fmt::Display, mem};
 
 use generic_array::GenericArray;
+use subtle::{Choice, ConstantTimeEq};
 
 use super::Field;
 use crate::{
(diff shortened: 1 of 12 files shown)


ipa-core/src/protocol/basics/check_zero.rs
```diff
@@ -61,7 +63,7 @@ where
             .expect("full reveal should always return a value"),
     );
 
-    Ok(rv == F::ZERO)
+    Ok(rv.ct_eq(&F::ZERO).into())
```
andyleiserson, line 66: In `Fp32BitPrime` this isn't going to matter, but we're using MACs in `Fp25519` now.

ipa-core/src/ff/prime_field.rs
```diff
@@ -241,6 +242,15 @@ macro_rules! field_impl {
             }
         }
 
+        // If the field value fits in a machine word, a naive comparison should be fine.
+        // But this impl is important for `[T]`, and useful to document where a
+        // constant-time compare is intended.
+        impl ConstantTimeEq for $field {
```
tyurek, line 248: Do you think it makes sense to make the Eq implementation for the `field_impl` macro default constant time as well?

akoshelev:
we should measure the performance impact of this change before merging. I agree constant time comparisons are important, but we need to understand whether this is going to make things significantly worse

andyleiserson:
Draft run, 4M, before: https://draft-mpc.vercel.app/query/view/obese-oat2024-11-11T2252, 24:51
After: https://draft-mpc.vercel.app/query/view/gone-coke2024-11-11T2333, 24:48

akoshelev (approved):

andyleiserson:
I am going to merge this. If we want to use the constant time algorithm for all equality checks, we can do that -- although there's nothing special about equality checks at that point -- it becomes a general exercise of scrubbing everything for constant-time execution. (Which it probably should be, but that is not a small undertaking.)

PR #1425, opened 2024-11-09
Reshard Hybrid reports according to PRF values

akoshelev:
Doing it along with computing PRF values helps with latency as reports get out as soon as OPRF is computed and we don't need to wait until the end.

Note that this requires malicious security with abort to be implemented across shards - if one shard detects malicious behavior, all shards must abort

I also noticed that we use semi-honest execution which is wrong as we need malicious, but I couldn't fix it as part of this PR - there were trait bounds errors that require more attention

### DIFF
--- a/ipa-core/src/report/hybrid.rs
+++ b/ipa-core/src/report/hybrid.rs
@@ -33,12 +33,15 @@ use bytes::{Buf, BufMut, Bytes};
 use generic_array::{ArrayLength, GenericArray};
 use hpke::Serializable as _;
 use rand_core::{CryptoRng, RngCore};
-use typenum::{Sum, Unsigned, U16};
+use typenum::{Sum, Unsigned, U12, U16};
 
 use crate::{
     const_assert_eq,
     error::{BoxError, Error},
-    ff::{boolean_array::BA64, Serializable},
+    ff::{
+        boolean_array::{BA3, BA64, BA8},
+        Serializable,
+    },
     hpke::{
         open_in_place, seal_in_place, CryptError, EncapsulationSize, PrivateKeyRegistry,
         PublicKeyRegistry, TagSize,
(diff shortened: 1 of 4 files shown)


ipa-core/src/report/hybrid.rs
```diff
@@ -751,6 +757,48 @@ where
     }
 }
 
+impl PrfHybridReport<BA8, BA3> {
```
eriktaubeneck, line 760: Question about code layout: I had put this in the `protocol/oprf.rs` since this file (`report/hybrid.rs`) is a very large file at this point, and it's functionality is needed for the OPRF protocol. Not opposed to moving this here, just want to understand the philosophy.

eriktaubeneck, line 760: On a separate note, I know that `PRFIndistinguishableHybridReport` is a bit unwieldy, but I'm not sure about `PRFHybridReport`. `HybridReport` is an enum, with `Impression` and `Conversion`, but `PRFHybridReport` doesn't map to impressions and conversion. 

We are also about to need a new struct that drops the PRF and only needs the secret shares of `breakdown_key` and `value` that get aggregated. We should make sure we have some logical naming across all these.

ipa-core/src/sharding.rs
```diff
@@ -106,6 +106,19 @@ impl TryFrom<u128> for ShardIndex {
     }
 }
 
+/// It is common to compute the shard index based on some integer value modulo
+/// total shard count
+impl Rem<ShardIndex> for u64 {
```
eriktaubeneck, line 111: 👍

eriktaubeneck (approved):
a couple small questions, but overall looks good. thanks for making these changes!

ipa-core/src/report/hybrid.rs
akoshelev: I think we can probably survive with one struct and type aliases here. I don't mind moving them around and assigning better names. You're right I was mostly focused on the part where we replace the match key type and not on indistinguishability. I can follow up on renaming it.

P.S. The type that does not have a match key can be expressed as `IndistinguishableHybridReport<BK, V, ()>`

PR #1426, opened 2024-11-11
Implement protocol to create cross-shard coordinated randomness

akoshelev:
This protocol will generate random seeds using PRSS and then distribute those seeds to all shards. Every shard will initialize cross-shard PRSS from those seeds. Follows up on #1410.


eriktaubeneck:
is this supposed to integrate with `ShardBinding::cross_shard_prss`? If so, it's not obvious how that happens.

otherwise, looks good (other than a test issue to resolve)

eriktaubeneck (approved):

akoshelev:
> Is this supposed to integrate with ShardBinding::cross_shard_prss? If so, it's not obvious how that happens.
yes, this will be used in real world implementation to initialize PRSS instances on different shards

PR #1427, opened 2024-11-11
[Draft] Macro for serialization-related trait bounds

tyurek:
First attempt at making a macro for generating some of the trait bounds.
The interface is rather clunky as we can't just put a macro invocation in the trait bounds section. I'm following the pattern from this SO post: https://stackoverflow.com/questions/68728105/how-can-i-generate-trait-bounds-in-a-declarative-macro

The committed code has an example of what this would look like, which I'll also reproduce here:

```
with_bounds_for! { BK, ( $($bounds:tt)* ) => (
impl<BK: SharedValue> Serializable for HybridImpressionReport<BK>
where
    // Additional bounds could go here
    $($bounds)*
{
    type Size = gen_size!(BK);
   // proceed to define whatever functions...
}
)}
```

If we think this is still worthwhile, then I can expand this and start replacing some of the trait bounds, but I'm worried about making things clunkier, so I'd like a second opinion or two.

Alternatively, I tried using serde, but I'm stuck at making AdditiveShare work with serde::Serialize/Deserialize. I'm not sure how to implement these traits for <V as vector::traits::Vectorizable<N>>::Array

### DIFF
--- a/ipa-core/src/report/hybrid.rs
+++ b/ipa-core/src/report/hybrid.rs
@@ -114,14 +114,32 @@ where
     pub breakdown_key: Replicated<BK>,
 }
 
+macro_rules! with_bounds_for {(
+    $T:ident, $($rules:tt)*
+) => (
+    macro_rules! __emit__ { $($rules)* }
+    __emit__! {
+        $T: SharedValue,
+        Replicated<$T>: Serializable,
+        <Replicated<$T> as Serializable>::Size: Add<U16>,
+        <<Replicated<$T> as Serializable>::Size as Add<<Replicated<BA64> as Serializable>::Size>>:: Output: ArrayLength,
+    }
+)}
+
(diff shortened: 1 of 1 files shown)


eriktaubeneck:
when you only have one generic type (in this example, `BK`), the trait bound isn't too crazy. where it get's complicated is when you start adding more and more generics. 

this approach is a good start, but we actually need to make it recursive, based on an array of generics. a good start for this one would be to make `BA64` generic as well (e.g., `MK`), and make this work recursively.

eriktaubeneck:
> Yeah, the macro itself will need some work, but do you think the interface for calling it is acceptable?

Not quite. It should take an array of generics, not a single one. For example, you need to be able to build this with `[MK, BK, V]`.

akoshelev:
is it something where a proc macro can help?

PR #1429, opened 2024-11-11
upgrade hybrid tests to use malicious context

eriktaubeneck:
This upgrades our tests to use malicious context, and implements a couple necessary traits.

~Unfortunately, two tests are stalling (at least when run locally) after this change. I'm not sure what about this change would cause a stall, especially since it's not all the tests. @akoshelev could you take a look?~

~Stalling tests:~

### DIFF
--- a/ipa-core/src/protocol/hybrid/oprf.rs
+++ b/ipa-core/src/protocol/hybrid/oprf.rs
@@ -21,7 +21,8 @@ use crate::{
         context::{
             dzkp_validator::{DZKPValidator, TARGET_PROOF_SIZE},
             reshard_try_stream, DZKPUpgraded, MacUpgraded, MaliciousProtocolSteps, ShardedContext,
-            UpgradableContext, Validator,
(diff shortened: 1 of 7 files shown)


ipa-core/src/query/runner/hybrid.rs
```diff
@@ -120,6 +120,8 @@ where
         )
         .await?;
 
+        unimplemented!("protocol::hybrid::hybrid_protocol is not fully implemented");
```
eriktaubeneck, line 123: temporary addition to identify stall location. adding this just above line 116 allows the tests to pass

ipa-core/src/protocol/hybrid/oprf.rs
```diff
@@ -235,7 +250,7 @@ mod test {
 
             // TODO: we need to use malicious circuits here
```
andyleiserson, line 251:
```suggestion
```

andyleiserson (approved):

PR #1430, opened 2024-11-13
Support sharded shuffle in executor

akoshelev:
This does not get it to the end, but it implements the necessary parts to plumb sharded contexts and streams into the sharded shuffle toy protocol.

We still need to test it end-to-end and do some blackbox testing inside transport module, but this work is currently blocked behind changes being made inside the processor

### DIFF
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -122,6 +122,7 @@ impl ShardConfiguration for &Gateway {
     }
 
     fn shard_count(&self) -> ShardIndex {
+        // total number of shards include this instance and all its peers, so we add 1.
         ShardIndex::from(self.transports.shard.peer_count() + 1)
     }
 }
--- a/ipa-core/src/helpers/gateway/mod.rs
+++ b/ipa-core/src/helpers/gateway/mod.rs
@@ -28,7 +28,7 @@ use crate::{
         ShardChannelId, TotalRecords, Transport,
     },
     protocol::QueryId,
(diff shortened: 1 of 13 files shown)


ipa-core/src/helpers/gateway/mod.rs
```diff
@@ -106,6 +106,26 @@ pub struct GatewayConfig {
...
+    }
+}
+
+impl ShardConfiguration for &Gateway {
+    fn shard_id(&self) -> ShardIndex {
+        self.transports.shard.identity()
+    }
+
+    fn shard_count(&self) -> ShardIndex {
+        ShardIndex::from(self.transports.shard.peer_count() + 1)
```
cberkhoff, line 125: Can you add a comment saying we add 1 because these are my peers *excluding* me but shard count needs to count me or something like that?

ipa-core/src/net/transport.rs
```diff
@@ -336,6 +340,10 @@ impl Transport for ShardHttpTransport {
         self.shard_count.iter().filter(move |&v| v != this)
     }
 
+    fn peer_count(&self) -> u32 {
+        u32::from(self.shard_count).saturating_sub(1)
```
cberkhoff, line 344: can you add a test, just an assertion on some test that exists already?

cberkhoff (approved):

ipa-core/src/helpers/gateway/mod.rs
akoshelev: I think the code here is pretty self-explanatory, but I don't mind adding a comment

PR #1432, opened 2024-11-13
Remove SharedValue trait bound from MaliciousShuffleable

akoshelev:
The only reason it was there was to be able to use `AdditiveShare` inside malicious shuffle, but it felt too constraining to me. AdditiveShare has a lot of stuff built in and a lot of these are not required for Shuffle and a simple wrapper could to the same trick.

While working on this, I noticed that it gets a bit clunky to use in tests that call `verify_shuffle`. I tried to fix that but it is a much bigger refactoring. The idea was to add more traits supporting maliciuos shuffle. Initially we could consume `MaliciousShuffleable` that could later be upgraded to `MaliciousWithTags` type and bound to the original through `ShuffleWithTag`. That type could expose shares with tags attached. More importantly, it should provide a means to go back to original `MaliciousShuffleable`.

I got half way through and abandoned that effort for now

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
+++ b/ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
@@ -42,7 +42,6 @@ use crate::{
     seq_join::seq_join,
     sharding::ShardIndex,
 };
-// use crate::protocol::ipa_prf::shuffle::sharded::MaliciousShuffleShare;
 
 /// Container for left and right shares with tags attached to them.
 /// Looks like an additive share, but it is not because it does not need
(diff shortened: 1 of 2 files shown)


ipa-core/src/protocol/ipa_prf/shuffle/malicious.rs
```diff
@@ -42,6 +42,45 @@ use crate::{
     seq_join::seq_join,
     sharding::ShardIndex,
 };
+// use crate::protocol::ipa_prf::shuffle::sharded::MaliciousShuffleShare;
```
andyleiserson, line 45:
```suggestion
```

andyleiserson (approved):

PR #1433, opened 2024-11-13
Add BatchInversion code

tyurek:
Performing inversions in finite fields is relatively expensive, and multiple inversions can be batched together using a trick described here: https://en.wikipedia.org/wiki/Modular_multiplicative_inverse

Before: https://draft-mpc.vercel.app/query/view/akin-rib2024-11-13T0857
After: https://draft-mpc.vercel.app/query/view/jumpy-tack2024-11-13T0846

So... it's slightly slower lol (though that could be noise). Anyone have any performance tips here?

### DIFF
--- a/ipa-core/proptest-regressions/ff/prime_field.txt
+++ b/ipa-core/proptest-regressions/ff/prime_field.txt
@@ -1,8 +1,0 @@
-# Seeds for failure cases proptest has generated in the past. It is
-# automatically read and these particular cases re-run before any
-# novel cases are generated.
-#
-# It is recommended to check this file in to source control so that
-# everyone who runs the test benefits from these saved cases.
-cc 928efec8e7e6e629845d55708bfa6250016460758d23808b9022961a3ee05f7c # shrinks to mut elements = [0_mod31]
-cc c183e98fa937467116235fdc749988b53b0231a8abb12434c8d8d79890f59d98 # shrinks to mut elements = []
--- a/ipa-core/proptest-regressions/ff/prime_field.txt
+++ b/ipa-core/proptest-regressions/ff/prime_field.txt
@@ -1,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 928efec8e7e6e629845d55708bfa6250016460758d23808b9022961a3ee05f7c # shrinks to mut elements = [0_mod31]
+cc c183e98fa937467116235fdc749988b53b0231a8abb12434c8d8d79890f59d98 # shrinks to mut elements = []
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -65,9 +65,6 @@ where
 {
     let N = field_elements.len();
     let mut prefix_products = field_elements.to_vec();
-    //.iter()
-    //.copied() /*.map(|i| *i)*/
-    //.collect::<Vec<P>>();
     for i in 1..N {
         let temp = prefix_products[i] * prefix_products[i - 1];
         prefix_products[i] = temp;
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -344,7 +344,10 @@ macro_rules! field_impl {
             };
 
             use super::*;
-            use crate::ff::Serializable;
+            use crate::{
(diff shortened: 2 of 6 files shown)


akoshelev (commented):
I think this is a good change and we should get this code, if not for Lagrange, but for other future use cases. We just need to make sure we have good test coverage for it

I have a theory why it may not lead to perf improvements. I was working around this code last week and at some point suspected that this invert may be expensive. We only use N = 15 and N = 31 for Lagrange tables, so I just precomputed the coefficients and inlined them in the code, so no invert operations were required at runtime.

At that time, there was no difference, even though it saved around 0.5M invert operations per query. So it may be that we just don't do enough inverts here to make a dent.

tyurek:
@akoshelev are these values always the same between runs? (i.e. are we always interpolating to the same coordinates?)

akoshelev:
> are these values always the same between runs? (i.e. are we always interpolating to the same coordinates?)
the denominator is fixed for a Lagrange polynomial of a degree `N`, so it can be precomputed I think.

tyurek:
So, the function ends up well-tested if we leave it in the current usage (20+ tests will fail if it's incorrect). If it does actually make things slower then I don't want to leave it in use, but it sounds like it's just not getting called in the benchmark?

akoshelev:
> So, the function ends up well-tested if we leave it in the current usage (20+ tests will fail if it's incorrect).

the danger of testing it through other modules is that they may stop using this code and we lose coverage. It could also be possible that we won't be hitting all code paths in this function through other code, so I usually prefer to add tests that use the new functionality directly w/o relying on 3-rd party code.

> If it does actually make things slower then I don't want to leave it in use, but it sounds like it's just not getting called in the benchmark?

to me the runs you posted looked the same, and I attributed the difference to the noise.

tyurek:
Good point! I just added a specific test that checks correctness by comparing the result of batch_invert with just running invert on everything

ipa-core/proptest-regressions/ff/prime_field.txt
```diff
@@ -0,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 928efec8e7e6e629845d55708bfa6250016460758d23808b9022961a3ee05f7c # shrinks to mut elements = [0_mod31]
```
akoshelev, line 7: we probably don't need to commit this - seed shows that it just generates an empty input which proptest always does

ipa-core/src/ff/prime_field.rs
```diff
@@ -53,6 +53,32 @@ pub trait PrimeField: Field + U128Conversions {
     }
 }
 
+/// Performs multiple field inversions at once for a lower cost
+/// Follows the "multiple inverses" algorithm in
+/// <https://en.wikipedia.org/wiki/Modular_multiplicative_inverse>
+/// ## Panics
+/// If any element is 0
+#[allow(non_snake_case)]
+pub fn batch_invert<P>(field_elements: &mut [P])
```
akoshelev, line 62: would it be worth making it `batch_invert<const N: usize, P>(field_elements: &mut [P; N])`? that saves an allocation and works for Lagrange case

akoshelev (approved):

ipa-core/proptest-regressions/ff/prime_field.txt
tyurek: ah sorry, I missed that that got generated

ipa-core/src/ff/prime_field.rs
akoshelev:
```rust
let mut prefix_products = field_elements.to_vec();
```

`.to_vec()` needs an allocation I think. It is small and nothing crazy but if we don't need it, we could avoid it

tyurek: Oh, I actually need that allocation. The algorithm requires having a second array, but you compute that from the first one

akoshelev: if it is an array in, can you put another array on the stack to hold the intermediates?

tyurek: Making sure I understand correctly: You're saying that if the function takes `[P; N]`, then you know the size at compile time and so you can avoid allocating `prefix_products` on the heap, and put it on the stack instead?

akoshelev: that's right - `N` is assumed to be small enough for array to be kept on the stack.

PR #1434, opened 2024-11-13
compiling version of aggregate_reports

eriktaubeneck:
This is a draft of implementing `aggregate_reports`, which adds the expected two shares from each match key (dropping any matchkeys without exactly 2 reports.)

The test is incomplete, and it stalls. Presumably this has to do with the malicious validation. Any feedback or pointers to other code that does something similar would be helpful. I was looking at [this implementation](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/protocol/ipa_prf/prf_sharding/mod.rs#L501-L548), and admittedly some of the comments there suggest there may be a better way to do this.

Additionally, this doesn't yet utilize vectorization. (cc @andyleiserson). Any feedback or pointers in that direction would also be greatly appreciated.

### DIFF
--- a/ipa-core/src/protocol/hybrid/agg.rs
+++ b/ipa-core/src/protocol/hybrid/agg.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 
-use futures::{stream, FutureExt, StreamExt, TryStreamExt};
+use futures::{stream, StreamExt, TryStreamExt};
 
 use crate::{
     error::Error,
@@ -9,7 +9,7 @@ use crate::{
     protocol::{
         boolean::step::EightBitStep,
         context::{
-            dzkp_validator::{DZKPValidator, TARGET_PROOF_SIZE},
+            dzkp_validator::{validated_seq_join, DZKPValidator, TARGET_PROOF_SIZE},
             Context, DZKPUpgraded, MaliciousProtocolSteps, ShardedContext, UpgradableContext,
         },
         hybrid::step::{AggregateReportsStep, HybridStep},
@@ -132,11 +132,8 @@ where
             }
         });
 
-    dzkp_validator
-        .validated_seq_join(agg_work)
+    validated_seq_join(dzkp_validator, agg_work)
         .try_collect()
-        // "Implementation of Send is not general enough" famous issue
-        .boxed()
         .await
 }
 
--- a/ipa-core/src/protocol/hybrid/agg.rs
+++ b/ipa-core/src/protocol/hybrid/agg.rs
@@ -1,0 +1,217 @@
+use std::collections::HashMap;
+
+use crate::{
+    error::Error,
+    ff::{boolean::Boolean, boolean_array::BooleanArray, ArrayAccess},
+    helpers::TotalRecords,
+    protocol::{
+        boolean::step::{EightBitStep, ThirtyTwoBitStep},
+        context::{
+            dzkp_validator::{DZKPValidator, TARGET_PROOF_SIZE},
+            Context, DZKPUpgraded, MaliciousProtocolSteps, UpgradableContext,
+        },
+        hybrid::step::HybridStep,
+        ipa_prf::boolean_ops::addition_sequential::integer_add,
+        BooleanProtocols, RecordId,
+    },
+    report::hybrid::{AggregateableHybridReport, PrfHybridReport},
+    secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
+};
+
+pub async fn aggregate_reports<BK, V, C>(
+    ctx: C,
+    reports: Vec<PrfHybridReport<BK, V>>,
+) -> Result<Vec<AggregateableHybridReport<BK, V>>, Error>
+where
+    C: UpgradableContext,
+    BK: SharedValue + BooleanArray,
+    V: SharedValue + BooleanArray,
+    Replicated<Boolean>: BooleanProtocols<DZKPUpgraded<C>>,
+{
+    let mut reports_by_matchkey = HashMap::new();
+    reports_by_matchkey.reserve(reports.len() / 2);
+
+    for report in reports {
+        reports_by_matchkey
+            .entry(report.match_key)
+            .or_insert_with(Vec::new)
+            .push(report);
+    }
+
+    // an honest client and report collector will only submit
+    // one report with a breakdown key and one report with a value.
+    // if there are less, it's unattributed. if more, something went wrong.
+    // we remove these (instead of erroring/panicing).
+    let report_pairs: Vec<[AggregateableHybridReport<BK, V>; 2]> = reports_by_matchkey
+        .into_iter()
+        .filter_map(|(_, value)| {
+            if value.len() == 2 {
+                Some([value[0].clone().into(), value[1].clone().into()])
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let mut agg_reports = Vec::new();
+
+    let chunk_size = TARGET_PROOF_SIZE;
+
+    let dzkp_validator = ctx.clone().dzkp_validator(
+        MaliciousProtocolSteps {
+            protocol: &HybridStep::GroupBySum,
+            validate: &HybridStep::GroupBySumValidate,
+        },
+        std::cmp::min(ctx.active_work().get(), chunk_size.next_power_of_two()),
+    );
+
+    let ctx = dzkp_validator
+        .context()
+        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
+
+    for (i, reports) in report_pairs.into_iter().enumerate() {
+        let record_id_bk = RecordId::FIRST + 2 * i;
+        let record_id_v = RecordId::FIRST + 2 * i + 1;
+        let (breakdown_key, _) = integer_add::<_, EightBitStep, 1>(
+            ctx.clone(),
+            record_id_bk,
+            &reports[0].breakdown_key.to_bits(),
+            &reports[1].breakdown_key.to_bits(),
+        )
+        .await?;
+        let (value, _) = integer_add::<_, ThirtyTwoBitStep, 1>(
+            ctx.clone(),
+            record_id_v,
+            &reports[0].value.to_bits(),
+            &reports[1].value.to_bits(),
+        )
+        .await?;
+
+        agg_reports.push(AggregateableHybridReport::<BK, V> {
+            match_key: (),
+            breakdown_key: breakdown_key.collect_bits(),
+            value: value.collect_bits(),
+        });
+    }
+
+    Ok(agg_reports)
+}
+
+#[cfg(all(test, unit_test, feature = "in-memory-infra"))]
+pub mod test {
+    use super::aggregate_reports;
+    use crate::{
+        ff::boolean_array::{BA3, BA8},
+        report::hybrid::{
+            AggregateableHybridReport, IndistinguishableHybridReport, PrfHybridReport,
+        },
+        test_executor::run,
+        test_fixture::{hybrid::TestHybridRecord, Runner, TestWorld, TestWorldConfig, WithShards},
+    };
+
+    #[test]
+    fn aggregate_reports_test() {
+        run(|| async {
+            let records = vec![
+                TestHybridRecord::TestImpression {
+                    match_key: 12345,
+                    breakdown_key: 2,
+                },
+                TestHybridRecord::TestImpression {
+                    match_key: 23456,
+                    breakdown_key: 4,
+                },
+                TestHybridRecord::TestConversion {
+                    match_key: 23456,
+                    value: 1,
+                }, // attributed
+                TestHybridRecord::TestImpression {
+                    match_key: 34567,
+                    breakdown_key: 1,
+                },
+                TestHybridRecord::TestImpression {
+                    match_key: 45678,
+                    breakdown_key: 3,
+                },
+                TestHybridRecord::TestConversion {
+                    match_key: 45678,
+                    value: 2,
+                }, // attributed
+                TestHybridRecord::TestImpression {
+                    match_key: 56789,
+                    breakdown_key: 5,
+                },
+                TestHybridRecord::TestConversion {
+                    match_key: 67890,
+                    value: 3,
+                }, // NOT attributed
+                TestHybridRecord::TestImpression {
+                    match_key: 78901,
+                    breakdown_key: 2,
+                },
+                TestHybridRecord::TestConversion {
+                    match_key: 78901,
+                    value: 4,
+                }, // attributed twice
+                TestHybridRecord::TestConversion {
+                    match_key: 78901,
+                    value: 5,
+                }, // attributed twice
+                TestHybridRecord::TestImpression {
+                    match_key: 89012,
+                    breakdown_key: 4,
+                },
+                TestHybridRecord::TestConversion {
+                    match_key: 89012,
+                    value: 6,
+                }, // attributed
+                TestHybridRecord::TestConversion {
+                    match_key: 90123,
+                    value: 7,
+                }, // NOT attributed
+            ];
+
+            // at this point, all unattributed values end up in index 0
+            // we will zero them out later.
+            // let expected = vec![
+            //     22, 0, 43, // 14 + 8, 12 + 31
+            //     13, 33, // 25 + 8
+            //     0,
+            // ];
+
+            let world = TestWorld::<WithShards<2>>::with_shards(TestWorldConfig::default());
+
+            let results: Vec<[Vec<AggregateableHybridReport<BA8, BA3>>; 3]> = world
+                .malicious(records.clone().into_iter(), |ctx, input| {
+                    let og_records = records.clone();
+                    async move {
+                        let indistinguishable_reports: Vec<
+                            IndistinguishableHybridReport<BA8, BA3>,
+                        > = input.iter().map(|r| r.clone().into()).collect::<Vec<_>>();
+
+                        let prf_reports: Vec<PrfHybridReport<BA8, BA3>> = indistinguishable_reports
+                            .iter()
+                            .zip(og_records.iter())
+                            .map(|(indist_report, test_report)| {
+                                let match_key = match test_report {
+                                    TestHybridRecord::TestConversion { match_key, .. } => match_key,
+                                    TestHybridRecord::TestImpression { match_key, .. } => match_key,
+                                };
+                                PrfHybridReport {
+                                    match_key: *match_key,
+                                    value: indist_report.value.clone(),
+                                    breakdown_key: indist_report.breakdown_key.clone(),
+                                }
+                            })
+                            .collect::<Vec<_>>();
+
+                        aggregate_reports(ctx.clone(), prf_reports).await.unwrap()
+                    }
+                })
+                .await;
+
+            println!("{:?}", results);
+            panic!();
+        })
+    }
+}
--- a/ipa-core/src/protocol/hybrid/agg.rs
+++ b/ipa-core/src/protocol/hybrid/agg.rs
@@ -8,14 +8,14 @@ use crate::{
     ff::{boolean::Boolean, boolean_array::BooleanArray, ArrayAccess},
     helpers::TotalRecords,
     protocol::{
-        boolean::step::{EightBitStep, ThirtyTwoBitStep},
+        boolean::step::EightBitStep,
         context::{
             dzkp_validator::{DZKPValidator, TARGET_PROOF_SIZE},
-            Context, DZKPUpgraded, MaliciousProtocolSteps, UpgradableContext,
+            Context, DZKPContext, DZKPUpgraded, MaliciousProtocolSteps, UpgradableContext,
         },
         hybrid::step::{AggregateReportsStep, HybridStep},
         ipa_prf::boolean_ops::addition_sequential::integer_add,
-        BooleanProtocols, RecordId,
+        BooleanProtocols,
     },
     report::hybrid::{AggregateableHybridReport, PrfHybridReport},
     secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
@@ -71,8 +71,6 @@ where
         .filter_map(|(_, v)| if v.1 == 2 { Some(v.0) } else { None })
         .collect::<Vec<_>>();
 
-    // let mut agg_reports = Vec::with_capacity(report_pairs.len());
-
     let chunk_size = TARGET_PROOF_SIZE;
 
     let dzkp_validator = ctx.clone().dzkp_validator(
@@ -85,29 +83,28 @@ where
 
     let ctx = dzkp_validator
         .context()
-        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
+        .set_total_records(TotalRecords::specified(report_pairs.len())?);
 
     let agg_work = stream::iter(report_pairs)
         .enumerate()
         .map(|(idx, reports)| {
-            let record_id_bk = RecordId::FIRST + 2 * idx;
-            let record_id_v = RecordId::FIRST + 2 * idx + 1;
             let agg_ctx = ctx.clone();
             async move {
                 let (breakdown_key, _) = integer_add::<_, EightBitStep, 1>(
-                    agg_ctx.narrow(&AggregateReportsStep::Add),
-                    record_id_bk,
+                    agg_ctx.narrow(&AggregateReportsStep::AddBK),
+                    idx.into(),
                     &reports[0].breakdown_key.to_bits(),
                     &reports[1].breakdown_key.to_bits(),
                 )
                 .await?;
-                let (value, _) = integer_add::<_, ThirtyTwoBitStep, 1>(
-                    agg_ctx.narrow(&AggregateReportsStep::Add),
-                    record_id_v,
+                let (value, _) = integer_add::<_, EightBitStep, 1>(
+                    agg_ctx.narrow(&AggregateReportsStep::AddV),
+                    idx.into(),
                     &reports[0].value.to_bits(),
                     &reports[1].value.to_bits(),
                 )
                 .await?;
+                agg_ctx.validate_record(idx.into()).await?;
                 Ok::<_, Error>(AggregateableHybridReport::<BK, V> {
                     match_key: (),
                     breakdown_key: breakdown_key.collect_bits(),
@@ -227,7 +224,7 @@ pub mod test {
                 })
                 .await;
 
-            println!("{results:?}");
+            println!("results: {results:?}");
             // todo: reconstruct results
             // assert_eq!(results, expected);
         });
--- a/ipa-core/src/protocol/hybrid/agg.rs
+++ b/ipa-core/src/protocol/hybrid/agg.rs
@@ -154,26 +154,34 @@ where
 
 #[cfg(all(test, unit_test))]
 pub mod test {
+    use rand::Rng;
+
     use super::{aggregate_reports, group_report_pairs_ordered};
     use crate::{
         ff::{
             boolean_array::{BA3, BA8},
             U128Conversions,
         },
+        helpers::Role,
         protocol::context::Context,
         report::hybrid::{
             AggregateableHybridReport, IndistinguishableHybridReport, PrfHybridReport,
         },
-        secret_sharing::replicated::{
-            semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing,
+        secret_sharing::{
+            replicated::{semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing},
+            SharedValue,
         },
         sharding::{ShardConfiguration, ShardIndex},
-        test_executor::run,
+        test_executor::{run, run_random},
         test_fixture::{
             hybrid::{TestAggregateableHybridReport, TestHybridRecord},
             Reconstruct, Runner, TestWorld, TestWorldConfig, WithShards,
         },
     };
+
+    // the inputs are laid out to work with exactly 2 shards
+    // as if it we're resharded by match_key/prf
+    const SHARDS: usize = 2;
 
     // we re-use these as the "prf" of the match_key
     // to avoid needing to actually do the prf here
@@ -293,7 +301,7 @@ pub mod test {
                 ],
             ];
 
-            let world = TestWorld::<WithShards<2>>::with_shards(TestWorldConfig::default());
+            let world = TestWorld::<WithShards<SHARDS>>::with_shards(TestWorldConfig::default());
             #[allow(clippy::type_complexity)]
             let results: Vec<[Vec<[AggregateableHybridReport<BA8, BA3>; 2]>; 3]> = world
                 .malicious(records.clone().into_iter(), |ctx, input| {
@@ -316,9 +324,6 @@ pub mod test {
                                 breakdown_key: indist_report.breakdown_key.clone(),
                             })
                             .collect::<Vec<_>>();
-                        println!("{:#?}", ctx.shard_id());
-                        println!("{:#?}", ctx.role());
-                        println!("{prf_reports:#?}");
                         group_report_pairs_ordered(prf_reports)
                     }
                 })
@@ -342,7 +347,6 @@ pub mod test {
                 })
                 .collect::<Vec<_>>();
 
-            println!("{results:#?}");
             assert_eq!(results, expected);
         });
     }
@@ -369,7 +373,7 @@ pub mod test {
                 },
             ];
 
-            let world = TestWorld::<WithShards<2>>::with_shards(TestWorldConfig::default());
+            let world = TestWorld::<WithShards<SHARDS>>::with_shards(TestWorldConfig::default());
 
             let results: Vec<[Vec<AggregateableHybridReport<BA8, BA3>>; 3]> = world
                 .malicious(records.clone().into_iter(), |ctx, input| {
@@ -411,7 +415,6 @@ pub mod test {
                 })
                 .collect::<Vec<_>>();
 
-            println!("{results:#?}");
             assert_eq!(results, expected);
         });
     }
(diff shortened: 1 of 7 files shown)


ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+        },
+        hybrid::step::HybridStep,
+        ipa_prf::boolean_ops::addition_sequential::integer_add,
+        BooleanProtocols, RecordId,
+    },
+    report::hybrid::{AggregateableHybridReport, PrfHybridReport},
+    secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
+};
+
+pub async fn aggregate_reports<BK, V, C>(
```
akoshelev, line 21: lets add comments explaining the high level algorithm here before merging

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+    reports: Vec<PrfHybridReport<BK, V>>,
+) -> Result<Vec<AggregateableHybridReport<BK, V>>, Error>
+where
+    C: UpgradableContext,
+    BK: SharedValue + BooleanArray,
+    V: SharedValue + BooleanArray,
+    Replicated<Boolean>: BooleanProtocols<DZKPUpgraded<C>>,
+{
+    let mut reports_by_matchkey = HashMap::new();
+    reports_by_matchkey.reserve(reports.len() / 2);
```
akoshelev, line 32:
```suggestion
    let mut reports_by_matchkey = HashMap::with_capacity(...);
```

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+            .entry(report.match_key)
+            .or_insert_with(Vec::new)
+            .push(report);
+    }
+
+    // an honest client and report collector will only submit
+    // one report with a breakdown key and one report with a value.
+    // if there are less, it's unattributed. if more, something went wrong.
+    // we remove these (instead of erroring/panicing).
+    let report_pairs: Vec<[AggregateableHybridReport<BK, V>; 2]> = reports_by_matchkey
```
akoshelev, line 45: it should be possible to filter out reports as you are adding them to the hashmap. That eliminates the need for a vector inside that map and postprocessing. @cberkhoff had somewhat similar problem to tackle in the aggregation, so that code can potentially be reused here

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+        .filter_map(|(_, value)| {
+            if value.len() == 2 {
+                Some([value[0].clone().into(), value[1].clone().into()])
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let mut agg_reports = Vec::new();
```
akoshelev, line 56:
```suggestion
    let mut agg_reports = Vec::with_capacity(report_pairs.len());
```

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let ctx = dzkp_validator
+        .context()
+        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
+
+    for (i, reports) in report_pairs.into_iter().enumerate() {
+        let record_id_bk = RecordId::FIRST + 2 * i;
+        let record_id_v = RecordId::FIRST + 2 * i + 1;
+        let (breakdown_key, _) = integer_add::<_, EightBitStep, 1>(
+            ctx.clone(),
```
akoshelev, line 76: you need to narrow here and that shouldn't require record id math: https://github.com/private-attribution/ipa/blob/29578dd8bee2a274b194e3e2620b69aabf49f4e7/ipa-core/src/protocol/ipa_prf/aggregation/mod.rs#L188

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+            validate: &HybridStep::GroupBySumValidate,
+        },
+        std::cmp::min(ctx.active_work().get(), chunk_size.next_power_of_two()),
+    );
+
+    let ctx = dzkp_validator
+        .context()
+        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
+
+    for (i, reports) in report_pairs.into_iter().enumerate() {
```
akoshelev, line 72: is there a reason why we need to add those sequentially or `seq_join` can work here too?

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,217 @@
...
+                }, // attributed twice
+                TestHybridRecord::TestConversion {
+                    match_key: 78901,
+                    value: 5,
+                }, // attributed twice
+                TestHybridRecord::TestImpression {
+                    match_key: 89012,
+                    breakdown_key: 4,
+                },
+                TestHybridRecord::TestConversion {
```
akoshelev, line 164: what happens if we have two impressions and one conversion for the same match key?

akoshelev (changes requested):

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: they should get filtered out, but that's a good test to include.

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: Another way to do this would be to have the values in the hashmap be a tuple: `([AggregateableHybridReport<BK, V>; 2], usize)`. where the second value is a count of the number of times that `match_key` appears. then we grab a vector of the values from that hashmap and filter down to instances where the count is exactly 2. 

that should be a bit more memory efficient since we'll have an upper bound on the size and know that every value is exactly an array with two reports and a usize.

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: ty, let me give this a try. do you think that's the issue will address the stall?

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: I can try that out. there shouldn't be any reason why it needs to be sequential.

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: I think stall occurs because of not using seq_join

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: you may run into Rust issues here - the most straightforward way to go around would be to use `[Option<Report>; 2]` as the value in that hashmap. It would be a bit clunky to deal with this array having 1 or 2 items, so you could create a wrapper type for it

eriktaubeneck: take a look at this, it uses an fixed array and count, and then filters the rows without count == 2 at the end.

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,235 @@
...
+                    match_key: (),
+                    breakdown_key: breakdown_key.collect_bits(),
+                    value: value.collect_bits(),
+                })
+            }
+        });
+
+    let agg_result = seq_join(ctx.active_work(), agg_work)
+        .try_collect::<Vec<_>>()
+        .await?;
```
eriktaubeneck, line 121: @akoshelev is this the right way to use `seq_join`? it compiles, but it's still stalling.

akoshelev: I don't think you need to do record id math here, just use `idx` as record id. Where does it stall?

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: it works but it would be better if this internal state magic happens somewhere outside of this function. Something like breakdown reveal aggregation struct would be great to have here

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,235 @@
...
+        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
+
+    let agg_work = stream::iter(report_pairs)
+        .enumerate()
+        .map(|(idx, reports)| {
+            let record_id_bk = RecordId::FIRST + 2 * idx;
+            let record_id_v = RecordId::FIRST + 2 * idx + 1;
+            let agg_ctx = ctx.clone();
+            async move {
+                let (breakdown_key, _) = integer_add::<_, EightBitStep, 1>(
```
eriktaubeneck, line 97: how do we pick the step size here? does it depend on the size of BK and V? if so, how do we do it with generics?

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: > I don't think you need to do record id math here, just use idx as record id.

is it ok to use the same record_id for both additions? I figured they needed to be distinct.

eriktaubeneck: here are the logs where it stalls:
```
2024-11-13T23:32:26.777285Z TRACE {role=H2}:receive{i=0 from=H3 gate="protocol/iter000/group_by_sum/add/bit0"}:send_stream{to=H2 gate=gate=protocol/iter000/group_by_sum/add/bit0}: ipa_core::helpers::gateway::send: new
2024-11-13T23:32:26.777336Z TRACE {role=H2}:receive{i=0 from=H3 gate="protocol/iter000/group_by_sum/add/bit0"}:send_stream{to=H2 gate=gate=protocol/iter000/group_by_sum/add/bit0}: ipa_core::helpers::gateway::send: close time.busy=3.25µs time.idle=47.2µs
2024-11-13T23:32:26.777381Z TRACE {role=H3}:receive{i=0 from=H1 gate="protocol/iter000/group_by_sum/add/bit0"}:send_stream{to=H3 gate=gate=protocol/iter000/group_by_sum/add/bit0}: ipa_core::helpers::gateway::send: new
2024-11-13T23:32:26.777475Z TRACE {role=H3}:receive{i=0 from=H1 gate="protocol/iter000/group_by_sum/add/bit0"}:send_stream{to=H3 gate=gate=protocol/iter000/group_by_sum/add/bit0}: ipa_core::helpers::gateway::send: close time.busy=2.75µs time.idle=90.9µs
2024-11-13T23:32:36.771245Z  WARN stall_detector{role=H2}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=4 state=
{
"gate=protocol/iter000/group_by_sum/add/bit0", from=H3. Waiting to receive records ["[0, 1]"].
}
2024-11-13T23:32:36.771278Z  WARN stall_detector{role=H1}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=4 state=
{
"gate=protocol/iter000/group_by_sum/add/bit0", from=H2. Waiting to receive records ["[0, 1]"].
}
2024-11-13T23:32:36.771271Z  WARN stall_detector{role=H3}: ipa_core::helpers::gateway::stall_detection::gateway: Helper is stalled sn=4 state=
{
"gate=protocol/iter000/group_by_sum/add/bit0", from=H1. Waiting to receive records ["[0, 1]"].
```

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,235 @@
...
+        MaliciousProtocolSteps {
+            protocol: &HybridStep::GroupBySum,
+            validate: &HybridStep::GroupBySumValidate,
+        },
+        std::cmp::min(ctx.active_work().get(), chunk_size.next_power_of_two()),
+    );
+
+    let ctx = dzkp_validator
+        .context()
+        .set_total_records(TotalRecords::specified(2 * report_pairs.len())?);
```
eriktaubeneck, line 88: I did `2 * report_pairs.len()` here because we do 2 additions per row in the vec, but if we're just using the index in the vec as the record_id, should this not be 2x? cc @akoshelev

akoshelev: yes, it should **not** be 2x I think

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,232 @@
...
+                    value: value.collect_bits(),
+                })
+            }
+        });
+
+    let agg_result = seq_join(ctx.active_work(), agg_work)
+        .try_collect::<Vec<_>>()
+        .await?;
+    Ok(agg_result)
+}
```
eriktaubeneck, line 120: stall fixed! thanks for the feedback @akoshelev 

now a new problem:
```
thread 'protocol::hybrid::agg::test::aggregate_reports_test' panicked at ipa-core/src/protocol/context/dzkp_validator.rs:937:32:
called `Result::unwrap()` on an `Err` value: ContextUnsafe("DZKPMaliciousContext")
```

it seems like this is related to validate not betting called, but it's called above on line 107. any suggestions?

akoshelev: I can't spot any obvious problems with this code - validation is happening. It could be a miscalculation of batch size for malicious validation but I'd expect a stall in this case. @andyleiserson any thoughts why ZKP stuff may be throwing this?

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: If we don't have enough bits in step, it will panic. Using a step that is too large may lead to compact step size blowout, but it may not be a concern for hybrid. We can probably pick a reasonable number of bits for both (whatever we use in the aggregation protocol)

eriktaubeneck: EightBit is probably sufficient for both of these for now, then. We'll need a larger one once we aggregate multiple values.

akoshelev: iirc for breakdown keys the additions don't matter as it is a sum with 0. for trigger values we may need to handle overflow

eriktaubeneck: at this point, we're assuming that there is exactly 1 conversion and 1 impression, so the same is true for the value here. not true when we have multiple values to aggregate though (later in the protocol)

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: a clue: if I remove the validation on 107, it still panics in the same way...

the stacktrace isn't that helpful and doesn't seem to point where inside this function the panic is happening

akoshelev: this happens when the validator is dropped, so it may be happening earlier than expected.

eriktaubeneck: do I need to validate for both steps?

eriktaubeneck: I added a second validate call and it didn't seem to help

akoshelev: oh yea, I think you need to have a single validate after both steps

eriktaubeneck: tried that but it didn't help. (didn't push it because it didn't fix it)

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,232 @@
...
+                Ok::<_, Error>(AggregateableHybridReport::<BK, V> {
+                    match_key: (),
+                    breakdown_key: breakdown_key.collect_bits(),
+                    value: value.collect_bits(),
+                })
+            }
+        });
+
+    dzkp_validator
+        .validated_seq_join(agg_work)
```
akoshelev, line 115: @andyleiserson This version works only with single batch and validated_seq_join. If number of batches is not `usize::MAX`, validation fails

akoshelev: also if I use `validate_record` directly inside the closure instead of `validated_seq_join`

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,232 @@
...
+            // If the count is 0, replace the first element with the new report
+            entry.0[0] = report.into();
+            entry.1 += 1;
+        } else if entry.1 == 1 {
+            // If the count is 1, replace the second element with the new report
+            entry.0[1] = report.into();
+            entry.1 += 1;
+        } else {
+            // If the count is 2 or more, increment the counter and drop the report
+            entry.1 += 1;
```
andyleiserson, line 64: You could define something like this as a helper to simplify this function:

```
enum MatchEntry {
    Single(Report),
    Pair(Report, Report),
    MoreThanTwo,
}
```

Whether or not you use the enum, you can match on `reports_by_matchkey.entry(match_key)` to avoid creating a zero entry temporarily. If `Vacant`, then insert an entry with one report, otherwise (`Occupied`), add a second report or mark as overflowed.

ipa-core/src/protocol/hybrid/agg.rs
andyleiserson: This seems concerning, I will look at it.

akoshelev: it may be HashMap but I don't think I fully understand why

eriktaubeneck: making this change:
```
let chunk_size: usize = TARGET_PROOF_SIZE / (BK::BITS as usize + V::BITS as usize);
```

and setting batch size to `chunk_size.next_power_of_two(),` seems to work!

andyleiserson: With or without your change, it only passes some of the time. With your change, it passes more often. Since the test only has two pairs that will actually be aggregated, it makes sense that it passes more often with more shards (less chance that both the pairs end up on the same shard). But I can't totally explain the pass rates that I am seeing. (I got 2 or 3 passes out of 40 before your change, 11 or 12 out of 40 after.)

andyleiserson: > making this change ... and setting batch size to chunk_size.next_power_of_two(), seems to work!

Are you sure it passes reliably? I really think the failures are due to `HashMap`.

 > it may be HashMap but I don't think I fully understand why

Considering a two party MPC for simplicity. If one helper has (share of A, share of B) and the other helper has (share of B, share of A), and they add those vectors together, the result is garbage. Taking proofs out of the picture, it would also fail a functional output check, the proofs just catch it sooner.

eriktaubeneck: @akoshelev and I just chatted about this live, and I think you're both right. the Hashmap will get things out of order, adn thus the proofs fail. I'm working on swapping out with a BTreeMap now.

akoshelev: > Considering a two party MPC for simplicity. If one helper has (share of A, share of B) and the other helper has (share of B, share of A), and they add those vectors together, the result is garbage. Taking proofs out of the picture, it would also fail a functional output check, the proofs just catch it sooner.

yea that makes sense. I think I got tests failing in almost 100% of the cases **without** validated seq join and passing locally **with** validated seq join

ipa-core/src/report/hybrid.rs
```diff
@@ -675,6 +676,40 @@ where
 /// Converted report where shares of match key are replaced with OPRF value
 pub type PrfHybridReport<BK, V> = IndistinguishableHybridReport<BK, V, u64>;
 
+/// After grouping `IndistinguishableHybridReport`s by the OPRF of thier `match_key`,
+/// that OPRF value is no longer required.
+pub type AggregateableHybridReport<BK, V> = IndistinguishableHybridReport<BK, V, ()>;
+
+impl<BK, V> AggregateableHybridReport<BK, V>
```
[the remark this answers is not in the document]
eriktaubeneck: @akoshelev suggested this elsewhere. I'm not opposed to making it it's own struct (especially since we will also want to use it with a larger value type for adding up lots of reports.)

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,253 @@
...
+{
+    let mut reports_by_matchkey: BTreeMap<u64, MatchEntry<BK, V>> = BTreeMap::new();
+
+    for report in reports {
+        let match_key = report.match_key;
+        match reports_by_matchkey.get(&match_key) {
+            Some(match_entry) => match match_entry {
+                MatchEntry::Single(s_report) => {
+                    reports_by_matchkey
+                        .insert(match_key, MatchEntry::Pair(report.into(), s_report.clone()));
```
eriktaubeneck, line 56: @andyleiserson can you take a look at this? I don't love that we have to `.clone()` here. Is there a better option?

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,250 @@
...
+                    match_key: (),
+                    breakdown_key: breakdown_key.collect_bits(),
+                    value: value.collect_bits(),
+                })
+            }
+        });
+
+    validated_seq_join(dzkp_validator, agg_work)
+        .try_collect()
+        .await
```
eriktaubeneck, line 137: cc @akoshelev this seems to work

ipa-core/src/protocol/hybrid/agg.rs
andyleiserson: If you clone into the same memory (you'll need to swap the order of the reports in the pair for this to be the case), then the compiler is smart enough to optimize it away. Given the size of the report here, I really doubt moving the report in memory would affect things compared to the overhead of everything else that's going on (map lookup and insertions).

If neither of those rationales works for you, then I don't think the enum is viable as-is, maybe something like:
```
struct MatchEntry {
    first_report: Report,
    matched_reports: MatchedReports,
}

enum MatchedReports {
    None,
    One(Report),
    TwoOrMore,
}
```

I was thinking that the logic that updates the entry with reports could also be made part of `MatchEntry`, like this:
```
impl MatchEntry {
    pub fn add_report(&mut self, new_report: Report) {
        match self {
            Self::Single(old_report) => {
                *self = Self::Pair(old_report.clone(), new_report);
            }
            Self::Pair { .. } | Self::MoreThanTwo => *self = Self::MoreThanTwo,
        }
    }
}
```
I think this is something like what @akoshelev had in mind [here](https://github.com/private-attribution/ipa/pull/1434#discussion_r1841289923). An advantage of encapsulation of this sort is that it's easier to adjust the memory representation of the data structure, if it is determined to be a performance issue.

eriktaubeneck: awesome, thanks for the suggestions.

ipa-core/src/report/hybrid.rs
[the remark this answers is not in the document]
akoshelev: maybe I misunderstand Andy's point here, I think he is advocating for writing `impl IndistinguishableHybridReport<BK, V, ()>` rather than not using type aliases at all

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: that's exactly right. I also think code that manages this one or two things in me logic does not really belong to the algorithm and distracts readers from the big picture. As Andy pointed already, having `add_report(..)` call inside the function gives enough information to understand what is going on

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: oh well...send bounds issue is really nasty

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+#[cfg(all(test, unit_test, feature = "in-memory-infra"))]
```
akoshelev, line 145:
```suggestion
#[cfg(all(test, unit_test)]
```

unit_test flag is a bit hard to find, but it is in `build.rs` file. You can see all features that are enabled by it and `in-memory-infra` is one of them

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,255 @@
...
+        },
+        hybrid::step::{AggregateReportsStep, HybridStep},
+        ipa_prf::boolean_ops::addition_sequential::integer_add,
+        BooleanProtocols,
+    },
+    report::hybrid::{AggregateableHybridReport, PrfHybridReport},
+    secret_sharing::replicated::semi_honest::AdditiveShare as Replicated,
+};
+
+enum MatchEntry<BK, V>
```
akoshelev, line 23: we shouldn't require `Empty` state, it can be modelled through the btreemap itself

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,255 @@
...
+    BK: BooleanArray,
+    V: BooleanArray,
+{
+    let mut reports_by_matchkey: BTreeMap<u64, MatchEntry<BK, V>> = BTreeMap::new();
+
+    for report in reports {
+        let match_entry = reports_by_matchkey
+            .entry(report.match_key)
+            .or_insert(MatchEntry::Empty);
+        match_entry.add_report(report.into());
```
akoshelev, line 71:
```suggestion
            .or_insert(MatchEntry::Single(report));
```

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,255 @@
...
+/// An honest report collector will only provide a single impression report per `match_key` and
+/// an honest client will only provide a single conversion report per `match_key`.
+fn group_report_pairs<BK, V>(
+    reports: Vec<PrfHybridReport<BK, V>>,
+) -> Vec<[AggregateableHybridReport<BK, V>; 2]>
+where
+    BK: BooleanArray,
+    V: BooleanArray,
+{
+    let mut reports_by_matchkey: BTreeMap<u64, MatchEntry<BK, V>> = BTreeMap::new();
```
akoshelev, line 65: can you please add a comment explaining why it is bitmap instead of hashmap? Just to make sure we don't accidentally refactor it

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: maybe I'm misunderstanding. isn't this better, since we no longer use boxed?

ipa-core/src/report/hybrid.rs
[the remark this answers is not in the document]
eriktaubeneck: ahh ok I misunderstood. I have no problem with changing this to what you suggested, @akoshelev

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    // we only keep the reports from match_keys that provided exactly 2 reports
+    reports_by_matchkey
+        .into_values()
+        .filter_map(|match_entry| match match_entry {
+            MatchEntry::Pair(r1, r2) => Some([r1, r2]),
+            _ => None,
+        })
+        .collect::<Vec<_>>()
```
akoshelev, line 81: I think this refactoring made it less efficient as you now need to collect the intermediates. I didn't think we need that, but you want to have a function, maybe you can make it borrowing from `reports` and producing an iterator, like this

```rust
fn group_report_pairs<'a, BK, V>(
    reports: &'a [PrfHybridReport<BK, V>],
) -> impl Iterator<Item = [AggregateableHybridReport<BK, V>; 2]>> + 'a
```

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,282 @@
...
+        match_entry.add_report(report.into());
+    }
+
+    // we only keep the reports from match_keys that provided exactly 2 reports
+    reports_by_matchkey
+        .into_values()
+        .filter_map(|match_entry| match match_entry {
+            MatchEntry::Pair(r1, r2) => Some([r1, r2]),
+            _ => None,
+        })
```
akoshelev, line 80: if you create a method `MatchEntry.to_pair(self) -> Option<[...; 2]>`, this can be made a one-liner: `filter_map(MatchEntry::to_pair)`

akoshelev (commented):
I think this is close and we just need to make sure tests pass

eriktaubeneck:
yes! the test produces real outputs, but not the expected outputs. will dig in tomorrow. 

one weird edge case though. because the reports are indistinguishable, in the case a malicious client gives exactly two conversion reports with the same match key AND there is no impression report, the sum of those two values will land in the breakdown_key 0 bucket. so I think we have to remove that bucket because we can't differentiate a malicious client from real values.

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: I think this would end up adding the same report twice (which is why I added the Empty value to the enum).

akoshelev: that should not be the case: https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html#method.or_insert

andyleiserson: The problem with using `or_insert` here is that it doesn't tell you whether or not it inserted something. So afterwards, you can't distinguish an entry that already had one report, from an entry that you just added one report. But you can do it by matching on the entry itself to distinguish the empty case (untested code):

```
match reports_by_matchkey.entry(...) {
    Vacant(entry) => entry.insert(/* new single entry */),
    Occupied(entry) => entry.get_mut().add_report(...),
}
```

akoshelev: yea, match works - late night reviews are not exactly find the best of me

eriktaubeneck: yea, exactly what @andyleiserson said. I solved it with the `Empty` state on the enum, but the match looks nicer.  I didn't see the `Vacant` / `Occupied` in the rust docs, thanks for the pointer!

eriktaubeneck: actually, looks like `and_modify` is the right way to go: https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.and_modify

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: I don't think this changed due to the refactor, because on line 106, we we're always using the `.len()` method to set the total records, and I don't know how we would have done that without collecting. We don't know until we're done with the whole vector how many pairs we'll have, so we need that in order to set total records.

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: oh nice, good to know!

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,476 @@
...
+    V: BooleanArray,
+{
+    let mut reports_by_matchkey: BTreeMap<u64, MatchEntry<BK, V>> = BTreeMap::new();
+
+    for report in reports {
+        reports_by_matchkey
+            .entry(report.match_key)
+            .and_modify(|e| e.add_report(report.clone().into()))
+            .or_insert(MatchEntry::Single(report.into()));
+    }
```
eriktaubeneck, line 85: This was my resolution to [this comment thread](https://github.com/private-attribution/ipa/pull/1434#discussion_r1843249312). However it does require a clone. The alternative (that @andyleiserson suggested) would end up looking like this:

```
    for report in reports {
        match reports_by_matchkey.entry(report.match_key) {
            Entry::Vacant(entry) => {
                entry.insert(MatchEntry::Single(report.into()));
            }
            Entry::Occupied(mut entry) => {
                entry.get_mut().add_report(report.into());
            }
        }
    }
```

Using `and_modify` feels like the Right Way™ to do it, but if we think this clone is concerning.

ipa-core/src/report/hybrid.rs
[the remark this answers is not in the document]
eriktaubeneck: turned out we don't need this anymore at all. the ZERO was used to fill out an array, but we use the enum now.

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: followed up here: https://github.com/private-attribution/ipa/pull/1434#discussion_r1844722799

resolving this thread.

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -0,0 +1,524 @@
...
+                            prf_reports[0].breakdown_key +=
+                                Replicated::new(BA8::ZERO, BA8::truncate_from(1_u128));
+                        }
+
+                        aggregate_reports(ctx.clone(), prf_reports).await.unwrap()
+                    }
+                })
+                .await;
+        });
+    }
```
eriktaubeneck, line 523: I wasn't able to get the intercepter to work, but I'm able to manually XOR one of the shares, and that causes the expected failure. cc @akoshelev @andyleiserson

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: entry API has my vote - not just because of clone, but also because it is more concise

ipa-core/src/protocol/hybrid/agg.rs
akoshelev: it is possible to compute when the btree map is constructed (keep a counter that tracks entries with exactly two rows) or do another pass after (over values), but we can skip it for now.

akoshelev (approved):

PR #1435, opened 2024-11-13
implement Shuffleable for IndistinguishableHybridReport

eriktaubeneck:
much simpler than the last go around, thanks for all the improvements here @andyleiserson!

note that the share type here is less than optimal, but I ran into issues when attempting to build smaller BooleanArrays. I think I have a lead on getting that to work, but it's better addressed in a different PR.

### DIFF
--- a/ipa-core/src/report/hybrid.rs
+++ b/ipa-core/src/report/hybrid.rs
@@ -39,15 +39,21 @@ use crate::{
     const_assert_eq,
     error::{BoxError, Error},
     ff::{
-        boolean_array::{BA3, BA64, BA8},
+        boolean_array::{
+            BooleanArray, BooleanArrayReader, BooleanArrayWriter, BA112, BA3, BA64, BA8,
+        },
         Serializable,
     },
     hpke::{
         open_in_place, seal_in_place, CryptError, EncapsulationSize, PrivateKeyRegistry,
         PublicKeyRegistry, TagSize,
     },
+    protocol::ipa_prf::shuffle::Shuffleable,
     report::hybrid_info::{HybridConversionInfo, HybridImpressionInfo, HybridInfo},
-    secret_sharing::{replicated::semi_honest::AdditiveShare as Replicated, SharedValue},
+    secret_sharing::{
+        replicated::{semi_honest::AdditiveShare as Replicated, ReplicatedSecretSharing},
+        SharedValue,
+    },
     sharding::ShardIndex,
 };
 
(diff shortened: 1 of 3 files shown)


ipa-core/src/report/hybrid.rs
```diff
@@ -682,8 +688,8 @@ pub type PrfHybridReport<BK, V> = IndistinguishableHybridReport<BK, V, u64>;
 #[derive(Clone, Debug, Eq, PartialEq, Default)]
 pub struct IndistinguishableHybridReport<BK, V, MK = Replicated<BA64>>
 where
-    BK: SharedValue,
-    V: SharedValue,
+    BK: SharedValue + BooleanArray,
+    V: SharedValue + BooleanArray,
```
andyleiserson, line 692:
```suggestion
    BK: BooleanArray,
    V: BooleanArray,
```

`BooleanArray` requires `SharedValue`, so I would favor conciseness.

ipa-core/src/report/hybrid.rs
```diff
@@ -757,6 +782,54 @@ where
     }
 }
 
+impl<BK, V> Shuffleable for IndistinguishableHybridReport<BK, V>
+where
+    BK: SharedValue + BooleanArray,
+    V: SharedValue + BooleanArray,
+{
+    // this requires BK:BAXX + B:BAYY  such that XX + YY <= 48
```
andyleiserson, line 790: `V:BAYY`?

ipa-core/src/report/hybrid.rs
```diff
@@ -757,6 +782,54 @@ where
...
+    // this is checked in a debud_assert call in ::new below
+    // PERF OPPORTUNITY
+    // note that BA96 would likely be a better fit here. however, that would require a `BA128`
+    // in order to use `impl_malicious_shuffle_share!` and `BA128` cannot currently be
+    // implemented with `the boolean_array_impl!` macro as the trait `secret_sharing::Block`
+    // is not implemented for `bitvec::array::BitArray<[u8; 16]>`
```
andyleiserson, line 796: There is a macro `store_impl` to implement `Block`. The only tricky thing here is that `Block` is used both for boolean arrays and Galois field types, and they can't both attempt to implement `Block` for the same type.

andyleiserson (approved):

ipa-core/src/report/hybrid.rs
eriktaubeneck: yes, good catch!

ipa-core/src/report/hybrid.rs
eriktaubeneck: I think I have this working, but I'll double check I'm not duplicating it, and open it in a different PR. There's also a weird tradeoff here because we want it to be efficiently sized but also not too small for future implementations. we may end up wanting a macro for this implementation...

PR #1436, opened 2024-11-13
Add eval_first to LagrangeTable

tyurek:
I noticed an instance where we were calculating a bunch of dot products but only needed the first ones so I just added an easy function to address that. Hopefully saves a few cycles.

Before: https://draft-mpc.vercel.app/query/view/finer-couch2024-11-13T2343
After: https://draft-mpc.vercel.app/query/view/weary-rough2024-11-13T2313

### DIFF
--- a/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
+++ b/ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
@@ -171,9 +171,9 @@ impl<F: PrimeField, const L: usize, const P: usize, const M: usize> ProofGenerat
                 let (u_chunk, v_chunk) = polynomial.borrow();
                 (
                     // new u value
-                    lagrange_table_r.eval(u_chunk)[0],
+                    lagrange_table_r.eval_first(u_chunk),
                     // new v value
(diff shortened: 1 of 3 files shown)


ipa-core/src/protocol/ipa_prf/malicious_security/prover.rs
```diff
@@ -171,9 +171,9 @@ impl<F: PrimeField, const L: usize, const P: usize, const M: usize> ProofGenerat
                 let (u_chunk, v_chunk) = polynomial.borrow();
                 (
                     // new u value
-                    lagrange_table_r.eval(u_chunk)[0],
+                    lagrange_table_r.eval_first(u_chunk),
```
akoshelev, line 174: this table has only one row 

```rust
let lagrange_table_r = LagrangeTable::<F, L, 1>::new(&denominator, &r);
```

I am not sure if we actually do throwaway work here

tyurek: ohh, that's why this isn't speeding up at all. Gonna just pretend I didn't write this one XD

PR #1438, opened 2024-11-14
Metrics exporter to Prometheus with OTLP

shinta-liem:
Added an exporter to Prometheus with opentelemetry


ipa-metrics/Cargo.toml
```diff
@@ -13,6 +13,11 @@ partitions = []
 crossbeam-channel = "0.5"
 # This crate uses raw entry API that is unstable in stdlib
 hashbrown = "0.15"
+# Open telemetry crates: opentelemetry-prometheus crate implementation is based on Opentelemetry API and SDK 0.23. (TBC)
+opentelemetry = "0.24"
```
akoshelev, line 17: this is great, but can you make a separate crate for it? `ipa-metrics` is intended to contain only core functionality and it is easier to maintain in the long term if it does not have extra dependencies

ipa-metrics/src/exporter.rs
```diff
@@ -0,0 +1,111 @@
...
+
+mod test {
+
+    use std::thread::{self, Scope, ScopedJoinHandle};
+
+    use super::PrometheusMetricsExporter;
+    use crate::{counter, install_new_thread, producer::Producer, MetricChannelType};
+    struct MeteredScope<'scope, 'env: 'scope>(&'scope Scope<'scope, 'env>, Producer);
+
+    impl<'scope, 'env: 'scope> MeteredScope<'scope, 'env> {
```
akoshelev, line 63: you don't need this complexity, each thread has access to its own metric store throuh `CurrentThreadMetricContext`

akoshelev (commented):
this is looking very promising, thanks for working on it!

ipa-core/src/net/server/handlers/query/metrics.rs
```diff
@@ -0,0 +1,67 @@
...
+) -> Result<Vec<u8>, Error> {
+    // match transport.dispatch(query_config, BodyStream::empty()).await {
+    //     Ok(resp) => Ok(Json(resp.try_into()?)),
+    //     Err(err @ ApiError::NewQuery(NewQueryError::State { .. })) => {
+    //         Err(Error::application(StatusCode::CONFLICT, err))
+    //     }
+    //     Err(err) => Err(Error::application(StatusCode::INTERNAL_SERVER_ERROR, err)),
+    // }
+
+    // TODO: Remove this dummy metrics and get metrics for scraper from ipa-metrics::PrometheusMetricsExporter (see ipa-metrics/exporter.rs)
```
shinta-liem, line 27: @akoshelev I'm still not clear how we can pull the metrics from the logging_handle. IpaHttpClient doesn't seem to have a reference to it?

akoshelev: that's right, it is not referenced from anywhere. `helper.rs` does have [logging_handle](https://github.com/private-attribution/ipa/blob/8c201ae1b955da2e38f6a0c634e7a23925e834a3/ipa-core/src/bin/helper.rs#L298) that has a reference to this controller. So we need to plumb it through to deliver `Controller` to the handlers

ipa-core/src/net/server/handlers/query/metrics.rs
```diff
@@ -0,0 +1,28 @@
...
+    helpers::{routing::RouteId, BodyStream},
+    net::{
+        http_serde::{self},
+        Error, MpcHttpTransport,
+    },
+};
+
+/// Takes details from the HTTP request and creates a `[TransportCommand]::CreateQuery` that is sent
+/// to the [`HttpTransport`].
+async fn handler(transport: Extension<MpcHttpTransport>) -> Result<Vec<u8>, Error> {
```
akoshelev, line 14: I'd encourage to write some tests here as it is often hard to figure out why HTTP server is not responding or responds with errors. There are some examples in this folder that exercise happy path and error path

akoshelev (approved):

ipa-core/src/net/server/handlers/mod.rs
```diff
@@ -6,12 +6,14 @@ use axum::Router;
...
-            .merge(query::h2h_router(transport)),
-    )
+    echo::router()
+        .nest(
+            http_serde::query::BASE_AXUM_PATH,
+            Router::new()
+                .merge(query::query_router(transport.clone()))
+                .merge(query::h2h_router(transport.clone())),
+        )
+        .merge(query::metric_router(transport))
```
cberkhoff, line 16: Nit: Does this handler belong inside the query module? You can perhaps park it alongside `echo`

ipa-metrics-prometheus/src/exporter.rs
```diff
@@ -0,0 +1,77 @@
...
+        })
+        .join()
+        .unwrap();
+
+        let mut store = controller.snapshot().unwrap();
+
+        let mut buff = Vec::new();
+        store.export(&mut buff);
+
+        let result = String::from_utf8(buff).unwrap();
```
cberkhoff, line 74: It would be good to validate the results somehow. Also consider removing the println.

cberkhoff (approved):

PR #1439, opened 2024-11-15
Sharded Query Status

cberkhoff:
Adding functionality to obtain the global status of a query in a sharded environment using two APIs `Processor::query_status` (modified) `Processor::shard_status` (new). Since `Transport::send` and by extension `Transport::broadcast` don't return any value to the caller, we decided that `shard_status` would compare its inner state against the incoming value from the leader shard. When the states match, shard_status returns 'ok', but if they don't, it returns an error. This way, the leader shard can reply back to the caller with a unified state or a `BroadcastError` with `DifferentStatus` with the shards which didn't matched the leader.

### DIFF
--- a/ipa-core/src/query/processor.rs
+++ b/ipa-core/src/query/processor.rs
@@ -10,7 +10,7 @@ use crate::{
     error::Error as ProtocolError,
     executor::IpaRuntime,
     helpers::{
-        query::{PrepareQuery, QueryConfig, QueryInput, QueryStatusRequest},
+        query::{CompareStatusRequest, PrepareQuery, QueryConfig, QueryInput},
         BroadcastError, Gateway, GatewayConfig, MpcTransportError, MpcTransportImpl, Role,
         RoleAssignment, ShardTransportError, ShardTransportImpl, Transport,
     },
@@ -112,6 +112,12 @@ pub enum QueryStatusError {
     NoSuchQuery(QueryId),
     #[error(transparent)]
     ShardBroadcastError(#[from] BroadcastError<ShardIndex, ShardTransportError>),
+    #[error("My status {my_status:?} for query {query_id:?} differs from {other_status:?}")]
+    DifferentStatus {
+        query_id: QueryId,
+        my_status: QueryStatus,
+        other_status: QueryStatus,
+    },
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -367,7 +373,7 @@ impl Processor {
             .get_status(query_id)
             .ok_or(QueryStatusError::NoSuchQuery(query_id))?;
 
-        let shard_query_status_req = QueryStatusRequest { query_id };
+        let shard_query_status_req = CompareStatusRequest { query_id, status };
 
         shard_transport.broadcast(shard_query_status_req).await?;
 
@@ -381,10 +387,20 @@ impl Processor {
     ///
     /// ## Panics
     /// If the query collection mutex is poisoned.
-    pub fn shard_status(&self, query_id: QueryId) -> Result<QueryStatus, QueryStatusError> {
+    pub fn shard_status(
+        &self,
+        req: &CompareStatusRequest,
+    ) -> Result<QueryStatus, QueryStatusError> {
         let status = self
-            .get_status(query_id)
-            .ok_or(QueryStatusError::NoSuchQuery(query_id))?;
+            .get_status(req.query_id)
+            .ok_or(QueryStatusError::NoSuchQuery(req.query_id))?;
+        if req.status != status {
+            return Err(QueryStatusError::DifferentStatus {
+                query_id: req.query_id,
+                my_status: status,
+                other_status: req.status,
+            });
+        }
         Ok(status)
     }
 
@@ -481,6 +497,14 @@ mod tests {
         },
         sharding::ShardIndex,
     };
+
+    fn prepare_query() -> PrepareQuery {
+        PrepareQuery {
+            query_id: QueryId,
+            config: test_multiply_config(),
+            roles: RoleAssignment::new(HelperIdentity::make_three()),
+        }
+    }
 
     fn create_handler<F, Fut, I: TransportIdentity>(cb: F) -> Arc<dyn RequestHandler<I>>
     where
@@ -797,14 +821,6 @@ mod tests {
         use super::*;
         use crate::query::QueryStatusError;
 
-        fn prepare_query() -> PrepareQuery {
-            PrepareQuery {
-                query_id: QueryId,
-                config: test_multiply_config(),
-                roles: RoleAssignment::new(HelperIdentity::make_three()),
-            }
-        }
-
         #[tokio::test]
         async fn happy_case() {
             let req = prepare_query();
@@ -924,23 +940,26 @@ mod tests {
     }
 
     mod query_status {
-        use crate::protocol::QueryId;
-
         use super::*;
+        use crate::protocol::QueryId;
 
         /// * From the standpoint of leader shard in Helper 1
         /// * On query_status
-        /// 
+        ///
         /// If one of my shards isn't ready
         #[tokio::test]
         async fn combined_status_response() {
             fn shard_handle(si: ShardIndex) -> Arc<dyn RequestHandler<ShardIndex>> {
                 create_handler(move |_| async move {
                     match si {
-                        ShardIndex(1) => Ok(HelperResponse::from(QueryStatus::AwaitingInputs)),
-                        ShardIndex(2) => Ok(HelperResponse::from(QueryStatus::Running)),
-                        ShardIndex(3) => Ok(HelperResponse::from(QueryStatus::Completed)),
-                        _ => Ok(HelperResponse::from(QueryStatus::Running))
+                        ShardIndex(3) => {
+                            Err(ApiError::QueryStatus(QueryStatusError::DifferentStatus {
+                                query_id: QueryId,
+                                my_status: QueryStatus::Completed,
+                                other_status: QueryStatus::Preparing,
+                            }))
+                        }
+                        _ => Ok(HelperResponse::ok()),
                     }
                 })
             }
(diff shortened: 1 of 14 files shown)


ipa-core/src/query/processor.rs
```diff
@@ -349,17 +361,60 @@ impl Processor {
...
+    pub fn shard_status(
+        &self,
+        shard_transport: &ShardTransportImpl,
+        req: &CompareStatusRequest,
+    ) -> Result<QueryStatus, QueryStatusError> {
+        let shard_index = shard_transport.identity();
+        if shard_index == ShardIndex::FIRST {
+            return Err(QueryStatusError::Leader);
+        }
+        let status = self
```
akoshelev, line 408: is Codecov hallucinating again or we don't have coverage here?

ipa-core/src/query/processor.rs
```diff
@@ -349,17 +361,60 @@ impl Processor {
...
+            return Err(QueryStatusError::NotLeader(shard_index));
+        }
+
         let status = self
             .get_status(query_id)
             .ok_or(QueryStatusError::NoSuchQuery(query_id))?;
+
+        let shard_query_status_req = CompareStatusRequest { query_id, status };
+
+        shard_transport.broadcast(shard_query_status_req).await?;
```
akoshelev, line 387: Using errors to communicate between shards is a workaround of our transport trait limitation but we don't have to do it with report collector. The error they will be getting is confusing - they asked for a query status, but got an error "DifferentStatus". We should report the "min" status back to them - if all shards except one is ready, we tell them pending. If at least one error occurred, we say error, etc.

akoshelev (changes requested):

ipa-core/src/query/processor.rs
cberkhoff: I agree. I wasn't sure about this approach for the RC collection. I only went for it because it was a clear implementation and the alternative didn't seemed a lot better. What I had in mind was a new `QueryStatus` called `MixedStatus`. This one was only used when returning for this global state. It didn't felt al that alright to have a state that was only used to reply to a global (sharded) state, not internally. Let me know if this is the route you want to take.

A more involved solution, depending on the current status of the leader vs the broadcast response would look something like the table below:


| Leader Status  | Response if DifferentStatus | Response If query isn't found | Notes |
| ------------- | ------------- | --- | --- |
| None  |  | special | If query IS found in other shard but not in leader we should return an `InconsistentGlobalStateError` (new) |
| Preparing  | Preparing  | Preparing | maybe the query status got earlier than the prepare_shard |
| Awaiting Inputs | Awaiting Inputs | `InconsistentGlobalStateError` |  |
| Running | Running | `InconsistentGlobalStateError` |  |
| Awaiting Completion | Running | `InconsistentGlobalStateError` |  |
| Completed | Running | `InconsistentGlobalStateError` |  |

For this table I'm assuming the helper is going through the states "normally". 

I'm concerned about overcomplicating this thing though.

akoshelev: I think we can just use the existing QueryStatus enum and encode these transitions somewhere. It does not feel that this needs to be complicated

cberkhoff: I'm not prescribing we're the implementation will be, just trying to clarify what each status will be. Can you please be more specific on the "complications"?

akoshelev: > I'm concerned about overcomplicating this thing though. 

it is basically a response to this

cberkhoff: One option is to not use the fact that a query exists (or not) in other shards. 

What I don't like about that though is that the RC will be able to submit queries even if some shard is still stuck in some state (because the leader won't be checking for query existence).

Any specific suggestions?

akoshelev: I think just returning an error in this case may be enough for now? as we said failure recovery is complicated and we may want to revisit it later

ipa-core/src/query/processor.rs
cberkhoff: I think we do have tests for this...

ipa-core/src/query/processor.rs
cberkhoff: So, instead of sending an error back, we just wrap the error into a non-error response message back?

ipa-core/src/query/processor.rs
akoshelev: is it possible to check that by adding a panic and observing that one or more tests fail?

ipa-core/src/query/processor.rs
```diff
@@ -349,17 +362,111 @@ impl Processor {
...
+        }
+        None
+    }
+
+    /// This helper is used by the in-memory stack to obtain the state of other shards via a
+    /// [`QueryStatusError::DifferentStatus`] error.
+    /// TODO: Ideally broadcast should return a value, that we could use to parse the state instead
+    /// of relying on errors.
+    #[cfg(feature = "in-memory-infra")]
+    fn get_state_from_error(
```
akoshelev, line 385: there are a few issues here
- This logic really belongs to `BroadcastError`. 
- what happens if we compile this with real-world-infra flag?

ipa-core/src/query/processor.rs
```diff
@@ -349,17 +362,111 @@ impl Processor {
...
+
+        let mut status = self
             .get_status(query_id)
             .ok_or(QueryStatusError::NoSuchQuery(query_id))?;
+
+        let shard_query_status_req = CompareStatusRequest { query_id, status };
+
+        let shard_responses = shard_transport.broadcast(shard_query_status_req).await;
+        if let Err(e) = shard_responses {
+            // The following silently ignores the cases where the query isn't found because those
```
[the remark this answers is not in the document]
akoshelev: there are a few ways to "ignore" the problem - we don't have to implement code to properly handle it, but choosing between swiping it under the rug and panicking, I feel I choose the latter.

ipa-core/src/query/processor.rs
```diff
@@ -362,16 +362,45 @@ impl Processor {
         Some(status)
     }
 
+    /// This helper function is used to transform a [`BoxError`] into a
+    /// [`QueryStatusError::DifferentStatus`] and retrieve it's internal state. Returns [`None`]
+    /// if not possible.
+    fn downcast_state_error(be: BoxError) -> Option<QueryStatus> {
+        let ae = be.downcast::<crate::helpers::ApiError>().ok()?;
```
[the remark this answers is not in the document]
cberkhoff: gave longer names

ipa-core/src/test_fixture/app.rs
```diff
@@ -118,12 +119,13 @@ impl TestApp {
...
     /// ## Panics
     /// Never.
-    pub fn query_status(&self, query_id: QueryId) -> Result<[QueryStatus; 3], ApiError> {
-        Ok((0..3)
-            .map(|i| self.drivers[i].query_status(query_id))
-            .collect::<Result<Vec<_>, _>>()?
-            .try_into()
-            .unwrap())
+    #[allow(clippy::disallowed_methods)]
+    pub async fn query_status(&self, query_id: QueryId) -> Result<[QueryStatus; 3], ApiError> {
```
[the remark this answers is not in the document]
cberkhoff: sounds good

ipa-core/src/query/state.rs
```diff
@@ -48,6 +48,21 @@ impl From<&QueryState> for QueryStatus {
     }
 }
 
+#[must_use]
+pub fn min_status(a: QueryStatus, b: QueryStatus) -> QueryStatus {
```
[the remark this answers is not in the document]
cberkhoff: makes sense

ipa-core/src/query/state.rs
```diff
@@ -226,3 +241,24 @@ impl Drop for RemoveQuery<'_> {
         }
     }
 }
+
+#[cfg(all(test, unit_test))]
+mod tests {
+    use crate::query::{state::min_status, QueryStatus};
+
+    #[test]
+    fn test_order() {
```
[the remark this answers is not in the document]
cberkhoff: ok, thanks for the provided code, it helps!

ipa-core/src/query/processor.rs
```diff
@@ -349,17 +362,111 @@ impl Processor {
...
+        let mut status = self
             .get_status(query_id)
             .ok_or(QueryStatusError::NoSuchQuery(query_id))?;
+
+        let shard_query_status_req = CompareStatusRequest { query_id, status };
+
+        let shard_responses = shard_transport.broadcast(shard_query_status_req).await;
+        if let Err(e) = shard_responses {
+            // The following silently ignores the cases where the query isn't found because those
+            // errors return `None` for [`BroadcasteableError::peer_state()`]
```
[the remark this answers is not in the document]
cberkhoff: nothing, removed the mention

ipa-core/src/query/processor.rs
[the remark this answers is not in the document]
cberkhoff: No, that's not what I mean.

When receiving the broadcast for query status, other shards can return "I don't have that query"... and that's fine. When returning the unified response we want to say that there IS a query (hence we "ignore" the response from the shards that don't have the query yet).

ipa-core/src/query/processor.rs
cberkhoff: 1. I don't think this belongs to BroadcastError. In fact, none of these new functions (`get_state_from_error` and `downcast_state_error`) have to deal with that type at all. Only `query_status` on broadcast.

2. Having 2 implementations under the same name won't compile. This is the same as `gateway::MpcTransportImpl`.
Also `InMemoryTransportError` won't be an available symbol in real-world-infra.

akoshelev: the semantic of error downcasting does not seem to be query processor job. Likely we would need to do more for real world

> Having 2 implementations under the same name won't compile. This is the same as gateway::MpcTransportImpl.

is it going to be an issue when we implement this for HTTP?

akoshelev (approved):

ipa-core/src/query/processor.rs
cberkhoff: Don't think I'm following, let's chat about this one.

The http implementation for the downcast is here. As long as we use AplicationError this should handle it. We can change it to be something more specific as well.

PR #1440, opened 2024-11-15
Use recursion factor of 4 for all proofs

andyleiserson:
Cleaned up version of #1407.

Performance was similar for recursion factors of 4 and 8. However, a recursion factor of 4 (at least for the first proof) makes it easier to implement the lookup table for Lagrange interpolation output in #1437, so that's what I've used here.

I change the calculation of proof batches to adjust downwards (rather than upwards) from the target to a power of two, so that there doesn't need to be a large margin between TARGET_PROOF_SIZE and the limit defined by MAX_PROOF_RECURSION.

I also change the names `LargeProofGenerator` and `SmallProofGenerator` to `FirstProofGenerator` and `CompressedProofGenerator`.


andyleiserson:
One more draft run at 7a7bdb1: https://draft-mpc.vercel.app/query/view/alien-fuse2024-11-16T2033

We can also try using λ = 4 for the first proof with λ = 8 for the other proofs. I doubt it will make much of a difference, but it's an easy experiment.

ipa-core/src/protocol/context/dzkp_validator.rs
```diff
@@ -50,11 +50,22 @@ const BIT_ARRAY_SHIFT: usize = BIT_ARRAY_LEN.ilog2() as usize;
...
 #[cfg(test)]
 pub const TARGET_PROOF_SIZE: usize = 8192;
 #[cfg(not(test))]
 pub const TARGET_PROOF_SIZE: usize = 50_000_000;
 
+/// Returns the largest power of two less than or equal to `self`.
+///
+/// Returns 0 if the input is outside the range `1..1 << (usize::BITS - 1)`.
+#[must_use]
+pub const fn prev_power_of_two(value: usize) -> usize {
```
akoshelev, line 65: there is another version of it [here](https://github.com/private-attribution/ipa/blob/8e61beebad2ed498c3b3b108c581903534a16aaf/ipa-core/src/helpers/gateway/send.rs#L261). 
Maybe we can follow up to merge them

akoshelev (approved):

PR #1442, opened 2024-11-16
[DRAFT] Proof recursion 8 redux

andyleiserson:


ipa-core/src/protocol/ipa_prf/validation_protocol/proof_generation.rs
```diff
@@ -80,39 +81,82 @@ impl ProofBatch {
...
+        });
+
+        let first_proof = FirstProofGenerator::compute_proof(table_index_tuple_inputs.chunk_array::<2>().map(
+            |chunk| {
+                let [(u0, v0), (u1, v1)] = chunk;
+                let u = array::from_fn(|i| table_right_0[usize::from(u0)][i] + table_right_1[usize::from(u1)][i]);
+                let v = array::from_fn(|i| table_left_0[usize::from(v0)][i] + table_left_1[usize::from(v1)][i]);
+                (u, v)
+            },
+        ));
```
andyleiserson, line 144: This could be optimized by deferring the reductions in these additions to happen at the same time as the reductions for the multiply-accumulate in `compute_proof`.

PR #1444, opened 2024-11-19
copy IPA breakdown_key reveal verbatim

eriktaubeneck:
nothing really to review here, just a copy of the breakdown_reveal.rs file to make #1445  easier to review.


akoshelev (approved):
I'd probably mention in the comments that this is the sharded version of this protocol and put some placeholders to fill in later where we would describe how it works

eriktaubeneck:
> I'd probably mention in the comments that this is the sharded version of this protocol and put some placeholders to fill in later where we would describe how it works 

I can add the shared part, but the description is there in #1445

PR #1445, opened 2024-11-19
Breakdown key reveal

eriktaubeneck:
This implements `breakdown_reveal` for hybrid, which covers padding reports, shuffling them, revealing the breakdown key, and then adding the secret shares of the values into histograms.

Note that this is stacked on top of #1444, but in the meantime, you can look just at commit `5dbaa63` to highlight the changes.

### DIFF
--- a/ipa-core/src/protocol/hybrid/breakdown_reveal.rs
+++ b/ipa-core/src/protocol/hybrid/breakdown_reveal.rs
@@ -12,7 +12,7 @@ use crate::{
         basics::{reveal, Reveal},
         context::{
             dzkp_validator::DZKPValidator, Context, DZKPUpgraded, MaliciousProtocolSteps,
-            UpgradableContext,
+            ShardedContext, UpgradableContext,
         },
         ipa_prf::{
             aggregation::{
@@ -66,7 +66,7 @@ pub async fn breakdown_reveal_aggregation<C, BK, V, HV, const B: usize>(
     padding_params: &PaddingParameters,
 ) -> Result<BitDecomposed<Replicated<Boolean, B>>, Error>
 where
-    C: UpgradableContext + Shuffle,
+    C: UpgradableContext + Shuffle + ShardedContext,
     Boolean: FieldSimd<B>,
     Replicated<Boolean, B>: BooleanProtocols<DZKPUpgraded<C>, B>,
     BK: BooleanArray + U128Conversions,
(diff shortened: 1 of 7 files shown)


ipa-core/src/protocol/hybrid/breakdown_reveal.rs
```diff
@@ -127,30 +60,29 @@ where
...
-        )
-        .await?;
+    let attributed_values_padded = apply_dp_padding::<_, AggregateableHybridReport<BK, V>, B>(
+        ctx.narrow(&Step::PaddingDp),
+        attributed_values,
+        padding_params,
+    )
+    .await?;
 
     let attributions = ctx
```
akoshelev, line 87: are we using sharded shuffle here?

ipa-core/src/protocol/hybrid/breakdown_reveal.rs
```diff
@@ -302,98 +231,38 @@ where
...
-            let result = result.iter().map(|&v| v.as_u128()).collect::<Vec<_>>();
-            assert_eq!(32, result.len());
-            assert_eq!(result, expectation);
-        });
-    }
-
-    #[test]
-    #[cfg(not(feature = "shuttle"))] // too slow
-    fn malicious_happy_path() {
+    fn breakdown_reveal_malicious_happy_path() {
```
akoshelev, line 265: is there a test that ensures it works for more than one shard?

ipa-core/src/protocol/hybrid/breakdown_reveal.rs
eriktaubeneck: good catch. added `ShardedContext` to the type bounds for the context, so we should be now.

ipa-core/src/protocol/hybrid/breakdown_reveal.rs
eriktaubeneck: updated!

ipa-core/src/protocol/hybrid/breakdown_reveal.rs
```diff
@@ -302,101 +231,45 @@ where
     }
 }
 
-#[cfg(all(test, any(unit_test, feature = "shuttle")))]
+#[cfg(all(test, unit_test, not(feature = "shuttle")))]
```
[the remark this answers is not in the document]
eriktaubeneck: shuttle was turned off for malicious in the IPA side (but there was a semi-honest version where it wasn't). we can replicate that here if we think that's useful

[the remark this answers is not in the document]
akoshelev: yes I think it would be useful not to lose coverage on this code path, even if it only works for semi-honest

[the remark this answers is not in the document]
eriktaubeneck: added!

akoshelev (approved):

PR #1446, opened 2024-11-19
Finalizer protocol for sharded executions

akoshelev:
for Hybrid protocol we will send all histograms back to the leader shard where they will be aggregated and returned back to the report collector.

For sharded shuffle (test protocol) we will send rows/reports back to the leader to concatenate them all together. 

Both of these will be supported by the newly added `ShardAssembledResult` trait, while this PR only brings the implementation for Hybrid


ipa-core/src/protocol/basics/shard_fin.rs
```diff
@@ -0,0 +1,158 @@
...
+    type SingleMessage = AdditiveShare<HV>;
+
+    fn empty() -> Self {
+        Vec::new()
+    }
+
+    fn merge(&mut self, other: Self) {
+        // this merges two histograms together by adding them up
+        for (a, b) in self.iter_mut().zip(other) {
+            *a += b;
```
eriktaubeneck, line 107: won't this just XOR the shares? I think we may need to use `integer_add` here.

ipa-core/src/protocol/basics/shard_fin.rs
```diff
@@ -0,0 +1,158 @@
...
+        test_executor::run,
+        test_fixture::{Reconstruct, Runner, TestWorld, TestWorldConfig, WithShards},
+    };
+
+    #[test]
+    fn shards_set_result() {
+        run(|| async {
+            let world: TestWorld<WithShards<3>> =
+                TestWorld::with_shards(TestWorldConfig::default());
+            let input = vec![BA64::ZERO, BA64::ZERO, BA64::ZERO];
```
eriktaubeneck, line 139: should probably test with some real values to make sure they add correctly.

eriktaubeneck (approved):
overall this looks good, but one comment about how it should work for histograms

ipa-core/src/protocol/basics/shard_fin.rs
```diff
@@ -0,0 +1,436 @@
...
+        Ok(R::empty())
+    }
+}
+
+/// This type exists to bind [`HV`] and [`B`] together and allow
+/// conversions from [`AdditiveShare<HV>`] to [`BitDecomposed<AdditiveShare<Boolean, B>>`]
+/// and vice versa. Decomposed view is used to perform additions,
+/// share is used to send data to other shards.
+#[derive(Debug, Default)]
+struct Histogram<HV: BooleanArray, const B: usize>
```
eriktaubeneck, line 199: as an aside, we should probably refactor the `breakdown_reveal` and `apply_dp_noise` to both take and return this type. the fact that one returns `BitDecomposed<AdditiveShare<Boolean, B>>` and the other returns `AdditiveShare<HV>` seems like it could use a touch up.

ipa-core/src/protocol/basics/shard_fin.rs
```diff
@@ -0,0 +1,436 @@
...
+        ctx: C,
+        record_id: RecordId,
+        other: Self,
+    ) -> impl Future<Output = Result<(), Error>> + Send + 'a
+    where
+        C: 'a,
+    {
+        async move {
+            // todo: EightBit only works for 256 breakdowns. EightBitStep will panic if we try
+            // to add larger values
```
eriktaubeneck, line 279: this isn't the only place this is an issue. this works for now, but we might want to think about a design that binds the step size to the number of breakdowns.

ipa-core/src/protocol/basics/mod.rs
```diff
@@ -3,6 +3,8 @@ mod if_else;
 pub(crate) mod mul;
 mod reshare;
 mod reveal;
+#[allow(dead_code)]
+mod shard_fin;
```
eriktaubeneck, line 7: the `protocol/hybrid/mod.rs` file is ready to have this included.

eriktaubeneck (approved):
nice! a couple small comments but overall looks good

ipa-core/src/protocol/basics/shard_fin.rs
akoshelev: yea, I think the reason for that is we want to keep the inputs to DP noise vectorized for faster additions. DP noise produces final output that does not have to be vectorized, so it uses shares. 

Hybrid would need to do the DP noise addition only on the leader, after this finalization step completes, so it would still probably want bit decomposed view

ipa-core/src/protocol/basics/shard_fin.rs
akoshelev: yep, it will panic if we add more than 8 bits, so it is not entirely out of control. But I agree - it is far less than ideal

ipa-core/src/protocol/basics/mod.rs
akoshelev: clippy will still yell at me, so I'll enable it in a follow up

akoshelev:
I made a small change to get rid of `Step` associated trait bound and use `MaliciousSteps` concrete type. It turns out (thanks sharded shuffle) that it is not possible to provide this type when working with generic contexts, unless we want extra bounds on it.

As we use malicious steps anyway in both implementations, it is just easier to manage this way

PR #1447, opened 2024-11-20
Complete query API for sharded environments

akoshelev:
As we discussed, we want to make the complete query API simple to implement. That means, we have an assumption that all shards communicate their results to the leader. That leaves complete API to just clean up the state on each shard and return response from the leader. This PR does exactly that.


ipa-core/src/app.rs
```diff
@@ -251,7 +251,10 @@ impl RequestHandler<HelperIdentity> for Inner {
             }
             RouteId::CompleteQuery => {
```
cberkhoff, line 252: I think we also need to add `qp.complete(...)` to the Shard Handler?

ipa-core/src/query/processor.rs
```diff
@@ -468,11 +486,11 @@ mod tests {
     }
 
     fn helper_respond_ok() -> Arc<dyn RequestHandler<HelperIdentity>> {
-        prepare_query_handler(|_| async { Ok(HelperResponse::ok()) })
+        make_owned_handler(move |_req, _| futures::future::ok(HelperResponse::ok()))
     }
 
     fn shard_respond_ok(_si: ShardIndex) -> Arc<dyn RequestHandler<ShardIndex>> {
-        prepare_query_handler(|_| async { Ok(HelperResponse::ok()) })
+        make_owned_handler(move |_req, _| futures::future::ok(HelperResponse::ok()))
```
cberkhoff, line 493: we will have git conflicts here with the query_status API...
what approach do you want to keep?

ipa-core/src/query/processor.rs
```diff
@@ -755,6 +804,81 @@ mod tests {
...
+            });
+
+            let t = TestComponents::new(args);
+            let query_id = t.new_running_query().await;
+
+            let _ = t
+                .processor
+                .complete(query_id, t.shard_transport.clone_ref())
+                .await
+                .unwrap_err();
```
cberkhoff, line 853: Can you check the error type as well?

ipa-core/src/query/processor.rs
```diff
@@ -755,6 +804,81 @@ mod tests {
...
+
+        #[tokio::test]
+        async fn complete_basic() {
+            let t = TestComponents::default();
+            let query_id = t.new_running_query().await;
+
+            t.processor
+                .complete(query_id, t.shard_transport.clone_ref())
+                .await
+                .unwrap();
```
cberkhoff, line 829: Do we also need to validate the return value?

ipa-core/src/query/processor.rs
```diff
@@ -755,6 +804,81 @@ mod tests {
...
+        #[tokio::test]
+        async fn only_leader_broadcasts() {
+            let mut args = TestComponentsArgs::default();
+
+            args.set_shard_handler(|shard_id| {
+                make_owned_handler(move |_req, _| {
+                    if shard_id == ShardIndex::FIRST {
+                        futures::future::err(ApiError::BadRequest(
+                            "Leader shard must not receive requests through shard channels".into(),
+                        ))
```
cberkhoff, line 865: does this need to be a `panic` instead to represent an unexpected bug?

ipa-core/src/query/state.rs
```diff
@@ -60,13 +60,14 @@ pub enum QueryState {
...
-        use QueryState::{AwaitingInputs, Empty, Preparing};
+        use QueryState::{AwaitingInputs, Empty, Preparing, Running};
 
         match (cur_state, &new_state) {
             // If query is not running, coordinator initial state is preparing
             // and followers initial state is awaiting inputs
             (Empty, Preparing(_) | AwaitingInputs(_, _, _))
-            | (Preparing(_), AwaitingInputs(_, _, _)) => Ok(new_state),
+            | (Preparing(_), AwaitingInputs(_, _, _))
+            | (AwaitingInputs(_, _, _), Running(_)) => Ok(new_state),
```
cberkhoff, line 70: How did this work before? I'm surprised we only had to add the "awaiting to running" transition just now...

cberkhoff (changes requested):

ipa-core/src/query/processor.rs
akoshelev: this one makes more sense to me as OK response for any request, not just prepare_query

cberkhoff: I mean this lines on my PR:

https://github.com/private-attribution/ipa/pull/1439/files#diff-49e05834cf3a03f064b79804a2c10dd7fa9606b61835ae9f71192e06cd6a8935R534-R548

ipa-core/src/net/server/handlers/mod.rs
```diff
@@ -17,6 +17,8 @@ pub fn mpc_router(transport: MpcHttpTransport) -> Router {
 pub fn shard_router(transport: ShardHttpTransport) -> Router {
     echo::router().nest(
         http_serde::query::BASE_AXUM_PATH,
-        Router::new().merge(query::s2s_router(transport)),
+        Router::new()
+            .merge(query::c2s_router(&transport))
```
cberkhoff, line 21: I would call it something like `shard_query_router` to match the one in mpc

ipa-core/src/net/server/handlers/query/mod.rs
```diff
@@ -55,10 +58,18 @@ pub fn h2h_router(transport: MpcHttpTransport) -> Router {
...
+        .merge(prepare::router(Arc::clone(&transport.inner_transport)))
+        .merge(results::router(transport.inner_transport))
         .layer(layer_fn(HelperAuthentication::<_, Shard>::new))
 }
 
+/// Client-to-shard routes. There are only a few cases where we expect parties
+/// to talk to individual shards. Input submission is one of them. This path does
+/// not require cert authentication.
+pub fn c2s_router(transport: &ShardHttpTransport) -> Router {
+    Router::new().merge(input::router(Arc::clone(&transport.inner_transport)))
```
cberkhoff, line 70: Why not leave input in the MPC routes, minimizing the number of changes?

I don't see advantage on sending this to the shard port instead.

ipa-core/src/net/test.rs
```diff
@@ -305,6 +305,19 @@ impl TestConfig {
         }
     }
 
+    /// Returns full set of clients to talk to each individual shard in the MPC.
+    #[must_use]
+    pub fn shard_clients(&self) -> [Vec<IpaHttpClient<Shard>>; 3] {
```
cberkhoff, line 310: You won't be able to set ClientIdentity from here. Is this a good place for this method?

I think it would be better to expose something you can iterate over. Something like:
```
pub fn shard_network_confs(&self) -> [NetworkConfig<Shard>; 3] {
...
}
```

Then on the tests `test_make_helpers` you can create the shard configs in a similar pattern to MPC:
```
let shard_clients = conf.shard_network_confs().map(|conf| {
IpaHttpClient::shards_from_conf(
                &IpaRuntime::current(),
                &conf,
                &ClientIdentity::None,
            )
})
```
Gives the test the option to set the settings as needed like MPC.

cberkhoff (changes requested):
There were also a few comments from the previous round that I don't think they were replied to

ipa-core/src/query/state.rs
akoshelev: this transition was managed entirely inside the `complete` handler. There are quite a few transitions involved, so it was never encoded here

ipa-core/src/net/server/handlers/mod.rs
akoshelev: I don't mind either. Regardless of scheme we choose here, the naming will be inconsistent. We either deviate from `s2s` or from MPC counterpart. Let me know which one you prefer

ipa-core/src/net/server/handlers/query/mod.rs
akoshelev: good catch, this is actually not required - we don't want clients to get shard clients. I removed it and that simplified this change by quite a bit

ipa-core/src/net/test.rs
akoshelev: yea, it is no longer relevant as we don't need shard clients anymore

ipa-core/src/net/server/handlers/mod.rs
akoshelev: ok, we don't need this anymore

ipa-core/src/app.rs
akoshelev: yep, this was addressed

akoshelev:
Draft is showing green light: https://draft-mpc.vercel.app/query/view/more-patio2024-11-21T1941

PR #1448, opened 2024-11-20
Create an abstraction for multiply-accumulate

andyleiserson:
Create an abstraction for multiply-accumulate and provide an optimized implementation for `Fp61BitPrime`. Currently it is still used only in `lagrange::dot_product`, but it will soon be used in `compute_proof` as well.

This change by itself is not meant to affect performance (up or down), it is meant to encapsulate the optimized implementation so that we don't need similar code working with field internals in both `dot_product` and `compute_proof` (and anywhere else we might use it in the future).

Without the `#[inline]` directives, this seemed to hurt performance in my local testing. After adding the `#[inline]` directives, I do not see a significant performance impact either locally or in draft ([before](https://draft-mpc.vercel.app/query/view/void-job2024-11-19T0059) / [after](https://draft-mpc.vercel.app/query/view/busty-poor2024-11-19T1758)) (these runs do include other changes not part of this PR).

Draft: https://draft-mpc.vercel.app/query/view/each-hurt2024-11-20T0056, 2:30.
The runs in #1440 were 2:27 and 2:23.

### DIFF
--- a/ipa-core/src/ff/prime_field.rs
+++ b/ipa-core/src/ff/prime_field.rs
@@ -6,7 +6,10 @@ use subtle::{Choice, ConstantTimeEq};
 use super::Field;
 use crate::{
     const_assert,
-    ff::{Serializable, U128Conversions},
+    ff::{
+        accumulator::{Accumulator, MultiplyAccumulate},
+        Serializable, U128Conversions,
+    },
     impl_shared_value_common,
     protocol::prss::FromRandomU128,
     secret_sharing::{Block, FieldVectorizable, SharedValue, StdArray, Vectorizable},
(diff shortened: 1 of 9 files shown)


ipa-core/src/ff/prime_field.rs
```diff
@@ -518,6 +537,14 @@ mod fp32bit {
 mod fp61bit {
     field_impl! { Fp61BitPrime, u64, u128, 61, 2_305_843_009_213_693_951 }
 
+    // For multiply-accumulate of `Fp61BitPrime` using `u128` as the accumulator, we can add 64
+    // products onto an original field element before reduction is necessary. i.e., 64 * (2^61 - 2)^2 +
+    // (2^61 - 2) < 2^128.
```
akoshelev, line 542: is the math here correct? I was wrong many times last couple of weeks, it feels that a sum of 64 products would be 64 * (2^61 - 1)^2

akoshelev (approved):

ipa-core/src/ff/prime_field.rs
andyleiserson: 2^61 - 1 is the prime, so 2^61 - 2 is the largest value a field element can have.

That aside, the math here is slightly different than before, because this is implementing an accumulator. The maximum value adds an additional 2^61 - 2 (not squared), representing the initial value of the accumulator (or the value after the previous reduction).

I did add some tests related to this, but that's not necessarily going to catch problems, since there are assumptions about the conditions where the maximum value occurs baked into the test.

PR #1452, opened 2024-11-23
Adding support to create sharded configurations

cberkhoff:
This PR contains the following changes

1) Created a new script `create-sharded-conf.py` which simply calls `helper keygen` in 2 loops to generate keys for all shards in all helpers.

2) Added a new subcommand to `helper` binary: `sharded-confgen`. This is similar to `confgen` (they share some code) but works for sharded environment. It's meant to be used in tandem with the new `create-sharded-conf.py` script.

3) Refactor: Both `clientconf.rs` and `config.rs` were becoming "polluted" with code related to parsing toml files. I decided to create a new module just for parsing; `config_parse.rs` which is used by `clientconf.rs`.  `config.rs` now is simpler and doesn't bother "how" configs are created outside of constructors. All the tests and logic remains the same.

4) Created a little extension trait `HelperNetworkConfigParseExt`  so that parts of the code that use `NetworkConfig<Helper>` can still call function `from_toml_str`, but using the new logic to parse.


cberkhoff:
the failed test is a flaky one, not a result from this change

ipa-core/src/cli/clientconf.rs
```diff
@@ -72,162 +72,157 @@ pub fn setup(args: ConfGenArgs) -> Result<(), BoxError> {
...
+            let id_nr: u8 = id.into();
+            shard_dir.push(format!("helper{id_nr}"));
+            shard_dir.push(format!("shard{ix}"));
+
+            let host_name = find_file_with_extension(&shard_dir, "pem").unwrap();
+            let tls_cert_file = shard_dir.join(format!("{host_name}.pem"));
+            let mk_public_key_file = shard_dir.join(format!("{host_name}_mk.pub"));
+            println!("Found {host_name} for helper {id_nr} shard {ix}:");
+            println!("\ttls_cert_file: {tls_cert_file:?}");
+            println!("\tmk_public_key_file: {mk_public_key_file:?}");
```
akoshelev, line 167: either use tracing or drop them maybe?

ipa-core/src/cli/clientconf.rs
```diff
@@ -72,162 +72,157 @@ pub fn setup(args: ConfGenArgs) -> Result<(), BoxError> {
...
+) -> impl Iterator<Item = HelperClientConf> {
+    (0..shard_count).flat_map(move |ix| {
+        let base_dir = keys_dir.clone();
+        HelperIdentity::make_three().into_iter().map(move |id| {
+            let mut shard_dir = base_dir.clone();
+            let id_nr: u8 = id.into();
+            shard_dir.push(format!("helper{id_nr}"));
+            shard_dir.push(format!("shard{ix}"));
+
+            let host_name = find_file_with_extension(&shard_dir, "pem").unwrap();
```
akoshelev, line 162: can we assume that the full file name is known and simplify this? We wouldn't need file search in this case

akoshelev (approved):

ipa-core/src/cli/clientconf.rs
cberkhoff: Discussed offline. We use the filename to pass the host name.

akoshelev (approved):

PR #1453, opened 2024-11-23
Implement e2e sharded shuffle test

akoshelev:
This is not an integration test, but it works all the way through HTTP stack similar to how test multiply works. This proves that Rust server and client code work to handle all interactions between shards and helpers.


cberkhoff (approved):
Thanks for filling the gaps in transport and handlers

PR #1455, opened 2024-11-25
update hybrid generator to provide max 1 conversion per impression

eriktaubeneck:
With the hybrid protocol, we only expect at most 1 conversion per impression. We're likely to still have cases where there are unmatched impression or conversions, but (aside from a malicious client), we'll never encounter multiple conversions per impression. The protocol filters out any match key with more than 2 reports associated (assuming it has 1 of each, though that's not confirmable), so event gen should match that case.


akoshelev (approved):
We discussed that we don't need multiple conversions per impression code as honest clients never generate that. For now, we focus on the main use case and we can get that code back if needed

PR #1456, opened 2024-11-26
Make the boolean_array_impl_large macro more flexible

andyleiserson:
This makes it usable e.g. for `BA4096`, but I'm not actually adding any large BA types in this PR.

### DIFF
--- a/ipa-core/src/ff/boolean_array.rs
+++ b/ipa-core/src/ff/boolean_array.rs
@@ -1,11 +1,14 @@
-use std::fmt::{Debug, Formatter};
+use std::{
+    fmt::{Debug, Formatter},
+    iter::repeat_with,
+};
 
 use bitvec::{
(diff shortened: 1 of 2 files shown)


ipa-core/src/ff/boolean_array.rs
```diff
@@ -917,40 +920,34 @@ macro_rules! boolean_array_impl {
...
 
         $crate::const_assert!(
-            $bits <= 256,
-            "Large BooleanArrays only support up to 256 bits."
+            $bits % 8 == 0,
+            "Large BooleanArrays only support whole bytes."
         );
 
-        // used to convert into Fp25519
-        impl From<(u128, u128)> for $name {
```
andyleiserson, line 929: Use of this for `Fp25519` was removed some time ago. The only remaining use was a few tests in `dzkp_validator`, which could easily be changed to use something else.

akoshelev (approved):

PR #1459, opened 2024-11-26
Adding Report Collector docker creation

cberkhoff:
Also adding some utilities to the docker. I can remove them after we're done setting up the sharded environment.

### DIFF
--- a/docker/report_collector.Dockerfile
+++ b/docker/report_collector.Dockerfile
@@ -3,15 +3,15 @@ ARG SOURCES_DIR=/usr/src/ipa
 FROM rust:bookworm AS builder
 ARG SOURCES_DIR
 
-# Prepare helper binaries
+# Prepare report collector binaries
 WORKDIR "$SOURCES_DIR"
 COPY . .
 RUN set -eux; \
-    cargo build --bin helper --release --no-default-features --features "web-app real-world-infra compact-gate"
+    cargo build --bin report_collector --release --no-default-features --features "cli test-fixture web-app real-world-infra compact-gate"
 
 # Copy them to the final image
 FROM rust:slim-bookworm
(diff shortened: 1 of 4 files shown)


docker/report_collector.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN set -eux; \
+    cargo build --bin report_collector --release --no-default-features --features "cli test-fixture web-app real-world-infra compact-gate"
+
+# Copy them to the final image
+FROM rust:slim-bookworm
+ENV RC_BIN_PATH=/usr/local/bin/report_collector
+ENV CONF_DIR=/etc/ipa
+ARG SOURCES_DIR
+
+RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
```
akoshelev, line 18: I am not sure if we need certs on RC. Helpers need them to start the web servers

docker/report_collector.Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1
+ARG SOURCES_DIR=/usr/src/ipa
+FROM rust:bookworm AS builder
+ARG SOURCES_DIR
+
+# Prepare report collector binaries
+WORKDIR "$SOURCES_DIR"
+COPY . .
+RUN set -eux; \
```
akoshelev, line 9: report collector only needs two features "cli" and "test-fixture"

akoshelev (approved):
we should fix the RC command and be good to go

docker/report_collector.Dockerfile
cberkhoff: Not sure either, I'll keep them

docker/report_collector.Dockerfile
cberkhoff: I checked and it needs all those features

PR #1461, opened 2024-11-29
New HTTP handler for query status status requests

akoshelev:
Quite a bit of changes here, mostly because of a need to have test server for shard endpoint.
This endpoint is required for shards to be able to communicate query status updates to each other.

### DIFF
--- a/ipa-core/src/net/client/mod.rs
+++ b/ipa-core/src/net/client/mod.rs
@@ -562,7 +562,7 @@ pub(crate) mod tests {
         ff::{FieldType, Fp31},
         helpers::{
             make_owned_handler, query::QueryType::TestMultiply, BytesStream, HelperIdentity,
-            HelperResponse, RequestHandler, RoleAssignment, Transport, MESSAGE_PAYLOAD_SIZE_BYTES,
+            HelperResponse, RequestHandler, RoleAssignment, MESSAGE_PAYLOAD_SIZE_BYTES,
         },
         net::test::TestServer,
         protocol::step::TestExecutionStep,
@@ -759,7 +759,7 @@ pub(crate) mod tests {
         resp_ok(resp).await.unwrap();
 
         let mut stream = transport
-            .receive(HelperIdentity::ONE, (QueryId, expected_step.clone()))
+            .receive(HelperIdentity::ONE, &(QueryId, expected_step.clone()))
             .into_bytes_stream();
 
         assert_eq!(
--- a/ipa-core/src/net/client/mod.rs
+++ b/ipa-core/src/net/client/mod.rs
@@ -33,10 +33,10 @@ use crate::{
     },
     executor::IpaRuntime,
     helpers::{
-        query::{PrepareQuery, QueryConfig, QueryInput},
+        query::{CompareStatusRequest, PrepareQuery, QueryConfig, QueryInput},
         TransportIdentity,
     },
-    net::{http_serde, Error, CRYPTO_PROVIDER},
+    net::{error::ShardQueryStatusMismatchError, http_serde, Error, CRYPTO_PROVIDER},
     protocol::{Gate, QueryId},
 };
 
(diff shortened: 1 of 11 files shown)


akoshelev:
hybrid query test is stalling because we don't have #1463 in main

ipa-core/src/net/test.rs
```diff
@@ -488,23 +510,72 @@ impl TestServer {
...
+    ) -> hyper::Response<axum::body::Body> {
+        let test_server = TestServerBuilder::<Shard>::default()
+            .with_request_handler(handler)
+            .build()
+            .await;
+        test_server.server.handle_req(req).await
+    }
+
+    #[cfg(all(test, unit_test))]
+    pub async fn oneshot_success(
```
cberkhoff, line 545: Nit: I think it would be better to have this function just handle the parsing of the ok response so that it's more composable. An extension fn from Response.

E.g. this goes from `oneshot_success` to `parse_success` and is used like this:

```
TestServer::<Shard>::oneshot(req, handler).parse_success().await;
```

ipa-core/src/net/test.rs
```diff
@@ -537,42 +608,157 @@ impl TestServerBuilder {
         self
     }
 
-    pub async fn build(self) -> TestServer {
-        let identities =
-            ClientIdentities::new(self.disable_https, ShardedHelperIdentity::ONE_FIRST);
-        let mut test_config = TestConfig::builder()
+    fn test_config(&self) -> TestConfig {
```
cberkhoff, line 611: do we need this function?

I would remove it and have TestServerBuilder just call
```
pub async fn build(self) -> TestServer<Shard> {
        let test_config = TestConfig::builder().build();

        let transport = ...
```

ipa-core/src/net/client/mod.rs
```diff
@@ -384,6 +384,31 @@ impl<F: ConnectionFlavor> IpaHttpClient<F> {
...
         resp_ok(resp).await
     }
+
+    /// This API is used by leader shards in MPC to request query status information on peers.
+    /// If a given peer has status that doesn't match the one provided by the leader, it responds
+    /// with 412 error and encodes its status inside the response body. Otherwise, 200 is returned.
+    ///
+    /// # Errors
+    /// If the request has illegal arguments, or fails to be delivered
+    pub async fn status_match(&self, data: CompareStatusRequest) -> Result<(), Error> {
```
cberkhoff, line 394: 1) Can you please add tests to this code? I would be good to have some checks around the parsing of the error and the status.

2) Does this belong in `IpaHttpClient<Shard>` instead of the generic `IpaHttpClient<F>` ?

ipa-core/src/net/test.rs
akoshelev: yes, this function is used to construct `TestServerBuilder<Shard>` and `TestServerBuilder<Helper>`

ipa-core/src/net/test.rs
akoshelev: probably, I wanted to mimic the existing `assert_success_with` behaviour. Will consider this change if I need to use it somewhere else

ipa-core/src/net/test.rs
```diff
@@ -465,16 +469,34 @@ impl TestConfigBuilder {
         TestConfig::new(self)
     }
 }
-pub struct TestServer {
+pub struct TestServer<F: ConnectionFlavor = Helper> {
```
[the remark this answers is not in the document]
akoshelev: Yea, it should  be possible to do in a follow up. Designing a new struct from scratch should not require default values for generic types. The story is different for existing types where we need to add a generic parameter. So it is mostly to avoid refactoring all tests that use it and there are quite a few.

ipa-core/src/net/test.rs
cberkhoff: Sure, it's used on a couple of places... but why not just call `TestConfig::builder().build()` instead  of `self.test_config()`.

I don't see the benefit of test_config method. It just hides the possibility of configuring things.

PR #1462, opened 2024-11-29
Upgrade to Rust 1.83

akoshelev:
Massive change due to new clippy lints


eriktaubeneck:
let's make sure to remember to upgrade draft too

PR #1465, opened 2024-12-02
use Replicated for AdditiveShare consistently

eriktaubeneck:
I was in this file, and realized that we were using both `Replicated` and `AdditiveShare` interchangeably within the same file. This opts for `Replicated` everywhere (though if we prefer `AdditiveShare` everywhere that's a simple change.)


andyleiserson (approved):
Some day, we will do away with these renamed imports entirely, but for now, sticking to one or the other in a single file is a good start.

PR #1466, opened 2024-12-02
Compile hybrid finalizer

eriktaubeneck:
This uses the finalizer within the hybrid protocol, however it is currently panicking with: 

```
Step index 8 out of bounds for EightBitStep with count 8.
```

### DIFF
--- a/ipa-core/src/query/runner/hybrid.rs
+++ b/ipa-core/src/query/runner/hybrid.rs
@@ -1,10 +1,12 @@
 use std::{
     convert::{Infallible, Into},
     marker::PhantomData,
+    ops::Add,
     sync::Arc,
 };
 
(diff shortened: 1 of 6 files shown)


ipa-core/src/query/runner/hybrid.rs
```diff
@@ -305,30 +315,26 @@ mod tests {
 
             // TODO: after landing #1446, refactor this to only take the first 3
```
akoshelev, line 316: does this PR address this TODO or we need to do something extra?

akoshelev (approved):

ipa-core/src/query/runner/hybrid.rs
eriktaubeneck: good call, this can go!

PR #1468, opened 2024-12-03
Hybrid query

eriktaubeneck:
Adding one more layer of plumbing to actually running a hybrid query.

This is an early draft to get some input on the direction. We likely won't want to merge this until @tyurek's work to load the `HybridInfo` from the reports directly is complete.

### DIFF
--- a/ipa-core/src/query/runner/hybrid.rs
+++ b/ipa-core/src/query/runner/hybrid.rs
@@ -8,31 +8,35 @@ use std::{
 use futures::{stream::iter, StreamExt, TryStreamExt};
 use generic_array::ArrayLength;
 
+use super::QueryResult;
 use crate::{
     error::{Error, LengthError},
     ff::{
         boolean::Boolean,
-        boolean_array::{BooleanArray, BA3, BA8},
+        boolean_array::{BooleanArray, BA16, BA3, BA8},
         curve_points::RP25519,
         ec_prime_field::Fp25519,
         Serializable, U128Conversions,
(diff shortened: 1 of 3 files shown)


akoshelev (commented):
Looks ok to me, but hard to tell w/o actually running this code

eriktaubeneck:
@akoshelev agreed. this will require an integration test, yea? @tyurek is making progress on #1469, which should then allow us to write an integration test for this.

ipa-core/src/query/runner/hybrid.rs
```diff
@@ -175,6 +180,36 @@ where
...
+    let sharded = Sharded {
+        shard_id: gateway.shard_id(),
+        shard_count: gateway.shard_count(),
+        prss: Arc::new(cross_shard_prss),
+    };
+
+    let ctx = ShardedMaliciousContext::new_with_gate(prss, gateway, gate, sharded);
+
+    // TODO: load this from individual reports
+    let hybrid_info =
```
akoshelev, line 203: @tyurek is it something you would need to fix along with the AAD change?

akoshelev (approved):

ipa-core/src/query/runner/hybrid.rs
eriktaubeneck: actually this should be fixed. I'll find time to rebase this evening

eriktaubeneck:
this is included in #1479, so closing this one.

PR #1469, opened 2024-12-03
Add info fields to Hybrid Report Structs

tyurek:
This PR adds `HybridInfo` structs to Hybrid Reports.
More specifically:
* Removes `Serializable` trait from `HybridReport`
* Rewrites `HybridInfo` and related structs to be serializable and deserializable
* Adds these info structs to report structs (as a struct for `HybridReport`, as bytes for `EncryptedHybridReport`)
* Changes the encryption/decryption interfaces to not require an Info argument (grab is from the report instead)
* Changes `TestHybridRecord` to include info fields

Current status: Functionality is mostly there. All the tests in `cargo test` pass. I'm currently trying to debug `test_hybrid()`, which passes in a string that doesn't deserialize into a `TestHybridRecord` any more since there are now more required field. I have a few todos open too, which I'll get to once everything works fully.

There was some discussion last week about including the info in EncryptedHybridReport, but not HybridReport. That may be feasible, but doing it this way seemed like the fastest way to get things to work. We can change if desired.


ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -193,65 +193,126 @@ pub mod test {
...
     const SHARD2_MKS: [u64; 7] = [23456, 23456, 45678, 56789, 67890, 67890, 67890];
 
+    #[allow(clippy::too_many_lines)]
     fn get_records() -> Vec<TestHybridRecord> {
         let shard1_records = [
             TestHybridRecord::TestImpression {
                 match_key: SHARD1_MKS[0],
                 breakdown_key: 45,
+                key_id: 0,
+                helper_origin: "HELPER_ORIGIN".to_string(),
```
eriktaubeneck, line 203: nit: I think you can add `let helper_origin = "HELPER_ORIGIN".to_string();` at the top of the function and then you can just provide `helper_origin` here. (Similarly below with `conversion_site_domain`)

ipa-core/src/report/hybrid_info.rs
```diff
@@ -1,22 +1,22 @@
-use crate::report::{hybrid::NonAsciiStringError, KeyIdentifier};
+use crate::report::{
+    hybrid::{InvalidHybridReportError, NonAsciiStringError},
+    KeyIdentifier,
+};
 
 const DOMAIN: &str = "private-attribution";
```
eriktaubeneck, line 6: `DOMAIN` is the `conversion_site_domain`? We shouldn't assume this is fixed (in certain types of queries it may come from multiple. We can assume `helper_origin` is fixed, like we [do here](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/report/ipa.rs#L52). 

It likely will simplify things if `conversion_site_domain` is the only variable width element on the report, like in [`EncryptedOprfReport`](https://github.com/private-attribution/ipa/blob/main/ipa-core/src/report/ipa.rs#L354-L358).

`conversion_site_domain` is still a valid property to have on this struct, but it's also safe to load it directly from the static constant.

eriktaubeneck:
looking good! a comment about which strings can be assumed to be fixed, but otherwise this seems to be the right approach.

ipa-core/src/protocol/hybrid/agg.rs
tyurek: use after move, but I can clone them at least

ipa-core/src/report/hybrid_info.rs
tyurek: Is `DOMAIN` just conversion_site_domain? The hybrid code has been assuming those were different things up until now. The [OPRF info struct](https://github.com/private-attribution/ipa/blob/dca5be774f9920dcc9a91e04418355802179f415/ipa-core/src/hpke/info.rs#L56) also appears to treat them as different things

ipa-core/src/protocol/hybrid/agg.rs
eriktaubeneck: ahh, yea. not quite as pretty as I was hoping

ipa-core/src/report/hybrid_info.rs
eriktaubeneck: (talked about in person but including here for posterity) my bad - `DOMAIN` (poorly named in this context) is used for encryption. it will be a fixed string that is known in advance.

eriktaubeneck (approved):
ship it!

PR #1470, opened 2024-12-03
Add proptest for breakdown-reveal aggregation

andyleiserson:
And some general proptest cleanup and improvements. Run longer versions of the MPC proptests when EXEC_SLOW_TESTS is set.

Prompted by #1463.


akoshelev (approved):

andyleiserson:
The bulk of the missing coverage is because I marked the OPRF version of the proptest as `#[ignore]`. The two versions are different enough in terms of code that it seemed worth keeping the OPRF version, but not different enough in coverage that it seemed worth the cost of running both.

PR #1472, opened 2024-12-04
add CLI command to run hybrid query

eriktaubeneck:
turns out a bunch of these functions aren't actually OprfIpa specific. I did one small "hack" to convert `HybridQueryParams` into `IpaQueryConfig` (it's just a super set and the extra fields are unused at this point) to avoid copying and pasting the entire `run_query_and_validate` function.

This is mostly complete, but we'll need to swap in `HTTPS` clients instead of `HTTP`, so we're blocked on that, cc @akoshelev

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -14,14 +14,16 @@ use hyper::http::uri::Scheme;
 use ipa_core::{
     cli::{
         playbook::{
-            make_clients, playbook_oprf_ipa, run_query_and_validate, validate, validate_dp,
-            InputSource,
+            make_clients, make_sharded_clients, playbook_oprf_ipa, run_hybrid_query_and_validate,
+            run_query_and_validate, validate, validate_dp, HybridQueryResult, InputSource,
         },
         CsvSerializer, IpaQueryResult, Verbosity,
     },
     config::{KeyRegistries, NetworkConfig},
     ff::{boolean_array::BA32, FieldType},
-    helpers::query::{DpMechanism, IpaQueryConfig, QueryConfig, QuerySize, QueryType},
+    helpers::query::{
+        DpMechanism, HybridQueryParams, IpaQueryConfig, QueryConfig, QuerySize, QueryType,
+    },
     net::{Helper, IpaHttpClient},
     report::{EncryptedOprfReportStreams, DEFAULT_KEY_ID},
     test_fixture::{
@@ -57,6 +59,9 @@ struct Args {
     /// The destination file for output.
     #[arg(long, value_name = "OUTPUT_FILE")]
     output_file: Option<PathBuf>,
+
+    #[arg(long, default_value_t = 1)]
+    shard_count: usize,
 
     #[command(subcommand)]
     action: ReportCollectorCommand,
@@ -132,6 +137,13 @@ enum ReportCollectorCommand {
         #[clap(flatten)]
         ipa_query_config: IpaQueryConfig,
     },
+    MaliciousHybrid {
+        #[clap(flatten)]
+        encrypted_inputs: EncryptedInputs,
+
+        #[clap(flatten)]
+        hybrid_query_config: HybridQueryParams,
+    },
 }
 
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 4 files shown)


akoshelev:
not sure what you mean by HTTPS clients - we do have them and they fully support HTTPS

akoshelev:
on the integration test side, #1473 should provide a way to write tests that leverage HTTPS

ipa-core/src/bin/report_collector.rs
```diff
@@ -329,6 +342,55 @@ fn write_ipa_output_file(
     Ok(())
 }
 
+async fn hybrid(
+    args: &Args,
+    hybrid_query_config: HybridQueryParams,
+    helper_clients: &[IpaHttpClient<Helper>; 3],
```
eriktaubeneck, line 348: sorry @akoshelev should have made a comment here directly. I'm not sure how this needs to be updated. it doesn't seem like oprf_ipa uses https (even for malicious). additionally, this needs to be 3 * number of shards, so presumably needs to be a vec, not an array.

akoshelev, line 348: this does not seem right - we need to have a client per shard. Sharded shuffle in `test_mpc` may provide a way to structure this for hybrid

akoshelev: eventual consistency is fun. Yea, we need at 3*number of shards clients and sharded shuffle uses that currently. As for `https` we seem to use [it](https://github.com/akoshelev/raw-ipa/blob/cf7efd485096b3132e44366a6c0aad7d263ecd21/ipa-core/src/bin/report_collector.rs#L172)

eriktaubeneck: agreed that we need to use `https` scheme, but I'm not seeing an `IpaHttpsClient` struct, so this specific type may not need to change. I'll take a look at sharded shuffle to try and sort it out.

ipa-core/src/cli/playbook/hybrid.rs
```diff
@@ -0,0 +1,135 @@
...
+            .zip(leader_clients)
+            .map(|(input_stream, client)| {
+                client.query_input(QueryInput {
+                    query_id,
+                    input_stream,
+                })
+            }),
+    )
+    .await
+    .unwrap();
```
eriktaubeneck, line 54: @akoshelev I ended up needing to create a new function here that could take `Vec<[IpaHttpClient<Helper>; 3]>`. right now it submits everything to the leader, but this seems like the natural place to do the round robin submission.

ipa-core/src/bin/report_collector.rs
```diff
@@ -329,6 +358,94 @@ fn write_ipa_output_file(
     Ok(())
 }
 
+fn write_hybrid_output_file(
```
akoshelev, line 361: this looks oddly familiar, is it possible to re-use the same code and avoid copy paste here?

ipa-core/src/bin/report_collector.rs
```diff
@@ -329,6 +358,94 @@ fn write_ipa_output_file(
...
+    // despite the name, this is generic enough to work with hybrid
+    let encrypted_report_streams = EncryptedOprfReportStreams::from(files);
+
+    let query_config = QueryConfig {
+        size: QuerySize::try_from(encrypted_report_streams.query_size).unwrap(),
+        field_type: FieldType::Fp32BitPrime,
+        query_type,
+    };
+
+    let query_id = helper_clients[0][0]
```
cberkhoff, line 423: should we first do a query status to make sure something else isn't running?

ipa-core/src/bin/report_collector.rs
eriktaubeneck: same old problem with Oprf and Hybrid. we can either spend time supporting both or just delete the other when we remove Oprf.

ipa-core/src/bin/report_collector.rs
eriktaubeneck: I believe it panics with "query already running" already, so that's effectively the same, no?

cberkhoff: Not really. The system does't have something like distributed transactions. Think of the following scenario; helper 1 gets the `create_query` and sends `prepare_query` to 2 and 3. Say helper 2 return ok but helper 3 says its running something return error. We don't "rollback" helper2 so it's left awaiting inputs.

This problem should be worse now with more states spread around.

eriktaubeneck: That's fair, and we have the "kill" endpoint to help clean up that state that sometimes happens. I'm inclined to try and get this merged as is (which is the same as Oprf Ipa, but with shards), and punt on making this better once we can actually run a query. For the immediate testing needs, we'll be restarting the binaries regularly anways.

cberkhoff: Don't think we have a sharded kill, but we can easily restart the k8s cluster.

I you think adding a `query_status` is going to take much time then I agree.

akoshelev (approved):

ipa-core/src/bin/report_collector.rs
eriktaubeneck: certainly something we should add, but we likely want to do that across all shards, so let's punt for now.

eriktaubeneck:
only check fail is code patch, which is coming in #1476. merging

PR #1473, opened 2024-12-04
HTTPS test for sharded shuffle

akoshelev:
This almost completes the setup for TLS/encryption infrastructure with one major exception - match key encryption keys are not re-used across shards. I have a PR almost ready, but decided to split that functionality to make PR somewhat revieweable


ipa-core/src/query/processor.rs
```diff
@@ -1150,16 +1151,18 @@ mod tests {
         #[tokio::test]
         async fn combined_status_response() {
             fn shard_handle(si: ShardIndex) -> Arc<dyn RequestHandler<ShardIndex>> {
+                const THIRD_SHARD: ShardIndex = ShardIndex::from_u32(3);
+                const SECOND_SHARD: ShardIndex = ShardIndex::from_u32(2);
```
eriktaubeneck, line 1155: wouldn't this be fourth and third?

ipa-core/src/protocol/hybrid/agg.rs
```diff
@@ -187,6 +187,7 @@ pub mod test {
     // the inputs are laid out to work with exactly 2 shards
     // as if it we're resharded by match_key/prf
     const SHARDS: usize = 2;
+    const SECOND_SHARD: ShardIndex = ShardIndex::from_u32(1);
```
eriktaubeneck, line 190: shouldn't this just be `from`?

```suggestion
    const SECOND_SHARD: ShardIndex = ShardIndex::from(1);
```

akoshelev: trait methods [are not supported](https://github.com/rust-lang/rust/issues/67792) in const contexts yet

eriktaubeneck (approved):

ipa-core/tests/common/mod.rs
```diff
@@ -189,18 +185,39 @@ pub fn spawn_shards(
                         .args(["-i", &id.to_string()])
                         .args(["--shard-index", &shard_index.to_string()])
                         .args(["--shard-count", &shard_count.to_string()])
-                        .args(["--network".into(), config_path.join("network.toml")])
-                        .arg("--disable-https")
```
cberkhoff, line 193: so this is the first time we have https tests?

cberkhoff (approved):
What changed in `sharded_keygen` or is it more of a cosmetical change?

ipa-core/tests/common/mod.rs
akoshelev: for sharded world yes

PR #1475, opened 2024-12-04
Using k8s supported DNS format

cberkhoff:
Updated the script to generate the hostnames in the way K8s likes them to be.

### DIFF
--- a/scripts/create-sharded-conf.py
+++ b/scripts/create-sharded-conf.py
@@ -3,6 +3,10 @@ import subprocess
 import argparse
 import shutil
 
+# The following format is what ipa-infra project expects.
+# See:
+# https://github.com/private-attribution/ipa-infra/blob/main/templates/service.yaml
+# https://github.com/private-attribution/ipa-infra/blob/main/templates/helpers.yaml
 DNS_FORMAT = "h{}-helper-shard-{}.h{}-helper-shard.default.svc.cluster.local"
 
 def generate_keys(base_dir, shard_count):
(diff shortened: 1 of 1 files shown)


scripts/create-sharded-conf.py
```diff
@@ -23,23 +40,26 @@ def generate_keys(base_dir, shard_count, helper_name_format):
...
                     "--mk-private-key",
                     f"{dir_path}/{helper_name}_mk.key",
                 ])
+
             subprocess.run(keygen_cmd)
 
             # Copy mk public and private keys from the first shard to other shards
             if j != 0:
-                src_dir = f"{base_dir}/helper{i}/shard{0}"
+                src_dir = f"{base_dir}/helper{i}/shard0"
```
akoshelev, line 56: I tried to trace all references to the shard directory naming convention in the Rust code and replace them with function call. We can't do it here, but it will be great if we had a comment here explaining that this needs to be kept in sync with the Rust code (ideally with code pointers https://github.com/private-attribution/ipa/blob/d8b73fb2521c8f90a8fd9897d462a5a9816adeb7/ipa-core/src/cli/clientconf.rs#L228)

scripts/create-sharded-conf.py
```diff
@@ -3,18 +3,35 @@
 import argparse
 import shutil
 
-def generate_keys(base_dir, shard_count, helper_name_format):
+DNS_FORMAT = "h{}-helper-shard-{}.h{}-helper-shard.default.svc.cluster.local"
```
akoshelev, line 6: can you provide a pointer to k8 configuration code that expects this format?

akoshelev (approved):

scripts/create-sharded-conf.py
cberkhoff: I can add some notes, but it's spread between the helper and service k8s templates

PR #1476, opened 2024-12-04
Hybrid integration test

eriktaubeneck:
This is stacked on #1472, and won't actually work until we can encrypt the input with the work that @tyurek is doing.

### DIFF
--- a/ipa-core/tests/hybrid.rs
+++ b/ipa-core/tests/hybrid.rs
@@ -4,7 +4,11 @@ mod common;
 
 use std::process::{Command, Stdio};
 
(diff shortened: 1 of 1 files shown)


ipa-core/tests/hybrid.rs
```diff
@@ -15,8 +19,15 @@ pub const IN_THE_CLEAR_BIN: &str = env!("CARGO_BIN_EXE_in_the_clear");
 #[test]
 fn test_hybrid() {
```
akoshelev, line 20:
```
#[ignore]
``` 

should work for now

akoshelev (approved):

PR #1477, opened 2024-12-05
Support streaming inputs in CLI tooling

akoshelev:
This change does not integrate streaming into report collector yet as we're not quite ready for it. But it provides a means to split a `BufRead` interface into `M` streams that read from it one at a time. This should not require reading the entire file in memory in order to submit it.

There are a few things that I decided not to bring in the first version:

* Reading from a file can be done in async runtime using `AsyncRead` interface. The benefit of it is not clear to me atm.
* reading more than one line at a time. We could support reading lets say 8kb chunks from file and then let HTTP streams to stream it out.

### DIFF
--- a/ipa-core/src/cli/playbook/streaming.rs
+++ b/ipa-core/src/cli/playbook/streaming.rs
@@ -1,0 +1,283 @@
+use std::{
+    io::BufRead,
+    pin::Pin,
+    task::{Context, Poll, Waker},
+};
+
+use bytes::Bytes;
+use futures::Stream;
+
+use crate::{
+    error::BoxError,
+    helpers::BytesStream,
+    sync::{Arc, Mutex},
+};
+
+/// Trait for submitting inputs as streams, rather than reading everything
+/// in memory. Should provide better performance for very large inputs.
+trait StreamingSubmission {
+    /// Spits itself into `count` instances of [`BytesStream`].
+    fn into_byte_streams(self, count: usize) -> Vec<impl BytesStream>;
+}
+
+/// Round-Robin strategy to read off the provided buffer
+/// and distribute them. Inputs is expected to be hex-encoded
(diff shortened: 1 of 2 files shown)


eriktaubeneck (approved):

ipa-core/src/cli/playbook/streaming.rs
```diff
@@ -0,0 +1,283 @@
...
+impl<R: BufRead> Stream for RoundRobinStream<R> {
+    type Item = Result<Bytes, BoxError>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut state = self.state.lock().unwrap();
+        // it can be made more efficient by using atomics to check index.
+        // but probably it does not matter for this code as it is not on the hot path
+        if state.idx == self.idx {
+            let poll = state.read_next();
+            if poll.is_ready() {
```
cberkhoff, line 53: does read_next only return Poll::Ready?

akoshelev: no, but only if it is ready, we want to advance the index

ipa-core/src/cli/playbook/streaming.rs
```diff
@@ -0,0 +1,283 @@
...
+        let mut expected = vec![String::new(); count];
+        for (i, next) in input.into_iter().enumerate() {
+            expected[i % count].push_str(next.as_ref());
+        }
+        assert_eq!(expected, drain_all(streams).await);
+    }
+
+    proptest! {
+        #[test]
+        fn proptest_round_robin(input: Vec<String>, count in 1_usize..953) {
```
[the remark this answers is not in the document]
akoshelev: nothing really, just something that is not large enough to cause proptests to run forever

PR #1483, opened 2024-12-07
Improve performance of hybrid encrypt CLI

akoshelev:
There were quite a few bottlenecks here:
* Writes were done serially, writing one file at a time.
* Shares were encrypted on a single CPU core

I almost used `rayon` to parallelize encryption, but the problem is that we need to get the output sorted to maintain total order across files. Rayon can do that, but requires collecting `ParallelIterator` which would be bad for generating 100M+ reports.

Our goal is to be able to share and encrypt 1B, so streaming and manual fiddling with thread pools is justified imo.

The way this CLI works right now: it keeps a compute pool for encryption (thread-per-core) and a separate pool of 3 threads to write data for each helper in parallel

I also made a few tweaks to improve code re-usability in this module.

## Benchmarks
Time it takes to encrypt 1M reports done locally on M1 Mac Pro (10 cores)

Before this change:
Encryption process is completed. 442.15834075s

After this change
Encryption process is completed. 55.63269625s

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -534,7 +534,7 @@ async fn ipa_test(
         r
     };
 
-    let mut key_registries = KeyRegistries::default();
+    let key_registries = KeyRegistries::default();
     let Some(key_registries) = key_registries.init_from(network) else {
         panic!("could not load network file")
     };
@@ -546,7 +546,7 @@ async fn ipa_test(
         helper_clients,
         query_id,
         ipa_query_config,
-        Some((DEFAULT_KEY_ID, key_registries)),
+        Some((DEFAULT_KEY_ID, key_registries.each_ref())),
     )
     .await;
 
--- a/ipa-core/src/cli/crypto/hybrid_encrypt.rs
+++ b/ipa-core/src/cli/crypto/hybrid_encrypt.rs
@@ -1,8 +1,13 @@
 use std::{
-    fs::{read_to_string, OpenOptions},
-    io::Write,
-    iter::zip,
+    array,
+    collections::BTreeMap,
+    fs::{read_to_string, File, OpenOptions},
+    io::{BufWriter, Write},
     path::{Path, PathBuf},
+    sync::mpsc::{channel, Sender},
+    thread,
+    thread::JoinHandle,
+    time::Instant,
 };
 
 use clap::Parser;
@@ -15,10 +20,22 @@ use crate::{
     },
     config::{KeyRegistries, NetworkConfig},
     error::BoxError,
+    hpke::{KeyRegistry, PublicKeyOnly},
     report::hybrid::{HybridReport, DEFAULT_KEY_ID},
     secret_sharing::IntoShares,
     test_fixture::hybrid::TestHybridRecord,
 };
+
+/// Encryptor takes 3 arguments: `report_id`, helper that the shares must be encrypted towards
+/// and the actual share to encrypt.
+type EncryptorInput = (usize, usize, HybridReport<BreakdownKey, TriggerValue>);
+/// Encryptor sends report id and encrypted bytes down to file worker to write those bytes
+/// down
+type EncryptorOutput = (usize, Vec<u8>);
+type FileWorkerInput = EncryptorOutput;
+
+/// This type is used quite often in this module
+type UnitResult = Result<(), BoxError>;
 
 #[derive(Debug, Parser)]
 #[clap(name = "test_hybrid_encrypt", about = "Test Hybrid Encrypt")]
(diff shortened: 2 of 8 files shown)


ipa-core/src/bin/report_collector.rs
```diff
@@ -534,7 +534,7 @@ async fn ipa_test(
         r
     };
 
-    let mut key_registries = KeyRegistries::default();
+    let key_registries = KeyRegistries::default();
     let Some(key_registries) = key_registries.init_from(network) else {
```
eriktaubeneck, line 538: IIRC, the `mut` was needed to call `init_from`? but if this is compiling that must not be needed.

ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -71,28 +89,199 @@ impl HybridEncryptArgs {
...
+                                    file_writer[helper_id].send((i, output))?;
+                                }
+
+                                Ok(())
+                            })
+                            .unwrap(),
+                    )
+                })
+                .collect(),
+            next_worker: 0,
```
eriktaubeneck, line 149: why is `next_worker` always 0? what's the point of it if it's constant?

ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -15,11 +20,23 @@ use crate::{
...
     config::{KeyRegistries, NetworkConfig},
     error::BoxError,
+    hpke::{KeyRegistry, PublicKeyOnly},
     report::hybrid::{HybridReport, DEFAULT_KEY_ID},
     secret_sharing::IntoShares,
     test_fixture::hybrid::TestHybridRecord,
 };
 
+/// Encryptor takes 3 arguments: `report_id`, helper that the shares must be encrypted towards
+/// and the actual share to encrypt.
```
eriktaubeneck, line 30: nit: maybe use `report` instead of `share` here, since all the types call the thing that's encrypted a `report`.

ipa-core/src/cli/crypto/hybrid_encrypt.rs
akoshelev: mutation occurs inside `encrypt_share`

ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -71,28 +89,199 @@ impl HybridEncryptArgs {
...
+impl EncryptorPool {
+    pub fn with_worker_threads(
+        thread_count: usize,
+        file_writer: [Sender<EncryptorOutput>; 3],
+        key_registries: [KeyRegistry<PublicKeyOnly>; 3],
+    ) -> Self {
+        Self {
+            pool: (0..thread_count)
+                .map(move |i| {
+                    let (tx, rx) = channel::<EncryptorInput>();
```
eriktaubeneck, line 125: `rx` is an `EncryptorInput`, but what is `tx`? a slightly more descriptive naming here would be helpful.

ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -71,28 +89,199 @@ impl HybridEncryptArgs {
...
+                    Ok(())
+                }
+
+                // write low watermark. All reports below this line have been written
+                let mut lw = 0;
+                let mut pending_reports = BTreeMap::new();
+
+                // Buffered writes should improve IO, but it is likely not the bottleneck here.
+                let mut writer = BufWriter::new(file);
+                for (report_id, report) in rx {
```
eriktaubeneck, line 259: is `rx` different here than above? above it also had a `helper_id`.

eriktaubeneck (approved):

ipa-core/src/cli/crypto/hybrid_encrypt.rs
akoshelev: it is common in Rust to name sending and receiving ends of a channel as `tx` (transmission) and rx (receiving): https://doc.rust-lang.org/std/sync/mpsc/index.html

ipa-core/src/cli/crypto/hybrid_encrypt.rs
akoshelev: helper_id is only needed for encryption pool to tell it which file needs to receive encrypted blob. file workers don't need that information because they are bound to a file already

ipa-core/src/bin/report_collector.rs
akoshelev: yea I changed the signature of this method as it is always used to fully consume `KeyRegistries`

ipa-core/src/cli/crypto/hybrid_encrypt.rs
eriktaubeneck: ahh ok. i haven't encountered this before.

PR #1484, opened 2024-12-07
use BA32 instead of BA16 for hybrid histogram values

eriktaubeneck:
As we scale up to 1B rows, the saturated_add results in all histogram values saturating at `2^16 - 1` (e.g. `65_535`) with `BA16`. Increasing this to `BA32` will move the saturation point up to `4_294_967_295`. 

Our current data generator has a 20% match rate. With a max_value of 7 (and thus expected value of 3.5), we'd expect histograms to have values around `(1B * 0.2 * 3.5) / 256 = 2_734_375`, safely under the saturation point.

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -21,10 +21,7 @@ use ipa_core::{
         CsvSerializer, IpaQueryResult, Verbosity,
     },
     config::{KeyRegistries, NetworkConfig},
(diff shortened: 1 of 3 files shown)


akoshelev (approved):
What is wrong with saturating at BA16?

eriktaubeneck:
> What is wrong with saturating at BA16?

it saturates at this scale, so we don't get the correct result

PR #1486, opened 2024-12-09
Provide backpressure on crypto binary

akoshelev:
We've been reading from a file faster than we were encrypting. That caused sender buffer to grow infinitely.


akoshelev:
I tested it locally and before this change, the memory used to encrypt 1M input was around 8Gb. After this change, I encrypted 30M and the memory usage stayed under 200Mb.

This change makes it a bit slower because once the channels are saturated, the producer (reading from a file) blocks on the first channel that is full while there may be others that are not full. Standard library does not expose channel size, so we would need to migrate to crossbeam if we want to do that

PR #1487, opened 2024-12-10
Remove Info from plaintext Hybrid Reports

tyurek:
* Remove HybridInfo from being included in plaintext HybridReports by default.
* HybridInfo is now a required argument for encryption
* Because we want the info in some contexts, adds a function to optionally return both the report and the info when decrypting
* Adds a `create_hybrid_info()` function to TestHybridRecord so that we can do the initial encryption
* Adjusts encryption/decryption scripts to handle refactor. Output format is unchanged.

### DIFF
--- a/ipa-core/src/cli/crypto/hybrid_decrypt.rs
+++ b/ipa-core/src/cli/crypto/hybrid_decrypt.rs
@@ -14,7 +14,10 @@ use crate::{
         U128Conversions,
     },
     hpke::{KeyRegistry, PrivateKeyOnly},
-    report::hybrid::{EncryptedHybridReport, HybridReport},
+    report::{
+        hybrid::{EncryptedHybridReport, HybridReport},
+        hybrid_info::HybridInfo,
+    },
     test_fixture::Reconstruct,
 };
 
@@ -104,7 +107,7 @@ impl HybridDecryptArgs {
         for (dec_report1, (dec_report2, dec_report3)) in
             decrypted_reports1.zip(decrypted_reports2.zip(decrypted_reports3))
         {
-            match (dec_report1, dec_report2, dec_report3) {
+            match (dec_report1.0, dec_report2.0, dec_report3.0) {
                 (
                     HybridReport::Impression(impression_report1),
                     HybridReport::Impression(impression_report2),
@@ -125,8 +128,9 @@ impl HybridDecryptArgs {
                     ]
                     .reconstruct()
                     .as_u128();
+                    let key_id = dec_report1.1.impression.key_id;
 
-                    writeln!(writer, "i,{match_key},{breakdown_key}")?;
+                    writeln!(writer, "i,{match_key},{breakdown_key},{key_id}")?;
                 }
                 (
                     HybridReport::Conversion(conversion_report1),
@@ -148,7 +152,14 @@ impl HybridDecryptArgs {
                     ]
                     .reconstruct()
                     .as_u128();
-                    writeln!(writer, "c,{match_key},{value}")?;
+
+                    let key_id = dec_report1.1.conversion.key_id;
+                    let conversion_site_domain = dec_report1.1.conversion.conversion_site_domain;
+                    let timestamp = dec_report1.1.conversion.timestamp;
+                    let epsilon = dec_report1.1.conversion.epsilon;
+                    let sensitivity = dec_report1.1.conversion.sensitivity;
+
+                    writeln!(writer, "c,{match_key},{value},{key_id},{conversion_site_domain},{timestamp},{epsilon},{sensitivity}")?;
                 }
                 _ => {
                     panic!("Reports are not all the same type");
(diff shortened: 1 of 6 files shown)


ipa-core/src/cli/crypto/hybrid_decrypt.rs
```diff
@@ -149,12 +148,7 @@ impl HybridDecryptArgs {
                     ]
                     .reconstruct()
                     .as_u128();
-                    let key_id = conversion_report1.info.key_id;
-                    let conversion_site_domain = conversion_report1.info.conversion_site_domain;
-                    let timestamp = conversion_report1.info.timestamp;
-                    let epsilon = conversion_report1.info.epsilon;
-                    let sensitivity = conversion_report1.info.sensitivity;
-                    writeln!(writer, "c,{match_key},{value},{key_id},{conversion_site_domain},{timestamp},{epsilon},{sensitivity}")?;
+                    writeln!(writer, "c,{match_key},{value}")?;
```
tyurek, line 151: Do we want to write the info fields to file?

tyurek: Spoke with Erik and he said we should keep those. Rearranged accordingly

ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -206,9 +216,12 @@ impl ReportWriter {
         &mut self,
         report_id: usize,
         shares: [HybridReport<BreakdownKey, TriggerValue>; 3],
+        info: &HybridInfo,
     ) -> UnitResult {
         for (i, share) in shares.into_iter().enumerate() {
-            self.encryptor_pool.encrypt_share((report_id, i, share))?;
+            //  todo: maybe a smart pointer to avoid cloning
+            self.encryptor_pool
+                .encrypt_share((report_id, i, share, info.clone()))?;
```
tyurek, line 224: probably some room for improvement here. What happens here is that we have three different reports (generated by share()), which all are encrypted with the same HybridInfo. I changed the encrypt_share function to take an info argument, but it really only needs a reference (I tried having it take a reference but ran into issues with lifetimes). Is it worth trying to use a smart pointer here? Or maybe rewrite encrypt_share to take three at a time but only one Info?

akoshelev: Yea it feels like using references here is the right thing to do. But lifetimes can be tricky to handle. I can take a look at it later today

ipa-core/src/report/hybrid_info.rs
```diff
@@ -18,10 +18,12 @@ impl HybridImpressionInfo {
...
 
     #[must_use]
-    pub fn byte_len(&self) -> usize {
+    /// # Panics
+    /// If report length does not fit in `u16`.
+    pub fn byte_len(&self) -> u16 {
         let out_len = std::mem::size_of_val(&self.key_id);
         debug_assert_eq!(out_len, self.to_bytes().len(), "Serialization length estimation is incorrect and leads to extra allocation or wasted memory");
-        out_len
+        out_len.try_into().unwrap()
```
akoshelev, line 26: One benefit of using usize here is that the caller cannot ignore the fact that report can be more than 65k. I thought we only need to do this cast once, when we encode the report length, so this module may not need to concern itself with such limitations

tyurek:
Looking to wrap this one up. I put out an example commit here: https://github.com/private-attribution/ipa/commit/13fafea654d01ea3ac2dcd8ddfbe8c2270c595ef which modifies the encryptor to encrypt three shares per thread and use the same aad info for each. I could do either this or use a smart pointer, but doing it with lifetimes seems complicated.

ipa-core/src/test_fixture/hybrid.rs
```diff
@@ -146,7 +146,7 @@ where
                         HybridReport::Impression::<BK, V>(HybridImpressionReport {
                             match_key: match_key_share,
                             breakdown_key: breakdown_key_share,
-                            info: HybridImpressionInfo::new(key_id),
+                            //info: HybridImpressionInfo::new(key_id),
```
akoshelev, line 149:
```suggestion

```

ipa-core/src/report/hybrid.rs
```diff
@@ -333,33 +334,39 @@ where
...
         &self,
         key_id: KeyIdentifier,
         key_registry: &impl PublicKeyRegistry,
+        info: &HybridConversionInfo,
         rng: &mut R,
     ) -> Result<Vec<u8>, InvalidHybridReportError> {
-        let mut out = Vec::with_capacity(usize::from(self.encrypted_len()));
-        self.encrypt_to(key_id, key_registry, rng, &mut out)?;
-        debug_assert_eq!(out.len(), usize::from(self.encrypted_len()));
+        let mut out = Vec::with_capacity(usize::from(self.ciphertext_len() + u16::try_from(info.byte_len()).unwrap()));
```
akoshelev, line 367: there seems to be a fair amount of similarities between this code and the code above. Any way to reduce the copy-paste?

PR #1488, opened 2024-12-10
Improve memory usage for computing hybrid results in the clear

akoshelev:
This operation was OOMing on 500M inputs

```
memory allocation of 73551314960 bytes failed
Aborted
```

The reason for that was that we were keeping all the information from original reports, including AAD and match key. 
The improved algorithm only keeps breakdown keys and values to perform attribution.

My back of the envelope calculation says that this improvement is enough to get us to 1 billion reports. Each entry would take ~8 bytes and for 1B entries we are looking at 8Gb memory consumption which should be doable

I ran it on 500M and it finished successfully.


eriktaubeneck (approved):
looks good!

PR #1489, opened 2024-12-11
Don't ignore errors on query status API

akoshelev:
We may be observing issues trying to run Hybrid at scale that are not visible to us due to the nature how query status handles those. If error is not a mismatch, it is just silently ignored. This could lead to a variety of problems - most importantly it will mask the failures and made other shards think that everything is OK, and they can proceed with finalizing results.

This PR changes this and makes this API return an error if there is at least one error that is not a status mismatch. If all of them are, the behaviour is the same as before

### DIFF
--- a/ipa-core/src/query/processor.rs
+++ b/ipa-core/src/query/processor.rs
@@ -369,13 +369,14 @@ impl Processor {
     /// [`QueryStatusError::DifferentStatus`] and retrieve it's internal state. Returns [`None`]
     /// if not possible.
     #[cfg(feature = "in-memory-infra")]
-    fn downcast_state_error(box_error: crate::error::BoxError) -> Option<QueryStatus> {
+    fn downcast_state_error(box_error: &crate::error::BoxError) -> Option<QueryStatus> {
         use crate::helpers::ApiError;
-        let api_error = box_error.downcast::<ApiError>().ok()?;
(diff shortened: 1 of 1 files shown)


tyurek:
yeah, this seems like a good thing to fix

ipa-core/src/query/processor.rs
```diff
@@ -431,17 +432,14 @@ impl Processor {
...
-            // TODO: this code is a ticking bomb - it ignores all errors, not just when
-            // query is not found. If there is no handler, handler responded with an error, etc.
-            // Moreover, any error may result in client mistakenly assuming that the status
-            // is completed.
-            let states: Vec<_> = e
-                .failures
-                .into_iter()
-                .filter_map(|(_si, e)| Self::get_state_from_error(e))
-                .collect();
-            status = states.into_iter().fold(status, min_status);
```
tyurek, line 444: for my education, why does this cause errors to be ignored? What is a minimum status and what would folding an iterator of failures even do?

cberkhoff (approved):

cberkhoff:
Dont we need to handle this error in the RC loop in some way?

ipa-core/src/query/processor.rs
akoshelev: Yea it is a bit subtle here. filter_map filters out all errors that are not status mismatch error. The closure maps them to None and filter_map excludes them

PR #1491, opened 2024-12-12
Multithreaded Encryption

tyurek:
Adds a `seq_join` and `async_move` to the shard code operating on encrypted report streams to make decryption parallelizable. 

To make sure it is in fact running in multiple cores, I added a print line to the lambda that calls the decrypt function and tested with the multi-threading feature
`cargo test --package ipa-core --test hybrid --no-default-features --features "cli compact-gate web-app real-world-infra test-fixture relaxed-dp multi-threading" -- test_hybrid --exact`
Example output: 
```
thread id: Thread { id: ThreadId(12), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(8), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(8), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(9), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(12), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(12), name: Some("query-executor"), .. }
thread id: Thread { id: ThreadId(8), name: Some("query-executor"), .. }
```

### DIFF
--- a/ipa-core/src/query/runner/hybrid.rs
+++ b/ipa-core/src/query/runner/hybrid.rs
@@ -46,6 +46,7 @@ use crate::{
         replicated::semi_honest::AdditiveShare as Replicated, BitDecomposed, TransposeFrom,
         Vectorizable,
     },
+    seq_join::seq_join,
     sharding::{ShardConfiguration, Sharded},
 };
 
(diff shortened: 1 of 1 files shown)


tyurek:
hmm, not sure why this fails but ./pre-commit runs okay...

eriktaubeneck:
> hmm, not sure why this fails but ./pre-commit runs okay...

the unit test needs the `mult-threading` flag now

akoshelev:
or worker_threads attribute if it is a tokio test

eriktaubeneck (approved):

PR #1492, opened 2024-12-12
Add Fixed-Rate Polling Specifier To ReportCollector

tyurek:
To test this, I added `.args(["--set-fixed-polling-ms", "100"])` before line 93 in the file containing `test_hybrid()`. I also verified that it fails with bad arguments

Test ran with `cargo test --package ipa-core --test hybrid --no-default-features --features "cli compact-gate web-app real-world-infra test-fixture relaxed-dp multi-threading" -- test_hybrid --exact`

I should say this usually passes... I ran it once (with the argument specified) and got
```
Unable to create query!: FailedHttpRequest { dest: "localhost:64473", status: 500, reason: "500 Internal Server Error: One or more peers rejected the request: [(ShardIndex(3), ShardError { shard_index: ShardIndex(0), source: ConnectError { dest: \"localhost:64493\", inner: hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidMessage(InvalidContentType) } }) } })]" }
```
But I don't think that's due to any incorrect logic in this PR

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -154,8 +154,8 @@ enum ReportCollectorCommand {
 
         // If set, use the specified fixed polling interval when running a query.
         // Otherwise, use exponential backoff.
-        #[arg(long, default_value_t = 0)]
-        set_fixed_polling_ms: u64,
+        #[clap(long)]
+        set_fixed_polling_ms: Option<u64>,
     },
 }
 
@@ -428,7 +428,7 @@ async fn hybrid(
     helper_clients: Vec<[IpaHttpClient<Helper>; 3]>,
     encrypted_inputs: &EncryptedInputs,
     count: usize,
-    set_fixed_polling_ms: u64,
+    set_fixed_polling_ms: Option<u64>,
 ) -> Result<(), Box<dyn Error>> {
     let query_type = QueryType::MaliciousHybrid(hybrid_query_config);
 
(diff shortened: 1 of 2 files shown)


ipa-core/src/bin/report_collector.rs
```diff
@@ -151,6 +151,11 @@ enum ReportCollectorCommand {
         /// Number of records to aggregate
         #[clap(long, short = 'n')]
         count: u32,
+
+        // If set, use the specified fixed polling interval when running a query.
+        // Otherwise, use exponential backoff.
+        #[arg(long, default_value_t = 0)]
+        set_fixed_polling_ms: u64,
```
akoshelev, line 158: is `Option<u64>` better represents this being not set?

akoshelev (approved):
Setting interval in millis is not great but probably the best we can do without polling extra crates. Humans tend to make typo errors when entering large numbers and if we want to poll every 5 mins, 300000 vs 30000 makes a big difference

ipa-core/src/bin/report_collector.rs
tyurek: oh, yeah, probably

tyurek:
> Setting interval in millis is not great but probably the best we can do without polling extra crates. Humans tend to make typo errors when entering large numbers and if we want to poll every 5 mins, 300000 vs 30000 makes a big difference

Are we expecting to have large polling intervals like that?

PR #1493, opened 2024-12-12
Add an option to put helper logs into a file

akoshelev:
We need it because stderr is often dropped by the external services (like Kubernetes)

This also adds some verification for our e2e tests to make sure we write something into them

### DIFF
--- a/ipa-core/src/cli/verbosity.rs
+++ b/ipa-core/src/cli/verbosity.rs
@@ -45,7 +45,14 @@ impl Verbosity {
         let filter_layer = self.log_filter();
         info!("Logging setup at level {}", filter_layer);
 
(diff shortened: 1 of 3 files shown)


ipa-core/src/cli/verbosity.rs
```diff
@@ -20,6 +24,11 @@ pub struct Verbosity {
     /// Verbose mode (-v, or -vv for even more verbose)
     #[arg(short, long, action = clap::ArgAction::Count, global = true)]
     verbose: u8,
+
+    /// This option is mutually exclusive with console logging. If set,
```
cberkhoff, line 28: can we have both? K8s logs are easy to pull for many shards whereas going into each host is a hassle

PR #1494, opened 2024-12-12
Abort on panics

akoshelev:
See the comment for reasoning


cberkhoff (approved):

eriktaubeneck (approved):

andyleiserson:
Is there a reason not to set this in debug builds as well?

akoshelev:
> Is there a reason not to set this in debug builds as well?
no particular reason except maybe getting better observability if we can something really stupid (with unsafe) that can only be seen on function stack cleanup?

akoshelev:
I can add abort on panics for debug builds as separate PR if that's what we want

PR #1496, opened 2024-12-13
Cache DNS responses for IpaClients

akoshelev:
Our client is not a general-purpose HTTP client and talks to only limited subset of services, so caching will be effective here to prevent overloading DNS


akoshelev:
I am seeing test hangs when running them locally, so I don't think this is ready to be merged. I can't figure out why it may be caused by the cache (we only resolve localhost and I confirmed that it does not block on Dashmap) but it is always DNS...

akoshelev:
We don't need this change as we were able to fix the DNS issues on the infrastructure side with increased TTL

PR #1501, opened 2024-12-17
Create a buffered wrapper around BytesStream

akoshelev:
The need for it is driven by the behavior we're observing from Report Collector sending bytes down to individual shards. It writes data as it becomes available and Hyper does not accumulate it before sending. On the receiver side we are seeing chunks of size 1 received and that creates thrashing on sender/receiver side.

This change paves the path to use buffering on RC side.

### DIFF
--- a/ipa-core/src/helpers/transport/stream/buffered.rs
+++ b/ipa-core/src/helpers/transport/stream/buffered.rs
@@ -99,15 +99,18 @@ mod tests {
         num::NonZeroUsize,
         pin::Pin,
         sync::{Arc, Mutex},
+        task,
         task::Poll,
     };
 
     use bytes::Bytes;
     use futures::{stream::TryStreamExt, FutureExt, Stream, StreamExt};
+    use pin_project::pin_project;
     use proptest::{
         prop_compose, proptest,
         strategy::{Just, Strategy},
     };
+    use task::Context;
 
     use crate::{
         error::BoxError, helpers::transport::stream::buffered::BufferedBytesStream,
@@ -214,10 +217,10 @@ mod tests {
         chunk: usize,
     }
 
+    #[pin_project]
     struct FallibleTestStream {
-        total_size: usize,
-        remaining: usize,
-        chunk: usize,
+        #[pin]
+        inner: TestStream,
         error_after: usize,
     }
 
@@ -231,9 +234,11 @@ mod tests {
 
     fn fallible_stream(total_size: usize, chunk: usize, error_after: usize) -> FallibleTestStream {
         FallibleTestStream {
-            total_size,
-            remaining: total_size,
-            chunk,
+            inner: TestStream {
+                total_size,
+                remaining: total_size,
+                chunk,
+            },
             error_after,
         }
     }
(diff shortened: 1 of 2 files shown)


ipa-core/src/helpers/transport/stream/buffered.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    #[test]
+    fn success() {
+        run(|| async move {
+            verify_success(infallible_stream(11, 2), 3).await;
+            // verify_success(infallible_stream(12, 3), 3).await;
+            // verify_success(infallible_stream(12, 5), 12).await;
+            // verify_success(infallible_stream(12, 12), 12).await;
+            // verify_success(infallible_stream(24, 12), 12).await;
+            // verify_success(infallible_stream(24, 12), 1).await;
```
eriktaubeneck, line 125: nit: clean these up

eriktaubeneck (approved):

ipa-core/src/helpers/transport/stream/buffered.rs
akoshelev: good catch, I want to get them back

ipa-core/src/helpers/transport/stream/buffered.rs
```diff
@@ -0,0 +1,313 @@
...
+use pin_project::pin_project;
+
+use crate::helpers::BytesStream;
+
+/// An adaptor to buffer items coming from the upstream
+/// [`BytesStream`](BytesStream) until the buffer is full, or the upstream is
+/// done. This may need to be used when writing into HTTP streams as Hyper
+/// does not provide any buffering functionality and we turn NODELAY on
+#[pin_project]
+pub struct BufferedBytesStream<S> {
```
cberkhoff, line 19: a few nits

To me `BufferedBytesStream` doesn't tell why this stream is special. I think the key thing is the poll with is being "chunked". I would rather call this `BufferedChunkedStream`

why not have the trait bound `S: BytesStream` here as well? Thinking more about documentation/readability

ipa-core/src/helpers/transport/stream/buffered.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Inner stream to poll
+    #[pin]
+    inner: S,
+    /// Buffer of bytes pending release
+    buffer: Vec<u8>,
+    /// Number of bytes released per single poll.
+    /// All items except the last one are guaranteed to have
+    /// exactly this number of bytes written to them.
+    sz: usize,
```
cberkhoff, line 28: nit: chunk_size?

ipa-core/src/helpers/transport/stream/buffered.rs
```diff
@@ -0,0 +1,313 @@
...
+            let next_chunk = (0..next_chunk_size)
+                .map(|v| u8::try_from(v % 256).unwrap())
+                .collect::<Vec<_>>();
+
+            self.remaining -= next_chunk_size;
+            Poll::Ready(Some(Ok(Bytes::from(next_chunk))))
+        }
+    }
+
+    impl Stream for FallibleTestStream {
```
cberkhoff, line 261: nit: we could save some duplicated code by having TestStream and FallibleTestStream share poll_next.

you could make error_after be and option.

cberkhoff (approved):

ipa-core/src/helpers/transport/stream/buffered.rs
akoshelev: Bounds need to be repeated everywhere if put on the struct - generally we try to avoid that if that's not necessary

akoshelev: this does not attempt to chunk the inner stream, the whole purpose of it is to accumulate enough bytes (buffer) before sending them down for processing.

cberkhoff: But doesn't the output get chunked? This adapter is buffering and chunking... another name would be paginating.

I still think BufferedBytesStream is vague. My 2 cents.

PR #1502, opened 2024-12-17
Upgrade jemalloc to 0.6; add jemalloc feature

andyleiserson:
jemalloc is used by default on linux. The feature enables it on non-linux platforms.

The `stats` feature is enabled by default on version 0.5 of the `tikv-jemalloc-ctl` crate. It needs to be specified on version 0.6 of the crate. I figured might as well upgrade the jemalloc bits now, when I'm aware of the breaking change, than have to remember (or somebody else rediscover) it later.

The jemalloc profiling enables are a bit confusing. This PR does not have all of the config that I set to do memory profiling. My intent was to have it compiled-in-but disabled, but I'm not sure that's actually what this PR does. I will do a draft run to check whether there is a performance impact.

### DIFF
--- a/ipa-core/Cargo.toml
+++ b/ipa-core/Cargo.toml
@@ -63,7 +63,9 @@ enable-benches = ["cli", "in-memory-infra", "test-fixture", "criterion", "iai"]
 # of unit tests use it. Real world infra uses HTTP implementation and is suitable for integration/e2e tests
 in-memory-infra = []
 real-world-infra = []
-dhat-heap = ["cli", "test-fixture"]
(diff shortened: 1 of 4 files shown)


akoshelev:
+1 on doing a draft run before merging this

akoshelev (approved):

andyleiserson:
Draft run is at https://draft-mpc.vercel.app/query/view/briny-hand2024-12-17T0158.

andyleiserson:
Draft run took 1:51, no change from previous.

I also did some further investigation to convince myself that this PR really does have profiling disabled. Since I'm curious, I opened a new one (#1505) to see if there is a performance impact of turning it on.

PR #1504, opened 2024-12-17
Use stream buffering in report collector

akoshelev:
Continuation of #1501, we want to avoid submitting reports one by one from RC to each individual shard. That likely leads to fragmentation and we've been observing slow execution on the client side.

This sets up the buffer size to be divisible by TCP MSS, but I don't have any real evidence that this is going to work well. We would need to experiment with it

### DIFF
--- a/ipa-core/src/bin/report_collector.rs
+++ b/ipa-core/src/bin/report_collector.rs
@@ -5,7 +5,6 @@ use std::{
     fs::{File, OpenOptions},
     io,
     io::{stdout, BufReader, Write},
-    num::NonZeroUsize,
     ops::Deref,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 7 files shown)


ipa-core/src/bin/report_collector.rs
```diff
@@ -430,6 +431,9 @@ async fn hybrid(
     count: usize,
     set_fixed_polling_ms: Option<u64>,
 ) -> Result<(), Box<dyn Error>> {
+    // twice the size of TCP MSS. This may get messed up if TCP options are used which is not
+    // in our control, but hopefully fragmentation is not too bad
+    const BUF_SIZE: NonZeroUsize = NonZeroUsize::new(1072).unwrap();
```
akoshelev, line 436: @andyleiserson do you have any suggestion on picking up the buffer size?

andyleiserson (approved):

ipa-core/src/bin/report_collector.rs
andyleiserson: I think what's important is that the kernel has at least a full packet's worth of data, so that it can send full packets.

akoshelev:
I don't see any difference in performance on 1B records. I think the bottleneck is on the upload, so I'll hold off merging this

akoshelev:
I intend to merge it - there is nothing bad about this change and buffered writes from RC can still be useful for submitting medium-sized inputs

PR #1505, opened 2024-12-17
[DRAFT] Performance impact of jemalloc profiling

andyleiserson:


andyleiserson:
The draft run with heap profiling on (`prof:true,prof_active:true`) took 2:05, about a 10% slowdown. I'm going to do another with `prof:true,prof_active:false`.

andyleiserson:
The [run with `prof:true,prof_active:false`](https://draft-mpc.vercel.app/query/view/bogus-give2024-12-17T1900) took 1:52, no change from the baseline without these settings (or without the `profiling` feature entirely).

PR #1507, opened 2024-12-18
Split encryption script

eriktaubeneck:
script used to turn the newline delimited files into N length delimited files


scripts/split-encrypted-files.py
```diff
@@ -0,0 +1,71 @@
...
+def split_hex_file(input_filename, output_stem, num_files):
+    """
+    Reads in a file of hex strings, one per line, splits it up into N files,
+    and writes out each line as length-delimited binary data.
+
+    :param input_filename: The name of the input file containing hex strings.
+    :param num_files: The number of output files to split the input into.
+    """
+    output_files = [
+        open(f"{output_stem}_shard_{i:03d}.bin", "wb") for i in range(num_files)
```
akoshelev, line 23: curious how do you encode helper id here? we would need to have N files per helper

eriktaubeneck: `helper_id` is in the `output_stem`, since I already have `helper_{i}.enc` files. here is an example invocation:

```
python ../scripts/split-encrypted-files.py -i encryptions/1B_cat/helper1.enc -o encryptions/1B_cat/30_shards/helper1 -n 30
```

akoshelev (approved):

akoshelev (approved):

PR #1508, opened 2024-12-18
Query input from URL

andyleiserson:
Support for helper to pull query input from a URL, rather than receiving it directly from the client in an HTTP request body.

Also adds a simple HTTP server in test_mpc to serve local files, for testing purposes.


akoshelev (approved):
lets ship it - tests for test_mpc will be added in #1510

PR #1509, opened 2024-12-19
add flag to encrypt_util for length delimited binary format

eriktaubeneck:
In order to support the new ability for helpers to load data from a file/url, we need the ability to produce length delimited encrypted data for testing. This adds a flag to the `crypto_util` called `--length-delimited` which will produce length delimited binary instead of newline delimited hex.

The next step is to use this with the integration test. The flow will be

1. generate raw input data files (one for each shard)
2. use `crypto_util` to encrypt each of these for each shard (and each helper)
3. pass these files into the report collector (is this implemented yet @akoshelev ?)
4. cat all input data files together and run `in-the-clear`
5. compare results

### DIFF
--- a/ipa-core/src/cli/crypto/hybrid_encrypt.rs
+++ b/ipa-core/src/cli/crypto/hybrid_encrypt.rs
@@ -52,15 +52,38 @@ pub struct HybridEncryptArgs {
     /// Path to helper network configuration file
     #[arg(long)]
     network: PathBuf,
+    /// a flag to produce length delimited binary instead of newline delimited hex
+    #[arg(long)]
+    length_delimited: bool,
+}
+
+#[derive(Copy, Clone)]
+enum FileFormat {
+    LengthDelimitedBinary,
+    NewlineDelimitedHex,
 }
 
 impl HybridEncryptArgs {
     #[must_use]
-    pub fn new(input_file: &Path, output_dir: &Path, network: &Path) -> Self {
+    pub fn new(
+        input_file: &Path,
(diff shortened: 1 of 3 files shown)


ipa-core/src/cli/crypto/hybrid_encrypt.rs
```diff
@@ -52,15 +52,38 @@ pub struct HybridEncryptArgs {
     /// Path to helper network configuration file
     #[arg(long)]
     network: PathBuf,
+    /// a flag to produce length delimited binary instead of newline delimited hex
+    #[arg(long)]
+    length_delimited: bool,
```
andyleiserson, line 57: Instead of a boolean argument it might be clearer to have descriptive options (something like the `FileFormat` choices -- possibly clap can even be taught to parse into `FileFormat`).

andyleiserson (approved):

ipa-core/src/cli/crypto/hybrid_encrypt.rs
eriktaubeneck: I generally think of [docopt](http://docopt.org) as the canonical way to design CLIs, and this would probably fit as a mutually exclusive element better than an element with an arg. Clap probably has a way of defining mutually exclusive elements (and as you say, clap could likely be taught to coerce that into the enum.) This would be even more important if there were more than two options. 

I likely won't get to this before I'm out, so up to @andyleiserson and @akoshelev if you want to refactor before merging.

akoshelev (approved):

PR #1511, opened 2024-12-19
s3 presigned urls hot fix

eriktaubeneck:
update presigned url script expires-in flag to 7 days - max allowed by aws

### DIFF
--- a/scripts/presigned-s3-urls.sh
+++ b/scripts/presigned-s3-urls.sh
@@ -1,29 +1,42 @@
 #!/bin/bash
 
 # Set the usage message
-usage="Usage: $0 <dir_path> <s3_uri> <output_file>"
+usage="Usage: $0 <s3_uri> <output_file> [<expires_in>]"
 
 # Example invocation
-# from ipa/input_data_S02/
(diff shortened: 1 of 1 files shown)


akoshelev:
We spoke about it offline, but it would be great to use s3 bucket + prefix as the source for signed urls

akoshelev (approved):

scripts/presigned-s3-urls.sh
```diff
@@ -1,29 +1,39 @@
...
+s3_uri="$1"
+output_file="$2"
+# Check if the output file already exists
+if [ -f "$output_file" ]; then
+  echo "Error: Output file '$output_file' already exists. Please remove it before running this script."
+  exit 1
+fi
+
+# default expires_in: 7 days (7 * 24 * 60 * 60) - 1. this is the max allowed
+expires_in="${3:-604799}"
```
akoshelev, line 25: nit - can we use hours/days here? I don't think there is a reason for us to use seconds

akoshelev (approved):

scripts/presigned-s3-urls.sh
eriktaubeneck: original intention was to keep it as close to the aws api as possible, but we can do whatever. preference for hours vs days?

PR #1520, opened 2025-01-10
Avoid using errors to communicate the state of other shards to the leader during query-status API.

cberkhoff:


akoshelev (commented):
I think it is better than what we have right now, but would love @andyleiserson opinion on this. Particularly the changes to transport API

PR #1523, opened 2025-01-14
Remove OPRF IPA code

akoshelev:
As we migrated everything to Hybrid, the old code is just a copy-paste and carries some dead weight we should avoid. There is a branch [oprf-ipa](https://github.com/private-attribution/ipa/tree/oprf-ipa) that has everything if we ever need to pick up this code again. 

Unfortunately we made quite a bit of shortcuts and there is currently no benchmark code for Hybrid


ipa-core/src/report/ipa.rs
andyleiserson, line 1: This file is now empty, should be removed.

andyleiserson (approved):

Issue #1527, opened 2025-03-11
Security advisories RUSTSEC-2024-0436, RUSTSEC-2024-0437, RUSTSEC-2025-0014


andyleiserson:
I have waived these in the workflow definition file for now.

* [RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436): paste crate is unmaintained
* [RUSTSEC-2024-0437](https://rustsec.org/advisories/RUSTSEC-2024-0437): crash due to uncontrolled recursion in protobuf crate. This is used via opentelemetry crates. Tracked by upstream issue [opentelemetry-rust#2769](https://github.com/open-telemetry/opentelemetry-rust/issues/2769).
* [RUSTSEC-2025-0014](https://rustsec.org/advisories/RUSTSEC-2025-0014): humantime crate is unmaintained

PR #1528, opened 2025-03-11
Fixes for security audit

andyleiserson:
* Grant additional permissions to the job
* Update rcgen to 0.13
* Waive several vulnerabilities that are not critical and cannot immediately be fixed. Tracked by #1527.


akoshelev:
I've been looking at protobuf errors but didn't have time to fix them. Should be fairly easy with crate version pinning.

I plan to look again next week

akoshelev (approved):

PR #1529, opened 2025-03-15
Remove RUSTSEC-2025-0014 audit lint suppression

akoshelev:
https://rustsec.org/advisories/RUSTSEC-2025-0014

> This advisory has been withdrawn and should be ignored. It is kept only for reference.

https://github.com/tailhook/humantime/issues/31


akoshelev:
Partially addresses #1527

martinthomson (approved):

PR #1533, opened 2026-02-13
Upgrade to Rust 1.93

akoshelev:
Various clippy fixes


martinthomson (commented):
This all looks fine, but why isn't it building?  (Why bother being the other question...)

akoshelev:
> This all looks fine, but why isn't it building? (Why bother being the other question...)

nostalgic notes and want to keep this code alive just in case if I need to borrow any ideas from it

akoshelev:
sorry, coverage, not today

